PR #2465, opened 2023-01-30
MDEV-30492 Crash when use mariabackup.exe with config 'innodb_flush_m…

grooverdan:
…ethod=async_unbuffered'

- [x] *The Jira issue number for this PR is: MDEV-30492*

## Description

The srv_file_flush_method has already has its range validated in the parsing of the options.

As such the assertion is unnecessary, and harmful if a user sets a Windows only flush method like async_unbuffered.

Thanks Mitchell Lee for reporting the bug.

## How can this PR be tested?

mariabackup.exe with innodb_flush_method=async_unbuffered

## Backward compatibility

N/A

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -1896,6 +1896,17 @@ xb_get_one_option(int optid,
     break;
 
   case OPT_INNODB_FLUSH_METHOD:
+#ifdef _WIN32
+    /* From: storage/innobase/handler/ha_innodb.cc:innodb_init_params */
+    switch (srv_file_flush_method) {
+    case SRV_ALL_O_DIRECT_FSYNC + 1 /* "async_unbuffered"="unbuffered" */:
+      srv_file_flush_method= SRV_ALL_O_DIRECT_FSYNC;
+      break;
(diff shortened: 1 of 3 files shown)


vaintroub:
Turns out that's not enough, and the assertion is valid.

Unfortunately the number of values in srv_flush_t enum and the number of elements in innodb_flush_method_names  differ, and  setting innodb_flush_method to "async_unbuffered", or "normal" would go outside of srv_flush_t range. 

"async_unbuffered" corresponds to SRV_ALL_O_DIRECT_FSYNC (synonym to "unbuffered"). "normal"  should correspond to SRV_FSYNC.

Server does this mapping (https://github.com/MariaDB/server/blob/7a98d232e42b66efc759d584b05214e91681c346/storage/innobase/handler/ha_innodb.cc#L4217) in rather ugly manner, but mariabackup does not ,so it would end up with invalid innodb_flush_method

grooverdan:
Are the extra Windows types valid in MariaBackup? Do you have a preference on restricting its upper value or applying the same ugly hack?

vaintroub:
I'd prefer the same ugly hack. the values might or might not have the same meaning (in --prepare they have the same meaning, but in --backup less ,although some files are still opened with innodb)

LinuxJedi (approved):
Approved on the basis that this is what we do in InnoDB too.

PR #2466, opened 2023-01-30
MDEV-30509: mariadb-plugin-connect: introduce curl as recommends

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-30509*

## Description
in order to be able to retrieve files using REST queries. Otherwise,
`ERROR 1105 (HY000): Curl not installed.` will be thrown.

This is upstreaming: [Debian Salsa MariaDB team MR 25](https://salsa.debian.org/mariadb-team/mariadb-server/-/merge_requests/25)

## How can this PR be tested?
Use plugin connect without curl and it will fail. Then install curl and it works

## Backward compatibility
Does not provide backward compatibility problems but solves user problem.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -525,6 +525,7 @@ Depends: libxml2,
          unixodbc,
          ${misc:Depends},
          ${shlibs:Depends}
+Recommends: curl
 Breaks: mariadb-connect-engine-10.1,
         mariadb-connect-engine-10.2,
         mariadb-connect-engine-10.3


illuusio:
Requested CLA writing from original author.

smohaorg:
I signed the CLA. Thank you @illuusio for this PR.

grooverdan:
Thanks @smohaorg , @illuusio . Its just missed the release cut < 10.11 so will probably be the later releases. Might still make 10.11+ as that's schedule for two weeks time.

PR #2467, opened 2023-01-30
MDEV-30502: Debian 12 deprecates package lsb-base and favor sysvinit-utils package

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-30502*

## Description
Debian 12 deprecates package lsb-base and favor sysvinit-utils package

Files:
 * support-files/mysql.server.sh
 * debian/mariadb-server-10.5.mariadb.init

Relies on LSB script file `/lib/lsb/init-functions` which has been moved from `lsb-base` to `sysvinit-utils` package

To maintain backward compability use `lsb-base` on older distributions likde Bookwork (11) and Buster (10) where `/lib/lsb/init-functions` is still in `lsb-base`.

## How can this PR be tested?
On Salsa-CI:
https://salsa.debian.org/illuusio/mariadb-server/-/pipelines/492495 (Which is against 10.5 and this is against 10.3 where it first happens)

Install current version on Debian 12/Sid and sysvinit scripts should work. Without they fail because of lack of ` /lib/lsb/init-functions`

## Backward compatibility
Should maintain backward compatibility


ottok:
That is not needed, you can just simply drop the lsb-base, right?

See
* https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/374f34cd1e973d52439aaa292123cc3c852dacb5
* https://salsa.debian.org/mariadb-team/galera-4/-/commit/696391572958f3f930e2d41e59c6002ff05fcc5a

illuusio:
It's still needed older distributions like Debian 11 and below. if those scripts are updated then it can be dropped. You still need `sysvinit-utils` if `mariadb-server-1*.mariadb.init` is in use as it requires `/lib/lsb/init-functions.sh` it only uses them for logging so I don't know should them be done like in `mysql.server.sh` that if it's not found they are just exposed as functions.

illuusio:
It seems removing had been done in a4cc6fb91f3855e0ed803a6e1762440cfcf8cb5c

illuusio:
I'll rebase this on 10.4 which has this init script moved to another place.

illuusio:
I've superseded this with PR #2501 so closing this to favor that.

PR #2468, opened 2023-01-31
MDEV-29959: UUID Sorting for UUID version <= 5

grooverdan:
UUIDv6 and v7 provide draft versopms for UUIDs that are in an already sorted format.

Applying the v1 method of sort order isn't sensible on this.

As v1 -> v5 are already defined by standard and assumed to be in use, the same packing of those is used for v1 to avoid a migration requirement.

v6+ versions can be stored without changing the order. Its possible because of the way its stored and constraints around the variant version (!=0) in version 6+ that these can be differenciated despite the swapped positions of version and variant in recorded version.

TODO on this is an upgrade process
(TABLE_SHARE::init_from_binary_frm_image)? that transforms existing v6+ versions in tables to an unswapped form.

## Description
TODO: fill description here

## How can this PR be tested?

## Backward compatibility


grooverdan:
> @grooverdan , do I understand correctly that after accepting this PR,

Its far from certain that this will be accepted.

I'll need to reject v8 variant 0 and come up with a migration so that data in existing table, like your work around, get migrated.

> it will be possible to store uuid v7 in a column with the uuid type and mariadb will sort these records correctly?

by not doing anything special on v6+. Yes.

> At the moment I use uuid v7 and before saving to the database I generate uuid v7 and then change the value myself so that it is sorted correctly.
> 
> Example code php 8.1
> 
> ```
> $uuid = Uuid::uuid7()->toString()

I'll try to test inserting this form. Thanks for the ref, my php programming is horrible.

> 
> $sortableUuid = substr($uuid, 24, 8)
>             . "-"
>             . substr($uuid, 32, 4)
>             . "-"
>             . substr($uuid, 19, 4)
>             . "-"
>             . substr($uuid, 14, 4)
>             . "-"
>             . substr($uuid, 0, 8)
>             . substr($uuid, 9, 4); 
> ```

grooverdan:
> @grooverdan It seems to me that until **lts** is accepted, you can not waste time transferring crutches like mine.

Yep, I misspoke, I wasn't fixing your crutches more than re-storing existing (mangled) v6+ uuids in the tables.

> The main thing is that mariadb should sort uuid v6+ correctly from the very beginning, then i won't have to invent such crutches :)
> 
> In the previous example, if you want to reproduce it, you will need to install this library https://github.com/ramsey/uuid

Thanks.

grooverdan:
Probably. :cry: , thanks for putting a good case forward.

someniatko:
> now everyone has to live with it

I see how you can solve it in a way without BC break: just add some sort of parameter to UUID type e.g. `UUID(rearrange=true)`, dunno, which defaults to `false` in new UUIDs, but for existing databases it's `true` to prevent breaking existing data.

vuvova:
that's fairly easy to do without a parameter and with almost no compatibility issues. The problem here is the code structure, to fit this into current type plugin architecture requires some non-trivial refactoring. More in the MDEV.

PR #2469, opened 2023-02-01
MDEV-23384 Server crashes in subselect_indexsubquery_engine::print with optimizer trace enabled

rex-johnston:
Pull request created in: https://jira.mariadb.org/browse/MDEV-23384

### DIFF
--- a/sql/item_subselect.cc
+++ b/sql/item_subselect.cc
@@ -4588,10 +4588,28 @@ void subselect_indexsubquery_engine::print(String *str,
   str->append(STRING_WITH_LEN("<index_lookup>("));
   tab->ref.items[0]->print(str, query_type);
   str->append(STRING_WITH_LEN(" in "));
-  str->append(tab->table->s->table_name.str, tab->table->s->table_name.length);
-  KEY *key_info= tab->table->key_info+ tab->ref.key;
-  str->append(STRING_WITH_LEN(" on "));
-  str->append(&key_info->name);
+  if (tab->table)
+  {
+    str->append(tab->table->s->table_name.str, tab->table->s->table_name.length);
+    KEY *key_info= tab->table->key_info+ tab->ref.key;
+    str->append(STRING_WITH_LEN(" on "));
+    str->append(&key_info->name);
+  }
+  else
+  {
+    // mysql_derived_fill() has removed tab->table, print out something from
+    // *tab that is more informative...
+    if( tab->tab_list && tab->tab_list->derived )
+    {
+      String dstring;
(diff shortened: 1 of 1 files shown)


sql/item_subselect.cc
```diff
@@ -4588,10 +4588,28 @@ void subselect_indexsubquery_engine::print(String *str,
   str->append(STRING_WITH_LEN("<index_lookup>("));
   tab->ref.items[0]->print(str, query_type);
   str->append(STRING_WITH_LEN(" in "));
-  str->append(tab->table->s->table_name.str, tab->table->s->table_name.length);
-  KEY *key_info= tab->table->key_info+ tab->ref.key;
-  str->append(STRING_WITH_LEN(" on "));
-  str->append(&key_info->name);
+  if (tab->table)
```
spetrunia, line 4591: Why can't we always get the TABLE* from `tab->tab_list->table` ?  I don't see where this won't be possible... 
(Perhaps, JOIN_TAB objects representing semi-join nests or aggregation nodes have tab->tab_list==NULL, but if we are in subselect_indexsubquery_engine or in a uniquesubquery_engine, it means the join had just one table and we are looking at the JOIN_TAB representing it... and not semi-join nest or aggregation node)

spetrunia (changes requested):
* Please add a testcase.
* Please fix also subselect_uniquesubquery_engine::print(). See the testcase in https://jira.mariadb.org/browse/MDEV-23384.
* Please change to use `tab->tab_list->table` (in all cases)
** Near  its use, please add a comment that the JOIN object might already have JOIN::destroy() called for it which might have set tab->table=NULL.  Other fields of JOIN_TAB are still intact so we can use them.

rex-johnston:
> * Please add a testcase.
> 
>     * Please fix also subselect_uniquesubquery_engine::print(). See the testcase in https://jira.mariadb.org/browse/MDEV-23384.
> 
>     * Please change to use `tab->tab_list->table` (in all cases)
>       ** Near  its use, please add a comment that the JOIN object might already have JOIN::destroy() called for it which might have set tab->table=NULL.  Other fields of JOIN_TAB are still intact so we can use them.

Item_in_subselect::setup_mat_engine
   subselect_hash_sj_engine::make_unique_engine() 

when initializing a new subselect_uniquesubquery_engine sets up a JOIN_TAB structure that is largely uninitialized, except for tab->table and tab->preread_init_done.  tab->tab_list remains unset and can cause a SEGV when printed out with subselect_uniquesubquery_engine::print.

I can do a memset on the structure and/or initialize it further, but it would only be used for the above.

PR #2470, opened 2023-02-01
MDEV-30534: Remove EOL Debian version 9 (stretch) from autobake-deb.sh

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-30534*

## Description
Debian 9 has EOL July 6th, 2020. This PR cleans it up from `debian/autobake-deb.sh` which is used to build official versions of MariaDB

## How can this PR be tested?
As this PR removes Debian 9 from building script only way to test this is try to build this on Debian 9 and see it failing 

## Backward compatibility
After merging this there is no official way to build it on Debian 9. One can still build it manually.


ottok (approved):
LGTM

PR #2471, opened 2023-02-01
MDEV-30503: Fix Breaks and Replaces on Debian Sid

illuusio:

## Description
Some files has been moved on debian/latest MariaDB (which is 10.11) and it makes `Breaks/Replaces` not working correctly

## How can this PR be tested?
Please see Salsa-CI
https://salsa.debian.org/illuusio/mariadb-server/-/jobs/3880590
and original version:
https://salsa.debian.org/illuusio/mariadb-server/-/jobs/3871617

## Backward compatibility
Adds backward compability more as there is possibility to install top of MariaDB which is on Debian 12


ottok (changes requested):
I think this is the wrong approach. Now with Debian sid having already MariaDB 10.11, we should no longer have CI that tries to install anything older than 10.11 on Sid. That is by design not going to work or be very brittle.

Better solution would be to adapt the 10.6 CI to not target Debian Sid anymore, but just make sure that 10.6 works on Debian Bullseye and Ubuntu Kinetic and older.

illuusio:
> I think this is the wrong approach. Now with Debian sid having already MariaDB 10.11, we should no longer have CI that tries to install anything older than 10.11 on Sid. That is by design not going to work or be very brittle.
> 
> Better solution would be to adapt the 10.6 CI to not target Debian Sid anymore, but just make sure that 10.6 works on Debian Bullseye and Ubuntu Kinetic and olde

illuusio:
Wrong button sorry... Have to reiterate this little bit

PR #2472, opened 2023-02-01
MDEV-30554 RocksDB risc64 libatomic linking on rocksdb tools missing: was: Update build_rocksdb.cmake

LocutusOfBorg:
Do not erase ${SYSTEM_LIBS} variable when setting it, this makes the latomic flag set for riscv64 lost

this is the current code
```
if(CMAKE_SYSTEM_PROCESSOR STREQUAL "riscv64")
 set(SYSTEM_LIBS ${SYSTEM_LIBS} -latomic)
endif()
find_package(Threads REQUIRED)
if(WIN32)
  set(SYSTEM_LIBS ${SYSTEM_LIBS} Shlwapi.lib Rpcrt4.lib)
else()
  set(SYSTEM_LIBS ${CMAKE_THREAD_LIBS_INIT} ${LIBRT} ${CMAKE_DL_LIBS})
endif()
```

So, in case of not WIN32, the SYSTEM_LIBS with latomic value is just lost. We should keep it to fix riscv64 build failure
https://launchpadlibrarian.net/648316103/buildlog_ubuntu-lunar-riscv64.mariadb-10.6_1%3A10.6.11-2_BUILDING.txt.gz


LocutusOfBorg:
@ottok I'm uploading the change in Ubuntu, feel free to upload in Debian if you agree please :)

grooverdan:
@LocutusOfBorg thanks very much for highlighting the obvious flaw and digging up the University of Illinois check.

It seems from the build above that the amd64 Centos 7 tries to include libatomic and it doesn't exist.

As it turns out the top level configure.cmake already does a similar check.

Do you mind checking #2477.

LocutusOfBorg:
@ottok we can sync as soon as you upload again in Debian
Actually I uploaded the patch to mariadb-10.6 to later discover that it can't build anymore because mariadb has higher version. So my upload is just lost.

ottok:
Thanks @LocutusOfBorg for your help. Could you please also review #2477 or any of the open MRs at https://salsa.debian.org/mariadb-team/mariadb-server/-/merge_requests to help the build issues and Debian packaging improve faster?

There is also a s390x test error visible on Debian builds at https://buildd.debian.org/status/package.php?p=mariadb which did not reproduce on Launchpad builds at https://launchpad.net/~mysql-ubuntu/+archive/ubuntu/mariadb-10.11/+builds?build_text=&build_state=all which I could use some help on. Merge Requests directly on the Debian (and Ubuntu) packaging welcome so I can do the next upload as soon as possible.

LocutusOfBorg:
Started 7 hours ago
Finished 7 minutes ago (took 7 hours, 26 minutes, 31.1 seconds)

grooverdan:
merged #2477 instead. Thanks for testing.

PR #2473, opened 2023-02-01
MDEV-30334 Optimizer trace produces invalid JSON with WHERE subquery

rex-johnston:
Pull request created in: https://jira.mariadb.org/browse/MDEV-30334


sql/sql_select.cc
```diff
@@ -27865,6 +27865,8 @@ void st_select_lex::print(THD *thd, String *str, enum_query_type query_type)
       }
     }
     str->append(" */ ");
+    if (join && join->cleaned)
+      return;
```
spetrunia, line 27869: This looks a little bit cryptic, considering the same logic below.

Please add a comment saying something like:
```
For selects with join->cleaned==true the code below will only print the 
select number. We've already printed it, so exit.
 ```

File: sql/sql_select.cc [hunk 2]
```diff
@@ -27865,6 +27865,8 @@ void st_select_lex::print(THD *thd, String *str, enum_query_type query_type)
...
   }
```
spetrunia, line 27870: Please add a testcase to the patch.

An interesting observation is that the "Verify we're writing valid JSON" feature didn't catch this.  Apparently it only checks the JSON object/array structures, but not the strings...  if we get any other cases like this, it might make sense to add verification..  but this is outside of the scope of this MDEV.

spetrunia (changes requested):

sql/sql_select.cc
rex-johnston: Added MDEV-30650 for future use.  Are we sticking to standard ASCII (space to ~) for these strings?

spetrunia: Good. 
No, it is Unicode.  Currently UTF8mb3:
```
MariaDB [test]> show create table information_schema.optimizer_trace\G
       Table: OPTIMIZER_TRACE
Create Table: CREATE TEMPORARY TABLE `OPTIMIZER_TRACE` (
  `QUERY` longtext NOT NULL,
  `TRACE` longtext NOT NULL,
  `MISSING_BYTES_BEYOND_MAX_MEM_SIZE` int(20) NOT NULL,
  `INSUFFICIENT_PRIVILEGES` tinyint(1) NOT NULL
) ENGINE=Aria DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci PAGE_CHECKSUM=0
1 row in set (0.004 sec)
```
(maybe we should change that to be UTF8MB4 in the next versions...)

Some characters need to be escaped, according to https://www.json.org/json-en.html, see the production for String.

spetrunia (approved):

iangilfillan:
@mariadb-RexJohnston  I presume this PR can be closed, as https://jira.mariadb.org/browse/MDEV-30334 was closed in July?

PR #2474, opened 2023-02-01
innodb: cmake - sched_getcpu removed - not used

grooverdan:
basic cleanup.


LinuxJedi (approved):
Easy one to grep, only RocksDB appears to use it.

PR #2475, opened 2023-02-02
MDEV-30498 Rename mysql_upgrade state file to mariadb_upgrade

christian7007:
- [x] *The Jira issue number for this PR is: MDEV-30498*

## Description
This PR includes the changes for renaming the upgrade state file from `mysql_upgrade_file` to `mariadb_upgrade_file` and ensures that the old file is automatically removed when `mariadb-upgrade` tool is executed.

## How can this PR be tested?

MTR tests referencing old `mysql_upgrade_file` have been updated with the new file name. Also, a new test case have been added on `main.mysql_upgrade` to test that the old file is removed after running `mariadb-upgrade`.

- [X] *This is a new feature and the PR is based against the latest MariaDB development branch*

## Backward compatibility
As the file will be renamed in a major version, there shouldn't be any backward compatibility issue, other than removing the old file and using the new name instead.

## Copyright

### DIFF
--- a/client/mysql_upgrade.c
+++ b/client/mysql_upgrade.c
@@ -700,7 +700,7 @@ static int get_upgrade_info_file_name(char* name)
 
   dynstr_free(&ds_datadir);
 
-  fn_format(name, "mysql_upgrade_info", name, "", MYF(0));
+  fn_format(name, "mariadb_upgrade_info", name, "", MYF(0));
   DBUG_PRINT("exit", ("name: %s", name));
   DBUG_RETURN(0);
 }
@@ -709,7 +709,7 @@ static char upgrade_info_file[FN_REFLEN]= {0};
 
 
 /*
-  Open or create mysql_upgrade_info file in servers data dir.
+  Open or create mariadb_upgrade_info file in servers data dir.
 
   Take a lock to ensure there cannot be any other mysql_upgrades
   running concurrently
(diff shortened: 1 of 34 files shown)


grooverdan (approved):
Very nice! Thank you.

Minor question (partly towards other reviewers), is if `mysql_upgrade_info` should be removed at the end of `open_mysql_upgrade_file`, or `finish_mysql_upgrade_info_file`, but overall I don't think it matters that much, even where it is is probably ok.

One C++ style comment, but another has existed the mysql_upgrade.c for 4 years without complaint.

christian7007:
Thanks for the fast review!

Yes, I guess we can remove the file at the end too. As you said, probably it doesn't matter, but if you or any other reviewers think that it would be better to remove the file at the end, just let me know and I'll update the PR, it should be an easy change. Also, let me know if you prefer me to change that comment format.

grooverdan:
From @elenst in [MDEV](https://jira.mariadb.org/browse/MDEV-30498?focusedCommentId=249827&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-249827) can you update the error messages to `mariadb-upgrade`. I suspect a backport would conflict a log with the mtr test results you've changed here so 11.0 is fine.

debian/mariadb-server.README.Debian also needs a mysql_upgrade -> mariadb-upgrade text change.

And debian/additions/mariadb.conf.d/50-mysql-clients.cnf to change to the mariadb groups if you have time.

christian7007:
Hi @grooverdan, I have no problem adding those changes, but please note that they refer to the rename of the `mariadb-upgrade` tool not the upgrade state file. I'm mentioning this just in case something else was missing when the `mariadb-upgrade` tool was renamed.

grooverdan:
Yep, so a MDEV-29582 commit for the renames of executable on top of this branch as the test result will conflict otherwise.

grooverdan:
Thank you.

LinuxJedi (approved):
Great work :)

PR #2479, opened 2023-02-03
MDEV-30448 No deprecation message shown for mysql_fix_extensions

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-30448*

## Description

This is a myisam script and we encourage the use of Aria. So we deprecate it.

Its useful to deprecate it as carrying large amount of low utility scripts makes it hard to determine what is useful and what isn't in the packaging.

## How can this PR be tested?

Run it.


LinuxJedi (approved):
Approving with the caveat that if this is determined to be useful it should probably be replaced with a BASH one-liner in documentation. It would only be someone familiar with a POSIX OS doing this anyway.

an3l (approved):

PR #2484, opened 2023-02-07
Deb: MariaDB naming

grooverdan:
May as well get some remaining bits of the naming right.

### DIFF
--- a/debian/mariadb-server.README.Debian
+++ b/debian/mariadb-server.README.Debian
@@ -1,5 +1,5 @@
-* MYSQL WON'T START OR STOP?
-============================
+* MARIADB WON'T START OR STOP?
+==============================
 
 The most common reasons the server does not start are:
 - AppArmor is enforced and something is wrong with the confinement profile.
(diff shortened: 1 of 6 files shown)


illuusio:
Is there reasoning behind this renaming stuff?

grooverdan:
Because we're not MySQL.

debian/mariadb-server.README.Debian
```diff
@@ -84,14 +84,14 @@ MariaDB in Debian is secure by default, because:
 - There is no root account with password anymore. The system admin needs to
   create one themselves if they need it. With no password, all issues related
   to password management and password leaking are gone. Sysadmins can access
-  the database without a password simply by running 'sudo mysql' thanks to
+  the database without a password simply by running 'sudo mariadb' thanks to
   socket based authentication, which detects the system root user and allows
-  them to use the mysqld console as the mysql root user. For details see
+  them to use the mariadb console as the MariaDB root user. For details see
```
Weijun-H, line 89: How do we choose to use MariaDB or mariadb?

grooverdan: "mariadb" is the console application formerly known as "mysql". "MariaDB root user" is referring to the user `root` inside the MariaDB server.

grooverdan:
@ottok , hi, this has been waiting for a review for 3 weeks. Is it possible to get a review (even if it needs to go to 11.0 now).

ottok (approved):
Any particular reason why this is split across 3 separate commits? One seems would be enough.

The changes make sense and look correct.

I was hoping to have time to push this on Salsa-CI to hunt for potential regressions, but I ran out of time to do that, so I am approving now just based on reading the changes.

grooverdan:
Thanks. 3 commits was just for an easy revert if one was a bad idea. I'll squash when I rebase.

PR #2485, opened 2023-02-08
MDEV-30543 Max_used_connections_time resolved

urmilkalaria:
- [x] *The Jira issue number for this PR is: MDEV-30543*

## Description
TODO: fill description here
This PR is trying to add a variable Max_used_connections_time that calculates the current date and time and stores in the prescribed format in the JIRA issue template.
This code doesn't mess with the existing code so, safe to go ;)

## How can this PR be tested?

## Backward compatibility

## PR quality check

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -166,6 +166,7 @@ extern "C" {					// Because of SCO 3.2V4.2
 #endif /* _WIN32 */
 
 #include <my_libwrap.h>
+#include <ctime>
 
 #ifdef _WIN32 
 #include <crtdbg.h>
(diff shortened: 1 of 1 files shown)


grooverdan:
Hi. Nice attempt.

To add a system status variable:
* See the [Slave_heartbeat_period](https://github.com/MariaDB/server/blob/cafba8761af55ae16cc69c9b53a341340a845b36/sql/mysqld.cc#L7519) / and its function [show_heartbeat_period](https://github.com/MariaDB/server/blob/cafba8761af55ae16cc69c9b53a341340a845b36/sql/mysqld.cc#L6962) that populates `buf` with the required string. As a pattern for definition, an population of the result.
* use `get_date` (`mysys/mf_getdate.c`) as the string writing function.
* use `static time_t max_used_connections_time` and `my_time(0);` to record the latest at the location you've added your current code.

There's some `max_used_connections` tests in mysql-test/main/status.test which you're encouraged to attempt extending. This can be run with `mysql-test/mtr main.status`. To record add `---record`. The test should have the same static output each test run so recording into a user variable and ensuring an increase would be one way. If you have troubles, do your best and we probably can refine the test.

vuvova:
already implemented in #2540

PR #2486, opened 2023-02-09
MDEV-30621: Türkiye is the correct current country naming

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-30621*

## Description
As requested to the UN the country formerly known as Turkey is to be refered to as Türkiye.

## How can this PR be tested?

Tests updated

This is a bug against the pre-GA version.

## Backward compatibility

Ideally it should be backported. But feedback welcome on the compatibility of such a change would be appreciated.

## PR quality check
- [ X ] I checked the [CODING_STANDARDS.md](CODING_STANDARDS.md) file and my PR conforms to this where appropriate.


sql/win_tzname_data.h
```diff
@@ -69,7 +69,7 @@
 {L"Libya Standard Time","Africa/Tripoli"},
 {L"Namibia Standard Time","Africa/Windhoek"},
 {L"Arabic Standard Time","Asia/Baghdad"},
-{L"Turkey Standard Time","Europe/Istanbul"},
+{L"Türkiye Standard Time","Europe/Istanbul"},
```
vaintroub, line 72: Nope. Windows does not have "Türkiye Standard Time" time zone name. It has "Turkey Standard Time".
image

Besides, that file is generated from Unicode consortium data

PR #2494, opened 2023-02-12
MDEV-24538: JSON_LENGTH does not return error upon wrong number of parameters

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-24538*

## Description
JSON_LENGTH does not return error upon wrong number of parameters
## How can this PR be tested?
The test is 'json_length.test'.

## PR quality check


grooverdan:
Hi @Weijun-H. Thanks for the patch.

To test for an error put `--error ER_WRONG_PARAMCOUNT_TO_NATIVE_FCT` in the test file before the SQL statement that results this error. Use `mysql-test/mtr --record {testname}` to record the result file to the expected results of the test. Add the result file to the commit.

Also rather than create a new test file, append to an existing test file that also is using JSON_LENGTH like mysql-test/main/func_int.test or mysql-test/main/func_json.test.

On the test itself, 0 arguments is also invalid so may as well error for that too.

grooverdan:
Good work. I just rebased it to 10.4 as merged into the 10.3 are restricted to give extra review to 10.3 merges.

grooverdan (approved):

PR #2495, opened 2023-02-12
MDEV-30412: JSON_OBJECTAGG doesn't escape double quote in key

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-30412*

## Description
JSON_OBJECTAGG doesn't escape double quote in key
## How can this PR be tested?
Append the tests to `func_json.test`

## PR quality check

### DIFF
--- a/sql/item_jsonfunc.cc
+++ b/sql/item_jsonfunc.cc
@@ -4060,7 +4060,7 @@ bool Item_func_json_objectagg::add()
     result.append(", ");
 
   result.append("\"");
-  result.append(*key);
+  st_append_escaped(&result,key);
   result.append("\":");
 
   buf.length(0);
--- a/mysql-test/main/func_json.result
+++ b/mysql-test/main/func_json.result
@@ -1623,5 +1623,17 @@ id	doc
 {"$oid":"611c0a463b150154132f6636"}	{ "_id" : { "$oid" : "611c0a463b150154132f6636" }, "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : [ { "a" : 1.0 } ] } ] } ] } ] } ] } ] } ] } ] } ] } ] } ] } ] } ] } ] } ] }
 DROP TABLE arrNestTest;
 #
+# MDEV-30412 JSON_OBJECTAGG doesn't escape double quote in key
+#
+SELECT JSON_OBJECTAGG('"', 1);
+JSON_OBJECTAGG('"', 1)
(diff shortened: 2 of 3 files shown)


grooverdan:
I suspect an existing function exists already for this as just escaping the quote key isn't the complete problem. The backslash would also need to be escaped correctly in the existing string potentially?

LinuxJedi:
In addition to @grooverdan's message, at the moment this doesn't compile in debug builds due to this warning escalated to an error:
```
/home/buildbot/amd64-ubuntu-2004-debug/build/sql/item_jsonfunc.cc: In member function ‘virtual bool Item_func_json_objectagg::add()’:
/home/buildbot/amd64-ubuntu-2004-debug/build/sql/item_jsonfunc.cc:4063:20: error: comparison of integer expressions of different signedness: ‘int’ and ‘uint32’ {aka ‘unsigned int’} [-Werror=sign-compare]
 4063 |   for (int i= 0; i < key->length(); i++)
      |                  ~~^~~~~~~~~~~~~~~
```

sql/item_jsonfunc.cc
```diff
@@ -4060,7 +4060,7 @@ bool Item_func_json_objectagg::add()
     result.append(", ");
 
   result.append("\"");
-  result.append(*key);
+  result.append_for_single_quote(key->ptr(), key->length());
```
grooverdan, line 4063: Except its not single quotes.

Found `st_append_escaped` that looks suitable. As `StringBuffer` is a subclass of `String`, `&buf` still can be used.

sql/sql_string.cc
```diff
@@ -1152,6 +1152,7 @@ bool String::append_for_single_quote(const char *st, size_t len)
     case '\n':   APPEND(STRING_WITH_LEN("\\n"));
     case '\r':   APPEND(STRING_WITH_LEN("\\r"));
     case '\032': APPEND(STRING_WITH_LEN("\\Z"));
+    case '\"':   APPEND(STRING_WITH_LEN("\\\""));
```
grooverdan, line 1155: Technically single quotes can take a double quote unescaped.

grooverdan (changes requested):

sql/item_jsonfunc.cc
Weijun-H: Yes, it is a better approach.

grooverdan (approved):

grooverdan:
Look good. Thanks. Tested with more options too an works as expected:

```
+SELECT JSON_OBJECTAGG('\t\'\0\b\n\r\Z', 1);
+JSON_OBJECTAGG('\t\'\0\b\n\r\Z', 1)
+{"\t'\u0000\b\n\r\u001A":1}
```

@Weijun-H , I'm really enjoying seeing these JSON bugs being fixed. Much appreciated.

PR #2496, opened 2023-02-13
MDBF-534: Coverity scan: fix client folder

an3l:
- [x] *The Jira issue number for this PR is: MDBF-534*

## Description
The purpose of this patch is to address scan by coverity on MariaDB client folder.
There are couple of commits for one or more files in client folder.
Each commit is addressing specific coverity status on found defect in specific file or files.

## How can this PR be tested?
Check commit message and check coverity scan.

### DIFF
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -3169,7 +3169,7 @@ int main(int argc, char** argv)
   /* Set delimiter back to semicolon */
   if (retval != ERROR_STOP)
   {
-    if (!stop_event_string.is_empty())
+    if (!stop_event_string.is_empty() && result_file)
       fprintf(result_file, "%s", stop_event_string.ptr());
     if (!opt_raw_mode && opt_flashback)
       fprintf(result_file, "DELIMITER ;\n");
--- a/client/mysqldump.c
+++ b/client/mysqldump.c
@@ -2480,7 +2480,10 @@ static uint dump_events_for_db(char *db)
       /* Get database collation. */
 
       if (fetch_db_collation(db_name_buff, db_cl_name, sizeof (db_cl_name)))
+      {
+        mysql_free_result(event_list_res);
         DBUG_RETURN(1);
+      }
     }
 
     if (switch_character_set_results(mysql, "binary"))
@@ -3262,7 +3265,10 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
       if (path)
       {
         if (!(sql_file= open_sql_file_for_table(table, O_WRONLY)))
+        {
+          mysql_free_result(result);
           DBUG_RETURN(0);
+        }
         write_header(sql_file, db);
       }
 
@@ -3663,7 +3669,7 @@ static int dump_triggers_for_table(char *table_name, char *db_name)
   char       name_buff[NAME_LEN*4+3];
   char       query_buff[QUERY_LENGTH];
   uint       old_opt_compatible_mode= opt_compatible_mode;
-  MYSQL_RES  *show_triggers_rs;
+  MYSQL_RES  *show_triggers_rs= NULL;
   MYSQL_ROW  row;
   FILE      *sql_file= md_result_file;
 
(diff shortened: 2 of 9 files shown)


client/mysqltest.cc
```diff
@@ -3630,9 +3630,12 @@ void do_system(struct st_command *command)
 /* returns TRUE if path is inside a sandbox */
 bool is_sub_path(const char *path, size_t plen, const char *sandbox)
 {
-  size_t len= strlen(sandbox);
-  if (!sandbox || !len || plen <= len || memcmp(path, sandbox, len - 1)
-      || path[len] != '/')
+  size_t len;
+  if (!sandbox)
+    return false;
+  else
```
cvicentiu, line 3636: No need for this else statement.

an3l: Thanks @cvicentiu  :).

LinuxJedi (changes requested):
Test is currently failing due to a crash in mariadb-dump:

```
#0  mysql_free_result (result=0xd8) at /home/buildbot/amd64-centos-7/build/libmariadb/libmariadb/mariadb_lib.c:598
598	    if (result->handle && result->handle->status == MYSQL_STATUS_USE_RESULT)
#0  mysql_free_result (result=0xd8) at /home/buildbot/amd64-centos-7/build/libmariadb/libmariadb/mariadb_lib.c:598
#1  0x000055cbaaa47c24 in dump_table (table=<optimized out>, db=db@entry=0x7ffc7b460d04 "test", hash_key=hash_key@entry=0x0, len=len@entry=0) at /home/buildbot/amd64-centos-7/build/client/mysqldump.c:4656
#2  0x000055cbaaa3e5af in dump_selected_tables (tables=<optimized out>, table_names=<optimized out>, db=0x7ffc7b460d04 "test") at /home/buildbot/amd64-centos-7/build/client/mysqldump.c:6140
#3  main (argc=2, argv=0x55cbabc35988) at /home/buildbot/amd64-centos-7/build/client/mysqldump.c:7265
```

This is likely the new free of `res` in `dump_table` crashing due to not being initialised to `NULL` at the top of the function. Could also be a double-free, but I couldn't spot that in a quick scan. GDB should spot this with a debug build to be sure.

client/mysqldump.c
```diff
@@ -3994,7 +3998,7 @@ static int dump_triggers_for_table(char *table_name, char *db_name)
 done:
   if (path)
     my_fclose(sql_file, MYF(0));
-
+  mysql_free_result(show_triggers_rs);
```
LinuxJedi, line 4001: This is another case that could cause a crash. Consider this code above:

```c
 if (mysql_query_with_error_report(mysql, &show_triggers_rs, query_buff))
    goto done;
```
This will not fill `show_triggers_rs` with anything if the result is not `0`. `show_triggers_rs` is not initialised, so there is a case now where a failure here could jump to `done` and run `mysql_free_result` on uninitialised memory.

LinuxJedi (approved):
I think this is all good now, although I think it should maybe be against 10.4 instead of 10.11.

@cvicentiu agreed?

client/mysqlbinlog.cc
```diff
@@ -3169,7 +3169,7 @@ int main(int argc, char** argv)
   /* Set delimiter back to semicolon */
   if (retval != ERROR_STOP)
   {
-    if (!stop_event_string.is_empty())
+    if (!stop_event_string.is_empty() && result_file)
```
cvicentiu, line 3172: @an3l @LinuxJedi I am confused about this check only happening here.

If this check is necessary here, why is the fprintf at line 3161 legal? That one doesn't check for opt_raw_mode either. Is `opt_raw_mode` mutually exclusive with `opt_flashback`?

an3l: @cvicentiu  coverity [report](https://scan5.scan.coverity.com/reports.htm#v58936/p10357/fileInstanceId=231728438&defectInstanceId=53074517&mergedDefectId=1519690&eventId=53074517-21) said that in case when`opt_raw_mode` is used and `opt_flasback` is 0, `result_file` is assigned to 0 (line 3071) and when comes to this line, `fprintf` is used on 0.

PR #2497, opened 2023-02-14
Fix mini-benchmark

robinnewhouse:
## Description

The mini-benchmark.sh script failed to run in the latest Fedora
distributions in GitLab CI. It requires `lscpu` resolved by installing
util-linux.

Additionally, executing the benchmark inside a Docker container had
failed because of increased Docker security in recent updates. In
particular the `renice` and `taskset` operations are not permitted.
Neither are the required `perf` operations.
https://docs.docker.com/engine/security/seccomp/

Allow these operations to fail gracefully, and test then skip `perf`,
allowing the remaining benchmark activities to proceed.

Other minor changes to the CI are included such as allowing sanitizer
jobs to fail and using "needs" in the mini-benchmark pipeline.

## How can this PR be tested?
The pipeline for these changes successfully complete and a [public pipeline](https://salsa.debian.org/robinnewhouse/mariadb-server-mirror/-/pipelines/499774) is available.

- [x] *This is a (CI) bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced*

The mini-benchmark test was introduced in 10.8 (the latest development branch at the time), so this correction targets the earliest branch where there is a CI failure to be fixed.

## Copyright
Amazon Web Services, Inc.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -331,6 +331,7 @@ mysql-test-run-asan:
   needs:
     - "fedora-sanitizer: [-DWITH_ASAN=YES]"
   <<: *mysql-test-run-def
+  allow_failure: true
   artifacts:
     when: always  # Also show results when tests fail
     reports:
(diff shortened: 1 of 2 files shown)


robinnewhouse:
> @robinnewhouse I'm not opposed to this, but can you please mention in the comments why this is against 10.8 and not 10.4?

Comment updated:

"The mini-benchmark test was introduced in 10.8 (the latest development branch at the time), so this correction targets the earliest branch where there is a CI failure to be fixed."

LinuxJedi (approved):

PR #2502, opened 2023-02-20
MDEV-30687: Make small facelifting to autobake-debs.sh

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-30687*

## Description
Currently autobake-debs.sh does not pass shellcheck it fails making errors:

 * SC1091 when using shellcheck -x it needs to know where to find ./VERSION. As this is not needed we just specify it as /dev/null as mentioned in shellcheck documentation: https://www.shellcheck.net/wiki/SC1091

 * SC2086 make sure that there is no globbing or word splitting in dpkg-buidpackage string. This not big problem or about to happen but now extra parameter parsing is more Bash compliant with using array. Make also sure that BUILDPACKAGE_PREPEND is always `eatmydata` and nothing else. see documentation: https://www.shellcheck.net/wiki/SC2086

Fix small script indentation problem.

## How can this PR be tested?
First try to build with `debian/autobake-debs.sh` and see if it still works. After that install shellcheck (with pip install shellcheck) or from Debian repos `apt install shellcheck`.
Then check script before `shellcheck debian/autobake-debs.sh` and with this PR. Problem which should arise are:
```
In debian/autobake-deb.sh line 19:
source ./VERSION
       ^-------^ SC1091 (info): Not following: ./VERSION was not specified as input (see shellcheck -x).


In debian/autobake-deb.sh line 209:
fakeroot $BUILDPACKAGE_PREPEND dpkg-buildpackage -us -uc -I $BUILDPACKAGE_FLAGS
         ^-------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                            ^-----------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
fakeroot "$BUILDPACKAGE_PREPEND" dpkg-buildpackage -us -uc -I "$BUILDPACKAGE_FLAGS"

For more information:
  https://www.shellcheck.net/wiki/SC1091 -- Not following: ./VERSION was not ...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

## Backward compatibility
Should not make any changes to backward compatibility

## PR quality check


debian/autobake-deb.sh
```diff
@@ -173,13 +174,26 @@ echo "Creating package version ${VERSION} ... "
 # during the entire build process (safe because a build can always be restarted)
 if which eatmydata > /dev/null
 then
-  BUILDPACKAGE_PREPEND=eatmydata
+  BUILDPACKAGE_PREPEND="yes"
```
ottok, line 177: This change and line 196 seems a bit complex. Why not continue using a simple variable that is either empty or has `eatmydata` in it?

ottok:
Looks good in general. I am a big fan of people using Shellcheck in general.

debian/autobake-deb.sh
illuusio: This is actually how shellcheck wants it. I think I'll rework it as it's not cleaner or better than it was before but I think I can make it shellcheck proof without that change as I don't like it by myself (Did it any way :-1: )

ottok:
Looks good! I would vote for merging this now if the CI was not failing. Did you check what it fails on?

illuusio:
It's some `fakeroot` problem. It should be sorted out before merging.

illuusio:
Worked fakeroot calling again to new system where there is variable where `eatmydata` is added if needed

ottok (approved):
I did not test running this, but it looks good to me!

PR #2503, opened 2023-02-20
MDEV-27317 wsrep_checkpoint order violation due to certification failure

temeo:
With binlogs enabled, debug assertion ut_ad(xid_seqno > wsrep_seqno) fired in trx_rseg_update_wsrep_checkpoint() when an applier thread synced the seqno out of order for write set which had failed certification. This was caused by releasing commit order too early when binlogs were on, allowing group commit to run in parallel and commit following transactions too early.

Fixed by extending the commit order critical section to cover call to wsrep_set_SE_checkpoint() also when binlogs are on.

---

Note that even though MDEV-27317 was opened against 10.6, this issue was reproducible with versions 10.4 and above, and therefore opening this PR against 10.4.

### DIFF
--- a/sql/wsrep_high_priority_service.cc
+++ b/sql/wsrep_high_priority_service.cc
@@ -472,7 +472,13 @@ int Wsrep_high_priority_service::log_dummy_write_set(const wsrep::ws_handle& ws_
     if (!WSREP_EMULATE_BINLOG(m_thd))
     {
       wsrep_register_for_group_commit(m_thd);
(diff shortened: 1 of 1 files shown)


sysprg (approved):
Thanks, everything works and no regression found on tests of mtr suites related to Galera

sysprg:
Merged with head revision as https://github.com/MariaDB/server/commit/cadc3efcddb078fe8fe19a6121cc2a95be9a97de

PR #2505, opened 2023-02-21
MDEV-30697, MDEV-30698 - raw [and] flashback options for mariadb-binlog test coverage and memory leak

an3l:


LinuxJedi:
@an3l nice work, but can you please rebase off 10.4?

an3l:
I can rebase with current state to 10.5, since `die()` function is added with e30a3048dacc  commit in 10.5+, not sure why not pushed to 10.4. 
For the sake of this patch 
now I will use `exit` function

LinuxJedi (approved):

PR #2392, opened 2022-12-21
MDEV-30205 Move /usr/share/mysql to /usr/share/mariadb

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-30205*

## Description

These are mainly internal files so is a low impact change.

The few scripts/mysql*sql where renames to mariadb_* prefix on the name.

mysql-test renamed to mariadb-test in the final packages.

## How can this PR be tested?

```
$ podman run -ti --rm ubuntu:20.04 
root@5ce193aeebd9:/#  apt-get update && apt-get install -y ca-certificates
root@5ce193aeebd9:/# echo "deb [trusted=yes] https://ci.mariadb.org/31091/amd64-ubuntu-2004-deb-autobake/debs ./" >>/etc/apt/sources.list
root@5ce193aeebd9:/# 
root@5ce193aeebd9:/# apt-get update
root@5ce193aeebd9:/# apt-get install mariadb-test
root@5ce193aeebd9:/#  cd /usr/share/mariadb/
root@5ce193aeebd9:/usr/share/mariadb# ls -la
total 3656
drwxr-xr-x. 30 root root    4096 Dec 21 06:11 .
drwxr-xr-x.  1 root root    4096 Dec 21 06:11 ..
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 bulgarian
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 charsets
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 chinese
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 czech
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 danish
-rwxr-xr-x.  1 root root    3015 Dec 21 05:29 debian-start.inc.sh
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 dutch
-rwxr-xr-x.  1 root root      27 Dec 21 05:29 echo_stderr
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 english
-rw-r--r--.  1 root root  718421 Dec 21 05:29 errmsg-utf8.txt
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 estonian
-rw-r--r--.  1 root root 2165918 Dec 21 05:29 fill_help_tables.sql
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 french
drwxr-xr-x.  2 root root    4096 Dec 21 06:11 german
[...]

PR #2507, opened 2023-02-22
Fix warning in mariadb-install-db

LinuxJedi:
If you are running mariadb-install-db from a source tree instead of installation it was executing `mysqld` instead of `mariadb` which showed the deprecation warning. This patch fixes that as well as fixing messages and links to other things that have been renamed.

### DIFF
--- a/scripts/mysql_install_db.sh
+++ b/scripts/mysql_install_db.sh
@@ -82,7 +82,7 @@ Usage: $0 [OPTIONS]
   --defaults-group-suffix=name
                        In addition to the given groups, read also groups with
                        this suffix
-  --force              Causes mysql_install_db to run even if DNS does not
+  --force              Causes mariadb-install-db to run even if DNS does not
                        work.  In that case, grant table entries that
                        normally use hostnames will use IP addresses.
   --help               Display this help and exit.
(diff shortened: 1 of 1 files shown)


scripts/mysql_install_db.sh
```diff
@@ -402,9 +402,9 @@ fi
 
 # Set up paths to SQL scripts required for bootstrap
 fill_help_tables="$srcpkgdatadir/fill_help_tables.sql"
-create_system_tables="$srcpkgdatadir/mysql_system_tables.sql"
+create_system_tables="$srcpkgdatadir/mariadb_system_tables.sql"
```
grooverdan, line 405: there SQL filenames haven't changed (yet), included in #2392

grooverdan (commented):
One error on SQL filenames. Otherwise fine and good to merge.

scripts/mysql_install_db.sh
LinuxJedi: You are correct, it appears I had that file lying around from something previous. I will fix now.

illuusio:
As naming of file is wrong it after name change there should be update of Debian `mariadb-server` and `mariadb-server-compat` this also affects RPM packaging?

PR #2508, opened 2023-02-28
MDEV-30713 field length handling for CONNECT engine

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-30713*

## Description

fp->field_length was unsigned and therefore the negative condition around it.

Backport of cc182aca9352 fixes it, however to correct the consistent use of types pcf->Length needs to be unsigned too.

At one point pcf->Precision is assigned from pcf->Length so that's also unsigned.

GetTypeSize is assigned to length and has a length argument. A -1 default value seemed dangerious to case, so at least 0 should assert if every hit.

## How can this PR be tested?

Existing test suite. Look at warnings generated.

## Backward compatibility

Hoping the -1 -> 0 change in GetType doesn't cascade to a meaningful change.

## PR quality check

### DIFF
--- a/storage/connect/catalog.h
+++ b/storage/connect/catalog.h
@@ -39,9 +39,9 @@ typedef struct _colinfo {
   PCSZ   Name;
   int    Type;
   int    Offset;
-  int    Length;
+  unsigned    Length;
   int    Key;
-  int    Precision;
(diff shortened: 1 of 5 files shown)


LinuxJedi (approved):
Good call.

PR #2512, opened 2023-03-01
MENT-1708 Donor in non-Primary causes assertion in wsrep-lib

temeo:
Constructed a test which makes donor to go into non-Primary configuration before `sst_sent()` is called, causing an assertion in wsrep-lib if the bug is present.

Updated wsrep-lib to version which contains the fix.


sysprg:
Thanks, fix merged with head revision:
https://github.com/mariadb-corporation/MariaDBEnterprise/commit/19b0a6c5916002b5169710d0163698db6f899adf
https://github.com/mariadb-corporation/MariaDBEnterprise/commit/983fae5888ce24053e5bec83aae65a76b0d824ce
https://github.com/mariadb-corporation/MariaDBEnterprise/commit/d1333e8dab495c7592c688b724538d3255f11fd8

PR #2513, opened 2023-03-01
Fix few vulnerabilities found by Cppcheck

HugoWenTD:
## Description

While performing SAST scanning using Cppcheck against source code of
commit 81196469, several code vulnerabilities were found.

Fix following issues:

1. Parameters of `snprintf` function are incorrect.

   Cppcheck error:

       client/mysql_plugin.c:1228: error: snprintf format string requires 6 parameters but only 5 are given.

   It is due to commit 630d7229 introduced option `--lc-messages-dir`
   in the bootstrap command. However the parameter was not even given
   in the `snprintf` after changing the format string.

   Fix:
   Restructure the code logic and correct the function parameters for
   `snprintf`.

2. Null pointer is used in a `snprintf` which could cause a crash.

   Cppcheck error:

       extra/mariabackup/xbcloud.cc:2534: error: Null pointer dereference

   The code intended to print the swift_project name, if the
   opt_swift_project_id is NULL but opt_swift_project is not NULL.
   However the parameter of `snprintf` was mistakenly using
   `opt_swift_project_id`.

   Fix:
   Change to use the correct string from `opt_swift_project`.

3. Potential double release of a memory

   Cppcheck error:

       plugin/auth_pam/testing/pam_mariadb_mtr.c:69: error: Memory pointed to by 'resp' is freed twice.

   A pointer `resp` is reused and allocated new memory after it has been
   freed. However, `resp` was not set to NULL after freed.
   Potential double release of the same pointer if the call back
   function doesn't allocate new memory for `resp` pointer.

   Fix:
   Set the `resp` pointer to NULL after the first free() to make sure
   the same address is not freed twice.

## How can this PR be tested?

The PR is to fix potential vulnerabilities. Existing MTR tests passed and no new test needed.

## Copyright

Services, Inc.


grooverdan (approved):

grooverdan:
Thanks @HugoWenTD. Looks good to me.

grooverdan:
request for next time, commit title to include more descriptive about the component changed and rough nature of change e.g. 'mysql_plugin: snprintf wrong number of args'.

HugoWenTD:
> request for next time, commit title to include more descriptive about the component changed and rough nature of change e.g. 'mysql_plugin: snprintf wrong number of args'.

Definitely will keep that in mind in future PRs. Thank you so much for reviewing and merging!

PR #2514, opened 2023-03-01
MDEV-35987 - Minimize unsafe C functions usage - replace strcat() and strcpy() - continues

Chaloff:
## Description

Similar to 567b681 continue to replace use of strcat() and strcpy() with safer
custom methods safe_strcat() and safe_strcpy().

## How can this PR be tested?

All build stages pass for these commits.

## Backward compatibility

that are either new files or modified ones, are contributed under the BSD-new
license. I am contributing on behalf of my employer Amazon Web Services


storage/connect/tabmul.cpp
```diff
@@ -467,7 +469,11 @@ int TDBMUL::ReadDB(PGLOBAL g)
         } // endif iFile
 
     } else if (rc == RC_FX)
-      strcat(strcat(strcat(g->Message, " ("), Tdbp->GetFile(g)), ")");
+    {
+      safe_strcat(g->Message, sizeof(g->Message), " (");
+      safe_strcat(g->Message, sizeof(g->Message), Tdbp->GetFile(g));
```
grooverdan, line 474: wouldn't this be more readable as `snprintf`.

storage/connect/tabmul.cpp
```diff
@@ -567,11 +573,12 @@ bool TDBMSD::InitFileNames(PGLOBAL g)
 	while (true)
```
grooverdan, line 573: I useful check before this point would be a check if `filename == NULL`. Also obscured by obsessive overloading here is that `FNSZ + 1` bytes is allocated.

storage/connect/tabmul.cpp
```diff
@@ -567,11 +573,12 @@ bool TDBMSD::InitFileNames(PGLOBAL g)
...
 		if ((rc = dirp->ReadDB(g)) == RC_OK) {
 #if defined(_WIN32)
-			strcat(strcpy(filename, dirp->Drive), dirp->Direc);
+			snprintf(filename, FNSZ, "%s%s", dirp->Drive, dirp->Direc);
 #else   // !_WIN32
-			strcpy(filename, dirp->Direc);
+			snprintf(filename, FNSZ, "%s", dirp->Direc);
 #endif  // !_WIN32
-			strcat(strcat(filename, dirp->Fname), dirp->Ftype);
+			safe_strcat(filename, FNSZ, dirp->Fname);
```
grooverdan, line 580: roll these into the previous `snprintf`s.

storage/connect/tabmysql.cpp
```diff
@@ -683,15 +684,21 @@ bool TDBMYSQL::MakeCommand(PGLOBAL g)
 
 
     // Make a lower case copy of the originale query
-    qrystr = (char*)PlugSubAlloc(g, NULL, strlen(Qrystr) + 5);
-    strlwr(strcpy(qrystr, Qrystr));
+    size_t qrystr_sz = strlen(Qrystr) + 5;
+    qrystr = (char*)PlugSubAlloc(g, NULL, qrystr_sz);
+    snprintf(qrystr, qrystr_sz, "%s", Qrystr);
+    strlwr(qrystr);
```
grooverdan, line 690: rather than copy a string and force it to lower case after the copy. Do this all in one step with an appropriately wrapped function.

storage/connect/tabmysql.cpp
```diff
@@ -760,7 +770,8 @@ int TDBMYSQL::MakeDelete(PGLOBAL g)
   {
   char *qc, cmd[8], from[8], tab[96], end[512];
 
-  Query = (char*)PlugSubAlloc(g, NULL, strlen(Qrystr) + 64);
+  size_t Query_sz = strlen(Qrystr) + 64;
+  Query = (char*)PlugSubAlloc(g, NULL, Query_sz);
```
grooverdan, line 774: allocation failure?

storage/connect/tabmysql.cpp
```diff
@@ -732,7 +742,8 @@ int TDBMYSQL::MakeUpdate(PGLOBAL g)
   {
   char *qc, cmd[8], tab[96], end[1024];
 
-  Query = (char*)PlugSubAlloc(g, NULL, strlen(Qrystr) + 64);
+  size_t Query_sz = strlen(Qrystr) + 64;
+  Query = (char*)PlugSubAlloc(g, NULL, Query_sz);
```
grooverdan, line 746: allocation failure?

storage/connect/tabmul.cpp
Chaloff: This not a part of my change but anyway I can add the check here. The question is - should we really do this?
`filename` can be  `NULL` here in case of `PlugSubAlloc()` fails but it should write error message and thow an exception then - see https://github.com/MariaDB/server/blob/10.4/storage/connect/plugutil.cpp#L581-L587

I agree it's not very clear here. I can add additional check if you still think it's right thing to do.

storage/connect/tabmysql.cpp
Chaloff: The logic here wasn't really changed - just added size variable to use it bellow. Allocation failure handled by `PlugSubAlloc()` - see https://github.com/MariaDB/server/blob/10.4/storage/connect/plugutil.cpp#L581-L587
Should we check the failure here as well?

storage/connect/tabmysql.cpp
Chaloff: The same comment as above (L745) Allocation failure handled by `PlugSubAlloc()`. We can add additional check though if you think we need

storage/connect/tabmysql.cpp
Chaloff: The problem here is - all memory safe function we have (`safe_strcpy()` `snprintf()`) return int value (number of bytes copied) We can only wrapped unsafe functions (`strcpy()` as before or `sprintf()`) but this makes no sense here - we are trying to do opposite...

grooverdan:
@Chaloff can you fix the vulnerability introduced in 567b68129943 described in #2640 and create a gcc bug report for `array-bounds` such that the obsessive cleaning of "unsafe" functions doesn't create "safe" vulnerabilities of human error. 

FWIW I saw the erroneous warnings in g++13.1.1 (20230511 (Red Hat 13.1.1-2)) in Fedora 38.

Chaloff:
> @Chaloff can you fix the vulnerability introduced in [567b681](https://github.com/MariaDB/server/commit/567b68129943a1cceab1d7b4c68e2a4ba011cdc0) described in #2640 and create a gcc bug report for `array-bounds` such that the obsessive cleaning of "unsafe" functions doesn't create "safe" vulnerabilities of human error.
> 
> FWIW I saw the erroneous warnings in g++13.1.1 (20230511 (Red Hat 13.1.1-2)) in Fedora 38.

This was addressed here https://github.com/MariaDB/server/pull/2692

storage/connect/tabmul.cpp
```diff
@@ -1367,10 +1377,10 @@ TDBDHR::TDBDHR(PTDBDHR tdbp) : TDBASE(tdbp)
...
   Hsearch = tdbp->Hsearch;
   iFile = tdbp->iFile;
-  strcpy(Drive, tdbp->Drive);
-  strcpy(Direc, tdbp->Direc);
-  strcpy(Fname, tdbp->Fname);
-  strcpy(Ftype, tdbp->Ftype);
+  snprintf(Drive, sizeof(Drive), tdbp->Drive);
+  snprintf(Direc, sizeof(Direc), tdbp->Direc));
+  snprintf(Fname, sizeof(Fname), tdbp->Fname));
+  snprintf(Ftype, sizeof(Ftype), tdbp->Ftype);
```
LinuxJedi, line 1383: Why not something like `safe_strcpy()` here? I'm pretty sure doing this without an `"%s"` will trigger warnings on some compilers anyway.

storage/connect/tabmul.cpp
```diff
@@ -1438,7 +1448,7 @@ int TDBDHR::GetMaxSize(PGLOBAL g)
 
       // Close the search handle.
       if (!FindClose(h) && n != -1)
-        strcpy(g->Message, MSG(SRCH_CLOSE_ERR));
+        snprintf(g->Message, sizeof(g->Message), MSG(SRCH_CLOSE_ERR));
```
LinuxJedi, line 1451: Likewise, should probably be `safe_strcpy()`.

storage/connect/tabmysql.cpp
```diff
@@ -683,15 +683,21 @@ bool TDBMYSQL::MakeCommand(PGLOBAL g)
 
 
     // Make a lower case copy of the originale query
-    qrystr = (char*)PlugSubAlloc(g, NULL, strlen(Qrystr) + 5);
-    strlwr(strcpy(qrystr, Qrystr));
+    size_t qrystr_sz = strlen(Qrystr) + 5;
+    qrystr = (char*)PlugSubAlloc(g, NULL, qrystr_sz);
+    snprintf(qrystr, qrystr_sz, "%s", Qrystr);
```
LinuxJedi, line 688: Another one that can probably be `safe_strcpy`.

LinuxJedi (changes requested):
Thank you for your work so far on this. I flagged a few instances where you are replacing `strcpy` with `snprintf` but not using format specifiers, there are more of them. I'm pretty sure this will trigger a warning, but worse, it is a security vulnerability. The string could be injected with a format specifier. It would probably be better to use `safe_strcpy` instead.

gkodinov (changes requested):
This is a preliminary review. Sorry for the very long delay! We're improving our procedures and making sure that no pull request is left behind. 

Same requests as the other one:
* please rebase to 10.11
* please make sure the pull request contains only your changes: right now it says there's about 1m lines changed.

gkodinov:
there has been no reply to my comments for more than 3 weeks now. Converting to draft. Please re-open when/if you intend to resume working on this.

PR #2517, opened 2023-03-02
MDEV-24005 Updated the --use-memory option usage message in Mariaback…

is-this-echo:
- [x] *The Jira issue number for this PR is: MDEV-24005*

## Description

The  --use-memory option is only relevant when performing a --prepare. The usage message is updated accordingly in this PR as it was not descriptive previously.

Previous:
```
debjyoti@d3viL3cho:~/build-mariadb-server-debug$ extra/mariabackup/mariabackup --help | grep use-memory -2
  --print-param       print parameter of mysqld needed for copyback.
  --use-memory=#      The value is used instead of buffer_pool_size
  --throttle=#        limit count of IO operations (pairs of read&write) per
```

Now : 
```
debjyoti@d3viL3cho:~/original-server/build-10.4$ extra/mariabackup/mariabackup --help | grep use-memory -2
  --print-param       print parameter of mysqld needed for copyback.
  --use-memory=#      The value is used in place of innodb_buffer_pool_size.
                      This option is only relevant when the --prepare option is
                      specified.
```

## How can this PR be tested?
NA

## PR quality check

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -1088,7 +1088,8 @@ struct my_option xb_client_options[]= {
      (G_PTR *) &xtrabackup_print_param, (G_PTR *) &xtrabackup_print_param, 0,
      GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
     {"use-memory", OPT_XTRA_USE_MEMORY,
-     "The value is used instead of buffer_pool_size",
+     "The value is used in place of innodb_buffer_pool_size. "
+     "This option is only relevant when the --prepare option is specified.",
      (G_PTR *) &xtrabackup_use_memory, (G_PTR *) &xtrabackup_use_memory, 0,
      GET_LL, REQUIRED_ARG, 100 * 1024 * 1024L, 1024 * 1024L, LONGLONG_MAX, 0,
      1024 * 1024L, 0},


extra/mariabackup/xtrabackup.cc
```diff
@@ -1088,7 +1088,8 @@ struct my_option xb_client_options[]= {
      (G_PTR *) &xtrabackup_print_param, (G_PTR *) &xtrabackup_print_param, 0,
      GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
     {"use-memory", OPT_XTRA_USE_MEMORY,
-     "The value is used instead of buffer_pool_size",
+     "The value is used in place of innodb_buffer_pool_size. " 
+	 "This option is only relevant when the --prepare option is specified.",
```
is-this-echo, line 1092: Not sure why this indentation shows up in git diff but not in my IDE

grooverdan: the line began with a tab. I've corrected this the latest push. Also previous line ended in a space.

Thanks for the fix.

grooverdan (approved):

is-this-echo:
Thanks for the correction @grooverdan , I see that the test runs have been stopped , that is why it shows failing.

LinuxJedi:
> Thanks for the correction @grooverdan , I see that the test runs have been stopped , that is why it shows failing.

We are aware, one of the builders stalled and required restarting. I've restarted the failed jobs.

PR #2520, opened 2023-03-03
MDEV-24931: Assertion `prefix_size <= width' failed in Bitmap<width>::is_prefix(uint) & UBSAN: shift exponent 32 is too large for 32-bit type 'int' in generate_derived_keys_for_table

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-24931*

## How can this PR be tested?
The test is append to `join.test`

## PR quality check

### DIFF
--- a/mysql-test/main/join.result
+++ b/mysql-test/main/join.result
@@ -3406,4 +3406,23 @@ id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
 1	SIMPLE	t3	ref	a	a	5	test.t2.key2	1	
 drop table t1,t2,t3;
 drop table t1000,t10,t03;
+#
+# MDEV-24931: Assertion `prefix_size <= width' failed
+#
+CREATE TABLE t1 (
+f01 int, f02 int, f03 int, f04 int, f05 int, f06 int, f07 int, f08 int,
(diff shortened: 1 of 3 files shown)


keeper:
Hi @Weijun-H , could you elaborate more on why does it make more sense using `DBUG_ASSERT(idx <= ARRAY_ELEMENTS);`  to check the boundary of the prefix in your commit log?

Also in your commit log stated that this commit also fix a potential UB but I believe the issue still exists in https://github.com/MariaDB/server/blob/70be59913c90e93fe5136d6f6df03c4254aa515d/sql/sql_select.cc#L13035, could you also verify that?

grooverdan (changes requested):
Good test case. Though including the other test case from the MDEV would be good.

Per @vaintroub's MDEV comment, the fix shouldn't be in the is_prefix, its why is_prefix is called with the greater value. `generate_derived_keys_for_table` is the most likely case for improvement.

Note sure if `typedef ulong key_part_map`; should be raised to ulonglong as well.

Same MDEV: #3139

svoj:
Closing as there was no response to requested changes. Feel free to reopen if you're aiming to continue work on it.

PR #2521, opened 2023-03-03
 Minimize unsafe C functions usage - replace sprintf() and strcpy()

christian7007:
## Description
Replace calls to `sprintf` and `strcpy` by the safer options `snprintf` and `safe_strcpy` in the following directories:

- libmysqld
- mysys
- sql-common
- strings

## How can this PR be tested?

All build stages pass for these commits. For the commit without the intentional bug, mysql-test-run is properly passing. The changes are not change the functionality itself they just replace unsafe C methods with the safe ones.

- [X] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced*

## Backward compatibility
The changes fully backward compatible.

## PR quality check

## Copyright

### DIFF
--- a/mysys/my_atomic_writes.c
+++ b/mysys/my_atomic_writes.c
@@ -112,7 +112,7 @@ static my_bool test_if_shannon_card_exists()
     char path[32];
     struct stat stat_buff;
 
-    sprintf(path, "/dev/df%c", dev_part);
+    snprintf(path, sizeof(path), "/dev/df%c", dev_part);
 #ifdef TEST_SHANNON
     if (lstat(path, &stat_buff) < 0)
     {
@@ -121,8 +121,10 @@ static my_bool test_if_shannon_card_exists()
     }
 #endif
     shannon_devices[shannon_found_devices].st_dev= stat_buff.st_rdev;
-    sprintf(shannon_devices[shannon_found_devices].dev_name, "/dev/sct%c",
-            dev_part);
+    snprintf(shannon_devices[shannon_found_devices].dev_name,
+             sizeof(shannon_devices[shannon_found_devices].dev_name),
+             "/dev/sct%c",
+             dev_part);
 
 #ifdef TEST_SHANNON
     printf("%s(): i=%d, stat_buff.st_dev=0x%lx, stat_buff.st_rdev=0x%lx, st_rdev=0x%lx, dev_name=%s\n",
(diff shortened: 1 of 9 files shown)


mysys/my_atomic_writes.c
```diff
@@ -112,7 +112,7 @@ static my_bool test_if_shannon_card_exists()
     char path[32];
     struct stat stat_buff;
 
-    sprintf(path, "/dev/df%c", dev_part);
+    snprintf(path, sizeof(path), "/dev/df%c", dev_part);
```
LinuxJedi, line 115: This one isn't required, the longest the path will ever be is 8 characters plus the NUL.

mysys/my_atomic_writes.c
```diff
@@ -121,8 +121,10 @@ static my_bool test_if_shannon_card_exists()
     }
 #endif
     shannon_devices[shannon_found_devices].st_dev= stat_buff.st_rdev;
-    sprintf(shannon_devices[shannon_found_devices].dev_name, "/dev/sct%c",
-            dev_part);
+    snprintf(shannon_devices[shannon_found_devices].dev_name,
```
LinuxJedi, line 124: Likewise enough space for this fixed size is already allocated, `snprintf` is redundant.

mysys/my_atomic_writes.c
```diff
@@ -145,13 +147,15 @@ static my_bool test_if_shannon_card_exists()
 
     for (dev_no= 1 ; dev_no < 9 ; dev_no++)
     {
-      sprintf(path, "/dev/df%c%d", dev_part, dev_no);
+      snprintf(path, sizeof(path), "/dev/df%c%d", dev_part, dev_no);
```
LinuxJedi, line 150: As above, this is longer but still always fits into the 32 bytes allocated.

mysys/my_likely.c
```diff
@@ -77,7 +77,7 @@ void end_my_likely(FILE *out)
   if (!(likely_file= out))
   {
     char name[80];
-    sprintf(name, "/tmp/unlikely-%lu.out", (ulong) getpid());
+    snprintf(name, sizeof(name), "/tmp/unlikely-%lu.out", (ulong) getpid());
```
LinuxJedi, line 80: This is redundant, 18+10+1 bytes needed max, 80 allocated.

mysys/my_thr_init.c
```diff
@@ -426,7 +426,7 @@ const char *my_thread_name(void)
   if (!tmp->name[0])
   {
     my_thread_id id= my_thread_dbug_id();
-    sprintf(name_buff,"T@%lu", (ulong) id);
+    snprintf(name_buff, sizeof(name_buff), "T@%lu", (ulong) id);
```
LinuxJedi, line 429: Also not needed, 100 bytes allocated, at most 13 used.

strings/uctypedump.c
```diff
@@ -225,7 +225,7 @@ int main(int ac, char ** av)
   {
     char plane_name[128]="NULL";
     if(uctype[plane].ctype){
-      sprintf(plane_name,"uctype_page%02X",(uint) plane);
+      snprintf(plane_name,sizeof(plane_name),"uctype_page%02X",(uint) plane);
```
LinuxJedi, line 228: Not needed, 128 bytes allocated, at most 15 used here.

LinuxJedi (changes requested):
Some of these aren't needed, but there are also some good catches here :)

strings/uctypedump.c
dlenski: The whole point of this is that individual humans don't need to go through and reason about the allocated-vs-used space for each and every individual call.

When we miss or misunderstand one, that's how C memory bugs happen.

dlenski:
> To me, it seems like in the cases where it is not a hotspot in the code, I'd encourage leaning in the direction of patterns of safety.


:100: agree. And adding a larger `n` to `snprintf` cannot possible make performance worse.

strings/uctypedump.c
grooverdan: Modern compilers however pick this up quite well.

LinuxJedi (approved):

LinuxJedi:
Approved after these and other discussions.

PR #2525, opened 2023-03-04
MDEV-27126 my_getopt compares option names case sensitively

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-28616*

## Description
make option name lookups and comparisons case insensitive
## How can this PR be tested?
I am not sure how to test this update.

## PR quality check

### DIFF
--- a/mysys/my_getopt.c
+++ b/mysys/my_getopt.c
@@ -23,6 +23,7 @@
 #include <mysys_err.h>
 #include <my_getopt.h>
 #include <errno.h>
+#include <ctype.h>
 
 typedef void (*init_func_p)(const struct my_option *option, void *variable,
                             longlong value);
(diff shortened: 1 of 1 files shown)


grooverdan (approved):
please rebase to 10.4 to merge.

Also optionally, correct the incorrect comment above the function that says "-1 if strings differ".

LinuxJedi (changes requested):
Hi @Weijun-H,

In general this looks good, and I wish to approve it, but a couple of things need to happen first:

1. A test case is needed. This is so that we don't accidentally regress or revert this behaviour later.
2. It will need to be moved to 10.5, because 10.4 is near EoL, so only blocker / security fixes will be going into the final release.

cvicentiu:
@Weijun-H I'm sorry for missing this PR. Indeed your contribution was first and effectively the same code made it in PR 3841.

PR #2526, opened 2023-03-06
MDEV-30771 Optimizer trace:  table_scan.rows is traced as integer, change to double

rex-johnston:
Pull request created in: https://jira.mariadb.org/browse/MDEV-30771

### DIFF
--- a/sql/my_json_writer.cc
+++ b/sql/my_json_writer.cc
@@ -217,7 +217,24 @@ void Json_writer::add_size(longlong val)
 void Json_writer::add_double(double val)
 {
   char buf[64];
-  size_t len= my_snprintf(buf, sizeof(buf), "%lg", val);
+  size_t len;
+  if (((val - (longlong)val) < OPTIMIZER_DOUBLE_TO_INT_PRECISION)
+     || (-((longlong)val - val) < OPTIMIZER_DOUBLE_TO_INT_PRECISION))
+    return Json_writer::add_ll((longlong)val);
+  if (  (val > OPTIMIZER_DOUBLE_TO_EXPONENT_VALUE)
+      ||(-val > OPTIMIZER_DOUBLE_TO_EXPONENT_VALUE)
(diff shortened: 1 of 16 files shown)


sql/my_json_writer.h
```diff
@@ -40,6 +40,16 @@ class Opt_trace_context;
...
+// See Json_writer::add_double() for usage
+#define OPTIMIZER_SHOW_DOUBLE_PRECISION        true
+#if OPTIMIZER_SHOW_DOUBLE_PRECISION
+#  define OPTIMIZER_DOUBLE_PRECISION   2
+#  define OPTIMIZER_DOUBLE_MIN_VALUE   0.005
+#else
+#  define OPTIMIZER_DOUBLE_SIGNIFICANT_DIGITS 6
+#endif
+#define OPTIMIZER_DOUBLE_TO_INT_PRECISION       0.00000001
+#define OPTIMIZER_DOUBLE_TO_EXPONENT_VALUE      999999
```
haidong, line 52: How was `999999` chosen? What if the `val` is greater than than?

sql/my_json_writer.cc
```diff
@@ -217,7 +217,24 @@ void Json_writer::add_size(longlong val)
 void Json_writer::add_double(double val)
 {
   char buf[64];
-  size_t len= my_snprintf(buf, sizeof(buf), "%lg", val);
+  size_t len;
```
haidong, line 220: Would it be easier, and also avoid edge cases when `val` is beyond than what's defined in `my_json_writer.h`, by using the following:
```
size_t len= my_snprintf(buf, sizeof(buf), "%.*lg",
          2, val);
```
If this works, we also avoid adding additional CONSTANT definitions in the header file.

haidong (commented):
I think this PR should address MDEV-30771.

I prefer simple changes if it does the job, hence the comments below. The simple approach, if it works, not only avoids adding additional code to the header, and but also makes the fairly complicated `if` conditions unnecessary.

PR #2528, opened 2023-03-06
MDEV-30269: Remove rpl_semi_sync_[slave,master] usage in code

an3l:
- Description:
  - Before 10.3.8 semisync was a plugin that is built into the server with
    MDEV-13073,starting with commit https://github.com/an3l/server/commit/cbc71485e24c31fc822277625512e55c2a8b650b.
    There are still some usage of `rpl_semi_sync_master` in mtr and
    `rpl_semi_sync_slave` keywords in the code.
    This patch should remove/explain the usage of them.

- Solution:
  - Found in `sys_vars.all_vars` test usage of plugins
  `rpl_semi_sync_master`, `rpl_semi_sync_slave` and this test is
  disabled by default (`sys_vars/disabled.def`) and marked as
  `obsolete`.
  Q1: However I would remove the usage of non-existing plugins, should
  we?
  - Found only single occurence of `rpl_semi_sync_master` system
  variable as part of compression in mtr, remove it.
  - To recognize the replica in the `dump_thread`, replica is creating
  local variable `rpl_semi_sync_slave` (the keyword of plugin) in
  function `request_transmit`, that is catched by primary in
  `is_semi_sync_slave`.
  This patch renamed the local variable `var_set_from_semi_sync_replica`.
  Q2: Additionally we can change the name of function `is_semi_sync_replica`?


LinuxJedi:
As far as I can tell, this would break semi-sync replication between two different versions of 10.4?

an3l:
Ah haven't considered that, thanks @LinuxJedi. I will update the patch @bnestere, just to remove obsolete mtr commands.

bnestere (approved):
This looks good, thanks! Before pushing, can you update the commit message to reflect the reduced scope of the patch?

an3l:
Commit message was updated. Now have updated PR description with the same.

PR #2529, opened 2023-03-06
MDEV-8587 Aria log encryption

LinuxJedi:
Contributed patch from MDEV-8587, adds the additional option `aria_encrypt_translog`.

- [x] *The Jira issue number for this PR is: MDEV-8587*

Author has declared in the MDEV that they are contributing this code under the 3-clause BSD license.


LinuxJedi:
Converted to draft as this will need more work before it is ready, as indicated in the MDEV.

PR #2533, opened 2023-03-08
MDEV-30811: fail to compile 11.0 on MacOS

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-30811*

## Description
Failed to compile 11.0 on MacOS Monterey, 12.5.1, ARM architecture, clang version 15.0.3.

## How can this PR be tested?
This is a specific issue on MacOS, so it is hard to be tested.

## PR quality check

### DIFF
--- a/mysys/my_init.c
+++ b/mysys/my_init.c
@@ -191,8 +191,9 @@ my_bool my_init(void)
       int res= 1;
 #ifdef EXE_LINKPATH
       res= my_readlink(link_name, EXE_LINKPATH, MYF(0));
+#else res = my_readlink(link_name, my_progname, MYF(0));
 #endif
-      if ((res == 0 || my_readlink(link_name, my_progname, MYF(0)) == 0) &&
+      if (res == 0 &&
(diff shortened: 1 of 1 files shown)


vuvova:
after this change, you will not issue a warning for a case like `ln -s mariadb mysql`,

vuvova:
That's not correct either. Note the comment in this file:
```c
#elif defined(__FreeBSD__)
/* unfortunately, not mounted by default */
#define EXE_LINKPATH "/proc/curproc/file"
#endif
```
So `/proc/curproc/file` could be a symlink to the current executable on FreeBSD, but `/proc` is not mounted by default there. That's why the code first tries to readlink `EXE_LINKPATH` and if that fails, it resorts to reading `my_progname`

dmitryshulga:
This issue is fixed by moving initialization of thread environment at very beginning of the main() 
The patch is created and sent for review. As soon as the patch be approved by reviewer it will be pushed into the branch 11.0.

vuvova:
fixed with 1310b3a02fe6c5eb14d61adbe13453ee0ab966fa

PR #2534, opened 2023-03-08
MDEV-13915: STOP SLAVE takes very long time on a busy system

bnestere:
The problem is that a parallel replica would not immediately stop
running/queued transactions when issued STOP SLAVE. This would lead
to long periods to wait for all waiting transactions to finish.

This patch updates a parallel replica to try and abort immediately
and roll-back any ongoing transactions. There are two exceptions:
   1. The transaction is next-to-commit and it has already modified
      non-transactional table and cannot be safely rolled back.
   2. We are at the commit step of the transaction.

The first commit on the branch is the MTR tests which show the problem.
The second commit is the code changes.
The third commit is updates to other MTR tests which needed to be fixed.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_row_par_stop_slave_quick.test
+++ b/mysql-test/suite/rpl/t/rpl_row_par_stop_slave_quick.test
@@ -1,0 +1,276 @@
+#
+# Validate that STOP SLAVE works in a timely manner on a parallel replica with
+# ROW binary logging format.
+#
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+--source include/master-slave.inc
+--source include/have_innodb.inc
+--source include/have_binlog_format_row.inc
+
+--echo #
+--echo # Setup
+--connection slave
+--source include/stop_slave.inc
+SET STATEMENT sql_log_bin=0 FOR call mtr.add_suppression("Can't find record");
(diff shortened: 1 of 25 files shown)


mysql-test/suite/rpl/t/rpl_row_par_stop_slave_quick.test
```diff
@@ -0,0 +1,276 @@
...
+--connect(slave_lock_extra,127.0.0.1,root,,test,$SLAVE_MYPORT)
+CHANGE MASTER TO MASTER_USE_GTID=SLAVE_POS;
+--source include/start_slave.inc
+
+--echo #
+--echo # Initialize test data
+--connection master
+SET STATEMENT sql_log_bin=0 FOR call mtr.add_suppression('Unsafe statement written to the binary log using statement format since BINLOG_FORMAT = STATEMENT.');
+--let $old_binlog_direct= `SELECT @@global.binlog_direct_non_transactional_updates`
+SET @@global.binlog_direct_non_transactional_updates= 0;
```
andrelkin, line 28: `direct <- 0` does not affect ROW format only STATEMENT one.

bnestere:
Closed in favor of the 10.5 re-target [PR-2663](https://github.com/MariaDB/server/pull/2663)

PR #2536, opened 2023-03-09
MDEV-30815 Printing a stack trace leaks file descriptors.

rex-johnston:
Pull request created in: https://jira.mariadb.org/browse/MDEV-30815

### DIFF
--- a/mysys/my_addr_resolve.c
+++ b/mysys/my_addr_resolve.c
@@ -115,7 +115,7 @@ int my_addr_resolve(void *ptr, my_addr_loc *loc)
   return 1;
 }
 
-const char *my_addr_resolve_init()
+const char *my_addr_resolve_init(void *ptr)
 {
   if (!bfdh)
   {
@@ -145,6 +145,11 @@ const char *my_addr_resolve_init()
 err:
   return bfd_errmsg(bfd_get_error());
 }
+
+
+void my_addr_resolve_close()
+{
+}
 #elif defined(HAVE_LIBELF_H)
 /*
   another possible implementation.
@@ -171,19 +176,11 @@ static char addr2line_binary[1024];
 static char output[1024];
 static struct pollfd poll_fds;
 static void *addr_offset;
+Dl_info info;
+
 
 int start_addr2line_fork(const char *binary_path)
 {
-
-  if (pid > 0)
-  {
-    /* Don't leak FDs */
-    close(in[1]);
-    close(out[0]);
-    /* Don't create zombie processes. */
-    waitpid(pid, NULL, 0);
-  }
-
   if (pipe(in) < 0)
     return 1;
   if (pipe(out) < 0)
@@ -209,6 +206,16 @@ int start_addr2line_fork(const char *binary_path)
   close(out[1]);
 
   return 0;
+}
+
+
+void my_addr_resolve_close()
+{
+  /* Don't leak FDs */
+  close(in[1]);
+  close(out[0]);
+  /* Don't create zombie processes. */
+  waitpid(pid, NULL, 0);
 }
 
 static int first_error= 0;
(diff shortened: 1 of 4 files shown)


cvicentiu:
@rex-johnston Thanks for the PR. I'll have a look at this shortly and let you know.

dlenski:
Why is this PR targeting MariaDB **10.5**?

Per [the PR template](https://github.com/MariaDB/server/blob/11.0/.github/pull_request_template.md?plain=1#L45-L47), PRs should normally be targeted to either the latest in-development major version (currently `11.0`) or the earliest version on which a bugfix applies.

Using the template here would have clarified that, if it is indeed the case that 10.5 is the earliest version where this bug exists.

mysys/stacktrace.c
```diff
@@ -185,7 +185,7 @@ static int print_with_addr_resolve(void **addrs, int n)
   int i;
   const char *err;
 
-  if ((err= my_addr_resolve_init()))
+  if ((err= my_addr_resolve_init(addrs[0])))
```
dlenski, line 188: If `n==0`, then this will SIGSEGV in a way that it never would have before, because you're unconditionally dereferencing `addrs[0]`.

Consider adding…
```diff
+if (n < 1)
+  return 1;    // Nothing to do
+
 if ((err= my_addr_resolve_init(addrs[0])))
 {
```

dlenski (changes requested):

mysys/stacktrace.c
rex-johnston: n==0 means backtrace() returns zero (or NULL), which implies the program has no stack (or there is something broken in libcX).

rex-johnston:
> Why is this PR targeting MariaDB **10.5**?
> 
> Per [the PR template](https://github.com/MariaDB/server/blob/11.0/.github/pull_request_template.md?plain=1#L45-L47), PRs should normally be targeted to either the latest in-development major version (currently `11.0`) or the earliest version on which a bugfix applies.
> 
> Using the template here would have clarified that, if it is indeed the case that 10.5 is the earliest version where this bug exists.

The only reason I issued a PR was to make it run all the BB checks.  Targeting 10.5 as that is what DBS is running and I wanted no issues with patching their code.  Best procedure here?  I expect my_print_stacktrace() to be rewritten in the near future anyway.

mysys/stacktrace.c
dlenski: In the corner cases where the stack trace can't be generated (we've seen a few :man_shrugging:), it'll help a lot to not generate a new `SIGSEGV` as a result.

vuvova:
This code was supposed to run only on crashes, like SIGSEGV or SIGABRT. There's one exec per trace and one trace per crash. Leaking two file descriptors in the crash signal handler is hardly an issue.

The leak got noticed when someone mistakenly started using this code in the running server, in a rather often executed part of the code. I'd say that fork+exec of addr2line is generally a wrong approach for that use case. Better to add support for some GPLv2 compatible symbol-resolving library instead. libelf might work.

mysys/my_addr_resolve.c
```diff
@@ -297,48 +304,50 @@ static int addr_resolve(void *ptr, my_addr_loc *loc)
...
+}
+
+
+const char *my_addr_resolve_init(void *ptr)
+{
   if (!dladdr(ptr, &info))
-    return 1;
+    return "dladdr() error";
 
-  if (strcmp(addr2line_binary, info.dli_fname))
```
vlad-lesin, line 305: As I understood it correctly, in the original code addr2line_binary was used for two purposes: a) to prevent fork for addr2line launching if it was already forked for the current shared object, b) launch addr2line for each shared object in backtrace. In the new version start_addr2line_fork() is invoked only from my_addr_resolve_init(), i.e. once per the whole backtrace. Suppose our backtrace contains function addresses from different shared libraries. Will it work correctly?

cvicentiu:
As @vlad-lesin pointed out, this fix is not correct:

To test this I did the following:

1. Introduced an intentional segfault in storage/connect/ha_connect.cc like so:
```diff
--- a/storage/connect/ha_connect.cc
+++ b/storage/connect/ha_connect.cc
@@ -6474,6 +6474,9 @@ int ha_connect::create(const char *name, TABLE *table_arg,
   PGLOBAL g= xp->g;
 
   DBUG_ENTER("ha_connect::create");
+  fp = (Field *) 0x01;
+  fp->field_name.length = 1;
+
   /*
     This assignment fixes test failures if some
     "ALTER TABLE t1 ADD KEY(a)" query exits on ER_ACCESS_DENIED_ERROR
```

2. Ran `./mtr --suite=connect`.

We end up with:
```
mysys/stacktrace.c:213(my_print_stacktrace)[0x55dd4cbd2d02]
sql/signal_handler.cc:241(handle_fatal_signal)[0x55dd4c26cb6a]
/usr/lib/libc.so.6(+0x38f50)[0x7fcab7c51f50]
/home/vicentiu/Workspace/MariaDB/11.0/mysql-test/var/plugins/ha_connect.so(_ZN10ha_connect6createEPKcP5TABLEP14HA_CREATE_INFO+0x129)[0x7fcab34c39db]
sql/handler.cc:5150(handler::ha_create(char const*, TABLE*, HA_CREATE_INFO*))[0x55dd4c27f573]
sql/handler.cc:5618(ha_create_table(THD*, char const*, char const*, char const*, HA_CREATE_INFO*, st_mysql_const_unsigned_lex_string*))[0x55dd4c280e78]
sql/sql_table.cc:5428(create_table_impl(THD*, st_mysql_const_lex_string const&, st_mysql_const_lex_string const&, st_mysql_const_lex_string const&, st_mysql_const_lex_string const&, char const*, DDL_options_st, HA_CREATE_INFO*, Alter_info*, int, bool*, st_key**, unsigned int*, st_mysql_const_unsigned_lex_string*))[0x55dd4bfe7d0d]
sql/sql_table.cc:5512(mysql_create_table_no_lock(THD*, st_mysql_const_lex_string const*, st_mysql_const_lex_string const*, Table_specification_st*, Alter_info*, bool*, int, TABLE_LIST*))[0x55dd4bfe814e]
sql/sql_table.cc:5616(mysql_create_table(THD*, TABLE_LIST*, Table_specification_st*, Alter_info*))[0x55dd4bfe861b]
sql/sql_table.cc:12367(Sql_cmd_create_table_like::execute(THD*))[0x55dd4bffd628]
sql/sql_parse.cc:6073(mysql_execute_command(THD*))[0x55dd4bee8cff]
sql/sql_parse.cc:8106(mysql_parse(THD*, char*, unsigned int, Parser_state*, bool, bool))[0x55dd4beef023]
sql/sql_parse.cc:1894(dispatch_command(enum_server_command, THD*, char*, unsigned int, bool, bool))[0x55dd4beda5af]
sql/sql_parse.cc:1375(do_command(THD*))[0x55dd4bed8d5e]
sql/sql_connect.cc:1416(do_handle_one_connection(CONNECT*, bool))[0x55dd4c09ba6d]
sql/sql_connect.cc:1320(handle_one_connection)[0x55dd4c09b7e9]
perfschema/pfs.cc:2203(pfs_spawn_thread)[0x55dd4c5ddbe4]
/usr/lib/libc.so.6(+0x85bb5)[0x7fcab7c9ebb5]
/usr/lib/libc.so.6(+0x107d90)[0x7fcab7d20d90]
```

Notice that the connect library was not properly resolved:
`test/var/plugins/ha_connect.so(_ZN10ha_connect6createEPKcP5TABLEP14HA_CREATE_INFO+0x129)[0x7fcab34c39db]`

Without the patch, the following result is produced instead:
`connect/ha_connect.cc:6478(ha_connect::create(char const*, TABLE*, HA_CREATE_INFO*))[0x7f9ac08c39db]`

Thus I consider this change to not be useful and the underlying MDEV has zero context about what the actual problem is. @rex-johnston can you please provide more context as to what the actual problem is by properly changing the MDEV-30815 description, creating a working patch that fixes the accurately stated problem and *then* asking for a review?

I'm closing this PR in the meantime.

PR #2537, opened 2023-03-09
WIP : [MDEV-23962] Removed obsolete code to handle 'arc' directories

is-this-echo:
- [x] *The Jira issue number for this PR is: MDEV-23962*

## Description
--**WIP**--
Deleted unused code to handle 'arc' directories in this files : 
sql/sql_db.cc: find_db_tables_and_rm_known_files()
sql/sql_db.cc: mysql_rm_arc_files()
sql/parse_file.cc:rename_in_schema_file()

## How can this PR be tested?
---

## PR quality check

### DIFF
--- a/sql/sql_db.cc
+++ b/sql/sql_db.cc
@@ -58,7 +58,6 @@ static TYPELIB deletable_extensions=
 static bool find_db_tables_and_rm_known_files(THD *, MY_DIR *, const char *,
                                               const char *, TABLE_LIST **);
 
-long mysql_rm_arc_files(THD *thd, MY_DIR *dirp, const char *org_path);
 my_bool rm_dir_w_symlink(const char *org_path, my_bool send_error);
 static void mysql_change_db_impl(THD *thd,
                                  LEX_CSTRING *new_db_name,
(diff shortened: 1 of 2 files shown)


sql/sql_db.cc
```diff
@@ -1463,82 +1443,6 @@ my_bool rm_dir_w_symlink(const char *org_path, my_bool send_error)
...
-
-  RETURN
-    > 0 number of removed files
-    -1  error
-
-  NOTE
-    A support of "arc" directories is obsolete, however this
-    function should exist to remove existent "arc" directories.
-*/
-long mysql_rm_arc_files(THD *thd, MY_DIR *dirp, const char *org_path)
```
robinnewhouse, line 1483: The code in here will need to be added in to `mysql_upgrade_db` to ensure arc directories are removed during upgrade

robinnewhouse (commented):
Hi @is-this-echo, thanks for taking a look at this. I realize this is a work in progress still, but I just wanted to point out that in [MDEV-23962](https://jira.mariadb.org/browse/MDEV-23962) the request is not only to remove the `mysql_rm_arc_files()` etc., but to also make sure that the directories are being removed at some point in `mysql_upgrade_db()`.

sql/sql_db.cc
is-this-echo: Hi @robinnewhouse, thanks for the review. Yes, I am aware that we need to add code to `mysql_upgrade` to remove existing arc directories, I am a bit busy right now, will resume working on this in a week.

client/mysql_upgrade.c
```diff
@@ -18,6 +18,9 @@
 
 #include "client_priv.h"
 #include <sslopt-vars.h>
+#include <my_dir.h>
+#include <m_ctype.h>
+#include "../include/mysql/psi/mysql_file.h"
```
grooverdan, line 23: `#include <mysql/psi/mysql_file.h>` seems sufficient elsewhere in the codebase.

client/mysql_upgrade.c
```diff
@@ -1401,6 +1410,114 @@ static int check_version_match(void)
...
+    revision= extension+5;
+    while (*revision && my_isdigit(system_charset_info, *revision))
+      revision++;
+    if (*revision)
+    {
+      found_other_files++;
+      continue;
+    }
+    strxmov(filePath, org_path, "/", file->name, NullS);
+    if (mysql_file_delete_with_symlink(key_file_misc, filePath, "", MYF(MY_WME)))
```
grooverdan, line 1460: `my_delete` is probably sufficient.  `key_file_misc` is related to peformance schema and is really only for within the server.

grooverdan (commented):
This looks just about ready right? It could to with a small test case, that we could handle if you don't have time.

mysql-test/main/mysql_upgrade.test - create some files of arc names before MYSQL_UPGRADE and see if they are removed afterwards (which mysql-test will do by default anyway).

PR #2538, opened 2023-03-10
MDEV-30825 innodb_compression_algorithm=0 (none) increments Innodb_num_pages_page_compression_error

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-30825*

## Description

fil_page_compress_low returns 0 for both innodb_compression_algorithm=0 and where there is compression errors. On the two callers to this function, don't increment the compression errors if the algorithm was none.

## How can this PR be tested?

MTR case included

## Backward compatibility

none

## PR quality check
- [ x ] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.

Marko, two questions:
* how to trigger buflush_do_batch so that `fil_page_compress_for_non_full_crc32` is triggered in a MTR test case?
* `mysql-test/mtr --gdb='b fil_page_compress_for_full_crc32;b fil_page_compress_for_non_full_crc32;r' innodb.innodb-page_compression_none` only hits  isn't hitting the `full_crc32` breakpoint, I'm sure I'm missing something obvious.


mysql-test/suite/innodb/r/innodb-page_compression_none.result
```diff
@@ -0,0 +1,20 @@
+#
+# MDEV-30825 innodb_compression_algorithm=0 (none) increments Innodb_num_pages_page_compression_error
+#
+SET GLOBAL innodb_compression_algorithm=0, innodb_checksum_algorithm=strict_full_crc32, innodb_buffer_pool_size=8*1024*1024;
```
dr-m, line 4: This is failing massively, for some reason:
```
CURRENT_TEST: innodb.innodb-page_compression_none
mysqltest: At line 8: query 'SET GLOBAL innodb_compression_algorithm=0, innodb_checksum_algorithm=strict_full_crc32, innodb_buffer_pool_size=8*1024*1024' failed: 1231: Variable 'innodb_buffer_pool_size' can't be set to the value of '8388608'
```
Wasn’t it so that in debug builds there is some extra flag that needs to be enabled to allow buffer pool resizing? Why does this test need to resize the buffer pool at all? Could the buffer pool size be specified in a `.opt` file?

mysql-test/suite/innodb/t/innodb-page_compression_none.test
```diff
@@ -0,0 +1,30 @@
...
+
+CREATE TABLE t (c INT) page_compressed=1 page_compression_level=4 ENGINE=InnoDB;
+INSERT INTO t VALUES (1);
+SET GLOBAL innodb_checksum_algorithm=crc32;
+CREATE TABLE t1 (c INT) page_compressed=1 page_compression_level=4 ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1);
+
+# do stuff to purge it out of innodb buffer pool - TODO how to do better.
+FLUSH TABLES;
+CREATE TEMPORARY TABLE b AS SELECT seq, REPEAT('aabbccddeeffgghhiijjkkllmmnn', 100000) FROM seq_1_to_1000;
```
dr-m, line 20: So, this was the reason for the buffer pool size change?
I do not know if it would work reliably in 10.4, but starting with 10.5, there is a test `mysql-test/suite/innodb/t/page_cleaner.test` that should work.
You could also simply try
```sql
FLUSH TABLES t, t1 FOR EXPORT;
UNLOCK TABLES;
```

mysql-test/suite/innodb/t/innodb-page_compression_none.test
```diff
@@ -0,0 +1,30 @@
...
+
+SELECT VARIABLE_VALUE - @compress_errors AS NUMBER_OF_ERRORS FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME = 'Innodb_num_pages_page_compression_error';
+
+SET GLOBAL innodb_checksum_algorithm=DEFAULT, innodb_compression_algorithm=DEFAULT, innodb_buffer_pool_size=DEFAULT, innodb_checksum_algorithm=DEFAULT;
```
dr-m, line 24: I think that you should simply copy `mysql-test/suite/innodb/t/table_flags.combinations` to a `.combinations` file of this test, to have separate combinations for `full_crc32` and `crc32`.

For restoring global variables, I think that a better way is
```sql
SET @save_compression_algorithm=@@GLOBAL.innodb_compression_algorithm;
SET GLOBAL innodb_compression_algorithm=0;
--…
SET GLOBAL innodb_compression_algorithm=@save_compression_algorithm;
```
That would work even when the test is being invoked with a non-default parameter, such as
```sh
./mtr --mysqld=--loose-innodb-compression-algorithm=lz4
```

dr-m (changes requested):
The code change looks reasonable to me, but I think that the test needs a little work.

mysql-test/suite/innodb/t/innodb-page_compression_none.test
```diff
@@ -0,0 +1,34 @@
...
+UNLOCK TABLES;
+
+SELECT VARIABLE_VALUE - @compress_errors AS NUMBER_OF_ERRORS FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME = 'Innodb_num_pages_page_compression_error';
+
+DROP TABLE t, t1;
+SET GLOBAL innodb_compression_algorithm=@save_compression_algorithm;
+
+# Seems to be needed reset INNODB_CHECKSUM_ALGORITHM to default
+let $shutdown_timeout = 0;
+--source include/restart_mysqld.inc
```
dr-m, line 28: Why do we even mess with `innodb_checksum_algorithm` in the test? You already added a `.combinations` file that covers both `strict_crc32` and `strict_full_crc32`.

dr-m (commented):
The code change looks OK to me, but I would avoid the kill+restart in the test, and just rely on the `.combinations` file.

mysql-test/suite/innodb/t/innodb-page_compression_none.test
grooverdan: because of hopefully temporary blindness and not being able to see the obvious in the middle of the test. Sigh. Thanks Marko.

dr-m (approved):

PR #2480, opened 2023-02-03
Minimize unsafe C functions usage - replace str…() and strn…() functions with custom safe_strcat() and safe_strcpy() functions - part 2

Chaloff:
# Description
This submission continues work started in 567b681 where `safe_strcat()` and `safe_strcpy()` functions were introduced.

The MariaDB code base uses `strcat()` and `strcpy()` in several places. These are known to have memory safety issues and their usage is discouraged. Common security scanners like Flawfinder flags them. In MariaDB we should start using modern and safer variants on these functions.

We changed old unsafe functions with the new safe ones and also use `snprintf()` in the places where series of `strcpy()/strcat()` were used – this also make code safe and fix more clear.

We would like to provide 2-3 smaller size changes to make code review and
merge easier. Next part will follow.

## How can this PR be tested?

All build stages pass for these commits.

## Basing the PR against the correct MariaDB version
- [x] *This is a bug fix and the PR is based against the earliest branch in which the bug can be reproduced*

## Backward compatibility
The changes are fully backward compatible.

All new code of the whole pull request, including one or several files
that are either new files or modified ones, are contributed under the BSD-new
license. I am contributing on behalf of my employer Amazon Web Services,
Inc.

PR #2540, opened 2023-03-11
[MDEV-30543] New status variable: max_used_connections_time

Chaloff:
- [x] *The Jira issue number for this PR is: MDEV-30543*

## Description

Max_used_connections_time - The time at which Max_used_connections reached its
current value. This will be helpful during the troubleshooting and helpful for
the user. MySQL 8 has a similar status variable

## How can this PR be tested?

Added MTR test `max_used_connections_time.test`

## Backward compatibility
The change does not affect previous branches

that are either new files or modified ones, are contributed under the BSD-new
license. I am contributing on behalf of my employer Amazon Web Services


ahmedibrahim404 (changes requested):
Otto Kekäläinen asked me to review this PR on JIRA.

Great Job! Looks very good.

I have only one comment here. Instead of returning the time in a unix timestamp format, we can convert it to MYSQL_TIME instead to be more clear (That's what MySQL does also).

One way to do so is having `show_max_used_connections_time` function to parse the unix timestamp to MYSQL_TIME

Chaloff:
> Otto Kekäläinen asked me to review this PR on JIRA.
> 
> Great Job! Looks very good.
> 
> I have only one comment here. Instead of returning the time in a unix timestamp format, we can convert it to MYSQL_TIME instead to be more clear (That's what MySQL does also).
> 
> One way to do so is having `show_max_used_connections_time` function to parse the unix timestamp to MYSQL_TIME

Hi @ahmedibrahim404 - thanks a lot for the review. Based on your comment and also on Daniel Black's comment from #2485  I added function show_max_used_connections_time() to represent  show_max_used_connections variable in string format (as suggested by Daniel) - looks like it's even more clear and user friendly.
Please review it again
Thanks in advance.
NB: there is some builds failing in `ubuntu-2204-debug-ps` with signature `worker[28] > Restart [mysqld.1 - pid: 279348, winpid: 279348] - running with different options '' != '--loose-enable-partition'` - looks like it's not related to the submission but some general infrastructure problem - I see the same signature in some other PR's now as well

sql/mysqld.cc
```diff
@@ -7004,11 +7008,24 @@ static int show_heartbeat_period(THD *thd, SHOW_VAR *var, char *buff,
   else
     var->type= SHOW_UNDEF;
   return 0;
-}
-
+}\
```
ahmedibrahim404, line 7011: I think this `\` was added by mistake, right?

ahmedibrahim404 (approved):
Except for the extra slash, everything looks great

sql/mysqld.cc
Chaloff: Yes - will remove it shortly

mysql-test/main/max_used_connections_time.test
```diff
@@ -0,0 +1,35 @@
+-- source include/not_embedded.inc
+
+--echo #
+--echo # MDEV-30543 New status variable: Max_used_connections_time
+--echo #
+
+FLUSH STATUS;
```
an3l, line 7: Do you need this at the beginning ?
Same for the next 2 lines?

sql/mysqld.cc
```diff
@@ -6173,7 +6174,10 @@ void create_new_thread(CONNECT *connect)
 
   uint sum= connection_count + extra_connection_count;
   if (sum > max_used_connections)
+  {
     max_used_connections= sum;
+    max_used_connections_time= time(nullptr);
```
an3l, line 6179: I haven't researched about the option itself, but just let me know the logic why reset here is important and how is different from `mysql_init_variables`?

sql/mysqld.cc
```diff
@@ -7006,9 +7010,22 @@ static int show_heartbeat_period(THD *thd, SHOW_VAR *var, char *buff,
   return 0;
 }
 
-
 #endif /* HAVE_REPLICATION */
 
+static int show_max_used_connections_time(THD *thd, SHOW_VAR *var, char *buff,
+                                 enum enum_var_type scope)
+{
+
```
an3l, line 7018: Extra line

sql/mysqld.cc
```diff
@@ -7006,9 +7010,22 @@ static int show_heartbeat_period(THD *thd, SHOW_VAR *var, char *buff,
...
+  char time_str[20];
+  get_date(time_str, GETDATE_DATE_TIME, max_used_connections_time);
+
+  var->type= SHOW_CHAR;
+  var->value= buff;
+
+  sprintf(buff, "%s", time_str);
```
an3l, line 7025: I quess you want to use safer version `snprintf`  here like you did with 567b68129943a1cceab1d7b4c68e2a4ba011cdc0 and PR #2480 ?

mysql-test/main/max_used_connections_time.test
```diff
@@ -0,0 +1,35 @@
...
+--sleep 1
+connect (con3,localhost,root,,);
+--sleep 1
+connect (con4,localhost,root,,);
+--sleep 1
+let $time_3=`SELECT VARIABLE_VALUE FROM INFORMATION_SCHEMA.SESSION_STATUS WHERE VARIABLE_NAME LIKE 'max_used_connections_time'`;
+--echo # This should report 1 as connection number was increased
+let $dif_2= `SELECT '$time_1' <> '$time_3'`;
+--echo $dif_2
+
```
an3l, line 32: Please add `show status` query example.

mysql-test/main/max_used_connections_time.test
```diff
@@ -0,0 +1,35 @@
+-- source include/not_embedded.inc
```
an3l, line 1: I would move this test to `main/status.test` where `max_used_connections` tests are.

sql/mysqld.cc
```diff
@@ -7006,9 +7010,22 @@ static int show_heartbeat_period(THD *thd, SHOW_VAR *var, char *buff,
...
+  return 0;
+}
+
```
an3l, line 7028: Functions should be diveded with 2 spaces(  we need to update that in Coding standard) @LinuxJedi

sql/mysqld.cc
```diff
@@ -7006,9 +7010,22 @@ static int show_heartbeat_period(THD *thd, SHOW_VAR *var, char *buff,
   return 0;
 }
 
-
 #endif /* HAVE_REPLICATION */
 
+static int show_max_used_connections_time(THD *thd, SHOW_VAR *var, char *buff,
```
an3l, line 7015: Functions should be diveded with 2 spaces ( I think we need to update that in Coding standard) @LinuxJedi

mysql-test/main/max_used_connections_time.test
```diff
@@ -0,0 +1,35 @@
...
+connect (con4,localhost,root,,);
+--sleep 1
+let $time_3=`SELECT VARIABLE_VALUE FROM INFORMATION_SCHEMA.SESSION_STATUS WHERE VARIABLE_NAME LIKE 'max_used_connections_time'`;
+--echo # This should report 1 as connection number was increased
+let $dif_2= `SELECT '$time_1' <> '$time_3'`;
+--echo $dif_2
+
+disconnect con4;
+disconnect con3;
+disconnect con1;
```
an3l, line 35: Adding the last empty line (this also should be added in Coding standard) @LinuxJedi .

sql/mysqld.cc
```diff
@@ -7314,7 +7331,6 @@ static int show_memory_used(THD *thd, SHOW_VAR *var, char *buff,
   return 0;
 }
 
-
```
an3l, line 7317: Not needed.

mysql-test/main/max_used_connections_time.test
```diff
@@ -0,0 +1,35 @@
...
+--echo # This should report 0 as it must have the same value as above
+let $dif_1= `SELECT '$time_1' <> '$time_2'`;
+--echo $dif_1
+--sleep 1
+connect (con3,localhost,root,,);
+--sleep 1
+connect (con4,localhost,root,,);
+--sleep 1
+let $time_3=`SELECT VARIABLE_VALUE FROM INFORMATION_SCHEMA.SESSION_STATUS WHERE VARIABLE_NAME LIKE 'max_used_connections_time'`;
+--echo # This should report 1 as connection number was increased
```
an3l, line 29: Missing test case of setting the value and generating the error message.

an3l (changes requested):
Hi @Chaloff,
thanks for the patch.
I gave brief review of current patch, since I haven't looked into how option is supposed to work itself, so I will add another round of review :).
Can you please add pointers and description in JIRA about MySQL usage and how option is expected to work with and without usage of `Max_used_connections` ?

mysql-test/main/max_used_connections_time.test
dlenski: `max_used_connections_time` appears in `SHOW STATUS` output; it is not a system variable so there is no SQL-level mechanism to even attempt to set it.

Are you sure this is relevant?

dlenski: You mentioned [`main/status.test`](https://github.com/MariaDB/server/blob/e416fb850bf06399201d59a5c711b7ff754e1d2f/mysql-test/main/status.test#L423-L431), where there are similarly no attempts to `SET` the values of the "status variables".

dlenski: I don't think I understand what you're asking for here.

This PR adds _one_ status variable, `max_used_connections_time`, with one clear purpose and simple implementation: "record the timestamp at which `max_used_connections` was last set."

@Chaloff has correspondingly added one test, which functionally verifies the behavior of this status variable.

mysql-test/main/max_used_connections_time.test
grooverdan: yes, other tests may have a higher number of max connections. Check `FLUSH STATUS` actually resets the value. I suspect it may not.

mysql-test/main/max_used_connections_time.test
grooverdan: There's no added error message. I don't understand either. I think the test case looks fine, though moving it with other test cases would be good.

mysql-test/main/max_used_connections_time.test
grooverdan: AFAIK this isn't a standard.

sql/mysqld.cc
grooverdan: This isn't a reset - see `man 2 time`. Look at `threadpool_add_connection` - `thd->start_utime` already contains the time, so using this condition to set `max_used_connections*` that after `connect->scheduler->add_connection(connect);` which set this this would avoid a system call.

sql/mysqld.cc
grooverdan: or better, why not just `get_date` straight into `buff` and avoid the copy.

sql/mysqld.cc
```diff
@@ -9200,6 +9218,7 @@ void refresh_status(THD *thd)
     connections.  This is not perfect, but status data is not exact anyway.
   */
   max_used_connections= connection_count + extra_connection_count;
+  max_used_connections_time= time(nullptr);
```
grooverdan, line 9221: could use `thd->{timevar}` here instead of `time` too.

dlenski: @grooverdan, which of the (many :cold_sweat:) timestamps in `class THD` would be relevant here?

https://github.com/mariadb/server/blob/11.0/sql/sql_class.h#L2854-L2859

sql/mysqld.cc
dlenski: > Functions should be diveded with 2 spaces ( I think we need to update that in Coding standard) @LinuxJedi

It already explains that blocks should be indented with 2 spaces: https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md#indentation

Do you mean that functions should be _separated_ with 2 _blank lines_? If so, yeah, that should be added to CODING_STANDARDS.md.

mysql-test/main/max_used_connections_time.test
dlenski: I suggested to @Chaloff that he should change this to `--eval SELECT '$time_3' > '$time_1' as expect_1`, so that the expected value of the comparison is apparent in _both_ the `.test` and the `.result` files.

.test will have:
```
--replace_regex /…/TIMESTAMP/
--eval SELECT '$time_3' > '$time_1' as expect_1
```

.result will have:
```
SELECT 'TIMESTAMP' > 'TIMESTAMP' as expect_1
expect_1
1
```

Chaloff: Yes, thanks - will update PR shortly

mysql-test/main/max_used_connections_time.test
dlenski: I'll put in a plug here for using [wtf](https://github.com/dlenski/wtf), which will automatically clean up whitespace inconsistencies including this one: in its default mode it will ensure that the final line ends with `\n`.

sql/mysqld.cc
Chaloff: > This isn't a reset - see `man 2 time`. Look at `threadpool_add_connection` - `thd->start_utime` already contains the time, so using this condition to set `max_used_connections*` that after `connect->scheduler->add_connection(connect);` which set this this would avoid a system call.

Thanks @grooverdan for the review, I addressed almost everything reviewers asked for except this particular comment - despite comments for the function `create_new_thread()` it does not contains `thd` parameter - the signature is `void create_new_thread(CONNECT *connect)` Socan you give me a hint how can I get `start_utime` parameter here?

mysql-test/main/max_used_connections_time.test
grooverdan: final line ending with `\n` makes sense, more than an empty blank one anyway.

sql/mysqld.cc
dlenski: > I was looking a lower overhead

Consulting the `SHOW STATUS` output is not in any plausible circumstance a fast-path operation… is it?

sql/mysqld.cc
grooverdan: `connect->prior_thr_create_utime` doesn't seem to work, so leave as is with `time`

an3l (changes requested):

mysql-test/main/status.result
```diff
@@ -447,3 +447,22 @@ Feature_json	2
...
+# MDEV-30543 New status variable: Max_used_connections_time
+#
+FLUSH STATUS;
+connect  con1,localhost,root,,;
+connect  con2,localhost,root,,;
+connection con1;
+disconnect con2;
+SELECT 'DTVALUE' = 'DTVALUE' AS expect_1;
+expect_1
+1
```
dlenski, line 460: Great! Much easier to interpret both the `.result` file and the `.test` file in this form.

grooverdan (approved):

grooverdan:
Thank you for writing it.

PR #2542, opened 2023-03-11
MDEV-30351 crash in Item_func_left::val_str

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-30351

## Description
crash in `Item_func_left::val_str`
## How can this PR be tested?
The test is appended to `func_str`

## PR quality check

### DIFF
--- a/mysql-test/main/func_str.result
+++ b/mysql-test/main/func_str.result
@@ -5240,6 +5240,15 @@ DROP TABLE crash_test_2;
 # Start of 10.4 tests
 #
 #
+# MDEV-30351 crash in Item_func_left::val_str
+#
+SELECT WEIGHT_STRING('aa') IN (LEFT(WEIGHT_STRING('aaa'),4),'bbb') as expect_1;
+expect_1
+1
+SELECT UNHEX('0032') in (LEFT(UNHEX('003200'), 2),'dog') as expect_1;
+expect_1
+1
+#
 # MDEV-21841 CONV() function truncates the result type to 21 symbol.
 #
 CREATE TABLE t1(i BIGINT);
--- a/mysql-test/main/func_str.test
+++ b/mysql-test/main/func_str.test
@@ -2256,15 +2256,21 @@ CREATE TABLE crash_test_2 (
 DROP TABLE t1;
 DROP TABLE crash_test_2;
 
-
 --echo #
 --echo # End of 10.3 tests
 --echo #
 
-
 --echo #
 --echo # Start of 10.4 tests
 --echo #
+
+--echo #
+--echo # MDEV-30351 crash in Item_func_left::val_str
+--echo #
+
+SELECT WEIGHT_STRING('aa') IN (LEFT(WEIGHT_STRING('aaa'),4),'bbb') as expect_1;
+
+SELECT UNHEX('0032') in (LEFT(UNHEX('003200'), 2),'dog') as expect_1;
 
 --echo #
 --echo # MDEV-21841 CONV() function truncates the result type to 21 symbol.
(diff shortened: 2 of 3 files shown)


Weijun-H:
I checked the failed tests, and they seem not to relate to my modification.

grooverdan (changes requested):
First mistake:

```
/home/dan/repos/mariadb-server-10.3/sql/item_strfunc.cc:1650:8: error: variable 'char_pos' is used uninitialized whenever '&&' condition is false [-Werror,-Wsometimes-uninitialized]
      (res->charset() != NULL && (res->length() <= (char_pos= res->charpos((int) length)))))
       ^~~~~~~~~~~~~~~~~~~~~~
/home/dan/repos/mariadb-server-10.3/sql/item_strfunc.cc:1653:26: note: uninitialized use occurs here
  tmp_value.set(*res, 0, char_pos);
                         ^~~~~~~~
/home/dan/repos/mariadb-server-10.3/sql/item_strfunc.cc:1650:8: note: remove the '&&' if its condition is always true
      (res->charset() != NULL && (res->length() <= (char_pos= res->charpos((int) length)))))
       ^~~~~~~~~~~~~~~~~~~~~~~~~
/home/dan/repos/mariadb-server-10.3/sql/item_strfunc.cc:1641:16: note: initialize the variable 'char_pos' to silence this warning
  uint char_pos;
               ^
                = 0
1 error generated.
```

Even after initializing char_pos to 0, while this prevents a crash, it doesn't generate the right result. Tracing this through it returns a empty string. It should return 10 characters of the timestamp (as examined in gdb around the changed code).

I got a solution of the following:
```
diff --git a/sql/item_strfunc.cc b/sql/item_strfunc.cc
index 90bc784ed01..5be5b8da82a 100644
--- a/sql/item_strfunc.cc
+++ b/sql/item_strfunc.cc
@@ -1646,8 +1646,10 @@ String *Item_func_left::val_str(String *str)
   /* if "unsigned_flag" is set, we have a *huge* positive number. */
   if ((length <= 0) && (!args[1]->unsigned_flag))
     return make_empty_result();
-  if ((res->length() <= (ulonglong) length) ||
-      (res->charset() != NULL && (res->length() <= (char_pos= res->charpos((int) length)))))
+  if (res->charset() == NULL)
+    char_pos= length;
+  else if ((res->length() <= (ulonglong) length) ||
+      (res->length() <= (char_pos= res->charpos((int) length))))
     return res;
 
   tmp_value.set(*res, 0, char_pos);
```

Can you construct a test case that effectively tests the result?

Can you put a [commit message](https://mariadb.com/kb/en/contributing-code/#what-to-have-in-a-commit-comment) that describes the problem/solution more fully.

Can you rebase on 10.4 (we might cherry pick to 10.3 later)?

grooverdan:
Simpler test: `SELECT UNHEX('0032') in (LEFT(UNHEX('003200'), 2),'dog');` returning 1 - crashes previous.

Getting a second review from @abarkov so wait off on test changes.

abarkov (changes requested):
It is not expected that `String::charset()` returns NULL.
The problem should be fixed in other place, where this String instance is initialized.

From a glance, `str->set_charset(&my_charset_bin)` should be added to `Item_func_unhex::val_str()`.

grooverdan:
> However, even if we fix `in_string`, the call for `str->set_charset(&my_charset_bin)` should still be added to `Item_func_unhex::val_str()`, because `str->m_charset` can be set to any arbitrary value when `str` is passed to `Item_func_unhex::val_str()`

I assume the same applies to `Item_func_weight_string::val_str` that was the original trigger in the MDEV too.

abarkov:
Correct. Any val_str() must set str->m_charset according to Item::collation.collation.

grooverdan:
Thanks @Weijun-H ,I just did some minor adjustments on test cases to show for WEIGHT_STRING that its returning the right value and listed the functions (and reviewers) in the commit message.

Thanks for making the change to fix this bug. I'll merge this once the CI passes the required tests.

PR #2543, opened 2023-03-11
Fix trivial spelling errors

ottok:
## Description

- agressively -> aggressively
- exising -> existing
- occured -> occurred
- releated -> related
- seperated -> separated
- sucess -> success
- use use -> use

## How can this PR be tested?

Most changes are pure text changes with no functional changes. If some strings in the output have changed, but if full MTR passes then also the equivalent string updates in tests were correct.

Most of these typos have been introduced after MariaDB 10.6. Basing submission on 10.11 as seems most fit in this case.

## Backward compatibility

String changes are not part of any API, should be fully backwards compatible.

## PR quality check


grooverdan:
Thanks

PR #2545, opened 2023-03-13
MDEV-30778: Remove Awk from mysql_install_db

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-30778*

## Description
Pull request reduces need of (G)AWK-command at least for Debian `mariadb-server-compat` package in `scripts/mysql_install_db.sh` script.

AWK command is replace by purely Posix sh compiliant version.

## How can this PR be tested?
Test script `mysql_install_db` before and after PR installation should go smoothly

## Backward compatibility
Should not affect backward compatibility

## PR quality check


ottok:
See also #2549 on the same topic. I am in favor of removing `awk` so we get rid of extra dependency. For this I have the same question as in the other - how did you test this, can you "prove" that the syntax is correct?

illuusio:
In my testing this didn't cause problems. I know as it's very mature script this kind of change can cause some backsplash effect.

This is proof that can be verified:

Script runs command (most of the times it's localhost):
```
/usr/bin/resolveip localhost
```
which is part of MariaDB and can be found on `extra/resolveip.c`
it outputs (you can test with it as it's installed) string like
```
IP address of localhost is 127.0.0.1
IP address of localhost is 127.0.0.1
```
AWK makes then (It's sixth on regex of space):
```
echo "$(/usr/bin/resolveip localhost)" | awk '/ /{print $6}'
127.0.0.1
127.0.0.1
```
Bash version is little stupid but also robust. It read line and remove all before last space so result should be the same in AWK. It can be done with cut also but I don't kwow if it's better.
```
echo  "$(/usr/bin/resolveip localhost)" | while read a; do echo ${a##* }; done
127.0.0.1
127.0.0.1
```

grooverdan (approved):
Thanks

PR #2546, opened 2023-03-13
Make 'move_file' command more reliable in mysqltest

lornacluo:
## Describe
The tests innodb.import_tablespace_race, innodn.restart, and innodb.innodb-wl5522 move the tablespace file between the data directory and the tmp directory specified by global environment variables. The code looks like: `--move_file $tmpdir/file $datadir/file`

However this is risky because it's not unusual that the set tmp directory (often under /tmp) is mounted on another disk partition or device, and 'move_file' command may fail with "Errcode: 18 'Invalid cross-device link'".

A screenshot of the error:

> CURRENT_TEST: innodb.import_tablespace_race

> /local/.../client//mariadb-test: Error on rename of '/tmp/1TzsNV4ZrR/7/t.cfg' to '/local/.../mysql-test/var/7/mysqld.1/data//test/t.cfg' (Errcode: 18 "Invalid cross-device link")

> mysqltest: At line 50: command "move_file" failed with error: 1  my_errno: 18  errno: 18

> The result from queries just before the failure was:

> < snip >
> ENGINE=InnoDB CHARSET latin1;
> INSERT INTO t SELECT seq, 'x' FROM seq_1_to_100;
> connect  con1,localhost,rdsadmin,,test;
> BEGIN NOT ATOMIC
> DECLARE a INT DEFAULT 0;
> REPEAT
> SET a= a+1;
> UPDATE t SET c = 'xx' WHERE pk = a;
> UNTIL a = 100
> END REPEAT;
> END
> $
> connection default;
> ALTER TABLE t NOWAIT ADD INDEX (c);
> connection con1;
> connection default;
> FLUSH TABLE t FOR EXPORT;
> UNLOCK TABLES;
> DROP TABLE t;
> ALTER TABLE t DISCARD TABLESPACE;

> More results from queries before failure can be found in /local/.../mysql-test/var/7/log/import_tablespace_race.log

To stabilize mysqltest in the described scenario, when move_file errors with error code
EXDEV, attempt copy file followed by remove file. If copy + remove is success, consider
move_file also success; if copy + remove errors, show error message of the original
move_file command and error code EXDEV.

## How to test this PR
Mount /tmp to a different disk partition and set $MYSQL_TMP_DIR to it, and then run the tests to reproduce the error. After the code change, the tests should pass.

Or `./mtr innodb.import_tablespace_race --tmpdir=/dev/shm/tmp`

- [x] This is a bug fix and the PR is based against the earliest branch in which the bug can be reproduced and bug fixes are still being accepted.

## Copyright
All new code of the whole pull request, including one or several files that
are either new files or modified ones, are contributed under the BSD-new license.
I am contributing on behalf of my employer Amazon Web Services, Inc.

### DIFF
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -3814,9 +3814,21 @@ void do_move_file(struct st_command *command)
                      sizeof(move_file_args)/sizeof(struct command_arg),
                      ' ');
 
-  if (bad_path(ds_to_file.str))
-    DBUG_VOID_RETURN;
+  size_t from_plen = strlen(ds_from_file.str);
+  size_t to_plen = strlen(ds_to_file.str);
+  const char *vardir= getenv("MYSQLTEST_VARDIR");
+  const char *tmpdir= getenv("MYSQL_TMP_DIR");
 
+  if (!((is_sub_path(ds_from_file.str, from_plen, vardir) && 
+        is_sub_path(ds_to_file.str, to_plen, vardir)) || 
+        (is_sub_path(ds_from_file.str, from_plen, tmpdir) && 
+        is_sub_path(ds_to_file.str, to_plen, tmpdir)))) {
+        report_or_die("Paths '%s' and '%s' are not both under MYSQLTEST_VARDIR '%s'"
+                "or both under MYSQL_TMP_DIR '%s'",
+                ds_from_file, ds_to_file, vardir, tmpdir);
+        DBUG_VOID_RETURN;
+  }
+  
   DBUG_PRINT("info", ("Move %s to %s", ds_from_file.str, ds_to_file.str));
   error= (my_rename(ds_from_file.str, ds_to_file.str,
                     MYF(disable_warnings ? 0 : MY_WME)) != 0);
--- a/mysql-test/suite/innodb/r/innodb-wl5522.result
+++ b/mysql-test/suite/innodb/r/innodb-wl5522.result
@@ -63,16 +63,24 @@ a	b	c
 # Done restarting server
 # List before t1 DISCARD
 db.opt
+t1.cfg.sav
 t1.frm
 t1.ibd
+t1.ibd.sav
+t2.cfg.sav
 t2.frm
 t2.ibd
+t2.ibd.sav
 ALTER TABLE t1 DISCARD TABLESPACE;
 # List after t1 DISCARD
 db.opt
+t1.cfg.sav
 t1.frm
+t1.ibd.sav
+t2.cfg.sav
 t2.frm
 t2.ibd
+t2.ibd.sav
 ALTER TABLE t1 IMPORT TABLESPACE;
 ALTER TABLE t1 ENGINE InnoDB;
 SELECT COUNT(*) FROM t1;
@@ -90,10 +98,14 @@ a	b	c
 638	Cavalry	..asdasdfaeraf
 db.opt
 t1.cfg
+t1.cfg.sav
 t1.frm
 t1.ibd
+t1.ibd.sav
+t2.cfg.sav
 t2.frm
 t2.ibd
+t2.ibd.sav
 SELECT COUNT(*) FROM t1;
 COUNT(*)
 640
@@ -112,7 +124,9 @@ ALTER TABLE t2 ROW_FORMAT=DYNAMIC;
 ALTER TABLE t2 DISCARD TABLESPACE;
 # List after t2 DISCARD
 db.opt
+t2.cfg.sav
 t2.frm
+t2.ibd.sav
 ALTER TABLE t2 IMPORT TABLESPACE;
 ERROR HY000: Schema mismatch (Table flags don't match, server table has 0x21 and the meta-data file has 0x1; .cfg file uses ROW_FORMAT=COMPACT)
 ALTER TABLE t2 IMPORT TABLESPACE;
--- a/mysql-test/suite/innodb/t/import_tablespace_race.test
+++ b/mysql-test/suite/innodb/t/import_tablespace_race.test
@@ -36,8 +36,8 @@ ALTER TABLE t NOWAIT ADD INDEX (c);
 
 FLUSH TABLE t FOR EXPORT;
 --let $create= query_get_value(SHOW CREATE TABLE t, Create Table, 1)
---copy_file $datadir/test/t.cfg $MYSQL_TMP_DIR/t.cfg
---copy_file $datadir/test/t.ibd $MYSQL_TMP_DIR/t.ibd
+--copy_file $datadir/test/t.cfg $datadir/test/t.cfg.sav
+--copy_file $datadir/test/t.ibd $datadir/test/t.ibd.sav
 UNLOCK TABLES;
 
 DROP TABLE t;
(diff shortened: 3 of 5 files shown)


dr-m (changes requested):
I would prefer a lower-level change. This would seem to make tests run slower when both files are within the same file system, which is a rather usual case, and I believe always the case on our CI systems.

Could we instead adjust the implementation of `move_file` in `client/mysqltest.cc` (`do_move_file()`) so that it would resort to copy&remove when `my_rename()` returns this type of an error?

lornacluo:
> Could we instead adjust the implementation of move_file in client/mysqltest.cc (do_move_file()) so that it would resort to copy&remove when my_rename() returns this type of an error?

Updated.

dr-m:
Thank you. The fix looks reasonable to me. I guess that this could be tested on a typical GNU/Linux system by executing the following:
```sh
MYSQL_TMP_DIR=/var/tmp ./mtr --mem
```
I did not test it myself. I reassigned the review to @vuvova.

lornacluo:
Thanks for the review. FYI: `./mtr --mem` doesn't reproduce the error on my machine, and `./mtr innodb.import_tablespace_race --tmpdir=/dev/shm/tmp` does. I'll add this into PR description.

vuvova:
wouldn't it be simpler to fix the test to not move files to `$MYSQL_TMP_DIR` ?
your fix makes mysqltest more smart, and that kind of encourages writing sloppy tests that sometimes rename a file but sometimes start copying it over for no good reason. I'd simply fix the test like
```diff
---copy_file $datadir/test/t.cfg $MYSQL_TMP_DIR/t.cfg
+--copy_file $datadir/test/t.cfg $datadir/t.cfg.sav
```

LornaL1900:
Yes I'd be glad to fix the test too by making those move-file commands under datadir context. But that doesn't do anything with tests in the future - developers should try to avoid the risky behavior but sometimes they just don't know... the test may pass on their machine and fail on others machine. so it was also reasonable for @dr-m to request a lower level change I think.

lornacluo:
Yes I'd be glad to fix the test too by making those move-file commands under datadir context. But that doesn't do anything with tests in the future - developers should try to avoid the risky behavior but sometimes they just don't know... the test may pass on their machine and fail on others machine. so it was also reasonable for @dr-m to request a lower level change I think.-Wno-unused-but-set-variable

vuvova:
code-wise the change looks fine. My concern is that, tests shouldn't be moving files between filesystems at all. grep for `move_file` in tests, You'll see lines like
```
mysql-test/suite/innodb/t/truncate.test:--move_file $MYSQLD_DATADIR/test/hidden.frm $MYSQLD_DATADIR/test/t1.frm
mysql-test/suite/innodb/t/default_row_format_compatibility.test:--move_file $MYSQLD_DATADIR/tab.cfg $MYSQLD_DATADIR/test/tab.cfg
mysql-test/suite/innodb/t/log_file_name.test:--move_file $MYSQLD_DATADIR/test/t.ibd $MYSQLD_DATADIR/test/t2.ibd
mysql-test/suite/encryption/t/innochecksum.test:--move_file $MYSQLD_DATADIR/test/t4.ibd.backup $MYSQLD_DATADIR/test/t4.ibd
```
etc.
It was one mistake in one test to move files to tmpdir. Perhaps, a low-level fix would be a check to ensure that move_file always renames files within the same dirname? Something like
```c
dirname_from=dirname_length(ds_from_file.str);
dirname_to=dirname_length(ds_to_file.str);
if (dirname_from != dirname_to || strncmp(ds_from_file.str, ds_to_file.str, dirname_from))
  ... error ...
  ```

lornacluo:
Yes that makes sense to me as a great measure to ensure code quality. But I'm a little not aure: is it a little bit too much to error on this? This is improper behavior but still is different from what errors feel like. And if we do that there might be tens of tests affected, most pf which now work fine.

If you are sure about this yes I can work with that!😊

vuvova:
Let me describe my whole line of thought, so that you could question it at any point:
* arbitrarily moving files in tests is bad, because the behavior is unpredictable, depending on the environment it'll be either a instant rename or a slow file copy, we don't want it to happen for big files.
* almost all other cases of `move_file` (that I've checked) rename files within the datadir, cannot possibly cross filesystem boundaries
* there was one single case where a developer moved files to tmpdir, apparently, without giving it much thought, as the test itself doesn't require specifically tmpdir.
* so it's trivial to fix your immediate problem (failing test) — don't move to tmpdir, move within datadir like other tests
* but you're right, the command is error-prone, it's easy to do a mistake like above, someone will definitely do it again
* allowing `move_file` to work across filesystem boundaries brings up the problem from the point one
* how can we make the command less error-prone?
* if it can only rename files, the ideal syntax would be, like `--rename_file $datadir/test/t1.ibd t1.ibd.sav` that is, rename the command to "rename_file" and specify the second argument without a path. Then it'd be impossible to do a mistake.
* but that would cause lots of changes across test files.
* a smaller change would be, as I proposed above, to make sure that "dirname" part of both names is the same
* few tests will need to be adjusted, they do, like `--move_file $datadir/test/t1.idb $datadir/t1.ibd` this will no longer work.
* another approach could be to require that all paths are always within datadir. Then the test above will work too. But there are tests that move files within vardir, so this would be too restrictive.

and this all converged on the approach I put in my previous comment.

_correction: there is more than one case of the datadir<->tmpdir move, but still they're in a clear minority_

FooBarrior:
A combination of three conditions can be permissive enough, yet safe:
+ Moving across datadir is allowed
+ Moving across vardir is allowed
+ Moving across a directory is allowed

Also, for a total happiness an unsafe version can be added

lornacluo:
@vuvova Thanks for the explanation! 

I think only 3 tests are really concerning: **innodb.import_tablespace_race**, **innodb.restart**, **innodb.innodb-wl5522**. 

In for example **rpl.rpl_binlog_index**, it has something like `--move_file $master_datadir/master-bin.000001 $tmpdir/master-bin.000001` but `$tmpdir` is set under `MYSQLTEST_VARDIR` with `--let $tmpdir= $MYSQLTEST_VARDIR/tmp/$dirname`, and the use of different directories seems necessary in this test so it looks alright.

For **innodb.import_tablespace_race** and **innodb.innodb-wl5522**, relocation is not necessary so in my new commit I just fixed them by renaming under the same directory.

**innodb.restart** however is a little tricky. In this query `eval CREATE TABLE td(a INT)ENGINE=InnoDB ROW_FORMAT=DYNAMIC STATS_PERSISTENT=0 DATA DIRECTORY='$MYSQL_TMP_DIR';`, the data directory cannot be anywhere under $datadir, as described in [MDEV-13285](https://jira.mariadb.org/browse/MDEV-13285). In this case it seems reasonable to use $tmpdir... not sure there's an easy way to fix without rewriting much of the test. It should be working fine with the code change in my first commit so I didn't remove the commit yet. If you have any idea how to fix **innodb.restart** we can remove it after that.

Except for these one or two cases, your idea makes perfect sense to me. I attempted something like, check if from_file and to_file are under the same directory, if not, instead of aborting with error, print some warning like `fprintf(stderr, "Please avoid moving files across directory unless it's essential to testing");` and let people handle it. Unfortunately this doesn't work and there seems to be no easy way to print a warning in mysqltest? `sql_print_warning()` doesn't work either. Maybe we should keep it simple and just fix the test for now?

vuvova:
Only fixing tests doesn't prevent the problem from happening again in the future, as you wrote earlier.

Currently, `mysqltest` has the "bad_path" check that checks that the path is either under `MYSQLTEST_VARDIR` or under `MYSQL_TMP_DIR`. We can add another check that for `--move_file` both source and destination are either both under `MYSQLTEST_VARDIR` or both under `MYSQL_TMP_DIR`. Two problems with that:
* few tests move from datadir (under vardir) to `$MYSQL_TMP_DIR`. To fix this they need to be changed to move to `$MYSQLTEST_VARDIR/tmp`
* normally MYSQL_TMP_DIR is under MYSQLTEST_VARDIR, so the check won't detect incorrect usage. To fix this we'll make buildbot to run `mtr` with `--tmpdir` outside of vardir.

lornacluo:
@vuvova The idea looks super reasonable to me! Thanks! Codes updated.

vuvova (approved):
looks good to me!

PR #2548, opened 2023-03-13
[WIP] MDEV-17398: Implement ST_IsValid Spatial (GIS) function

ahmedibrahim404:
- [x] *The Jira issue number for this PR is: MDEV-17398*

## Description
Implement `ST_ISVALID` function as part of Spatial (GIS) functions that are included in MySQL and not in MariaDB
To Implement that, I have followed OGC Specifications in Geometries, so

*If it's a polygon:*
- The polygon must have at least four coordinates.
- The polygon must not have duplicate vertices.
- The polygon must be simple, meaning that the edges do not cross and the interior does not intersect itself.

*Point/Line:*
- A point geometry is considered valid if it has exactly one coordinate, while a MultiPoint geometry is considered valid if it has at least one point.
- Similarly, a LineString or MultiLineString geometry is considered valid if it has at least two coordinates

## How can this PR be tested?
I have added some tests to `gis.test` and their results in `gis.result` and made sure it's working

## Backward compatibility
This shouldn't break any previous versions.

## PR quality check


ahmedibrahim404:
Hello,

After some investigation, I figured out that for polygons, we still need one more condition (so I made the PR WIP) which is:
- Check for multi-rings polygon that its valid (One exterior and the rest are interiors)

I will try to implement this and push it asap

sql/spatial.cc
```diff
@@ -3066,6 +3191,50 @@ bool Gis_multi_polygon::get_data_as_json(String *txt, uint max_dec_digits,
...
+
+    Geometry *temp;
+    swkb->set_charset(&my_charset_bin);
+    swkb->length(0);
+
+    if(this->geometry_n(i, swkb))
+      return 1;
+
+    temp= Geometry::construct(&buff_temp, swkb->ptr(), swkb->length());
+    if(!temp)
```
grooverdan, line 3220: below doesn't free `delete swkb`

sql/item_geofunc.cc
```diff
@@ -37,6 +37,7 @@
 #include "set_var.h"
 #ifdef HAVE_SPATIAL
 #include <m_ctype.h>
+#include <stack>
```
grooverdan, line 40: what uses `stack`?

sql/spatial.h
```diff
@@ -289,6 +289,7 @@ class Geometry
   virtual int geom_length(double *len, const char **end) const  { return -1; }
   virtual int area(double *ar, const char **end) const { return -1;}
   virtual int is_closed(int *closed) const { return -1; }
+  virtual int is_valid(int *valid) const { return -1; }
```
grooverdan, line 292: I'm confused by the interface that accepts a pointer that seems to in most case be populated with the negated value of the return. Is there a reason both are needed? If not, just return a value.

sql/spatial.cc
```diff
@@ -3538,6 +3707,50 @@ bool Gis_geometry_collection::get_data_as_json(String *txt, uint max_dec_digits,
...
+      return 1;
+
+    int tempValid;
+    if(temp->is_valid(&tempValid))
```
grooverdan, line 3740: style nit: space between `if` and `(condition)` please.

sql/item_geofunc.cc
```diff
@@ -2061,6 +2062,23 @@ String *Item_func_buffer::val_str(String *str_value)
...
+  Geometry *g;
+
+  if ((args[0]->null_value || !(g= Geometry::construct(&buffer, swkb->ptr(), swkb->length()))))
+    return -1;
+
+  int valid;
+  if(g->is_valid(&valid))
+    return -1;
+
+  // Simplcity should be removed from here and checked for each polygon alone
```
grooverdan, line 2079: Simpl[e] typo

grooverdan (commented):
First draft review.

Please include a full commit message - https://mariadb.com/kb/en/contributing-code/#what-to-have-in-a-commit-comment

Good test of tests, well done.

sql/spatial.h
ahmedibrahim404: `is_valid` returns 0 if the function completed successfully (regardless the geometry is valid or not), and returns 1 otherwise (Any error happens inside). This same logic exist in all other functions like `is_closed`

Then we check in `Item_func_isvalid::val_int()`
```cpp
if(g->is_valid(&valid))
  return -1;
```
If it returns true (error happened) just return -1 in `Item_func_isvalid::val_int()`

sql/item_geofunc.cc
ahmedibrahim404: Done

sql/spatial.cc
ahmedibrahim404: Done

sql/item_geofunc.cc
ahmedibrahim404: Removed.

sql/spatial.cc
ahmedibrahim404: Done

grooverdan:
Thank you for your contribution.  st_isvalid was implemented as part of a broader project for Google Summer of Code in 2024 and is now merged in 1656b5c10f27.

PR #2550, opened 2023-03-14
MDEV-30839: Improvements to mini-benchmark (add CLI options and fixes)

uoysip:
## Description

[MDEV-30839](https://jira.mariadb.org/browse/MDEV-30839)

[`mini-benchmark.sh`](https://github.com/MariaDB/server/blob/10.8/support-files/mini-benchmark.sh) has been useful for locally-run benchmarks, recently it has been used to provide benchmarks for [MDEV-30628](https://jira.mariadb.org/browse/MDEV-30628). However, using the script can be tedious since it requires manually modifying the code to change the benchmark configuration. Also, the current implementation doesn't save standard output of the script (`lscpu` and other output is lost).

This PR adds command line options that address the above issues and improve usage of the script.

##### The following options have been added:

- _`--log`_: If set, logs `mini-benchmark.sh` output (`lscpu`, etc.) to a file within the benchmarks folder.
- _`--duration`_: Sets the duration in seconds.
- _`--name`_: Sets the name of the benchmark (for the folder name and _`--log`_ file).
- _`--workload`_: The sysbench workload (`oltp_read_write`, `oltp_write_only`, etc.).
- _`--threads`_: The threads to run the benchmark with.

##### Besides the command line options, three minor issues have been fixed:

- The condition for the cycle count based verdict originally checked if the cycle count was greater than 850 billion, but this was always false because `$RESULT` is only 3 digits and represents the number of cycles _in billions_.

- The `3000px` width of generated flamegraphs (if _`--perf-flamegraph`_ is used) causes the SVG's to be cut-off horizontally in the browser on most screen sizes.

- [ShellCheck](https://github.com/koalaman/shellcheck) warnings and suggestions.

- The condition to check if `perf` had permission to run on the system had a bug that caused `perf` to always be disabled.

## How can this PR be tested?

The command line options were tested using valid values/formats as documented in _`mini-benchmark.sh --help`_. When none of the options are used the behaviour of the script is unchanged (besides changes due to the minor fixes in the section above).

The script was passed through [ShellCheck](https://github.com/koalaman/shellcheck) and had no new errors/warnings (and all pre-existing warnings/suggestions were resolved).

- [x] This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced

This PR adds new functionality _as well as bug fixes_. ~It is based on the 10.8 branch which introduced `mini-benchmark.sh`.~ I have rebased this PR to be based on the 10.11 branch.

## PR quality check

## Copyright

### DIFF
--- a/support-files/mini-benchmark.sh
+++ b/support-files/mini-benchmark.sh
@@ -10,12 +10,30 @@ display_help() {
   echo "regressions."
   echo
   echo "optional arguments:"
+  echo "  --name STRING      identifier for the benchmark, added to the "
+  echo "                     folder name and (if --log is set) the log file "
+  echo "  --threads \"STRING\" quoted string of space-separated integers "
+  echo "                     representing the threads to run."
+  echo "                     example: --threads \"1 32 64 128\""
+  echo "                     default: \"1 2 4 8 16\""
+  echo "  --duration INTEGER duration of each thread run in seconds"
+  echo "                     default: 60"
+  echo "  --workload STRING  sysbench workload to execute"
+  echo "                     default: oltp_read_write"
+  echo "  --log              logs the mini-benchmark stdout/stderr into the"
+  echo "                     benchmark folder."
   echo "  --perf             measure CPU cycles and instruction count in for "
   echo "                     sysbench runs"
   echo "  --perf-flamegraph  record performance counters in perf.data.* and"
   echo "                     generate flamegraphs automatically"
   echo "  -h, --help         display this help and exit"
 }
+
+# Default parameters
+BENCHMARK_NAME='mini-benchmark'
+THREADS='1 2 4 8 16'
+DURATION=60
+WORKLOAD='oltp_read_write'
 
 while :
 do
@@ -27,6 +45,31 @@ do
     --version)
       display_version
       exit 0
+      ;;
+    --name)
+      shift
+      BENCHMARK_NAME+='-'
+      BENCHMARK_NAME+=$1
+      shift
+      ;;
+    --threads)
+      shift
+      THREADS=$1
+      shift
+      ;;
+    --duration)
+      shift
+      DURATION=$1
+      shift
+      ;;
+    --workload)
+      shift
+      WORKLOAD=$1
+      shift
+      ;;
+    --log)
+      LOG=true
+      shift
       ;;
     --perf)
       PERF=true
@@ -47,6 +90,13 @@ do
   esac
 done
 
+# Save results of this run in a subdirectory so that they are not overwritten by
+# the next run
+TIMESTAMP="$(date -Iseconds)"
+mkdir "$BENCHMARK_NAME-$TIMESTAMP"
+cd "$BENCHMARK_NAME-$TIMESTAMP" || exit 1
+
+(
 # Check that the dependencies of this script are available
 if [ ! -e /usr/bin/pgrep ]
 then
@@ -62,6 +112,7 @@ fi
 
 # If there are multiple processes, assume the last one is the actual server and
 # any potential other ones were just part of the service wrapper chain
+# shellcheck disable=SC2005
 MARIADB_SERVER_PID="$(echo "$(pgrep -f mariadbd || pgrep -f mysqld)" | tail -n 1)"
 
 if [ -z "$MARIADB_SERVER_PID" ]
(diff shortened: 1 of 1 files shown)


support-files/mini-benchmark.sh
```diff
@@ -156,28 +207,23 @@ mariadb -e "
   CREATE USER IF NOT EXISTS sbtest@localhost;
   GRANT ALL PRIVILEGES ON sbtest.* TO sbtest@localhost"
 
-sysbench oltp_read_write prepare --tables=20 --table-size=100000 | tee sysbench-prepare.log
+sysbench "$WORKLOAD" prepare --tables=20 --table-size=100000 | tee sysbench-prepare.log
```
grooverdan, line 210: any reason not to parametrize table and table size at the same time?

support-files/mini-benchmark.sh
```diff
@@ -240,3 +286,6 @@ case $RESULT in
     fi
     ;;
 esac
+# Record the output into the log file, if requested
+) 2>&1 | ($LOG && tee "$BENCHMARK_NAME"-"$TIMESTAMP".log)
```
grooverdan, line 290: easier way:
```
exec 2>&1 | ($LOG && tee "$BENCHMARK_NAME"-"$TIMESTAMP".log)
```
At the place where the `(` begins.

grooverdan (approved):
very nice cleanup. Thanks

Optional changes, let me know if you want to enact them.

But otherwise looks good to me.

support-files/mini-benchmark.sh
```diff
@@ -156,28 +207,23 @@ mariadb -e "
...
 # profile, and also guaranteed to be CPU bound because of being limited to a
 # single CPU with 'tasksel'.
-for t in 1 2 4 8 16
+for t in $THREADS
 do
   # Prepend command with perf if defined
-  # Output stderr to stdout as perf outpus everything in stderr
-  $PERF $TASKSET_SYSBENCH sysbench oltp_read_write run --threads=$t --time=60 --report-interval=10 2>&1 | tee sysbench-run-$t.log
+  # Output stderr to stdout as perf outputs everything in stderr
+  # shellcheck disable=SC2086
```
grooverdan, line 222: `$t` and `$DURATION` could still be quoted. 

Optional: To eliminate the shell check warning `THREADS`, `PERF` and `TASK_SETSYSBENCH` could all be changed to arrays.

support-files/mini-benchmark.sh
uoysip: If I understand correctly, the suggestion is to remove the [outermost `(...)` wrapper (at L99)](https://github.com/MariaDB/server/pull/2550/files/362529acac0ac9b030a47dd9c9d9f61179c923d7#diff-31ffc6ead737b0fde857304ec2501c45b01753115079125a5a9f55f0d7ec4c55R99) and instead use `exec` to redirect the output cleanly.

Using the exact `exec` line above, the output was not being logged to the file. The following worked instead, but readability is slightly worse:
```bash
exec > >($LOG && tee "$BENCHMARK_NAME"-"$TIMESTAMP".log) 2>&1
```

There are other ways of achieving the same result, though I prefer the original implementation (using the [outermost `(...)` wrapper](https://github.com/MariaDB/server/pull/2550/files/362529acac0ac9b030a47dd9c9d9f61179c923d7#diff-31ffc6ead737b0fde857304ec2501c45b01753115079125a5a9f55f0d7ec4c55R99)); it makes it clear what is being logged (everything within the `(...)` block) and the redirection/logging statement is readable.

support-files/mini-benchmark.sh
uoysip: It _might_ be worth leaving `$t` and `$DURATION` unquoted for readability/consistency with the rest of the line; if either of the values were globbed or word split then that would mean the user supplied bad inputs. Either way (quoting/unquoting) `sysbench` would complain that the values are in an unrecognized format.

If we agree that the values (`$t`, `$DURATION`) should be left unquoted, that would mean `shellcheck disable=SC2086` would remain, regardless of converting the other arguments into arrays ([`SC2086`](https://www.shellcheck.net/wiki/SC2086) complains about _all_ the arguments on the line being unquoted).

robinnewhouse:
Hi @grooverdan , wpould you be able to re-review this with the latest changes? Thank you.

ottok (approved):
These changes look good to me and this affects only the benchmark script which currently has low usage, so merging this is safe.

grooverdan:
Sorry for losing track of this one. Good now and thank you.

PR #2551, opened 2023-03-15
MDEV-30838 Assertion `m_thd == _current_thd()'

sciascid:
- Update wsrep-lib which contains fix for the assertion
- Fix error handling for appending fragment to streaming log, make sure tables are closed after rollback.


sciascid:
I amended the commit message in this pull request. It was referencing MDEV-30383 instead of MDEV-30838.

sysprg:
Thanks, merged with head revision as https://github.com/MariaDB/server/commit/ef227762b1c08b6259dfebb541ebb226e390203d

PR #2553, opened 2023-03-15
MDEV-30855 Remove test galera.galera_bf_abort_group_commit

sciascid:
This test was re-enabled in commit 0174a9ff3d412ada22205edc19066, and has been failing since then.
The test is configured such that Galera runs with commit ordering disabled, a configuration which is which was meant for testing the performance penalty of commit ordering (not meant to be used in practice).
Moreover, we have test galera_sr.galera_sr_bf_abort, which is identical, but runs with commit ordering enabled.
No reasons to keep the failing test around.


sysprg:
Thanks, merged with head revision after review: https://github.com/MariaDB/server/commit/1ac00c5e9f6168cfbb1934e23fc81f4891a678c0

PR #2554, opened 2023-03-16
MDEV-30473 Remove test galera.MDEV-27713

sciascid:
Remove test galera.MDEV-27713. This test relies on GET_LOCK() and has stopped working since commit 844ddb1 (see MDEV-30473). This commit disabled GET_LOCK() in combination with Galera.


sysprg:
Thanks, merged with head revision: https://github.com/MariaDB/server/commit/7d55eb00f357e876fc81631919dd7f7defaf5bec

PR #2557, opened 2023-03-17
[MDEV-30824] Fix binlog to use 'String' for setting 'character_set_client'

PhysicsTing:
- [x] *The Jira issue number for this PR is: MDEV-30824*

## Description
Commit a923d6f4 disabled numeric setting of character_set_* variables with non-default values:
```
MariaDB [(none)]> set character_set_client=224;
ERROR 1115 (42000): Unknown character set: '224'
```
However the corresponding binlog functionality still write numeric values for log event, and this will break binlog replay if the value is not default. Now make the server use 'String' type for 'character_set_client' when generating binlog event.

Before:
```
/*!\C utf8mb4 *//*!*/;
SET @@session.character_set_client=224,@@session.collation_connection=224,@@session.collation_server=33/*!*/;
```
After:
```
/*!\C utf8mb4 *//*!*/;
SET @@session.character_set_client=utf8mb4,@@session.collation_connection=33,@@session.collation_server=8/*!*/;
```

Note: prior to the previous commit, setting with '224' or '45' or 'utf8mb4' have the same effect, as they all set the parameter to 'utf8mb4'.

## How can this PR be tested?
MTR binglog related tests are rerecorded. The difference are as expected.
The rest of test cases in main suite all passed.

- [x] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced: 10.9. In addition this should also be merged into 10.10, 10.11 and 11.0 where the bug also presents*

## Backward compatibility
The setting of character_set_client originally supports variable type 'String', so this will not compromise any backward compatibility.

## PR quality check

## Copyright
Services, Inc.


PhysicsTing:
Some test failed due to some MTR tests not passing I'm addressing them.

an3l:
The changes from this PR are pushed with commit https://github.com/MariaDB/server/commit/dccbb5a6dba21b241e1796af82a0db85de28d195 to 10.5 @abarkov ok to close PR?

abarkov:
I changed the target version from 10.9 to 10.5, so a "mysqlbinlog" output from versions between 10.5 and 10.8 can also be loaded by a 10.9+ server.

ottok:
For the record, this was committed to 10.5 in https://github.com/MariaDB/server/commit/dccbb5a6dba21b241e1796af82a0db85de28d195

PR #2561, opened 2023-03-21
MDEV-30862 Assertion `mode_ == m_high_priority' failed

sciascid:
CREATE TABLE AS SELECT is not supported in combination with streaming replication.


sysprg:
Thanks, commit merged with head revision after review and testing: https://github.com/MariaDB/server/commit/bc3bfcf943b817b19a41e4f599b4f2e9a259b263

PR #2563, opened 2023-03-23
MDEV-29630: Remove unneeded code in join_read_always_key()

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-29630*

## Description
Remove unneeded code in join_read_always_key() and merge conditions in one.

## PR quality check

### DIFF
--- a/sql/sql_select.cc
+++ b/sql/sql_select.cc
@@ -23786,13 +23786,13 @@ join_const_unlock_row(JOIN_TAB *tab)
 
   DESCRIPTION
     This is "read_fist" function for the "ref" access method.
-   
+
     The functon must leave the index initialized when it returns.
     ref_or_null access implementation depends on that.
 
   RETURN
     0  - Ok
-   -1  - Row not found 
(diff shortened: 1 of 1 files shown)


sql/sql_select.cc
```diff
@@ -23803,14 +23803,12 @@ join_read_always_key(JOIN_TAB *tab)
   TABLE *table= tab->table;
 
   /* Initialize the index first */
-  if (!table->file->inited)
+  if ((!table->file->inited) && (unlikely((error=
```
an3l, line 23806: According to MDEV description `!table->file->inited` should be removed, right?

Weijun-H: Yes, we should remove it. And wait for testing pass.

Weijun-H: @Olernov  Remove `!table->file->inited` block would cause error, maybe we should flattern the block instead of remove it.

Weijun-H:
Turns out the code is actually is use

PR #2564, opened 2023-03-23
MDEV-30892 test galera.galera_log_bin is not deterministic

sjaakola:
galera.galera_log_bin test created the test tables and executed initial DML into node 2 Then connection is switched to node 1, where ALTER TABLE was attempted. But there is no guarantee that the table to alter was yet replicated to node 1.

The fix in this commit, creates the test tables in node 1 instead, so it is guaranteed that they are available for the later ALTER

- [x] *The Jira issue number for this PR is: MDEV-30892*

## Description
Fix for the undeterministic test galera.galera_log_bin

## How can this PR be tested?
PR has the test fixed

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/galera_log_bin.result
+++ b/mysql-test/suite/galera/r/galera_log_bin.result
@@ -1,10 +1,10 @@
 connection node_2;
 connection node_1;
-connection node_1;
+connection node_2;
 set global wsrep_on=OFF;
 reset master;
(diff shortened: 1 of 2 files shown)


sysprg:
Thanks, I've added a bit of a fix and merged with head revision as https://github.com/MariaDB/server/commit/9b6f87b62afb32ea238e39b362b8e761f53e541c

PR #2432, opened 2023-01-12
Use memory safe snprintf() in client

Chaloff:
## Description

Replace sprintf(buf, ...) with snprintf(buf, sizeof(buf), ...),
where buf is allocated with a size known at compile time.

This makes sure we are not writing outside array/string bounds which will
lead to undefined behaviour. In case the code is trying to write outside bounds - safe version of functions simply cut the string messages so we process this
gracefully.

Similar fixes were done in 19af1890 and 9de9f105.

## How can this PR be tested?

All build stages pass for these commits. For the commit without the intentional
bug, mysql-test-run is properly passing.  The changes are not change the
functionality itself they just replace unsafe C methods with the safe ones.

## Basing the PR against the correct MariaDB version
- [x] *This is a bug fix and the PR is based against the earliest branch in which the bug can be reproduced*

## Backward compatibility

The changes fully backward compatible.

All new code of the whole pull request, including one or several files
that are either new files or modified ones, are contributed under the BSD-new
license. I am contributing on behalf of my employer Amazon Web Services,
Inc.

PR #2429, opened 2023-01-10
Fix three occurrences of use-of-uninitialized-value

Chaloff:
GitLab CI backported to version 10.3 in this commit
[24e86b8](https://github.com/MariaDB/server/pull/2418/commits/24e86b862277b853842ebba9e0183312cb28b9c0)
was able to find several problems with similarsignature. Here we are fixing 3
first issues:

```
[ 34%] Built target aes-t
==7597==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55a56cdee49b in my_read_charset_file /CPACK_BUILD_SOURCE_DIRS_LONG_NAME_REQUIREMENT/mysys/charset.c:507:62
    #1 0x55a56cde5367 in init_available_charsets /CPACK_BUILD_SOURCE_DIRS_LONG_NAME_REQUIREMENT/mysys/charset.c:641:3
    #2 0x7fb7887e2a67 in __pthread_once_slow (/lib64/libc.so.6+0xa4a67)
    #3 0x55a56cde545f in get_charset_number /CPACK_BUILD_SOURCE_DIRS_LONG_NAME_REQUIREMENT/mysys/charset.c:705:3
    #4 0x55a56cdddba9 in create_sys_files /CPACK_BUILD_SOURCE_DIRS_LONG_NAME_REQUIREMENT/extra/comp_err.c:350:18
    #5 0x55a56cdddba9 in main /CPACK_BUILD_SOURCE_DIRS_LONG_NAME_REQUIREMENT/extra/comp_err.c:200:9
    #6 0x7fb78877deaf in __libc_start_call_main (/lib64/libc.so.6+0x3feaf)
    #7 0x7fb78877df5f in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3ff5f)
    #8 0x55a56cd555d4 in _start (/CPACK_BUILD_SOURCE_DIRS_LONG_NAME_REQUIREMENT/builddir/extra/comp_err+0x465d4)

  Uninitialized value was created by an allocation of 'stat_info' in the stack frame of function 'my_read_charset_file'
    #0 0x55a56cdedfb0 in my_read_charset_file /CPACK_BUILD_SOURCE_DIRS_LONG_NAME_REQUIREMENT/mysys/charset.c:500

[...]

PR #2573, opened 2023-03-26
Replace sprintf function to snprintf

arun-esh:
sprintf function is deprecated and recommended to use snprintf instead.

## Description
TODO: fill description here
sprintf function accepts a pointer to a buffer but does not check whether the buffer is large enough to contain the resulting string. 

Instead snprintf has one of its parameter as size of the buffer. 

## How can this PR be tested?
I have replaced the `sprintf` function wherever I got the error  while building MariaDB on macOS.

macOS: Ventura 13.2.1 \ Intel

- [*] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced*

## Backward compatibility

## PR quality check
- [*] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.

### DIFF
--- a/sql/gcalc_slicescan.cc
+++ b/sql/gcalc_slicescan.cc
@@ -102,7 +102,7 @@ const char *gcalc_ev_name(int ev)
 
 static int gcalc_pi_str(char *str, const Gcalc_heap::Info *pi, const char *postfix)
 {
-  return sprintf(str, "%s %d %d | %s %d %d%s",
+  return snprintf(str, sizeof(str), "%s %d %d | %s %d %d%s",
                      GCALC_SIGN(pi->node.shape.ix[0]) ? "-":"", FIRST_DIGIT(pi->node.shape.ix[0]),pi->node.shape.ix[1],
                      GCALC_SIGN(pi->node.shape.iy[0]) ? "-":"", FIRST_DIGIT(pi->node.shape.iy[0]),pi->node.shape.iy[1],
                      postfix);
@@ -148,12 +148,11 @@ static void GCALC_DBUG_PRINT_SLICE(const char *header,
   for (; slice; slice= slice->get_next())
   {
     size_t lnbuf= nbuf;
-    lnbuf+= sprintf(buf + lnbuf, "%d\t", slice->thread);
-    lnbuf+= sprintf(buf + lnbuf, "%s\t", gcalc_ev_name(slice->event));
+    lnbuf+= snprintf(buf + lnbuf, sizeof(buf) - lnbuf, "%d\t%s\t", slice->thread, gcalc_ev_name(slice->event));
 
     lnbuf+= gcalc_pi_str(buf + lnbuf, slice->pi, "\t");
     if (slice->is_bottom())
-      lnbuf+= sprintf(buf+lnbuf, "bt\t");
+      lnbuf+= snprintf(buf+lnbuf, sizeof(buf) - lnbuf, "bt\t");
     else
       lnbuf+= gcalc_pi_str(buf+lnbuf, slice->next_pi, "\t");
     buf[lnbuf]= 0;
--- a/sql/item_inetfunc.cc
+++ b/sql/item_inetfunc.cc
@@ -827,7 +827,7 @@ size_t Inet6::to_string(char *dst, size_t dstsize) const
       //
       // If it is not the last field, append closing ':'.
 
-      p += sprintf(p, "%x", ipv6_words[i]);
+      p += snprintf(p, sizeof(p), "%x", ipv6_words[i]);
 
       if (i + 1 != IN6_ADDR_NUM_WORDS)
       {
--- a/sql/item_inetfunc.cc
+++ b/sql/item_inetfunc.cc
@@ -827,7 +827,7 @@ size_t Inet6::to_string(char *dst, size_t dstsize) const
       //
       // If it is not the last field, append closing ':'.
 
-      p += snprintf(p, sizeof(p), "%x", ipv6_words[i]);
+      p += sprintf(p, "%x", ipv6_words[i]);
 
       if (i + 1 != IN6_ADDR_NUM_WORDS)
       {
--- a/storage/innobase/dict/dict0crea.cc
+++ b/storage/innobase/dict/dict0crea.cc
@@ -1750,7 +1750,7 @@ dict_foreign_def_get(
 				tbname, strlen(tbname), trx->mysql_thd);
 	tablebuf[bufend - tablebuf] = '\0';
 
-	snprintf(fk_def, sizeof(fk_def), 
+	sprintf(fk_def,
 		(char *)"CONSTRAINT %s FOREIGN KEY (", (char *)tablebuf);
 
 	for(i = 0; i < foreign->n_fields; i++) {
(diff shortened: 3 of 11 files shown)


sql/item_inetfunc.cc
```diff
@@ -827,7 +827,7 @@ size_t Inet6::to_string(char *dst, size_t dstsize) const
       //
       // If it is not the last field, append closing ':'.
 
-      p += sprintf(p, "%x", ipv6_words[i]);
+      p += snprintf(p, sizeof(p), "%x", ipv6_words[i]);
```
grooverdan, line 830: p isn't a character array.

storage/innobase/dict/dict0crea.cc
```diff
@@ -1750,7 +1750,7 @@ dict_foreign_def_get(
 				tbname, strlen(tbname), trx->mysql_thd);
 	tablebuf[bufend - tablebuf] = '\0';
 
-	sprintf(fk_def,
+	snprintf(fk_def, sizeof(fk_def),
```
grooverdan, line 1753: `fk_def` isn't a char array.

storage/innobase/dict/dict0dict.cc
```diff
@@ -1564,7 +1564,8 @@ dict_table_rename_in_cache(
 					strcat(foreign->id,
 					       old_id + ut_strlen(old_name));
 				} else {
-					sprintf(strchr(foreign->id, '/') + 1,
+					snprintf(strchr(foreign->id, '/') + 1,
+						sizeof(strchr(foreign->id, '/') + 1),
```
grooverdan, line 1568: dict_foreign_t.id is ptr, not a char array.

storage/innobase/include/dict0crea.inl
```diff
@@ -54,7 +54,7 @@ dict_create_add_foreign_id(
 		if (dict_table_t::is_temporary_name(name)) {
 
 			/* no overflow if number < 1e13 */
-			sprintf(id, "%s_ibfk_%lu", name,
+			snprintf(id, sizeof(id), "%s_ibfk_%lu", name,
```
grooverdan, line 57: namelen + 20 is the length

grooverdan (changes requested):
there's just too many pointers being used in sizeof which just doesn't evaluate correctly. `sizeof(ptr)` is 4 bytes on 64bit machines which isn't what's allocated.

sql/item_inetfunc.cc
arun-esh: @grooverdan `sprintf` also accepts character pointer by definition same as `snprintf. Just trying to understand why would array is required?

grooverdan:
```
#include <stdio.h>
#include <stdlib.h>

int main()
{ char *p, a[300];
  p= malloc(20);
  printf("sizeof(p) = %zu, sizeof(a) = %zu\n", sizeof(p), sizeof(a));
  return 0;
}
```

```
$ /tmp/main
sizeof(p) = 8, sizeof(a) = 300
```

See that sizeof isn't the allocated size except with where its known at compile time.

arun-esh: Here what I think...
So, the problem is pointer's size will be different than that of Array. Pointer's size will be calculated at the time of compiling based on OS, but Array could be of any value. But as `p` is just a pointer, `sizeof(p)` is not going to be correct. 
Is that correct understanding?

arun-esh (commented):
I have updated the files as per the recommendation. Can you please have a look.

storage/innobase/include/dict0crea.inl
arun-esh: Instead of `sizeof(id)` -> `sizeof(nameless + 20)`

storage/innobase/dict/dict0dict.cc
arun-esh: Reverted back to sprintf function

storage/innobase/dict/dict0crea.cc
arun-esh: Reverted back to sprintf function

sql/item_inetfunc.cc
arun-esh: Reverted back to sprintf function

ottok:
@arun-esh Do you want to review the similar PRs #2429, #2432, #2480, #2514 and #2590?

arun-esh:
> @arun-esh Do you want to review the similar PRs #2429, #2432, #2480, #2514 and #2590?

Sure! So not only `sprintf` but `strcpy` and others also needs to be replaced.

sql/gcalc_slicescan.cc
```diff
@@ -102,7 +102,7 @@ const char *gcalc_ev_name(int ev)
 
 static int gcalc_pi_str(char *str, const Gcalc_heap::Info *pi, const char *postfix)
 {
-  return sprintf(str, "%s %d %d | %s %d %d%s",
+  return snprintf(str, sizeof(str), "%s %d %d | %s %d %d%s",
```
Chaloff, line 105: size of str is not known on compiling time. sizeof(str) will return the size of pointer not the size of array. I suppose it's not that you intended to do

storage/innobase/include/dict0crea.inl
```diff
@@ -54,7 +54,7 @@ dict_create_add_foreign_id(
 		if (dict_table_t::is_temporary_name(name)) {
 
 			/* no overflow if number < 1e13 */
-			sprintf(id, "%s_ibfk_%lu", name,
+			snprintf(id, sizeof(namelen + 20), "%s_ibfk_%lu", name,
```
Chaloff, line 57: namelen is ulint type (=strlen(name). You probably want to write snprintf(id, namelen + 20, "%s_ibfk_%lu", name, ulong) (*id_nr)++); ?

storage/innobase/include/fts0priv.inl
```diff
@@ -50,14 +50,14 @@ fts_write_object_id(
 	/* Use this to construct old(5.6.14 and 5.7.3) windows
 	ambiguous aux table names */
 	DBUG_EXECUTE_IF("innodb_test_wrong_windows_fts_aux_table_name",
-			return(sprintf(str, "%016llu", (ulonglong) id)););
+			return(snprintf(str, sizeof(str), "%016llu", (ulonglong) id)););
```
Chaloff, line 53: The same thing - size of str is not known on compiling time, sizeof(str) will return the size of pointer not an array

storage/innobase/ut/ut0ut.cc
```diff
@@ -129,7 +129,7 @@ ut_sprintf_timestamp(
 	time(&tm);
 	localtime_r(&tm, &cal_tm);
 	cal_tm_ptr = &cal_tm;
-	sprintf(buf, "%02d%02d%02d %2d:%02d:%02d",
+	snprintf(buf, sizeof(buf), "%02d%02d%02d %2d:%02d:%02d",
```
Chaloff, line 132: the same - sizeof(buf) will return size of pointer

dlenski:
> @arun-esh Do you want to review the similar PRs https://github.com/MariaDB/server/pull/2429, https://github.com/MariaDB/server/pull/2432, https://github.com/MariaDB/server/pull/2480, https://github.com/MariaDB/server/pull/2514 and https://github.com/MariaDB/server/pull/2590?

It appears that _everything_ in this PR is already covered with greater care and accuracy in @Chaloff's memory-safety PRs. (Specifically, in https://github.com/MariaDB/server/pull/2432 and https://github.com/MariaDB/server/pull/2590 which address `sprintf` → `snprintf`.)

Perhaps this one should just be closed.

PR #2137, opened 2022-05-26
MDEV-7567 Add slow queries variable aliases

michaelqin3:
- [x] *The Jira issue number for this PR is: MDEV-7567*

## Description
Add aliases with prefix "log_slow" for system variables relating to slow query log.

Thus, all these variables will be grouped together and more logically named.

Descriptions for the old variables were updated to indicate they are
now aliases for the newly introduced variables with prefix log_slow.

log_slow_queries_not_using_indexes_filter will not be addressed
in this merge request.
log_throttle_queries_not_using_indexes seems to no longer be in use.

MTR tests are also updated to include the new variable names.

## How can this PR be tested?

All tests pass. MTR tests have been updated using --result.

## Basing the PR against the correct MariaDB version
- [x] *This is a new feature and the PR is based against the latest MariaDB development branch*
Target branch is 10.10

## Copyright
All new code of the whole pull request, including one or several files that are either new files or modified ones, are contributed under the BSD-new license. I am contributing on behalf of my employer Amazon Web Services, Inc.

PR #2114, opened 2022-05-10
MDEV-22023 Update man pages titles to say MariaDB instead of MySQL

PhysicsTing:
## Description
When reading the man page of e.g. 'mysql' on a system with MariaDB installed one would actually see the man page of 'mariadb'. However the man page had no indication of the page being for 'mariadb', which was confusing for users.

Fix this by updating the man page title lines to use mariadb-* instead of mysql* for MariaDB binaries that are drop-in replacements for MySQL equivalents, indicating that the commands are actually of the MariaDB version.

In long term, all the commands in man pages should be replaced by their MariaDB counterparts. Update the title lines as a start, and only those that exist as symlinks to their MariaDB counterparts.

Before:

    man mariadb-upgrade | head -n 1
    MYSQL_UPGRADE(1) ...

After:

    man mariadb-upgrade | head -n 1
    MARIADB-UPGRADE(1) ...

## How can this PR be tested?
Only man pages are changed so it is not amenable to test automation. However mtr main suite still runs and passes. Also man pages are read using man tool to verify that the outputs are as expected.

## Basing the PR against the correct MariaDB version
- [x] *This is a new feature and the PR is based against the latest MariaDB development branch*
Target 10.5 which was when that MariaDB names became default.

## Copyright
All new code of the whole pull request, including one or several files that are either new files or modified ones, are contributed under the BSD-new license. I am contributing on behalf of my employer Amazon Web Services, Inc.

PR #2579, opened 2023-03-28
Update pull request template to suggest making PRs editable by maintainers

cvicentiu:
Often there are small "nitpicky" changes that need to be done to a PR in order to get it merged. This tends to require a lot of back and forth that doesn't add any meaningful value to either the maintainer and the contributor. By encouraging contributors to make their PR editable by maintainers, we can shortcut this step and merge PRs faster.

## PR quality check


grooverdan (approved):

ericherman (approved):

ottok:
Two points of feedback:

## Point 1: This does not seem to help fixing the issue it tries to fix

> This tends to require a lot of back and forth that doesn't add any meaningful value to either the maintainer or the contributor. 

I disagree on this. In a code review at least 1-2 rounds of back and forth are expected, and they do add meaningful value.

This template change is a response to my feedback about https://github.com/MariaDB/server/pull/2557 where the reviewer gave zero feedback - in fact their first comment on the PR was 'merged on 10.5'. It would have been more in the open source spirit to ask the contributor to rebase it on 10.5, instead of pushing the commit directly to 10.5 outside the PR, and then closing the PR. Yes, this a round of **feedback does add a little bit of work, but it is important and meaningful for the overall process**.

Note that the PR had 'Allow edits by maintainers' enabled. Also, it it is enabled by default anyway. Asking submitters to enable it does not help prevent similar things.

You could tweak the text to ask permission to rewrite submissions in general without emphasize on the Github 'Allow edits by maintainer' checkbox, but I would still argue that to grow a good base of contributors, giving feedback and waiting for at least N days for the submitter to respond/refresh their submission is at the core of how open source works.

For example in https://github.com/MariaDB/server/pull/2137 the most distinguished MariaDB core developer gave feedback to a contributor, but also at the same time asked permission to finalise the submission himself. The response from the submitter (on the same day) was naturally to agree and give permission. While the intent surely was to make this one contribution get committed faster and well meaning, the end result unfortunately was a demovating as a first-time contribution experience.

To end on a positive example, in https://github.com/MariaDB/server/pull/2114 Daniel asked Ting on May 12th at 17:48 PDT to rebase on 10.5, and Ting did it on May 13th 9:14 PDT. All parties gave and responded to feedback, and did it promptly. The PR got merged and everyone were happy. This should continue and be upheld as the core value, not dismissed as "back-and-forth that does not add meaningful value".

## Point 2: The template is already very long and making it longer might put off first-time contributors

The template is already quite long, and opening a new PR can be a daunting experience for first-timers. Some might get discouraged, but there is no stats on Github on how started to draft a PR but gave up. There are numerous examples of submissions where people just ignore the template (e.g. 2570 by looking at the ones that came in from new contributors this week).

Adding more stuff to the template should not happen lightly, and asking people to allow edits (which is by default anyway in enabled) seems like something that the project probably can do without as well.

The template should strive to be short and easy to understand so that the overall contribution process is welcoming. I understand where this proposal came from and that it has a good intent, but please also consider that making things optimally convenient for the reviewer and communicates a mentality that it was the submitters fault that they didn't get any review nor  their PR merged.

In open source having the PR merged and getting the git credits is often the only reward the submitter gets, and having a lengthy template that states that one's submission might get rewritten by the reviewer can potentially make more harm than good to the overall goal of getting more contributors, and thus my vote goes to not making this addition to the template.


Please note that the feedback here is written as constructive and as factual as possible. I thank the MariaDB Foundation staff for taking time to try to improve the template in the first place even though I disagree on this particular change, and I thank for the continued efforts to improve the contribution process.

cvicentiu:
@ottok Thank you for taking the time to consider this change.

I think we both agree that communication is key and that reviewers have failed at properly communicating on various ocasions. I can definitely think of PRs I've merged that could have been handled better. There is definitely a big area to improve upon here. Moving forward, we will instruct reviewers to communicate intent and allow sufficient time for the contributor to reply / acknowledge. 

I can imagine scenarios where reviewers will prefer to do the edits themselves, but should that be the case, they should ask for permission and explain that the alternative will be much more time consuming (or perhaps not an option due to time constraints), before taking any action.

Nonetheless, while we can't fully control how non-foundation engineers work, we will do our best to act as moderators.

As for the PR template change itself:
* I disagree with the PR template being too long. I do not think we are really asking for too much, and the biggest chunk of text is "comments".
* I will also amend the wording, to encourage contributors to submit the PR, even if they do not believe it's perfect. We'll keep an eye on how many PRs follow the template. If many drop the template, we should consider changing it then.
* I didn't know the "allow edits by maintainers" is on by default. I'll pay more attention here and make use of the feature more in the future. I'll change the wording of the check box accordingly.

PR #2581, opened 2023-03-28
[MDEV-30178] Explicit errors on ssl transport auth

Chupsy:
- [x] *The Jira issue number for this PR is: MDEV-30178*

## Description
The error message for user connections using insecure transport when connection requires a secured transport.

To make the error message more relevant, introduce a new error'ER_ACCESS_SSL_REQUIRED'
as sibling of ACCESS_DENIED errors with the same
error code 28000, and update the corresponding error message.

```
Before: ER_ACCESS_DENIED "Access denied for user '%s'@'%s'"

Now: ER_ACCESS_SSL_REQUIRED "Access denied for user '%s'@'%s', connection requires secured transport."
```

This new error code does not check wether the error comes from specific user configuration that requires secured transport or if require_secure_transport is enabled. This is to prevent potential attackers from having too much information regarding the configuration of a specific user vs server configuration.

## How can this PR be tested?

Modified the existing test for require_secure_transport to expect the new error.

## Backward compatibility
This PR adds  a new error message, this should be fully backward compatible.

## PR quality check

## Copyright
All new code of the whole pull request, including one or several files that
are either new files or modified ones, are contributed under the BSD-new license.
I am contributing on behalf of my employer Amazon Web Services, Inc.


sql/sql_acl.cc
```diff
@@ -13183,14 +13183,19 @@ struct MPVIO_EXT :public MYSQL_PLUGIN_VIO
 /**
   a helper function to report an access denied error in most proper places
 */
-static void login_failed_error(THD *thd)
+static void login_failed_error(THD *thd, uint err_override = ER_ACCESS_DENIED_ERROR)
```
grooverdan, line 13186: I tend to dislike default values on arguments as its easy to get wrong. Can you change the other callers of `login_failed_error` to pass `ER_ACCESS_DENIED` rather than the default being here.

sql/sql_acl.cc
```diff
@@ -13183,14 +13183,19 @@ struct MPVIO_EXT :public MYSQL_PLUGIN_VIO
...
 {
-  my_error(access_denied_error_code(thd->password), MYF(0),
+  uint err = access_denied_error_code(thd->password);
+
+  if(err_override != ER_ACCESS_DENIED_ERROR)
+    err = err_override;
```
grooverdan, line 13191: Is this more readable?
`err= (err_override == ER_ACCESS_DENIED_ERROR) ? access_denied_error_code(thd->password) : err_override;`

alternately set `err= access_denied_error_code(` in `else` branch.

note style - no space on lhs of assignment.

sql/sql_acl.cc
```diff
@@ -14181,29 +14186,29 @@ static bool acl_check_ssl(THD *thd, const ACL_USER *acl_user)
...
-#else
-#ifndef _WIN32
-      return type != VIO_TYPE_SOCKET;
-#else
-      return type != VIO_TYPE_NAMEDPIPE;
-#endif
-#endif
+
+      bool has_secured_transport =
+      #ifdef HAVE_OPENSSL
```
grooverdan, line 14195: note the VIO_TYPE defines aren't limited by defines at all. Also Wolfssl is a SSL implementation. This could be a simple 3 condition `||` test.  You can use `#ifdef _WIN32` around the named pipe.

sql/sql_acl.cc
```diff
@@ -14540,12 +14545,13 @@ bool acl_authenticate(THD *thd, uint com_change_user_pkt_len)
       as an additional layer, not instead of the password
       (in which case it would've been a plugin too).
     */
-    if (acl_check_ssl(thd, acl_user))
+    uint acl_check_error = acl_check_ssl(thd, acl_user);
```
grooverdan, line 14548: new line just for readability.

grooverdan (changes requested):
nice to see this bug being addressed. Thank you.

note the other test failures in BB (perfschema.hostcache_ipv6_ssl perfschema.hostcache_ipv4_ssl), check if they are returning the right error code and adjust accordingly.

@vuvova as the error code is the same can this be a 10.5 fix? or did you want the MySQL error code 3159? Did you want to do a final review?

sql/sql_acl.cc
grooverdan: I looked closer, use of `#ifdef` seems common around these macros and `HAVE_OPENSSL` is defined for Wolfssl (that provides a compatible interface).

so fine as is.

grooverdan (approved):
thanks

Chupsy:
Thank you for your review! I see that buildbot is failing at 2 places, is that "expected"? When checking, it does not seem to be due to my changes

grooverdan:
> Thank you for your review! I see that buildbot is failing at 2 places, is that "expected"? When checking, it does not seem to be due to my changes

Yes. Both seem unrelated to your change and have occurred before.

vuvova (changes requested):
I'm not sure this is a good approach in general. I'd think a user-independent check for `opt_require_secure_transport` should be done rather early, almost immediately after the connection is established. And it'll fail with `ER_ACCESS_SSL_REQUIRED`. It's not part of the authentication, it checks whether the connection itself can be allowed to exist.

While `acl_check_ssl()` checks **user specific** SSL requirements, is done after the password is verified and returns ER_ACCESS_DENIED, as it's part of the authentication.

Also, as it's the error message that MySQL has, can it have the same text and the same error number as in MySQL? For compatibility reasons.

Chupsy:
Thank you for your review @vuvova

I understand that we should follow errors from MySQL, here is the original error
```
Mysql error:SQLSTATE[HY000] [3159] Connections using insecure transport are prohibited while --require_secure_transport=ON
```

This is a bit different from what is proposed in this CR, the error would only concern this small part
```
    if (opt_require_secure_transport)
    {
      enum enum_vio_type type= vio_type(vio);

      bool has_secured_transport =
      #ifdef HAVE_OPENSSL
        (type == VIO_TYPE_SSL) ||
      #endif
      #ifdef _WIN32
        (type == VIO_TYPE_NAMEDPIPE);
      #else
        (type == VIO_TYPE_SOCKET);
      #endif

      if(!has_secured_transport){
        return ER_ACCESS_SSL_REQUIRED;
      }
    }
```

We can move this part to check it earlier in the function, before the check of the password?

Chupsy:
For some reason, amd64-ubuntu-2204-msan is failing on tests unrelated to my changes... 
```
    innodb.undo_truncate
    mariabackup.incremental_page_compressed
    innodb.innodb_page_compressed
    innodb.innodb_page_compressed
    innodb.innodb_page_compressed
    innodb.innodb_page_compressed
    innodb.innodb_page_compressed
    innodb_zip.create_options innodb_zip.recover innodb_zip.innodb-zip innodb_zip.bug56680
```
I'm not sure why these come up

Chupsy:
@vuvova the CI is green, feel free to review at any time

Chupsy:
CI failure fedora-sanitizer: [-DWITH_UBSAN=YES]  is known as a result of building with performance schema. This is reported in https://jira.mariadb.org/browse/MDEV-22940. An upcoming PR will allow this particular UBSAN job to fail.

The same upcoming PR will also fix the fedora upgrade failed test job. This is caused by [mysql -e "SHUTDOWN;"](https://github.com/MariaDB/server/blob/9ea51a436dd0d207491abee54f5ce88b0ea17cfb/.gitlab-ci.yml#L465) being called after uninstall of the engine.

Chupsy:
The only CI failing is due to a test losing connection, I see that all PRs have the same issue, with a different test

dlenski:
@LinuxJedi, @vuvova, could you please take another look at this?

(And perhaps re-start the `amd64-ubuntu-2204-msan` build which appears to be failing in a way that's unrelated to this PR.)

sql/sql_connect.cc
```diff
@@ -826,6 +826,26 @@ bool init_new_connection_handler_thread()
...
 
+static bool check_require_secured_transport(THD *thd)
+{
+  Vio *vio= thd->net.vio;
+  if (opt_require_secure_transport)
+    {
+      enum enum_vio_type type= vio_type(vio);
+
+      return
+      #ifdef HAVE_OPENSSL
```
LinuxJedi, line 837: The `#ifdef` and alike should start at character zero, not padded.

LinuxJedi (commented):
In general I think this is a lot better. One minor thing to fix. Likely this will need rebasing against 11.2 once it is the main branch for the tree.

dlenski:
@LinuxJedi, there's a new failure in the MSAN CI build (`buildbot/amd64-ubuntu-2204-msan`) after @Chupsy's latest push…

```
Completed: Failed 1/5931 tests, 99.98% were successful.
Failing test(s): innodb_zip.recover
```

It appears to have nothing to do with the changes in this PR.

LinuxJedi:
Agreed, this is likely unrelated. I'll look into it.

sql/sql_connect.cc
```diff
@@ -826,6 +826,26 @@ bool init_new_connection_handler_thread()
   return 0;
 }
 
+static bool check_require_secured_transport(THD *thd)
+{
+  Vio *vio= thd->net.vio;
+  if (opt_require_secure_transport)
+    {
```
LinuxJedi, line 833: This brace indentation is off.

sql/sql_connect.cc
```diff
@@ -1078,6 +1098,16 @@ static int check_connection(THD *thd)
     return 1; /* The error is set by alloc(). */
   }
 
+  if(check_require_secured_transport(thd)){
```
LinuxJedi, line 1101: Opening brace on its own line.

LinuxJedi (changes requested):
A few more style issues to fixup. Looking good though :)

sql/sql_acl.cc
```diff
@@ -14184,25 +14184,7 @@ static bool acl_check_ssl(THD *thd, const ACL_USER *acl_user)
...
-#endif
-#else
-#ifndef _WIN32
-      return type != VIO_TYPE_SOCKET;
-#else
-      return type != VIO_TYPE_NAMEDPIPE;
-#endif
-#endif
-    }
+  case SSL_TYPE_NONE:                           // SSL is not required FOR THIS SPECIFIC USER
```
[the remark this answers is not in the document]
dlenski: What is the reason for the 80-character limit?

- It seems obsolete, in that no one is running an editor on a fixed-width terminal that's limited to around 80 characters.
- It's mostly not followed anyway. By my quick check (`git ls-files '*.c' '*.h' '*.cc' | xargs wc -L | sort -n`), there are currently about *1800 source files* with lines of >80 characters in the repository.

Chupsy:
I fixed the minor issues (and apparently the CI is working now)

dlenski (approved):

sql/share/errmsg-utf8.txt
```diff
@@ -10780,3 +10780,5 @@ ER_CM_OPTION_MISSING_REQUIREMENT
         eng "CHANGE MASTER TO option '%s=%s' is missing requirement %s"
 ER_SLAVE_STATEMENT_TIMEOUT 70100
         eng "Slave log event execution was interrupted (slave_max_statement_time exceeded)"
+ER_SECURE_TRANSPORT_REQUIRED 08004
```
vuvova, line 10783: let's try to have it the same number as in MySQL, if possible

sql/sql_connect.cc
```diff
@@ -826,6 +826,26 @@ bool init_new_connection_handler_thread()
...
+
+    return
+#ifdef HAVE_OPENSSL
+      (type != VIO_TYPE_SSL) &&
+#endif
+#ifdef _WIN32
+      (type != VIO_TYPE_NAMEDPIPE);
+#else
+      (type != VIO_TYPE_SOCKET);
+#endif
```
vuvova, line 844: do you need all these #ifdef's ? they make the code difficult to read. Values are defined unconditionally, so it seems that you can remove ifdefs and the code will still compile file.

sql/sql_connect.cc
```diff
@@ -1078,6 +1098,17 @@ static int check_connection(THD *thd)
     return 1; /* The error is set by alloc(). */
   }
 
+  if(check_require_secured_transport(thd))
+  {
+    Host_errors errors;
+    errors.m_ssl= 1;
+    inc_host_errors(thd->main_security_ctx.ip, &errors);
+    status_var_increment(thd->status_var.access_denied_errors);
```
vuvova, line 1106: may be `thd->status_var.denied_connections` ? not sure

sql/sql_connect.cc
dlenski: Those `#ifdef`s predate this PR, as you can see from [the diff of `sql_acl.cc` above](https://github.com/MariaDB/server/pull/2581/files#diff-df544694418bef1c4bc6cdc5211ca133e7ad4d31901f16d0fdee8df6e4debe89). @Chupsy has simplified them considerably here.

> Values are defined unconditionally, …

Right. https://github.com/MariaDB/server/blob/HEAD/include/violite.h#L38-L45

> … so it seems that you can remove ifdefs and the code will still compile file.

There are a fairly enormous number of other places in the code which are wrapped in unnecessary `#ifdef HAVE_OPENSSL`, in that case :sweat_smile:

sql/share/errmsg-utf8.txt
dlenski: - MySQL 5.7: https://github.com/mysql/mysql-server/blob/5.7/sql/share/errmsg-utf8.txt#L7574-L7575
- MySQL 8.0: https://github.com/mysql/mysql-server/blob/8.0/share/messages_to_clients.txt#L7650-L7651

sql/sql_connect.cc
Chupsy: I switched to denied_connections since it does make more sense

sql/sql_connect.cc
Chupsy: I am not totally certain on how #ifdef work so I just optimized them with @dlenski, not sure what would happen if we removed them. Tell me if you want me to remove them completely from the code

dlenski: Based on what @vuvova said and what I see in the code, it seems like you could simply change this section as follows…

```cpp
    return
      (type != VIO_TYPE_SSL) &&
      (type != VIO_TYPE_NAMEDPIPE) &&
      (type != VIO_TYPE_SOCKET);
```

… and it should still compile and work correctly on all platforms, even if MariaDB is (somehow) built without TLS support, and even though named pipes are only used on Windows builds, and Unix sockets are only used on POSIX builds.

sql/share/errmsg-utf8.txt
vuvova: How did you do it? I don't see how it was done :(

Chupsy: Isn't it defined only through `sql/share/errmsg-utf8.txt` ? If so, I modified the error number to match the new number

sql/sql_connect.cc
LinuxJedi: `denied_connections` is a global var I believe, not a session var, so it would be `status_var_increment(denied_connections)`

Chupsy: I put back the original part, should be good now

sql/sql_connect.cc
```diff
@@ -826,6 +826,21 @@ bool init_new_connection_handler_thread()
...
+{
+  Vio *vio= thd->net.vio;
+  if (opt_require_secure_transport)
+  {
+    enum enum_vio_type type= vio_type(vio);
+
+    return
+      (type != VIO_TYPE_SSL) &&
+      (type != VIO_TYPE_NAMEDPIPE);
+      (type != VIO_TYPE_SOCKET);
```
dlenski, line 839: @Chupsy, these should be separated by `;` rather than `&&`… probably should be caught by a compiler warning.

sql/share/errmsg-utf8.txt
```diff
@@ -10780,3 +10780,5 @@ ER_CM_OPTION_MISSING_REQUIREMENT
         eng "CHANGE MASTER TO option '%s=%s' is missing requirement %s"
 ER_SLAVE_STATEMENT_TIMEOUT 70100
         eng "Slave log event execution was interrupted (slave_max_statement_time exceeded)"
+ER_SECURE_TRANSPORT_REQUIRED 3159
```
vuvova, line 10783: eh, no. you didn't change the error number, you've made sqlstate to be 3159, which isn't even a valid sqlstate value.

you can see error numbers in the include/mysqld_error.h file, they're generated, error messages are simply numbered sequentially.

you can see that the last MySQL error in errmsg-utf8.txt is currently ER_ALTER_OPERATION_NOT_SUPPORTED_REASON_GIS, number 3060. So if you want to add a new one with the number 3159, you need to add ~100 new error messages. Like this
```diff
         eng "Do not support online operation on table with GIS index"
         spa "No soporta operación en línea en tabla con índice GIS"
+
+ER_MYSQL_3061
+        eng ""
+ER_MYSQL_3062
+        eng ""
... and so on ...
+ER_MYSQL_3157
+        eng ""
+ER_MYSQL_3158
+        eng ""
+
+ER_SECURE_TRANSPORT_REQUIRED 08004
+        eng "Connections using insecure transport are prohibited while --requi>
+
```

dlenski: > let's try to have it the same number as in MySQL, if possible

I completely understand the goal here, but how is this going to be possible in the long run?

If the wire protocol associates error codes with specific meanings purely based on their numeric codes, which are sequentially increasing numbers assigned by two diverging codebases (MySQL and MariaDB), then there's inevitably going to be a point where MySQL and MariaDB assign different meanings to the same code, right?

(Maybe it has happened already :man_shrugging:)

> So if you want to add a new one with the number 3159, you need to add ~100 new error messages.

Couldn't we simply skip to a new position using `start-error-number`, as done at several other points in the file? E.g.

```diff
 ER_ALTER_OPERATION_NOT_SUPPORTED_REASON_GIS
         chi "不要支持使用GIS索引的表中的在线操作"
         eng "Do not support online operation on table with GIS index"
         spa "No soporta operación en línea en tabla con índice GIS"
+
+ start-error-number 3159
+ ER_SECURE_TRANSPORT_REQUIRED 08004
+        eng "Connections using insecure transport are prohibited while --requi>
+
```

vuvova: 1. Ranges are different. 1000-1999 is the historical range for MySQL/MariaDB common errors, 2000-2999 is for client-side errors from Connector/C, 3000-3999 is the new MySQL error range, 4000-4999 is the new MariaDB error range.
2. no we cannot, I even tried that before writing a previous comment. it is assumed that ranges are all 1000 messages wide, see how `ER()` macro is defined:
```c
#define ERRORS_PER_RANGE 1000

#define ER_THD(thd,X) ((thd)->variables.errmsgs[((X)-ER_ERROR_FIRST) / ERRORS_PER_RANGE][(X) % ERRORS_PER_RANGE])
#define ER(X)         ER_THD(current_thd, (X))
```

dlenski (approved):

Chupsy:
I fixed the error number and the error that @dlenski found, should be good to go now :)

vuvova (approved):
Looks good to me, thanks @Chupsy !

Chupsy:
Thanks @vuvova !
@LinuxJedi anything else needed ? Can we merge the PR?

PR #2583, opened 2023-03-30
MDEV-29646: sformat('Num [{:20}]', 42) gives incorrect result in view

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-29646*

## Description
The problem is that sformat does not assign the enough space for the result string. The result string is allocated with the max_length of argument, but the correst max_length should be based on the format string.

The patch fixes the problem by calculating the max_length of the result string based on the format string.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/main/func_sformat.test
+++ b/mysql-test/main/func_sformat.test
@@ -253,3 +253,24 @@ set names latin1;
 echo #;
 echo # End of 10.7 tests;
 echo #;
+
+echo #;
+echo # Start of 10.8 tests;
+echo #;
+
+echo #;
+echo # MDEV-29646 sformat('Num [{:20}]', 42) gives incorrect result in view;
+echo #;
+
+create view v1 as select sformat('Num [{:20}]', 42);
+select * from v1;
+drop view v1;
+
+create view v1 as SELECT sformat('Square root of [{:d}] is [{:.20}]', 2, sqrt(2));
+select * from v1;
+drop view v1;
+
+create table t1 (a text, b int, c text);
+insert t1 values ('[{} -> {}]', 10, '{}'), ('[{:20} <- {}]', 1, '{:30}');
+select sformat(a,b,c) from t1;
+drop table t1;
--- a/sql/item_strfunc.cc
+++ b/sql/item_strfunc.cc
@@ -1348,13 +1348,13 @@ bool Item_func_sformat::fix_length_and_dec()
 
   for (uint i=0 ; i < arg_count ; i++)
   {
-    char_length+= args[i]->max_char_length();
     if (args[i]->result_type() == STRING_RESULT &&
         Type_std_attributes::agg_item_set_converter(c, func_name_cstring(),
                                                     args+i, 1, flags, 1))
       return TRUE;
   }
 
+  char_length= MAX_BLOB_WIDTH;
   fix_char_length_ulonglong(char_length);
   return FALSE;
 }
--- a/mysql-test/main/func_sformat.result
+++ b/mysql-test/main/func_sformat.result
@@ -434,7 +434,7 @@ create table t1 as select sformat(_ucs2 x'003D007B007D003D', _ucs2 x'04420435044
 show create table t1;
 Table	Create Table
 t1	CREATE TABLE `t1` (
-  `x` varchar(8) CHARACTER SET ucs2 COLLATE ucs2_general_ci DEFAULT NULL
+  `x` longtext CHARACTER SET ucs2 COLLATE ucs2_general_ci DEFAULT NULL
 ) ENGINE=MyISAM DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
 drop table t1;
 set names latin1;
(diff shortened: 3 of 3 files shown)


mysql-test/main/func_sformat.test
```diff
@@ -250,6 +250,14 @@ set @a=3.14;
...
 set names latin1;
 
+echo #;
+echo # MDEV-29646 sformat('Num [{:20}]', 42) gives incorrect result in view;
+echo #;
+
+create view v1 as select sformat('Num [{:20}]', 42);
+select * from v1;
+drop view v1;
+
```
grooverdan, line 260: good basic test, but probably more test cases are needed. Also this is in 10.8 so its after the "End of 10.7". Adding a "Start of 10.8" will help merging eventually.

sql/item_strfunc.cc
```diff
@@ -1348,13 +1348,31 @@ bool Item_func_sformat::fix_length_and_dec()
 
   for (uint i=0 ; i < arg_count ; i++)
   {
-    char_length+= args[i]->max_char_length();
+    // scan the number in {:d} and add it to the char_length
+    size_t num= 0;
+    const char* str= args[i]->name.str;
+    char delimiter1[]= "{:";
+    char delimiter2[]= "}";
```
grooverdan, line 1355: A couple of problems without a full solution.
`{{` is an escaped `{` so should be processed ref: https://fmt.dev/latest/syntax.html

The width isn't always the next argument - after `:` - https://fmt.dev/latest/syntax.html#format-specification-mini-language

sql/item_strfunc.cc
```diff
@@ -1348,13 +1348,31 @@ bool Item_func_sformat::fix_length_and_dec()
...
+
+    const char* pos1 = str;
+    const char* pos2 = NULL;
+
+    char num_str[100]; // allocate a buffer for the number string
+    while ((pos1 = strstr(pos1, delimiter1)) != NULL &&
+           (pos2 = strstr(pos1 + strlen(delimiter1), delimiter2)) != NULL) {
+        strncpy(num_str, pos1 + strlen(delimiter1), pos2 - pos1 - strlen(delimiter1));
+        num_str[pos2 - pos1 - strlen(delimiter1)] = '\0';
+        num+= strtol(num_str, NULL, 10);
```
grooverdan, line 1365: `strtol` with a not-`NULL` end bufferptr would avoid the need to use a `num_str` buffer.

sql/item_strfunc.cc
grooverdan: I'd be looking for a solution in the fmt API. @vuvova any other ideas of a solution to this one?

Weijun-H: Yes, using fmt api is a better choice. I also think about regrex to extract the number. What do you think?

Weijun-H: Also I found this api could return the length of result https://fmt.dev/latest/api.html#_CPPv4IDpEN3fmt14formatted_sizeE6size_t13format_stringIDp1TEDpRR1T

mysql-test/main/func_sformat.test
```diff
@@ -253,3 +253,21 @@ set names latin1;
...
+echo #;
+echo # MDEV-29646 sformat('Num [{:20}]', 42) gives incorrect result in view;
+echo #;
+
+create view v1 as select sformat('Num [{:20}]', 42);
+select * from v1;
+drop view v1;
+
+SELECT sformat('Square root of [{:d}] is [{:.20}]', 2, sqrt(2));
```
vuvova, line 269: this should be in a view too, right? or in `create ... select` or may be `enable_metadata` will show the length. but without any of that I don't see what you're testing

vuvova: add tests when some arguments aren't constants and when all arguments aren't constants. Like
```sql
create table t1 (a text, b int, c text);
insert t1 values ('{} -> {}', 10, '{}'), '{:20} <- {}', 1, '{:30}');
select sformat(a,b,c) from t1;
```

sql/item_strfunc.cc
vuvova: This is not easy. @grooverdan already commented about `{{`. You would basically need to have a complete frm format parser to get the length right. And we don't want to have a second fmt format parser.

`fmt::formatted_size()` needs the string and all arguments. To call it you need to use dummy values for all arguments, like, `0` for `INT_RESULT`, `""` for `STRING_RESULT`, etc. This is quite a trick, so make sure to avoid it when it's not needed, like, when one of the arguments is already blob, so the length is already at the maximum.

perhaps it'd be enough to always return `MAX_BLOB_WIDTH` here without trying to estimate the length more precisely.

mysql-test/main/func_sformat.test
Weijun-H: nice test

grooverdan:
Sorry for the delay. Can you squash these commits up and rebase onto the latest 10.8.

Weijun-H:
> Sorry for the delay. Can you squash these commits up and rebase onto the latest 10.8.

Ok, I will do this latter today.

grooverdan (approved):

grooverdan:
Thanks @Weijun-H .

PR #2587, opened 2023-03-31
MDEV-30955 Explicit locks released too early in rollback path

sciascid:
Assertion `thd->mdl_context.is_lock_owner()` fires when a client is disconnected, while transaction and and a table is opened through `HANDLER` interface.
Reason for the assertion is that when a connection closes, its ongoing transaction is eventually rolled back in
`Wsrep_client_state::bf_rollback()`. This method also releases explicit which are expected to survive beyond the transaction lifetime. This patch also removes calls to `mysql_ull_cleanup()`. User level locks are not supported in combination with Galera, making these calls unnecessary.

### DIFF
--- a/mysql-test/suite/galera/r/MDEV-30955.result
+++ b/mysql-test/suite/galera/r/MDEV-30955.result
@@ -1,0 +1,26 @@
+connection node_2;
+connection node_1;
+CREATE TABLE t (a CHAR(1) KEY);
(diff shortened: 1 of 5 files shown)


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/feeeacc4d747868c234425dc12c157c6e5fa8fbb

PR #2588, opened 2023-04-04
 MDEV-30414 sporadic failures with galera var retry autocommit

sjaakola:
- [x] *The Jira issue number for this PR is: MDEV-30414*

## Description
the test phase 2, was not deterministic. In problematic scheduling case, the autocommit retry would be too fast and enter once again to block the applier for second time.
The fix, in this PR, is to stop the retrying in sync point, before the actual retrying begins. And let the applier continue first. 
## How can this PR be tested?
PR is about a mtr test fix

- [x ] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced*

## PR quality check


sysprg:
Thanks, merged with head revision as https://github.com/MariaDB/server/commit/4e942bcd93e961fe9e5ee5502f7817633600db4c

PR #2589, opened 2023-04-04
MDEV-30879 Add conversion to based 62 for CONV function

Weijun-H:
- [x] *The Jira issue number for this PR is: MDEV-30879*

## Description
This patch adds conversion to based 62 for CONV function.

This patch splits the CONV function into two situation: first, convert the number from 2 to 36, second, convert the number from 36 to 62. Use `convert_str_to_ull` and `convert_str_to_ll` to handle the case lager than 36.
## How can this PR be tested?
Append the tests to `func_str`

## PR quality check

### DIFF
--- a/sql/item_strfunc.cc
+++ b/sql/item_strfunc.cc
@@ -3712,6 +3712,202 @@ err:
   return 0;
 }
 
+/*
+  This is a helper function for converting a string based on the number between 32 and 62
+  to unsigned long long in Item_func_conv::val_str
+*/
+inline longlong convert_str_to_ull(CHARSET_INFO *cs,
+                             const char *nptr, size_t l, int base,
+                             char **endptr, int *err, bool* null_value)
+{
+  int negative;
+  ulonglong cutoff;
+  uint cutlim;
+  ulonglong i;
+  const char *s, *e;
+  const char *save;
+  int overflow;
+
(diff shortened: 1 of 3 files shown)


sql/item_strfunc.cc
```diff
@@ -3712,6 +3712,202 @@ String *Item_func_lpad::val_str(String *str)
...
+  ulonglong cutoff;
+  uint cutlim;
+  ulonglong i;
+  const char *s, *e;
+  const char *save;
+  int overflow;
+
+  *err= 0;				/* Initialize error indicator */
+
+  s = nptr;
```
LinuxJedi, line 3733: No space before the equals here and a few other places in the code.

LinuxJedi (changes requested):
Many thanks for your contribution. A few minor changes needed I think. Please also add some tests for the extremes such as overflows and maximum allowed / minimum allowed for negatives.

LinuxJedi:
Judging by other implementations and this wiki page (https://en.wikipedia.org/wiki/Base62), the case processing is inverted in this patch. I'm working on a fix for it.

LinuxJedi:
I've re-written this in #2589 so that it is in the charset code instead of duplicates of the charset code. Many thanks for your contribution, this guided the final patch and some of your tests were used (as marked by the Co-authored-by tag).

PR #2593, opened 2023-04-11
rocksdb: Define _GNU_SOURCE during fallocate CMake probe

fweimer-rh:
The glibc headers declare fallocate only if _GNU_SOURCE is defined. Without this change, the probe fails with C compilers which do not support implicit function declarations even if the system does in fact support the fallocate function.

Upstream rocksdb does not need this because the probe is run with the C++ compiler, and current g++ versions define _GNU_SOURCE automatically.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


grooverdan (approved):
Thanks Florian!

The linux/falloca.h header seemed unnecessary too.

PR #2594, opened 2023-04-12
MDEV-31042: Fixing table corruption after incremental backup restore in TSE enabled databases

thejaka-kanewala:
- [x] *The Jira issue number for this PR is: MDEV-31042*

## Description
The Problem
==========
Consider the following steps:

Suppose there is an encrypted table "t"
Take a base backup of the database
Create a table like t and name it tmp_t
Rename t to z_t and tmp_t to t
Start an increment backup, while the increment backup is ongoing, execute
optimize table on t (make sure this is executed after copying t to the backup)
Restore the incremental backup and start the server.
When we try to rename z_t table to t_prime, we get the following error:
errno: 155 "The table does not exist in the storage engine"

The Cause
========
An incremental backup collect pages modified since a given LSN number.
The modified pages are written to the backup as .delta files.
Further, tables created/optimized while the backup is on-going on are
saved with a suffix such as ".new", ".ren".

The step 5 above creates a t.new file. These files are processed
at the begining of the apply log stage when preparing the incremental
backup. When t.new file is processed, mariabackup deletes the
existing t.ibd file from the base directory and copies t.new as t.ibd.
However, the new t.ibd file carries a tablespace id different from
the old t.ibd file.

To apply the delta, mariabackup finds the proper tablespace file
in "xb_delta_open_matching_space". Mariabackup fails to
find a matching tablespace for z_t (created at step 4 above) inside
"xb_delta_open_matching_space", because there isn't a matching tablespace
in the base backup by the name z_t and also there isn't a matching space
with z_t's tablespace id. At this point, mariabackup creates new tablespace
file. However, this newly created tablespace file does not contain
crypt data and it just have 4 allocated pages. If delta file for
z_t table is empty, the restore operation succeeds but when
we try to access z_t after restore it remains corrupted. If delta file
contains pages (i.e., z_t is not empty), then we will see mariabackup
throwing an error saying that delta file cannot be applied because it cannot
read page 4 from the ibd file.

The Fix
=====
When processing .new files, deleting original t.ibd file and replacing
it with t.new file is the main cause for the issue.

Therefore, in the fix, when processing t.new files we do not
remove original t.ibd file, instead we rename the existing t.ibd
file to xtrabackup_tmp#<space_id>.ibd.

With this, when mariabackup try to find a matching space for z_t.delta.ibd, it will
return the xtrabackup_tmp#<space_id>.ibd and will be renamed to z_t.ibd in the
base backup. Then, afterwards, mariabackup should be able to apply the delta
to the z_t.ibd.

## How can this PR be tested?
I have included 3 MTRs to test this issue.
MDEV-31042_1 -- Verifies the expected behaviour when the table is empty.
MDEV-31042_2 -- Verifies the expected behaviour when the table is not empty.
MDEV-31042_3 -- To make sure we dont leave any temp tables (xtrabackup_tmp<space_id>.ibd) after restore.

You might consider answering some questions like:
1. Does this affect the on-disk format used by MariaDB? - No.
2. Does this change any behavior experienced by a user
   who upgrades from a version prior to this patch? - Yes.
3. Would a user be able to start MariaDB on a datadir
   created prior to your fix? - Yes.

## PR quality check

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -5408,6 +5408,69 @@ static void rename_force(const char *from, const char *to) {
 }
 
 
+/**
+ * Given an ibd file path, this returns the space id.
+ * @param file The ibd file
+ * @return space id for the ibd file, -1 if an error.
+ */
+static
+ulint get_space_id(const char *file) {
+	// make sure this is an ibd file ...
+	if (!ends_with(file, ".ibd"))
+		die("Error: Requesting space id for a non-ibd file: %s", file);;
+
+	int fd = open(file, O_RDONLY, 0);
+	if (fd < 0)
+		die("Error: Unable to open file for reading: %s", file);
+
+	// read first page to the buffer ...
+	// we are reading the base backup ibd file, therefore,
+	// srv_page_size should work (no need to read page size
+	// from delta meta file)
+	unsigned char *page = new unsigned char[srv_page_size];
+	memset(page, 0, srv_page_size);
+
+	ulong bytes_read;
+	if ((bytes_read = read(fd, page, srv_page_size)) == srv_page_size) {
+		// read the space id ...
+		ulint space_id = mach_read_from_4(
+				page + FIL_PAGE_ARCH_LOG_NO_OR_SPACE_ID);
+
+		delete[] page;
+		close(fd);
+
+		return space_id;
+	} else {
+		delete[] page;
+		close(fd);
+
+		die("Error: Unable to open file for reading: %s", file);
+	}
+}
+
+
+/** Very similar to rename_force except if file exists,
+ * then, rename it to xtrabackup#<space_id> and populate it to
+ * .new existing set.*/
+static
+void safe_rename(const char *from, const char *to, const char *dest_db_dir) {
+	if (access(to, R_OK) == 0) {
+		ulint to_space_id = get_space_id(to);
+
+		// do the rename ...
+		char new_to_name[FN_REFLEN];
+		snprintf(new_to_name, FN_REFLEN, "%s/xtrabackup_tmp_#" ULINTPF ".ibd",
+				dest_db_dir, to_space_id);
+
+		msg("mariabackup: .new file found for %s. Renaming existing %s to %s.ibd",
+			to, to, new_to_name);
+		rename_file(to, new_to_name);
+	}
+
+	rename_file(from, to);
+}
+
+
 /** During prepare phase, rename ".new" files, that were created in
 backup_fix_ddl() and backup_optimized_ddl_op(), to ".ibd". In the case of
 incremental backup, i.e. of arg argument is set, move ".new" files to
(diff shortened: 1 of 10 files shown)


extra/mariabackup/xtrabackup.cc
```diff
@@ -5408,6 +5408,69 @@ static void rename_force(const char *from, const char *to) {
 }
 
 
+/**
+ * Given an ibd file path, this returns the space id.
+ * @param file The ibd file
+ * @return space id for the ibd file, -1 if an error.
+ */
+static
+ulint get_space_id(const char *file) {
```
dr-m, line 5417: InnoDB tablespace identifiers are 32-bit. The return type should be `uint32_t` accordingly.

`-1` is never returned by the function.

extra/mariabackup/xtrabackup.cc
```diff
@@ -5408,6 +5408,69 @@ static void rename_force(const char *from, const char *to) {
...
+	// srv_page_size should work (no need to read page size
+	// from delta meta file)
+	unsigned char *page = new unsigned char[srv_page_size];
+	memset(page, 0, srv_page_size);
+
+	ulong bytes_read;
+	if ((bytes_read = read(fd, page, srv_page_size)) == srv_page_size) {
+		// read the space id ...
+		ulint space_id = mach_read_from_4(
+				page + FIL_PAGE_ARCH_LOG_NO_OR_SPACE_ID);
```
dr-m, line 5437: The page checksum is not being validated here.

Starting with 86dc7b4d4cfe15a2d37f8b5f60c4fce5dba9491d it is more likely that data files will not contain valid tablespace identifiers.

I think that any fix of this needs to be carefully tested in such a way that incremental backups are being made during a DDL intensive workload.

mysql-test/suite/mariabackup/MDEV-31042_1.opt
```diff
@@ -0,0 +1,7 @@
+--innodb-encryption-rotate-key-age=2
+--innodb_encrypt_tables=ON
+--innodb-encryption-threads=4
+--innodb-tablespaces-encryption
+--plugin-load-add=$FILE_KEY_MANAGEMENT_SO
+--loose-file-key-management
+--loose-file-key-management-filename=$MYSQL_TEST_DIR/std_data/logkey.txt
```
dr-m, line 7: There is a missing newline character at the end of the file.

I would like test names to be descriptive, such as `incremental_encrypted`. Instead of creating new test files, could that existing test be extended?

mysql-test/suite/mariabackup/MDEV-31042_3.test
```diff
@@ -0,0 +1,48 @@
+# This is to make sure that the xtrabackup_tmp_#<tablespace_id> is removed when not utilized.
+
+--source include/have_file_key_management.inc
+--source include/have_debug.inc
```
dr-m, line 4: Tests should not be skipped on non-debug builds unless there is a very good reason. I do not see any fault injection being present in this test.

dr-m (changes requested):
Because 10.2 reached EOL in 0ba528fe56f6c637d9fbc9d177a62610038fd519 and 10.3 will reach EOL soon too, I think that this needs to target 10.4.

This needs to be tested with a DDL-heavy workload also on 10.5 and 10.6.

extra/mariabackup/xtrabackup.cc
midenok: @dr-m Any ideas how to fix this?
@thejaka-kanewala Are you sure there can be no clash on `xtrabackup#<space_id>` after multiple incremental backups?

midenok:
Deprecated since 10.5, not fitting into 10.4 LTS due to unaddressed comments.

PR #2424, opened 2023-01-09
Deb: Sync Salsa-CI from Debian MariaDB 10.6

ottok:
- [x] *The Jira issue number for this PR is: MDEV-MDEV-30322*

## Description

Compare to Debian packaging of MariaDB 1:10.6.11-2 release at commit https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/2934e8a795cbd1eba79e33b8d6c82ed2b0897128 and sync upstream everything that is relevant for upstream and safe to import on a stable 10.6 release.

* Use OpenSSL 1.1 from Debian Snapshots https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/0040c272bf8a9a253fcd393b235b18cd5cdc9289 Related: https://jira.mariadb.org/browse/MDEV-30322

* Prefer using bullseye-backports in mosts tests over buster-bpo https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/daa827ecded9dc1f096ecb05242116173298d602

* Add new upgrade test for MySQL Community Cluster 8.0 https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/3c71bec9b7626d25a55e23b2e570b4e125d5d5df

* Enable automatic datadir move also on upgrades from MySQL.com packages https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/4cbbcb7e56d9604a67230d11a68543d5a83960e3

* Update Breaks/Replaces https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/2cab13d05959fe8f6ea56b866df51f3bfe17dd3f

* Normalize apt-get and curl commands https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/8754ea2578bd214d75a2a9e6a0f7b4b74c062b41

* Standardize on using capitalized 'ON' in CMake build options https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/938757a85aee44c727f3f6996677ef1126c2f5aa

* Apply wrap-and-sort -av and other minor tweaks and inline documentation

NOTE TO MERGERS: This commit is made on 10.6 branch and can be merged to all later branches (10.7, 10.8, ..., 10.11).

### Salsa-CI

Before: https://salsa.debian.org/mariadb-team/mariadb-server/-/pipelines/481117

image

After: https://salsa.debian.org/otto/mariadb-server/-/pipelines/481866

image

[...]

PR #2597, opened 2023-04-13
MDEV-31045: Fix regression building on Ubuntu 18.04

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-31045*

## Description
Github PR #2424 regressed Ubuntu 18.04 building other than x86_64 machines. Architecture that are impacted are **PPC64** and **ARM64**. This was because of changes in debian/rules file

which caused removing dependency to package 'libpmem-dev' and CMake which '-DWITH_PMEM' removing not working correctly. Package libpmem-dev was removed but it still required to have PMEM with CMake which. Commit make change that -DWITH_PMEM is correctly removed if it's not wanted.

## How can this PR be tested?
This is easiest to test with MariaDB testsuite on correct architecture PPC64 and ARM64

## Backward compatibility
This should bring back backward compatibility

## PR quality check

### DIFF
--- a/debian/autobake-deb.sh
+++ b/debian/autobake-deb.sh
@@ -64,7 +64,7 @@ replace_uring_with_aio()
 disable_pmem()
 {
   sed '/libpmem-dev/d' -i debian/control
-  sed '/-DWITH_PMEM=YES/d' -i debian/rules
+  sed '/-DWITH_PMEM=ON/d' -i debian/rules
 }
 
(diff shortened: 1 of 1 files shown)


grooverdan (approved):

illuusio:
I also made by hand PPC64 and ARM64 testing. I'll test also #2598 as it should provide this one on 10.9

ottok:
I see, thanks for the clarification!

PR #2598, opened 2023-04-13
MDEV-31045: Update debian/rules to be in sync to MariaDB LTS 10.6

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-31045*

## Description
To make sure that PMEM problem does not happen again sync 10.9 and up debian/rules to make sure that they are not making anymore problems with autobake-debs.sh

MariaDB 10.8 have correct changes but 10.9 and above does not have needed changes to make PR #2597 to work

## How can this PR be tested?
This can tested to run on Debian machine `debian/autobake-debs.sh` and see if it works.

## Backward compatibility
This ensures more backward compatibility

## PR quality check


debian/autobake-deb.sh
```diff
@@ -58,8 +58,8 @@ remove_rocksdb_tools()
 replace_uring_with_aio()
 {
   sed 's/liburing-dev/libaio-dev/g' -i debian/control
-  sed -e '/-DIGNORE_AIO_CHECK=YES/d' \
-      -e '/-DWITH_URING=YES/d' -i debian/rules
+  sed -e '/-DIGNORE_AIO_CHECK=ON/d' \
+      -e '/-DWITH_URING=ON/d' -i debian/rules
 }
 
 disable_pmem()
```
grooverdan, line 65: so below needs to be =ON too, or alternately we're waiting for a #2597 merge up.

illuusio: Yes this was the thing that I was thinking back and forward and thinked that waiting #2597 but after thinking it while how long it takes and do we get to same point that PR fixed if we don't do it here.

PR #2600, opened 2023-04-13
MDEV-30430: Enabling system versioning on tables without primary key breaks replication

bnestere:
When replicating MDL events for a table that uses system versioning
without primary keys, ensure that for data sets with duplicate
records, the updates to these records with duplicates are enacted on
the correct row. That is, there was a bug (reported in MDEV-30430)
such that the function to find the row to update would stop after
finding the first matching record. However, in the absence of
primary keys, the version of the record is needed to compare the row
to ensure we are updating the correct one.

The fix, therefore, updates the record comparison functionality to
use system version columns when there are no primary keys on the
table.

The first commit on the branch is the regression test, and the second
commit on the branch is the code fix.

### DIFF
--- a/mysql-test/suite/versioning/r/rpl.result
+++ b/mysql-test/suite/versioning/r/rpl.result
@@ -172,9 +172,11 @@ insert into t1 values (1);
 delete from t1;
 connection slave;
 include/diff_tables.inc [master:test.t1,slave:test.t1]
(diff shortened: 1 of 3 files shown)


knielsen:
I can see that this was pushed some time ago.
So closing the pull request, hope that's ok.

PR #2601, opened 2023-04-13
MDEV-35988 - Fix insecure use of strcpy, strcat and sprintf in Client and SQL (continues)

Chaloff:
…tinues)

Old style C functions `strcpy()`, `strcat()` and `sprintf()` are vulnerable to security issues due to lacking memory boundary checks. Replace these in the Client and SQL with safe new and/or custom functions such as `snprintf()` `safe_strcpy()` and `safe_strcat()`.
Continue, started 75e394f6b14645b703fa8d41632a3858999b03bd.

## Description

Old style C functions `strcpy()`, `strcat()` and `sprintf()` are vulnerable to
security issues due to lacking memory boundary checks. Replace these in the
Client and SQL with safe new and/or custom functions such as
`snprintf()` `safe_strcpy()` and `safe_strcat()`.
Continue, started 75e394f6b14645b703fa8d41632a3858999b03bd.

With this change FlawFinder and other static security analyzers report 287 fewer
findings.

## How can this PR be tested?

The FlawFinder log before changes:

```
$ cat flawfinder-all-vulnerabilities.html | grep "Hits ="
Hits = 14955
```
After the changes:
```
$ cat flawfinder-all-vulnerabilities.html | grep "Hits ="
Hits = 14919
```
The number of fixes - `36`

Additionally all CI passes for this change.

## Backward compatibility


sql/sql_connect.cc
```diff
@@ -985,7 +985,7 @@ static int check_connection(THD *thd)
                       /* See RFC 5737, 192.0.2.0/24 is reserved. */
                       const char* fake= "192.0.2.4";
                       inet_pton(AF_INET,fake, ip4);
-                      strcpy(ip, fake);
+                      snprintf(ip, sizeof(ip), "%s", fake);
```
grooverdan, line 988: is a const char copy, compilers can detect length, esp with `ip` bloated to `NI_MAXHOST` (man 3 getnameinfo). Likewise below.

sql/mysqld.cc
```diff
@@ -2430,7 +2430,7 @@ static MYSQL_SOCKET activate_tcp_port(uint port)
   if (ret < 0)
   {
     char buff[100];
-    sprintf(buff, "Can't start server: Bind on TCP/IP port. Got error: %d",
+    snprintf(buff, sizeof(buff), "Can't start server: Bind on TCP/IP port. Got error: %d",
```
grooverdan, line 2433: obviously fits

grooverdan (changes requested):
this takes a lot of review time for things compilers detect anyway.

client/mysql.cc
```diff
@@ -1239,11 +1239,10 @@ int main(int argc,char *argv[])
...
     else if (getenv("HOME"))
     {
-      histfile=(char*) my_malloc((uint) strlen(getenv("HOME"))
-				 + (uint) strlen("/.mysql_history")+2,
-				 MYF(MY_WME));
+      size_t histfile_sz = (uint) strlen(getenv("HOME")) + (uint) strlen("/.mysql_history") + 2;
+      histfile=(char*) my_malloc(histfile_sz, MYF(MY_WME));
       if (histfile)
-	sprintf(histfile,"%s/.mysql_history",getenv("HOME"));
+	snprintf(histfile, histfile_sz, "%s/.mysql_history",getenv("HOME"));
```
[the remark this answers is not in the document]
Chaloff: the flawfinder message for this particular string:
```./client/mysql.cc:1246: [4] (buffer) sprintf: Does not check for buffer overflows ([CWE-120](https://cwe.mitre.org/data/definitions/120.html)). Use sprintf_s, snprintf, or vsnprintf.```
I agree - in case the change is going to make some merge problem we probably need to remove it.

In general - the code as it is now does not create overflow. But the reason why Flawfinder complains (and some other static analyzer also can) - the code is not safe in general. In case messages or variable memory allocation is change we'll get seg fault here
Exchanging `sprintf` with `snprintf` for variables with known on compiling time size does not create any overhead just simply makes code cleaner and safer. I think we discussed it already with Andrew earlier

This applies to other places in the code which have no problem and strings fit now

File: client/mysql.cc [hunk 4]
```diff
@@ -3514,7 +3514,7 @@ static char *fieldflags2str(uint f) {
   ff2s_check_flag(ON_UPDATE_NOW);
 #undef ff2s_check_flag
   if (f)
-    sprintf(s, " unknows=0x%04x", f);
+    snprintf(s, buf_sz, " unknows=0x%04x", f);
```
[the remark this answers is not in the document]
Chaloff: Yes - thanks, didn't noticed it. Need to remove this part for now

client/mysql.cc
```diff
@@ -4230,7 +4230,7 @@ com_edit(String *buffer,char *line __attribute__((unused)))
   if ((error= system(buff)))
   {
     char errmsg[100];
-    sprintf(errmsg, "Command '%.40s' failed", buff);
+    snprintf(errmsg, sizeof(errmsg), "Command '%.40s' failed", buff);
```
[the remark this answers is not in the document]
vmarora: @grooverdan : is there any advantage of continuing to use sprintf?

sql/mysqld.cc
dlenski: The point of these changes is very specifically to make them _**even where it's “obvious” that there's no overrun possibility**_ in the current code.

1. This will reduce the cognitive load and potential for oversights and errors _in the future_: future reviewers won't have to manually (re)review so many occurrences of (unbounded) `sprintf` or `strcpy` to (re)confirm that they're “obviously” correct.
2. Compilers can optimize `snprintf` into `sprintf`, `strncpy` into `strcpy`, etc, when they can verify at compile time that no truncation is possible.
3. The vast majority of these changes are in error paths / slow paths.

    If any of them is in a truly hot path where it was plausibly performance-limiting and might require hand-optimization, we could revisit it individually.

client/mysql.cc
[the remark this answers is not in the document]
dlenski: In https://github.com/MariaDB/server/pull/2601#discussion_r1186384599, I make the case that there is _no disadvantage_ to switching to `snprintf`, and some key  _advantages_ to doing so, in terms of reducing the cognitive load on _human_ code reviewers.

knielsen:
@Chaloff , in case of overflow, snprintf() will silently truncate the result string, leaving a corrupt result.
Did you carefully understand the impact of each change, making sure you fully understand the full consequences of a truncated string in each place?

Sometimes the output string is just an error message where truncation has limited consequences, other times truncation can be very serious. Silent truncation can hide bugs from dynamic analysing tools like Valgrind and ASAN used in the testsuite. So it's very important not to do these kind of modifications blindly, and be sure not to make a change without fully understanding the code around it, even though this can be very time consuming.

Hope this helps,

 \- Kristian.

sql/sql_connect.cc
LinuxJedi, line 988: Is there an advantage to this over `safe_strcpy`?

client/mysql.cc
```diff
@@ -1263,19 +1261,19 @@ int main(int argc,char *argv[])
...
-      sprintf(histfile_tmp, "%s.TMP", histfile);
+      snprintf(histfile_tmp, histfile_tmp_sz, "%s.TMP", histfile);
     }
   }
 
 #endif
 
-  sprintf(buff, "%s",
+  snprintf(buff, sizeof(buff), "%s",
 	  "Type 'help;' or '\\h' for help. Type '\\c' to clear the current input statement.\n");
```
LinuxJedi, line 1277: I suspect this could be changed to an `safe_strcpy()`

LinuxJedi (commented):
In general, I approve. I don't think the scenario @knielsen mentions here is something to worry about with this particular PR, but it is something to watch out for.

I have questioned the advantage of using `snprintf` over `safe_strcat` in some places.

gkodinov (changes requested):
I'm terribly sorry that this has stalled for such a long time. But, I hope that you'd be willing to resume working on it nevertheless.

Here's how to get this started: 

* please rebase your changes to 10.11: this is where I've pushed my last sprintf->snprintf change. 
* please make sure the diff is correct: right now it says there's 8k changes in this PR. 

Once we get the PR cleaned up I will do a proper review.

gkodinov:
There has been no reply on this for more than 3 weeks now. Converting to draft. Please re-open if/when you intend to resume working on this.

PR #2604, opened 2023-04-14
[MDEV-30854] Do not use " as string delimiter in mariadb-tzinfo-to-sql

dlenski:
- [x] *The Jira issue number for this PR is: MDEV-30854*

## Description
If SQL_MODE contains ANSI_QUOTES (https://mariadb.com/kb/en/sql-mode/), then
the double-quote character (") is not a legal string delimiter.

In https://github.com/MariaDB/server/commit/13e77930e615f05cc74d408110e887b00e1abcc9#diff-a333d4ebb2d73b6361ef7dfebc86d883f7e19853b4a9eb85984b039058fae47cR2431-R2435,
@grooverdan introduced a case where the double-quote character would be used as
a string delimiter in the SQL queries generated by mariadb-tzinfo-to-sql.

This tool tool generates SQL queries which should be able to run on any
MariaDB server of the matching version.  Therefore, it should be extremely
conservative in the SQL that it outputs, in order to maximize the chance
that it can run regardless of the build or execution environment of the
server.

All new code of the whole pull request, including one or several files that are
either new files or modified ones, are contributed under the BSD-new license. I
am contributing on behalf of my employer Amazon Web Services, Inc.

## How can this PR be tested?

The preexisting MTR test `mysql_tzinfo_to_sql_symlink.test` 
actually [runs `mariadb-tzinfo-to-sql … > tz.sql` and executes the SQL which it outputs](https://github.com/MariaDB/server/blob/10.6/mysql-test/main/mysql_tzinfo_to_sql_symlink.test#L32-L35).
It continues to pass with minor modifications reflecting the expected changes in the SQL syntax; this
demonstrates that the generated SQL continues to work (in the environment where it is tested) after this PR.

See MDEV-18778, MDEV-28263, and MDEV-28782 for previous cases where MariaDB
has FAILED TO ENSURE that the generated timezone.sql actually works in
different build and execution environments. More test coverage is clearly
needed here to proactively detect such cases.

## Backward compatibility

This PR _improves_ compatibility, backwards and otherwise, because it makes the `mariadb-tzinfo-to-sql` output
compatible with MariaDB's parsing regardless of whether or not `SQL_MODE` includes `ANSI_QUOTES`.

## PR quality check


dlenski:
Build is all-:heavy_check_mark:.

Would be great to get some quick :eyes: on this from @grooverdan, since it's essentially a fix to his https://github.com/MariaDB/server/commit/13e77930e615f05cc74d408110e887b00e1abcc9.

**_Note that this is against 10.6, since that change was first merged in 10.6.8_**

grooverdan:
Ok. Build all passed previously too but that didn't mean it was right.

Can some tests that sets `ansi_quotes` and other SQL modes relevant to parsing the type of output generated by mariadb-tzinfo-to-sql be included in mtr.

The other/additional solution (if other sql modes generate a parsing fault and not fixable so easily) is to set the sql_mode to something known to work in the beginning of the output and restore its value at the end of the output.

dlenski:
> Ok. Build all passed previously too but that didn't mean it was right.
>
> Can some tests that sets ansi_quotes and other SQL modes …

Yes, those tests would be good to have to prevent _future_ bugs, but this here fix will at least fix the _already-known_ bug.

> … relevant to parsing the type of output generated by mariadb-tzinfo-to-sql be included in mtr.

That's the trouble with SQL_MODE. It can change the allowed syntax and semantics of the commands in various _extremely subtle_ ways which are hard to predict. Particularly [`SQL_MODE='ORACLE'`](https://mariadb.com/kb/en/sql-mode/#oracle)/`MSSQL`/`POSTGRESQL`… yikes.

> The other/additional solution (if other sql modes generate a parsing fault and not fixable so easily) is to set the sql_mode to something known to work in the beginning of the output and restore its value at the end of the output.

As discussed in [MDEV-28782](https://jira.mariadb.org/browse/MDEV-28782) and elsewhere, we can't guarantee that the `timezone.sql` is being run in a mode where it's possible to set variables. (`--bootstrap` would be the now-known exception).

If you're okay with a _backwards-incompatible_ change, where the `mariadb-tzinfo-to-sql` output becomes specified as legal only in specific execution environments, then saving and restoring `SQL_MODE` would be a reasonable approach too.

grooverdan:
I'd like to work in bootstrap too, regardless of the `sql_mode` there too. You're quite right in previous comments, bootstrap needs a cleanup/more tests, but not for here.

vuvova:
@dlenski what did you mean by
> As discussed in [MDEV-28782](https://jira.mariadb.org/browse/MDEV-28782) and elsewhere, we can't guarantee that the `timezone.sql` is being run in a mode where it's possible to set variables. (`--bootstrap` would be the now-known exception).

how could it be not possible to set variables?

_this question is unrelated to your change, I think the change is good regardless of whether it's possible to set variables or not_

grooverdan (approved):

dlenski:
I see this has been merged to 11.0, but not (yet?) to 11.1.

grooverdan:
Not yet, the [11.1 RC is scheduled](https://jira.mariadb.org/projects/MDEV?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page&status=unreleased) a week later than the <=11.0 releases.

PR #2606, opened 2023-04-17
MDEV-31062 : Reduce number of wsrep calls to server code from InnoDB

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-31062*

## Description

Thread executing wsrep transaction can't change during transaction execution. Similarly, thread executing high priority brute force (BF) transaction does not change during transaction execution. Therefore, in both cases there is no need to call server code after transaction has initialized.

InnoDB already stores information is this wsrep transaction to trx_t::wsrep and this is checked using trx->is_wsrep() function. Because, brute force transaction is always a wsrep transaction we can extend trx_t::wsrep variable so that value

0 == not wsrep transaction (and not BF)
1 == normal wsrep transaction
2 == high priority BF wsrep transaction
4 == high priority BF transaction is performing unique secondary index scan

These values can be set by calling server code on innobase_trx_init(). After that we can use trx_t::is_wsrep() and new function introduced in this patch trx_t::is_wsrep_BF(). Unique secondary index scan is determined later but it implies BF, so this patch removes unnecessary variable trx_t::wsrep_UK_scan. Patch introduces new functions trx_t::is_wsrep_UK_scan(), wsrep_begin_UK_scan() and wsrep_end_UK_scan() to handle unique secondary index scan.

This change reduces number of calls to server code from InnoDB and reduces code bloat on performance critical stages like acquiring record locks. Furthermore, it simplifies code on several locations.

## How can this PR be tested?

- [ X] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced*

## PR quality check

### DIFF
--- a/storage/innobase/include/trx0trx.h
+++ b/storage/innobase/include/trx0trx.h
@@ -734,14 +734,27 @@ public:
 	Transitions to COMMITTED are protected by trx_t::mutex. */
 	trx_state_t	state;
 #ifdef WITH_WSREP
-	/** whether wsrep_on(mysql_thd) held at the start of transaction */
-	bool		wsrep;
+	/** (1) whether wsrep_on(mysql_thd) held at the start of transaction
+	    (2) whether wsrep_thd_is_BF(mysql_thd) held at start of transaction
+	    (4) whether BF thread is performing unique secondary index scan
+	**/
+	static const byte TRX_WSREP_ON = 1;
+	static const byte TRX_WSREP_BF = 2;
+	static const byte TRX_WSREP_UK_SCAN = 4;
+	byte		wsrep;
+	/** true, if thread has wsrep_on  */
 	bool is_wsrep() const { return UNIV_UNLIKELY(wsrep); }
+	/** true, if thread is high priority brute force (BF)  */
+	bool is_wsrep_BF() const { return UNIV_UNLIKELY(wsrep & TRX_WSREP_BF); }
 	/** true, if BF thread is performing unique secondary index scanning */
-	bool wsrep_UK_scan;
-	bool is_wsrep_UK_scan() const { return UNIV_UNLIKELY(wsrep_UK_scan); }
+	bool is_wsrep_UK_scan() const { return UNIV_UNLIKELY(wsrep & TRX_WSREP_UK_SCAN); }
+	/** begin unique secondary index scan in wsrep BF thread */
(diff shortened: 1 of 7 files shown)


storage/innobase/include/trx0trx.h
```diff
@@ -734,14 +734,27 @@ struct trx_t {
...
-	/** whether wsrep_on(mysql_thd) held at the start of transaction */
-	bool		wsrep;
+	/** (1) whether wsrep_on(mysql_thd) held at the start of transaction
+	    (2) whether wsrep_thd_is_BF(mysql_thd) held at start of transaction
+	    (4) whether BF thread is performing unique secondary index scan
+	**/
+	static const byte TRX_WSREP_ON = 1;
+	static const byte TRX_WSREP_BF = 2;
+	static const byte TRX_WSREP_UK_SCAN = 4;
+	byte		wsrep;
```
dr-m, line 744: There is no Doxygen comment for the data member `wsrep` anymore. The constants are missing `constexpr`.

storage/innobase/trx/trx0trx.cc
```diff
@@ -374,7 +373,6 @@ trx_t *trx_create()
 
 #ifdef WITH_WSREP
 	trx->wsrep_event= NULL;
```
dr-m, line 375: Shouldn’t this be checked as part of `trx_t::assert_freed()`? Also the assertions on `trx_t::mod_tables` and `trx_t::lock` belong there.

dr-m (changes requested):

storage/innobase/trx/trx0trx.cc
janlindstrom: Sure but that is out of scope of this change.

storage/innobase/include/trx0trx.h
janlindstrom: I will fix the Doxygen (not sure its really used) but I used here same const as undo_no_t constant (see 10.6 version where constexpr was used). Can you use C++11 on 10.4?

dr-m: d9613b750c6d975beed0e1d7f0cb1a8dd3ad106b enabled the use of C++11

janlindstrom:
10.4 is EOL so not going to push this.

PR #2607, opened 2023-04-17
MDEV-31062 : Reduce number of wsrep calls to server code from InnoDB

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-31062*

## Description

Thread executing wsrep transaction can't change during transaction execution. Similarly, thread executing high priority brute force (BF) transaction does not change during transaction execution. Therefore, in both cases there is no need to call server code after transaction has initialized.

InnoDB already stores information is this wsrep transaction to trx_t::wsrep and this is checked using trx->is_wsrep() function. Because, brute force transaction is always a wsrep transaction we can extend trx_t::wsrep variable so that value

0 == not wsrep transaction (and not BF)
1 == normal wsrep transaction
2 == high priority BF wsrep transaction
4 == high priority BF transaction is performing unique secondary index scan

These values can be set by calling server code on innobase_trx_init(). After that we can use trx_t::is_wsrep() and new function introduced in this patch trx_t::is_wsrep_BF(). Unique secondary index scan is determined later but it implies BF. Patch introduces new functions wsrep_begin_UK_scan() and wsrep_end_UK_scan() to handle unique secondary index scan.

This change reduces number of call to server code from InnoDB and reduces code bloat on performance critical stages like acquiring record locks. Furthermore, it simplifies code on several locations.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -2182,14 +2182,17 @@ btr_cur_ins_lock_and_undo(
 			GAP-locking we mark that this transaction
 			is using unique key scan here. */
 			if ((type & (DICT_CLUSTERED | DICT_UNIQUE)) == DICT_UNIQUE
-			    && trx->is_wsrep()
-			    && wsrep_thd_is_BF(trx->mysql_thd, false)) {
-				trx->wsrep = 3;
+			    && trx->is_wsrep_BF()) {
+				trx->wsrep_begin_UK_scan();
 			}
 #endif /* WITH_WSREP */
-			if (dberr_t err = lock_rec_insert_check_and_lock(
-				    rec, btr_cur_get_block(cursor),
-				    index, thr, mtr, inherit)) {
+			dberr_t err = lock_rec_insert_check_and_lock(
+					rec, btr_cur_get_block(cursor),
+					index, thr, mtr, inherit);
+#ifdef WITH_WSREP
+			trx->wsrep_end_UK_scan();
+#endif /* WITH_WSREP */
+			if (err) {
 				return err;
 			}
 		}
--- a/storage/innobase/include/trx0trx.h
+++ b/storage/innobase/include/trx0trx.h
@@ -715,12 +715,27 @@ public:
   alignas(CPU_LEVEL1_DCACHE_LINESIZE) trx_lock_t lock;
 
 #ifdef WITH_WSREP
-  /** whether wsrep_on(mysql_thd) held at the start of transaction */
-  byte wsrep;
+  /** (1) whether wsrep_on(mysql_thd) held at the start of transaction
+      (2) whether wsrep_thd_is_BF(mysql_thd) held at start of transaction
+      (4) whether BF thread is performing unique secondary index scan
+  **/
+  static constexpr byte TRX_WSREP_ON = byte{1};
+  static constexpr byte TRX_WSREP_BF = byte{2};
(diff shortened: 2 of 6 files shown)


storage/innobase/include/trx0trx.h
```diff
@@ -715,12 +715,31 @@ struct trx_t : ilist_node<>
...
+  /** wsrep_on(mysql_thd) held at the start of transaction */
+  static constexpr byte TRX_WSREP_ON = byte{1};
+  /** wsrep_thd_is_BF(mysql_thd) held at start of transaction */
+  static constexpr byte TRX_WSREP_BF = byte{2};
+  /** BF thread is performing unique secondary index scan */
+  static constexpr byte TRX_WSREP_UK_SCAN = byte{4};
+  /** whether this transaction is running wsrep transaction with following sub-states:
+      (1) whether wsrep_on(mysql_thd) held at the start of transaction
+      (2) whether wsrep_thd_is_BF(mysql_thd) held at start of transaction
+      (4) whether BF thread is performing unique secondary index scan
```
dr-m, line 727: The `TRX_` prefix is redundant, because these are `static` members of `trx_t`.

The comment of `wsrep` could simply mention that it is a combination of the flags `WSREP_ON`, `WSREP_BF`, `WSREP_UK_SCAN`. The semantics of those flags is already documented in the per-flag comments.

storage/innobase/btr/btr0cur.cc
```diff
@@ -2182,14 +2182,17 @@ btr_cur_ins_lock_and_undo(
...
 #endif /* WITH_WSREP */
-			if (dberr_t err = lock_rec_insert_check_and_lock(
-				    rec, btr_cur_get_block(cursor),
-				    index, thr, mtr, inherit)) {
+			dberr_t err = lock_rec_insert_check_and_lock(
+					rec, btr_cur_get_block(cursor),
+					index, thr, mtr, inherit);
+#ifdef WITH_WSREP
+			trx->wsrep_end_UK_scan();
+#endif /* WITH_WSREP */
```
dr-m, line 2194: The indentation of the continuation lines is one TAB too deep.

Could we define `trx_t::wsrep_end_UK_scan()` as dummy empty functions `#ifndef WITH_WSREP`? That would remove some `#ifdef` pollution.

storage/innobase/lock/lock0lock.cc
```diff
@@ -5554,15 +5554,15 @@ lock_sec_rec_read_check_and_lock(
 	there is a probability for lock conflicts between two wsrep
 	high priority threads. To avoid this GAP-locking we mark that
 	this transaction is using unique key scan here. */
-	if (trx->is_wsrep() && wsrep_thd_is_BF(trx->mysql_thd, false))
-		trx->wsrep = 3;
+	if (trx->is_wsrep_BF())
+		trx->wsrep_begin_UK_scan();
```
dr-m, line 5558: The condition should be part of `trx_t::wsrep_begin_UK_scan()`.

dr-m (changes requested):
Looks good. I only have some style comments.

storage/innobase/include/trx0trx.h
janlindstrom: Done

storage/innobase/lock/lock0lock.cc
janlindstrom: Done

ankit-gautam23 (approved):
LGTM

PR #2608, opened 2023-04-18
MDEV-29293 MariaDB stuck on starting commit state

temeo:
- [x] *The Jira issue number for this PR is: MDEV-29293*

## Description
MariaDB stuck on starting commit state

The problem seems to be a deadlock between KILL command execution
and BF abort issued by an applier, where:
* KILL has locked victim's LOCK_thd_kill and LOCK_thd_data.
* Applier has innodb side global lock mutex and victim trx mutex.
* KILL is calling innobase_kill_query, and is blocked by innodb
  global lock mutex.
* Applier is in wsrep_innobase_kill_one_trx and is blocked by
  victim's LOCK_thd_kill.

The fix in this commit removes the TOI replication of KILL command
and makes KILL execution less intrusive operation. Aborting the
victim happens now by using awake_no_mutex() and ha_abort_transaction().
If the KILL happens when the transaction is committing, the
KILL operation is postponed to happen after the statement
has completed in order to avoid KILL to interrupt commit
processing.

Notable changes in this commit:
* wsrep client connections's error state may remain sticky after
  client connection is closed. This error message will then pop
  up for the next client session issuing first SQL statement.
  This problem raised with test galera.galera_bf_kill.
  The fix is to reset wsrep client error state, before a THD is
  reused for next connetion.
* Release THD locks in wsrep_abort_transaction when locking
  innodb mutexes. This guarantees same locking order as with applier
  BF aborting.
* BF abort from MDL was changed to do BF abort on server/wsrep-lib
  side first, and only then do the BF abort on InnoDB side. This
  removes the need to call back from InnoDB for BF aborts which originate
  from MDL and simplifies the locking.
* Removed wsrep_thd_set_wsrep_aborter() from service_wsrep.h.
  The manipulation of the wsrep_aborter can be done solely on
  server side. Moreover, it is now debug only variable and
  could be excluded from optimized builds.
* Remove LOCK_thd_kill from wsrep_thd_LOCK/UNLOCK to allow more
  fine grained locking for SR BF abort which may require locking
  of victim LOCK_thd_kill. Added explicit call for
  wsrep_thd_kill_LOCK/UNLOCK where appropriate.
* Wsrep-lib was updated to version which allows external
  locking for BF abort calls.

Changes to MTR tests:
* Disable galera_bf_abort_group_commit. This test is going to
  be removed (MDEV-30855).
* Record galera_gcache_recover_manytrx as result file was incomplete.
  Trivial change.
* Make galera_create_table_as_select more deterministic:
  Wait until CTAS execution has reached MDL wait for multi-master
  conflict case. Expected error from multi-master conflict is
  ER_QUERY_INTERRUPTED. This is because CTAS does not yet have open
  wsrep transaction when it is waiting for MDL, query gets interrupted
  instead of BF aborted. This should be addressed in separate task.
* A new test galera_kill_group_commit to verify correct behavior
  when KILL is executed while the transaction is committing.

## How can this PR be tested?

Test by running MTR suites galera, galera_3nodes, wsrep, galera_sr, galera_3nodes_sr.
See also JIRA ticket https://jira.mariadb.org/browse/MDEV-29293.

## PR quality check


temeo:
A note about merging. This fix does not merge cleanly to 10.6, which requires separate PR.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -18799,10 +18755,45 @@ wsrep_innobase_kill_one_trx(
...
-  /* Here we need to lock THD::LOCK_thd_data to protect from
-  concurrent usage or disconnect or delete. */
+  /* Grab reference to victim_trx before releasing the mutex, this will
+     prevent victim to release locks or commit while the mutex is
+     unlocked. The state may change to TRX_STATE_COMMITTED_IN_MEMORY.
+     See skip_lock_inheritance_n_ref in trx0trx.h. */
+  const trx_id_t victim_trx_id= victim_trx->id;
+retry_lock:
+  victim_trx->reference();
+  trx_mutex_exit(victim_trx);
```
dr-m, line 18765: Why is `lock_sys.mutex` not being released during the retry loop? It should be a bigger bottleneck than `victim_trx->mutex`.

temeo: Is it safe to release `lock_sys.mutex` here, this function is called from 
```
lock_rec_other_has_conflicting() -> wsrep_kill_victim() -> wsrep_innobase_kill_one_trx()
```
and
```
lock_table_other_has_incompatible() -> wsrep_kill_victim() -> wsrep_innobase_kill_one_trx()
```

Looking around `lock_rec_other_has_conflicting()` and `lock_table_other_has_incompatible()`, releasing `lock_sys.mutex` here does not look safe as the call happens in the middle of `lock_sys.rec_hash` iteration.

dlenski:
> A note about merging. This fix does not merge cleanly to 10.6, which requires separate PR.

It appears that this has been created as a new PR in #2609.

@temeo Perhaps you can update the PR titles to indicate the target versions, to distinguish them more easily?

sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/3f59bbeeaec751e9aabdc544324546f3c8326f0f

PR #2609, opened 2023-04-19
MDEV-29293 MariaDB stuck on starting commit state

temeo:
- [x] *The Jira issue number for this PR is: MDEV-29293*

## Description

This commit contains a merge from 10.5-MDEV-29293-squash into 10.6.

Although the bug MDEV-29293 was not reproducible with 10.6, the fix contains several improvements for wsrep KILL query and BF abort handling, and addresses the following issues:

* MDEV-30307 KILL command issued inside a transaction is problematic for galera replication: This commit will remove KILL TOI replication, so Galera side transaction context is not lost during KILL.
* MDEV-21075 KILL QUERY maintains nodes data consistency but breaks GTID sequence: This is fixed as well as KILL does not use TOI, and thus does not change GTID state.
* MDEV-30372 Assertion in wsrep-lib state: This was caused by BF abort or KILL when local transaction was in the middle of group commit. This commit disables THD::killed handling during commit, so the problem is avoided.
* MDEV-30963 Assertion failure !lock.was_chosen_as_deadlock_victim in trx0trx.h:1065: The assertion happened when the victim was BF aborted via MDL while it was committing. This commit changes MDL BF aborts so that transactions which are committing cannot be BF aborted via MDL. The RQG grammar attached in the issue could not reproduce the crash anymore.

Original commit message from 10.5 fix:

    MDEV-29293 MariaDB stuck on starting commit state

    The problem seems to be a deadlock between KILL command execution
    and BF abort issued by an applier, where:
    * KILL has locked victim's LOCK_thd_kill and LOCK_thd_data.
    * Applier has innodb side global lock mutex and victim trx mutex.
    * KILL is calling innobase_kill_query, and is blocked by innodb
      global lock mutex.
    * Applier is in wsrep_innobase_kill_one_trx and is blocked by
      victim's LOCK_thd_kill.

    The fix in this commit removes the TOI replication of KILL command
    and makes KILL execution less intrusive operation. Aborting the
    victim happens now by using awake_no_mutex() and ha_abort_transaction().
    If the KILL happens when the transaction is committing, the
    KILL operation is postponed to happen after the statement
    has completed in order to avoid KILL to interrupt commit
    processing.

    Notable changes in this commit:
    * wsrep client connections's error state may remain sticky after
      client connection is closed. This error message will then pop
      up for the next client session issuing first SQL statement.
      This problem raised with test galera.galera_bf_kill.
      The fix is to reset wsrep client error state, before a THD is
      reused for next connetion.
    * Release THD locks in wsrep_abort_transaction when locking
      innodb mutexes. This guarantees same locking order as with applier
      BF aborting.
    * BF abort from MDL was changed to do BF abort on server/wsrep-lib
      side first, and only then do the BF abort on InnoDB side. This
      removes the need to call back from InnoDB for BF aborts which originate
      from MDL and simplifies the locking.
    * Removed wsrep_thd_set_wsrep_aborter() from service_wsrep.h.
      The manipulation of the wsrep_aborter can be done solely on
      server side. Moreover, it is now debug only variable and
      could be excluded from optimized builds.
    * Remove LOCK_thd_kill from wsrep_thd_LOCK/UNLOCK to allow more
      fine grained locking for SR BF abort which may require locking
      of victim LOCK_thd_kill. Added explicit call for
      wsrep_thd_kill_LOCK/UNLOCK where appropriate.
    * Wsrep-lib was updated to version which allows external
      locking for BF abort calls.

    Changes to MTR tests:
    * Disable galera_bf_abort_group_commit. This test is going to
      be removed (MDEV-30855).
    * Make galera_var_retry_autocommit result more readable by echoing
      cases and expectations into result. Only one expected result for
      reap to verify that server returns expected status for query.
    * Record galera_gcache_recover_manytrx as result file was incomplete.
      Trivial change.
    * Make galera_create_table_as_select more deterministic:
      Wait until CTAS execution has reached MDL wait for multi-master
      conflict case. Expected error from multi-master conflict is
      ER_QUERY_INTERRUPTED. This is because CTAS does not yet have open
      wsrep transaction when it is waiting for MDL, query gets interrupted
      instead of BF aborted. This should be addressed in separate task.
    * A new test galera_bf_abort_registering to check that registering trx gets
      BF aborted through MDL.
    * A new test galera_kill_group_commit to verify correct behavior
      when KILL is executed while the transaction is committing.

    Co-authored-by: Seppo Jaakola <seppo.jaakola@iki.fi>
    Co-authored-by: Jan Lindström <jan.lindstrom@galeracluster.com>

## How can this PR be tested?

See MTR tests in this pull request.

## PR quality check
- [* ] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.


julienfritsch44:
I see no reviewer assigned and in Jira it's you Teemu who is assigned. So, who should review this? Julius?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -18668,103 +18668,91 @@ void lock_wait_wsrep_kill(trx_t *bf_trx, ulong thd_id, trx_id_t trx_id)
...
-	}
-
-	wsrep_thd_kill_UNLOCK(victim_thd);
-	DBUG_VOID_RETURN;
+static void wsrep_abort_transaction(handlerton *, THD *bf_thd, THD *victim_thd,
+                                    my_bool signal)
+{
+  DBUG_ENTER("wsrep_abort_transaction");
+  ut_ad(bf_thd);
+  ut_ad(victim_thd);
```
dr-m, line 18709: I think that it would be better to add `MY_ATTRIBUTE((nonnull))` on the function instead of adding assertions for checking the same.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -18668,103 +18668,91 @@ void lock_wait_wsrep_kill(trx_t *bf_trx, ulong thd_id, trx_id_t trx_id)
...
+    DEBUG_SYNC(bf_thd, "before_wsrep_thd_abort");
+    DBUG_EXECUTE_IF("sync.before_wsrep_thd_abort", {
+      const char act[]= "now "
+                        "SIGNAL sync.before_wsrep_thd_abort_reached "
+                        "WAIT_FOR signal.before_wsrep_thd_abort";
+      DBUG_ASSERT(!debug_sync_set_action(bf_thd, STRING_WITH_LEN(act)));
+    };);
+    victim_trx->lock.set_wsrep_victim();
+    victim_trx->mutex_unlock();
+    lock_sys.cancel_lock_wait_for_trx(victim_trx);
```
dr-m, line 18749: Is it absolutely necessary to enter a `DEBUG_SYNC` wait while holding `victim_trx->mutex`?

There is no guarantee that the transaction exists at the time `lock_sys_t::cancel_lock_wait_for_trx(trx_t *)` is called. The transaction commit could be blocked in `trx_t::commit_state()` until the moment `victim_trx->mutex_unlock()` is called. After that point, the call to `wsrep_commit_ordered()` might cause some synchronization; I am not familiar with that code. A little after that call, `trx_t::commit_cleanup()` can free the transaction object back to the transaction pool, and the memory can be reused for an unrelated transaction. In other words, the call `lock_sys.cancel_lock_wait_for_trx(victim_trx);` here may kill an innocent transaction if the intended `victim_trx` was committed right before the call.

I think that safe operation here needs to follow a similar locking pattern as `lock_sys_t::cancel_lock_wait_for_trx(trx_t*)`.

There should also be a comment that explains what the intended impact is on transactions that are not blocked in `lock_wait()` but waiting for something outside InnoDB, such as the next SQL statement to execute, or metadata lock on some table.

dr-m (changes requested):
This seems to break the test `perfschema.nesting`.

storage/innobase/handler/ha_innodb.cc
janlindstrom: My suggestion to this: Cache victim_trx->id while we hold victim_trx mutex. Then either extend ock_sys_t::cancel_lock_wait_for_trx(trx_t*) or introduce new wsrep version where after we have acquired lock_sys latch and lock_sys wait mutex we validate victim_trx id and state, we could even use trx_sys::find() to avoid referencing victim_trx before we know it is valid.

storage/innobase/handler/ha_innodb.cc
temeo: This function is called via handlerton function pointer, so the attribute should be added also to handlerton function pointer declaration to be effective. As server layer does not seem to have MY_ATTRIBUTE macro, this kind of change is out of scope of this PR.

dr-m (changes requested):
This looks good from the correctness point of view, but I would suggest some optimization.

storage/innobase/handler/ha_innodb.cc
dr-m: Sorry. I should have suggested `__attribute__`. There is a compatibility `#define` for it in `include/my_attribute.h`.

storage/innobase/handler/ha_innodb.cc
dr-m: Yes, storing the `trx_t::id` should be safe. Transaction identifiers are not reusable and InnoDB does not support wrap-around. At a rate of 10⁶ transactions per second, the space of 2⁴⁸ identifiers (actually 2⁴⁷ because a second one is consumed on transaction commit) would be consumed in 4 years and 5 months.

storage/innobase/handler/ha_innodb.cc
dr-m: I think that the assertions will cause warnings in newer `clang++` or `g++` now that you added the `__attribute__((nonnull))` to the function prototype.

dr-m (commented):
The InnoDB part looks OK now. The changes outside InnoDB will need to be reviewed by someone else.

sql/handler.h
```diff
@@ -1530,9 +1530,9 @@ struct handlerton
                             const char *query, uint query_length,
                             const char *db, const char *table_name);
 
-   void (*abort_transaction)(handlerton *hton, THD *bf_thd,
-			    THD *victim_thd, my_bool signal);
-   int (*set_checkpoint)(handlerton *hton, const XID* xid);
+   void (*abort_transaction)(handlerton *hton, THD *bf_thd, THD *victim_thd,
+                             my_bool signal) __attribute((nonnull));
```
dr-m, line 1534: It seems that this file is missing `#include <my_attribute.h>`, to provide a compatibility `#define` for Microsoft Windows.

sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/f307160218f8f9ed2528ffc685f49c4e2ae050b3

PR #2619, opened 2023-04-29
MDEV-23865 Create malloc function attribute

Yasuhiro-gh:
Add ATTRIBUTE_MALLOC macro.

ATTRIBUTE_MALLOC defines allocator that used at my_malloc func


grooverdan:
Can you squash commits up please.

Can you sign the CLA or provide consent of submission under the BSD-3 license?

Was there a reason that the deallocator was remove from the attribute?

Do you wish to annotate any other functions?

Did you perform some benchmarking?

https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

Yasuhiro-gh:
>Was there a reason that the deallocator was remove from the attribute?

the deallocator is not supported by clang, I didn't make a separate version for the GNU compiler

>Do you wish to annotate any other functions?

why not, do you have any suggestions?

>Did you perform some benchmarking?

I didn't. Is it required? If so, can you suggest, how can I make some basic benchmarking related?

dlenski:
As I understand it, the [`__attribute__((malloc))` annotation](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) is unlikely to have substantial _performance_ effects.

Its main effects are to help the compiler emit better warnings about potential memory safety issues.

FooBarrior:
@dlenski malloc attribute makes a better grained aliasing rules, so it improves the optimization process.
Though specifying the destructor only helps the data flow analysis, yes.

grooverdan (approved):

PR #2622, opened 2023-05-02
deb: autobake - add lunar

grooverdan:
fixes - https://buildbot.mariadb.org/#/builders/543/builds/1357


grooverdan:
lunar has 10.11 so that's the minimum version - new pr coming.

PR #2625, opened 2023-05-04
MDEV-28433 : Server crashes when wsrep_sst_donor and wsrep_cluster_ad…

janlindstrom:
…dress set to NULL

- [x] *The Jira issue number for this PR is: MDEV-28433*

## Description
Do not allow setting wsrep_sst_donor as NULL as it is incorrect value. User can use value '' (default) that represents same as NULL. Setting wsrep_cluster_address to NULL is already handled correctly.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/mdev-28433.result
+++ b/mysql-test/suite/galera/r/mdev-28433.result
@@ -1,0 +1,25 @@
+connection node_2;
+connection node_1;
+connection node_1;
+connection node_2;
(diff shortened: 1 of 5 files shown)


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/f102b595e8a65d910e34ea943eb562dc16a4256c

PR #2626, opened 2023-05-05
Hartmut mdev 30904

hholzgra:
- [x] *The Jira issue number for this PR is: MDEV-30904*

## Description
1. See MDEV-30905, ownership for the auth_pam_tool_dir
   were set in the POSTINST hook script, not by RPM file
   list entry itself, so "rpm --setugids" would reset
   the directory ownership from the "mysql" system user
   back to "root, breaking PAM authentication as the
   auth_pam_tool utility would no longer be visible to
   the PAM subprocesses
2. No actual server process output was changed, just
   the RPM file contents information and the POSTINST
   hook script
3. I do not expect any side effects at all
## How can this PR be tested?

I'm not sure what test mechanisms for RPM package installation we have in place, if any?

The PR can be tested by:

1) building RPM packages from branch with this patch applied
2) verify RPM contents

rpm -qp --dump MariaDB-server-10.[...].x86_64.rpm  | grep auth_pam_tool

on current distributed server RPMs:

```
Downloads > rpm -qp --dump MariaDB-server-10.5.18-1.el8.x86_64.rpm  | grep auth_pam_tool
[...]
/usr/lib64/mysql/plugin/auth_pam_tool_dir 0 1667398508 0000000000000000000000000000000000000000000000000000000000000000 040700 root root 0 0 0 X
/usr/lib64/mysql/plugin/auth_pam_tool_dir/auth_pam_tool 12480 1667398641 70644fc08bfb64845ef7877a23b72e39641b70d6dbe0d771b3e34d9fc04cee56 0104755 root root 0 0 0 X
```

on server RPM built with this patch applied:

```
[...]
/usr/lib64/mysql/plugin/auth_pam_tool_dir 0 1683232792 0000000000000000000000000000000000000000000000000000000000000000 040700 mysql root 0 0 0 X
/usr/lib64/mysql/plugin/auth_pam_tool_dir/auth_pam_tool 11912 1683232845 b1eb81ce5cb8076e6fb79bb37ca81728505ae881de49efd5393dcba157da59cf 0104755 root root 0 0 0 X
```

Note how the auth_pam_tool_dir ownership changed from "root root" to "mysql root"

3) install from the newly built RPMs and check that /usr/lib64/mysql/plugin/auth_pam_tool_dir still is owned by user "mysql" as before

4) run `rpm --setugids` and check that auth_pam_tool_dir is still owned by "mysql", that ownership did not switch from "mysql" to "root" as it happened with distributed RPMs so far

## Backward compatibility
Now backwards compatibility implications, file ownerships after rpm/yum/dnf/... install will be the same as before, just achieved in a slightly different way that will make them survive a "rpm -set

## PR quality check


grooverdan:
cursory looks ok. Pushed to https://buildbot.mariadb.org/#/grid?branch=bb-10.4-MDEV-30904-harmut-pkgtest so that generated packages on https://ci.mariadb.org/34918 (after build) can be tested against requirements.

hholzgra:
Turns out this does not work on a freshly installed VM where no "mysql" user exists yet.

PR #2629, opened 2023-05-10
MDEV-30456 ALTER TABLE algorithm clause not replicated

sjaakola:
- [x] *The Jira issue number for this PR is: MDEV-30456*

## Description
If user has specified the desired alter algorithm by using session variable alter_algorithm, and not specifying the algorithm in the ALTER SQL statement, then the ALTER will be processed in sending node according the algorithm chosen by the alter_algorithm variable. But in receiving nodes, appliers cannot see this session variable, and will use the default algorithm when executing the ALTER.

The fix in this commit, will check if user has set alter_algorithm variable and has not explicitly specified the algorithm clause in the ALTER statement, and in such case will rewrite the original ALTER statement appended by the algorithm clause. This rewritten ALTER statement will be used in replication.

## How can this PR be tested?
The commit has also new mtr test: galera.galera_alter_table_algorithm which verifies that ALTER query rewrite happens when needed

- [x ] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced*

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/galera_alter_table_algorithm.result
+++ b/mysql-test/suite/galera/r/galera_alter_table_algorithm.result
@@ -1,0 +1,64 @@
+connection node_2;
+connection node_1;
+CREATE TABLE t1 (f1 INTEGER) ENGINE=InnoDB;
+set global wsrep_on=OFF;
+reset master;
+set global wsrep_on=ON;
+connection node_2;
+set global wsrep_on=OFF;
+reset master;
+set global wsrep_on=ON;
+connection node_1;
+ALTER TABLE t1 ADD COLUMN f2 INTEGER, ALGORITHM=INSTANT;
+show binlog events limit 1 offset 4;
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+binlog.000001	367	Query	1	489	use `test`; ALTER TABLE t1 ADD COLUMN f2 INTEGER, ALGORITHM=INSTANT
+connection node_2;
+show binlog events limit 1 offset 4;
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+binlog.000001	367	Query	1	489	use `test`; ALTER TABLE t1 ADD COLUMN f2 INTEGER, ALGORITHM=INSTANT
+connection node_1;
+set global wsrep_on=OFF;
+reset master;
+set global wsrep_on=ON;
+connection node_2;
+set global wsrep_on=OFF;
+reset master;
+set global wsrep_on=ON;
+connection node_1;
+SET SESSION alter_algorithm='INSTANT';
+ALTER TABLE t1 ADD COLUMN f3 INTEGER;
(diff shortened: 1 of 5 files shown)


mysql-test/suite/galera/r/galera_alter_table_algorithm.result
```diff
@@ -0,0 +1,64 @@
...
+set global wsrep_on=ON;
+connection node_2;
+set global wsrep_on=OFF;
+reset master;
+set global wsrep_on=ON;
+connection node_1;
+ALTER TABLE t1 ADD COLUMN f2 INTEGER, ALGORITHM=INSTANT;
+show binlog events limit 1 offset 4;
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+binlog.000001	367	Query	1	489	use `test`; ALTER TABLE t1 ADD COLUMN f2 INTEGER, ALGORITHM=INSTANT
```
dlenski, line 15: Many of the fields here (`Pos`, `End_log_pos`, and maybe `Log_name` and `Server_id`?) will vary from run to run of this test.

The only thing the test result actually cares about (I think) is the `Info` column, right?

Consider adding this before each invocation of `show binlog events …` in the `.test` file:

```
--replace_column 1 LOG_NAME 2 POS 4 SERVER_ID 5 END_LOG_POS
```

See [mysql-test/main/alter_table.test](https://github.com/MariaDB/server/blob/11.0/mysql-test/main/alter_table.test#L252-L261), for example, where this is used in other pre-existing tests.

mysql-test/suite/galera/t/galera_alter_table_algorithm.cnf
```diff
@@ -0,0 +1,10 @@
+!include ../galera_2nodes.cnf
+
+[mysqld.1]
+log_slave_updates=ON
+log_bin=binlog
+
+[mysqld.2]
+log_slave_updates=ON
+log_bin=binlog
+
```
dlenski, line 10: Nit: extra blank line at end of file.

sql/sql_alter.h
```diff
@@ -108,6 +108,12 @@ class Alter_info
   // Number of partitions.
   uint                          num_parts;
 private:
+#ifdef WITH_WSREP
+  /* wsrep patch needs to peak the algorithm clause used in ALTER statement
```
dlenski, line 112: 1. Typo: I believe you meant the word *`pick`, not `peak`, here.
2. Why not simply make the `algorithm` member `public:` _unconditionally_?

dlenski (changes requested):
It appears that this bug and this PR apply specifically to the case of _Galera_ replication, and _not_ to other types of replication that MariaDB supports.

I suggest modifying the PR title and description to clarify this.

Perhaps something like the following\?

```
MDEV-30456 Galera replication ignores ALTER_ALGORITHM session variable
```

PR #2631, opened 2023-05-11
MDEV-30388 : Assertion `!wsrep_has_changes(thd) || (thd->lex->sql_com…

janlindstrom:
…mand == SQLCOM_CREATE_TABLE && !thd->is_current_stmt_binlog_format_row()) || thd->wsrep_cs().transaction().state() == wsrep::transaction::s_aborted' failed

- [x] *The Jira issue number for this PR is: MDEV-30388*

## Description

Problem for Galera is the fact that sequences are not really transactional. Sequence operation is committed immediately in sql_sequence.cd and later Galera could find out that we have changes but actual statement is not there anymore.

Therefore, we must make some restrictions what kind of sequences Galera can support.

(1) Galera cluster supports only sequences implemented by InnoDB storage engine. This is because Galera replication supports currently only InnoDB.

(2) We do not allow LOCK TABLE on sequence object and we do not allow sequence creation under LOCK TABLE, instead lock is released and we issue warning.

(3) We allow sequences with NOCACHE definition or with INCREMEMENT BY 0 CACHE=n definition. This makes sure that sequence values are unique accross Galera cluster.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/MDEV-18832.result
+++ b/mysql-test/suite/galera/r/MDEV-18832.result
@@ -1,17 +1,14 @@
 connection node_2;
 connection node_1;
-CREATE SEQUENCE Seq1_1  START WITH 1  INCREMENT BY 1;
+CREATE SEQUENCE Seq1_1  START WITH 1  INCREMENT BY 1 NOCACHE;
 CREATE TABLE t1 (Id int(11) NOT NULL, PRIMARY KEY (Id));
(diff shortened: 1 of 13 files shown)


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/28eaf66e18fbb2fc15c20c14f835c7c276c35cb2
Additional correction: https://github.com/MariaDB/server/commit/3a7b3113505b317f8f1288d18dd0fa19a6ac360a

PR #2632, opened 2023-05-11
Re-enable MTR test galera_sr.GCF-1060

sciascid:
Test seems to pass reliably now.

### DIFF
--- a/mysql-test/suite/galera_sr/disabled.def
+++ b/mysql-test/suite/galera_sr/disabled.def
@@ -10,6 +10,5 @@
 #
 ##############################################################################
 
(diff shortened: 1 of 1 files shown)


sysprg:
@sciascid @janlindstrom The failure happened again, the mtr log is here: https://buildbot.mariadb.net/buildbot/builders/kvm-rpm-centos74-amd64/builds/38140/steps/mtr-galera/logs/stdio
galera node logs are here: https://buildbot.mariadb.net/buildbot/builders/kvm-rpm-centos74-amd64/builds/38140/steps/mtr-galera/

PR #2633, opened 2023-05-11
server.cnf: adjust major version to 10.11

karmengc:
## Description
Update rpm file server.cnf to bump major version to 10.11 group.

## How can this PR be tested?
For example, adding some options for enabling SSL (ssl_key, ssl_cert, ssl_ca, ssl_cipher)

## PR quality check


LinuxJedi (approved):
Good catch, many thanks!

PR #2634, opened 2023-05-12
MDEV-30013 : Assertion `state() == s_aborting || state() == s_must_re…

janlindstrom:
…play' failed in int wsrep::transaction::after_rollback()

- [x] *The Jira issue number for this PR is: MDEV-30013*

## Description
This must be some kind of merge error because at ha_check_engine we just find out used engine or default engine. There is no need to roll-back transaction here even if engine is not supported as it will be handled later.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/mdev-30013.result
+++ b/mysql-test/suite/galera/r/mdev-30013.result
@@ -1,0 +1,17 @@
+connection node_2;
+connection node_1;
+INSTALL PLUGIN ARCHIVE SONAME 'ha_archive.so';
+CREATE TABLE t (a CHAR(1)) ENGINE=ARCHIVE;
+SHOW CREATE TABLE t;
+Table	Create Table
+t	CREATE TABLE `t` (
+  `a` char(1) DEFAULT NULL
+) ENGINE=ARCHIVE DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
+INSERT INTO t VALUES ('A');
+UNINSTALL SONAME 'ha_archive';
+Warnings:
+Warning	1620	Plugin is busy and will be uninstalled on shutdown
+ALTER TABLE t CHANGE COLUMN a a CHAR(2);
+INSERT INTO t (a) VALUES ('AB');
+ERROR 42000: Unknown storage engine 'ARCHIVE'
+DROP TABLE t;
--- a/mysql-test/suite/galera/t/mdev-30013.test
+++ b/mysql-test/suite/galera/t/mdev-30013.test
@@ -1,0 +1,15 @@
+--source include/galera_cluster.inc
+
+if (!$HA_ARCHIVE_SO) {
+  skip Needs Archive loadable plugin;
+}
+
+INSTALL PLUGIN ARCHIVE SONAME 'ha_archive.so';
+CREATE TABLE t (a CHAR(1)) ENGINE=ARCHIVE;
+SHOW CREATE TABLE t;
+INSERT INTO t VALUES ('A');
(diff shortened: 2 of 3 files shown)


mysql-test/suite/galera/r/mdev-30013.result
```diff
@@ -0,0 +1,17 @@
...
+t	CREATE TABLE `t` (
+  `a` char(1) DEFAULT NULL
+) ENGINE=ARCHIVE DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
+INSERT INTO t VALUES ('A');
+UNINSTALL SONAME 'ha_archive';
+Warnings:
+Warning	1620	Plugin is busy and will be uninstalled on shutdown
+ALTER TABLE t CHANGE COLUMN a a CHAR(2);
+INSERT INTO t (a) VALUES ('AB');
+ERROR 42000: Unknown storage engine 'ARCHIVE'
```
dlenski, line 16: Does the result captured by this test actually demonstrate the expected and desired behavior?

It appears to me that it reveals a further bug, or at least a misleading error message.

The error message says, "Plugin is busy and will be uninstalled _on shutdown_", but then the table is _clearly unusable_ immediately, rather than continuing to work until the engine is shut down.

mysql-test/suite/galera/t/mdev-30013.test
```diff
@@ -0,0 +1,15 @@
+--source include/galera_cluster.inc
+
+if (!$HA_ARCHIVE_SO) {
+  skip Needs Archive loadable plugin;
+}
+
+INSTALL PLUGIN ARCHIVE SONAME 'ha_archive.so';
```
dlenski, line 7: This should be fixed to work on Windows.

See many similar examples in the `archive` test suite, e.g. https://github.com/MariaDB/server/blob/11.1/mysql-test/suite/archive/archive_plugin.test#L7-L8

mysql-test/suite/galera/r/mdev-30013.result
janlindstrom: I tested this without Galera and as you can see result is exactly the same:

`SELECT @@global.wsrep_on;
@@global.wsrep_on
0
INSTALL PLUGIN ARCHIVE SONAME 'ha_archive.so';
CREATE TABLE t (a CHAR(1)) ENGINE=ARCHIVE;
SHOW CREATE TABLE t;
Table	Create Table
t	CREATE TABLE `t` (
  `a` char(1) DEFAULT NULL
) ENGINE=ARCHIVE DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
INSERT INTO t VALUES ('A');
UNINSTALL SONAME 'ha_archive';
Warnings:
Warning	1620	Plugin is busy and will be uninstalled on shutdown
ALTER TABLE t CHANGE COLUMN a a CHAR(2);
INSERT INTO t (a) VALUES ('AB');
ERROR 42000: Unknown storage engine 'ARCHIVE'
`

Therefore, while I agree this looks not desired behavior it belong to different MDEV and to MariaDB developers.

mysql-test/suite/galera/t/mdev-30013.test
janlindstrom: Galera does not support Windows and these tests are not run on Windows. Therefore, I do not see point on this.

sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/ffd5d74c4f0e0812b461b5c6990abdf99ed038bf

PR #2635, opened 2023-05-12
MDEV-31118: Make Salsa-CI file work again for MariaDB 10.9 and 10.10

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-31118*

## Description
PR contains small changes to re-sync Salsa-CI file and some minor changes to ensure MariaDB 10.9 and 10.10 can also be tested on Debian Salsa-CI. These commit are not intended to 10.11 and above as they need small amount of adaptation to make sure there won't be breakage.

Things that are beyond thins PR are:

- Package lsb-base on Lintian. It has to be addressed in other PR probably just turning OFF.
- MariaDB 10.9 installation on MariaDB 10.9 should be checked what is the main reason why it's not working

## How can this PR be tested?
Current situation of Salsa-Ci file is:
https://salsa.debian.org/illuusio/mariadb-server/-/pipelines/526974

With this PR situation changes:
https://salsa.debian.org/illuusio/mariadb-server/-/pipelines/526990

## Backward compatibility
PR ensures Backward compatibility

## PR quality check


ottok:
The changes in this PR look mostly OK and would be fine for 10.9 and 10.10.

While bisecting for the arm64 cross-build regression in 10.11.3 I also created https://salsa.debian.org/otto/mariadb-server/-/commit/8337f9d57325c1fa57783823f95757236467e76b from which you can take some inspiration if you want.

illuusio:
> The changes in this PR look mostly OK and would be fine for 10.9 and 10.10.
> 
> While bisecting for the arm64 cross-build regression in 10.11.3 I also created https://salsa.debian.org/otto/mariadb-server/-/commits/10.9-salsa-ci-2023-05 from which you can take some inspiration if you want.

Thank you for pointing out I'll see how this can be incorporated.

illuusio:
@ottok I just can't figure out how to get Debian native build not fail with not enough disk space error. Any tips?

ottok:
> @ottok I just can't figure out how to get Debian native build not fail with not enough disk space error. Any tips?

Try building it locally and then run [ncdu](https://dev.yorhel.nl/ncdu/scr) on the build directory to see what takes up most of the disk space. After that slim down the build to fit within limits. Without any evidence I would take a wild guess that Columnstore hogs a huge amount of disk space, but could be something else too.

debian/salsa-ci.yml
```diff
@@ -112,35 +120,64 @@ blhc:
...
+  apt-get -qq install --no-install-recommends --yes ./libreadline5_5.2+dfsg-3+b13_amd64.deb
+
 .test-enable-bullseye-repos: &test-enable-bullseye-repos
   # Replace any old repos with just Sid
   - echo 'deb http://deb.debian.org/debian bullseye main' > /etc/apt/sources.list
   # Upgrade minimal stack first
   - apt-get update -qq
   - apt-get install -y apt
 
-.test-enable-bullseye-backports-repos: &test-enable-bullseye-backports-repos |
```
ottok, line 122: Why do you rename this to `buster` and then you do the exact opposite rename in the section below? Can't they just stay as-is?

ottok (commented):
Most of the changes make sense, but should we fix the 10.5 branch first, and then merge upwards adding per version fixes after each merge?

PR #2636, opened 2023-05-12
MDEV-30951: Fix small perlcritic and enable modern Perl

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-30951*

## Description
- Add Modern Perl headers. Perl 5.16 is still fairly old from 2012.
- Enable UTF-8, warnings and make script 'strict'
- Small fixes for perlcritic reported problems and some crashes
```
I/O layer ":utf8" used at line 268, column 16.  Use ":encoding(UTF-8)" to get strict validation.  (Severity: 5)
"return" statement with explicit "undef" at line 806, column 4.  See page 199 of PBP.  (Severity: 5)
"return" statement with explicit "undef" at line 6844, column 4.  See page 199 of PBP.  (Severity: 5)
"return" statement with explicit "undef" at line 7524, column 4.  See page 199 of PBP.  (Severity: 5)
"return" statement with explicit "undef" at line 7527, column 4.  See page 199 of PBP.  (Severity: 5)
"return" statement with explicit "undef" at line 7599, column 4.  See page 199 of PBP.  (Severity: 5)
"return" statement with explicit "undef" at line 7602, column 4.  See page 199 of PBP.  (Severity: 5)
Expression form of "eval" at line 7784, column 4.  See page 161 of PBP.  (Severity: 5)
Expression form of "eval" at line 7806, column 4.  See page 161 of PBP.  (Severity: 5)
Glob written as <...> at line 8016, column 25.  See page 167 of PBP.  (Severity: 5)
"return" statement followed by "sort" at line 9195, column 60.  Behavior is undefined if called in scalar context.  (Severity: 5)
Expression form of "eval" at line 9846, column 10.  See page 161 of PBP.  (Severity: 5)
```

## How can this PR be tested?
This should be tested as root or user which has admin rights. If user does not have admin rights it can't get some of needed information. 
Command that can be used to test this PR is:
`perl debian/additions/innotop/innotop`

It show different views with keys like A, B, C, and D (there is lot of other view and they can get with key ?). For example before PR B and D view crashes with: `Use of uninitialized value $text in pattern match (m//) at /home/tuukka/src/c/mariadb/mariadb-server/debian/additions/innotop/innotop line 615` and after PR they should all work.

## Backward compatibility
This breaks little bit of  Backward compatibility as Perl 5.16 (released in 2012) is minimum required version but even Debian  9 (which is EOL) has newer version of Perl. Also warnings and strict is now required so it'll crash if there is problem.

## PR quality check

### DIFF
--- a/debian/additions/innotop/innotop
+++ b/debian/additions/innotop/innotop
@@ -20,6 +20,9 @@
 # Street, Fifth Floor, Boston, MA 02110-1335 USA
 
 use strict;
+use warnings;
+use utf8;
+use feature ':5.16';
 use warnings FATAL => 'all';
 
 our $VERSION = '1.11.4';
@@ -265,7 +268,7 @@ sub get_dbh {
             $dbh->do($sql);
             MKDEBUG && _d('Enabling charset for STDOUT');
             if ( $charset eq 'utf8' ) {
-               binmode(STDOUT, ':utf8')
+               binmode(STDOUT, ':encoding(UTF-8)')
                   or die "Can't binmode(STDOUT, ':utf8'): $OS_ERROR";
             }
             else {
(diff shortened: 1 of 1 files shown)


LinuxJedi (approved):
I think a 11 year old version of Perl is a more than acceptable requirement :)

LinuxJedi:
> What is the status of `mytop` @illuusio ? If it doesn't pass can we include it too in this PR?

Whether or not it should be modified, that is out of scope for the Jira and PR. We don't want to tack too much on in one go.

PR #2638, opened 2023-05-14
Fix building with modern versions of LibreSSL

brad0:
Fix building with LibreSSL with opaque EVP_MD_CTX and EVP_CIPHER_CTX.

### DIFF
--- a/include/ssl_compat.h
+++ b/include/ssl_compat.h
@@ -62,6 +62,13 @@
 #endif /* HAVE_OPENSSL11 */
 #endif
 
+#ifdef LIBRESSL_VERSION_NUMBER
+#define EVP_MD_CTX_SIZE			64
+#define EVP_CIPHER_CTX_SIZE             192
+#define EVP_MD_CTX_init(X) do { memset((X), 0, EVP_MD_CTX_SIZE); EVP_MD_CTX_reset(X); } while(0)
+#define EVP_CIPHER_CTX_init(X) do { memset((X), 0, EVP_CIPHER_CTX_SIZE); EVP_CIPHER_CTX_reset(X); } while(0)
+#else
+
 #ifdef HAVE_WOLFSSL
 #define EVP_MD_CTX_SIZE                 sizeof(wc_Md5)
 #endif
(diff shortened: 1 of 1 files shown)


include/ssl_compat.h
```diff
@@ -89,6 +96,7 @@
 #define X509_get0_notAfter(X) X509_get_notAfter(X)
 #endif
 #endif
+#endif
```
dlenski, line 99: Seems like there's a very high likelihood of confusion with all of these nested but not indented preprocessor `#if`s.

Should they be indented?

include/ssl_compat.h
```diff
@@ -62,6 +62,13 @@
 #endif /* HAVE_OPENSSL11 */
 #endif
 
+#ifdef LIBRESSL_VERSION_NUMBER
+#define EVP_MD_CTX_SIZE			64
+#define EVP_CIPHER_CTX_SIZE             192
+#define EVP_MD_CTX_init(X) do { memset((X), 0, EVP_MD_CTX_SIZE); EVP_MD_CTX_reset(X); } while(0)
+#define EVP_CIPHER_CTX_init(X) do { memset((X), 0, EVP_CIPHER_CTX_SIZE); EVP_CIPHER_CTX_reset(X); } while(0)
+#else
```
grooverdan, line 70: To avoid another nesting and have an `#endif` here, afaik  only a `#ifndef EVP_CIPHER_CTX_SIZE` around the `#define EVP_CIPHER_CTX_SIZE` below is needed.

include/ssl_compat.h
brad0: It gets confusing pretty quickly with such mazes.

LinuxJedi: Or maybe add things like `/* ! HAVE_WOLFSSL */` to the `#endif`s for clarity?

brad0:
Circling back to this.

I noticed this commit. I believe with this alone we should be good without needing this patch.

https://github.com/MariaDB/server/commit/9c0e91a27cc58d1b5076435e125c70cd277fd989

brad0:
So it looks like things are good with the 10.9 branch as is. It's just the LibreSSL / OpenSSL 3.0.0 and the CiscoSSL commits needs to be merged back and forth on the branches.

PR #2643, opened 2023-05-18
Add preprocessor whitespace to CODING_STANDARDS.md

LinuxJedi:
## Description
Adds minor section about preprocessor indentation to coding standards.

Against 11.0 since that is where the first version of the coding standards document lives.

## PR quality check

### DIFF
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -125,14 +125,52 @@ Consecutive functions should be separated with 2 empty lines in between
 ```cpp
 void my_function_1()
 {
-  <logic>  
+  <logic>
 }
 
 
 void my_function_2()
 {
-  <logic>  
+  <logic>
 }
+```
+
+#### Preprocessor directives
+
+Compiler preprocessor directives should have no indentation to them, even if in the middle of indented code.
+For example:
+
+```c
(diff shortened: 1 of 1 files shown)


grooverdan:
Do you want to add something like "comments reflecting the original #if condition can be appended to #else / #endif where the complexity/nesting of preprocessor directives requires additional clarity." e.g:

```
#else /* ! EMBEDDED_LIBRARY*/
```

ericherman:
Thanks for cleaning up the pre-EOL whitespace.

dlenski:
> Do you want to add something like "comments reflecting the original #if condition can be appended to #else / #endif where the complexity/nesting of preprocessor directives requires additional clarity." e.g:
> 
> ```
> #else /* ! EMBEDDED_LIBRARY*/
> ```

I agree with this suggestion. Adding these comments makes it a lot less error-prone to _read_ and modify nested `#if`s.

LinuxJedi:
> Do you want to add something like "comments reflecting the original #if condition can be appended to #else / #endif where the complexity/nesting of preprocessor directives requires additional clarity." e.g:
> 
> ```
> #else /* ! EMBEDDED_LIBRARY*/
> ```

Great suggestion, this has now been added.

grooverdan (approved):

PR #2646, opened 2023-05-28
Deb: Fix blocksize check to use $mysql_datadir/$datadir correctly

ottok:
## Description

In commit f99a8918 this line was changed to not use awk, and new version copied both to init file and preinst file but overlooking that they use different variable names.

Also fix minor syntax issues to make Shellcheck happy.

## PR quality check


illuusio:
This is fix that should get in.  Could you take a look at #2650 as it should be fixed before getting this in?

grooverdan:
in the interest of time, just merged this first while protected branches was green. #2650 should be a quick rebase now.

PR #2651, opened 2023-05-30
MDEV-35826 Fix COLUMNSTORE in Debian CI builds

LinuxJedi:
Debian building tried to implicitly enable ColumnStore, but it needs to be explicitly enabled to work. This should fix the AARCH64 BuildBot builders.


LinuxJedi:
Updated patch due to `CMAKE_SYSTEM_PROCESSOR` being `arm64` on Debian 11 BuildBot, not `aarch64`.

ottok (commented):
Based on the changes I see now, the commit message should be:

```
Add `arm64` to the supported platforms for ColumnStore Debian buils

Debian build tried to implicitly enable ColumnStore on arm64, but for this
to work, the boost/flex dependencies need to be there.
```

LinuxJedi:
> Based on the changes I see now, the commit message should be:
> 
> ```
> Add `arm64` to the supported platforms for ColumnStore Debian buils
> 
> Debian build tried to implicitly enable ColumnStore on arm64, but for this
> to work, the boost/flex dependencies need to be there.
> ```

That isn't quite right though. ColumnStore has been supported on ARM64 Debian for a long time. We had a builder that should have been testing it but we found it wasn't working.

The commit message may change before this is merged.

grooverdan:
Part of the reason columnstore/CMakeLists.txt needed changing is debian/rules `-DCMAKE_SYSTEM_PROCESSOR=$(DEB_HOST_ARCH)` is a really dumb thing to do. Debian's architecture terms are so non-standard it was bound to be trouble.

fauust:
@grooverdan please remember to update https://github.com/MariaDB/buildbot/commit/6e20c378738970314173235c08e3d6c30da87af3 once this is merged. Not sure how we could automate this and make it an alert, but it would be very good.

grooverdan:
with #2893, cmake change no longer required.

PR #2653, opened 2023-05-30
MDEV-30921: Use safe function snprintf

Chaloff:
- [x] *The Jira issue number for this PR is: MDEV-30921*

## Description
Replace unsafe usage of sprintf() with snprintf in storage/perfschema/table_helper.h and replace macro with inline function to make code safer and fix MDEV-30921

## How can this PR be tested?
All build stages pass for these commits.

## Backward compatibility


Chaloff:
I'm closing this PR. The problem on MacOS seems to be unreproducible hence we don't need this change

PR #2655, opened 2023-05-31
MDEV-29894: Calling a function from a different database in a slave side trigger crashes

bnestere:
When opening and locking tables, if triggers will be invoked in a
separate database, thd->set_db() is invoked, thus frees the memory
and headers which thd->db had previously pointed to. In row based
replication, the event execution logic initializes thd->db to point
to the database which the event targets, which is owned by the
corresponding table share. The problem then, is that during the
table opening and locking process for a row event, memory which
belongs to the table share would be freed, which is not valid.

This patch replaces the thd->reset_db() calls to thd->set_db(),
which copies-by-value, rather than by reference. Then when the
memory is freed, our copy of memory is freed, rather than memory
which belongs to a table share.

Additionally, the function backup_current_db_name() from sql_db.cc
is exposed for use in replication.

Note for review: an alternative approach would be to use the
functions already exposed in sql_db.cc: mysql_opt_change_db()
and mysql_change_db(); however, mysql_change_db() has the following
comment:

/*
...
  This function is not the only way to switch the database that is
  currently employed. When the replication slave thread switches the
  database before executing a query, it calls thd->set_db directly.
  However, if the query, in turn, uses a stored routine, the stored routine
  will use this function, even if it's run on the slave.
...
*/

and so I kept the logic as-is for now for consistency.

The first commit in this PR is the MTR regression test which shows
the problem, and the second commit is the code patch.

## Description
TODO: fill description here

## How can this PR be tested?

## Backward compatibility

## PR quality check

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_row_trigger_multi_db.test
+++ b/mysql-test/suite/rpl/t/rpl_row_trigger_multi_db.test
@@ -1,0 +1,98 @@
+#
+#   This test ensures that a table share's database name is not freed when
+# using row based replication with triggers that open different databases
+#
+#
+# References:
+#   MDEV-29894: Calling a function from a different database in a slave side
+#               trigger crashes
+#
+--source include/master-slave.inc
+--source include/have_binlog_format_row.inc
+
+--connection slave
+--let $old_slave_run_triggers= `SELECT @@global.slave_run_triggers_for_rbr`
+set global slave_run_triggers_for_rbr=1;
+
+--connection master
+CREATE TABLE t1 (a int);
+--sync_slave_with_master
+
+--connection slave
+CREATE DATABASE db2;
+CREATE FUNCTION db2.get_value(a INT) RETURNS int(2)  RETURN 0;
+
+--echo #
+--echo # Test Insert_rows_log_event
+
+--connection slave
+DELIMITER //;
+CREATE TRIGGER tr_ins BEFORE INSERT ON t1 FOR EACH ROW BEGIN
+DECLARE a INT;
+SET a = db2.get_value(1);
+END//
+DELIMITER ;//
+
+--connection master
+INSERT INTO t1 VALUES (1);
+--sync_slave_with_master
+
+--connection slave
+DROP TRIGGER tr_ins;
+
+
+--echo #
+--echo # Test Update_rows_log_event
+--connection master
+--let $row_val=5
+--eval INSERT INTO t1 VALUES ($row_val)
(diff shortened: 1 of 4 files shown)


mysql-test/suite/rpl/t/rpl_row_trigger_multi_db.test
```diff
@@ -0,0 +1,95 @@
...
+#
+# References:
+#   MDEV-29894: Calling a function from a different database in a slave side
+#               trigger crashes
+#
+--source include/master-slave.inc
+--source include/have_binlog_format_row.inc
+
+--connection slave
+--let $old_slave_run_triggers= `SELECT @@global.slave_run_triggers_for_rbr`
```
andrelkin, line 14: We need RAII here too :-). Not today though.

mysql-test/suite/rpl/t/rpl_row_trigger_multi_db.test
```diff
@@ -0,0 +1,95 @@
...
+DELIMITER //;
+CREATE TRIGGER tr_upd BEFORE UPDATE ON t1 FOR EACH ROW BEGIN
+DECLARE a INT;
+SET a = db2.get_value(1);
+END//
+DELIMITER ;//
+
+--connection master
+--eval UPDATE t1 SET a=a+1 WHERE a=$row_val
+--sync_slave_with_master
```
andrelkin, line 60: For better readability I found good to underline a new connection context one newline skip from the preceding block. `--sync_slave_with_master` of course begins a new one.

sql/log_event_server.cc
```diff
@@ -5749,6 +5750,7 @@ int Rows_log_event::do_apply_event(rpl_group_info *rgi)
      */
     rgi->set_row_stmt_start_timestamp();
 
+
```
andrelkin, line 5753: A residual of edits to remove.

sql/log_event_server.cc
```diff
@@ -7601,15 +7603,13 @@ Write_rows_log_event::do_exec_row(rpl_group_info *rgi)
 {
   DBUG_ASSERT(m_table != NULL);
   const char *tmp= thd->get_proc_info();
-  LEX_CSTRING tmp_db= thd->db;
```
andrelkin, line 7604: This is a victorious trains of deletion. I did not expect that :-)!

andrelkin (approved):
Looks good though the test lines can be brushed.

PR #2657, opened 2023-06-01
Fixup warnings for safe_strcpy

LinuxJedi:
This fixes warnings in safe_strcpy when compiling with mysql_release build config.

This should actually be slightly less expensive than the previous implementation. It works similarly to the glibc `strncpy()` and now only executes one `memset()` on the destination. Previously one would have been executed in the `strncpy()` and another in this function.

## How can this PR be tested?

Compile with `-DBUILD_CONFIG=mysql_release` and watch to see if the array bounds and truncation warnings associated with this function go away. MTR should still pass fine.

## PR quality check


illuusio:
Just wondering is this better approach than `strncpy` as this seems to be `strncpy` code?

LinuxJedi:
> Just wondering is this better approach than `strncpy` as this seems to be `strncpy` code?

The problems with `strncpy` are you do not know if there has been a truncation, and you can't guarantee it is `NUL` terminated when overflowing. This solves these problems.

The original function used `strncpy` with some extra logic to do the `NUL` and truncation checks. Unfortunately that logic caused compiler warnings due to the compiler not understanding a condition would never be hit if the source character array was too small. This could be solved with an extra `strlen`, but this becomes expensive. Warnings could be turned off for this segment of code, but that masks the occasional real problem.

This patch uses something similar to the glibc `strncpy` (note that other libc implementations and behaviours are different), but with several key differences (actually most of it is different) to handle the truncation without performance penalty. In fact this method should be a little faster for most cases as there is only one `memset` now instead of two.

cvicentiu:
@LinuxJedi Can you have a look at this patch from Monty to see if it gets rid of the warnings? It's actually a bit cleaner than my initial code and makes use of documented behaviour of strncpy, where it does a memset on its own (yep... I didn't notice this when I wrote my patch).

https://github.com/MariaDB/server/commit/a7adfd4c52307876d68ad3386cefd3757ee66e92

LinuxJedi:
> @LinuxJedi Can you have a look at this patch from Monty to see if it gets rid of the warnings? It's actually a bit cleaner than my initial code and makes use of documented behaviour of strncpy, where it does a memset on its own (yep... I didn't notice this when I wrote my patch).
> 
> https://github.com/MariaDB/server/commit/a7adfd4c52307876d68ad3386cefd3757ee66e92
> 
> 

If I'm reading his correctly it will flag as a truncation when src and dest are the same length.

LinuxJedi:
> > If I'm reading his correctly it will flag as a truncation when src and dest are the same length.
> 
> I don't think I understand your comment. Can you show the failure case?

You are right, in my head the "aa" case would have truncated, but I realise I was wrong. I'll close this in favour of Monty's solution then.

PR #2659, opened 2023-06-02
Fix random InnoDB MSAN failures

LinuxJedi:
This is difficult to trigger, but is a false positive that will fire on certain InnoDB tests in MSAN depending on several random factors.

The hit is a false-positive and there is already a code comment indicating this.

Patch was written by Marko and tested by Andrew.

## How can this PR be tested?

Running MSAN buildbot many times over. Appears to be easier to trigger on the older Ubuntu MSAN builder, but I'm able to trigger it locally on using the Debian builder.

## PR quality check


LinuxJedi:
Against 10.5 as we don't do MSAN testing in 10.4.

LinuxJedi:
I'm going to close this pull request as after discussions with @dr-m, further testing is required.

PR #2662, opened 2023-06-05
MDEV-30587: TLS/SSL error: unexpected eof while reading

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-30587*

## Description

Like the libmariadb commit of the same title (https://github.com/mariadb-corporation/mariadb-connector-c/pull/222), we don't want logs filling up with TLS errors due to abrupt termination.

The server with this commit tolerates a client quiet shutdown which is set the the libmariadb ma_tls_close in libmariadb/secure/openssl.c.

We also adjust some of the tests here so they will pass successfully when run with mtr --ssl. These add the 2026 (TLS alert) as an allowed error if the connection if force closed. This isn't a complete set of test case fixes to make mtr --ssl work for all.

## How can this PR be tested?

```
mysql-test/mtr --parallel=auto --ssl --force --max-test-fail=100
```

Most spider tests fail with:
```
CURRENT_TEST: spider.direct_left_right_join_nullable
mysqltest: Could not open connection 'default': 2026 TLS/SSL error: self-signed certificate in certificate chain
```

Passes most other tests, a few main/perf schema tests are affected by a use of a TLS monitoring test thread.

## Backward compatibility

The define only occurs if the test is there. Adjusting the tests for an additional error makes addition to buildbot later easier.

## PR quality check


vuvova:
Let's not hide the new OpenSSL 3.x error, but only update tests to expect it (in case someone runs mtr with `--ssl`)

grooverdan:
This isn't just a test case fix. Its hiding an error we've induced because of the `SSL_set_quiet_shutdown(ssl, 1)` call on both client and server. To hide it in the test cases is leaving the users exposed to frequent warnings of little value. Its not a true warning as fixed userspace protocol prevents this being exploitable.

vuvova:
I didn't see this error in most of the tests, only there where a server or a connection is killed from within the test. A normal client disconnect doesn't cause it, why would users be exposed to frequent warnings of little value?

mysql-test/include/wait_until_connected_again.inc
```diff
@@ -11,7 +11,7 @@ let $counter= 5000;
 let $mysql_errno= 9999;
 while ($mysql_errno)
 {
-  --error 0,ER_ACCESS_DENIED_ERROR,ER_SERVER_SHUTDOWN,ER_CONNECTION_KILLED,ER_LOCK_WAIT_TIMEOUT,2002,2006,2013,HA_ERR_NO_ENCRYPTION
+  --error 0,ER_ACCESS_DENIED_ERROR,ER_SERVER_SHUTDOWN,ER_CONNECTION_KILLED,ER_LOCK_WAIT_TIMEOUT,2002,2006,2013,2026,HA_ERR_NO_ENCRYPTION
```
dlenski, line 14: The source and meaning of these "client side" errors is very non-obvious in MTR tests, and hard to maintain or audit for completeness/correctness.

For example, when !2581 is merged, will _client-side_ error [2026 `CR_SSL_CONNECTION_ERROR`](https://github.com/mariadb-corporation/mariadb-connector-c/blob/master/include/errmsg.h#L44-L87) still be the right one to detect here? Or might it change to the _server-side_ `ER_SECURE_TRANSPORT_REQUIRED` which is added by that PR?

In general, seems to me like MTR ought to be aware of the symbolic names of [the client-side `CR_*` errors defined in `libmariadb`](https://github.com/mariadb-corporation/mariadb-connector-c/blob/master/include/errmsg.h#L44-L87), so that the tests don't contain these long lists of hard-to-discover and hard-to-remember integer constants.

mysql-test/suite/binlog_encryption/my.cnf
```diff
@@ -1,5 +1,4 @@
-!include include/default_mysqld.cnf
-!include include/default_client.cnf
+!include include/default_my.cnf
```
dlenski, line 1: Appears that the vast majority of the changes in this PR are due to replacing `default_mysqld.cnf` and/or `default_client.cnf` with `default_my.cnf`… rather than with the stated purpose of addressing https://jira.mariadb.org/browse/MDEV-30587.

The actual changes for MDEV-30587 are addressed mainly in the changes to `userstat.{test,result}` and `wait_until_connected_again.inc`… along with a few other files (`ha.result` and `rpl*.test`) which are _very_ easy to overlook amidst all the include file replacements.

At least to me, it's not apparent from this PR…

1. If [`default_my.cnf`](https://github.com/MariaDB/server/blob/11.1/mysql-test/include/default_my.cnf) is in fact "equivalent" to the combination of `default_mysqld.cnf` and `default_client.cnf`, or if it differs from them in some way.
2. Why this change is being made here.
3. Whether it's related in some way to fixing the test result mismatches caused by MTR connecting with TLS. (I think the answer is **no**, but I'm definitely not 100% certain.)

Given all that, shouldn't these changes be split out into a separate PR, or a separate commit, with a separate description for what's being changed and why?

vuvova (changes requested):
please remove all unrelated changes from this PR, as @dlenski said.
Only keep strictly bug fixing changes here

mysql-test/suite/binlog_encryption/my.cnf
grooverdan: @vuvova picked this up in b0fc90ac8a64486ede34999c7e3a2dea9bc5fc46

PR #2663, opened 2023-06-06
MDEV-13915 (10.5): STOP SLAVE takes very long time on a busy system

bnestere:
The problem is that a parallel replica would not immediately stop running/queued transactions when issued STOP SLAVE. That is, it allowed the current group of transactions to run, and sometimes the transactions which belong to the next group could be started and run through commit after STOP SLAVE was issued too, if the last group had started committing. This would lead to long periods to wait for all waiting transactions to finish.

This patch updates a parallel replica to try and abort immediately and roll-back any ongoing transactions. The exception to this is any transactions which are non-transactional (e.g. those modifying sequences or non-transactional tables), and any prior transactions, will be run to completion.

The specifics are as follows:

 1. A new stage was added to SHOW PROCESSLIST output for the SQL Thread when it is waiting for a replica thread to either rollback or finish its transaction before stopping. This stage presents as “Waiting for worker thread to stop”

 2. Worker threads which error or are killed no longer perform GCO cleanup if there is a concurrently running prior transaction. This is because a worker thread scheduled to run in a future GCO could be killed and incorrectly perform cleanup of the active GCO.

 3. Refined cases when the FL_TRANSACTIONAL flag is added to GTID binlog events to disallow adding it to transactions which modify both transactional and non-transactional engines when the binlogging configuration allow the modifications to exist in the same event, i.e. when using binlog_direct_non_trans_update == 0 and binlog_format == statement.

 4. A few existing MTR tests relied on the completion of certain transactions after issuing STOP SLAVE, and were re-recorded (potentially with added synchronizations) under the new rollback behavior.

## Description
TODO: fill description here

## How can this PR be tested?

## Backward compatibility

## PR quality check


knielsen:
Please see commit 7470535f02e63d8f1ab52449fdc53132498b3a5e that I pushed to branch knielsen_faster_stop_slave.

This fixes the main problem, which is that the current STOP SLAVE method waits for too long before starting to skip events. With this simple fix, each worker thread will finish the currently running event group, and then the slave will stop. I think this will solve the real problem that the original user was having.

I think this can be extended to roll back the currently running event groups at stop slave, though I think that's not really desirable as it can cause massive and unnecessary rollback if many threads are waiting to commit. I can also see if I can come up with a better fix for MDEV-31448.

Some parts of this patch are good, like the stage_waiting_for_worker_stop and the and moving the initialization of rgi->gtid_ev_flags2 to event_group_new_gtid().

sql/log_event_server.cc
```diff
@@ -3287,7 +3287,10 @@ Gtid_log_event::Gtid_log_event(THD *thd_arg, uint64 seq_no_arg,
       thd_arg->transaction->all.has_created_dropped_temp_table() ||
       thd_arg->transaction->all.trans_executed_admin_cmd())
     flags2|= FL_DDL;
-  else if (is_transactional && !is_tmp_table)
+  else if (is_transactional && !is_tmp_table &&
+           !(thd_arg->transaction->all.modified_non_trans_table &&
+             thd->variables.binlog_direct_non_trans_update == 0 &&
+             !thd->is_current_stmt_binlog_format_row()))
     flags2|= FL_TRANSACTIONAL;
```
knielsen, line 3294: This change seems good.

sql/log_event_server.cc
```diff
@@ -3450,7 +3453,6 @@ Gtid_log_event::do_apply_event(rpl_group_info *rgi)
   thd->variables.server_id= this->server_id;
   thd->variables.gtid_domain_id= this->domain_id;
   thd->variables.gtid_seq_no= this->seq_no;
-  rgi->gtid_ev_flags2= flags2;
```
knielsen, line 3453: This change also seems good (moving the init into event_group_new_gtid()).

sql/log_event_server.cc
```diff
@@ -5708,8 +5710,26 @@ int Rows_log_event::do_apply_event(rpl_group_info *rgi)
...
+      /*
+        Exit early, and let the Event-level exit logic take care of the cleanup
+        and rollback.
+      */
+      if (rgi->rli->mi->using_parallel() &&
+          rgi->parallel_entry->stop_abrupt(rgi->rli) &&
+          rgi->parallel_entry->rgi_is_safe_to_terminate(rgi))
+        break;
+
       error= do_exec_row(rgi);
```
knielsen, line 5722: Why only in parallel replication? Changing STOP SLAVE to abort in the middle of a row event - if we really want that - would make sense for non-parallel as well as parallel?

sql/rpl_parallel.cc
```diff
@@ -395,13 +395,14 @@ do_gco_wait(rpl_group_info *rgi, group_commit_orderer *gco,
     } while (wait_count > entry->count_committing_event_groups);
   }
 
-  if (entry->force_abort && wait_count > entry->stop_count)
```
knielsen, line 398: Why this change? It seems wrong?

knielsen (changes requested):
1. The main problem with slow STOP SLAVE is that the worker threads don't attempt to stop until the end of the GCO. This is not addressed this patch. I implemented a simple fix for this here: https://github.com/MariaDB/server/commit/7470535f02e63d8f1ab52449fdc53132498b3a5e

With this fix, I suspect there will be no need to roll back active transactions at STOP SLAVE. And doing so is not desirable, I think. Many of those transactions will in general be already done, just waiting to commit, and forcing immediate stop will cause massive rollback. At least this is a big functional change and should not be done in a GA/stable release.

2. The fix included in this patch for MDEV-31488 is not correct. I committed a better fix for this here: https://github.com/MariaDB/server/commit/6ce9c839997c1fc78c2989540dd21155a96fb419.

3. Should I push the above two fixes? If so, in which branch? And maybe you want to adapt the new test cases from this patch to the above two alternative fixes?

4. The stage_waiting_for_worker_stop seems good. It's not appropriate for a GA release though (as users may have monitoring scripts that expect the existing behaviour).

5. The fix to not set FL_TRANSACTIONAL on binlog_direct_non_trans_update==0 seems good.

6. Why change forced_abort to be an atomic? Most accesses are left as default, which means that they will use costly std::memory_order_seq_cst in, which seems wrong?

7. unsafe_rollback_marker_sub_id, what is the scenario where this is useful? Checking it doesn't seem very useful, as the value could change immediately after. And besides, non-FL_TRANSACTIONAL event groups will not start until prior event groups have committed (or started to commit), so it should not be necessary either?

8. If we really want to roll back active transactions, maybe an optional STOP SLAVE FORCE or some other appropriate syntax could be used? I would expect this could be done using the existing stop mechanism, where an error is signalled for the aborted event group. And some code to silence the error messages that would normally end up in the error log.

sql/sql_class.cc
```diff
@@ -8001,9 +8035,25 @@ wait_for_commit::wait_for_prior_commit2(THD *thd)
...
+    If this worker has been killed prior to this wait, e.g. in do_gco_wait(),
+    then it should not perform thread cleanup if there are threads which
+    have yet to commit. This is to prevent the cleanup of resources that
+    the prior RGI may need, e.g. its GCO. This is achieved by skipping
+    the unregistration of the waitee, such that each subsequent call to
+    wait_for_prior_commit() will exit early (while maintaining the
+    dependence), thus allowing the final call to
+    thd->wait_for_prior_commit() within finish_event_group() to wait.
+  */
+  if (!loc_waitee
```
andrelkin, line 8051: @knielsen, for your attention, the block intends to cover a `SPECULATE_WAIT`:ers, not what MDEV-31448 description means - I gather my mistake.
A  killed `SPECULATE_WAIT`:er is supposed to exit from here without unlinking from the (gc) `order` parent. In the following path down to `finish_event_group` the link to `waitee` must be preserved.
Your [fixes](https://github.com/MariaDB/server/commit/6ce9c839997c1fc78c2989540dd21155a96fb419#diff-1e7bbf2eedc9ef2e8c81bbbf3dcac4287fa55b8bfb6250daae0578814e506a67) of MDEV-31448 do help to preserve the link which I thought to implement with following the pattern of l.8036 (BASE) that is to set `worker_errror` which avoids `unmark()`.

knielsen: @andrelkin , sorry I did not understand what you mean?

I checked my MDEV-31448 fix against the testcase from Jira, the testcase passes with my fix (and without this change to in wait_for_prior_commit2()).

Is there some other problem that is fixed by these changes, and is there a testcase that I can try? Or did you mean that my fix to MDEV-31448 does solve the problem and is a valid alternative to these changes in wait_for_prior_commit2() ?

The wait_for_prior_commit is a very basic mechanism with a precise semantics. I really don't want it to depend on replication internals like thd->rgi_slave->did_mark_start_commit. So if there's anything else needed than my two commits, I want to know what the problem is so I can come up with a good solution.

I don't understand from the comment in Brandon's original patch or the discussions what these changes in wait_for_prior_commit2() are meant to do, or why they are necessary.

andrelkin: MDEV-31448 description test I am still reviewing. What I meant is in just attached test which is about the optimistic mode SPECULATE_WAIT worker that get killed in that waiting. The test asserts on your branch too.

I agree the introduction of ``did_mark_start_commit`  is worth of discussion.
`did_mark_start_commit` helps to recognize the worker's wait is not about to commit but rather about `SPECULATE_WAIT` condition to start. The latter is killable unlike the former when the worker's state fits to your [comments](https://github.com/MariaDB/server/blob/bb-10.5-MDEV-13915/sql/sql_class.cc#L8064)
``` 
 But if a wakeup is already in progress, then we must ignore the kill and not give error 
```
I realize now that the MDEV-31448 fixes are still incomplete to leave out a subclass of the killed which is neither `!thd->rgi_slave->did_mark_start_commit` nor `loc_waitee->wakeup_subsequent_commits_running`. So a waiting-for-prior-commit worker that gets killed while the waitee has not started its commit phase still [unlinks](https://github.com/MariaDB/server/blob/bb-10.5-MDEV-13915/sql/sql_class.cc#L8084) from the commit parent. 
Now I am thinking about common coverage of this and  the `did_mark_start_commit` case. It looks it's just enough 
not to do [`remove_from_list()` etc] (https://github.com/MariaDB/server/blob/bb-10.5-MDEV-13915/sql/sql_class.cc#L8082) here, to defer that either to `retry_event_group`, or to `finish_event_group` calling `wait_for_prior_commit2(... force_wait == true)`.

The `wait_for_prior_commit2()` comments indeed need updating. Let me do that short in a separate commit.

andrelkin: To follow up
> I realize now that the MDEV-31448 fixes are still incomplete

indeed my SPECULATE_WAIT version of its test asserts in 13915 branch.
> It looks it's just enough not to do remove_from_list() 

And that must be a correct approach. The link with `waitee` remains intact, to play its order-safety role in `finish_event_group` 's `wait_for_prior_commit(... force_wait == true)`.

I am about pushing a commit to that with an elaborated test covering both the SPECULATIVE and commit-time waits.

knielsen: Hm, valid point that do_gco_wait() also continues to wait after being killed (and also after signalling the error).

And I like much better the approach that wait_for_prior_commit doesn't have logic that refers to rgi etc., just skips the unlinking if killed. It's at least a clear semantics, wait_for commit will unlink only once the prior commit either commits or signals that it aborts, not if interrupted by kill. It's a pity that the return doesn't easily distinguish between prior transaction getting error or this transaction getting kill, but we probably don't really need that (or we could add something if we need it).

sql/sql_class.cc
```diff
@@ -6465,6 +6469,36 @@ int THD::decide_logging_format(TABLE_LIST *tables)
...
+          }
+          else
+          {
+            // todo: issue a warning/errror
+            mysql_mutex_unlock(&e->LOCK_parallel_entry);
+            DBUG_RETURN(-1);
+          }
+          mysql_mutex_unlock(&e->LOCK_parallel_entry);
+        }
+#endif
```
knielsen, line 6500: 1. This code surely does not belong here? decide_logging_format() has the task to select between logging in statement or row. This logic seems to be about tracking non-transactional changes for faster stop of parallel replication, completely unrelated. And what if binlogging isn't enabled on the slave, is this code then not called?
2. Parallel replication does not support different engine on master and slave (I think the purpose of this code is to handle a table that is eg. InnoDB on the master but altered to MyISAM on the slave). This situation will cause failures in many different ways, not just for fast STOP SLAVE. And even this case seems not correctly handled by this code; an earlier event group could check unsafe_rollback_marker_sub_id and decide to roll back just before this code runs and sets the new value of unsafe_rollback_marker_sub_id ? It is a valid task to make parallel replication safe for replicating between different engines on master and slave, but it does not make sense to implement a partial solution.

sql/slave.cc
```diff
@@ -3692,6 +3694,11 @@ has_temporary_error(THD *thd)
   if (!likely(thd->is_error()))
     DBUG_RETURN(0);
 
+  if (rgi->rli->mi->using_parallel() &&
+      rgi->parallel_entry->stop_abrupt(rli) &&
+      rgi->parallel_entry->rgi_is_safe_to_terminate(rgi))
+    DBUG_RETURN(0);
```
knielsen, line 3700: This logic doesn't belong in has_temporary_error(), it is about whether to stop the slave, not about classification of an error.

sql/sql_class.cc
andrelkin: Completely agree with p.2. (I was against doing that for the very same reason; however we pushed it to satisfy @montywi 's review notes on the corporate slack).

To p.1, indeed the changes  in this function are ineffectual in `log_bin = ON` mode. That's a good point of yours.

knielsen:
I now implemented the STOP SLAVE FORCE idea I mentioned, it can be found on my branch knielsen_faster_stop_slave:

    https://github.com/MariaDB/server/commits/knielsen_faster_stop_slave
    https://github.com/MariaDB/server/commit/c20ce0dff404890df59f1fd305bd85be0ada86f8

I believe with this patch, my branch should now have all the functionality of this pull request. comments?

bnestere:
Closing PR, as a simplified version was pushed as 60bec1d54d320e86f2c09407e1981c901a675c1c

PR #2667, opened 2023-06-12
MDEV-31242: Make sure every Debian post/pre script is using bash

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-31242*

## Description
Make sure that every Debian post and prescript are using bash and have set -e in next to shebang. Commands set -e makes sure that if script fails in one command it will fail whole script.

## How can this PR be tested?
As Salsa-CI on 11.2 in currently broken and this has to be tested by hand:

```
debian/autobake-debs.sh
dpkg -i ../*.deb
```
If everything goes well it should be ok as most of the people had sh either zsh or bash. Now we just want to use Bash as it makes testing easier.

## PR quality check

### DIFF
--- a/debian/mariadb-common.postinst
+++ b/debian/mariadb-common.postinst
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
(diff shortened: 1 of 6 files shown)


ericherman (approved):
Thank you, and nice to standardize on `set -e` rather than a mix of that and `#!/bin/bash -e` for no obvious reason.

PR #2675, opened 2023-06-22
MDEV-20286 mariabackup fails when innodb_max_dirty_pages_pct contains…

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-20286*

## Description

This is a port of the Percona Server commit 5265f42e290573e9591f8ca28ab66afc051f89a3

which is the same as their bug PXB-1807: xtrabackup does not accept fractional values for innodb_max_dirty_pages_pct

Problem:

Variable specified as double in MySQL server, but read as long in the xtrabackup. This causes xtrabackup to fail at startup when the value contains decimal point.

Fix:

Make xtrabackup to interpret the value as double to be compatible with server.

## How can this PR be tested?

tested with:
```
$ mysql-test/mtr --suite=mariabackup --mysqld=--innodb_max_dirty_pages_pct=0.5
```

## PR quality check


extra/mariabackup/xtrabackup.cc
```diff
@@ -1638,8 +1638,11 @@ struct my_option xb_server_options[] =
    "Path to InnoDB log files.", &srv_log_group_home_dir,
    &srv_log_group_home_dir, 0, GET_STR, REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
   {"innodb_max_dirty_pages_pct", OPT_INNODB_MAX_DIRTY_PAGES_PCT,
-   "Percentage of dirty pages allowed in bufferpool.", (G_PTR*) &srv_max_buf_pool_modified_pct,
-   (G_PTR*) &srv_max_buf_pool_modified_pct, 0, GET_ULONG, REQUIRED_ARG, 90, 0, 100, 0, 0, 0},
+   "Percentage of dirty pages allowed in bufferpool.",
+   (G_PTR*) &srv_max_buf_pool_modified_pct,
+   (G_PTR*) &srv_max_buf_pool_modified_pct, 0, GET_DOUBLE, REQUIRED_ARG,
+   (longlong)getopt_double2ulonglong(75), (longlong)getopt_double2ulonglong(0),
```
LinuxJedi, line 1644: This appears to set the default to 75% instead of 90%. Is that intentional?

grooverdan: oops. Nope. corrected.

LinuxJedi (approved):
Much better :)

PR #2680, opened 2023-06-28
MDEV-30182: Optimize open_tables to take O(N) time

Vanislavsky:
- [x] *The Jira issue number for this PR is: MDEV-30182*

## Description
find_fk_prelocked_table traverses all opened tables in connection, MDL_context::find_ticket traverses all mdl tickets which is also not less amount of steps. Both functions are called for each table, taking O(tables^2) time total, during opening tables.

## How can this PR be tested?

## PR quality check


sql/sql_base.cc
```diff
@@ -4604,23 +4653,83 @@ bool DML_prelocking_strategy::handle_routine(THD *thd,
   return FALSE;
 }
 
+#include <string>
+#include <unordered_map>
+#include <string_view>
+#include <cstring>
+#include <functional>
```
LinuxJedi, line 4660: These should probably be at the top of the file.

LinuxJedi (dismissed):
Thank you for your contribution.

I've not looked at the implementation at this stage, just on things such as coding style. I've commented on a few things, but it will likely need a few more minor cleanups.

The CI did not run properly because there are conflicts you will need to resolve first.

Also, can you please either sign the CLA or declare that you are contributing under the 3-Clause BSD license?

Vanislavsky:
I am contributing my new code of the whole pull request, including one or several files that are either new files or modified ones, under the BSD-new license

FooBarrior:
@LinuxJedi hi! @Vanislavsky was doing this work under my mentorship. This PR is a little-bit WIP yet, we've decided to publish it ASAP, since the diploma thesis presntation is soon.

Let me do the kick-off review with all the style checks.

There's also a performance benchmark that will be presented a few days later, showing the benefit.

LinuxJedi:
> @LinuxJedi hi! @Vanislavsky was doing this work under my mentorship. This PR is a little-bit WIP yet, we've decided to publish it ASAP, since the diploma thesis presntation is soon.
> 
> Let me do the kick-off review with all the style checks.
> 
> There's also a performance benchmark that will be presented a few days later, showing the benefit.

Ah! Thanks for letting me know, assigned it over to you :)

FooBarrior:
@Vanislavsky Please also add a description of what was done, like:
* A broad description of a hash table you're adding. What features it has.
* What components were changed and why.
* Why we don't remove an array of ticket lists

unittest/sql/open_address_hash-t.cc
```diff
@@ -64,8 +69,14 @@ int main(int argc __attribute__((unused)),char *argv[])
   hashie.insert(data[0]+5);
   ok(hashie.size() == 3, "wrong size, %u", hashie.size());
   // collision
+  hashie.insert(data[1] + 1); 
+  found= hashie.find(data[1] + 1);
```
FooBarrior, line 73: no, `found` was supposed to be the old one. How does it work now, if it inserts the same element twice?

sql/mdl.cc
```diff
@@ -2498,7 +2498,7 @@ MDL_context::acquire_lock(MDL_request *mdl_request, double lock_wait_timeout)
   DBUG_ASSERT(wait_status == MDL_wait::GRANTED);
 
   m_tickets[mdl_request->duration].push_front(ticket);
-  ticket_hash.insert(ticket); // TODO error handling
+  ticket_hash.insert(ticket);
```
FooBarrior, line 2501: why did you remove a TODO? Or doesn't `insert` need an error handling anymore?

sql/sql_class.h
```diff
@@ -2621,6 +2621,8 @@ class table_key
   }
 };
 
+
+
```
FooBarrior, line 2625: junk change

sql/sql_class.h
```diff
@@ -2735,7 +2737,7 @@ class THD: public THD_count, /* this must be first */
   */
   mutable mysql_mutex_t LOCK_thd_kill;
 
-  open_address_hash<table_key, hash_trait<TABLE_LIST, TABLE_LIST*>> pr_table_hash;
+  open_address_hash<MDL_key_trait<TABLE_LIST>, hash_trait<TABLE_LIST, TABLE_LIST *>> pr_table_hash;
```
FooBarrior, line 2740: better make a space after the first closing `>` bracket -- not all C++ parsers are good at distinguishing it with a shift operator, and though it's now covered by 11 standard, still can be a good gesture

FooBarrior (changes requested):
forgot to press the "request changes" button

sql/open_address_hash.h
```diff
@@ -0,0 +1,300 @@
...
+        second= nullptr;
+        return true;
+      }
+      else
+      {
+        return false;
+      }
+    }
+
+    if (capacity > 7 && static_cast<double>(size - 1) <
```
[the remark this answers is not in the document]
FooBarrior: okay, as we discussed, capacity really can be < 8. I thought we agreed on 8+, but it's CAPASITY_INITIAL is 4 now.
Let's leave it so, but if we'll change it to 8 or more, then there'll be no need to evaluate `capacity > 7` in runtime.

Let's add the hint for a compiler to optimize that out in such case:
```
if ((CAPACITY_INITIAL >= 8 || capacity >= 8) ...
```
This humanly folds into `if ((MAX(CAPACITY_INITIAL, capacity) >= 8...`, but it's risky that a compiler won't recognize the optimization, so the first one is better imo.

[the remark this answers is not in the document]
FooBarrior: okay, most of the performance issue is fixed, and the hint can be added by us, once we'll decide to raise the initial capacity

sql/open_address_hash.h
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  void rehash(const uint _capacity)
+  {
+    uint past_capacity= capacity;
+    capacity= _capacity;
+
+    if (past_capacity > capacity)
+    {
+      _size= 0;
```
FooBarrior, line 108: should we tear this _size value back ad forth during rehash?
It seems that the only place it should be increased is `insert`!

Same for the decrease -- just put is in `remove` and clean these torturings elsewere.

sql/sql_class.h
```diff
@@ -2586,6 +2586,40 @@ struct thd_async_state
   }
 };
 
+#include "open_address_hash.h"
```
FooBarrior, line 2589: this #include should be on the top of the file

sql/mdl.h
```diff
@@ -861,6 +861,94 @@ typedef I_P_List<MDL_request, I_P_List_adapter<MDL_request,
...
+  key_type_pair(MDL_key *_mdl_key, enum_mdl_type _type)
+  {
+    mdl_key= _mdl_key;
+    type= _type;
+  }
+  MDL_key *mdl_key;
+  enum_mdl_type type;
+};
+
+#include "open_address_hash.h"
```
FooBarrior, line 876: same

sql/mdl.h
```diff
@@ -861,6 +861,94 @@ typedef I_P_List<MDL_request, I_P_List_adapter<MDL_request,
                  I_P_List_counter>
         MDL_request_list;
 
+class key_type_pair
```
FooBarrior, line 864: I suppose that we don't need it anymore

FooBarrior (changes requested):

sql/mdl.h
```diff
@@ -934,18 +887,13 @@ template <typename T, typename K> class hash_trait
...
-           lhs->has_stronger_or_equal_type(rhs.type);
-  }
 
   static bool is_empty(const elem_type &el) { return el == nullptr; }
   static void set_null(elem_type &el) { el= nullptr; }
 
   static my_hash_value_type get_hash_value(T *t)
   {
-    return t->get_key()->tc_hash_value();
+    return MDL_key_trait<T>::get_hash_value_from_key(t->get_key());
```
FooBarrior, line 896: This way you're making a class named `hash_trait` bound to MDL_key_trait

sql/sql_base.cc
```diff
@@ -1001,9 +1001,6 @@ void close_thread_table(THD *thd, TABLE **table_ptr)
     table->file->ha_reset();
   }
 
-  /*
```
FooBarrior, line 1004: will we have a new comment?

FooBarrior (changes requested):

sql/open_address_hash.h
```diff
@@ -205,9 +242,15 @@ template <typename key_trait, typename trait> class open_address_hash
     }
 
     if (unlikely(capacity > 7 && (_size - 1) * LOW_LOAD_FACTOR < capacity))
-      rehash(0.5 * capacity);
+      shrink(0.5 * capacity);
```
FooBarrior, line 245: if it's `grow(capacity << 1);`, then maybe let's make it symmetric with capacity >> 1?

sql/open_address_hash.h
```diff
@@ -0,0 +1,316 @@
...
+
+    if (past_capacity > capacity)
+    {
+      _size= 0;
+      for (uint i= capacity; i < past_capacity; i++)
+      {
+        if (hash_array[i])
+        {
+          auto temp_el= hash_array[i];
+          hash_array[i]= nullptr;
```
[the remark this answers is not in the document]
Vanislavsky: I fixed it. Should work correctly

sql/mdl.h
Vanislavsky: It really is. Are you not happy with this way, or are you just suggesting renaming the class?

sql/sql_base.cc
Vanislavsky: There seems to be no need for a comment

sql/open_address_hash.h
Vanislavsky: it's done

sql/mdl.h
FooBarrior: Suppose we'll have another class, that would be stored in our hash, it can be a table of plugins for example. 
The developer would study the usage of our hash, make their own key trait (since plugins don't use MDL keys), and try to reuse hash_trait, but they would fail because of this dependency in `get_hash_value`.

FooBarrior: my suggestion is to leave `get_key()` here and call `get_hash_value_from_key` (also please rename it to just `get_hash_value`) from the hash:
```cpp
template <class key_trait> // since we don't know the exact return type and we don't have `auto` here
static const key_trait::key_type& get_key(const T *value)
{
  return value->get_key();
}
```

Now in the hash code:
```cpp
  my_hash_value_type get_hash_value(const T *value)
  { return key_trait::get_hash_value(value_trait::get_key(value)); }
```

Note that `key_trait::get_hash_value` should accept const key_type& in the way i wrote it here. 
You can also rephrase it in pointers, if you like. But `const` is mandatory.

sql/mdl.h
```diff
@@ -1102,7 +1102,7 @@ class MDL_context
   /**
     Ticket hash. Stores only locked tickets.
   */
-  open_address_hash<MDL_key_trait<MDL_ticket>, hash_trait<MDL_ticket> > ticket_hash;
+  open_address_hash<MDL_key_trait<MDL_ticket>, hash_trait<MDL_ticket, MDL_key_trait<MDL_ticket> > > ticket_hash;
```
FooBarrior, line 1105: you see, this way it makes the declaration harder, which better should be avoided. Actually, the only thing we're concerned is the return type of `get_key` the rest can be cared out inside the hash code.

unittest/sql/open_address_hash-t.cc
```diff
@@ -32,7 +32,7 @@ struct pointer_trait
 
 struct test_value_trait: public pointer_trait<uint32>
 {
-  static const uint32 &get_key(uint32 *elem) { return *elem; }
+  static const uint32* get_key(uint32 *elem) { return elem; }
```
FooBarrior, line 35: 1. put `*` sticked to `get_key`: `uint32 *get_key`
2. add `const` to the argument.

Vanislavsky: adding const to an argument causes a compilation error. unable to convert TABLE_LIST

FooBarrior: there's no TABLE_LIST here, it's a unit test

sql/sql_base.cc
FooBarrior: I mean [the one for the declaration](https://github.com/MariaDB/server/pull/2680#discussion_r1276456880)

Vanislavsky: i added this

FooBarrior: thanks, missed it.

sql/open_address_hash.h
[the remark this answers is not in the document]
FooBarrior: Do you mean `realloc`? That's the different issue, I haven't addresed it yet.

Here I mean that in the end `hash_array[i]` gets the same value it had before `nullptr` assigned in 50% of cases.
Maybe we can first insert to the right position, and then, _if this position differs_, zero the old cell?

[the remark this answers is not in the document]
Vanislavsky: I made it so that zeroing and rehashing only happens if i is different from the key
`if (hash_array[i] && i != to_index(hash_from_value(hash_array[i])))`

sql/open_address_hash.h
```diff
@@ -0,0 +1,315 @@
...
+  };
+
+  uint rehash_subsequence(uint i)
+  {
+    for (uint j= to_index(i + 1); hash_array[j] != nullptr; j= to_index(j + 1))
+    {
+      auto temp_el= hash_array[j];
+      hash_array[j]= nullptr;
+      //_size--;
+      insert_helper(temp_el);
```
FooBarrior: Sorry, initially I wanted to add the comment here, but sent it to the wrong `nullptr+insert_helper` pair:) But it als applies there.

Now here, concider we have this collision in memory:
```
|5|6|7|8|9|
|0|A|B|C|0|
```
And we're removing A, i.e. `hash_array[6]`. Now what happens in memory
```
1. hash_array[6] gets 0
2. hash_array[6] gets B
3. hash_array[7] gets 0
4. hash_array[7] gets C
5. hash_array[8] gets 0
```
Steps 1 and 3 are obsolete. If you'll find a way to avoid it, you'll make twice fewer stores during the collision rehashing!

sql/open_address_hash.h
```diff
@@ -0,0 +1,334 @@
...
+  {
+    uint past_capacity= capacity;
+    capacity= _capacity;
+
+    for (uint i= capacity; i < past_capacity; i++)
+    {
+      if (hash_array[i])
+      {
+        auto temp_el= hash_array[i];
+        hash_array[i]= nullptr;
```
FooBarrior, line 130: you don't need to zero it at all

sql/open_address_hash.h
```diff
@@ -0,0 +1,334 @@
...
+      }
+    }
+  }
+
+  void shrink(const uint _capacity)
+  {
+    uint past_capacity= capacity;
+    capacity= _capacity;
+
+    for (uint i= capacity; i < past_capacity; i++)
```
FooBarrior, line 125: good. indeed we don't need to run it for `i < capacity` because of the number theory:)
*If* `i = k (mod 2^n) such that k < 2^{n-1}`
*Then* `i = k (mod 2^{n-1})`

and actually, if it's vice-versa, i.e. `k >= 2^{n-1}`,
then `i = k - 2^{n-1}  (mod 2^{n-1})`

FooBarrior (changes requested):

sql/open_address_hash.h
Vanislavsky: I understand what you are saying, but I have no idea how to implement it

sql/mdl.h
```diff
@@ -1102,7 +1102,7 @@ class MDL_context
   /**
     Ticket hash. Stores only locked tickets.
   */
-  open_address_hash<MDL_key_trait<MDL_ticket>, hash_trait<MDL_ticket, MDL_key_trait<MDL_ticket> > > ticket_hash;
+  open_address_hash<MDL_key_trait<MDL_ticket>, hash_trait<MDL_ticket,MDL_key> > ticket_hash;
```
FooBarrior, line 1105: This only makes it trickier to understand. How do you imagine a developer who wants just an `std::set<Key, Value>` analogue use this?

sql/open_address_hash.h
```diff
@@ -0,0 +1,333 @@
...
+    return false;
+  }
+
+  void grow(const uint _capacity)
+  {
+    uint past_capacity= capacity;
+    capacity= _capacity;
+    hash_array= (T *) realloc(hash_array, capacity * sizeof(T));
+
+    for (uint i= past_capacity; i < capacity; i++)
```
FooBarrior, line 104: just bzero it

sql/open_address_hash.h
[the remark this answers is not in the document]
FooBarrior: Oh, i see, you are probably trying to avoid the collision-wraparound problem this way... Okay, it's ineffective, but if it's correct, we probably can move on with it in scope of a pull request. But at least i have to think on its correctness for a few days, so I'll respond later.

PR #2681, opened 2023-06-28
[MDEV-15935] Connection redirection mechanism

dlenski:
- [x] *The Jira issue number for this PR is: MDEV-15935*

## Description

This PR proposes a new feature in [the MariaDB client-server protocol](https://mariadb.com/kb/en/clientserver-protocol): application-layer redirection of client connections.

It also provides an initial server-side implementation of that feature; the corresponding client-side implementation is in https://github.com/mariadb-corporation/mariadb-connector-c/pull/226.

We want the MariaDB server to be able to tell clients connecting to it, “Sorry, this server is unavailable. Connect to an alternate server instead.” This mechanism is inspired by HTTP 302 (“temporary redirect”) mechanism familiar to all developers of web applications, and is intended to have similar semantics and security properties, since these have now been widely deployed and tested for decades.

### Automated testing

- [ ] TODO: write new MTR tests to demonstrate and validate the redirection functionality
- [x] Pre-existing MTR tests are updated and passing.

### Manual testing

- Build the server from this PR, along with the updated client from https://github.com/mariadb-corporation/mariadb-connector-c/pull/226 (should be included as a submodule).
- Start the server with `sql/mariadbd --server-redirect-mode=ON --server-redirect-target='some-other-mariadb.server.com:3306' --port=3306 --extra-port=3307 --socket=/tmp/mysql.sock`
- Connect with the updated client and observe the redirection behavior, including the fact that connections to the [`EXTRA_PORT`](https://mariadb.com/kb/en/thread-pool-in-mariadb/#configuring-the-extra-port) and to the Unix socket  are _not_ redirected:
  - `client/mariadb --host 127.0.0.1 --port=3306 -uUSERNAME -pPASSWORD` → redirected
  - `client/mariadb --host 127.0.0.1 --port=3307 -uUSERNAME -pPASSWORD` → _not_ redirected (`EXTRA PORT`)
  - `client/mariadb --socket=/tmp/mysql.sock -uUSERNAME -pPASSWORD` → _not_ redirected (local/non-network-based connection)

- [x] *This is a new feature and the PR is based against the latest MariaDB development branch, **11.2**.*

## PR quality check

### DIFF
--- a/mysql-test/main/ssl_7937,nossl.result
+++ b/mysql-test/main/ssl_7937,nossl.result
@@ -3,15 +3,13 @@ select if(variable_value > '','yes','no') as 'have_ssl'
   from information_schema.session_status
 where variable_name='ssl_cipher';
 mysql --ssl-ca=cacert.pem -e "call test.have_ssl()"
-have_ssl
-no
+ERROR 2026 (HY000): Client requires TLS/SSL, but the server does not support it
 mysql --ssl -e "call test.have_ssl()"
-have_ssl
-no
+ERROR 2026 (HY000): Client requires TLS/SSL, but the server does not support it
 mysql --ssl-ca=cacert.pem --ssl-verify-server-cert -e "call test.have_ssl()"
-ERROR 2026 (HY000): TLS/SSL error: SSL is required, but the server does not support it
+ERROR 2026 (HY000): Client requires TLS/SSL, but the server does not support it
 mysql --ssl --ssl-verify-server-cert -e "call test.have_ssl()"
-ERROR 2026 (HY000): TLS/SSL error: SSL is required, but the server does not support it
+ERROR 2026 (HY000): Client requires TLS/SSL, but the server does not support it
 #
 # MDEV-27105 --ssl option set as default for mariadb CLI
 #
--- a/sql/share/errmsg-utf8.txt
+++ b/sql/share/errmsg-utf8.txt
@@ -10786,3 +10786,7 @@ ER_JSON_SCHEMA_KEYWORD_UNSUPPORTED
         eng "%s keyword is not supported"
 ER_EVENTS_NO_ACL
         eng "Event scheduler cannot function with --skip-grant-tables, --bootstrap, or embedded build"
+ER_SERVER_REDIRECT
+        eng "|Server is redirecting clients to '%1$s'|%1$s"
+        fra "|Ce serveur rédirige ses clients vers '%1$s'|%1$s"
+        spa "|Este servidor redirige sus clientes hacia '%1$s'|%1$s"
--- a/sql/sql_acl.cc
+++ b/sql/sql_acl.cc
@@ -13743,6 +13743,84 @@ static ulong parse_client_handshake_packet(MPVIO_EXT *mpvio,
   DBUG_PRINT("info", ("parse_client_handshake STARTING: vio_type=%s",
                       safe_vio_type_name(thd->net.vio)));
 
+  enum enum_vio_type type= vio_type(thd->net.vio);
+  bool local_connection= (type == VIO_TYPE_SOCKET) || (type == VIO_TYPE_NAMEDPIPE);
+  uint16_t server_port = 0;
+  bool is_extra_port= FALSE;
+
+  if (!local_connection) {
+    /* We want to determine if the client is connecting to the TCP
+     * port optionally configured via the system variable
+     * EXTRA_PORT. This is intended for emergency/administrative
+     * use. (See
+     * https://mariadb.com/kb/en/thread-pool-in-mariadb/#configuring-the-extra-port)
+     *
+     * Based on its name and usage elsewhere in the server codebase,
+     * I expected that the net->vio->mysql_socket data structure
+     * would be populated at this point, including the member
+     * .is_extra_port whose name suggests it is precisely intended
+     * to convey this information.
+     *
+     * However, debugging with GDB demonstrates that
+     * net->vio->mysql_socket is NOT reliably populated, and in
+     * particular .is_extra_port appears to have the value 127
+     * regardless of whether client has connected via the extra
+     * port, or not.
+     *
+     * This means that we need to compare the local socket
+     * address to mysqld_extra_port (which holds the value of the
+     * EXTRA_PORT system vairable).
+     *
+     * That leads to another data structure which is inexplicably
+     * unpopulated at this point: net->vio->local. Since we cannot
+     * rely on that data structure either, we have to populate it
+     * ourselves here.
+     *
+     * FIXME: If net->vio->mysql_socket were sanely and reliably
+     * populated here, this entire 'if'-block could be replaced
+     * with:
+     *
+     *   is_extra_port= net->vio->mysql_socket->is_extra_port.
+     */
+
+    union _sockaddr_ipv46 {
+      struct sockaddr_storage _nvl; /* This is the actual type of net->vio->local */
+      struct sockaddr_in inaddr;
+      struct sockaddr_in6 in6addr;
+      struct sockaddr addr;
+    } *a = (union _sockaddr_ipv46 *)&net->vio->local;
+
+    if (a->_nvl.ss_family == AF_UNSPEC)
+    {
+      /* FIXME: If we could rely on net->vio->local having already
+       * been populated, we could remove this entire 'if' block.
+       */
+      socklen_t addrlen = sizeof(a->_nvl);
+      memset(a, 0, addrlen);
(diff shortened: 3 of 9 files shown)


sql/sql_acl.cc
```diff
@@ -13736,10 +13738,91 @@ static ulong parse_client_handshake_packet(MPVIO_EXT *mpvio,
...
+     *
+     *   is_extra_port= net->vio->mysql_socket->is_extra_port.
+     */
+
+    union _sockaddr_ipv46 {
+      struct sockaddr_storage _nvl; /* This is the actual type of net->vio->local */
+      struct sockaddr_in inaddr;
+      struct sockaddr_in6 in6addr;
+      struct sockaddr addr;
+    } *a = (union _sockaddr_ipv46 *)&net->vio->local;
```
dlenski, line 13791: This `union` is needed to strictly comply with C/C++ [strict aliasing](https://en.cppreference.com/w/c/language/object#Strict_aliasing) rules.

Even though `sockaddr_in` and `sockaddr_in6` store the `_family` and `_port` fields at identical positions, the compiler flags used by MariaDB cause GCC and Clang to reject sloppier type-punning approaches here.

mysql-test/main/ssl_7937,nossl.result
```diff
@@ -3,15 +3,13 @@ select if(variable_value > '','yes','no') as 'have_ssl'
   from information_schema.session_status
 where variable_name='ssl_cipher';
 mysql --ssl-ca=cacert.pem -e "call test.have_ssl()"
-have_ssl
-no
+ERROR 2026 (HY000): Client requires TLS/SSL, but the server does not support it
 mysql --ssl -e "call test.have_ssl()"
-have_ssl
-no
+ERROR 2026 (HY000): Client requires TLS/SSL, but the server does not support it
```
dlenski, line 8: This is changes because this PR incorporates https://github.com/mariadb-corporation/mariadb-connector-c/pull/226 as a submodule, and that includes the fix for [MDEV-28634](https://jira.mariadb.org/browse/MDEV-28634).

Without that change, the command-line client can be **instructed to use SSL** and then _**silently downgraded**_ to an insecure/plaintext connection. :scream:

With that change, Connector/C based clients with TLS enabled (e.g. `mysql --ssl`) properly refuse to connect to servers which don't support TLS.

dlenski:
It appears that _at least **some**_ of the CI failures are caused by Buildbot _not using the updated_ Connector/C client…

… even though this PR includes it as a submodule change. :man_shrugging:

Example in: https://buildbot.mariadb.org/#/builders/576/builds/1093/steps/7/logs/stdio

mysql-test/main/ssl_7937,nossl.result
LinuxJedi: I haven't checked, but I suspect the submodule update doesn't work because it links to a commit ID that isn't in the main mariadb-connector-c repo.

dlenski: > I haven't checked, but I suspect the submodule update doesn't work because it links to a commit ID that isn't in the main mariadb-connector-c repo.

Hmmm… is there anything in the Buildbot log that indicates that this is an issue? I don't see anything, or maybe don't have access to the logs that would show it.

From _Github's_ view of things, everything appears correct:

1. The "Changes" tab of this PR shows the intended change in the `libmariadb` submodule:

    image

2. The target commit is indeed visible in the web-based interface for the main/upstream `mariadb-connector-c` repo (e.g. at `https://github.com/mariadb-corporation/mariadb-connector-c/commit/abad64b9cf46b0f9d6b85bf8a59f513856ea75fb`), with the usual caveat or warning message about how this commit comes from a fork:

    image

sql/sys_vars.cc
```diff
@@ -7173,3 +7173,23 @@ static Sys_var_optimizer_cost Sys_optimizer_scan_cost(
...
+        "OFF", "ON", "ALL", 0 };
+
+static Sys_var_enum Sys_server_redirect_mode(
+       "server_redirect_mode",
+       "Server redirection mode. "
+       "Possible modes are: OFF - No redirection, "
+       "ON: Redirect all new clients attempting to connect over the network to the standard server port (no redirection of local socket-based connections, nor of connections to the EXTRA_PORT), "
+       "ALL: Unconditionally redirect new clients attempting to connect (even via local socket-based connections).",
+       GLOBAL_VAR(server_redirect_mode), CMD_LINE(REQUIRED_ARG),
+       server_redirect_mode_names, DEFAULT(0));
```
dlenski, line 7195: Over at https://jira.mariadb.org/browse/MDEV-15935?focusedCommentId=263403#comment-263403, @markus456 suggested renaming the `ON` value to `FORCE`.

> It would cover use-cases other than forceful redirection and also make it possible to have a longer redirection period. As such, I'd name the value for `server_redirect_mode` as `FORCE` rather than `ON`: this way the door is left open for a less intrusive method of redirection

That's a good idea :+1:. It's clearer and leaves open more options for future "less-forceful" redirection modes.

sql/sys_vars.cc
```diff
@@ -7173,3 +7173,23 @@ static Sys_var_optimizer_cost Sys_optimizer_scan_cost(
   CMD_LINE(REQUIRED_ARG),
   VALID_RANGE(0, 100000000), DEFAULT(DEFAULT_TABLE_SCAN_SETUP_COST),
   COST_ADJUST(1000));
+
+static Sys_var_charptr_fscs Sys_server_redirect_target(
+	"server_redirect_target",
+        "Server redirection target. This should be a hostname, an IP address, or "
+        "a hostname or IP address followed by ':PORT'. Server redirection will "
+        "not be activated unless SERVER_REDIRECT_MODE is also set.",
```
dlenski, line 7181: On [the mailing list](https://lists.mariadb.org/hyperkitty/list/developers%40lists.mariadb.org/message/QZDOCLK7TRTSAZOJFNANX5VLBY4PKS45/), @vuvova wrote:

> I'd really rather avoid introducing a new connection string format.
> …
> MDEV-15935 says that we "might" optionally implement URI validation in
the server. Without it one can literally send an arbitrary string.
This might be error-prone. With the server side URI validation
we can limit redirection URIs to "mysql://host[:port]" which basically
is what you wanted and postpones all the URI complexity till we have
some standard specs for it.

I agree.

Allowing the server to send an arbitrary string here is error-prone, and introducing Yet Another Connection String Format seems unwise, even in the name of simplicity.

So what I should do is…

1. Change this help text to explain the format as `mysql://host[:port]`, where `host` can be either a hostname or a URL-format IP address (bracket-enclosed for IPv6, e.g. `[2000::dead:beef]`), and where `port` defaults to 3306 if unspecified.
2. Add a validation function to prevent the `SERVER_REDIRECT_TARGET` sysvar from being set to a value not matching this format.
3. Modify https://github.com/mariadb-corporation/mariadb-connector-c/pull/226 to expect this format.

sql/share/errmsg-utf8.txt
```diff
@@ -10786,3 +10786,7 @@ ER_JSON_SCHEMA_KEYWORD_UNSUPPORTED
         eng "%s keyword is not supported"
 ER_EVENTS_NO_ACL
         eng "Event scheduler cannot function with --skip-grant-tables, --bootstrap, or embedded build"
+ER_SERVER_REDIRECT
+        eng "|Server is redirecting clients to '%1$s'|%1$s"
```
markus456, line 10790: The use of pipes as a way to make this easy to parse makes it look (IMO) like a hack. What about using a non-localized message like @vaintroub suggested on MDEV-15935? This way the URL is at a constant offset which makes it simpler to deal with. This would eliminate the need to parse the message as the URL would be at a constant offset.

If localization is something that's desired, I think the leading pipe and duplication of data is redundant: you know from the error number that the message will be of some predefined format. You could drop the pipes altogether if you change it to something like `Server is redirecting clients to: %1$s` and make sure the localized message doesn't use colons. This gets rid of the "ugly" error message and also removes the duplication of the URL string. I think it also simplifies the client-side parsing by a little bit:
```c
    char *p= mysql->net.last_error;
    p = strstr(p, ": ");
    if (p) {
        host= p + 2;
        p= strchr(host, ':');
        if (p) {
            *p++ = '\0';
            port= atoi(p);
        }
        else
        {
            /* Restore to the default port, rather than reusing our current one */
            port= 0;
        }
        fprintf(stderr, "Got server redirect to '%s' (port %d)\n", host, port);
        ++server_redirects;
        goto tcp_redirect;
    }
```

vuvova:
We went on with the sysvar tracker approach as more flexible for a DBA to use.

It's unrelated to the forced/optional discussion, because a server can disconnect after sending the redirection info to the client either way, whether it sends an error message or a sysvar tracking packet.

PR #2682, opened 2023-06-29
MDEV-30619: Parallel Slave SQL Thread Can Update Seconds_Behind_Master with Active Workers

bnestere:
If the workers of a parallel replica are busy (potentially with long
queues), but the SQL thread has no events left to distribute (so it
goes idle), then the next event that comes from the primary will
update mi->last_master_timestamp with its timestamp, even if the
workers have not yet finished.

This patch changes the parallel replica logic which updates
last_master_timestamp after idling from using sql_thread_caught_up
(added in MDEV-29639) to using rli queued/dequeued event counters.
That is, if the queued count is equal to the dequeued count, it
means all events have been processed and the replica is idle.

Notes:
 * MDEV-29639 changed when sql_thread_caught_up was updated so it
could be used to set last_master_timestamp. This patch additionally
reverts those changes.
 * Renamed rpl_delayed_parallel_slave_sbm.test to
rpl_parallel_sbm.test to be more general to cover this test case as
well.

The first commit on the branch renames a test to serve a more general
case so it can be extended for the purpose of this patch. The second
commit on the branch is the regression to show the issue. The last
commit on the branch is the patch to fix the issue.

### DIFF
--- a/sql/slave.cc
+++ b/sql/slave.cc
@@ -1857,8 +1857,10 @@ static int get_master_version_and_clock(MYSQL* mysql, Master_info* mi)
       (master_row= mysql_fetch_row(master_res)))
   {
     mysql_mutex_lock(&mi->data_lock);
-    mi->clock_diff_with_master=
-      (long) (time((time_t*) 0) - strtoul(master_row[0], 0, 10));
+    mi->clock_diff_with_master= DBUG_EVALUATE_IF(
+        "negate_clock_diff_with_master", 0,
+        (long) (time((time_t *) 0) - strtoul(master_row[0], 0, 10)));
+
     mysql_mutex_unlock(&mi->data_lock);
   }
   else if (check_io_slave_killed(mi, NULL))
@@ -3187,6 +3189,14 @@ static bool send_show_master_info_data(THD *thd, Master_info *mi, bool full,
       else
       {
         idle= mi->rli.sql_thread_caught_up;
+
+        /*
+          The idleness of the SQL thread is needed for the parallel slave
+          because events can be ignored before distribution to a worker thread.
+          That is, Seconds_Behind_Master should still be calculated and visible
+          while the slave is processing ignored events, such as those skipped
+          due to slave_skip_counter.
+        */
         if (mi->using_parallel() && idle && !mi->rli.parallel.workers_idle())
           idle= false;
       }
@@ -4146,7 +4156,6 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli,
               thd,
               STRING_WITH_LEN(
                   "now SIGNAL paused_on_event WAIT_FOR sql_thread_continue")));
-          DBUG_SET("-d,pause_sql_thread_on_next_event");
           mysql_mutex_lock(&rli->data_lock);
         });
 
@@ -4163,7 +4172,8 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli,
       the user might be surprised to see a claim that the slave is up to date
       long before those queued events are actually executed.
      */
-    if ((!rli->mi->using_parallel()) && event_can_update_last_master_timestamp(ev))
+    if ((!rli->mi->using_parallel()) &&
+        event_can_update_last_master_timestamp(ev))
     {
       rli->last_master_timestamp= ev->when + (time_t) ev->exec_time;
       rli->sql_thread_caught_up= false;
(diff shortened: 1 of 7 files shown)


knielsen (commented):
I think to push this to a stable release as a bugfix, there needs to be a clear description of how Seconds_Behind_Master precisely should work, from the point of view of the user. If that is not clear, then it's not possible to know if a bugfix actually makes the behaviour correct. And the current documentation in the knowledgebase seems insufficient:

  "Difference between the timestamp logged on the master for the event that the replica is currently processing, and the current timestamp on the replica. Zero if the replica is not currently processing an event. With [parallel replication](https://mariadb.com/kb/en/parallel-replication/), seconds_behind_master is updated only after transactions commit."

The Seconds_Behind_Master I think has been modified a number of times and the actual semantics seems quite complex. Here is some old discussions:

    https://lists.launchpad.net/maria-developers/msg08958.html
    https://lists.launchpad.net/maria-developers/msg08961.html

I *think* the behaviour has changed somewhat since then due to various user requests etc., but I wonder if anyone actually understands exactly how it's supposed to work in detail.
So I think it's necessary to create a precise description of the behaviour of Seconds_behind_master, to show that the expected behaviour is actually known and can be matched to the code, and which could be put in the documentation when this is released.

bnestere:
Hi @knielsen!

Thanks for the feedback; that is a good point, the documentation is indeed lagging behind. As a starting point in its extension, I propose something like:
 
"Difference between the timestamp logged on the master for the event that the replica is currently processing, and the current timestamp on the replica. Zero if the replica is not currently processing an event. With serial replication, Seconds_Behind_Master is updated when the SQL thread reads a transaction from the relay log for execution. With parallel replication, Seconds_Behind_Master is updated only after transactions commit. Starting in 10.3.38, 10.4.28, 10.5.19, 10.6.12, 10.8.7, 10.9.5, 10.10.3, and 10.11.2, an exception is drawn on the parallel replica to additionally update Seconds_Behind_Master when the first transaction after idling is read from the relay log, to provide a more reliable lag estimate for the duration until the transaction commits."

Where MDEV-29639 is the basis for the extension, as it added a check when an event is read from the relay log to update `SBM` dependent on the idleness of the SQL thread. Then note that MDEV-30619 was opened alongside MDEV-29639's closure because the solution was incomplete, with the intention to extend the check to also check the state of the worker threads. However, after working on MDEV-30619, the former's approach being dependent on the idleness of the SQL thread now seems wrong entirely to me (rather than just incomplete), and this changes the "idleness check" to just use the progress of the worker threads. I tried to capture both changes in the documentation proposition above :)

knielsen:
Thanks, @bnestere , looks pretty good to me. It's good that this can be described in a single paragraph, even if the rules for SBM are somewhat complex.
(There's another can of worms with exactly what the "timestamp logged on the master" becomes set to, if it's still like it used to be. But that's not really related to this patch, I think it's fine like this for now.)

sql/rpl_parallel.cc
```diff
@@ -2435,21 +2435,9 @@ rpl_parallel::stop_during_until()
 bool
 rpl_parallel::workers_idle()
 {
-  struct rpl_parallel_entry *e;
```
andrelkin, line 2438: What do you think to keep this block in `#ifndef DBUG_OFF` and assert its `return` is equal the main branch's one?

sql/rpl_parallel.cc
```diff
@@ -2435,21 +2435,9 @@ rpl_parallel::stop_during_until()
...
-    e= (struct rpl_parallel_entry *)my_hash_element(&domain_hash, i);
-    mysql_mutex_lock(&e->LOCK_parallel_entry);
-    active= e->current_sub_id > e->last_committed_sub_id;
-    mysql_mutex_unlock(&e->LOCK_parallel_entry);
-    if (active)
-      break;
-  }
-  return (i == max_i);
+  DBUG_ASSERT(rli);
+  return rli->last_inuse_relaylog->queued_count ==
```
andrelkin, line 2439: For safety let's install an assert claiming equality for the two at slave start time, e.g into [rpl_parallel::reset()](https://github.com/MariaDB/server/blob/bb-10.4-MDEV-30619/sql/rpl_parallel.cc#L2275).

sql/slave.cc
```diff
@@ -4161,9 +4168,22 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli,
...
+
+      rli->sql_thread_caught_up is negated here to ensure that the value of
+      Seconds_Behind_Master in SHOW SLAVE STATUS is consistent with the update
+      of last_master_timestamp. It was previously unset immediately after
+      reading an event from the relay log; however, for the duration between
+      that unset and the time that LMT would be updated could lead to spikes
+      in SBM.
      */
-    if ((!rli->mi->using_parallel()) && event_can_update_last_master_timestamp(ev))
+    if ((unlikely(rli->last_master_timestamp == 0) ||
```
andrelkin, line 4182: As it pointed earlier `rli->last_master_timestamp == 0` must fall under `rli->last_master_timestamp < ev->when`, *and* (now it got clear to me) at that point `queued_count == dequeued_count` we could remove l.4182 disjunction argument or convert into `assert(lmt != 0 | queued == dequeued)`. For safely I suggest to do so in 11.3.

andrelkin (changes requested):

PR #2684, opened 2023-07-03
[MDEV-31585] Stop trusting or relying on client identifying information sent prior to the TLS handshake

dlenski:
- [x] *The Jira issue number for this PR is: MDEV-31585*

## Description

The server has heretofore improperly mishandled—and TRUSTED—information sent
in the [_plaintext_ login request packet](https://mariadb.com/kb/en/connection/#sslrequest-packet) sent prior to the TLS handshake.

As a result of this, the client is *forced* to send excessive and
exploitable identifying information in the pre-TLS-handshake plaintext login
packet. That client-side vulnerability is [CONC-654](https://jira.mariadb.org/browse/CONC-654).

This modifies the server to stop relying on any of the information in
the pre-TLS-handshake plaintext login packet EXCEPT for the single bit
that tells it that a TLS handshake will follow. It furthermore adds an
"extended capability" bit to the server greeting packet, which informs
the client that it is safe to send a bare-bones dummy packet containing
ONLY the instruction that a TLS handshake will follow:

```c
/* Server does not grievously mishandle information sent in the plaintext
 * login request packet sent prior to the TLS handshake. As a result, the
 * client can safely send an empty/dummy packet contianing no
 * identifying information. Indicates that MDEV-31585 has been fixed.
 * Since ??.?.
 */
#define MARIADB_CLIENT_CAN_SEND_DUMMY_HANDSHAKE_PACKET    (1ULL << 37)
```

## How can this PR be tested?

- [ ] TODO: automated MTR tests to verify backwards compatibility with older clients
- [ ] Build MariaDB Connector/C PR #???, with the corresponding client-side fix, and observe from Wireshark/tcpdump that the [pre-TLS-handshake login request packet](https://mariadb.com/kb/en/connection/#sslrequest-packet) now contains nothing except for the `CLIENT_SSL` bit, but the connection continues to work correctly.
image

  ***This is a bugfix and *should* be based on 10.2 (oldest maintained branch?) but is currently based on 11.2.***

## PR quality check


sql/sql_acl.cc
```diff
@@ -13736,45 +13739,37 @@ static ulong parse_client_handshake_packet(MPVIO_EXT *mpvio,
...
-      // it is client with mariadb extensions
-      ulonglong ext_client_capabilities=
-        (((ulonglong)uint4korr(net->read_pos + 28)) << 32);
-      client_capabilities|= ext_client_capabilities;
-    }
-  }
-
-  /* Disable those bits which are not supported by the client. */
-  compile_time_assert(sizeof(thd->client_capabilities) >= 8);
-  thd->client_capabilities&= client_capabilities;
```
dlenski, line 13768: The diff makes this look more complicated than it really is.

All that is really changing is that these few lines move from _before_ the TLS handshake to _after_ the TLS handshake.

These line tells the server to ***read capability bits sent by the client and to trust them***: the server alters its view of how the connection should work (`thd->client_capabilities&= `) based on them.

- If the server reads these bits from the pre-TLS handshake plaintext version of the login request packet (https://mariadb.com/kb/en/connection/#sslrequest-packet), it **forces the client to send this juicy identifying information in the clear**.
- Making the server instead read these bits from the post-TLS handshake version of the login request packet (https://mariadb.com/kb/en/connection/#handshake-response-packet), it allows the client to *only* send its complete capability bits (and other identifying information) over the TLS-secured connection.
   - … and the corresponding Connector/C change allows the client to do just that: ***TBD***

dlenski:
I've rebased this on top of 10.2, since it's a bugfix. This required an additional cherry-pick commit, e54c24c64b6, to add some support code (`vio_type_name` and `safe_vio_type_name` functions) which had never been added to 10.2.

Merging this correctly into all versions ranging from 10.2-11.2 is non-trivial.

The version of this branch targeted at **11.2** is available at: https://github.com/dlenski/mariadb-server/tree/dlenski-github/MDEV-31585_targeting_11.2

vuvova:
Looks rather risky to me. The server was ignoring capability flags from the after-ssl client packet for >20 years. It is totally possible that some third-party clients don't even send legitimate flags in that packet.

The part with `MARIADB_CLIENT_CAN_SEND_DUMMY_HANDSHAKE_PACKET` is safe, of course, I'll comment more in https://github.com/mariadb-corporation/mariadb-connector-c/pull/227

dlenski:
@vuvova wrote:

> Looks rather risky to me. The server was ignoring capability flags from the after-ssl client packet for >20 years. It is totally possible that some third-party clients don't even send legitimate flags in that packet.

Indeed, I would consider that highly probable.

Are you saying that the server must continue _indefinitely_ to support a buggy and insecure behavior _by default_ in the name of backwards compatibility?

A much more reasonable option would be to require the client and/or server to "opt in" to the old/buggy/insecure behavior (`--backwards-compatible-insecure-ssl`) in order to continue using it.

dlenski:
> A reasonable approach, followed by much other software, would be to aggressively fix the TLS vulnerabilities in both the client and server ***in a backwards-incompatible way***… but to allow the client and/or server to "opt in" to the old/buggy/insecure behavior (`--backwards-compatible-insecure-ssl`) in order to continue using it while the incompatibilities are worked out.

I've already implemented this.

See:

- https://github.com/dlenski/mariadb-connector-c/commit/force_client_to_opt_in_to_old_insecure_TLS:
  - Built on top of this PR
  - Adds the `backwards-compatible-insecure-ssl` option to Connector/C
  - Client _will not_ downgrade itself to CONC-648 or CONC-654 vulnerabilities unless specified
- https://github.com/dlenski/mariadb-server/commit/allow_client_to_opt_in_to_old_insecure_TLS
  - Adds the `--backwards-compatible-insecure-ssl` option to the `mariadb` client.
  - Trying to connect to a server without the MDEV-31585 fix *will fail*:  
    ```sh
    $ client/mariadb --host 127.0.0.1  --port=3306 --ssl
    ERROR 2026 (HY000): Client requires secure TLS/SSL, but server only supports old, buggy, insecure TLS/SSL. Specify backwards-compatible-insecure-ssl to override.
    ```
  - Adding `--backwards-compatible-insecure-ssl` allows it to succeed, with prominent warnings:
    ```sh
    $ client/mariadb --host 127.0.0.1 --port=3306 --ssl
    WARNING: Server only supports old, buggy, insecure TLS/SSL, and client has
      opted to use this (backwards-compatible-insecure-ssl set)
    ...
    ```
  - The same option will allow the client to continue accepting pre-(TLS handshake) error packets:
    ```
    $ client/mariadb --host 127.0.0.1 --port=3306 --ssl
    WARNING: Received error packet sent by server before completion of TLS
       handshake. This is insecure, but we allow it because client is configured
       with backwards-compatible-insecure-ssl.
    ...
    ```

vuvova:
> @vuvova wrote:
> 
> > Looks rather risky to me. The server was ignoring capability flags from the after-ssl client packet for >20 years. It is totally possible that some third-party clients don't even send legitimate flags in that packet.
> 
> Indeed, I would consider that highly probable…
> Unless you think that the server must continue _indefinitely_ to support buggy and insecure behaviors _by default_ in the name of backwards compatibility, with an unknown universe of client implementations?

That's pretty much what I'd suggest. But don't forget, after your mariadb-corporation/mariadb-connector-c#227 this new mode will be enabled by default, so only old third-party clients will still have the issue and it'll be their vulnerability, not MariaDB's — we'll update all our connectors by the next release.

dlenski:
@vuvova, I've added an additional commit 056d083972f which will make the server:

1. **_Detect_** the case where an active MITM is downgrading the client by stripping the `CLIENT_CAN_SEND_DUMMY_HANDSHAKE_PACKET` bit from our server greeting packet.
2. Abort the connection if this happens.

This won't prevent an active MITM attacker from stripping the bit to gather information on the client, but it _will_ cause a connection MITM'ed in this way to promptly fail, which should discourage attackers from using it.

markus456:
Could you document any changes to the MariaDB protocol [on the MariaDB KB](https://mariadb.com/kb/en/connection/) when the pull request gets accepted? This makes it easier to refer to the protocol without having to dig through the server source code.

LinuxJedi:
The CI won't run on this because it is against 10.2, which is way past EOL. If you make this against 10.4, it should run.

dlenski:
@LinuxJedi wrote:

> The CI won't run on this because it is against 10.2, which is way past EOL. If you make this against 10.4, it should run.

Done, rebased on `10.4`.

Are you actually planning to merge this and https://github.com/mariadb-corporation/mariadb-connector-c/pull/227? The Connector/C PR has received absolutely no feedback from its maintainer.

LinuxJedi:
> @LinuxJedi wrote:
> 
> > The CI won't run on this because it is against 10.2, which is way past EOL. If you make this against 10.4, it should run.
> 
> Done, rebased on `10.4`.
> 
> Are you actually planning to merge this and [mariadb-corporation/mariadb-connector-c#227](https://github.com/mariadb-corporation/mariadb-connector-c/pull/227)? The Connector/C PR has received absolutely no feedback from its maintainer.

This one I would like to see included in some form, it won't be my final decision to make. But I don't see the intention being difficult to argue. That being said, the submodule update is breaking tests and I suspect this would either need to go or be done differently.

The Connector/C one I have zero influence over. This is a decision for @9E0R9 to make. I agree that "something" needs to be done. I have not looked at the current state of the protocol on the MySQL side to see whether this is the correct way of resolving the problem. The spare bit appears to be contentious.

In an ideal world that would be a separate port that does a TLS handshake first, but I can not see a way of doing that in a MySQL compatible way. At which point it might as well be part of a new protocol implementation.

dlenski:
@LinuxJedi wrote:

> the submodule update is breaking tests and I suspect this would either need to go or be done differently.

What does "be done differently" mean?

There is absolutely ***nothing that I as an internal contributor can do*** to remedy this situation.

- The PR _correctly identifies_ the commit (https://github.com/mariadb-corporation/mariadb-connector-c/commit/23cbf7cc72ba56c93a99b6462a6774b939f8918f) to which the `libmariadb` submodule needs to be updated in order for the PR to function as advertised.
- Nevertheless, the submodule update seems to breaks Buildbot's brain. It doesn't seem to know how to check out and build correctly using the submodule.
- If you or another of the upstream devs can check this out and build it locally, you'll see that it works as advertised in the “How can this PR be tested?” section of the description.

> The Connector/C one I have zero influence over. This is a decision for @9E0R9 to make. I agree that "something" needs to be done. I have not looked at the current state of the protocol on the MySQL side to see whether this is the correct way of resolving the problem.

The substantial flaws in the protocol appears to have been ignored for many years, which is why we're in this situation.

> The spare bit appears to be contentious.

I don't know what this means.

The 

> In an ideal world that would be a separate port that does a TLS handshake first, but I can not see a way of doing that in a MySQL compatible way. At which point it might as well be part of a new protocol implementation.

Indeed, that might be ideal, but it is also not necessary.

There is a ton of low-hanging fruit in terms of poor design of the {MySQL,MariaDB}-over-TLS protocol, and poor implementation.

All of this can be simplified, improved, and made more secure **without** switching to an entirely separate TCP port.

LinuxJedi:
> @LinuxJedi wrote:
> 
> > the submodule update is breaking tests and I suspect this would either need to go or be done differently.
> 
> What does "be done differently" mean?
>
> There is absolutely _**nothing that I as an internal contributor can do**_ to remedy this situation.

Sorry, I should have expanded on that. I meant that it appears the submodule is pushing Connector/C quite far ahead of its current position, to a different major version. Which is not something we normally do in an older release. Usually, such changes would be in the major version of Connector/C paired with this MariaDB version.
 
>     * The PR _correctly identifies_ the commit ([mariadb-corporation/mariadb-connector-c@23cbf7c](https://github.com/mariadb-corporation/mariadb-connector-c/commit/23cbf7cc72ba56c93a99b6462a6774b939f8918f)) to which the `libmariadb` submodule needs to be updated in order for the PR to function as advertised.

Is it not possible to make the server rely on post-TLS handshake without a Connector/C change?

>     * Nevertheless, the submodule update seems to breaks Buildbot's brain. It doesn't seem to know how to check out and build correctly using the submodule.

It breaks in buildbot because it is switching major versions which have different error messages.

>     * If you or another of the upstream devs can check this out and build it locally, you'll see that it works as advertised in the “How can this PR be tested?” section of the description.

I suspect it would still break the regression suite.

> > The Connector/C one I have zero influence over. This is a decision for @9E0R9 to make. I agree that "something" needs to be done. I have not looked at the current state of the protocol on the MySQL side to see whether this is the correct way of resolving the problem.
> 
> The substantial flaws in the protocol appears to have been ignored for many years, which is why we're in this situation. The server and client implementation are tightly entangled, in terms of how they implement the {MySQL,MariaDB}-over-TLS protocol.
> 
> How can either side possibly make a decision about how to proceed him if the developers aren't actively communicating around such issues?

I know this situation is frustrating, and things in the background are actively changing to improve this situation. The Unconference in particular will help with trying to put better processes in place.

> > The spare bit appears to be contentious.
> 
> I don't know what this means.
> 
> Are you referring to the discussion (starting with [mariadb-corporation/mariadb-connector-c#227 (comment)](https://github.com/mariadb-corporation/mariadb-connector-c/pull/227#issuecomment-1631431897)) about exactly where we should place the "SSL_V2" capability bit?

Yes, this.

> > In an ideal world that would be a separate port that does a TLS handshake first, but I can not see a way of doing that in a MySQL compatible way. At which point it might as well be part of a new protocol implementation.
> 
> Indeed, that might be ideal, but it is also not necessary.
> 
> There is a ton of low-hanging fruit in terms of poor design of the {MySQL,MariaDB}-over-TLS protocol, and poor implementation.
> 
> All of this can be simplified, improved, and made more secure **without** switching to an entirely separate TCP port.

I agree, but we also have to be careful not to break compatibility with MySQL applications and connectors. This is why we are particularly cautious with protocol changes.

dlenski:
> Is it not possible to make the server rely on post-TLS handshake without a Connector/C change?

Indeed, it is _not possible_ to do so in a way that would be guaranteed to be backwards-compatible with existing clients.

- Because existing _servers_ don't _use_ the capabilities bits from the **post**-TLS client greeting packet, there is no guarantee that all clients are _sending_ them. (As @vuvova wrote in the first comment, https://github.com/MariaDB/server/pull/2684#issuecomment-1620131410)
- … and, because existing _servers_ only _use_ the capabilities bits from the **pre**-TLS client greeting packet, clients _have to_ send them.

Personally, I think the value of such bug-for-bug backwards-compatibility is very low, but if you insist on maintaining it then there's no wiggle room here. :man_shrugging:

> > * If you or another of the upstream devs can check this out and build it locally, you'll see that it works as advertised in the “How can this PR be tested?” section of the description.
>
> I suspect it would still break the regression suite.

What is the regression suite? How could I run or check this?

markus456:
I'll also comment the same thing here that I commented on the MariaDB Zulip. Instead of reserving a bit in the MariaDB extended capabilities, you could reserve the first byte of the filler space for this capability. This way the patch could be applied to all MariaDB and MySQL compatible databases without forcing them to adapt the MariaDB extended capability bits. This would allow you to offer the patch to all implementations which would make it easier to adopt into use.

dlenski:
So far I have proposed…

* Using one of the extended MariaDB capability bits (`#define CLIENT_CAN_SSL_V2    (1ULL << 37)`)
  * Criticized because MySQL doesn't use these capability bits (although there is no reason they _couldn't_ start, as @bgrainger from MySQL.NET [pointed out](https://github.com/mariadb-corporation/mariadb-connector-c/pull/227#issuecomment-1632532972)).
* Using (`#define CLIENT_CAN_SSL_V2    (1ULL << 28)`)
  * Oracle/MySQL [has already defined a use for this bit](https://github.com/mariadb-corporation/mariadb-connector-c/pull/227#issuecomment-1631634167) as well.

@markus456 wrote:
 
> Instead of reserving a bit in the MariaDB extended capabilities, you could reserve the first byte of the filler space for this capability.

That would be fine with me but… do you think that Oracle/MySQL is _more likely_ to adopt a protocol change that uses a bit in the "unused" area than they are to agree with using a bit in the area that MariaDB currently uses for its extended capabilities?

From the perspective of Oracle/MySQL servers and clients, the "unused" bits and the "MariaDB extended capabilities" are **all** unused bits :smiling_face_with_tear:

image

dlenski:
Based on the discussion on Zulip, @markus456 thinks this will have a better chance of adoption by MySQL if we pick a _new_ part of the unused bits and declare that to be something like the **"cross-implementation extended capabilities"**.

I'm all for it. :+1:

I assume it makes sense to start using the bits at the beginning, rather than the end, of the space that's currently unused by any implementation.

@LinuxJedi @vuvova, are you on board with proceeding in this direction?

LinuxJedi:
> Based on the discussion on Zulip, @markus456 thinks this will have a better chance of adoption by MySQL if we pick a _new_ part of the unused bits and declare that to be something like the **"cross-implementation extended capabilities"**.
> 
> I'm all for it. 👍
> 
> I assume it makes sense to start using the bits at the beginning, rather than the end, of the space that's currently unused by any implementation.
> 
> @LinuxJedi @vuvova, are you on board with proceeding in this direction?

I am a little cautious that it is possible for MySQL to use those bits for anything in the future and could in-theory break other connectors from talking to us. I'm wondering if this is something we could somehow work together with the MySQL community on, to ensure that whatever change is made will be supported both sides in the future.

My personal ideal solution (which @markus456 mentioned) is to have a MARIADBS port that does an immediate initial TLS handshake. Then a vast majority of problems immediately go away. This will require adoption over time by other connectors. But with some upstream community work for things such as PHP, as long as the rest of the protocol doesn't deviate too much, I think this is doable.

LinuxJedi:
Alternatively, if enough of the community group together, we can build a protocol definition working group (suggested by @grooverdan)

markus456:
From what I've seen, one of the major arguments against the changes in pull request has been that it doesn't improve the security enough to be worth the effort of getting this change accepted in all connectors and server implementations. As for whether the changes are "big enough" is not something that can be logically proven and the discussion around it is, in my opinion, going in circles.

I think that nobody can deny the fact that a separate TLS port is more secure and would immediately improve the security situation. It also improves the performance of the connection creation by removing the redundant SSLRequest packet. Performance is always something that people care about even if security is ranked above it. In the TLS port case, you get the best of both worlds: better security with less data sent over the wire.

Something which has not been discussed is the need for documenting how the protocol is used and what sorts of combinations of packets might be sent. This pull request introduces a new packet type that requires its own documentation and it changes the "state machine" required to implement the MariaDB/MySQL protocol. As someone who works ona projec that will eventually have to implement both the client and the server side implementations of this, simplicitly of the protocol extensions is a factor that affects us the most. The more complex a packet is to parse, the more open it is for errors and misinterpretations of the protocol. The TLS-first approach, on the other hand, simplifies the protocol by eventually removing the need for the SSLRequest packet. The TLS-first approach also simplifies the server-side implementation compared to this pull request by removing the need to check if the capabilities have been downgraded.

The big problem with the TLS port is the need to add a new option flag for all client programs and any connector that implements it. The practical effort of having to then configure two ports for nearly all installations is something that cannot be ignored. This is a heroic amount of effort from the community but it allows all MariaDB/MySQL compatible client and server programs to implement it rather trivially.

The only other approach is a new network protocol but I believe that this is an idea that's not going to be picked up simply because it requires several competing server implementations to implement a new protocol in unison and a large change on the client side for all connectors. The current protocol also offers no backwards compatible way of downgrading to an earlier protocol which requires that the client side sends some form of a "I want the new protocol" packet to the server's handshake response that further slows down and complicates the implementation. This is kind of what this pull request implements by adding a new packet to the network protocol.

@dlenski also pointed out after I asked on Zulip about the importance of the server capabilities and the version string being "leaked" in the handshake packet: neither the current suggested protocol change nor the TLS port approach solve this problem as the server still sends the exact version in the first packet. The TLS port would make it encrypted so it's not broadcast over the network but it would still be trivial to find out if one is actively scanning for ports. I think that it would be an improvement to "freeze" the version to something like `MariaDB 11.0` so that the exact version is only visible post-authentication.

Like @LinuxJedi mentioned (and @grooverdan suggested), perhaps it would be smart to first reach out to the other involved parties (Oracle, community connectors, hosting providers etc.) and see if a modified version of this patch (the "shared capability bits" thing) is acceptable for them. In addition, asking how willing they'd be to adapt to a separate TLS port would be a good idea.

vuvova:
To clarify the comment

> The big problem with the TLS port is the need to add a new option flag for all client programs and any connector that implements it.

A separate port, while, indeed, "nobody can deny the fact that a separate TLS port is more secure", would mean
* new server option `--tls-port` (for example)
* new server opton `--extra-tls-port`, because the server can listen on a second port, `--extra-port`
* new server option to disable non-tls port, `--port=0` won't do, it's already taken, so `--disable-non-tls-port`, for example (may be simply `--disable-port` ?)
* let's say we don't want different a `bind-address` on tls port, so this won't need a new option like `--tls-bind-address`
* new client option `--tls-port`. **For all connectors and clients.**
* client likely won't need more options, `--ssl --ssl-verify-server-cert` would mean to connect to the tls port. `--ssl --skip-ssl-verify-server-cert` would mean connect to tls port and, if failed, connect to the non-tls port. Could be slow (so, discouraging the old fallback behavior, which is a nice side effect)
* specifying a port on the command line would be less convenient, one cannot use `mariadb --port 1234`, one would need to know whether to specify `--tls-port` or `--port`. Which means knowing whether ssl is enabled in the config files. (why the user should need to know that?)

So, I suspect that this approach, while, indeed, undeniably more secure, has a rather slim chance of being adopted by other connectors and clients. Not unless it'll provide distinct benefits clearly outweighing the efforts of implementing (and explaining/documenting) it. Reaching out first, as @markus456 suggested, could be a sensible way forward.

svaroqui:
To give an humble opinion on this an extra port that only response to TLS from the first handshake looks better for any layer 4 proxies like nginx and proxysql that can route TCP traffic  base on the Certificate SNI extension and this could really help adoption in the cloud and with security pentests

vaintroub:
> To clarify the comment
> 
> > The big problem with the TLS port is the need to add a new option flag for all client programs and any connector that implements it.
> 
> A separate port, while, indeed, "nobody can deny the fact that a separate TLS port is more secure", would mean
> 
> * new server option `--tls-port` (for example)
> * new server opton `--extra-tls-port`, because the server can listen on a second port, `--extra-port`
> * new server option to disable non-tls port, `--port=0` won't do, it's already taken, so `--disable-non-tls-port`, for example (may be simply `--disable-port` ? it'll mean we'll also have `--disable-tls-port` and when both are disabled it'll mean `--skip-networking`)
> * let's say we don't want a different `bind-address` on tls port, so this won't need a new option like `--tls-bind-address`
> * new client option `--tls-port`. **For all connectors and clients.**
> * client likely won't need more options, `--ssl --ssl-verify-server-cert` would mean to connect to the tls port. `--ssl --skip-ssl-verify-server-cert` would mean connect to tls port and, if failed, connect to the non-tls port. Could be slow (so, discouraging the old fallback behavior, which is a nice side effect)
> * specifying a port on the command line would be less convenient, one cannot use `mariadb --port 1234`, one would need to know whether to specify `--tls-port` or `--port`. Which means knowing whether ssl is enabled in the config files. (why the user should need to know that?)
> 
> So, I suspect that this approach, while, indeed, undeniably more secure, has a rather slim chance of being adopted by other connectors and clients. Not unless it'll provide distinct benefits clearly outweighing the efforts of implementing (and explaining/documenting) it. Reaching out first, as @markus456 suggested, could be a sensible way forward.

While separate SSL port would require a couple of  server options, there is no substantial effort for the client. Client only has to specify if he connects to SSL port, that's one bit of information. Many popular connectors have an idea of "connection string,URL, or DNS. and adding useSSLPort=true to that is not a huge deal. Doing SSL handshake before first packet, based on that bit, is not that huge deal either.

gkodinov:
I will close the PR since there was no activity on it for close to 2 years. 

But I'd also like to offer my 2 cents on this: 

When you open the TCP socket to the mariadb server's listener, the MySQL protocol is such that it's the server sending the first packet and then waiting for the client to reply. 

Whereas TLS is the reverse: it's the client sending the first packet. 

Maybe we could have the server port have dual function? i.e. before sending the MySQL server hello, the server can peek the network to check if the client has sent something in (presumably the TLS ClientHello) and if it did, do TLS first prior to server hello. 

If it did not send anything, do the usual MySQL ServerHello. 

This sounds like a possible idea, since the TCP ACK packet and the TLS client hello can be batched (https://eyakubovich.github.io/2013-06-04-folding-ssltls-into-tcp-to-gain/)

Not sure how viable is this. Just throwing it out there.

PR #2685, opened 2023-07-04
MDEV-31618: Server crashes in process_i_s_table_temporary_tables/get_all_tables

an3l:
[x] *The Jira issue number for this PR is: MDEV-31618*
## PR quality check

### DIFF
--- a/mysql-test/main/information_schema_temp_table.result
+++ b/mysql-test/main/information_schema_temp_table.result
@@ -231,3 +231,38 @@ def	test	t1	BASE TABLE	MyISAM	10	Fixed	1	7	X	X	X	X	NULL	X	X	NULL	latin1_swedish_
 def	test	t2	TEMPORARY	MRG_MyISAM	10	Fixed	0	0	X	X	X	X	NULL	X	X	NULL	latin1_swedish_ci	NULL			X	Y
 def	test	t3	BASE TABLE	MRG_MyISAM	10	Fixed	1	5	X	X	X	X	NULL	X	X	NULL	latin1_swedish_ci	NULL			X	N
 DROP TABLE t1,t2,t3;
+#
+# MDEV-31618: Server crashes in
+#             process_i_s_table_temporary_tables/get_all_tables
+#
+CREATE TEMPORARY SEQUENCE seq1;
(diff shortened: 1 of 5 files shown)


mysql-test/main/information_schema_temp_table.result
```diff
@@ -231,3 +231,38 @@ def	test	t1	BASE TABLE	MyISAM	10	Fixed	1	7	X	X	X	X	NULL	X	X	NULL	latin1_swedish_
...
+#             process_i_s_table_temporary_tables/get_all_tables
+# 
+CREATE TEMPORARY SEQUENCE seq1;
+SHOW FULL TABLES;
+Tables_in_test	Table_type
+seq1	TEMPORARY SEQUENCE
+SELECT table_schema, table_name FROM  INFORMATION_SCHEMA.TABLES WHERE table_type='temporary sequence';
+table_schema	table_name
+test	seq1
+mysqltest	s2
```
an3l, line 245: @vuvova not sure how this happened, when `mysqltest` is removed before? 
`tmp_mem_root` maybe?

vuvova: sorry, I don't understand. what do you mean?

sql/sql_show.cc
```diff
@@ -5310,6 +5310,8 @@ int get_all_tables(THD *thd, TABLE_LIST *tables, COND *cond)
       }
 
       TABLE *tmp_tbl= share_temp->all_tmp_tables.front();
+      if(!tmp_tbl)
+        continue;
```
vuvova, line 5314: No, this is completely wrong. There cannot be a temporary table **share** without a single temporary **table**.

If this is the case here — it's a bug.

vuvova (changes requested):

mysql-test/main/information_schema_temp_table.result
montywi: The temporary sequences are shown because they where created in previous tests that did not clean up itself properly.
Fix by moving:
DROP TABLE mysqltest.s1;
DROP TABLE mysqltest.s2;
before the line
#MDEV-31618: Server crashes in

PR #2688, opened 2023-07-06
MDEV-31369 Disable TLS v1.0 and 1.1 for MariaDB

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-31369*

## Description

Remove TLSv1.1 from the default tls_version system variable.

Output a warning if TLSv1.0 or TLSv1.1 are selected.

Thanks Tingyao Nian for the feature request.

## How can this PR be tested?

Covered by existing tests.

- [ X ] *This is a new feature and the PR is based against the latest MariaDB development branch.*

## PR quality check

### DIFF
--- a/mysql-test/main/ssl_cipher.result
+++ b/mysql-test/main/ssl_cipher.result
@@ -66,3 +66,5 @@ Variable_name	Value
 Ssl_cipher_list	AES128-SHA
 disconnect ssl_con;
 connection default;
+call mtr.add_suppression("TLSv1.0 and TLSv1.1 are insecure");
+FOUND 2 /TLSv1.0 and TLSv1.1 are insecure/ in mysqld.1.err
--- a/mysql-test/main/ssl_cipher.test
+++ b/mysql-test/main/ssl_cipher.test
@@ -101,3 +101,9 @@ SHOW STATUS LIKE 'Ssl_cipher';
 SHOW STATUS LIKE 'Ssl_cipher_list';
 disconnect ssl_con;
 connection default;
+
+# MDEV-31369 Disable TLS v1.0 and 1.1 for MariaDB
+call mtr.add_suppression("TLSv1.0 and TLSv1.1 are insecure");
+--let SEARCH_FILE=$MYSQLTEST_VARDIR/log/mysqld.1.err
+--let SEARCH_PATTERN= TLSv1.0 and TLSv1.1 are insecure
+--source include/search_pattern_in_file.inc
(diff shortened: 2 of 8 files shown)


PhysicsTing:
Thanks @grooverdan for addressing this! Code change looks good to me especially the addition of the warning message.

I still believe it will be very beneficial if this can be merged to 10.4. Impact to backward compatibility will be very minimum because it only breaks if the client program explicitly uses only TLSv1.1 (TLSv1.0 is already disabled before). Even then, users can still re-enable TLSv1.1 easily.

LinuxJedi (approved):

fraser-nds:
Hi @PhysicsTing & @grooverdan, we are currently on 10.4 and experiencing this issue in production with one of our third-party applications. It would be good to re-enable TLSv1.1 (as you suggest) to fix our problem while we work with the third party to update their application to use latest TLS.

Do you have brief instructions on how to re-enable TLSv1.1?

Thanks,
Fraser

vuvova:
it's `--tls-version` server option. The commit has only changed its default value, you can still use `--tls-version=TLSv1.1,TLSv1.2,TLSv1.3`

fraser-nds:
Thanks @vuvova 

Do you have any advice on how to apply this change? In particular we're using an AWS RDS managed database so we might not have full control over server functionality. For example, could it be applied with a query?

e.g.
```SET tls_version='TLSv1.1,TLSv1.2,TLSv1.3';```

grooverdan:
Docs show that [tls_version](https://mariadb.com/kb/en/ssltls-system-variables/#tls_version) isn't Dynamic, and therefor cannot be set at runtime.

On the [AWS RDS page](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/MariaDB.Concepts.FeatureSupport.html), tls_version is a parameter from 10.4 upwards so I assume is configurable somewhere in their interfaces;

PhysicsTing:
Hi Fraser,

You can config parameter 'tls_version' by creating a customized RDS
parameter group and associate it to your instance. However due to security
requirements RDS stops allowing using 'tls_version' TLSv1.0 and TLSv1.1
from MariaDB 10.4.32 and above. If you have to use TLSv1.1, please use
MariaDB 10.4.31 or lower.

Working with parameter group:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_WorkingWithDBInstanceParamGroups.html

Tingyao,

fraser-nds:
Thanks @grooverdan and @PhysicsTing 

It looks like RDS doesn't allow you to set the TLS back to 1.1 due to security requirements on their side. And AWS already upgraded us to 10.4.32 so it's not possible to go back a version with RDS. I've raised a support ticket with AWS to see if there is anything they can do.

PR #2691, opened 2023-07-10
MDEV-31413 : Node has been dropped from the cluster on Startup / Shut…

janlindstrom:
…down with async replica

- [x] *The Jira issue number for this PR is: MDEV-31413*

## Description
There was two related problems:

(1) Galera node that is defined as a slave to async MariaDB master at restart might do SST (state stransfer) and part of that it will copy mysql.gtid_slave_pos table. Problem is that updates on that table are not replicated on a cluster. Therefore, table from donor that is not slave is copied and joiner looses gtid position it was and start executing events from wrong position of the binlog. This incorrect position could break replication and causes node to be dropped and requiring user action.

(2) Slave sql thread might start executing events before galera is ready (wsrep_ready=ON) and that could also cause node to be dropped from the cluster.

In this fix we enable replication of mysql.gtid_slave_pos table on a cluster. In this way all nodes in a cluster will know gtid slave position and even after SST joiner knows correct gtid position to start.

Furthermore, we wait galera to be ready before slave sql thread executes any events to prevent too early execution.

## How can this PR be tested?

- [ X] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [X ] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

### DIFF
--- a/mysql-test/suite/galera/r/galera_as_slave_gtid_myisam.result
+++ b/mysql-test/suite/galera/r/galera_as_slave_gtid_myisam.result
@@ -17,12 +17,15 @@ SELECT LENGTH(@@global.gtid_binlog_state) > 1;
 LENGTH(@@global.gtid_binlog_state) > 1
 1
 connection node_1;
+SELECT COUNT(*) AS EXPECT_1 FROM t1;
+EXPECT_1
+1
 gtid_binlog_state_equal
 0
 connection node_2;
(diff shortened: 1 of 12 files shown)


sysprg:
Thanks, merged with head revision as https://github.com/MariaDB/server/commit/277968aa4cfa46a290af1c44e9d756381005939a

PR #2372, opened 2022-12-13
Minimize unsafe C functions usage - replace strcat() and strcpy() (and strncat() and strncpy()) with custom safe_strcat() and safe_strcpy() functions. Part one.

Chaloff:
Minimize unsafe C functions usage - replace strcat() and strcpy() with strncat() and strncpy(). Part one.

 ## Description
The MariaDB code base uses strcat() and strcpy() in several
places. These are known to have memory safety issues and their usage is
discouraged. Common security scanners like Flawfinder flags them. In MariaDB we
should start using modern and safer variants on these functions.

Those changes continue the similar memory issues fixes started at
[8298d32](https://github.com/MariaDB/server/pull/2210/commits/8298d32d8ef9b2207f267ff6e308c43450d2a97e) and
[19af189](https://github.com/MariaDB/server/commit/19af1890b56c6c147c296479bb6a4ad00fa59dbb).
Now we are taking care of some of the usage of strcat() and strcpy(). We are
replacing them with the safer options - strncat() and strncpy() but since these
2 functions also have well known disadvantages - it's not guaranteed what new
string will be null-terminated - we are adding '\0' forcefully to make sure the
result string is correct:
```
size_t dest_len = sizeof(g->Message);
strncpy(g->Message, "Null json tree", dest_len); strncat(g->Message, ":",
sizeof(g->Message) - strlen(g->Message)); size_t wrote_sz = strlen(g->Message);
size_t cur_len = wrote_sz >= dest_len ? dest_len - 1 : wrote_sz;
g->Message[cur_len] = '\0';
```

We would like to provide 2-3 smaller size changes to make code review and
merge easier. Next part will follow.

All build stages pass for these commits. For the
commit without the intentional bug, mysql-test-run is properly passing. The
changes do not change the functionality itself they just replace unsafe C
methods with the safe ones.

- [x] *This is a bug fix and the PR is based against the earliest branch in which the bug can be reproduced*

The changes are fully backward compatible.

All new code of the whole pull request, including one or several files
that are either new files or modified ones, are contributed under the BSD-new
license. I am contributing on behalf of my employer Amazon Web Services,
Inc.

PR #2692, opened 2023-07-10
Replace safe_strcpy with a more efficient version that doesn't trigger GCC 8+ warnings

dlenski:
## Description

The `safe_strcpy()` function was added in
https://github.com/mariadb/server/commit/567b68129943#diff-23f88d0b52735bf79b7eb76e2ddbbebc96f3b1ca16e784a347525a9c43134d77

Unfortunately, its current implementation triggers many GCC 8+ string truncation and array bounds warnings, particularly due to the potential for a false positive `-Warray-bounds`.

For example, the line `safe_strcpy(delimiter, sizeof(delimiter), ";")` in `client/mysqldump.c` causes the following warning:

    [1669/1914] Building C object client/CMakeFiles/mariadb-dump.dir/mysqldump.c.o
    In file included from /PATH/include/my_sys.h:20,
                     from /PATH/mysqldump.c:51:
    In function ?safe_strcpy?,
        inlined from ?dump_events_for_db.isra? at /PATH/client/mysqldump.c:2595:3:
    /PATH/include/m_string.h:258:39: warning: array subscript 1535 is outside array bounds of ?const char[2]? [-Warray-bounds=]
      258 |   if (dst[dst_size - 2] != '\0' && src[dst_size - 1] != '\0')
          |                                    ~~~^~~~~~~~~~~~~~

GCC is reporting that the `safe_strcpy` function *could* cause an out-of-bounds read from the constant *source* string `";"`, however this warning is unhelpful and confusing because it can only happen if the size of the *destination* buffer is incorrectly specified, which is not the case here.

In https://github.com/MariaDB/server/pull/2640, Andrew Hutchings proposed fixing this by disabling the `-Warray-bounds` check in this function (specifically in
https://github.com/MariaDB/server/pull/2640/commits/be382d01d08739d081f6cf40f350f7414f29b49d#diff-23f88d0b52735bf79b7eb76e2ddbbebc96f3b1ca16e784a347525a9c43134d77R255-R262).

However, this was rejected because it also disables the *helpful* `-Warray-bounds` check on the destination buffer.

This commit should solve these problems:

1. It reimplements `safe_strcpy` a bit more efficiently, preventing the need for a duplication `memset(dst, 0, dst_size)` since `strncpy` already pads `dst` with 0 bytes.
2. It will not trigger the `-Warray-bounds` warning, because `src` is not read based on an offset determined from `dst_size`.
3. It does trigger the `-Wstringop-truncation` warning in `strncpy` (https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wstringop-truncation), so we need to disable that.  However, this is a less broad and generally-useful warning so there is no loss of static analysis value caused by disabling it.

## How can this PR be tested?

Build with GCC 8 or newer in release mode.

Verify the absence of build failures caused by `-Warray-bounds -Werror` false positives.

- [x] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced. 
   - 10.3 is the earliest branch to which https://github.com/mariadb/server/commit/567b68129943 was merged, causing the false-positive `-Warray-bounds`, so that's where I've applied the fix.

## PR quality check


dlenski:
@grooverdan and @LinuxJedi, this is intended as a replacement for https://github.com/MariaDB/server/pull/2640.

On that PR, both of you expressed misgivings about disabling the extremely useful `-Warray-bounds` warning in order to avoid the false positives in it.  (https://github.com/MariaDB/server/pull/2640#issuecomment-1552442599, https://github.com/MariaDB/server/pull/2640#issuecomment-1552774654)

This PR is intended to _prevent_ the false positives, _without_ requiring us to disable the very useful array-bounds checking.

include/m_string.h
```diff
@@ -235,14 +235,31 @@ static inline void lex_string_set3(LEX_CSTRING *lex_str, const char *c_str,
...
+#if defined(__GNUC__) && __GNUC__ >= 8
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+#endif
+  strncpy(dst, src, dst_size);
+#if defined(__GNUC__) && __GNUC__ >= 8
+#pragma GCC diagnostic pop
+#endif
+
+  if (dst[dst_size - 1] != '\0') {
```
dlenski, line 258: This version avoid reading `src` (no `&& src[dst_size - 1] != '\0'`) so it avoids triggering a false-positive `-Warray-bounds` here.

dlenski:
> This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.
> * 10.3 is the earliest branch to which https://github.com/MariaDB/server/commit/567b68129943 was merged, causing the false-positive `-Warray-bounds`, so that's where I've applied the fix.

@LinuxJedi, I targeted this PR to 10.3, because that's the earliest branch where the false-positive warnings exist.

Should I rebase and retarget the PR to 10.4, since 10.3 is no longer maintained?

vuvova:
Is `safe_strcpy()` significantly different from `strmake()` ?
(`strmake()` is defined in `strings/strmake.c`)

dlenski:
@vuvova wrote:

> Is `safe_strcpy()` significantly different from `strmake()` ?

I'm not sure who this question is for.

Whether or not the overlapping functionality of `safe_strcpy` and `strmake` is desirable, it seems somewhat orthogonal to the issue of reducing false-positive warnings from `safe_strcpy`.

vuvova:
Sorry, I thought you might know, as you're investing your time into fixing `safe_strcpy`.

If it completely duplicates `strmake` then it should simply be removed and this will stop GCC 8+ warnings too.

dlenski:
> Sorry, I thought you might know, as you're investing your time into fixing `safe_strcpy`.

To be clear, I'm _only_ trying to change it to reduce spurious warnings on GCC 8+.

> If it completely duplicates `strmake` then it should simply be removed and this will stop GCC 8+ warnings too.

It seems that other core developers thought `safe_strcpy` was specifically useful, based on the lengthy discussion in #2372, which added it, and [the decision by @cvicentiu to merge it](https://github.com/MariaDB/server/pull/2372#issuecomment-1398489453).

dlenski:
> If it completely duplicates `strmake`

These functions have inconsistent APIs…

# Comparison of API

## `strmake()`

1. Takes parameters `dst`, `src`, `length` in that order
2. :warning: `dst` needs to have space for `length`***`+1`*** bytes
3. Returns pointer to the `\0` byte written at the end of `dst`

https://github.com/MariaDB/server/blob/584c2351dee93b7691c6726a0f55bac6557aa696/strings/strmake.c#L32-L42

## `safe_strcpy()`

1. Takes parameters `dst`, `dst_size`, `src`, in that order
2. :green_circle: `dst` needs to have space for `dst_size` bytes  
3. Returns 1 if `src` was truncated, 0 if `src` was not truncated

https://github.com/MariaDB/server/blob/HEAD/include/m_string.h#L242-L262

# Summary

The parameter order difference is annoying, and could easily be fixed.

The off-by-1 caveat in the `length` parameter of `strmake` seems extremely error-prone, and could be fixed.

The return values of these two functions make them useful for different purposes which are difficult to reconcile:

- `strmake` is useful for repeated concatenation to the same buffer, without running into the `O(length^2)` problem of repeated `strcat` (known as [Schelemiel the Painter's Algorithm](https://en.wikichip.org/wiki/schlemiel_the_painter%27s_algorithm)).
- `safe_strcpy` is useful for detecting checking that buffers are big enough, and going to an error path, rather than silently truncating output

LinuxJedi (changes requested):
Good call. Fundamentally I'm completely in favour of this. But Monty implemented it in an almost identical way in https://github.com/MariaDB/server/commit/a7adfd4c52307876d68ad3386cefd3757ee66e92 which was merged into 10.5.

I'd prefer if you could cherry-picking backporting this patch to 10.3 instead, even though it is pretty much functionally identical to your patch so we don't have a confusing history. Then if the stringop-truncation warning needs to be disabled, layer that as a second commit on top. This will then easily apply when merging up.

LinuxJedi:
Oh yes, and to answer the question, yes please with the 10.4 instead of 10.3 :)

LinuxJedi:
Also, can you please add a BSD declaration to the PR comments for posterity?

dlenski:
@LinuxJedi wrote:

> Fundamentally I'm completely in favour of this. But Monty implemented it in an almost identical way in https://github.com/MariaDB/server/commit/a7adfd4c52307876d68ad3386cefd3757ee66e92 which was merged into 10.5.

Aha, thanks, I hadn't seen that. It's essentially identical to this PR, _except_ that it doesn't disable the `-Wstringop-truncation` warning. (Probably @montywi wasn't testing with GCC v8+ with `-Wstringop-truncation -Werror`.)

> I'd prefer if you could cherry-picking this patch to 10.4 instead, even though it is pretty much functionally identical to your patch so we don't have a confusing history. Then if the stringop-truncation warning needs to be disabled, layer that as a second commit on top. This will then easily apply when merging up.

Sure, that makes sense. So, the TODO here is …

- [ ] Take https://github.com/MariaDB/server/commit/a7adfd4c52307876d68ad3386cefd3757ee66e92 and cherry-pick it onto 10.4
- [ ] Add the disabling of `-Wstringop-truncation` as a second commit on top of this
- [ ] Retarget this PR to 10.4

dlenski:
> Sure, that makes sense. So, the TODO here is …
> 
> * [x]  Take [a7adfd4](https://github.com/MariaDB/server/commit/a7adfd4c52307876d68ad3386cefd3757ee66e92) and cherry-pick it onto 10.4
> * [x]  Add the disabling of `-Wstringop-truncation` as a second commit on top of this
> * [x]  Retarget this PR to 10.4

Done, done, and done. Please re-review @LinuxJedi!

vuvova:
some random thoughts about `strmake` vs `safe_strcpy`:
* we don't need two nearly identical functions
* agree that `+1` is a nuisance and error prone, but it was like that since at least 1996 (the oldest MySQL sources google was able to find) and for me its removal is not worth fighting for
* safe_strcpy(D,L,S), can be replaced, like, `strmake(D,S,L-1)==D+L`
* on the other hand it's not 1996 anymore and `strmake` is doing bytewise copying, while `strncpy` is gcc builtin and likely not even a function call
* so perhaps `strmake` should be rewritten in terms of `safe_strcpy()`
* on yet another hand, compilers, I suspect, can recognize the pattern of bytewise copying pretty well and replace it with something optimized
* that needs studying the generated assembly (I haven't done it yet)

Anyway, a cherry-pick of a commit to an earlier version shouldn't concern with any of that, I suppose @LinuxJedi will review and merge it when appropriate. Thanks, @dlenski, for doing it!

And the above thoughts will stay for some future time when I'll get to reconciling two very similar functions.

dlenski:
> * on yet another hand, compilers, I suspect, can recognize the pattern of bytewise copying pretty well and replace it with something optimized
> * that needs studying the generated assembly (I haven't done it yet)

https://godbolt.org/z/b568bMWo4

LinuxJedi:
> some random thoughts about `strmake` vs `safe_strcpy`:
> 
> * we don't need two nearly identical functions
> * agree that `+1` is a nuisance and error prone, but it was like that since at least 1996 (the oldest MySQL sources google was able to find) and for me its removal is not worth fighting for
> * safe_strcpy(D,L,S), can be replaced, like, `strmake(D,S,L-1)==D+L`
> * on the other hand it's not 1996 anymore and `strmake` is doing bytewise copying, while `strncpy` is gcc builtin and likely not even a function call
> * so perhaps `strmake` should be rewritten in terms of `safe_strcpy()`
> * on yet another hand, compilers, I suspect, can recognize the pattern of bytewise copying pretty well and replace it with something optimized
> * that needs studying the generated assembly (I haven't done it yet)
> 
> Anyway, a cherry-pick of a commit to an earlier version shouldn't concern with any of that, I suppose @LinuxJedi will review and merge it when appropriate. Thanks, @dlenski, for doing it!
> 
> And the above thoughts will stay for some future time when I'll get to reconciling two very similar functions.

I do agree that longer term these two could well be reconciled in an optimised way. IIRC glibc implements `strncpy` as a `strlen`, `memset` and `memcpy`, which are typically well optimised.

vuvova:
> https://godbolt.org/z/b568bMWo4

Thanks, @dlenski ! This is very interesting. Neither latest gcc nor clang optimize `strmake` loop, they both compile it into a byte-wise copy. So `strmake` definitely needs fixing for 2023+

dlenski:
> Excellent, many thanks!

@LinuxJedi, looks like you approved this and rebased it to the latest HEAD of 10.4, but didn't merge. Did you mean to?

dlenski:
@vuvova wrote:

> * agree that +1 is a nuisance and error prone, but it was like that since at least 1996 (the oldest MySQL sources google was able to find) and for me its removal is not worth fighting for

These kinds of off-by-one offsets add an additional cognitive load to developers, which makes them error-prone to use. 

Because of problems like these, which pretty much all C-based projects share, Linux kernel + libc devs have been working quite hard on making it easier to statically check more arrays (https://lwn.net/Articles/908817/) and to use less error-prone string copying functions (https://lwn.net/Articles/905777/).

> * `safe_strcpy(D,L,S)`, can be replaced, like, `strmake(D,S,L-1)==D+L`

These are _not_ equivalent.

`safe_strcpy(D, L, S)` will return 1 to indicate truncation if `strlen(S) >= L`

`strmake(D, S, L-1) == D+L` will _never_ be true. `strmake` returns a pointer to the 0 byte written after copying the string, so the maximum return value is `D+L-1`.

Even if you fix _that_ off-by-one mistake, there's still another subtler one: `strmake(D, S, L-1) == (D+L-1)` will be true if `strlen(S) >= L-1`, not if `strlen(S) >= L`. In other words, it can't distinguish the case where the destination buffer was _just barely big enough_, 0 bytes to spare, from the case where the destination buffer is too small.

vuvova:
@dlenski wrote:

> @vuvova wrote:
> 
> > * agree that +1 is a nuisance and error prone, but it was like that since at least 1996 (the oldest MySQL sources google was able to find) and for me its removal is not worth fighting for
> 
> These kinds of off-by-one offsets add an additional cognitive load to developers, which makes them error-prone to use.

I totally agree. What I meant was that _some_ old-time developers will insist that this `+1` behavior is very good and intuitive and useful. And while I dislike +1 behavior, I don't dislike it enough to fight for its removal. I am definitely not going to defend it.


> > * `safe_strcpy(D,L,S)`, can be replaced, like, `strmake(D,S,L-1)==D+L`
> 
> These are _not_ equivalent.
> 
> `safe_strcpy(D, L, S)` will return 1 to indicate truncation if `strlen(S) >= L`
> 
> `strmake(D, S, L-1) == D+L` will _never_ be true. `strmake` returns a pointer to the 0 byte written after copying the string, so the maximum return value is `D+L-1`.

Right, thanks. That was supposed to show the idea, I didn't test it to see whether it was actually correct :innocent: 
 
> Even if you fix _that_ off-by-one, and use `strmake(D, S, L-1) == (D+L-1)`, there's a subtler one: this expression will be true if `strlen(S) >= L-1`, not if `strlen(S) >= L`. In other words, it can't distinguish the case where the destination buffer was _just barely big enough_, 0 bytes to spare, from the case where the destination buffer is too small.

That's true.

LinuxJedi:
> > Excellent, many thanks!
> 
> @LinuxJedi, looks like you approved this and rebased it to the latest HEAD of 10.4, but didn't merge. Did you mean to merge?

I did mean to, unfortunately after a rebase I can't merge until after CI does the required passes, and then I got pulled away for something else. Sorry about that.

PR #2693, opened 2023-07-11
MDEV-7850: MariaDB doesn't show thread_id for ROW-based events in binlog

bnestere:
Where the thread_id that executed a statement is present in the
mysqlbinlog output of a Query_log_event, it has been requested
that GTID_log_events also present this value. 

This PR extends GTID log events to store the thread_id of the
executing thread. It is initially separated into four commits:

 1. Andrei's refinement of the original patch of Sujatha and
Sachin (uploaded into the patch as (`7850.patch`)
 2. Removal of the `7850.patch` file
 3. Updated the GTID extra flags to account for new flags
in later versions of the server. Also added in the original test
case from the `7850.patch`, and refined/extended it.
 4. Fixed other MTR tests which fail due to these changes.


andrelkin:
@bnestere, I've made a "contribution" review , to push [an amended patch](https://github.com/MariaDB/server/commit/98e86dc0646) to bb-10.5-andrei. 
Could you please rebase it onto 11.4 and follow further preview steps.

PR #2698, opened 2023-07-17
MDEV-31714 Debian 50-mariadb_safe.cnf has syslog enabled

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-31714*

## Description

50-mariadb_safe.cnf was only ever intended to be a reference file. With mariadbd-safe falling into disuse in Debian since systemd became a service it was assumed this files wasn't used except on a few old legacy init systems. Its assumed these users will have a configuration already set.

It is however read by the mariabackup galera-sst script. This forks off the logger process to write the output to /dev/log, which on a systemd service the journalctl is reading anyway.

The real harm is on containers where there is now a MariaDB operator that runs Galera based containers. In containers there is no /dev/log, so information gets lost.

It would be much more direct just to go straight to stdout/error like what would happen without this configuration.

Rather than risk the galera-sst scripts moving to the [mariadbd-safe] group name introduced in a1211a4eda9c7348295ce4a6fdfa8e94860e5e14, we remove the file for cleanness.

## How can this PR be tested?

Instigate a SST like https://github.com/MariaDB/mariadb.org-tools/tree/master/daniel/galera-sst-test.

- [ X ] *This is a new feature and the PR is based against the latest MariaDB development branch.*

## PR quality check


ottok:
Nitpick: The current Markdown syntax for the checkboxes you use:

```
- [ X ] *This is a new feature and the PR is based against the latest MariaDB development branch.*

## PR quality check
```

.. leads to this
image

image

Correct would be:

```

## PR quality check
```

ottok (approved):
This config file has been around since the [dawn](https://github.com/MariaDB/server/commit/cfd4fcb0bc3d469dfca74dae30d17250d65fdd91) to MariaDB Debian packaging, with filename `etc/mysql/conf.d/mysqld_safe_syslog.cnf`.

The section has always been under `[mysqld_safe]` or `[mariadb-safe]` which I assume is why it worked and was only read by the legacy init script. However if this is not honored by the mariabackup galera-sst script as your commit message states, removing this is fine.

If we don't want to support mariadb-safe and traditional SystemV init, we should probably later also remove the mariadb-safe thing completely.

grooverdan:
> This config file has been around since the [dawn](https://github.com/MariaDB/server/commit/cfd4fcb0bc3d469dfca74dae30d17250d65fdd91) to MariaDB Debian packaging, with filename `etc/mysql/conf.d/mysqld_safe_syslog.cnf`.

ack
> 
> The section has always been under `[mysqld_safe]` or `[mariadb-safe]` which I assume is why it worked and was only read by the legacy init script.

I was too for a long time too.

> However if this is not honored by the mariabackup galera-sst script as your commit message states, removing this is fine.

Not so much not honored but treated incorrectly as an operationally active setting. While probably more correctly this is a galera-sst problem that also needs fixing, its operational importance to most usages of the file is still none.

> If we don't want to support mariadb-safe and traditional SystemV init, we should probably later also remove the mariadb-safe thing completely.

Not so much unsupported as those who want it know how to use it, know how to use it and its still documented.

PR #2699, opened 2023-07-18
MDEV-31545 GCC 13 -Wdangling-pointer in execute_show_status()

grooverdan:

## Description

The pointer was used deep in the call path.

Resolve this by setting the pointer to NULL at the end of the function.

Tested with gcc-13.3.1 (fc38)

The warning disable 38fe266ea953 can be reverted in 10.6+ on merge.

## How can this PR be tested?

Compiler with gcc-13.3.1+ and run mtr.

## PR quality check


LinuxJedi (approved):
Seems sane to me.

PR #2700, opened 2023-07-18
MDEV-31727: pcre stack size not functioning on clang-16

grooverdan:
Thanks Zhixu Liu for the great bug report.

- [x] *The Jira issue number for this PR is: MDEV-31727*

## Description

noinline attribute was being ignored by clang-16 and reporting 32 stack size on Gentoo, 16 locally on Fedora 38.

Based on https://stackoverflow.com/questions/54481855/clang-ignoring-attribute-noinline changed the noinline to noopt.

After that the -pcre_exec(NULL, NULL, NULL, -999, -999, 0, NULL, 0) returned 1056, simlar to gcc.

From https://bugs.gentoo.org/910188.

## How can this PR be tested?

Manual compile against bundled pcre with code per gentoo bug report.

## PR quality check


clan:
Does gcc support __attribute__((optnone)) ?
If not, maybe we should add a compiler check?

grooverdan:
gcc-13.1.1 doesn't support the `optnone`, and clang didn't support the `noclone`, but both seem to ignore the ones they don't support.

vuvova:
ideally, the fix should work with the system pcre too. If we patch the bundled version, MariaDB will not be able to use system pcre anymore

vuvova (approved):

grooverdan:
And luckily for the most part system builds are still gcc based.

@vuvova still got an account to update https://bugs.exim.org/show_bug.cgi?id=2173, apparently other's had trouble registering.

thesamesam:
> ideally, the fix should work with the system pcre too. If we patch the bundled version, MariaDB will not be able to use system pcre anymore
> 
> UPD: ok, it's 10.4-only, and we check whether system pcre is usable or not, then I suppose we can fix only bundled. It'll be up to Gentoo/Fedora to patch their distro version of pcre if they want to.

Done for Gentoo, thanks!

dr-m:
I wonder if this is in any way related to [MDEV-31605](https://jira.mariadb.org/browse/MDEV-31605), which involves JSON functions.

I was thinking of adding the following to `include/my_compiler.h`:
```c
#ifdef __clang__
# define ATTRIBUTE_OPTNONE __attribute__((optnone))
#elif defined __GNUC__
# define ATTRIBUTE_OPTNONE __attribute__((optimize(0)))
#else
# define ATTRIBUTE_OPTNONE /* empty */
#endif
```
As you probably know, Clang defines `__GNUC__` (as 4), so we would need the checks in that order. The oldest GCC version that we currently support is 4.8.5, and it does recognize the `optimize` attribute.

PR #2704, opened 2023-07-19
Cleanups in ACL related code. Preparation for future development

cvicentiu:
- [x] *The Jira issue number for this PR is: MDEV-31741*

## Description
This is a set of refactoring commits and a few bug fixes. Not significant logic changes, but it does make the code easier to extend down the line.

## PR quality check


sql/sql_hset.h
```diff
@@ -72,6 +72,14 @@ class Hash_set
   {
     return my_hash_delete(&m_hash, reinterpret_cast<uchar*>(value));
   }
+
+  T *find(const T *other) const
+  {
+    size_t klen;
+    uchar *key= m_hash.get_key(other, &klen, false);
+    return (T*)my_hash_search(&m_hash, reinterpret_cast<uchar *>(key), klen);
```
vuvova, line 80: may be
```c++
  return find(key, klen);
```
?

sql/sql_acl.cc
```diff
@@ -3609,7 +3609,7 @@ bool check_some_access(THD *thd, privilege_t want_access, TABLE_LIST *table)
     1	Access denied.  In this case an error is sent to the client
 */
 
-bool check_global_access(THD *thd, privilege_t want_access, bool no_errors)
+bool check_global_access(THD *thd, const privilege_t want_access, bool no_errors)
```
vuvova, line 3633: I don't get it. in the previous commit all prototypes of `check_global_access` don't have `const`, where is this one coming from?

sql/sql_acl.cc
```diff
@@ -2464,13 +2464,16 @@ static bool set_user_salt_if_needed(ACL_USER *user_copy, int curr_auth,
...
   if (lex_string_eq(&auth->plugin, &old_password_plugin_name))
+  {
     auth->plugin= old_password_plugin_name;
-  else
-    return true;
-  return false;
+    return false;
+  }
+  return true;
 }
```
vuvova, line 2503: eh, I kind of liked the old one here. It had a nice regular structure, easy to read.

sql/sql_acl.cc
```diff
@@ -903,7 +903,7 @@ class User_table: public Grant_table_base
   virtual int get_auth(THD *, MEM_ROOT *, ACL_USER *u) const= 0;
   virtual bool set_auth(const ACL_USER &u) const = 0;
   virtual privilege_t get_access() const = 0;
-  virtual void set_access(const privilege_t rights, bool revoke) const = 0;
+  virtual void set_access(const privilege_t rights) const = 0;
```
vuvova, line 940: No, I don't agree with this one. It wasn't hiding anything, it was **set**ting values of specific privilege bits or columns to the value specified by the second argument. It was a quite clear and well defined semantics. And `set()` already was idempotent, as far as I understand.

sql/sql_acl.cc
```diff
@@ -12182,8 +12183,9 @@ int mysql_alter_user(THD* thd, List<LEX_USER> &users_list)
   {
     LEX_USER* lex_user= get_current_user(thd, tmp_lex_user, false);
     if (!lex_user ||
-        replace_user_table(thd, tables.user_table(), lex_user, NO_ACL,
-                           false, false, true))
+        replace_user_table(thd, tables.user_table(), lex_user,
+                           NO_ACL, false,
+                           false, true))
```
vuvova, line 12251: I prefer less lines of code, not more. I've long noticed that the speed of reading the code is generally measured in "lines per minute" — that is, "less lines" is translated verbatim into the "less time to read".

so, ok, as you prefer. but if I stumble upon this fragment some day and I might fairly automatically reformat it back to occupy one line less.

sql/grant.h
```diff
@@ -94,7 +89,17 @@ class Grant_privilege
...
                        SELECT_LEX *sel,
                        privilege_t with_grant_option);
-  const List<LEX_COLUMN> & columns() const { return m_columns; }
+  void set_all_privileges()
+  {
+    m_object_privilege= GLOBAL_ACLS;
+    m_all_privileges= true;
+  }
+  /* Getters */
+  List<LEX_COLUMN> & columns() { return m_columns; }
```
vuvova, line 98: why did you remove `const` here?

sql/sql_acl.h
```diff
@@ -386,15 +386,16 @@ class Sql_cmd_grant_proxy: public Sql_cmd_grant
 };
 
 
-class Sql_cmd_grant_object: public Sql_cmd_grant, public Grant_privilege
+class Sql_cmd_grant_object: public Sql_cmd_grant
 {
 protected:
+  Grant_privilege &m_gp;
```
vuvova, line 377: pointer, please, it's not `const`.

sql/sql_yacc.yy
```diff
@@ -17224,8 +17224,9 @@ grant_privileges:
           object_privilege_list
         | ALL opt_privileges
           { 
-            if (!($$= new (thd->mem_root) Lex_grant_privilege(GLOBAL_ACLS, true)))
+            if (!($$= new (thd->mem_root) Lex_grant_privilege()))
               MYSQL_YYABORT;
+            $$->set_all_privileges();
```
vuvova, line 17229: why is it better than a constructor?

sql/sql_acl.cc
```diff
@@ -8055,10 +8071,11 @@ static int mysql_table_grant(THD *thd, TABLE_LIST *table_list,
     /* Create user if needed */
     error= copy_and_check_auth(Str, Str, thd) ||
            replace_user_table(thd, tables.user_table(), Str,
-                               NO_ACL, revoke_grant, create_new_users,
-                               MY_TEST(!is_public(Str) &&
-                                       (thd->variables.sql_mode &
-                                        MODE_NO_AUTO_CREATE_USER)));
+                              {NO_ACL, revoke_grant},
+                              create_new_users,
+                              MY_TEST(!is_public(Str) &&
```
vuvova, line 8082: you could've as well removed MY_TEST

sql/sql_acl.cc
```diff
@@ -8694,19 +8712,18 @@ static bool mysql_grant(THD *thd, const char *db, List<LEX_USER> &list,
...
     if (copy_and_check_auth(Str, tmp_Str, thd) ||
         replace_user_table(thd, tables.user_table(), Str,
-                           (!db ? rights : NO_ACL),
-                           revoke_grant, create_new_users,
-                           MY_TEST(!is_public(Str) &&
-                                   (thd->variables.sql_mode &
-                                    MODE_NO_AUTO_CREATE_USER))))
+                           {!db ? priv_spec.access : NO_ACL, priv_spec.revoke},
+                           create_new_users,
+                           !is_public(Str) && (thd->variables.sql_mode &
```
vuvova: oh! precisely :)

sql/sql_acl.cc
```diff
@@ -13087,11 +13117,34 @@ bool Sql_cmd_grant_proxy::execute(THD *thd)
     return true;
 
   WSREP_TO_ISOLATION_BEGIN(WSREP_MYSQL_DB, NULL, NULL);
-  /* Conditionally writes to binlog */
-  if (mysql_grant(thd, NULL/*db*/, lex->users_list, {m_grant_option, is_revoke()},
-                  true/*proxy*/))
+
+  if (tables.open_and_lock(thd, Table_user | Table_proxies_priv, TL_WRITE))
     return true;
```
vuvova, line 13129: old code had here
```c
   if ((result= tables.open_and_lock(thd, tables_to_open, TL_WRITE)))
     DBUG_RETURN(result != 1);
```
now you return an error when `result == 1`, this ignores the logic from the comment
```
     @retval  1 replication filters matched. Abort the operation,
                but return OK (!)
```

sql/sql_acl.cc
```diff
@@ -13196,11 +13257,75 @@ bool Sql_cmd_grant_table::execute_table_mask(THD *thd)
...
+
+    /* Only allow DB level grants. */
+    if ((m_gp.object_privilege() & DB_ACLS) != m_gp.object_privilege())
+    {
+        my_error(ER_WRONG_USAGE, MYF(0), "DB GRANT", "GLOBAL PRIVILEGES");
+        return true;
+    }
+
+    if (tables.open_and_lock(thd, Table_user | Table_db, TL_WRITE))
+      return true;
```
vuvova: here, again, you return an error for return value of 1. And also you've lost an assert
```c
   DBUG_ASSERT(!thd->is_current_stmt_binlog_format_row());
```

sql/sql_acl.cc
```diff
@@ -8651,110 +8652,134 @@ bool mysql_grant_role(THD *thd, List <LEX_USER> &list, bool revoke)
...
+    /* Resolve if it the proxy is "CURRENT_USER" or "CURRENT_ROLE" or
+       a username / rolename without a hostname specified. */
+    LEX_USER *resolved_user= get_current_user(thd, user, false);
+    if (!resolved_user)
+    {
+      result= true;
+      continue;
+    }
+    /* TODO(cvicentiu) This might actually not be necessary for grant_proxy.
+       Carry over auth information. */
```
vuvova: this comment is in `mysql_grant_db`, not in `mysql_grant_proxy`. looks out of place

sql/sql_acl.cc
```diff
@@ -8651,110 +8652,134 @@ bool mysql_grant_role(THD *thd, List <LEX_USER> &list, bool revoke)
...
     {
-      my_error(ER_WRONG_DB_NAME ,MYF(0), db);
-      DBUG_RETURN(TRUE);
+      result= true;
+      continue;
     }
-    my_casedn_str(files_charset_info, tmp_db);
-    db=tmp_db;
+    /* TODO(cvicentiu) This might actually not be necessary for grant_proxy.
+       Carry over auth information. */
```
vuvova: this comment is in `mysql_grant_global`, not in `mysql_grant_proxy`. looks out of place

sql/sql_acl.cc
```diff
@@ -8651,110 +8652,134 @@ bool mysql_grant_role(THD *thd, List <LEX_USER> &list, bool revoke)
...
+    }
+    /* TODO(cvicentiu) This might actually not be necessary for grant_proxy.
+       Carry over auth information. */
+    resolved_user->auth= user->auth;
+    if (check_if_auth_can_be_changed(resolved_user, thd) ||
+        replace_user_table(thd, tables.user_table(),
+                           resolved_user,
+                           {NO_ACL, priv_spec.revoke},
+                           create_new_users,
+                           !is_public(resolved_user) && no_auto_create_users) ||
```
vuvova: why not to move this (common) part to a separate function? find a user, copy auth, create a user if needed — all your new mysql_grant_* functions do that

vuvova: also the loop over users, propagation of role grants

sql/sql_acl.cc
```diff
@@ -8651,110 +8652,134 @@ bool mysql_grant_role(THD *thd, List <LEX_USER> &list, bool revoke)
...
+  DBUG_ENTER("mysql_grant_db");
+  mysql_mutex_assert_owner(&acl_cache->lock);
+  DBUG_ASSERT((priv_spec.access & DB_ACLS) == priv_spec.access);  // Only DB_ACLs are accepted.
+  /* go through users in user_list */
+  for (LEX_USER *user= it++; user != NULL; user= it++)
   {
-    DBUG_ASSERT(!db);
-    proxied_user= str_list++;
-  }
+    /* Resolve if it the proxy is "CURRENT_USER" or "CURRENT_ROLE" or
```
vuvova: "proxy"? it's `mysql_grant_db`

sql/sql_acl.cc
```diff
@@ -8651,110 +8652,134 @@ bool mysql_grant_role(THD *thd, List <LEX_USER> &list, bool revoke)
...
+  bool result= false;
+  DBUG_ENTER("mysql_grant_global");
 
-  if (lower_case_table_names && db)
+  /* go through users in user_list */
+  for (LEX_USER *user= it++; user != NULL; user= it++)
   {
-    char *end= strnmov(tmp_db,db, sizeof(tmp_db));
-    if (end >= tmp_db + sizeof(tmp_db))
+    /* Resolve if it the proxy is "CURRENT_USER" or "CURRENT_ROLE" or
```
vuvova: "proxy"? it's `mysql_grant_global`

sql/sql_acl.cc
vuvova: `Sql_cmd_grant_sp::execute` becomes quite different now, it doesn't assert, doesn't binlog, doesn't open tables. Unlike other `Sql_cmd_grant*::execute` methods

sql/sql_acl.cc
vuvova: as a summary, there's quite a lot of duplicated code now. I feel that it can be isolated even further

sql/sql_acl.cc
```diff
@@ -8247,43 +8246,36 @@ static bool mysql_routine_grant(THD *thd, TABLE_LIST *table_list,
 
   DBUG_PRINT("info",("now time to iterate and add users"));
 
-  while ((tmp_Str= str_list++))
+  for (LEX_USER *user= it++; user != NULL; user= it++)
```
vuvova, line 8253: by the way, `List<>` supports iterator interface, so you can use
```c++
for (auto &user: user_list)
{
```

sql/sql_acl.cc
```diff
@@ -13059,36 +13018,51 @@ void Sql_cmd_grant::warn_hostname_requires_resolving(THD *thd,
...
-  while ((tmp_user= it++))
+    This resolved list is needed within all grant ops so we store it as
+    a class member.
+  */
+  List_iterator<LEX_USER> it(thd->lex->users_list);
+  m_resolved_users.empty();
+  for (LEX_USER *tmp_user= it++; tmp_user != NULL; tmp_user= it++)
   {
-    if (!(user= get_current_user(thd, tmp_user)))
+    LEX_USER *resolved_user= get_current_user(thd, tmp_user);
```
vuvova, line 13069: ok, so you add an extra loop over users. Where do you eliminate a number of mallocs?

sql/sql_acl.cc
```diff
@@ -13145,7 +13143,67 @@ bool Sql_cmd_grant_object::grant_stage0_exact_object(THD *thd,
 }
 
 
-bool Sql_cmd_grant_table::execute_exact_table(THD *thd, TABLE_LIST *table)
+bool Sql_cmd_grant_table::execute_grant_database(THD *thd)
+{
+  Grant_tables tables;
+  bool result;
+  const bool no_auto_create_users= MY_TEST(thd->variables.sql_mode &
+                                           MODE_NO_AUTO_CREATE_USER);
```
vuvova, line 13184: why do you still keep `MY_TEST()` around?

sql/sql_acl.cc
```diff
@@ -13145,7 +13143,67 @@ bool Sql_cmd_grant_object::grant_stage0_exact_object(THD *thd,
...
+                         m_create_new_users, no_auto_create_users);
+
+  mysql_mutex_unlock(&acl_cache->lock);
+  /* Write to binlog if statement succeeded. */
+  if (!result)
+    result= write_bin_log(thd, TRUE, thd->query(), thd->query_length());
+  mysql_rwlock_unlock(&LOCK_grant);
+
+  if (!result)
+    my_ok(thd);
```
vuvova, line 13228: why not to put this tail into the caller? now you have it duplicated.
and the head is pretty much the same, except for `lower_case_table_names` part. you can move the head into the caller too

sql/sql_acl.cc
```diff
@@ -3173,18 +3144,39 @@ check_access(THD *thd, privilege_t want_access,
     DBUG_RETURN(FALSE);
   }
 
+  /*
+    Shortcut to not check database specific access rights. One caveat:
+      If we don't have a global SELECT privilege, we have to get the
+      database specific access rights to be able to handle queries of type
+      UPDATE t1 SET a=1 WHERE b > 0
+      In that case the shortcut won't work.
```
vuvova, line 3167: the original code didn't make any sense to me, so the refactoring preserving the same logic didn't help much :)
a better comment for this 20-year old code could've been
```c++
/*
  for UPDATE and DELETE the server only checks for the UPDATE_ACL
  (or DELETE_ACL) first, so want_access is UPDATE_ACL or
  DELETE_ACL here. later it'll check for SELECT_ACL if there old
  column values are used in the statement. to simplify this later
  check let's merge global and database privileges in *saved_priv
  now when global privileges alone don't include SELECT_ACL
*/
```
I think it's not a nice solution at all, more like a hack. Instead db privileges have to be consulted when they're actually needed (during the check for SELECT_ACL).

sql/item.cc
```diff
@@ -6201,16 +6201,16 @@ bool Item_field::fix_fields(THD *thd, Item **reference)
...
   {
-    const char *db, *tab;
-    db=  field->table->s->db.str;
-    tab= field->table->s->table_name.str;
     if (!(have_privileges= (get_column_grant(thd, &field->table->grant,
-                                             db, tab, field_name.str) &
+                                             field->table->s->db,
+                                             field->table->s->table_name,
+                                             field_name) &
                             VIEW_ANY_ACL)))
```
vuvova, line 6208: perhaps you can also move the assignment out of the `if()`? :innocent:

sql/sql_base.cc
```diff
@@ -8583,20 +8583,32 @@ insert_fields(THD *thd, Name_resolution_context *context, const char *db_name,
...
         DBUG_ASSERT(item->type() == Item::FIELD_ITEM);
         Item_field *fld= (Item_field*) item;
-        const char *field_db_name= field_iterator.get_db_name();
-        const char *field_table_name= field_iterator.get_table_name();
+        /*
+          TODO(cvicentiu) constructing these lex_cstrings could be avoided
+          if we updated all the users of field_iterator::get_xxx_name to
+          use LEX_CSTRING.
+        */
+        const LEX_CSTRING field_db_name= {
```
vuvova, line 8591: btw, generally, please, prefer `Lex_table_name` (for table and db names) and `Lex_ident` (for other identifiers) to `LEX_CSTRING`. They're binary compatible, you can pass the former where the latter is expected, but they carry additional metadata for the compiler and can be compared easier.

sql/share/errmsg-utf8.txt
```diff
@@ -3571,31 +3571,30 @@ ER_COLUMNACCESS_DENIED_ERROR 42000
...
-        nor "Illegal GRANT/REVOKE command; please consult the manual to see which privleges can be used"
-        norwegian-ny "Illegal GRANT/REVOKE command; please consult the manual to see which privleges can be used"
-        pol "Illegal GRANT/REVOKE command; please consult the manual to see which privleges can be used"
-        por "Comando GRANT/REVOKE ilegal. Por favor consulte no manual quais privilégios podem ser usados"
-        rum "Comanda GRANT/REVOKE ilegala. Consultati manualul in privinta privilegiilor ce pot fi folosite"
-        rus "Неверная команда GRANT или REVOKE. Обратитесь к документации, чтобы выяснить, какие привилегии можно использовать"
+        chi "非法的 GRANT/REVOKE/DENY 命令；请查阅文档查看可以使用哪些权限"
+        cze "Neplatný příkaz GRANT/REVOKE/DENY. Prosím, přečtěte si v manuálu, jaká privilegia je možné použít"
+        dan "Forkert GRANT/REVOKE/DENY kommando. Se i brugervejledningen hvilke privilegier der kan specificeres"
+        eng "Illegal GRANT/REVOKE/DENY command; please consult the manual to see which privileges can be used"
```
vuvova, line 3577: a bit too early, don't you think?

sql/sql_acl.cc
```diff
@@ -10796,12 +10796,55 @@ static bool show_table_and_column_privileges(THD *thd, const char *username,
...
+  append_identifier(thd, result, db, strlen(db));
+  result->append('.');
+  append_identifier(thd, result, routine_name, strlen(routine_name));
+  add_to_user(thd, result, username, *hostname, hostname);
+  if (routine_priv & GRANT_ACL)
+    result->append(STRING_WITH_LEN(" WITH GRANT OPTION"));
+
+  return false;
+}
+
```
vuvova, line 10855: why did you extract it?

sql/sql_acl.cc
```diff
@@ -3346,12 +3346,12 @@ static bool check_show_access(THD *thd, TABLE_LIST *table)
   case SCH_TRIGGERS:
   case SCH_EVENTS:
   {
-    const char *dst_db_name= table->schema_select_lex->db.str;
+    const LEX_CSTRING &dst_db_name= table->schema_select_lex->db;
 
-    DBUG_ASSERT(dst_db_name);
+    DBUG_ASSERT(dst_db_name.str);
```
vuvova, line 3351: may be `DBUG_ASSERT(dst_db_name.length);` ?

PR #603, opened 2018-02-11
MDEV-14296 wsrep sst common from bindir to libexecdir

grooverdan:
I submit this under the MCA.

PR #2706, opened 2023-07-20
MDEV-14296 wsrep sst common from bindir to libexecdir

LinuxJedi:
This is a port of @grooverdan PR #603 to MariaDB 11.2. It includes a fix for the comment made by @illuusio.

- [x] *The Jira issue number for this PR is: MDEV-14296*

## PR quality check

### DIFF
--- a/cmake/install_layout.cmake
+++ b/cmake/install_layout.cmake
@@ -45,6 +45,7 @@
 # - INSTALL_SYSCONF2DIR     (additional config files, e.g. /etc/mysql/conf.d)
 #
 # - INSTALL_LIBDIR          (directory with client end embedded libraries)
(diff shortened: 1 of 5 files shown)


LinuxJedi:
Removing the Foundation tag as this pre-dates Daniel being at the Foundation.

LinuxJedi:
Breaks Debian 10 packaging, need to fix that prior to this being reviewed.

LinuxJedi:
This actually needs a lot of refinement. I see several issues with the original PR. There are user executable binaries such as `mytop` that are going into `libexec` and some binaries that should be `libexec` going to `bin`.

We probably need to look on a case by case basis for this.

PR #2719, opened 2023-08-03
Convert user input to lower case to achive case independent behaviour from validate_reply Function

shalom2727:
Convert user input to lower case. After that we get case independent response from the validate_reply function. if tr binary not available or corrupted on targeted host, it will fallback to previous behavior mean case dependent behavior.

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check


LinuxJedi (changes requested):
Thank you for your contribution. I have a question, and then I'll get to the reason I'm requesting changes at this time:

What problem does this PR solve? The check already checks for lowercase, uppercase, and camel case. The only thing missing are cases such as "nO" or "yEs". Is this an issue to be solved?

Now, the reason I'm requesting changes is that you have your commit, and then you have done a merge commit. Instead, can you go back one commit (`git reset HEAD^ --hard`) and then do a `git rebase` on the latest 11.2 instead? Otherwise, the merge commits pollute the tree, making it more difficult to debug issues later on.

Additionally, it would help us out if you could fill in the rest of the PR template.

scripts/mysql_secure_installation.sh
```diff
@@ -207,7 +207,15 @@ validate_reply () {
 	reply=y
 	return $ret
     fi
-    case $1 in
+    #
+    # Convert user input to lower case. After that we get case independent response from 
+    # the validate_reply function. if tr binary not available or corrupted on targeted host,
+    # it will fallback to previous behaviour mean case dependent behaviour.   
+    yes_or_no=$(echo $1 | tr '[:lower:]' '[:upper:]')
```
ericherman, line 214: The comment says "# Convert user input to lower case." but the `tr` command translates lower-case characters to upper-case.

Thus, if I am reading this correctly, I believe the following `$yes_or_no` would never be equal to either "yes" or "no", which will simply set `$yes_or_no` to `$1` .... which is the same as `case $1 in` ....

vuvova:
> Thank you for your contribution. I have a question, and then I'll get to the reason I'm requesting changes at this time:
> 
> What problem does this PR solve? The check already checks for lowercase, uppercase, and camel case. The only thing missing are cases such as "nO" or "yEs". Is this an issue to be solved?

if one wants to handle "yES", I'd rather do
```diff
-        y|Y|yes|Yes|YES) reply=y ;;
+        [yY]|[yY][eE][sS]) reply=y ;;
```

LinuxJedi:
> > Thank you for your contribution. I have a question, and then I'll get to the reason I'm requesting changes at this time:
> > What problem does this PR solve? The check already checks for lowercase, uppercase, and camel case. The only thing missing are cases such as "nO" or "yEs". Is this an issue to be solved?
> 
> if one wants to handle "yES", I'd rather do
> 
> ```diff
> -        y|Y|yes|Yes|YES) reply=y ;;
> +        [yY]|[yY][eE][sS]) reply=y ;;
> ```

Good call, I was thinking that there had to be a simpler way, I didn't consider that way.

scripts/mysql_secure_installation.sh
LinuxJedi: You are 100% correct here, good catch I didn't dig too much into the implementation details because I'm uncertain if this is a problem that needs to be solved.

ericherman: I'm not sure I understand, as even "YES" and "NO" are not equal to "yes" or "no" and thus `$yes_or_no` would be reset to $1 .... however ... it really doesn't matter! If we wish to be fully case-insensitive, @vuvova presented a more obvious way to do so.

shalom2727:
> Thank you for your contribution. I have a question, and then I'll get to the reason I'm requesting changes at this time:
> 
> What problem does this PR solve? The check already checks for lowercase, uppercase, and camel case. The only thing missing are cases such as "nO" or "yEs". Is this an issue to be solved?
> 
> Now, the reason I'm requesting changes is that you have your commit, and then you have done a merge commit. Instead, can you go back one commit (`git reset HEAD^ --hard`) and then do a `git rebase` on the latest 11.2 instead? Otherwise, the merge commits pollute the tree, making it more difficult to debug issues later on.
> 
> Additionally, it would help us out if you could fill in the rest of the PR template.

Yes you are right , i just want to pass every possible combination of case like yes, YeS, yEs etc.

scripts/mysql_secure_installation.sh
shalom2727: > The comment says "# Convert user input to lower case." but the `tr` command translates lower-case characters to upper-case.
> 
> Thus, if I am reading this correctly, I believe the following `$yes_or_no` would never be equal to either "yes" or "no", which will simply set `$yes_or_no` to `$1` .... which is the same as `case $1 in` ....

Yes you are right , thanks for pointing out problem in my code

shalom2727:
> > Thank you for your contribution. I have a question, and then I'll get to the reason I'm requesting changes at this time:
> > What problem does this PR solve? The check already checks for lowercase, uppercase, and camel case. The only thing missing are cases such as "nO" or "yEs". Is this an issue to be solved?
> 
> if one wants to handle "yES", I'd rather do
> 
> ```diff
> -        y|Y|yes|Yes|YES) reply=y ;;
> +        [yY]|[yY][eE][sS]) reply=y ;;
> ```

I think it is much better solution. i request please create it and merge on 11.2 branch. i am closing pull request.

PR #2720, opened 2023-08-04
MDEV-31837: Server crashes in  process_i_s_table_temporary_tables/get_all_tables after alter in rename query

an3l:
- Open temporary tables in rename, in order to find it, since
during failed alter it may be marked to be reopened and removed from
share.
- Reviewer: <>

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -6146,7 +6146,11 @@ static TABLE *find_temporary_table_for_rename(THD *thd,
     }
   }
   if (!found)
+  {
+    if (thd->has_temporary_tables())
+      thd->open_temporary_table(table);
     res= thd->find_temporary_table(table, THD::TMP_TABLE_ANY);
+  }
   DBUG_RETURN(res);
 }
 
--- a/mysql-test/main/information_schema_temp_table.result
+++ b/mysql-test/main/information_schema_temp_table.result
@@ -266,3 +266,56 @@ mysqltest	s1
 DROP TABLE seq1;
 DROP TABLE mysqltest.s1;
 DROP TABLE mysqltest.s2;
+#
+# MDEV-31618: Server crashes in
+#             process_i_s_table_temporary_tables/get_all_tables after alter in rename
+#
+CREATE  table seq1 (a bigint, b int);
+CREATE  TEMPORARY TABLE tmp LIKE seq1;
+INSERT  tmp SELECT * FROM seq1;
+ALTER  TABLE tmp RENAME TO seq1;
+ALTER TABLE seq1 CHANGE a b int ;
+Got one of the listed errors
+RENAME  TABLE seq1 TO seq1;
+ERROR 42S01: Table 'seq1' already exists
+show  full tables;
+Tables_in_test	Table_type
+seq1	TEMPORARY TABLE
+seq1	BASE TABLE
+drop table seq1;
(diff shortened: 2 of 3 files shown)


sql/temporary_tables.cc
```diff
@@ -795,7 +795,8 @@ void THD::mark_tmp_table_as_free_for_reuse(TABLE *table)
 
   table->query_id= 0;
   table->file->ha_reset();
-
+  if (unlikely(table->needs_reopen()) && table->s->table_type == TABLE_TYPE_SEQUENCE)
```
an3l, line 798: Without checking for sequences we get the failure
```
+++ /home/anel/GitHub/mariadb/server/src/11.2/mysql-test/main/alter_table_errors.reject	2023-08-06 10:38:03.508451890 +0200
@@ -15,7 +15,7 @@
 Table	Create Table
 t1	CREATE TEMPORARY TABLE `t1` (
   `a` int(11) DEFAULT NULL,
-  `v` int(11) GENERATED ALWAYS AS (`a`) VIRTUAL
+  `v` int(11) GENERATED ALWAYS AS (`?`) VIRTUAL
 ) ENGINE=MyISAM DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
 create temporary table t2 (a int, v int as (a));
 lock table t2 write;
@@ -25,7 +25,7 @@
 Table	Create Table
 t2	CREATE TEMPORARY TABLE `t2` (
   `a` int(11) DEFAULT NULL,
-  `v` int(11) GENERATED ALWAYS AS (`a`) VIRTUAL
+  `v` int(11) GENERATED ALWAYS AS (`?`) VIRTUAL
 ) ENGINE=MyISAM DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
 drop temporary table t1, t2;
 create temporary table t1 (a int);
```
and 
```
--- /home/anel/GitHub/mariadb/server/src/11.2/mysql-test/main/temp_table.result2023-08-02 16:36:05.578615581 +0200
+++ /home/anel/GitHub/mariadb/server/src/11.2/mysql-test/main/temp_table.reject2023-08-06 10:38:02.664453558 +0200
@@ -595,8 +595,8 @@
 UPDATE t1 SET b = 3;
 SELECT * FROM t1;
 a	b
-2001-01-01 10:20:30	3
-2001-01-01 10:20:30	3
+NULL	3
+NULL	3
 DROP TEMPORARY TABLE t1;
```
Is this related to the [MDEV-31853](https://jira.mariadb.org/browse/MDEV-31853) ?

sql/sql_parse.cc
```diff
@@ -6146,7 +6146,11 @@ static TABLE *find_temporary_table_for_rename(THD *thd,
     }
   }
   if (!found)
+  {
+    if (thd->has_temporary_tables())
+      thd->open_temporary_table(table);
     res= thd->find_temporary_table(table, THD::TMP_TABLE_ANY);
+  }
   DBUG_RETURN(res);
```
vuvova, line 6154: I think it's the wrong place to fix it. The rule is that there cannot be a `TMP_TABLE_SHARE` without a single `TABLE`. It's violated not here, it's violated in `THD::find_temporary_table`, so I believe it must be fixed there.

vuvova (changes requested):

PR #2721, opened 2023-08-07
MDEV-31833 replication breaks when using optimistic replication and r…

sjaakola:
…eplica is a galera node

MariaDB async replication SQL thread was stopped for any failure in applying of replication events and error message logged for the failure was: "Node has dropped from cluster". The assumption was that event applying failure is always due to node dropping out.
With optimistic parallel replication, event applying can fail for natural reasons and applying should be retried to handle the failure. This retry logic was never exercised because the slave SQL thread was stopped with first applying failure.

To support optimistic parallel replication retrying logic this commit will now skip replication slave abort, if node remains in cluster (wsrep_ready==ON) and replication is configured for optimistic or aggressive retry logic.

During the development of this fix, galera.galera_as_slave_nonprim test showed some problems. The test was analyzed, and it appears to need some attention. One excessive sleep command was removed in this commit, but it will need more fixes still to be fully deterministic. After this commit galera_as_slave_nonprim is successful, though.

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check

### DIFF
--- a/sql/rpl_parallel.cc
+++ b/sql/rpl_parallel.cc
@@ -819,6 +819,10 @@ do_retry:
   event_count= 0;
   err= 0;
   errmsg= NULL;
+#ifdef WITH_WSREP
+  thd->wsrep_cs().reset_error();
+  WSREP_DEBUG("retrying async replication event");
+#endif /* WITH_WSREP */
 
   /*
     If we already started committing before getting the deadlock (or other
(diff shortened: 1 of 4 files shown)


sql/rpl_parallel.cc
```diff
@@ -918,6 +922,7 @@ retry_event_group(rpl_group_info *rgi, rpl_parallel_thread *rpt,
       err= rgi->worker_error= 1;
       my_error(ER_PRIOR_COMMIT_FAILED, MYF(0));
       mysql_mutex_unlock(&entry->LOCK_parallel_entry);
+
```
knielsen, line 925: Please avoid unrelated formatting changes to the existing code, as they complicate merges and history tracking.

sysprg:
Thanks, merged with head revision: https://github.com/MariaDB/server/commit/a3cbc44b24ec467f33e445f57e2022e038b88623

PR #2724, opened 2023-08-09
use special executable prefix for Debian

eworm-de:
The default systemd service file set `PermissionsStartOnly=true`, which makes everything but `ExecStart=` (including `ExecStartPre=`, `ExecStartPost=`, etc) run as user root. This is not necessary.

Looks like the reason for this is an addition for Debian, that adds an extra call to `/usr/bin/install` to create `@MYSQL_UNIX_DIR@`.

Let's clean things up... Drop the configuration from default service file, then make the specific `ExecStartPre=` use special executable prefix for full privileges. (See systemd.service(5) for details.)

## PR quality check


eworm-de:
CC @LinuxJedi 
This is to (hopefully) resolve the failing bits from #1143 first.

Is there a way to trigger a Debian CI run on this?

eworm-de:
Oh, hold on... We can not do this independent from Galera changes...

eworm-de:
Closing, this is now contained in #2726.

PR #2736, opened 2023-08-24
MDEV-27247: Add keywords "SQL_BEFORE_GTIDS" and "SQL_AFTER_GTIDS" for START SLAVE UNTIL

bnestere:
This PR extends the START SLAVE UNTIL command with options
SQL_BEFORE_GTIDS and SQL_AFTER_GTIDS to allow user control of
whether the replica stops before or after a provided GTID state. Its
syntax is:

START SLAVE UNTIL (SQL_BEFORE_GTIDS|SQL_AFTER_GTIDS)=”<gtid_list>”

When providing SQL_BEFORE_GTIDS=”<gtid_list>”, the replica will
execute all transactions up to the first GTID found in the provided
list, and stop immediately. In contrast to the default behavior of
UNTIL, this will execute transactions from all domains on the
primary until the replica stops due to seeing a GTID on the list.

START SLAVE UNTIL SQL_AFTER_GTIDS=”<gtid_list>” is an alias to the
default behavior of START SLAVE UNTIL master_gtid_pos=”<gtid_list>”.
That is, the replica will only execute transactions originating from
domain ids provided in the list, and will stop once all transactions
provided in the UNTIL list have all been executed.

Example:
=========
If a primary server has a binary log consisting of the following GTIDs:

0-1-1
1-1-1
0-1-2
1-1-2
0-1-3
1-1-3

If a replica is started with SQL_BEFORE_GTIDS, i.e.
START SLAVE UNTIL SQL_BEFORE_GTIDS=”1-1-2”

The resulting gtid_slave_pos of the replica will be “0-1-2,1-1-1”.
This is because the replica will execute all events until it sees
the transaction with GTID 1-1-2 and immediately stop without
executing it.

However, if a replica is started with SQL_AFTER_GTIDS, i.e.
START SLAVE UNTIL SQL_AFTER_GTIDS=”1-1-2”

then the resulting gtid_slave_pos of the replica will be “1-1-2”.
This is because it will only execute events from domain 1 until it
has executed the provided GTID.

The initial commits are ordered as follows:
1. The MTR test for the feature
2. Extension to the code base 
3. Fixes for existing MTR tests which the patch broke.


sql/sql_repl.cc
```diff
@@ -1861,7 +1874,7 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
       if (until_gtid_state)
       {
         gtid= until_gtid_state->find(event_gtid.domain_id);
-        if (gtid == NULL)
+        if (gtid == NULL && !info->is_until_before_gtids)
```
knielsen, line 1877: Minor style suggestion:
I would add the new condition inside an extra if(!info->is_until_before_gtids){...} ...

sql/sql_repl.cc
```diff
@@ -1870,7 +1883,8 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
           info->gtid_skip_group = (flags2 & Gtid_log_event::FL_STANDALONE ?
                               GTID_SKIP_STANDALONE : GTID_SKIP_TRANSACTION);
         }
-        else if (event_gtid.server_id == gtid->server_id &&
+        else if (gtid != NULL &&
+                 event_gtid.server_id == gtid->server_id &&
```
knielsen, line 1887: ... so that you avoid a double check for NULL here.

Eg.
  if (gtid==NULL) {
    if (!info->is_until_before_gtids) {
    ...
    }
} else if (event_gtid.server_id == gtid->server_id &&
                 event_gtid.seq_no >= gtid->seq_no) {
  ...

mysql-test/suite/rpl/include/rpl_gtid_until_before_gtids.test
```diff
@@ -0,0 +1,329 @@
...
+--inc $t1_ctr
+--eval INSERT INTO t1 VALUES ($t1_ctr)
+--inc $t1_ctr
+
+
+--echo # Skip a seq_no
+--let $binlog_pos= `SELECT @@gtid_binlog_pos`
+--let $domain_id= `SELECT @@gtid_domain_id`
+--let $server_id= `SELECT @@server_id`
+--let $last_seq_no=  `SELECT REGEXP_SUBSTR(REGEXP_SUBSTR('$binlog_pos','[0-9]+\\\$'),'[0-9]+')`
```
knielsen, line 136: Do you need the double REGEXP_SUBSTR() here? I think the outer one should be redundant, just matching /[0-9]+$/ should be enough?

mysql-test/suite/rpl/include/rpl_gtid_until_before_gtids.test
```diff
@@ -0,0 +1,329 @@
...
+    # Will have GTIDs for both domains 0 and 1
+    --let $expected_stop_gtid= `SELECT @@gtid_binlog_pos`
+}
+set session gtid_domain_id=0;
+--eval INSERT INTO t1 VALUES ($t1_ctr)
+--inc $t1_ctr
+
+--let $binlog_pos= `SELECT @@gtid_binlog_pos`
+# Just the GTID for domain 0
+--let $until_gtid=  `SELECT REGEXP_SUBSTR('$binlog_pos','0-[0-9]-[0-9]+')`
```
knielsen, line 204: I'd put a "+" in the match aginst server-id here, '0-[0-9]+-[0-9]+', just for consistency.
Functionally it doesn't matter of course since server-id is always 1.

knielsen (approved):
Hi Brandon,

Looks good, thanks! Just a handful of minor comments.

I like the commit message description.
I suggest to add in the example "Assume the slave starts with the empty GTID
position, @@gtid_slave_pos=''", or similar (just to be precise about which
position will be reached after stop).

Test case looks good too, I think it's good that it --sources the include file twice once for "before" and once for "after". I suggest to rename the files to reflect this fact, eg. "rpl_gtid_until_before_after_gtids.*" or similar.

mysql-test/suite/rpl/include/rpl_gtid_until_before_gtids.test
bnestere: Nope; my test wasn't working and during debugging, I wondered if an EoL delimiter was invalidly kept, but that wasn't the case, and I forgot to clean it up. Thanks!

sql/sql_repl.cc
bnestere: Oh nice! I like that :)

mysql-test/suite/perfschema/r/digest_view.result
```diff
@@ -191,17 +191,17 @@ SELECT SCHEMA_NAME, DIGEST, DIGEST_TEXT, COUNT_STAR
...
+test	4a6602212e0e1386dada25a12dc2315f	EXPLAIN SELECT * FROM `test` . `v1` WHERE `a` = ? 	1
+test	7b3a13b2c268ba0a72aade33fb3b9d7d	EXPLAIN SELECT * FROM `test` . `v1` WHERE `b` > ? 	1
+test	840a987076e3bd7b25603240595f47b9	EXPLAIN SELECT `a` , `b` FROM `test` . `v1` 	1
+test	955679ae1c2068ade1003c820f1cdd58	EXPLAIN SELECT `b` , `a` FROM `test` . `v1` 	1
+test	3fe51e497b828a521e0b404f7c0e0ac6	SELECT * FROM `test` . `v1` 	1
+test	2bec69e086533e335e311b159eee104a	SELECT * FROM `test` . `v1` WHERE `a` = ? 	1
+test	4e995969bf7999998047ed967d8ebf45	SELECT * FROM `test` . `v1` WHERE `b` > ? 	1
+test	51f783c563f3af0cdca3065baace1236	SELECT `a` , `b` FROM `test` . `v1` 	1
+test	be34ebc622c2d81afca10cc511a06af9	SELECT `b` , `a` FROM `test` . `v1` 	1
+test	af6cd26b7f88eafb1ff05602a49ce39e	TRUNCATE TABLE `performance_schema` . `events_statements_summary_by_digest` 	1
```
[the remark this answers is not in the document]
knielsen: Ah, I see. Yeah, I suppose it's fine, I was mostly just curious.

PR #2737, opened 2023-08-24
Remove duplicated default client include from replication my.cnf

an3l:
- `default_client` is included already in `rpl_1slave_base.cnf`, so remove it from `my.cnf`
- Remove option group for `mysqld` server as and add comment how to override specific settings for specific server

- Reviewer: <brandon.nesterenko@mariadb.com>

- Dependency graph (red component in the diagram is related to the PR):

image

### DIFF
--- a/mysql-test/suite/rpl/my.cnf
+++ b/mysql-test/suite/rpl/my.cnf
@@ -1,8 +1,8 @@
 # Use settings from rpl_1slave_base.cnf
 # add setting to connect the slave to the master by default
 !include rpl_1slave_base.cnf
-!include include/default_client.cnf
-
-
-[mysqld.2]
-
+# Override specific server settings using [mariadb-x.y] option group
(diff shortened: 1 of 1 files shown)


mysql-test/suite/rpl/my.cnf
```diff
@@ -1,8 +1,4 @@
 # Use settings from rpl_1slave_base.cnf
 # add setting to connect the slave to the master by default
 !include rpl_1slave_base.cnf
-!include include/default_client.cnf
-
-
-[mysqld.2]
-
+# Override specific server settings using [mariadb-x.y] option group from test.cnf file
```
bnestere, line 4: If we're going to add this comment snippet to mention how to override (which I like!), we should probably be more complete about it. That is, to briefly define `x` and `y`, and also add a little example

bnestere (approved):
Thanks Anel for another good cleanup :) See my note.

PR #2738, opened 2023-08-25
MDEV-29861 : Galera "notify" test cases hang

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-29861*

## Description

Problem was that if wsrep_notify_cmd was set it was called with a new status "joined" it tries to connect to the server to update some table, but the server isn't initialized yet, it's not listening for connections. So the server waits for the script to finish, script waits for mariadb client to connect, and the client cannot connect, because the server isn't listening.

Fix is to call script only when Galera has already formed a view or when it is synched or donor.

This fix also enables following test cases:
* galera.MW-284
* galera.galera_binlog_checksum
* galera_var_notify_ssl_ipv6

## How can this PR be tested?

## PR quality check


sysprg:
Thanks, merged with head revision: https://github.com/MariaDB/server/commit/632a503ce7726547d8d858ae89fc3833ff95fe83

PR #2739, opened 2023-08-26
MDEV-32018 Allow the setting of Auto_increment on FK referenced columns

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32018*

## Description

In MDEV-31086, SET FOREIGN_KEY_CHECKS=0 cannot bypass checks that make column types of foreign keys incompatible. An unfortunate consequence is that adding an AUTO_INCREMENT is considered incompatible in Field_{num,decimal}::is_equal and for the purpose of FK checks this isn't relevant.

The reason for the auto_increment check in is_equal seems to resolve mysql bug #14573 which still affects Aria and MyISAM storage engines but not InnoDB.

## How can this PR be tested?

mtr test included.

- [*] *This is a feature addition for a limited set of functionality that was removed in a bug fix and hence cased a regression for @horrockss.

## PR quality check
- [*] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.

### DIFF
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -9135,8 +9135,7 @@ enum fk_column_change_type
   @retval FK_COLUMN_NO_CHANGE    No significant changes are to be done on
                                  foreign key columns.
   @retval FK_COLUMN_DATA_CHANGE  ALTER TABLE might result in value
-                                 change in foreign key column (and
-                                 foreign_key_checks is on).
+                                 change in foreign key column.
   @retval FK_COLUMN_RENAMED      Foreign key column is renamed.
   @retval FK_COLUMN_DROPPED      Foreign key column is dropped.
 */
@@ -9172,7 +9171,18 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
         return FK_COLUMN_RENAMED;
       }
 
-      if ((old_field->is_equal(*new_field) == IS_EQUAL_NO) ||
+      /*
+        Field_{num|decimal}::is_equal evaluates to IS_EQUAL_NO where
+        the new_field adds an AUTO_INCREMENT flag on a column due to a
+	limitation in MyISAM/ARIA. For the purposes of FK determination
+        it doesn't matter if AUTO_INCREMENT is there or not.
+      */
+      const uint flags= new_field->flags;
+      new_field->flags&= ~AUTO_INCREMENT_FLAG;
+      const bool equal_result= old_field->is_equal(*new_field);
+      new_field->flags= flags;
+
+      if ((equal_result == IS_EQUAL_NO) ||
           ((new_field->flags & NOT_NULL_FLAG) &&
            !(old_field->flags & NOT_NULL_FLAG)))
       {
--- a/mysql-test/suite/innodb/r/foreign_key.result
+++ b/mysql-test/suite/innodb/r/foreign_key.result
@@ -892,4 +892,19 @@ DROP INDEX fx ON t1;
 INSERT INTO t1 VALUES (2,11,11);
 DROP TABLE t1;
 SET FOREIGN_KEY_CHECKS=DEFAULT;
+#
+# MDEV-32018 Allow the setting of Auto_increment on FK referenced columns
+#
+CREATE TABLE t1 (
+id int unsigned NOT NULL PRIMARY KEY
+);
+CREATE TABLE t2 (
+id int unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+t1_id int unsigned DEFAULT NULL,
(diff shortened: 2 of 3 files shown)


mysql-test/main/auto_increment.test
```diff
@@ -433,3 +439,28 @@ SET @engine='MyISAM';
 --echo # End of 10.2 tests
 --echo #
 
+--echo #
+--echo # Start of 10.4 tests
```
dr-m, line 443: I don’t think that a "Start of" marker adds any value. The value of the "End of" marker is that it helps avoid merge conflicts.

mysql-test/main/auto_increment.test
```diff
@@ -433,3 +439,28 @@ SET @engine='MyISAM';
...
+  id int unsigned NOT NULL PRIMARY KEY
+);
+
+CREATE TABLE t2 (
+  id int unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  t1_id int unsigned DEFAULT NULL,
+  CONSTRAINT FK_t1_id FOREIGN KEY (t1_id) REFERENCES t1 (id)
+);
+
+ALTER TABLE t1 MODIFY id INT unsigned AUTO_INCREMENT;
```
dr-m, line 460: This test does not appear to enable InnoDB. Therefore, the foreign key constraint should be ignored by whatever is the default storage engine.

I would suggest to add a test that uses explicit `ALGORITHM=INPLACE` so that the error message that was added to `ha_innobase::check_if_supported_inplace_alter()` will be covered.

dr-m (commented):
The code change looks reasonable to me.

mysql-test/main/auto_increment.test
grooverdan: wasn't reachable so left as commit.

mysql-test/main/auto_increment.test
grooverdan: ack

sql/sql_table.cc
```diff
@@ -9139,17 +9138,27 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
...
       }
 
+      /*
+        Field_{num|decimal}::is_equal evaluates to IS_EQUAL_NO where
+        the new_field adds an AUTO_INCREMENT flag on a column due to a
+	limitation in MyISAM/ARIA. For the purposes of FK determination
+        it doesn't matter if AUTO_INCREMENT is there or not.
+      */
+      const auto flags= new_field->flags;
+      new_field->flags&= ~AUTO_INCREMENT_FLAG;
```
FooBarrior, line 9148: Can you please reorganize the code this way:
```
      const uint flags= new_field->flags;
      new_field->flags&= ~AUTO_INCREMENT_FLAG;
      const bool equal_result= old_field->is_equal(*new_field);
      new_field->flags= flags;

      if ((equal_result == IS_EQUAL_NO) || ...
```

Then you won't need to reset flags twice.

sql/sql_table.cc
```diff
@@ -9139,17 +9138,27 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
         return FK_COLUMN_RENAMED;
       }
 
+      /*
+        Field_{num|decimal}::is_equal evaluates to IS_EQUAL_NO where
+        the new_field adds an AUTO_INCREMENT flag on a column due to a
+	limitation in MyISAM/ARIA. For the purposes of FK determination
+        it doesn't matter if AUTO_INCREMENT is there or not.
+      */
+      const auto flags= new_field->flags;
```
FooBarrior, line 9147: He had an agreeement with serg not to use `auto` in simple cases , like integers, for example.

I know it's not in coding style yet, but I would like to add it (as well other c++-related style elements that we're used to follow. We have none covered in current doc).

mysql-test/main/auto_increment.test
```diff
@@ -433,3 +439,24 @@ SET @engine='MyISAM';
...
 
+--echo #
+--echo # MDEV-32018 Allow the setting of Auto_increment on FK referenced columns
+--echo #
+
+CREATE TABLE t1 (
+  id int unsigned NOT NULL PRIMARY KEY
+);
+
+CREATE TABLE t2 (
```
FooBarrior, line 450: This file doesn't use innodb at all. I think that better to move the test somewhere into `innodb` suite. A good pick can be `foreign_key.test`. I check other options like alter and autoinc there in the suite, but they are mostly debug-only, unlike `foreign_key`.

FooBarrior (changes requested):

FooBarrior:
Thank you! While a hash, it will do. Quick and easy.

FooBarrior (approved):

dr-m (approved):

PR #2740, opened 2023-08-28
MDEV-31303 - Key not used when IN clause has both signed and unsigned values

cvicentiu:
This patch enables possible optimizations for the query optimizer when the where clause has an IN condition of the form:

`expr_bigint IN (<[signed|unsigned] int>, <[signed|unsigned] int>)` etc.

The current code would check each pair between the expr and the arguments in the IN clause. For a case such as:

`CAST(123 AS UNSIGNED BIGINT) IN (1, LONGLONG_MAX + 1)` the comparison between
* `123` and `1` produced a `new_decimal` type handler because 1 was seen as a `signed long long`,
* `123` and `LONGLONG_MAX + 1` produced an `unsigned long long` type handler comparator because the types matched.

This lead to the optimizer considering that the types in the IN clause were not compatible, hence an INDEX range lookup was not considered possible. This is however false, as the IN list can be upcasted to UNSIGNED or NEW DECIMAL and it can be compared.

This patch introduces a check for this sort of case, forcing a complete upcast to NEW DECIMAl for all INT columns, regardless if their signness matches.

- [x] *The Jira issue number for this PR is: MDEV-31303*
## How can this PR be tested?
Test cases updated.

I would like to have this in 10.4, but first let's agree if the patch is good enough.


abarkov (commented):
Hi Vicențiu. Thanks for the patch!
I propose to modify the patch slightly:
- There is no a need to run through aggregation of all (predicant,valueN) pairs again. We can analyze the state of `Predicant_to_list_comparator` after the `prepare_predicant_and_values()` call, then just iterate through all arguments to check if they are all integers.
- in_longlong supports a mixture of integers of different signess. There is no a need to use in_decimal.

Please find a patch implementing these idea here:
https://github.com/MariaDB/server/commit/6f43b292a4c35ae6e08d6f5dfbd3cd46494076e3

Thanks.

https://github.com/MariaDB/server/commit/6f43b292a4c35ae6e08d6f5dfbd3cd46494076e3

abarkov:
It also applies as is so 10.5:

https://github.com/MariaDB/server/commit/3096cdb57b3c5b3c0c38e4538186f5c6f5fc4d03

Backporting to 10.4 would need some additional changes.

PR #2741, opened 2023-08-28
MDEV-31177: SHOW SLAVE STATUS Last_SQL_Errno Race Condition on Errored Slave Restart

bnestere:
The SQL thread and a user connection executing SHOW SLAVE STATUS
have a race condition on Last_SQL_Errno, such that a slave which
previously errored and stopped, on its next start, SHOW SLAVE STATUS
can show that the SQL Thread is running while the previous error is
also showing.

The fix is to move when the last error is cleared when the SQL
thread starts to occur before setting the status of
Slave_SQL_Running.

The PR commits are as follows:
 1. Cleanup of test rpl_xa_prepare_gtid_fail, unrelated to the bug
 2. Extend rpl_xa_prepare_gtid_fail with a test case which shows the presence of the bug
 3. The patch which fixes the bug

The remainder of the commits are to address reviews.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_xa_prepare_gtid_fail.test
+++ b/mysql-test/suite/rpl/t/rpl_xa_prepare_gtid_fail.test
@@ -19,6 +19,8 @@
 source include/master-slave.inc;
 source include/have_binlog_format_row.inc;
 source include/have_innodb.inc;
+source include/have_debug.inc;
+source include/have_debug_sync.inc;
 
 --connection slave
 --source include/stop_slave.inc
@@ -26,6 +28,7 @@ source include/have_innodb.inc;
 --let $save_par_thds= `SELECT @@global.slave_parallel_threads`
 --let $save_strict_mode= `SELECT @@global.gtid_strict_mode`
 --let $save_innodb_lock_wait_timeout= `SELECT @@global.innodb_lock_wait_timeout`
+--let $save_debug= `SELECT @@global.debug_dbug`
 
 change master to master_use_gtid=slave_pos;
 set @@global.slave_parallel_threads= 4;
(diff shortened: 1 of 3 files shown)


knielsen (approved):

mysql-test/suite/rpl/t/rpl_xa_prepare_gtid_fail.test
```diff
@@ -25,9 +25,9 @@ source include/have_debug_sync.inc;
 --connection slave
```
andrelkin, line 25: While we're here, let's shift l.19 `master-slave.inc` to the end of the sourced inc:s.

mysql-test/suite/rpl/t/rpl_xa_prepare_gtid_fail.test
```diff
@@ -25,9 +25,9 @@ source include/have_debug_sync.inc;
...
 --source include/stop_slave.inc
 
---let $save_par_thds= `SELECT @@global.slave_parallel_threads`
---let $save_strict_mode= `SELECT @@global.gtid_strict_mode`
---let $save_innodb_lock_wait_timeout= `SELECT @@global.innodb_lock_wait_timeout`
+set @save_par_thds= @@global.slave_parallel_threads;
```
andrelkin, line 28: As a remark, I always thought of a macro which would assign to a new value while save in background into magic "save_" list of items so that the eventual `rpl_end.inc` would restrore from them.

andrelkin (approved):
The patch looks nice.  Thanks!

bnestere:
I realized the bug itself is present in 10.4, so I ported the test/code to 10.4 (into [bb-10.4-MDEV-31177](https://github.com/MariaDB/server/tree/bb-10.4-MDEV-31177), if you want to look at it before I push); and then I'll still push the test cleanup into 10.5.

knielsen:
Sounds ok Brandon, thanks!

bnestere:
Closing because the 31177 changes were put into 10.4 as [1407f9996](https://github.com/MariaDB/server/commit/1407f99963481d2c59c109642a6ed81a1ae201b4) and the test code cleanup were pushed as [7de0c7b56](https://github.com/MariaDB/server/commit/7de0c7b569af9120d8df38f7bf5931853901a3b2)

PR #2744, opened 2023-08-31
MDEV-32043 mariadb-upgrade should remove bundled plugins from mysql.plugin

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32043*

## Description

Also in the startup, lets not "Error" on attempting to install a mysql.plugin that is already there. We use the 'if_not_exists' parameter to true to downgrade this to a "Note".

## How can this PR be tested?

MTR updated + MDEV shows the case for error log.

## PR quality check


sql/sql_plugin.cc
```diff
@@ -1930,7 +1930,7 @@ static void plugin_load(MEM_ROOT *tmp_root)
       the mutex here to satisfy the assert
     */
     mysql_mutex_lock(&LOCK_plugin);
-    plugin_add(tmp_root, false, &name, &dl, MYF(ME_ERROR_LOG));
+    plugin_add(tmp_root, true, &name, &dl, MYF(ME_ERROR_LOG));
```
vuvova, line 1933: I don't think that's a good idea

grooverdan: To have plugin_load_add like `INSTALL PLUGIN IF NOT EXISTS`? I think its a good idea.

vuvova: Okay. I actually cannot think of any problem it could possibly cause :innocent:

vuvova: still, I'm doubtful. `DROP TABLE` doesn't always imply `IF NOT EXISTS`, and I also cannot think of any problem it could possibly cause. So, at least, plugin behavior will be inconsistent.

May be, let's only imply `IF NOT EXISTS` on startup? I mean, when loading from a table, but not from an explicit `INSTALL PLUGIN` command.

---
Okay. I have come up with a reason why implying `IF NOT EXISTS` is harmful. When a user runs `INSTALL PLUGIN` they have all the reasons to expect that a new plugin is installed and will work. If FederatedX cannot be installed because Federated already is, if an old version of the plugin is marked for deletion, but still used, in all these cases an error will indicate that `INSTALL PLUGIN` has failed and the new plugin is not in force. You cannot just swallow this error up.

grooverdan: "INSTALL PLUGIN" calls plugin add:

```
Thread 14 "mariadbd" hit Breakpoint 2, plugin_add (tmp_root=0x7f8694006a50, if_not_exists=false, name=name@entry=0x7f8694005c98, dl=dl@entry=0x7f86cc1c5890, MyFlags=MyFlags@entry=0)
    at /home/dan/repos/mariadb-server-10.5/sql/sql_plugin.cc:1115
1115	  if (name->str && plugin_find_internal(name, MYSQL_ANY_PLUGIN))
(gdb) bt
#0  plugin_add (tmp_root=0x7f8694006a50, if_not_exists=false, name=name@entry=0x7f8694005c98, dl=dl@entry=0x7f86cc1c5890, MyFlags=MyFlags@entry=0)
    at /home/dan/repos/mariadb-server-10.5/sql/sql_plugin.cc:1115
#1  0x0000000000709c67 in mysql_install_plugin (thd=thd@entry=0x7f8694000f38, name=0x7f8694005c98, dl_arg=<optimized out>) at /home/dan/repos/mariadb-server-10.5/sql/sql_plugin.cc:2313
#2  0x00000000006fbd4e in mysql_execute_command (thd=thd@entry=0x7f8694000f38) at /home/dan/repos/mariadb-server-10.5/sql/sql_parse.cc:5987
#3  0x00000000006f769e in mysql_parse (thd=thd@entry=0x7f8694000f38, rawbuf=0x7f8694012c20 "install soname 'locales'", length=<optimized out>, 
    parser_state=parser_state@entry=0x7f86cc1c65b0, is_com_multi=false, is_next_command=<optimized out>) at /home/dan/repos/mariadb-server-10.5/sql/sql_parse.cc:8120
#4  0x00000000006f5a18 in dispatch_command (command=command@entry=COM_QUERY, thd=thd@entry=0x7f8694000f38, packet=packet@entry=0x7f86940080f9 "install soname 'locales'", 
    packet_length=packet_length@entry=24, is_com_multi=false, is_next_command=false) at /home/dan/repos/mariadb-server-10.5/sql/sql_parse.cc:1891
#5  0x00000000006f7a82 in do_command (thd=0x7f8694000f38) at /home/dan/repos/mariadb-server-10.5/sql/sql_parse.cc:1375
#6  0x00000000007e9f69 in do_handle_one_connection (connect=<optimized out>, connect@entry=0x2a81308, put_in_cache=true) at /home/dan/repos/mariadb-server-10.5/sql/sql_connect.cc:1416
#7  0x00000000007e9ddb in handle_one_connection (arg=arg@entry=0x2a81308) at /home/dan/repos/mariadb-server-10.5/sql/sql_connect.cc:1318
#8  0x0000000000aedeb6 in pfs_spawn_thread (arg=0x2326228) at /home/dan/repos/mariadb-server-10.5/storage/perfschema/pfs.cc:2201
#9  0x00007f86cd4ae947 in start_thread () from /lib64/libc.so.6
#10 0x00007f86cd534860 in clone3 () from /lib64/libc.so.6
```

Only the behaviour of `plugin_load_add` is changed.

PR #2745, opened 2023-09-04
MDEV-23924 s390x patch  srw_lock.cc

Nayana-ibm:
- [x] *The Jira issue number for this PR is: MDEV-23924*

## Description
Fix debug mode build failure on s390x. 
Replaced builtin_ttest by __builtin_tx_nesting_depth() > 0 as a s390x equivalent version of the expression.

## How can this PR be tested?
cmake -DCMAKE_BUILD_TYPE=Debug
make 
make install
mysql-test
./mtr --suite=unit --force --max-test-fail=0

## PR quality check

### DIFF
--- a/storage/innobase/sync/srw_lock.cc
+++ b/storage/innobase/sync/srw_lock.cc
@@ -117,8 +117,13 @@ bool transactional_lock_enabled()
 __attribute__((target("htm"),hot))
 bool xtest()
 {
+# ifdef __s390x__
+  return have_transactional_memory &&
+    __builtin_tx_nesting_depth() > 0;
+# else
   return have_transactional_memory &&
     _HTM_STATE (__builtin_ttest ()) == _HTM_TRANSACTIONAL;
+# endif
 }
 # endif
 #endif


storage/innobase/sync/srw_lock.cc
```diff
@@ -117,8 +117,13 @@ bool transactional_lock_enabled()
 __attribute__((target("htm"),hot))
 bool xtest()
 {
+# ifdef __s390x__
+  return have_transactional_memory &&
+    __builtin_tx_nesting_depth() > 0;
```
dr-m, line 122: Thank you for this fix. I found that the [oldest available GCC documentation that mentions HTM for this platform](https://gcc.gnu.org/onlinedocs/gcc-6.5.0/gcc/S_002f390-and-zSeries-Options.html) was for GCC 6.5. This is also the first documentation where I found a mention of this built-in function.

Because the built-in function was introduced at the same time as the {{htm}} target attribute, I do not think that we need any additional compile-time checks for using this.

dr-m (commented):
Great work. I was a bit in the dark when implementing this, guided by @grooverdan. I did not run any performance tests, only checked that the (non-debug) builds and tests on our CI pass.

Can you please submit this against the 10.6 branch, which is the oldest applicable version? I think that we need an MDEV bug report for this as well.

storage/innobase/sync/srw_lock.cc
dr-m: Based on what I tested on https://godbolt.org, `__builtin_tx_nesting_depth()` is specific to s390x and not applicable to the POWER implementation of HTM (v2.07 Hardware Transactional Memory, not v2.09 Hardware Trace Monitor).

dr-m:
> Sure. So one more PR for 10.6 branch?

It might be possible to retain this PR if you rebase your commit to the 10.6 branch and force-push it.

storage/innobase/sync/srw_lock.cc
dr-m: Just to be clear, there is no "issue" here, only an observation why the change looks correct.

grooverdan:
> > Sure. So one more PR for 10.6 branch?
> 
> It might be possible to retain this PR if you rebase your commit to the 10.6 branch and force-push it.

Yes.
```
git fetch {remote}
git rebase --onto {remote}/10.6 HEAD^
git push -f git@github.com:linux-on-ibm-z/server.git
# Then edit this PR title, change the target branch to 10.6

# Note:remote is the one that corresponds to https://github.com/MariaDB/server.git
```

dr-m:
> > > Sure. So one more PR for 10.6 branch?
> > 
> > 
> > It might be possible to retain this PR if you rebase your commit to the 10.6 branch and force-push it.
> 
> Yes.

Apparently, something went wrong, because the `VERSION` file in your branch still says 11.3 and not 10.6.

Perhaps the following would work (referring to your initial commit by the commit hash):
```sh
git fetch {remote}
git checkout {remote}/10.6
git cherry-pick --no-commit 85782405d760e5826339c19d04f068fac6d5f8ee
git commit --date=now
git push -f git@github.com:linux-on-ibm-z/server.git
```

grooverdan:
note the above `git push -f git@github.com:linux-on-ibm-z/server.git HEAD:11.3` will push is back to this PR. The branch name will be incorrect but it will contain the right information for the PR. Editing the target version is still required.

Alternately if you allows maintainer edits on this PR I can help out and do this for you.

image

grooverdan:
@Nayana-ibm off topic request, Galera used for MariaDB HA is [failing on Ubuntu Focal on s390x](https://buildbot.mariadb.org/#/builders/325/builds/847), always the same assertion. [Ubuntu Jammy](https://buildbot.mariadb.org/#/builders/366/builds/456) succeeds, and [other distros](https://buildbot.mariadb.org/#/grid?branch=refs/pull/11/merge).

I'm assuming its a gcc compiler fault on Focal or some less defined behaviour resulting in the error. If you have a work around (like a compiler flag) or code change https://github.com/MariaDB/galera (branch mariadb-4.x) that would be much appreciated.

As these Galera versions are included in the Docker Official Images of MariaDB. Focal is still used for the 10.5 and 10.6 MariaDB versions on s390x.

grooverdan (approved):

dr-m:
Thank you, @Nayana-ibm. For what it is worth, this does not affect the stand-alone version https://github.com/dr-m/atomic_sync because it does not implement any `xtest()` predicate. In InnoDB, we only use the predicate in order to relax some debug assertions when locks are being elided. (That might be completely unnecessary after all. If an assertion failure would result in a system call being invoked, the memory transaction should be rolled back (at least with Intel TSX-NI) and the code re-executed without lock elision.)

Nayana-ibm:
> @Nayana-ibm off topic request, Galera used for MariaDB HA is [failing on Ubuntu Focal on s390x](https://buildbot.mariadb.org/#/builders/325/builds/847), always the same assertion. [Ubuntu Jammy](https://buildbot.mariadb.org/#/builders/366/builds/456) succeeds, and [other distros](https://buildbot.mariadb.org/#/grid?branch=refs/pull/11/merge).
> 
> I'm assuming its a gcc compiler fault on Focal or some less defined behaviour resulting in the error. If you have a work around (like a compiler flag) or code change https://github.com/MariaDB/galera (branch mariadb-4.x) that would be much appreciated.
> 
> As these Galera versions are included in the Docker Official Images of MariaDB. Focal is still used for the 10.5 and 10.6 MariaDB versions on s390x.

Tests are passing when we install gcc/g++10 on ubuntu 20.04. Can you please check by upgrading gcc/g++ to v10 on ubuntu 20.04 node?
can you please ask to upgrade gcc/g++ to v10 on ubuntu 20.04 node?

PR #1829, opened 2021-05-16
[WIP] Try to fix dist-upgrade

fauust:
Let's see if BB pass.
Discussion is in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=988089 for the moment.

PR #2748, opened 2023-09-06
MDEV-30953: Add package for Galera server

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-30953*

## Description
Add new package mariadb-server-galera to Debian to recude mariadb-server package dependencies. Galera cluster scripts and libraries are currently in package mariadb-server. Move Galera cluster scripts and libraries to new package mariadb-server-galera as most of the people does not need them and it allows to slim down mariadb-server package

## How can this PR be tested?
As Salsa-CI is currently running out of space most of time (this should be addressed) this could check out:
branch: https://salsa.debian.org/illuusio/mariadb-server/-/pipelines/575821
11.3: https://salsa.debian.org/illuusio/mariadb-server/-/pipelines/575819
or build by hand in Debian sid with `debian/autobake-deb.sh`

This should also go thru MariaDB BB builds

- [x ] *This is a new feature and the PR is based against the latest MariaDB development branch.*

## PR quality check


debian/control
```diff
@@ -916,6 +912,28 @@ Description: MySQL compatibility links to mariadb-server binaries/scripts.
...
+Architecture: any
+Depends: galera-4 (>= 26.4),
+         gawk,
+         iproute2 [linux-any],
+         lsof [linux-any],
+         mariadb-server,
+         rsync,
+         socat,
+         ${misc:Depends},
+         ${perl:Depends},
```
grooverdan, line 925: I didn't think anything here was perl.

Depends on mariadb-backup too?

debian/control
```diff
@@ -916,6 +912,28 @@ Description: MySQL compatibility links to mariadb-server binaries/scripts.
...
+         ${shlibs:Depends}
+Conflicts: mariadb-server (<< 11.3)
+Breaks: mariadb-server (<< 1:11.3)
+Description: MariaDB database scripts needed by Galera cluster
+ MariaDB Galera Cluster is a virtually synchronous multi-primary
+ cluster for MariaDB. It is available on Linux only, and only supports
+ the InnoDB storage engine (although there is experimental support for
+ MyISAM and, from MariaDB 10.6, Aria.
```
grooverdan, line 933: "from MariaDB 10.6" looks odd in a 11.3 package - omit.

grooverdan (commented):
move and enable debian/additions/mariadb.conf.d/60-galera.cnf with "ON" settings.
```
[galera]
# Mandatory settings
wsrep_on                 = ON
wsrep_provider      = /usr/lib/galera/libgalera_smm.so
binlog_format            = row
innodb_autoinc_lock_mode = 2

#wsrep_cluster_name       = "MariaDB Galera Cluster"
#wsrep_cluster_address    = gcomm://
#default_storage_engine   = InnoDB
```
Uncommentted ones changed because they aren't the default. Need to test more if wsrep_cluster_address can be unset/blank.

Is RPM changes from #1829 moving here? Including change the CMakeLists.txt COMPONENT for every part here so that RPM is as close as possible.

TODO check is wsrep_provider path on MariaDB.org deb package is in the same path as debian, and ubuntu.

Also `pv` is an optional package used by SST.

debian/control
illuusio: Perl is correct. I don't know mariadb-backup. I'll check it out

debian/control
illuusio: Ok i've reworded. Should read what copy-paste

debian/control
grooverdan: mariadb-backup is one of the SST mechanisms, probably the one that should be used, but not quite enough to override the default.

Comments where intentional in the configuration file - not having things in there that are default. Odd this was in a client package.

illuusio: This should now be resolved if build goes well should check RPMs

illuusio:
Ok no it's working as expected. Galera is turned OFF when it's not installed. What should be done is to ask user or just enable things if Galera package is installed.

grooverdan:
Can this be made so the the upgrade from mariadb-server <= 11.7 automaticity goes to mariadb-server-galera when available? This is the main criteria for acceptance.

PR #2750, opened 2023-09-06
MDEV-21465: expose bind-address parameter via cmd options in client

YVEF:
- [x] *The Jira issue number for this PR is: MDEV-21465*

## Description
1. The PR is intended to provide the command line parameter to use existing functionality of address binding for outbound connections in mariadb client.
2. __
3. No side-effects.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -45,6 +45,7 @@
 #if defined(HAVE_LOCALE_H)
 #include <locale.h>
 #endif
+#include <client_connect.h>
 
 const char *VER= "15.2";
 
@@ -240,11 +241,10 @@ static MYSQL mysql;			/* The connection */
 static my_bool ignore_errors=0,wait_flag=0,quick=0,
                connected=0,opt_raw_data=0,unbuffered=0,output_tables=0,
 	       opt_rehash=1,skip_updates=0,safe_updates=0,one_database=0,
-	       opt_compress=0, using_opt_local_infile=0,
-	       vertical=0, line_numbers=1, column_names=1,opt_html=0,
+               using_opt_local_infile=0, vertical=0, line_numbers=1,
+               column_names=1,opt_html=0,
                opt_xml=0,opt_nopager=1, opt_outfile=0, named_cmds= 0,
-	       tty_password= 0, opt_nobeep=0, opt_reconnect=1,
-	       opt_secure_auth= 0,
+               tty_password= 0, opt_nobeep= 0, opt_reconnect= 1,
                default_pager_set= 0, opt_sigint_ignore= 0,
                auto_vertical_output= 0, show_query_cost= 0,
                show_warnings= 0, executing_query= 0,
@@ -254,17 +254,14 @@ static my_bool column_types_flag;
 static my_bool preserve_comments= 0;
 static my_bool in_com_source, aborted= 0;
 static ulong opt_max_allowed_packet, opt_net_buffer_length;
-static uint verbose=0,opt_silent=0,opt_mysql_port=0, opt_local_infile=0;
+static uint verbose= 0, opt_silent= 0, opt_local_infile= 0;
 static uint my_end_arg;
-static char * opt_mysql_unix_port=0;
 static int connect_flag=CLIENT_INTERACTIVE;
 static my_bool opt_binary_mode= FALSE;
 static my_bool opt_connect_expired_password= FALSE;
 static int interrupted_query= 0;
-static char *current_host,*current_db,*current_user=0,*opt_password=0,
-            *current_prompt=0, *delimiter_str= 0,
-            *default_charset= (char*) MYSQL_AUTODETECT_CHARSET_NAME,
-            *opt_init_command= 0;
+static char *current_db;
+static char *current_prompt=0, *delimiter_str= 0, *opt_init_command= 0;
 static char *histfile;
 static char *histfile_tmp;
 static String glob_buffer,old_buffer;
@@ -272,9 +269,8 @@ static String processed_prompt;
 static char *full_username=0,*part_username=0,*default_prompt=0;
 static int wait_time = 5;
 static STATUS status;
-static ulong select_limit,max_join_size,opt_connect_timeout=0;
+static ulong select_limit,max_join_size;
 static char mysql_charsets_dir[FN_REFLEN+1];
-static char *opt_plugin_dir= 0, *opt_default_auth= 0;
 static const char *xmlmeta[] = {
   "&", "&amp;",
   "<", "&lt;",
@@ -296,10 +292,9 @@ static char delimiter[16]= DEFAULT_DELIMITER;
 static uint delimiter_length= 1;
 unsigned short terminal_width= 80;
 
-static uint opt_protocol=0;
-static const char *opt_protocol_type= "";
-
-#include "sslopt-vars.h"
+static CLNT_CONNECT_OPTIONS cl_opts= CLNT_INIT_OPTS_WITH_PRG_NAME_DEFCHAR(
+  "mariadb",
+  (char*) MYSQL_AUTODETECT_CHARSET_NAME);
 
 const char *default_dbug_option="d:t:o,/tmp/mariadb.trace";
 
@@ -330,8 +325,7 @@ static int com_nopager(String *str, char*), com_pager(String *str, char*),
 #endif
 
 static int read_and_execute(bool interactive);
-static int sql_connect(char *host,char *database,char *user,char *password,
-		       uint silent);
+static int sql_connect(CLNT_CONNECT_OPTIONS *opts, uint silent);
 static const char *server_version_string(MYSQL *mysql);
 static int put_info(const char *str,INFO_TYPE info,uint error=0,
 		    const char *sql_state=0);
@@ -1292,8 +1286,7 @@ int main(int argc,char *argv[])
   glob_buffer.realloc(512);
   completion_hash_init(&ht, 128);
   init_alloc_root(PSI_NOT_INSTRUMENTED, &hash_mem_root, 16384, 0, MYF(0));
-  if (sql_connect(current_host,current_db,current_user,opt_password,
-		  opt_silent))
+  if (sql_connect(&cl_opts, opt_silent))
   {
     quick= 1;					// Avoid history
     status.exit_status= 1;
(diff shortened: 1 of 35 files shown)


sql-common/client.c
```diff
@@ -1071,6 +1071,10 @@ void mysql_read_default_options(struct st_mysql_options *options,
           break;
         case OPT_enable_cleartext_plugin:
           break;
+        case OPT_bind_address:
+          my_free(options->bind_address);
+            options->bind_address = opt_strdup(opt_arg, MYF(MY_WME));
```
grooverdan, line 1076: this can be `EXTENSION_SET_STRING(options, bind_address, opt_arg);` for consistency with other args right?

sql-common/client.c
```diff
@@ -3895,6 +3899,10 @@ mysql_options(MYSQL *mysql,enum mysql_option option, const void *arg)
       }
     }
     break;
+  case MYSQL_OPT_BIND:
```
grooverdan, line 3902: same as above `EXTENSION_SET_STRING(options, bind_address, opt_arg);`

client/mysql.cc
```diff
@@ -1632,6 +1637,10 @@ static struct my_option my_long_options[] =
    0, 0, 0, GET_NO_ARG, NO_ARG, 0, 0, 0, 0, 0, 0},
   {"binary-as-hex", 0, "Print binary data as hex", &opt_binhex, &opt_binhex,
    0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
+  {"bind-address", OPS_BIND_ADDRESS,
+   "name IP address to bind to.",
```
grooverdan, line 1641: `OPS_BIND_ADDRESS` can just be `0` (unless I've missed something significant on how this works).
Omit `"name '", this is filled out in help and `--bind-address=name name IP address to bind to.` looks quite odd.

client/mysql.cc
```diff
@@ -265,6 +265,7 @@ static char *current_host,*current_db,*current_user=0,*opt_password=0,
             *current_prompt=0, *delimiter_str= 0,
             *default_charset= (char*) MYSQL_AUTODETECT_CHARSET_NAME,
             *opt_init_command= 0;
+static char *opt_bind_address=0;
```
grooverdan, line 268: technically `static` vars are always initialized to 0, so omit or use the consistent style of `opt_bind_address= 0;`

grooverdan (changes requested):
Thanks for the contribution.

Can you also update man/mariadb.1 to include the option?

While the MDEV did include only the `mysql`/`mariadb` as the client, there are other clients that could be handled the same way.

In client:
* mysqladmin.cc
* mysqlbinlog.cc
* mysqlcheck.c
* mysqldump.c
* mysqlimport.c
* mysqlshow.c
* mysqlslap.c
* mysqltest.cc

And extra/mariabackup/backup_mysql.cc.

If you feel particular keen, and I'd appreciate it (though obviously optional), consolidate the connection methods across the client into a separate file like `sql-common/connect.c`  and include in CMakeLists.txt like:
```
MYSQL_ADD_EXECUTABLE(mariadb-dump mysqldump.c ../sql-common/my_user.c ../sql-common/connect.c)
```

sql-common/client.c
YVEF: Sure, but to use the macro the field should be part of st_mysql_options_extention. While I'm not entirely sure, whether this parameter need to be part of the main options or extensions structure, the [comment](https://github.com/MariaDB/server/blob/8d9bc61d0bf783fa792e6c3be37b0eceecbeec89/include/mysql.h#L196) here helped me to decide. What do you think?

client/mysql.cc
YVEF: fixed

client/client_priv.h
```diff
@@ -111,6 +111,7 @@ enum options_client
   OPT_DO_DOMAIN_IDS,
   OPT_IGNORE_SERVER_IDS,
   OPT_DO_SERVER_IDS,
+  OPS_BIND_ADDRESS,
```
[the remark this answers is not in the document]
YVEF: thank you. fixed

client/mysql.cc
YVEF: thanks for the notes. fixed

YVEF:
> If you feel particular keen, and I'd appreciate it (though obviously optional), consolidate the connection methods across the client into a separate file like `sql-common/connect.c` and include in CMakeLists.txt like:
> 
Sure, It's might be important point of refactoring and I would appreciate some hints on this. So, as far as I can see, all clients (except mariadb-backup) use the pvio plugin from connector. Once the functionality is implemented there, it is already spread across clients. But only in backup util I had to do it separately. Maybe it worth doing this the same way as it is done in other clients?

sql-common/client.c
grooverdan: You made a good choice. Thanks.

grooverdan:
mysqlcheck.cc:dbConnect
mysql.cc:do_connect
backup_mysql.cc:xb_mysql_connect

And the other functions you've added to bind to are effectively the same function.

They have a program name (which would appreciate an update to the latest version ('mysql' -> 'mariadb',  `mysql_dump` -> `mariadb-dump` etc).

So an interfaces of:
```
bool client_connect(MYSQL *mysql, const char *clientname, const char *host, const char *user,
                          const char *password, const char *database, const char *bind_addr, ulong flags)
{

   mysql_options(&mysql_connectio ....

...
  return  mysql_real_connect(&mysql_connection,host,user,passwd,
                            NULL,opt_mysql_port,opt_mysql_unix_port, 0))


}
```

Callers would do their:
```
 mysql_init()

any items specific to that client.
ret = client_connect(..... )

any post error handling or post connection options.
```

extra/mariabackup/backup_mysql.cc
```diff
@@ -124,6 +124,11 @@ xb_mysql_connect()
 	mysql_options(connection, MYSQL_OPT_PROTOCOL, &opt_protocol);
 	mysql_options(connection,MYSQL_SET_CHARSET_NAME, "utf8");
 
+    if(opt_bind_address && *opt_bind_address) {
+        mysql_options(connection, MYSQL_OPT_BIND, opt_bind_address);
+    }
```
grooverdan, line 129: just watch the whitespace formatting, normally what you did is right, however there's tabs in this section of file so best not to mess that up too much (unless you where rewriting most of a function, in which its perfectly ok to change the entire function)..

man/mariadb-admin.1
```diff
@@ -800,6 +800,22 @@ Display help and exit\&.
 .sp -1
 .IP \(bu 2.3
 .\}
+.\" mysql: bind-address option
+.\" bind-address option: mysql
+\fB\-\-bind\-address=\fR\fB\fIip_address\fR\fR
+.sp
+On a computer having multiple network interfaces, use this option to select
+which interface to use for connecting to the MySQL server\&.
```
grooverdan, line 808: can we call it "the MariaDB server". There are a few lingering instances of MySQL server in the man pages but we'll get to those.

sql-common/client.c
```diff
@@ -2660,19 +2674,59 @@ set_connect_attributes(MYSQL *mysql, char *buff, size_t buf_len)
...
-  NET		*net= &mysql->net;
+  char		    buff[NAME_LEN+USERNAME_LENGTH+100];
+  int               scramble_data_len, UNINIT_VAR(pkt_scramble_len);
+  char              *end,*host_info= 0, *server_version_end, *pkt_end;
+  char              *scramble_data;
+  const char        *scramble_plugin;
+  ulong		    pkt_length;
+  NET		    *net= &mysql->net;
+  int               rc= 0;
+  time_t start_t= time(NULL);
```
grooverdan, line 2728: consistent style please.

sql-common/client.c
```diff
@@ -2885,6 +2954,21 @@ CLI_MYSQL_REAL_CONNECT(MYSQL *mysql,const char *host, const char *user,
...
+      {
+        for (bres= bind_addr; bres; bres= bres->ai_next)
+        {
+          if (!(rc= bind(sock, bres->ai_addr, (int)bres->ai_addrlen)))
+            break;
+        }
+        if (rc)
+        {
+          closesocket(sock);
+          sock= INVALID_SOCKET;
```
grooverdan, line 2967: Does `saved_error` like the previous block need to happen here too?

man/mariadb-admin.1
YVEF: Sure. I was also wondering if I could use MariaDB here since I didn't know the intention. Fixed.

sql-common/client.c
YVEF: Fixed

sql-common/client.c
YVEF: You are right. Fixed

YVEF:
While hiding the common infrastructure for the all clients connection is a good idea, I didn't see any meaningful reason to only use this parameters as it is ( + bind_address). As long as the code use the same own general variables, the extensibility such method is not the pleasant copy-pasting. So, I suggest using the shared options pack to distribute the shared connection method. The idea is extremely simple but I had to make enough changes. That's why I marked the pr as draft.

include/mysql.h
```diff
@@ -189,7 +189,10 @@ enum mysql_option
   /* MariaDB options */
   MYSQL_PROGRESS_CALLBACK=5999,
   MYSQL_OPT_NONBLOCK,
-  MYSQL_OPT_USE_THREAD_SPECIFIC_MEMORY
+  MYSQL_OPT_USE_THREAD_SPECIFIC_MEMORY,
+
+  /* Libmariadb compatibility */
+  MARIADB_OPT_TLS_VERSION
```
YVEF, line 195: Backup utility exposes the tls version option though the cmd but didn't setup it during the connection. I suggest to use it for compatibility. If no, hiding it for the backup util is also an option.

grooverdan (approved):
Very well done. This is so much neater.

Small changes:
* rebased and changed a few trivial conflicts where `get_tty_password` -> `my_get_tty_password` was renamed.
* Added a couple of missing includes in libmysqld/examples/ for an embedded build.
* squashed commits up and the commit message

Otherwise looks perfect.

I was concerned at one point there was missing `my_free(cl_opts.bind_address);` in mysql.cc:mysql_end, mysqlbinlog.cc:cleanup, mysqldump.cc:free_resources, and also the inconsistent `my_free(cl_options.{user,password))`, but it all seems correct to match the strdups that apply slightly differently in each function.

grooverdan:
@YVEF how are the defined meant to work for the Windows build - its not currently.

vuvova:
Note that this is for 11.4, not 11.3.
And better wait for 11.3.1 release, when commits c8d903e5922455c1c0424a888b306463217dca95 and c275dc48d736e4505e7819cbbbdb09b588c6e035 will be pushed, as your commit will not apply cleanly after them, there will be merge conflicts

YVEF:
@grooverdan your are right. that was fixed and included into the main commit.

LinuxJedi (approved):
Excellent cleanup here too, approved for 11.4

grooverdan:
Thank you @YVEF

vuvova:
1. refactoring and a feature must be in different commits
2. sql-common is for the code *common* between a client and a server (thus the name). do_client_connect is client only, should be in client/

vuvova:
So, it's not merged, needs many changes, way past the 11.4 preview deadline.
I'll try to fix it, but might not have enough time for that. Thus it likely won't be in 11.4 despite the confusing status.

PR #2755, opened 2023-09-13
MDEV-22418 mysqladmin wrong error with simple_password_check

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-22418*

## Description

mysqladmin's presumption about the cause of an error by looking at the error code was presumptious. Server knows best, so pass the error along.

## How can this PR be tested?

mtr included.

## PR quality check


grooverdan:
@vaintroub  can I get your thoughts there:

Current behaviour under mariadb-admin error - returns 255 on linux or -1 on windows.
a) normalise to 1 as error (per PR)
b) change windows to 255 so MTR test is easier to write.
c) extended mtr to accept `--error -1`

vaintroub:
> a) normalise to 1 as error (per PR)
is the best I think

vaintroub (approved):
Looks good to me

PR #2760, opened 2023-09-15
MDEV-31768: Alias MASTER_DEMOTE_TO_REPLICA for MASTER_DEMOTE_TO_SLAVE

bnestere:
Per MDEV-20601, REPLICA should be an alias for SLAVE in SQL statements.


knielsen (approved):
The keyword "MASTER_DEMOTE_TO_REPLICA" is seriously ugly :-)
But the patch is good to push.

 \- Kristian.

PR #2762, opened 2023-09-18
MDEV-32168: slave_error_param condition is never checked from the wait_for_slave_param.inc

an3l:


mysql-test/suite/binlog_encryption/encrypted_master_switch_to_unencrypted_coords.test
```diff
@@ -136,6 +136,8 @@ start slave;
 SHOW TABLES;
 
 --disable_connect_log
+# IO thread is stopped, stop SQL thread only
+--let $rpl_only_running_threads= 1
```
bnestere, line 140: I think using `rpl_only_running_threads=1` is fine to keep as is (in the context of your current commit), but for future reference, there is `--source include/stop_slave_sql.inc` and `stop_slave_io.inc` which would set the behavior expectation more explicitly.

mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
```diff
@@ -116,6 +116,9 @@ show status like 'Rpl_semi_sync_master_no_tx';
 --connection server_2
 --eval SET @@GLOBAL.debug_dbug= "$sav_server_2_dbug"
 --eval SET @@GLOBAL.rpl_semi_sync_slave_enabled= 0
+--let $slave_io_errno=1595
+--source include/wait_for_slave_io_error.inc
```
bnestere, line 120: The `rpl_semi_sync_shutdown_await_ack` test will be a little more complicated to fix. If you look at the "driver" test, `rpl_semi_sync_shutdown_await_ack.test`, it calls this `.inc` file 4 times, each having different expectations for servers 2 and 3 to error or not.. When `corrupt_queue_event`, is provided to the server, the IO thread will fail with this error (and it can be either server 2 or 3, depending on which test case is running). If a server doesn't have `corrupt_queue_event` I'd imagine this errno would probably be `2003` (lost connection to server) instead. One way to do this would be using `SELECT REGEXP`, something like (using `server_2` as the example, and note it is untested :)):

```
# 2003 is the default..
--let $slave_io_errno=2003

# and if corrupt_queue_event is specified, we should await a relay log error instead
if (`SELECT REGEXP("corrupt_queue_event","$server_2_dbug")`)
{
--let $slave_io_errno=1595
}
```

mysql-test/suite/multi_source/gtid_slave_pos.test
```diff
@@ -118,8 +118,11 @@ set default_master_connection = 'slave1';
 STOP SLAVE;
 --source include/wait_for_slave_to_stop.inc
 set default_master_connection = 'slave2';
+--let $slave_sql_errno= 1942
+--source include/wait_for_slave_sql_error.inc
 STOP SLAVE;
---source include/wait_for_slave_to_stop.inc
+--let $rpl_only_running_threads= 1
```
bnestere, line 124: You should be able to remove the above `STOP SLAVE` as it is now redundant with the addition of `source include/stop_slave.inc`

mysql-test/suite/rpl/include/rpl_shutdown_wait_slaves.inc
```diff
@@ -75,17 +75,26 @@ EOF
 --connection server_1
 DROP TABLE t1;
 
+# Slaves IO thread will receive the disconnect error when master was shutdown
+# so we are allowing error on start.
```
bnestere, line 79: This is a bit vague with the following behavior essentially starting twice:
```
--let rpl_allow_error=1
--source include/wait_for_slave_io_to_start.inc
```

I think it should be reworded to mention that the IO thread will automatically try to reconnect, and the error is allowed during this time, rather than "on start"

mysql-test/suite/rpl/t/rpl_mdev_17614.test
```diff
@@ -41,7 +41,8 @@ SELECT * FROM t1;
 
 # restart replication for the next testcase
 stop slave;
---source include/wait_for_slave_to_stop.inc
+--let $rpl_only_running_threads= 1
```
bnestere, line 44: Here too, the above `stop slave` can be removed

mysql-test/suite/rpl/t/rpl_semi_sync_master_shutdown.test
```diff
@@ -30,8 +30,9 @@ connection master;
 --let $rpl_server_number=1
 source include/rpl_stop_server.inc;
 
---connection slave
---source include/stop_slave.inc
+# After stoping the master, slave receives the disconnect error (2003)
```
bnestere, line 33: We should ensure this and wait to see the error before going on, and also the following 2 line comment should be deleted.

mysql-test/suite/rpl/t/rpl_semi_sync_master_shutdown.test
```diff
@@ -42,8 +43,10 @@ source include/rpl_start_server.inc;
 # Clean up
 #
 --connection slave
---source include/stop_slave.inc
---source include/start_slave.inc
+--source include/wait_for_slave_sql_to_start.inc
+--let rpl_allow_error=1
+--source include/wait_for_slave_io_to_start.inc
```
bnestere, line 48: With the above change to await the error, I imagine the IO thread will have stopped and that it should be re-started manually, i.e. it should have missed the window for automatic restart.

bnestere (approved):
Generally looks like a good fix and cleanup, thanks Anel! See my few notes too.

sql/slave.cc
```diff
@@ -4775,7 +4775,11 @@ pthread_handler_t handle_slave_io(void *arg)
         goto err;
       goto connected;
     }
-    DBUG_EXECUTE_IF("fail_com_register_slave", goto err;);
+    DBUG_EXECUTE_IF("fail_com_register_slave", {
```
bnestere, line 4778: The braces should be on new lines.

bnestere (approved):
Left a comment about a quick syntax fix-up, otherwise looks good. Thanks @an3l!

PR #2768, opened 2023-09-20
MDEV-31285 : Assertion `state() == s_executing || state() == s_prepar…

janlindstrom:
…ing || state() == s_prepared || state() == s_must_abort || state() == s_aborting || state() == s_cert_failed || state() == s_must_replay' failed

- [x] *The Jira issue number for this PR is: MDEV-31285*

## Description

When applier tries to execute write rows event it find out in table_def::compatible_with that value is not compatible and sets error and thd->is_slave_error but thd->is_error() is false. Later in rpl_group_info::slave_close_thread_tables we commit stmt. This is bad for Galera because later in apply_write_set we notice that event apply was not successful and try to rollback transaction, but wsrep transaction is already in s_committed state.

This is fixed on rpl_group_info::slave_close_thread_tables so that in Galera case we rollback stmt if thd->is_slave_error or thd->is_error() is set. Then later we can rollback wsrep transaction.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/mdev-31285.result
+++ b/mysql-test/suite/galera/r/mdev-31285.result
@@ -1,0 +1,23 @@
+connection node_2;
+connection node_1;
+connection node_1;
+connection node_2;
+connection node_1;
+CREATE TABLE t ENGINE=InnoDB WITH SYSTEM VERSIONING AS SELECT 1 AS i;
+SHOW CREATE TABLE t;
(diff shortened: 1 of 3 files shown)


sysprg:
Thanks, fix merged with head revision as https://github.com/MariaDB/server/commit/3c65434b789be9b06235bbecec08a7e97301bb52

PR #2771, opened 2023-09-25
MDEV-18200 MariaBackup - remove symlink requirement from mariabackup.data_directory

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-18200*

## Description

Test if this runs on Windows.

## PR quality check


LinuxJedi (commented):
The very purpose of this test is to test the `DATA DIRECTORY` option. If you don't have symlinks (.isl files) enabled, the `DATA DIRECTORY` option doesn't work. So, isn't skipping this test correct behaviour? Or am I missing something?

vaintroub:
> The very purpose of this test is to test the `DATA DIRECTORY` option. If you don't have symlinks (.isl files) enabled, the `DATA DIRECTORY` option doesn't work. So, isn't skipping this test correct behaviour? Or am I missing something?


DATA DIRECTORY, with .isl files used to work on Windows without problems.  Now it does not, and there is no way to enable it. something somewhere got broken.

LinuxJedi:
> DATA DIRECTORY, with .isl files used to work on Windows without problems. Now it does not, and there is no way to enable it. something somewhere got broken.

Sure, but fudging the test to not test that doesn't seem to be the fix. There should probably be a separate MDEV to fix that.

mysql-test/suite/mariabackup/data_directory.test
```diff
@@ -1,8 +1,9 @@
---source include/have_symlink.inc
 let $table_data_dir=$MYSQLTEST_VARDIR/ddir;
 mkdir $table_data_dir;
+--disable_warnings
 --replace_result $table_data_dir table_data_dir
 EVAL CREATE TABLE t(a INT) ENGINE=InnoDB DATA DIRECTORY='$table_data_dir';
+--enable_warnings
```
LinuxJedi, line 6: This test is for the "DATA DIRECTORY" option which requires symlinks to actually work. Bypassing in this way just fakes a test pass rather than fixing the actual problem (see my previous comments in this PR).

LinuxJedi (changes requested):
I don't think this is the correct way to fix the problem. The problem isn't well defined in the PR, so it is difficult to say for sure.

Also, the description links to a closed MDEV. This appears to be a subset of that and should have its own MDEV so that we can better track what the actual intention is.

PR #2772, opened 2023-09-27
MDEV-32265: Seconds_Behind_Master is inaccurate for Delayed replication

bnestere:
MDEV-32265: seconds_behind_master is inaccurate for Delayed replication

If a replica is actively delaying a transaction when restarted (STOP
SLAVE/START SLAVE), when the sql thread is back up,
Seconds_Behind_Master will present as 0 until the configured
MASTER_DELAY has passed. That is, before the restart,
last_master_timestamp is updated to the timestamp of the delayed
event. Then after the restart, the negation of sql_thread_caught_up
is skipped because the timestamp of the event has already been used
for the last_master_timestamp, and their update is grouped together
in the same conditional block.

This patch fixes this by separating the negation of
sql_thread_caught_up out of the timestamp-dependent block, so it is
called any time an idle parallel slave queues an event to a worker.

This patch is organized into two commits:
1. The first commit is the regression test
2. The second commit is the fix

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_parallel_sbm.test
+++ b/mysql-test/suite/rpl/t/rpl_parallel_sbm.test
@@ -36,10 +36,6 @@ create table t2 (a int);
 --echo #
 --echo # Pt 1) Ensure SBM is updated immediately upon arrival of the next event
 
---echo # Lock t1 on slave so the first received transaction does not complete/commit
---connection slave
-LOCK TABLES t1 WRITE;
-
 --connection master
 --echo # Sleep 2 to allow a buffer between events for SBM check
 sleep 2;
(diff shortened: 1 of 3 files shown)


mysql-test/suite/rpl/t/rpl_parallel_sbm.test
```diff
@@ -65,8 +61,29 @@ if (`SELECT $sbm_trx1_arrive > ($seconds_since_idling + 1)`)
 }
 --echo # ..done
 
+--echo # MDEV-32265. If a replica is actively delaying a transaction when
+--echo # restarted, when the sql thread is back up, Seconds_Behind_Master
+--echo # should pick back up where it it left-off.
+--source include/stop_slave.inc
+--source include/start_slave.inc
```
andrelkin, line 68: I'd love to see
``` Seconds_Behind_Master = SQL_Delay  - SQL_Remaining_Delay ```
that holds at this point, but I don't insist esp not being sure if any round-up can come to play.
Yet you can give it some though.

andrelkin (approved):
Looks good. Thanks!

mysql-test/suite/rpl/t/rpl_parallel_sbm.test
bnestere: I think this would result in an unstable test, as `query_get_value()` would run separate queries to get each field. Perhaps we can implement `query_get_values()` and this would be our first refactor with it :)

PR #2774, opened 2023-09-30
MDEV-32282 Galera node remains paused after interleaving FTWRLs

temeo:
- [x] *The Jira issue number for this PR is: MDEV-32282*

## Description

After two concurrent FTWRL/UNLOCK TABLES, the node stays in paused state and the following CREATE TABLE fails with

  ER_UNKNOWN_COM_ERROR (1047): Aborting TOI: Replication paused on
  node for FTWRL/BACKUP STAGE.

The cause is the use of global `wsrep_locked_seqno` to determine if the node should be resumed on UNLOCK TABLES. In some executions the `wsrep_locked_seqno` is cleared by the first UNLOCK TABLES after the second FTWRL gets past `make_global_read_lock_block_commit()`.

As a fix, use `thd->wsrep_desynced_backup_stage` to determine if the thread should resume the node on UNLOCK TABLES.

Add MTR test galera.galera_ftwrl_concurrent to reproduce the race. The test contains also cases for BACKUP STAGE which uses similar mechanism for desyncing and pausing the node.

## How can this PR be tested?

New test case `mysql-test/suite/galera/t/galera_ftwrl_concurrent.test` is added to verify that the bug is fixed.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/galera_ftwrl_concurrent.result
+++ b/mysql-test/suite/galera/r/galera_ftwrl_concurrent.result
@@ -1,0 +1,122 @@
+connection node_2;
+connection node_1;
+connect node_1_ctrl, 127.0.0.1, root, , test, $NODE_MYPORT_1;
+SET SESSION wsrep_sync_wait=0;
+#
+# Case 1: FTWRL
+#
+connection node_1;
+SET SESSION wsrep_sync_wait=0;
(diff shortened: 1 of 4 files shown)


sysprg:
Thanks, fix merged with head revision (https://github.com/MariaDB/server/commit/ef7fc586aea1048bc5526192158a8e8e935ddd1a)

PR #2775, opened 2023-10-01
MDEV-32336 deb default config - use collation-server = utf8mb4_uca1400_ai_ci

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32336*

## Description

utf8mb4_general_ci has been outdated for a while. Lets use our modern standard.

## How can this PR be tested?

my_print_defaults --mysqld on install in debian

- [] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check

### DIFF
--- a/debian/additions/mariadb.conf.d/50-server.cnf
+++ b/debian/additions/mariadb.conf.d/50-server.cnf
@@ -84,10 +84,10 @@ expire_logs_days        = 10
 # * Character sets
 #
 
-# MySQL/MariaDB default is Latin1, but in Debian we rather default to the full
+# MariaDB default is Latin1, but in Debian we rather default to the full
 # utf8 4-byte character set. See also client.cnf
-character-set-server  = utf8mb4
-collation-server      = utf8mb4_general_ci
+character-set-server     = utf8mb4
+character-set-collations = utf8mb4=uca1400_ai_ci
 
 #
 # * InnoDB


ottok (changes requested):
The migration to utf-8 was done in Debian by having Debian defaults that override the upstream defaults (latin1) as they seem outdated and unsuitable for most users. According to https://mariadb.com/kb/en/server-system-variables/#character_set_server and https://mariadb.com/kb/en/server-system-variables/#collation_server this still seems to be the case.

Would it not be the best solution to simply remove these customizations in `debian/` and default to upstream server defaults, and ensure upstream has modern and sensible values?

If changing upstream default is not an option, the change in `debian/` should be properly documented. The Jira https://jira.mariadb.org/browse/MDEV-32336 does not explain why the new default should be specifically `utf8mb4_uca1400_ai_ci` nor does the commit message explain *why* this change should be done to this value:

```
MDEV-32336 deb default config - use collation-server =
 utf8mb4_uca1400_ai_ci

utf8mb4_general_ci has been outdated for a while. Lets use our modern
standard.
```

See [tips 2 & 3 in about good commit message](https://optimizedbyotto.com/post/good-git-commit/). The commit message should say something about *why* `utf8mb4_uca1400_ai_ci` is the best value and why this change is done now in 11.3 and not say in 11.4 as https://jira.mariadb.org/browse/MDEV-25829 says.

grooverdan:
Good point on commit message. Later and actually a standard is the general criteria however I'll find some better words.

Looking back on why utf8mb4_general_ci, was copied from utf8 comment from 2012 in 438ed0408c69faa9e5c4c026e24ef52eac62f08e in and in 7c2079f600bacbd4d24762159550b3d40ad856c1 was made non-commented, but updated to utf8mb4.

[MDEV-25829](https://jira.mariadb.org/browse/MDEV-25829), I'm not sure is a genuine target version of 11.4, but I hadn't looked for it either.

I'd certainly welcome better upstream defaults too.

11.3 was chosen as only non-packaged releases have been done on this so I was assuming its still compatible and not causing packaging regressions.

illuusio:
What is the real difference between `utf8mb4_uca1400_ai_ci'` and `utf8mb4_general_ci`. Is it just newer standard or all the pages are supported?

grooverdan:
Mainly from https://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci - utf8mb4_general_ci isn't a standard, it was just a (slightly) quicker (and dirty) implementation.

ottok:
You probably want to have the git commit message updated with the text you posted in PR comments here, and use as title something like "MDEV-32336: Use utf8mb4_uca1400_ai_ci as default collation in Debian"

grooverdan:
Acceptable?

ottok (approved):

illuusio:
Is that correct new notation that I don't know `utf8mb4=uca1400_ai_ci`. I suppose so. I have to admit look very weird but if it work then I'm cool with it.

PR #2783, opened 2023-10-12
MDEV-32439 INSERT IGNORE on constraints result in ERROR rather than w…

grooverdan:
…arning

- [x] *The Jira issue number for this PR is: MDEV-32439*

## Description

INSERT IGNORE had a pecular undocumented case that when one row was inserted, there was an error rather than a warning.

As LOAD DATA IGNORE, UPDATE IGNORE, INSERT IGNORE SELECT, and INSERT IGNORE VALUES (single row, for foreign key violation) all behave the same way with a warning lets keep the behaviour normalized.

In compatibility, previously a error was generated, now a warning is generated.

This behaviour is now consistent with MySQL-8.0 too.

## How can this PR be tested?

test cases included.

## PR quality check

### DIFF
--- a/mysql-test/main/check_constraint.result
+++ b/mysql-test/main/check_constraint.result
@@ -308,5 +308,61 @@ t1	CREATE TABLE `t1` (
 ) ENGINE=MyISAM DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
 drop table t1;
 #
-# End of 10.4 tests
+# MDEV-32439 INSERT IGNORE VALUES (one row) errors on constraint
+#
+CREATE TABLE t1 (v1 varchar(10), v2 varchar(10), constraint unequal check (v1 != v2));
+INSERT IGNORE INTO t1 VALUES (1,1);
+Warnings:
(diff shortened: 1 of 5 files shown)


vuvova:
whoa, wait a bit. As you write yourself, it's not a bug. That is, it is not a mistake, someone intentionally wrote this, so this is intended behavior. And I suspect it's very old, like, Monty's code from 25 years ago?

You have to discuss it with Monty. Also realize that you'll be changing an intentional behavior that existed basically forever, it cannot be done in any GA version. May be it'll need an option to keep the old behavior (old-mode). And you need to know why it was done that way too.

I don't know the answer but my guess would be that it's because after the first row was successfully inserted, MyISAM (actually, ISAM, this behavior predates MyISAM) could no longer abort the statement cleanly, rolling back its effects, so there were only two options — abort dirty (keeping some rows already inserted) or keep going. But when the error happened on the very first row, it could, so there were no need to ignore errors in that case. I feel weird writing it in 2023 when everyone uses InnoDB.

PR #2281, opened 2022-09-27
MDEV-18200 MariaBackup full backup failed with InnoDB: Failing assert…

grooverdan:
…ion: success

There are many filesystem related errors that can occur with MariaBackup. These already outputed to stderr with a good description of the error. Many of these are permission or resource (file descriptor) limits where the assertion and resulting core crash doesn't offer developers anything more than the log message. To the user assertions and corecrashes come across as poor error handling.

There are other case where the operation of MariaBackup and server haven't kept status which are legitimate bugs, but the server state is more useful than the state of MariaBackup which is fairly documented in the log.

Being a change in InnoDB where it isn't the server this also changes the behavior of InnoDB utilities like innochecksum in a similar way to MariaBackup.

PR #2784, opened 2023-10-12
MDEV-18200 MariaBackup full backup failed with InnoDB: Failing assertion: success (10.5)

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-18200*

## Description

There are many filesystem related errors that can occur with MariaBackup. These already outputed to stderr with a good description of the error. Many of these are permission or resource (file descriptor) limits where the assertion and resulting core crash doesn't offer developers anything more than the log message. To the user, assertions and core crashes come across as poor error handling.

As such we return an error and handle this all the way up the stack.

## How can this PR be tested?

mtr test included. Leaving tests versioned as 10.4 as #2281 merge up will be in the same spot(?).

## PR quality check

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -4436,16 +4436,24 @@ bool Backup_datasinks::backup_low()
 
 		if (recv_find_max_checkpoint(&max_cp_field) == DB_SUCCESS
 		    && log_sys.log.format != 0) {
-			if (max_cp_field == LOG_CHECKPOINT_1) {
-				log_sys.log.read(max_cp_field,
-						 {log_sys.checkpoint_buf,
-						  OS_FILE_LOG_BLOCK_SIZE});
+			switch (max_cp_field) {
+			case LOG_CHECKPOINT_1:
+				if (log_sys.log.read(max_cp_field,
+						     {log_sys.checkpoint_buf,
+						     OS_FILE_LOG_BLOCK_SIZE}))
+				{
+					/* metadata_to_lsn still 0 so error returns below */
+					msg("Error: recv_find_max_checkpoint() failed.");
+					break;
+				}
+				/* fallthrough */
+			default:
+				metadata_to_lsn = mach_read_from_8(
+					log_sys.checkpoint_buf + LOG_CHECKPOINT_LSN);
+				msg("mariabackup: The latest check point"
+				    " (for incremental): '" LSN_PF "'",
+				    metadata_to_lsn);
 			}
-			metadata_to_lsn = mach_read_from_8(
-				log_sys.checkpoint_buf + LOG_CHECKPOINT_LSN);
-			msg("mariabackup: The latest check point"
-			    " (for incremental): '" LSN_PF "'",
-			    metadata_to_lsn);
 		} else {
 			msg("Error: recv_find_max_checkpoint() failed.");
 		}
@@ -4661,14 +4669,14 @@ reread_log_header:
 		checkpoint_lsn_start = log_sys.log.get_lsn();
 		checkpoint_no_start = log_sys.next_checkpoint_no;
 
-		log_sys.log.read(max_cp_field, {buf, OS_FILE_LOG_BLOCK_SIZE});
+		err = log_sys.log.read(max_cp_field, {buf, OS_FILE_LOG_BLOCK_SIZE});
 
-		if (checkpoint_no_start
+		if (err == DB_SUCCESS && (checkpoint_no_start
 		    != mach_read_from_8(buf + LOG_CHECKPOINT_NO)
 		    || checkpoint_lsn_start
 		    != mach_read_from_8(buf + LOG_CHECKPOINT_LSN)
 		    || log_sys.log.get_lsn_offset()
-		    != mach_read_from_8(buf + LOG_CHECKPOINT_OFFSET))
+		    != mach_read_from_8(buf + LOG_CHECKPOINT_OFFSET)))
 			goto reread_log_header;
 	}
 	mysql_mutex_unlock(&log_sys.mutex);
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -363,6 +363,7 @@ fil_node_t* fil_space_t::add(const char* name, pfs_os_file_t handle,
 	return node;
 }
 
+__attribute__((warn_unused_result, nonnull))
 /** Open a tablespace file.
 @param node  data file
 @return whether the file was successfully opened */
@@ -391,9 +392,9 @@ static bool fil_node_open_file_low(fil_node_t *node)
                                  : OS_FILE_OPEN | OS_FILE_ON_ERROR_NO_EXIT,
                                  OS_FILE_AIO, type,
                                  srv_read_only_mode, &success);
-    if (node->is_open())
+
+    if (success && node->is_open())
     {
-      ut_ad(success);
 #ifndef _WIN32
       if (!node->space->id && !srv_read_only_mode && my_disable_locking &&
           os_file_lock(node->handle, node->name))
--- a/storage/innobase/include/log0log.h
+++ b/storage/innobase/include/log0log.h
@@ -506,7 +506,8 @@ public:
     /** reads buffer from log file
     @param[in]	offset		offset in log file
     @param[in]	buf		buffer where to read */
-    void read(os_offset_t offset, span<byte> buf);
+    dberr_t MY_ATTRIBUTE((warn_unused_result)) read(os_offset_t offset,
+                                                    span<byte> buf);
     /** Tells whether writes require calling flush() */
     bool writes_are_durable() const noexcept;
     /** writes buffer to log file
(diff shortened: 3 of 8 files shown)


storage/innobase/fil/fil0fil.cc
```diff
@@ -391,6 +391,9 @@ static bool fil_node_open_file_low(fil_node_t *node)
                                  : OS_FILE_OPEN | OS_FILE_ON_ERROR_NO_EXIT,
                                  OS_FILE_AIO, type,
                                  srv_read_only_mode, &success);
+    if (!success)
+      return false;
+
     if (node->is_open())
```
dr-m, line 397: Will the added code make the code after `if (node->is_open())` unreachable?

extra/mariabackup/xtrabackup.cc
```diff
@@ -4661,7 +4669,10 @@ static bool xtrabackup_backup_func()
 		checkpoint_lsn_start = log_sys.log.get_lsn();
 		checkpoint_no_start = log_sys.next_checkpoint_no;
 
-		log_sys.log.read(max_cp_field, {buf, OS_FILE_LOG_BLOCK_SIZE});
+		if (log_sys.log.read(max_cp_field, {buf, OS_FILE_LOG_BLOCK_SIZE})) {
+			mysql_mutex_unlock(&log_sys.mutex);
+			goto fail;
+		}
 
 		if (checkpoint_no_start
```
dr-m, line 4677: We could simplify the above by:
```c++
err = log_sys.log.read(…);
if (err != DB_SUCCESS);
else if (checkpoint_no_start …
```
In this way, we would also avoid duplicating the call to `mysql_mutex_unlock(&log_sys.mutex)`.

storage/innobase/log/log0recv.cc
```diff
@@ -1409,13 +1414,15 @@ static dberr_t recv_log_recover_10_4()
...
 
 	if (!redo_file_sizes_are_correct()) {
 		return DB_CORRUPTION;
 	}
 
-	recv_sys.read(source_offset & ~lsn_t(OS_FILE_LOG_BLOCK_SIZE - 1),
-		      {buf, OS_FILE_LOG_BLOCK_SIZE});
+	if ((err= recv_sys.read(source_offset & ~lsn_t(OS_FILE_LOG_BLOCK_SIZE - 1),
+		      {buf, OS_FILE_LOG_BLOCK_SIZE})))
+		return err;
```
dr-m, line 1425: I’d declare the variable `err` just inside the `if` statement, in C++98 style.

storage/innobase/log/log0recv.cc
```diff
@@ -1467,13 +1474,15 @@ recv_find_max_checkpoint(ulint* max_field)
 	ib_uint64_t	checkpoint_no;
 	ulint		field;
 	byte*		buf;
+	dberr_t		err;
```
dr-m, line 1477: Instead of declaring a local variable for the entire function, I’d narrow the scope of the variable to each `if` block that is invoking `log_sys.log.read()`.

storage/innobase/fil/fil0fil.cc
```diff
@@ -366,7 +366,7 @@ fil_node_t* fil_space_t::add(const char* name, pfs_os_file_t handle,
 /** Open a tablespace file.
 @param node  data file
 @return whether the file was successfully opened */
-static bool fil_node_open_file_low(fil_node_t *node)
+static MY_ATTRIBUTE((warn_unused_result)) bool fil_node_open_file_low(fil_node_t *node)
```
dr-m, line 369: All callers of a `static` function must be in the same compilation unit. It’s rather easy to check that the return value is not being ignored by any caller.

If you think that the attribute is useful, I’d put it right before the comment:
```c++
__attribute__((warn_unused_result, nonnull))
/** Open a tablespace file.
@param node  data file
@return whether the file was successfully opened */
static bool fil_node_open_file_low(fil_node_t *node)
```

storage/innobase/log/log0log.cc
```diff
@@ -280,7 +280,7 @@ dberr_t file_os_io::close() noexcept
   return DB_SUCCESS;
 }
 
-dberr_t file_os_io::read(os_offset_t offset, span<byte> buf) noexcept
+dberr_t MY_ATTRIBUTE((warn_unused_result)) file_os_io::read(os_offset_t offset, span<byte> buf) noexcept
```
dr-m, line 283: To avoid too long lines, put the attribute on a line on its own, before the return type. I think that the attribute here is redundant, and the attribute in the member function declaration should be sufficient.

storage/innobase/log/log0log.cc
```diff
@@ -382,7 +382,8 @@ class file_pmem_io final : public file_io
                                                                    : DB_ERROR;
   }
   dberr_t close() noexcept final { return m_file.unmap(); }
-  dberr_t read(os_offset_t offset, span<byte> buf) noexcept final
+  dberr_t MY_ATTRIBUTE((warn_unused_result)) read(os_offset_t offset,
+                                                  span<byte> buf) noexcept final
```
dr-m, line 386: Please add the attribute on a separate line, right before the function return type `dberr_t`. It’s nicer for `grep` if the main function signature fits in a single line.

storage/innobase/log/log0recv.cc
```diff
@@ -762,14 +762,15 @@ void recv_sys_t::open_log_files_if_needed()
...
+dberr_t
+recv_sys_t::read(os_offset_t total_offset, span<byte> buf)
 {
   open_log_files_if_needed();
 
   size_t file_idx= static_cast<size_t>(total_offset / log_sys.log.file_size);
   os_offset_t offset= total_offset % log_sys.log.file_size;
-  dberr_t err= recv_sys.files[file_idx].read(offset, buf);
-  ut_a(err == DB_SUCCESS);
+  return recv_sys.files[file_idx].read(offset, buf);
```
dr-m, line 773: While you are at this, you might want to include parts of c72ddeea4f5400f02b65df19c2e2c02da8166c1f, to make sure that the first log file is only registered in one place, and that `recv_sys.close_files()` will actually close the files.

dr-m (commented):
Thank you, the high-level idea looks good. Most of this code is covered by the existing test `innodb.log_corruption`.

storage/innobase/log/log0recv.cc
dr-m: Sorry, I did not realize that I had refactored `log_file_t` in 685d958e38b825ad9829be311f26729cccf37c46. Please disregard my previous comment.

storage/innobase/log/log0recv.cc
```diff
@@ -1272,6 +1274,7 @@ inline uint32_t log_block_calc_checksum_format_0(const byte *b)
 ATTRIBUTE_COLD static dberr_t recv_log_recover_pre_10_2()
 {
   uint64_t max_no= 0;
+  dberr_t err;
```
dr-m, line 1277: Please use code similar to `recv_log_recover_10_4()` here, and do not declare a local variable for the entire function scope.

storage/innobase/log/log0recv.cc
dr-m: The line becomes too long. To avoid that and to be somewhat in line with the old-style formatting, I would suggest the following:
```c++
	if (dberr_t err =
            recv_sys.read(source_offset & ~lsn_t(OS_FILE_LOG_BLOCK_SIZE - 1),
			  {buf, OS_FILE_LOG_BLOCK_SIZE})) {
		return err;
	}
```

storage/innobase/log/log0recv.cc
```diff
@@ -762,14 +762,15 @@ void recv_sys_t::open_log_files_if_needed()
   }
 }
 
-void recv_sys_t::read(os_offset_t total_offset, span<byte> buf)
+MY_ATTRIBUTE((warn_unused_result))
+dberr_t
+recv_sys_t::read(os_offset_t total_offset, span<byte> buf)
```
dr-m, line 767: The return type should fit in the same line here. Only in the old InnoDB formatting style the return type was put in a separate line. The code is nicer to `grep` if the entire function signature (possibly excluding attributes) fits in a single line.

storage/innobase/include/log0log.h
```diff
@@ -506,7 +506,8 @@ struct log_t{
     /** reads buffer from log file
     @param[in]	offset		offset in log file
     @param[in]	buf		buffer where to read */
-    void read(os_offset_t offset, span<byte> buf);
+    dberr_t MY_ATTRIBUTE((warn_unused_result)) read(os_offset_t offset,
+                                                    span<byte> buf);
```
dr-m, line 510: I’d add the attribute in a separate line and define this function inline, because the implementation was simplified:
```c++
    MY_ATTRIBUTE((warn_unused_result))
    /** reads buffer from log file
    @param[in]	offset		offset in log file
    @param[in]	buf		buffer where to read
    @return error code */
    dberr_t read(os_offset_t offset, span<byte> buf)
    { return fd.read(offset, buf); }
```

dr-m (approved):
Thank you. OK to push after addressing my remaining comments.

PR #2788, opened 2023-10-16
MDEV-31272 Statement rollback causes empty writeset replication

sciascid:
This patch fixes cases where a transaction caused empty writeset to be replicated. This could happen in the case where a transaction executes a statement that initially manages to modify some data and therefore appended keys some for  certification. The statement is however rolled back at some later stage due to some error (for example, a duplicate key error). After statement rollback the transaction is still alive, has no other changes. When committing such transaction, an empty writeset was replicated through Galera.

The fix is to avoid calling into commit hook only when transaction has appended one or keys for certification *and* has some data in binlog cache to replicate. Otherwise, the commit is considered empty, and goes through usual empty commit path.

### DIFF
--- a/mysql-test/suite/galera/r/MDEV-31272.result
+++ b/mysql-test/suite/galera/r/MDEV-31272.result
@@ -1,0 +1,36 @@
+connection node_2;
+connection node_1;
+connection node_1;
+CREATE TABLE t1 (f1 int primary key, f2 int);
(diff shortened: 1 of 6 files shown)


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/pull/2788/commits/aaaa819c1b6014ab184ceba3529426b456a14172

PR #2789, opened 2023-10-16
MDEV-32050 UNDO logs still growing for write-intensive workloads

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32050*

## Description
Multi-version concurrency control in InnoDB requires that undo log records from committed transactions be preserved until the oldest read view that could “unsee” any committed changes has been closed. The `purge_sys.view` usually points to the oldest active view. These changes fix a performance issue with the purge of committed transaction history, reducing the “purge lag” or the “history list length” (number of committed transactions to purge); concurrency in the purge subsystem is improved instead.

The main improvements are as follows:
 * The `purge_coordinator_task` will no longer free any undo log pages, except when `innodb_undo_log_truncate=ON`. A new `purge_truncation_task` may execute concurrently with the purge coordinator, freeing pages that were processed in the previous purge batch. The previous throttling parameter `innodb_purge_rseg_truncate_frequency` will be deprecated and ignored.
 * The `purge_coordinator_task` will look up and cache all tables and undo log pages that are to be processed in the current purge batch. Undo log records will no longer be copied. This reduces contention on `dict_sys.latch`, the MDL subsystem, as well as `buf_pool.page_hash`.
 * During `innodb_fast_shutdown=0` (slow shutdown), the maximum `innodb_purge_batch_size` will be used.

## How can this PR be tested?

Stress tests with RQG as well as performance tests are needed. The changes related to MDL and DDL may potentially introduce hangs. Some adjustments may be necessary.

## PR quality check


dr-m:
I had pushed this accidentally to 10.6 instead of another branch. This is still pretty much open.

Test status:
* We have hangs that seem to involve `LOCK TABLES` and metadata locks (MDL). I  This could be an existing bug with MDL; maybe the deadlock detection there does not work? I do not see how 5128d04b63717a5d843dea6feb78dd2aa44a2c33 could cause it. We’ll do one more 4-hour stress test run without `LOCK TABLES`.
* We may have to increase the maximum value of `innodb_purge_batch_size` as well as the default value, to reduce the history list length. A performance test run with 200000 (instead of the current maximum of 5000) is in progress.

dr-m:
@vaintroub, can you please review the MDL change 5128d04b63717a5d843dea6feb78dd2aa44a2c33?

FooBarrior:
Apparently, commit 5128d04b6 has an incorrect caption. Allow cloning of shared MDL? Should be like "Allow table to be guarded by an MDL lock from another thread"

dr-m:
@FooBarrior That’s better; I’ll change that when I rebase the branch again for the final review.

vaintroub:
Re https://github.com/MariaDB/server/commit/5128d04b63717a5d843dea6feb78dd2aa44a2c33, I do not have much to say, can only suggest a better name for "emissary". While it is neat, and ´perhaps a sentimental reference to USSR action/spy movies, I fail to understand it's meaning in the context of MDL . I suggest something more dull, like "owner" maybe?

FooBarrior:
@vaintroub there is already `MDL_context_owner *m_owner`, which is ultimately a thd holding thid mdl context.

dr-m:
#2792 addresses @vaintroub’s comment by using the names `MDL_context::lock_warrant` and `MDL_context::is_lock_warrantee()`.

PR #2792, opened 2023-10-19
MDEV-32050 UNDO logs still growing for write-intensive workloads

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32050*

## Description
Multi-version concurrency control in InnoDB requires that undo log records from committed transactions be preserved until the oldest read view that could “unsee” any committed changes has been closed. The `purge_sys.view` usually points to the oldest active view. These changes fix a performance issue with the purge of committed transaction history, reducing the “purge lag” or the “history list length” (number of committed transactions to purge); concurrency in the purge subsystem is improved instead.

The main improvements are as follows:

* The `purge_coordinator_task` will no longer free any undo log pages, except when `innodb_undo_log_truncate=ON`. A new `purge_truncation_task` may execute concurrently with the purge coordinator, freeing pages that were processed in the previous purge batch. The previous throttling parameter `innodb_purge_rseg_truncate_frequency` will be deprecated and ignored, and each purge batch will involve an attempt of freeing undo log pages and shrinking the history list.
* The `purge_coordinator_task` will look up and cache all tables and undo log pages that are to be processed in the current purge batch. Undo log records will no longer be copied. This reduces contention on `dict_sys.latch`, the MDL subsystem, as well as `buf_pool.page_hash`.
* During `innodb_fast_shutdown=0` (slow shutdown), the maximum `innodb_purge_batch_size` will be used to maximize throughput.

## How can this PR be tested?
Stress tests with RQG as well as performance tests have been conducted.

Based on performance tests, we might want to increase the default value of `innodb_purge_batch_size` from the current 300. Increasing the current maximum of 5000 undo log pages does not seem to have much impact.

The changes related to MDL and DDL may potentially introduce hangs. So far, the only observed hangs have been unrelated to these changes, due to something outside InnoDB, for example, [MDEV-31393](https://jira.mariadb.org/browse/MDEV-31393) as well a MDL deadlock between two `LOCK TABLES` statements. We got one core dump of a hang that was caused by an apparent acquisition of a `rseg->latch` that was never released, but no `rr replay` trace for it. This could theoretically be related to 64b77bc664a5885babd53fc0c3b85427d1a2b256.

## PR quality check


dr-m:
I got a core dump of a 10.6 hang (without these changes) where the culprit is the `purge_coordinator_task` that is waiting for another shared MDL on a table on which a DDL operation has started to wait for an exclusive MDL:
```
#6  0x0000559b327026ca in MDL_context::acquire_lock (this=this@entry=0x559b34d89460, mdl_request=mdl_request@entry=0x7f4b5f71c780, lock_wait_timeout=<optimized out>) at /data/Server/10.6/sql/mdl.h:482
#7  0x0000559b32ed2057 in dict_acquire_mdl_shared<false, true> (table=table@entry=0x7f4b100c6f00, thd=thd@entry=0x559b34d89308, mdl=mdl@entry=0x559b34e3d668, table_op=table_op@entry=DICT_TABLE_OP_NORMAL) at /data/Server/10.6/sql/mdl.h:610
#8  0x0000559b32ed21ea in dict_table_open_on_id<true> (table_id=4423, dict_locked=dict_locked@entry=false, table_op=table_op@entry=DICT_TABLE_OP_NORMAL, thd=0x559b34d89308, mdl=mdl@entry=0x559b34e3d668) at /usr/include/c++/9/bits/atomic_base.h:413
#9  0x0000559b32d4cc4d in row_purge_parse_undo_rec (node=node@entry=0x559b34e3d4e0, undo_rec=undo_rec@entry=0x7f4ac40463f8 "", thr=thr@entry=0x559b34e3d300, updated_extern=updated_extern@entry=0x7f4b5f71ca6f) at /data/Server/10.6/storage/innobase/row/row0purge.cc:1082
#10 0x0000559b32d54e54 in row_purge (node=node@entry=0x559b34e3d4e0, undo_rec=undo_rec@entry=0x7f4ac40463f8 "", thr=thr@entry=0x559b34e3d300) at /data/Server/10.6/storage/innobase/row/row0purge.cc:1260
#11 0x0000559b32d54fec in row_purge_step (thr=thr@entry=0x559b34e3d300) at /data/Server/10.6/storage/innobase/row/row0purge.cc:1326
#12 0x0000559b32cc548f in que_thr_step (thr=0x559b34e3d300) at /data/Server/10.6/storage/innobase/que/que0que.cc:588
#13 que_run_threads_low (thr=thr@entry=0x559b34e3d300) at /data/Server/10.6/storage/innobase/que/que0que.cc:644
#14 0x0000559b32cc56fa in que_run_threads (thr=0x559b34e3d300) at /data/Server/10.6/storage/innobase/que/que0que.cc:664
#15 0x0000559b32dba49e in trx_purge (n_tasks=<optimized out>, history_size=history_size@entry=15985) at /data/Server/10.6/storage/innobase/trx/trx0purge.cc:1303
```
3a10d062d70c6860d909981b54a2fb79a8848cc6 fixes this by avoiding the table lookup and MDL wait at the low level. In this scenario, `purge_sys_t::close_and_reopen()` would release all MDL to allow the DDL operation to proceed, and then try to acquire them again.

storage/innobase/srv/srv0srv.cc
```diff
@@ -1304,12 +1304,31 @@ static tpool::waitable_task purge_truncation_task
...
-      ++purge_state.m_running == 1)
-    srv_thread_pool->submit_task(&purge_coordinator_task);
+  if (enabled() && !paused() && !purge_state.m_running)
+  {
+    if (srv_undo_log_truncate);
+    else if (!trx_sys.history_exists())
+      return;
+    else
+    {
+      purge_coordinator_timer->disarm();
```
vaintroub, line 1299: Can we move this complicated logic/locking into purge_coordinator_task? basically doing as before? A false wakeup would not happen very often with the old logic either?

dr-m: Done in ee1ff06819b624446b50cb3c14019b7c15469498

storage/innobase/trx/trx0purge.cc
```diff
@@ -1409,6 +1381,12 @@ static void trx_purge_wait_for_workers_to_complete()
 TRANSACTIONAL_INLINE
 void purge_sys_t::clone_end_view(const purge_sys_t::iterator &head)
 {
+  /* Release the undo pages. */
+  for (auto p : pages)
```
vlad-lesin, line 1381: There is only one caller of clone_end_view(). And it's said that it just updates end_view in the documentation. It does not look like a callback to clean everything after purging. I propose to move the code of pages cleaning to clone_end_view() caller.

storage/innobase/include/row0purge.h
```diff
@@ -89,155 +89,69 @@ struct trx_purge_rec_t
...
+  /** memory heap used as auxiliary storage; must be emptied between rows */
+  mem_heap_t *heap;
+  /** persistent cursor to the clustered index record */
+  btr_pcur_t pcur;
+
+  /** Undo recs to purge */
+  std::queue<trx_purge_rec_t> undo_recs;
+
+  /** map of table identifiers to table handles and meta-data locks */
+  std::unordered_map<table_id_t, std::pair<dict_table_t*,MDL_ticket*>> tables;
```
vlad-lesin, line 128: Is there any sense to point buckets count explicitly here? We don't know what exactly value stl proposes for us by default.

vlad-lesin (approved):
The code looks good to me. There are some minor notes, but I did not find any errors in the logic. It's not obvious of when purge_sys.rseg->latch is acquired and released, because it's spreader among functions, and the callers of acquirer are responsible to release the latch, it takes time to understand what exactly callers do this, but it's not impossible, besides, this is the common pattern in this part of code. It's also good to have such big changes split into several commits, where each commit is well described in commit message.

vaintroub (approved):
It looks good, but I think needs to go through performance testing before pushed, in purge intensive large workloads.
Some changes are obviously improving performance (like this MDL caching), some changes   could turn both ways.  In the past,  iirc, there was a 10 ms pause between purge batches, which is now gone together with the purge timer. While cleanup is good, the full effect of it is still unknown (maybe this pause relaxed the contention between foreground and background threads), so it  would be good to measure before push

storage/innobase/trx/trx0purge.cc
dr-m: The reason why I added the cleanup here is that the single caller is not a member function of `purge_sys_t`. We could add another single-caller function to do the cleanup, or maybe a better solution would be to rename the `clone_end_view()` to `batch_cleanup()` or something. The name `purge_sys_t::clone_end_view()` was introduced in ab0190101b0587e0e03b2d75a967050b9a85fd1b in 10.6.

storage/innobase/include/row0purge.h
dr-m: I checked that `libstdc++` 6.0.32 would default to `_M_h._M_bucket_count=1` and supposedly resize and rehash on demand. e4e00a07bda3ceb0b2dc569f9cb13860b1c0bafa introduces `constexpr size_t TRX_PURGE_TABLE_BUCKETS=128`, which I hope will be a reasonable compromise.

storage/innobase/trx/trx0purge.cc
```diff
@@ -67,8 +67,9 @@ TrxUndoRsegsIterator::TrxUndoRsegsIterator()
 /** Sets the next rseg to purge in purge_sys.
 Executed in the purge coordinator thread.
 @return whether anything is to be purged */
-TRANSACTIONAL_INLINE inline bool TrxUndoRsegsIterator::set_next()
+inline bool TrxUndoRsegsIterator::set_next()
```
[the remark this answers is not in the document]
dr-m: Done in 9ac62e81be6564596c3f92a47166d42c9d89eca0.

storage/innobase/trx/trx0purge.cc
dr-m: 9ac62e81be6564596c3f92a47166d42c9d89eca0 renames `clone_end_view()` to `batch_cleanup()`.

dr-m:
> It looks good, but I think needs to go through performance testing before pushed, in purge intensive large workloads. 
Some changes are obviously improving performance (like this MDL caching), some changes could turn both ways.

@vaintroub, you are absolutely right about testing this. I mainly tested the performance of `innodb_fast_shutdown=0` after some short and write intensive `sysbench` workloads, because that was a quick and easy test for me.

@andrelkin has tested this with replication, and the history list length in his test remained around 70 committed transactions, after he increased `innodb_purge_threads` and `innodb_purge_batch_size` so that the purge could keep up with the workload.

In fact, the MDL and `dict_table_t*` caching turned out to have a _negative_ impact in @hgxl64’s TPC-C like testing, compared to the first changes that mainly reduced operations on `purge_sys->rseg.latch`. As far as I understand, all in all this is a clear improvement. Instead of seeing a monotonically growing history list length, we see that the purge gets a chance every now and then. We definitely need to identify why there are some “purge stalls” (maybe it’s just due to old read views in the workload?) and do more in this area.

>In the past, iirc, there was a 10 ms pause between purge batches, which is now gone together with the purge timer. While cleanup is good, the full effect of it is still unknown (maybe this pause relaxed the contention between foreground and background threads), so it would be good to measure before push

@hgxl64 retested the TPC-C workload after c0fa0e1cd01b5b47a8c8bfcab97d159041f5b60f, and there was no significant change. It could be good to test some simpler workloads as well.

PR #2793, opened 2023-10-20
MDEV-32024 : Galera library 26.4.16 fails with every server version

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-32024*

## Description

Problem was that total order isolation (TOI) is started before we know sequence implementing storage engine. This led to situation where table implementing persistent storate for sequence in case of MyISAM was created on applier causing errors later in test execution.

Therefore, in both CREATE SEQUENCE and ALTER TABLE to implementing persistent storage we need to check implementing storage engine after open_tables and this check must be done in both master and applier, because if implementing storage engine is MyISAM it does not support rollback.

Added tests to make sure that if sequence implementing storage engine is MyISAM or we try to alter it to MyISAM user gets error and changes are not replicated.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/galera_concurrent_ctas.result
+++ b/mysql-test/suite/galera/r/galera_concurrent_ctas.result
@@ -1,5 +1,30 @@
 connection node_2;
 connection node_1;
+connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1;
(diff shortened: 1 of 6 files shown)


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/e913f4e11e1e519196f276d7c5689f653e724547

PR #2798, opened 2023-10-23
MDEV-30260: Slave crashed:reload_acl_and_cache during shutdown

bnestere:
The signal handler thread can use various different runtime
resources when processing a SIGHUP (e.g. master-info information)
due to calling into reload_acl_and_cache(). Currently, the shutdown
process waits for the termination of the signal thread after
performing cleanup. However, this could cause resources actively
used by the signal handler to be freed while reload_acl_and_cache()
is processing.

Where it is still desirable for the signal handler to receive
signals during the shutdown process, e.g. to process kill -9 on a
slow shutdown; it must be ensured that a SIGHUP has finished
processing before performing any cleanup related activities. This
patch adds in additional synchronization such that the server will
wait for any active SIGHUP processing to finish before killing any
connections or starting cleanup of resources.

This PR consists of two commits:
1. The regression which shows the crash
2. The patch itself


vaintroub (commented):
please add --source include/not_windows.inc to the test, so that if we ever run rpl suite on Windows, it will be properly skipped.

sql/mysqld.cc
```diff
@@ -2084,16 +2089,31 @@ static void clean_up(bool print_message)
 */
 static void wait_for_signal_thread_to_end()
 {
-  uint i;
+  uint i, n_waits= DBUG_EVALUATE("force_sighup_processing_timeout", 5, 100);
+  int err;
```
vaintroub, line 2093: should be initialized `int err=ESRCH;`  or so.  Now fails on Windows with uninitialized variable `err` in line 2106 . The variable is not initialized, because `signal_thread_in_use` is never set, and loop body never runs.  That's because Windows does not use signal_hand thread,  entire shutdown sequence runs in the main thread (there is also no sigwait or SIGHUP here either). So this function should be no-op on Windows, but it could also be no-op if signal thread did yet start, who knows.

sql/mysqld.cc
```diff
@@ -1730,6 +1730,12 @@ static void close_connections(void)
   while (CONNECT::count)
     my_sleep(100);
 
+  /*
+    If the signal thread is actively using server resources, let it finish
+    before killing any connections
+  */
+  wait_for_signal_thread_to_end();
```
vaintroub, line 1737: I think it makes more sense to put that one at the very end of the function. Advantages are:
- it shuts down use/background THDs earlier, rather than waiting in worst case for 10 sec before informing them about shutdown.
- there is less CPU load at the end of this function (all THDs are shut down), so chances are high 10 seconds will be enough for  `wait_for_signal_thread_to_end()`  to finish, even if it had to flushing table cache.

sql/mysqld.cc
```diff
@@ -2084,16 +2089,22 @@ static void clean_up(bool print_message)
 */
 static void wait_for_signal_thread_to_end()
 {
-  uint i;
+  uint i, n_waits= DBUG_EVALUATE("force_sighup_processing_timeout", 5, 100);
   /*
     Wait up to 10 seconds for signal thread to die. We use this mainly to
     avoid getting warnings that my_thread_end has not been called
```
vaintroub, line 2096: The comment might be a little outdated by now. There is a potentially long-ish operation SIGHUP handler that needs to be waited on.

vaintroub:
I like current state, once it is squashed and rebased.

andrelkin (changes requested):

andrelkin (approved):
The whole patch looks good.
I suggest to make into the commit message the gist of the issue and the fixes.
In my version:
the real issue is a race between `Main_thread.end_slave().delete active_mi` vs `Signal_handler.Master_info_index::flush_all_relay_logs.read(active_mi)`.
Your fixes make sure the destructive latter is unreachable, while the signal handler thread is around, thanks to `wait_for_signal_thread_to_end()` relocation to in front of end_slave().

bnestere:
Closing this PR, as the approved patch was pushed into 10.5 as 952ab9a59659b28f10ce2770e17249e0ff5b0681

PR #2799, opened 2023-10-24
MDEV-32549 Cluster inconsistent after SAVEPOINT is rolled back

sciascid:
Attempting to set a SAVEPOINT when one of the involved storage engines does not support savepoints, raises an error, and results in statement rollback. If Galera is enabled with binlog emulation, the above scenario was not handled correctly, and resulted in cluster wide inconsistency.

The problem was in wsrep_register_binlog_handler(), which is called towards the beginning of SAVEPOINT execution. This function is supposed to mark the beginning of statement position in trx cache through `set_prev_position()`. However, it did so only on condition that `get_prev_position()` returns `MY_OFF_T_UNDEF`. This before statement position is typically reset to undefined at the end of statement in `binlog_commit()` / `binlog_rollback()`. However that's not the case with Galera and binlog emulation, for which binlog commit / rollback hooks are not called due to the optimization that avoids internal 2PC (MDEV-16509).

### DIFF
--- a/mysql-test/suite/galera/r/MDEV-32549.result
+++ b/mysql-test/suite/galera/r/MDEV-32549.result
@@ -1,0 +1,24 @@
+connection node_2;
+connection node_1;
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY) engine=innodb;
+CREATE TABLE t2 (f1 INTEGER PRIMARY KEY) engine=aria;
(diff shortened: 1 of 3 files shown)


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/362c0950e8999a88f5866fc7361e10615364e4a3

PR #2800, opened 2023-10-24
Disable ps-protocol second execution on test that do not support it

sciascid:
MDEV-31003 has introduced second execution for SELECTs that execute under ps-protocol. The following tests in galera suites do not support this mode of execution, disable it:

  galera.MDEV-27862
  galera.galera_log_output_csv
  galera.galera_query_cache
  galera.galera_query_cache_sync_wait
  galera_3nodes_sr.GCF-336
  galera_3nodes_sr.galera_sr_isolate_master
  galera_sr.galera_sr_large_fragment
  galera_sr.galera_sr_many_fragments

### DIFF
--- a/mysql-test/suite/galera/t/MDEV-27862.test
+++ b/mysql-test/suite/galera/t/MDEV-27862.test
@@ -1,6 +1,8 @@
 --source include/galera_cluster.inc
 --source include/have_innodb.inc
 
+--disable_ps2_protocol
+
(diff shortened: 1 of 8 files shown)


sysprg:
Thanks, fix merged with head revision as https://github.com/MariaDB/server/commit/c554f2683201df4731b1e3c61fb5011d7d8bbb86

PR #2801, opened 2023-10-24
MDEV-32568: Fix wrong use of my_read() return type.

mariadb-LuisLizardo:
- [x] *The Jira issue number for this PR is: MDEV-32568*

## Description
The `my_read()` function currently returns an unsigned `size_t` value. However, in cases of errors during file reading, an error state is signaled by a macro constant named `MY_FILE_ERROR`, which is set to the maximum value of `size_t` (essentially -1 in two's complement representation). Unfortunately, this unconventional approach has led to widespread confusion and misuse of the `my_read()` function throughout the codebase. This code change fixes all the callers that mistakenly used the `my_read()` return type.

## How can this PR be tested?

## PR quality check


libmysqld/libmysql.c
```diff
@@ -619,17 +619,17 @@ static int default_local_infile_init(void **ptr, const char *filename,
...
 
-  if ((count= (int) my_read(data->fd, (uchar *) buf, buf_len, MYF(0))) < 0)
+  const size_t bytes_count= my_read(data->fd, (uchar *) buf, buf_len, MYF(0));
+  if (bytes_count == MY_FILE_ERROR)
   {
     data->error_num= EE_READ; /* the errmsg for not entire file read */
     my_snprintf(data->error_msg, sizeof(data->error_msg)-1,
 		EE(EE_READ),
 		data->filename, my_errno);
   }
```
grooverdan, line 631: this should be returning < 0 on error. I'm not sure casting does this, but even if it does, its not particularly clear.

extra/replace.c
```diff
@@ -996,8 +996,7 @@ static int fill_buffer_retaining(File fd, int n)
     return 0;
 
   /* Read in new stuff. */
-  if ((i=(int) my_read(fd, (uchar*) buffer + bufbytes,
-                       (size_t) bufread, MYF(MY_WME))) < 0)
+  if (my_read(fd, (uchar*) buffer + bufbytes, (size_t) bufread, MYF(MY_WME)) == MY_FILE_ERROR)
```
grooverdan, line 999: from builder:
```
/home/buildbot/amd64-ubuntu-2204-debug-ps/build/extra/replace.c: In function ‘fill_buffer_retaining’:
/home/buildbot/amd64-ubuntu-2204-debug-ps/build/extra/replace.c:1003:6: error: ‘i’ may be used uninitialized [-Werror=maybe-uninitialized]
 1003 |   if (i == 0 && bufbytes > 0 && buffer[bufbytes - 1] != '\n')
 ```

File: sql/ha_partition.cc [hunk 3]
```diff
@@ -3190,17 +3190,21 @@ int ha_partition::read_par_file(const char *name)
...
     }
     connect_string.length= uint4korr(buffer);
     connect_string.str= tmp= (char*) alloc_root(&m_mem_root,
                                                 connect_string.length+1);
-    if (my_read(file, (uchar*) connect_string.str, connect_string.length,
-                MYF(MY_NABP)))
+
+    bytes_read= my_read(file, (uchar*) connect_string.str, connect_string.length, MYF(MY_NABP));
+    if (bytes_read != MY_FILE_ERROR && bytes_read > 0 )
+    {
```
grooverdan, line 3205: A bit beyond what you where changing but looks like this should be a `goto err2`?

grooverdan (commented):
Hi, I assume this is going to be rebased back?

libmysqld/libmysql.c
mariadb-LuisLizardo: The logic is the same. It is returning -1 on line 632 after casting. 
I can add a comment there explaining though.

mariadb-LuisLizardo: Or return -1 explicitly inside the if-block.

midenok:
> Hi, I assume this is going to be rebased back?

@grooverdan To avoid merge conflicts it's better pushed to 10.4. but this is against stable policy. @vuvova Does this patch fit to 10.4?

sql/ha_partition.cc
mariadb-LuisLizardo: @grooverdan to be honest, I don't know. 
The if-block is asserting that some bytes were read from the connection arguments, so it doesn't look like it is an error to return 2.  By breaking the loop it returns 0.  Just like the other read attempt on line 3193. 
Its up to you.

extra/replace.c
mariadb-LuisLizardo: fixed

libmysqld/libmysql.c
grooverdan: I do like an explicit -1.

Compilers warn on existing casts depending on [arch/compiler/version](https://godbolt.org/z/oa6onEj9e).

sql/ha_partition.cc
grooverdan: Given the other accesses on `m_connect_string` imply something is there an error jump seems prudent. Monty seems to be the only one who's touched this code recently so I'd check with him.

grooverdan:
> > Hi, I assume this is going to be rebased back?
> 
> @grooverdan To avoid merge conflicts it's better pushed to 10.4. but maybe this is against stable policy. @vuvova Does this patch fit to 10.4?

You haven't actually changed the prototype definition, just doing some bug fixes and error handling to ensure the errors are handled. So its a bug fix.

seems my_read (and probably others) are suitable for the attribute `warn_unused_result` (and ` nonnull`) would catch some missing ones in the merge up.

grooverdan:
and obviously my_write equally flawed:
```
27	  uint errors;
28	  DBUG_ENTER("my_write");
29	  DBUG_PRINT("my",("fd: %d  Buffer: %p  Count: %lu  MyFlags: %lu",
30			   Filedes, Buffer, (ulong) Count, MyFlags));
31	  errors= 0; written= 0;
32	  if (!(MyFlags & (MY_WME | MY_FAE | MY_FNABP)))
33	    MyFlags|= my_global_flags;
34	
35	  /* The behavior of write(fd, buf, 0) is not portable */
36	  if (unlikely(!Count))
(gdb) n
36	  if (unlikely(!Count))
(gdb) 
49	    writtenbytes= write(Filedes, Buffer, Count);
(gdb) 
62	    if (writtenbytes == Count)
(gdb) p writtenbytes
$1 = 18446744073709551615
(gdb) p Count
$2 = 22
(gdb) list
57	                     if (!errors) {
58	                       errno= ENOSPC;
59	                       writtenbytes= (size_t) -1;
60	                     });
61	
62	    if (writtenbytes == Count)
63	      break;
64	    if (writtenbytes != (size_t) -1)
65	    {						/* Safeguard */
66	      written+= writtenbytes;
(gdb) n
64	    if (writtenbytes != (size_t) -1)
(gdb) info locals
errors = 0
written = 0
writtenbytes = 18446744073709551615
(gdb) n
70	    my_errno= errno;
(gdb) n
74	    if (my_thread_var->abort)
(gdb) n
77	    if ((my_errno == ENOSPC || my_errno == EDQUOT) &&
(gdb) n
85	    if ((writtenbytes == 0 || writtenbytes == (size_t) -1))
(gdb) n
87	      if (my_errno == EINTR)
(gdb) n
94	      if (!writtenbytes && !errors++)		/* Retry once */
(gdb) n
106	    if (MyFlags & (MY_NABP | MY_FNABP))
(gdb) n
117	  if (MyFlags & (MY_NABP | MY_FNABP))
(gdb) n
119	  DBUG_RETURN(writtenbytes+written);
(gdb) info locals
errors = <optimized out>
written = 0
writtenbytes = <optimized out>
(gdb) finish
Run till exit from #0  my_write (Filedes=11, Buffer=<optimized out>, Count=<optimized out>, MyFlags=16)
    at /home/dan/repos/mariadb-server-10.11/mysys/my_write.c:119
0x0000000000cd572d in mem_pressure::setup (this=0x201f1a8 <mem_pressure_obj>)
    at /home/dan/repos/mariadb-server-10.11/storage/innobase/buf/buf0buf.cc:792
792	      if (my_write(m_fds[m_num_fds].fd, *trig, slen, MYF(MY_WME)) < slen)
Value returned is $3 = 18446744073709551615
(gdb) p my_errno
No symbol "my_errno" in current context.
(gdb) p errno
$4 = 22
```

libmysqld/libmysql.c
grooverdan: approved with this change.

midenok:
If this is a bugfix there must be test cases in commit. @mariadb-LuisLizardo can you make some?

libmysqld/libmysql.c
mariadb-LuisLizardo: I added the `return -1` inside the if block

mariadb-LuisLizardo:
> If this is a bugfix there must be test cases in commit. @mariadb-LuisLizardo can you make some?

@midenok I don't know how to do so. I don't know how to trigger it to write a mtr. 
Service Now reported this problem without seeing any crash or bug itself, but by just spotting this in the code.

grooverdan:
> > If this is a bugfix there must be test cases in commit. @mariadb-LuisLizardo can you make some?
> 
> @midenok I don't know how to do so. I don't know how to trigger it to write a mtr.

Short of a fuse filesystem (that I haven't found implemented), NFS or some heavy real world test I can't think of a non-debug build test.

Its partially already instrumented in debug builds with `DBUG_EXECUTE_IF ("simulate_file_read_error"`. No idea who though ENOSPC was a read error but alas it doesn't matter too much, and can be changed, because nothing currently uses it.

With the implementation on `my_read` I'd move this to before the actual read and return so the buffer isn't populated.

Debug builds would trigger this with:
```
SET @@global.debug_dbug="d,simulate_file_read_error";
```

Also:

Seems my_read (and probably others) are suitable for the attribute warn_unused_result (and nonnull for its buffer parameter) would catch missing error handing in the merge up. (though I suggest testing and not relying on the merger).

PR #2803, opened 2023-10-26
MDEV-32589 FULL_NODUP mode for binlog_row_image

SongLibing:
This patch provides a new mode FULL_NODUP to binlog_row_image system
variable. With FULL_NODUP mode, all columns are included in before
image, but only updated columns are included in after image for UPDATE.
While all columns are included in the after image for INSERT.

FULL_NODUP is for replacing FULL mode. It includes all data of
the before and after image as FULL mode, but it uses less storage
especially in the case that only a few columns are updated.

Note: It will binlog full before and after image for all modes if the
      table has no primary key. FULL_NODUP follows the behavior.

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check


LinuxJedi (approved):
Many thanks for this contribution. In my first pass this looks good to me. @andrelkin or has said he or @bnestere will give it a second pass next week.

bnestere:
Thanks for the contribution, @SongLibing !

I haven't gone through the code yet (I think @andrelkin is planning on doing the actual review), but an initial thought comes to mind that the term `FULL_DIFFERENT` is a little awkward and ambiguous/not self-explanatory. A convention already somewhat exists with `NOBLOB`, as it writes all columns except unnecessary blob-like ones. Perhaps we could change this to something like `NOSTAGNANT` or `NOUNCHANGED` to indicate that unchanged columns aren't written in the output :thinking:. Though note I'm not set on this either way.

SongLibing:
Hi  @bnestere , thanks for your feedback. FULL_DIFFERENT just express that before image is full and after image only includes the changed. I agree with you that it is not as good as NOBLOB.  How about NODUP, it can explain the idea that there is no dup values between before and after image?

andrelkin:
I'd go with `FULL_AFTER_IMAGE`.

sql/table.cc
```diff
@@ -7820,6 +7820,11 @@ void TABLE::mark_columns_needed_for_insert()
   if (vfield)
     mark_virtual_columns_for_write(TRUE);
   mark_columns_per_binlog_row_image();
+
+  /* FULL_DIFFERENT is for replacing FULL mode, insert includes all columns. */
+  if (in_use->variables.binlog_row_image == BINLOG_ROW_IMAGE_FULL_DIFFERENT)
+    rpl_write_set= read_set;
```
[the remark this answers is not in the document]
SongLibing: As I mentioned in the comment, the purpose of this mode is to replace FULL mode. So FULL_DIFFERENT mode should includes all columns, thus users can use it just like FULL mode. E.g, people may replicate data from mariadb  other databases, they may have different behavior on default value. So all data should be binlogged and replicated.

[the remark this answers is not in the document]
andrelkin: I am fine with this clarification which needs documenting and I'll be offering in the patch lines.

To look at this issue more generally, perhaps we could consider for future a user-defined column replication, which also would cover the different defaults on master and slave. Something like `ALTER table `rpl_write_set` = (1,2,3,7,10... )`.

sql/sys_vars.cc
```diff
@@ -6773,15 +6773,20 @@ static Sys_var_mybool Sys_binlog_encryption(
...
                          PRIV_SET_SYSTEM_VAR_BINLOG_ROW_IMAGE>
 Sys_binlog_row_image(
        "binlog_row_image",
-       "Controls whether rows should be logged in 'FULL', 'NOBLOB' or "
-       "'MINIMAL' formats. 'FULL', means that all columns in the before "
-       "and after image are logged. 'NOBLOB', means that mysqld avoids logging "
+       "Controls whether rows should be logged in 'FULL', 'FULL_DIFFERENT', "
+       "'NOBLOB' or 'MINIMAL' formats. 'FULL', means that all columns in the "
+       "before and after image are logged. 'FULL_DIFFERENT', means that all "
+       "columns are logged in before image, but only changed columns are "
```
andrelkin, line 6812: Considering [clarification](https://github.com/MariaDB/server/pull/2803/commits/300ead12bfa08dd15d1dd15dc201371e0e605ff6#r1399923916), let's specify about the inserted columns. ? How about 
```
'FULL_DIFFERENT', means that all columns are logged in before image, but only updated columns or all columns of the inserted record are logged  in before image,
```

sql/sys_vars.cc
```diff
@@ -6773,15 +6773,20 @@ static Sys_var_mybool Sys_binlog_encryption(
        READ_ONLY GLOBAL_VAR(encrypt_binlog), CMD_LINE(OPT_ARG),
        DEFAULT(FALSE));
 
-static const char *binlog_row_image_names[]= {"MINIMAL", "NOBLOB", "FULL", NullS};
+static const char *binlog_row_image_names[]=
+{
+  "MINIMAL", "NOBLOB", "FULL", "FULL_DIFFERENT", NullS
```
andrelkin, line 6802: Trying out another option of the name. How about
`FULL_NODUP` , where `NODUP` obviously :innocent: stands for ... what the FR actually wants, right? (That is only changed cols in `AI`)

SongLibing: That is better.

sql/sys_vars.cc
SongLibing: sure, it will update.

sql/table.cc
```diff
@@ -7890,6 +7895,10 @@ void TABLE::mark_columns_per_binlog_row_image()
         /* Set of columns that should be written (all) */
         rpl_write_set= read_set;
         break;
+      case BINLOG_ROW_IMAGE_FULL_DIFFERENT:
+        bitmap_set_all(read_set);
+        rpl_write_set= versioned() ? &s->all_set : write_set;
```
andrelkin, line 7991: We need the same comment 
```
TODO: After MDEV-18432
```
 as on l.7941 (the patch line #), need not we?

andrelkin (approved):
The patch looks good. 
I would still invite @midenok at least to become aware of possible connection of this work with MDEV-18432.

mysql-test/suite/rpl/t/rpl_row_img_sequence_full_nodup.test
```diff
@@ -0,0 +1,51 @@
+#
+# Purpose:
+#   The rpl_row_img_sequence group of tests verify that sequence MDL updates,
+# i.e. NEXTVAL and SETVAL, respect the binlog_row_image variable value when
+# written into the binary log. In particular, it ensures that only changed
+# columns are written with MINIMAL image mode, and all columns are written
+# otherwise. This test focuses on validating the behavior of
+# binlog_row_img=FULL.
```
andrelkin, line 8: `= FULL_NODUP`, right?

SongLibing: right

andrelkin (approved):
Thanks for addressing the notes, LiBing!
I found one word to change in a test comments.

andrelkin (approved):
Looks good.

LinuxJedi (approved):
Rebased off 11.4 as that is where it should end up. All looks good to me!

PR #2804, opened 2023-10-27
MDEV-28430: Fix memory barrier missing of lf_alloc on Arm64

gelidNiu:
When testing MariaDB on Arm64, a stall issue will occur, jira link:
https://jira.mariadb.org/browse/MDEV-28430.

The stall occurs because of an unexpected circular reference in the
LF_PINS->purgatory list which is traversed in lf_pinbox_real_free().

We found that on Arm64, ABA problem in LF_ALLOCATOR->top list was not
solved, and various undefined problems will occur, including circular
reference in LF_PINS->purgatory list.

- [x] *The Jira issue number for this PR is: MDEV-28430*

## Description
When testing MariaDB on Arm64, a stall issue will occur, jira link: https://jira.mariadb.org/browse/MDEV-28430.

## How can this PR be tested?

Run unit.lf.

## PR quality check

### DIFF
--- a/mysys/lf_alloc-pin.c
+++ b/mysys/lf_alloc-pin.c
@@ -292,7 +292,7 @@ static int harvest_pins(LF_PINS *el, struct st_harvester *hv)
   {
     for (i= 0; i < LF_PINBOX_PINS; i++)
     {
-      void *p= el->pin[i];
+      void *p= my_atomic_loadptr((void **)&el->pin[i]);
       if (p)
         *hv->granary++= p;
     }
@@ -317,7 +317,7 @@ static int match_pins(LF_PINS *el, void *addr)
   LF_PINS *el_end= el+LF_DYNARRAY_LEVEL_LENGTH;
   for (; el < el_end; el++)
     for (i= 0; i < LF_PINBOX_PINS; i++)
-      if (el->pin[i] == addr)
+      if (my_atomic_loadptr((void **)&el->pin[i]) == addr)
         return 1;
   return 0;
 }
@@ -502,7 +502,8 @@ void *lf_alloc_new(LF_PINS *pins)
     {
       node= allocator->top;
      lf_pin(pins, 0, node);
-    } while (node != allocator->top && LF_BACKOFF());
+    } while (node != my_atomic_loadptr((void **)(char *)&allocator->top)
+             && LF_BACKOFF());
     if (!node)
     {
       node= (void *)my_malloc(allocator->element_size, MYF(MY_WME));


grooverdan (approved):
Looks really good and obviously is well thought out and passes tests.

Just going to run those unit tests a touch longer and under POWER too as a more rigorous test.

I do note the one outstanding issue that was pre-existing.
 
Thank you very much.

gelidNiu:
Rebase to the latest code of version 10.4, and update the commit message to be clearer.

I would like to ask about the testing situation. Is there anything we can help you with?

mysys/lf_alloc-pin.c
```diff
@@ -292,7 +292,7 @@ static int harvest_pins(LF_PINS *el, struct st_harvester *hv)
   {
     for (i= 0; i < LF_PINBOX_PINS; i++)
     {
-      void *p= el->pin[i];
+      void *p= my_atomic_loadptr((void **)&el->pin[i]);
```
dr-m, line 295: I see that each changed piece of code is changing a non-atomic read to an atomic one. In `include/atomic/gcc_builtins.h` this would expand to `__atomic_load_n((P), __ATOMIC_SEQ_CST)`. Can we use `my_atomic_loadptr_explicit` with something else than `MY_MEMORY_ORDER_SEQ_CST`, such as `MY_MEMORY_ORDER_ACQUIRE`?

I am aware that the rest of this subsystem is using the sequential consistent memory order. I also noticed that mysql/mysql-server@67936ecf9fd32b2e218c81e07f98c2390cab8e4d is using plain `std::atomic`, which defaults to `std::memory_order_seq_cst`.

Based on https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html anything else than the C++11 `std::memory_order_seq_cst` should not affect affect systems that implement total store order (which include typical implementations of the IA-32, AMD64 or SPARC ISA).

gelidNiu: If use MY_MEMORY_ORDER_ACQUIRE, there will be problem on some arm servers, as shown in link https://godbolt.org/z/jGczbEssf, when march=Armv8.3-A 
(1) The assembly code corresponding to C11 load-acquire is "ldapr", ldapr may be reordered before strl, which will cause dekker's algorithm to be invalid.
(2) The assembly code corresponding to C11 sec-cst is "ldar", and ldar will not be reordered before strl.
According to C11 memory model, C11's load-acquire may be reordered before store-release, while load-sec-cst will not be reordered before store-sec-cst.

This bug should be fixed based on the C11 memory model to ensure that it works on all platforms, e.g. whether the platform is based on Armv8.2-A or Armv8.3-A, etc. 
With C11 sec-cst semantics, dekker's algorithm can be guaranteed to work on all platforms.

dr-m: Thank you for the example. I created another one, which I think should correspond closely to the `match_pins()` function: https://godbolt.org/z/6W138nY5q
```c
int match_pins_seq_cst(void*** el, void *addr)
{
  int i;
  void ***el_end = el + 256;
  for (; el < el_end; el++)
    for (i= 0; i < 4; i++)
#if 1
      if (__atomic_load_n((*el)[i], __ATOMIC_SEQ_CST) == addr)
#else
      if ((*el)[i] == addr)
#endif
        return 1;
  return 0;
}
```
Only the `el++` in the outer `for` loop is different from the actual code, because it will advance the pointer by `sizeof(void*)` rather than `sizeof(LF_PINS)`. The inner `for` loop is equivalent to the original, because `offsetof(LF_PINS, pin) == 0`.

In my link, I intentionally chose GCC 4.8.5 for AMD64, because that is the oldest compiler that we currently target, because of Red Hat Enterprise Linux 7 and its derivatives. I also used `-O2` instead of `-O3`, because I seem to remember that it is what our own build scripts employ.

The two variants allocate a little different registers, and for some reason, the statement of interest would be translated in two slightly different ways:
```asm
        mov     rdx, QWORD PTR [rcx+rax]
        cmp     rsi, rdx
```
and the following for the current (buggy) version:
```asm
        cmp     QWORD PTR [rdx+rax], rsi
```
However, I do not think that this makes any difference from the memory bus perspective. Both are normal reads, not with any `lock` prefix or any special instruction like `xchg` or `mfence`. This also corresponds to the link that I posted yesterday: sequential consistent reads would not incur any penalty on typical TSO implementations.

I think that this is sufficient evidence to show that your suggested patch should not have any performance impact on other ISA than those that employ a weak memory model.

I have not studied this algorithm; maybe I should one day because of [MDEV-21423](https://jira.mariadb.org/browse/MDEV-21423). It would be nice if we could use acquire-release ordering somehow.

dr-m (approved):

PR #25, opened 2015-03-11
Mdev 5536 systemd socket activation notification

grooverdan:
Adds system functionality to mariadb including socket activation and notification.

PR #24, opened 2015-03-11
MDEV-8557: Correct comments before mysql_socket_{g|s}etfd to refer to the right fun...

grooverdan:
This just changes the comments before the functions as they where transposed.

PR #23, opened 2015-03-11
MDEV-7390 Alter online table x (no options) possible

grooverdan:
This no-op of an operations should be able to occur without locks and and occur online.

MDEV-7390

PR #22, opened 2015-03-11
Alter table x CHARACTER SET y, CONVERT TO CHARACTER SET z fail

grooverdan:
Both these terms set the default_table_charset attribute internally so whenever used they should be the same value.

Alter table xxx CONVERT TO CHARACTER SET latin1, CHARACTER SET utf8 has protection mechanism that result in an error.

However Alter table xxx CHARACTER SET utf8, CONVERT TO CHARACTER SET latin1 will ignore the utf8 and just use the latin1 for the default charset and conversion.

How to repeat:
Alter table xxx CHARACTER SET utf8, CONVERT TO CHARACTER SET latin1
on any table.

The resulting table will have DEFAULT CHARSET=latin1

This modification causes and error consistent with Alter table xxx CONVERT TO CHARACTER SET latin1, CHARACTER SET utf8 to fail.

PR #2805, opened 2023-10-27
MDEV-31953 madvise(..., MADV_FREE) is causing a performance regression / MDEV-24670 avoid OOM by linux kernel co-operative memory management

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-31953 / MDEV-24670*

## Description

The madvice(FREE) added in 10.11 cased performance regressions. To trigger it less often the memory pressure mechanism of Linux was used.

## How can this PR be tested?

Manually like in MDEV-25341.

```
2023-10-27 19:37:44 0 [Note] InnoDB: Loading buffer pool(s) from /tmp/build-mariadb-server-10.11-datadir/ib_buffer_pool
2023-10-27 19:37:44 0 [Note] Plugin 'FEEDBACK' is disabled.
2023-10-27 19:37:44 0 [Note] sql/mariadbd: ready for connections.
Version: '10.11.6-MariaDB'  socket: '/tmp/build-mariadb-server-10.11.sock'  port: 0  Source distribution
2023-10-27 19:37:45 0 [Note] InnoDB: Buffer pool(s) load completed at 231027 19:37:45
2023-10-27 19:37:54 0 [Note] Memory pressure event freed 5314 pages
```

+ mtr case

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/mem_pressure.result
+++ b/mysql-test/suite/innodb/r/mem_pressure.result
@@ -1,0 +1,27 @@
+#
+# MDEV-24670 avoid OOM by linux kernel co-operative memory management
+#
+set @save_dbug=@@debug_dbug;
+set @save_limit=@@GLOBAL.innodb_limit_optimistic_insert_debug;
+set @save_lag_wait=@@GLOBAL.innodb_max_purge_lag_wait;
+set GLOBAL innodb_max_purge_lag_wait=0;
+CREATE TABLE t1 (t TEXT) ENGINE=InnoDB;
+SET GLOBAL innodb_limit_optimistic_insert_debug=2;
+SET unique_checks=0, foreign_key_checks=0;
+INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR)) FROM seq_1_to_1000;
+SET GLOBAL innodb_limit_optimistic_insert_debug=@save_limit;
+SET GLOBAL innodb_max_purge_lag_wait=@save_lag_wait;
+DROP TABLE t1;
+SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+set debug_dbug="d,trigger_garbage_collection";
+SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
+SELECT CAST(VARIABLE_VALUE AS INTEGER) < @dirty_prev AS LESS_DIRTY_IS_GOOD
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+LESS_DIRTY_IS_GOOD
+1
+FOUND 1 /InnoDB: Memory pressure event freed/ in mysqld.1.err
+set debug_dbug=@save_dbug;
+# End of 10.11 tests
--- a/mysql-test/suite/innodb/t/mem_pressure.test
+++ b/mysql-test/suite/innodb/t/mem_pressure.test
@@ -1,0 +1,44 @@
+--source include/have_debug.inc
+--source include/linux.inc
+--source include/not_embedded.inc
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+
+--echo #
+--echo # MDEV-24670 avoid OOM by linux kernel co-operative memory management
+--echo #
+
+set @save_dbug=@@debug_dbug;
+
+set @save_limit=@@GLOBAL.innodb_limit_optimistic_insert_debug;
+set @save_lag_wait=@@GLOBAL.innodb_max_purge_lag_wait;
+set GLOBAL innodb_max_purge_lag_wait=0;
+
+CREATE TABLE t1 (t TEXT) ENGINE=InnoDB;
+SET GLOBAL innodb_limit_optimistic_insert_debug=2;
+SET unique_checks=0, foreign_key_checks=0;
+INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR)) FROM seq_1_to_1000;
+
+SET GLOBAL innodb_limit_optimistic_insert_debug=@save_limit;
+SET GLOBAL innodb_max_purge_lag_wait=@save_lag_wait;
+
+DROP TABLE t1;
+
+SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+
+set debug_dbug="d,trigger_garbage_collection";
+SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
+
+SELECT CAST(VARIABLE_VALUE AS INTEGER) < @dirty_prev AS LESS_DIRTY_IS_GOOD
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+let SEARCH_PATTERN= InnoDB: Memory pressure event freed;
+--source include/search_pattern_in_file.inc
+
+set debug_dbug=@save_dbug;
+
+--echo # End of 10.11 tests
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -742,6 +742,219 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
 
 #ifndef UNIV_INNOCHECKSUM
 
+#ifdef __linux__
+#include <poll.h>
+#include <fstream>
+
+/** Memory Pressure
+
+based off https://www.kernel.org/doc/html/latest/accounting/psi.html#pressure-interface
+and https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#memory */
+class mem_pressure
+{
+  /* triggers + pipe */
+  struct pollfd m_fds[3];
+  nfds_t m_num_fds;
+  int m_pipe[2]= {-1, -1};
+
+  std::thread m_thd;
+  /* mem pressure garbarge collection resticted to interval */
+  static constexpr ulonglong max_interval_us= 60*1000000;
+public:
+  mem_pressure() : m_num_fds(0) {}
+
+  bool setup()
+  {
+    static_assert(array_elements(m_fds) == (array_elements(m_triggers) + 1),
+                  "insufficient fds");
+    std::string memcgroup{"/sys/fs/cgroup"};
+    std::string cgroup;
+    {
+      std::ifstream selfcgroup("/proc/self/cgroup");
+      std::getline(selfcgroup, cgroup, '\n');
+    }
+
+    cgroup.erase(0, 3); // Remove "0::"
+    memcgroup+= cgroup + "/memory.pressure";
+
+    m_num_fds= 0;
+    for (auto trig= std::begin(m_triggers); trig!= std::end(m_triggers); ++trig)
+    {
+      if ((m_fds[m_num_fds].fd=
+             open(memcgroup.c_str(), O_RDWR | O_NONBLOCK | O_CLOEXEC)) < 0)
+      {
+        switch (errno) {
+        case EPERM:
+        /* https://lore.kernel.org/all/CAMw=ZnQ56cm4Txgy5EhGYvR+Jt4s-KVgoA9_65HKWVMOXp7a9A@mail.gmail.com/T/#m3bd2a73c5ee49965cb73a830b1ccaa37ccf4e427 */
+          sql_print_information("InnoDB: Failed to initialize memory pressure EPERM, "
+                                "file permissions.");
+          break;
+        default:
+          sql_print_information("InnoDB: Failed to initialize memory pressure: %s",
+                                strerror(errno));
+        }
+        shutdown();
+        return false;
+      }
+      my_register_filename(m_pipe[0], memcgroup.c_str(), FILE_BY_OPEN, 0, MYF(0));
+      ssize_t slen= strlen(*trig);
+      if (write(m_fds[m_num_fds].fd, *trig, slen) < slen)
+      {
+        sql_print_warning("InnoDB: Failed create trigger for memory pressure \"%s\"", *trig);
+        my_close(m_fds[m_num_fds].fd, MYF(MY_WME));
+        continue;
+      }
+      m_fds[m_num_fds].events= POLLPRI;
+      m_num_fds++;
+    }
+    if (m_num_fds < 1)
+      return false;
+
+    if (pipe2(m_pipe, O_CLOEXEC | O_DIRECT))
+    {
+      sql_print_warning("InnoDB: No memory pressure - can't create pipe");
+      shutdown();
+      return false;
+    }
+    my_register_filename(m_pipe[0], "mem_pressure_read", FILE_BY_DUP, 0, MYF(0));
+    my_register_filename(m_pipe[1], "mem_pressure_write", FILE_BY_DUP, 0, MYF(0));
+    m_fds[m_num_fds].fd= m_pipe[0];
+    m_fds[m_num_fds].events= POLLIN;
+    m_num_fds++;
+    m_thd= std::thread(pressure_routine, this);
+    return true;
+  }
+
+  void shutdown()
+  {
+    while (m_num_fds)
+    {
+      m_num_fds--;
+      my_close(m_fds[m_num_fds].fd, MYF(MY_WME));
+      m_fds[m_num_fds].fd= -1;
+    }
+    /* note m_pipe[0] closed in above loop */
+    if (m_pipe[1] >= 0)
+    {
+      my_close(m_pipe[1], MYF(MY_WME));
+      m_pipe[1]= -1;
+    }
+  }
+
+  static void pressure_routine(mem_pressure *m);
+
+#ifdef UNIV_DEBUG
+  void trigger_collection()
+  {
+    const char c= 'G';
+    if (m_pipe[1] >=0 && write(m_pipe[1], &c, 1) < 1)
+      sql_print_information("InnoDB: (Debug) Failed to trigger memory pressure");
+    else /* assumed failed to meet intialization criteria, so trigger directy */
+      buf_pool.garbage_collect();
+  }
+#endif
+
+  void quit()
+  {
+    const char c= 'Q';
+    if (m_num_fds && write(m_pipe[1], &c, 1) < 1)
+      sql_print_warning("InnoDB: Failed to write memory pressure quit message");
+  }
+
+  void join()
+  {
+    if (m_thd.joinable())
+    {
+      quit();
+      m_thd.join();
+    }
+  }
+
+  static const char* const m_triggers[2];
+};
+
+
+/*
+  ref: https://docs.kernel.org/accounting/psi.html
+  maximum window size (second number) 10 seconds.
+  window size in multiples of 2 second interval required (for Unprivileged)
+  Time is in usec.
+*/
+const char* const mem_pressure::m_triggers[]=
+  {"some 5000000 10000000", /* 5s out of 10s */
+   "full 10000 2000000"}; /* 10ms out of 2s */
+
+static mem_pressure mem_pressure_obj;
+
+void mem_pressure::pressure_routine(mem_pressure *m)
+{
+  DBUG_ASSERT(m == &mem_pressure_obj);
+  if (my_thread_init())
+  {
+    m->shutdown();
+    return;
+  }
+
+  ulonglong last= microsecond_interval_timer() - max_interval_us;
+  while (1)
+  {
+    if (poll(&m->m_fds[0], m->m_num_fds, -1) < 0)
+    {
+      if (errno == EINTR)
+        continue;
+      else
+      {
+        sql_print_warning("InnoDB: memory pressure poll error %s",
+                          strerror(errno));
+        break;
+      }
+    }
+    for (pollfd &p : st_::span<pollfd>(m->m_fds, m->m_num_fds))
+    {
+      if (p.revents & POLLPRI)
+      {
+        ulonglong now= microsecond_interval_timer();
+        if ((now - last) > max_interval_us)
+        {
+          last= now;
+          buf_pool.garbage_collect();
+        }
+      }
+
+      if (p.revents & POLLIN)
+      {
+        char c= '\0';
+        /* signal to quit */
+        if (read(p.fd, &c, 1) >=0)
+          switch (c) {
+          case 'Q':
+            goto shutdown;
+#ifdef UNIV_DEBUG
+          case 'G':
+            buf_pool.garbage_collect();
+#endif
+          }
+      }
+    }
+  }
+shutdown:
+  m->shutdown();
+
+  my_thread_end();
+}
+
+/** Initialize mem pressure. */
+ATTRIBUTE_COLD void buf_mem_pressure_detect_init()
+{
+  mem_pressure_obj.setup();
+}
+
+ATTRIBUTE_COLD void buf_mem_pressure_shutdown()
+{
+  mem_pressure_obj.join();
+}
+#endif /* __linux__ */
+
 #if defined(DBUG_OFF) && defined(HAVE_MADVISE) &&  defined(MADV_DODUMP)
 /** Enable buffers to be dumped to core files
 
@@ -1099,6 +1312,10 @@ bool buf_pool_t::create()
   chunk_t::map_ref= chunk_t::map_reg;
   buf_LRU_old_ratio_update(100 * 3 / 8, false);
   btr_search_sys_create();
+
+#ifdef __linux__
+  buf_mem_pressure_detect_init();
+#endif
   ut_ad(is_initialised());
   return false;
 }
@@ -1536,6 +1753,7 @@ struct find_interesting_trx
 inline void buf_pool_t::resize()
 {
   ut_ad(this == &buf_pool);
+  ut_ad(srv_shutdown_state < SRV_SHUTDOWN_CLEANUP);
 
 	bool		warning = false;
 
@@ -1878,6 +2096,100 @@ calc_buf_pool_size:
 	return;
 }
 
+#ifdef __linux__
+inline void buf_pool_t::garbage_collect()
+{
+  mysql_mutex_lock(&mutex);
+  size_t freed= 0;
+
+#ifdef BTR_CUR_HASH_ADAPT
+  /* buf_LRU_free_page() will temporarily release and reacquire
+  buf_pool.mutex for invoking btr_search_drop_page_hash_index(). Thus,
+  we must protect ourselves with the hazard pointer. */
+rescan:
+#else
+  lru_hp.set(nullptr);
+#endif
+  for (buf_page_t *bpage= UT_LIST_GET_LAST(LRU), *prev; bpage; bpage= prev)
+  {
+    prev= UT_LIST_GET_PREV(LRU, bpage);
+#ifdef BTR_CUR_HASH_ADAPT
+    lru_hp.set(prev);
+#endif
+    auto state= bpage->state();
+    ut_ad(state >= buf_page_t::FREED);
+    ut_ad(bpage->in_LRU_list);
+
+    /* We try to free any pages that can be freed without writing out
+    anything. */
+    switch (bpage->oldest_modification()) {
+    case 0:
+    try_to_evict:
+      if (buf_LRU_free_page(bpage, true))
+      {
+      evicted:
+        freed++;
+#ifdef BTR_CUR_HASH_ADAPT
+        bpage= prev;
+        prev= lru_hp.get();
+        if (!prev && bpage)
+          goto rescan;
+#endif
+      }
+      continue;
+    case 1:
+      break;
+    default:
+      if (state >= buf_page_t::UNFIXED)
+        continue;
+    }
+
+    if (state < buf_page_t::READ_FIX && bpage->lock.u_lock_try(true))
+    {
+      ut_ad(!bpage->is_io_fixed());
+      lsn_t oldest_modification= bpage->oldest_modification();
+      switch (oldest_modification) {
+      case 1:
+        mysql_mutex_lock(&flush_list_mutex);
+        oldest_modification= bpage->oldest_modification();
+        if (oldest_modification)
+        {
+          ut_ad(oldest_modification == 1);
+          delete_from_flush_list(bpage);
+        }
+        mysql_mutex_unlock(&flush_list_mutex);
+        /* fall through */
+      case 0:
+        bpage->lock.u_unlock(true);
+        goto try_to_evict;
+      default:
+        if (bpage->state() < buf_page_t::UNFIXED &&
+            oldest_modification <= log_sys.get_flushed_lsn())
+        {
+          release_freed_page(bpage);
+          goto evicted;
+        }
+        else
+          bpage->lock.u_unlock(true);
+      }
+    }
+  }
+
+#if defined MADV_FREE
+  /* FIXME: Issue fewer calls for larger contiguous blocks of
+  memory. For now, we assume that this is acceptable, because this
+  code should be executed rarely. */
+  for (buf_page_t *bpage= UT_LIST_GET_FIRST(free); bpage;
+       bpage= UT_LIST_GET_NEXT(list, bpage))
+    madvise(bpage->frame, srv_page_size, MADV_FREE);
+#endif
+  mysql_mutex_unlock(&mutex);
+  sql_print_information("InnoDB: Memory pressure event freed %zu pages",
+                        freed);
+  return;
+}
+#endif /* __linux__ */
+
 /** Thread pool task invoked by innodb_buffer_pool_size changes. */
 static void buf_resize_callback(void *)
 {
@@ -1906,12 +2218,23 @@ static tpool::waitable_task buf_resize_task(buf_resize_callback,
 
 void buf_resize_start()
 {
-	srv_thread_pool->submit_task(&buf_resize_task);
+#if !defined(DBUG_OFF) && defined(__linux__)
+  DBUG_EXECUTE_IF("trigger_garbage_collection",
+  {
+    mem_pressure_obj.trigger_collection();
+  }
+  );
+#endif
+
+  srv_thread_pool->submit_task(&buf_resize_task);
 }
 
 void buf_resize_shutdown()
 {
-	buf_resize_task.wait();
+#ifdef __linux__
+  buf_mem_pressure_shutdown();
+#endif
+  buf_resize_task.wait();
 }
 
 
--- a/storage/innobase/include/buf0buf.h
+++ b/storage/innobase/include/buf0buf.h
@@ -656,12 +656,9 @@ public:
     access_time= 0;
   }
 
-  void set_os_unused()
+  void set_os_unused() const
   {
     MEM_NOACCESS(frame, srv_page_size);
-#ifdef MADV_FREE
-    madvise(frame, srv_page_size, MADV_FREE);
-#endif
   }
 
   void set_os_used() const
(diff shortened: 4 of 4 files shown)


storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,211 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+        continue;
+      for (nfds_t i= 0; i < m->m_num_fds; ++i)
+      {
+        if (m->m_fds[i].revents & POLLPRI)
+        {
+          ulonglong now= microsecond_interval_timer();
+          if ((now - last) > max_frequency_ms)
+          {
+            last= now;
+            garbage_collect();
```
dr-m, line 866: Do we really need the separate `garbage_collect()` function? We could just call `buf_pool.garbage_collect()` and output the message here.

storage/innobase/buf/buf0buf.cc
```diff
@@ -1878,39 +2088,99 @@ inline void buf_pool_t::resize()
...
 void buf_resize_start()
 {
-	srv_thread_pool->submit_task(&buf_resize_task);
+#if defined(UNIV_DEBUG) && defined(__linux__)
+  DBUG_EXECUTE_IF("trigger_gargbage_collection",
+  {
+    mem_pressure_obj.trigger_collection();
+  }
+  );
+#endif
```
dr-m, line 2174: I would not add this. Let the resizing be entirely unchanged by this.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17416,18 +17416,12 @@ innodb_stopword_table_validate(
 
 extern void buf_resize_start();
 
-/** Update the system variable innodb_buffer_pool_size using the "saved"
-value. This function is registered as a callback with MySQL.
-@param[in]	save	immediate result from check function */
 static
-void
-innodb_buffer_pool_size_update(THD*,st_mysql_sys_var*,void*, const void* save)
+void innodb_buffer_pool_size_update(THD*,st_mysql_sys_var*,void*,const void*)
```
dr-m, line 17420: Let’s avoid this change.

dr-m (changes requested):
Thank you. I think that the general idea is good.

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,211 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+#ifdef UNIV_PFS_THREAD
+    pfs_register_thread(mem_pressure_thread_key);
+#endif
+
+    ulonglong last= microsecond_interval_timer() - max_frequency_ms;
+    while (1)
+    {
+      int retval= poll(&m->m_fds[0], m->m_num_fds, -1);
+      if (retval == EINTR)
+        continue;
```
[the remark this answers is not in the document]
dr-m: Actually, depending on the `errno` value, we should probably quit the loop and report an error. We would not want one thread to keep calling a contantly failing `poll()` in a busy loop.

storage/innobase/buf/buf0buf.cc
grooverdan: I've still got the resize call back as a trigger for the debug test, but implementation is the same.

mysql-test/suite/innodb/r/mem_pressure.result
```diff
@@ -0,0 +1,23 @@
...
+SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
+SELECT SLEEP(5);
+SLEEP(5)
+0
+SELECT VARIABLE_VALUE < @dirty_prev AS LESS_DIRTY_IS_GOOD FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+LESS_DIRTY_IS_GOOD
+1
+SELECT VARIABLE_VALUE > @flush_prev AS MORE_FLUSHED_IS_GOOD FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_flushed';
+MORE_FLUSHED_IS_GOOD
+1
```
dr-m, line 18: The `buf_pool::garbage_collect()` avoids any writes in order to minimize the I/O impact. If something was written out during the 5-second sleep, that should be a coincidence. I’d rather omit this part of the test.

mysql-test/suite/innodb/r/mem_pressure.result
```diff
@@ -0,0 +1,23 @@
+#
+# MDEV-24670 avoid OOM by linux kernel co-operative memory management
+#
+CREATE TABLE t1 (t TEXT) ENGINE=InnoDB;
+INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR)) FROM seq_1_to_1000;
+SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @flush_prev FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_flushed';
+set debug_dbug="d,trigger_garbage_collection";
```
dr-m, line 8: We should also
```sql
set @save_dbug=@@debug_dbug;
```
and restore the variable after we are done.

mysql-test/suite/innodb/r/mem_pressure.result
```diff
@@ -0,0 +1,23 @@
...
+SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
```
dr-m, line 9: Why are we checking this? This could actually result in different values depending on the build parameters, because the `buf_block_t` will be allocated from the buffer pool as well.

mysql-test/suite/innodb/r/mem_pressure.result
```diff
@@ -0,0 +1,23 @@
...
+SELECT SLEEP(5);
+SLEEP(5)
+0
```
dr-m, line 12: Instead of sleeping in the test, I think that it would be better to revise the debug function `mem_pressure::trigger_collection()` so that it will poll until the garbage collection has been invoked. That could make use of a debug counter.

dr-m (changes requested):
This is almost good to go. Please squash my commit to yours, and just note in the commit message that the logic of `buf_pool_t::garbage_collect()` was provided by me.

I think that the `buf_pool_t::garbage_collect()` needs to be reviewed by @vlad-lesin or @Thirunarayanan. We’d also want subject this to some stress testing with a [MDEV-25341](https://jira.mariadb.org/browse/MDEV-25341) type workload and maybe also using the revised debug instrumentation.

mysql-test/suite/innodb/r/mem_pressure.result
dr-m: Oh, sorry, we are not displaying the size of `innodb_buffer_pool_size`, but using the dummy `SET GLOBAL` as a special trigger mechanism that is enabled by the `debug_dbug`.

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,215 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
 
 #ifndef UNIV_INNOCHECKSUM
 
+/** Memory Pressure */
+
+#ifdef __linux__
+mysql_pfs_key_t mem_pressure_thread_key;
```
dr-m, line 748: If `PLUGIN_PERFSCHEMA=NO`, `mysql_pfs_key_t` will not be defined here. I’d seriously consider using `std::mutex` here, because this is not expected to be a contended mutex, and it is not declared in `all_innodb_mutexes[]` either.

storage/innobase/buf/buf0buf.cc
```diff
@@ -1878,6 +2092,71 @@ inline void buf_pool_t::resize()
...
+    ut_ad(bpage->in_LRU_list);
+
+    /* We try to free any pages that can be freed without writing out
+    anything. */
+    switch (bpage->oldest_modification()) {
+    case 0:
+    try_to_evict:
+      if (buf_LRU_free_page(bpage, true))
+        freed++;
+      continue;
```
dr-m, line 2115: Sorry, I had forgotten that this may temporarily release `buf_pool.mutex`. Please revise as follows:
```diff
@@ -2095,12 +2095,22 @@ inline void buf_pool_t::resize()
 inline void buf_pool_t::garbage_collect()
 {
   mysql_mutex_lock(&mutex);
-  lru_hp.set(nullptr);
   size_t freed= 0;
 
+#ifdef BTR_CUR_HASH_ADAPT
+  /* buf_LRU_free_page() will temporarily release and reacquire
+  buf_pool.mutex for invoking btr_search_drop_page_hash_index(). Thus,
+  we must protect ourselves with the hazard pointer. */
+rescan:
+#else
+  lru_hp.set(nullptr);
+#endif
   for (buf_page_t *bpage= UT_LIST_GET_LAST(LRU), *prev; bpage; bpage= prev)
   {
     prev= UT_LIST_GET_PREV(LRU, bpage);
+#ifdef BTR_CUR_HASH_ADAPT
+    buf_pool.lru_hp.set(prev);
+#endif
     auto state= bpage->state();
     ut_ad(state >= buf_page_t::FREED);
     ut_ad(bpage->in_LRU_list);
@@ -2111,7 +2121,16 @@ inline void buf_pool_t::garbage_collect()
     case 0:
     try_to_evict:
       if (buf_LRU_free_page(bpage, true))
+      {
+      evicted:
         freed++;
+#ifdef BTR_CUR_HASH_ADAPT
+        bpage= prev;
+        prev= buf_pool.lru_hp.get();
+        if (!prev && bpage)
+          goto rescan;
+#endif
+      }
       continue;
     case 1:
       break;
@@ -2136,7 +2155,10 @@ inline void buf_pool_t::garbage_collect()
       default:
         if (bpage->state() < buf_page_t::UNFIXED &&
             oldest_modification <= log_sys.get_flushed_lsn())
+        {
           release_freed_page(bpage);
+          goto evicted;
+        }
         else
           bpage->lock.u_unlock(true);
       }
```
This will also make us count marked-as-freed pages that were removed both from `flush_list`  and `LRU`.

mysql-test/suite/innodb/r/mem_pressure.result
grooverdan: The entire garbage collection is performed in the same thread as 'SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;'. As its not a background task, no `SLEEP` or addition synchronization is required.

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,214 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+        {
+          sql_print_warning("InnoDB: memory pressure poll error %s",
+                            strerror(errno));
+          m->shutdown();
+          break;
+        }
+      }
+      for (pollfd &p : st_::span<pollfd>(m->m_fds, m->m_num_fds))
+      {
+        if (p.revents & POLLPRI)
```
dr-m, line 862: Do we really need the loop here? Only for this first pipe we subscribed to `POLLPRI` events, and only for the second pipe we subscribed to `POLLIN`. Could we just check
```c++
if (m->fds[0].revents & POLLPRI) { … }
if (m->fds[1].revents & POLLIN) { … }
```

storage/innobase/buf/buf0buf.cc
```diff
@@ -1906,12 +2209,23 @@ static tpool::waitable_task buf_resize_task(buf_resize_callback,
 
 void buf_resize_start()
 {
-	srv_thread_pool->submit_task(&buf_resize_task);
+#if defined(UNIV_DEBUG) && defined(__linux__)
+  DBUG_EXECUTE_IF("trigger_garbage_collection",
```
dr-m, line 2213:
```c++
#if !defined DBUG_OFF && defined __linux__
```

dr-m (commented):
Thank you. Hopefully this time the stress test run will not reveal any problems with `buf_pool_t::garbage_collect()`.

storage/innobase/buf/buf0buf.cc
grooverdan: So the first 2 are POLLPRI - two triggers, and I've written the `setup` to tolerate one of the failing to initialize (on EPERM)., and then its the last that's POLLIN only. So yes, we've over checking slightly, but on 3 memory structures that are already in cache that aren't hotly used I was going to readability.

dr-m: OK, then let’s retain the loop.

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,214 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+      }
+    }
+shutdown:
+    m->shutdown();
+
+    my_thread_end();
+
+#ifdef UNIV_PFS_THREAD
+    pfs_delete_thread();
+#endif
```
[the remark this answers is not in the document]
grooverdan: I'm thinking I should have removed this.

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,214 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+  void shutdown()
+  {
+    while (m_num_fds)
+    {
+      m_num_fds--;
+      my_close(m_fds[m_num_fds].fd, MYF(MY_WME));
+    }
+    if (m_pipe[1] >= 0)
+      my_close(m_pipe[1], MYF(MY_WME));
+  }
```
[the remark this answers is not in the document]
grooverdan: never a re-init. But sure easy enough.

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,214 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+    {
+      if ((m_fds[m_num_fds].fd=
+             open(memcgroup.c_str(), O_RDWR | O_NONBLOCK | O_CLOEXEC)) < 0)
+      {
+        switch (errno) {
+        case EPERM:
+        /* https://lore.kernel.org/all/CAMw=ZnQ56cm4Txgy5EhGYvR+Jt4s-KVgoA9_65HKWVMOXp7a9A@mail.gmail.com/T/#m3bd2a73c5ee49965cb73a830b1ccaa37ccf4e427 */
+          sql_print_information("InnoDB: Failed to initialize memory pressure EPERM, "
+                                "file permissions, CAP_SYS_RESOURCE, or Kernel fix required"
+				"(applies to 6.5, 6.6 kernels)");
```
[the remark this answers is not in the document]
grooverdan: yep, omit, but the time this gets out hopefully there's a distro update that includes the patch (at least fc39).

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,214 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+      my_close(m_fds[m_num_fds].fd, MYF(MY_WME));
+    }
+    if (m_pipe[1] >= 0)
+      my_close(m_pipe[1], MYF(MY_WME));
+  }
+
+  static void pressure_routine(mem_pressure *m)
+  {
+    if (my_thread_init())
+      return;
```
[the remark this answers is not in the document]
grooverdan: I couldn't see how to forward declare the class and object enough to compile this debug assert in this position.

mysql-test/suite/innodb/t/mem_pressure.test
```diff
@@ -0,0 +1,33 @@
+--source include/have_debug.inc
+--source include/linux.inc
+--source include/not_embedded.inc
```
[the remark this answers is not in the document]
grooverdan: Haven't worked it out, but it isn't:
```
innodb.mem_pressure 'innodb'             [ fail ]
        Test ended at 2023-11-06 20:04:21

CURRENT_TEST: innodb.mem_pressure
Warning: /home/dan/repos/build-mariadb-server-10.11-debug/libmysqld/examples/mysqltest_embedded: unknown variable 'loose-ssl-ca=/home/dan/repos/mariadb-server-10.11/mysql-test/std_data/cacert.pem'
Warning: /home/dan/repos/build-mariadb-server-10.11-debug/libmysqld/examples/mysqltest_embedded: unknown variable 'loose-ssl-cert=/home/dan/repos/mariadb-server-10.11/mysql-test/std_data/client-cert.pem'
Warning: /home/dan/repos/build-mariadb-server-10.11-debug/libmysqld/examples/mysqltest_embedded: unknown variable 'loose-ssl-key=/home/dan/repos/mariadb-server-10.11/mysql-test/std_data/client-key.pem'
Warning: /home/dan/repos/build-mariadb-server-10.11-debug/libmysqld/examples/mysqltest_embedded: unknown option '--loose-skip-ssl'
--- /home/dan/repos/mariadb-server-10.11/mysql-test/suite/innodb/r/mem_pressure.result	2023-11-06 16:48:34.466207820 +1100
+++ /home/dan/repos/mariadb-server-10.11/mysql-test/suite/innodb/r/mem_pressure.reject	2023-11-06 20:04:20.942017980 +1100
@@ -10,9 +10,9 @@
 SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
 SELECT VARIABLE_VALUE < @dirty_prev AS LESS_DIRTY_IS_GOOD FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 LESS_DIRTY_IS_GOOD
-1
+0
 DROP TABLE t1;
-FOUND 1 /InnoDB: Memory pressure event freed/ in mysqld.1.err
+NOT FOUND /InnoDB: Memory pressure event freed/ in mysqld.1.err
 set debug_dbug=@save_dbug;
 #
 # End of 10.11 tests
```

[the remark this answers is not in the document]
dr-m: OK, I think that we can leave it at that.

storage/innobase/buf/buf0buf.cc
```diff
@@ -1878,6 +2095,96 @@ inline void buf_pool_t::resize()
 	return;
 }
 
+static std::mutex garbage_collect_mutex{};
+
+inline void buf_pool_t::garbage_collect()
+{
+  std::unique_lock<std::mutex> lk(garbage_collect_mutex);
```
dr-m, line 2102: I don’t think we need an extra mutex here. If the adaptive hash index is not being used on the pages that are being freed here, the entire critical section will be covered by `buf_pool.mutex`. Else, the `buf_pool.LRU` traversal is protected by the combination of `buf_pool.mutex` and `buf_pool.lru_hp`.

storage/innobase/buf/buf0buf.cc
```diff
@@ -1878,6 +2095,96 @@ inline void buf_pool_t::resize()
 	return;
 }
 
+static std::mutex garbage_collect_mutex{};
+
+inline void buf_pool_t::garbage_collect()
```
dr-m, line 2100: This code could be inside `#ifdef __linux__` because it is only being used on Linux.

dr-m (commented):
Let’s wait for the stress test results.

storage/innobase/buf/buf0buf.cc
grooverdan: its covering the RQG tests where multiple triggers of garbage_collect can occur the same time. I wasn't sure if this was sufficiently protected. 

Could be under a debug build only if needed.

storage/innobase/buf/buf0buf.cc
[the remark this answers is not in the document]
dr-m:
```c++
class mem_pressure
{
// …
  static void pressure_routine(mem_pressure *m);
// …
};
static mem_pressure mem_pressure_obj;

void mem_pressure::pressure_routine(mem_pressure *m)
{
  DBUG_ASSERT(m == &mem_pressure_obj);
```

mleich1:
The RQG testing on 
      origin/bb-10.11-MDEV-24670-memory-pressure 826b1a0fc3adf6209dc98e737f4c23fc5aec83d0 2023-11-06T20:06:47+11:00
showed none rare
[rr 4186708 251852]mariadbd: /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/include/buf0buf.h:2031: void buf_page_t::clear_oldest_modification(): Assertion `oldest_modification()' failed.
(rr) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=89646028527168) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=89646028527168) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=89646028527168, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x000079f57f728476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x000079f57f70e7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x000079f57f70e71b in __assert_fail_base (fmt=0x79f57f8c3150 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x55cf19977740 "oldest_modification()", 
    file=0x55cf19976460 "/data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/include/buf0buf.h", line=2031, function=<optimized out>) at ./assert/assert.c:92
#6  0x000079f57f71fe96 in __GI___assert_fail (assertion=0x55cf19977740 "oldest_modification()", file=0x55cf19976460 "/data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/include/buf0buf.h", line=2031, 
    function=0x55cf199776a0 "void buf_page_t::clear_oldest_modification()") at ./assert/assert.c:101
#7  0x000055cf18a83e3e in buf_page_t::clear_oldest_modification (this=this@entry=0x7f311180d430) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/include/buf0buf.h:2031
#8  0x000055cf18a6fdbf in buf_pool_t::delete_from_flush_list (this=this@entry=0x55cf1a6f1b40 <buf_pool>, bpage=bpage@entry=0x7f311180d430)
    at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/buf/buf0flu.cc:170
#9  0x000055cf18a45017 in buf_pool_t::garbage_collect (this=this@entry=0x55cf1a6f1b40 <buf_pool>) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/buf/buf0buf.cc:2155
#10 0x000055cf18a27a68 in mem_pressure::trigger_collection (this=0x55cf1b20b5c0 <mem_pressure_obj>) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/buf/buf0buf.cc:909
#11 buf_resize_start () at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/buf/buf0buf.cc:2217
#12 0x000055cf1835e434 in innodb_buffer_pool_size_update (save=<optimized out>) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/handler/ha_innodb.cc:17430
#13 0x000055cf16ecfebb in sys_var_pluginvar::global_update (this=0x6210000cc888, thd=0x62c000380218, var=0x6290017664d8) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/sql_plugin.cc:3677
#14 0x000055cf16b7f16b in sys_var::update (this=0x6210000cc888, thd=<optimized out>, var=0x6290017664d8) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/set_var.cc:207
#15 0x000055cf16b804e5 in set_var::update (this=<optimized out>, thd=<optimized out>) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/set_var.cc:863
#16 0x000055cf16b83a96 in sql_set_variables (thd=thd@entry=0x62c000380218, var_list=var_list@entry=0x62c000385630, free=free@entry=true) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/set_var.cc:745
#17 0x000055cf16e7264e in mysql_execute_command (thd=thd@entry=0x62c000380218, is_called_from_prepared_stmt=is_called_from_prepared_stmt@entry=false) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/sql_parse.cc:5064
#18 0x000055cf16e7c6b0 in mysql_parse (thd=thd@entry=0x62c000380218, rawbuf=<optimized out>, length=<optimized out>, parser_state=parser_state@entry=0x518857af75a0)
    at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/sql_parse.cc:8030
#19 0x000055cf16e8353c in dispatch_command (command=command@entry=COM_QUERY, thd=thd@entry=0x62c000380218, 
    packet=packet@entry=0x62900175c219 "SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size /* E_R Thread1 QNO 5185 CON_ID 13 */ ", packet_length=packet_length@entry=98, blocking=blocking@entry=true)
    at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/sql_parse.cc:1894
#20 0x000055cf16e88980 in do_command (thd=0x62c000380218, blocking=blocking@entry=true) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/sql_parse.cc:1407
#21 0x000055cf1731ef63 in do_handle_one_connection (connect=<optimized out>, connect@entry=0x608000003238, put_in_cache=put_in_cache@entry=true) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/sql_connect.cc:1416
#22 0x000055cf1731f779 in handle_one_connection (arg=arg@entry=0x608000003238) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/sql/sql_connect.cc:1318
#23 0x000055cf181568f1 in pfs_spawn_thread (arg=0x617000007098) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/perfschema/pfs.cc:2201
#24 0x000079f57f77ab43 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#25 0x000079f57f80bbb4 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:100
(rr)

sdp:/data1/results/1699453476/TBR-2080$ _RR_TRACE_DIR=./1/rr rr replay --mark-stdio

Testing scenario:
1. Start the server with sql_mode =      'STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION';
2. create 3 tables
    oltp2 is used later and contains 100 rows. 
    CREATE TABLE `oltp2` (
            `pad` char(60) NOT NULL default '',
            `k` int(10) unsigned NOT NULL default '0',
            `id` int(10) unsigned NOT NULL auto_increment,
            `c` char(120) NOT NULL default '',
             KEY `k` (`k`),
             PRIMARY KEY (`id`)) ENGINE=innodb
2. Two threads connect around the same point of time
    thread1
       runs once, SET debug_dbug="d,trigger_garbage_collection" ;
       runs ~"endless" in a loop   SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size ;
    thread2
       runs ~"endless" in a loop    INSERT IGNORE INTO oltp2 ( k ) VALUES ( random _smallint_unsigned )

dr-m:
The failure that @mleich1 reported on 826b1a0fc3adf6209dc98e737f4c23fc5aec83d0 is in the `buf_pool_t::garbage_collect()` implementation that I provided. We have it waiting for `buf_pool.flush_list_mutex`:
```
#13 buf_pool_t::garbage_collect (this=this@entry=0x55cf1a6f1b40 <buf_pool>) at /data/Server/bb-10.11-MDEV-24670-memory-pressure/storage/innobase/buf/buf0buf.cc:2154
```
While this thread is waiting, `buf_flush_page_cleaner()` is invoking `buf_pool_t::get_oldest_modification()`, which will collect the garbage by invoking `buf_pool_t::delete_from_flush_list()`. We must reload the `oldest_modification` after acquiring the mutex. @mleich1 can you please apply this patch and test again?
```diff
diff --git a/storage/innobase/buf/buf0buf.cc b/storage/innobase/buf/buf0buf.cc
index d46446a4b58..ed434e77fa4 100644
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2148,11 +2148,16 @@ inline void buf_pool_t::garbage_collect()
     if (state < buf_page_t::READ_FIX && bpage->lock.u_lock_try(true))
     {
       ut_ad(!bpage->is_io_fixed());
-      const lsn_t oldest_modification= bpage->oldest_modification();
+      lsn_t oldest_modification= bpage->oldest_modification();
       switch (oldest_modification) {
       case 1:
         mysql_mutex_lock(&buf_pool.flush_list_mutex);
-        buf_pool.delete_from_flush_list(bpage);
+        oldest_modification= bpage->oldest_modification();
+        if (oldest_modification)
+        {
+          ut_ad(oldest_modification == 1);
+          buf_pool.delete_from_flush_list(bpage);
+        }
         mysql_mutex_unlock(&buf_pool.flush_list_mutex);
         /* fall through */
       case 0:
```

storage/innobase/buf/buf0buf.cc
dr-m: I do not think that multiple concurrent calls to `buf_pool.garbage_collect()` should be an issue. Each of them will attempt a scan of the buffer pool. If we have `block->index` for any of the evicted pages (the AHI is disabled by default), then in the worst case, each iteration will `goto rescan` several times. Progress should be guaranteed even in that case.

In the normal case, one of the threads at a time will acquire and hold `buf_pool.mutex` for the entire duration of the iteration. Basically, the `garbage_collect_mutex` is duplicating the role of `buf_pool.mutex`.

mysql-test/suite/innodb/t/mem_pressure.test
```diff
@@ -0,0 +1,32 @@
...
+CREATE TABLE t1 (t TEXT) ENGINE=InnoDB;
+INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR)) FROM seq_1_to_1000;
+
+DROP TABLE t1;
+SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+
+set debug_dbug="d,trigger_garbage_collection";
+SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
+
+SELECT VARIABLE_VALUE < @dirty_prev AS LESS_DIRTY_IS_GOOD FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
```
dr-m, line 22: We are missing a `CAST` in the second `SELECT`. For some reason, the return value of `get_one_variable()` is a string. I did not track down where the comparison is being executed and why it would fail, but I assume that one of the operands is a string and the other is `INTEGER`.

mysql-test/suite/innodb/t/mem_pressure.test
```diff
@@ -0,0 +1,32 @@
...
+--source include/have_sequence.inc
+
+--echo #
+--echo # MDEV-24670 avoid OOM by linux kernel co-operative memory management
+--echo #
+
+set @save_dbug=@@debug_dbug;
+
+CREATE TABLE t1 (t TEXT) ENGINE=InnoDB;
+INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR)) FROM seq_1_to_1000;
```
dr-m, line 14: This will fill the buffer pool with undo log pages due to row-level undo logging. We should enable the bulk insert interface to fill the buffer pool mainly with the pages of the table `t1`. Therefore, it can happen that before and after the `buf_pool_t::garbage_collect()`, the buffer pool is filled with undo log pages.

Because we are using debug instrumentation anyway, we can also make use of `innodb_limit_optimistic_insert_debug` and write considerably less redo log.

We’d better wait that all undo logs have been purged before we start the test. Before I added that, the test would fail after a couple of dozen repetitions.
```diff
diff --git a/mysql-test/suite/innodb/t/mem_pressure.test b/mysql-test/suite/innodb/t/mem_pressure.test
index 2d140fb22ad..60f5acc291d 100644
--- a/mysql-test/suite/innodb/t/mem_pressure.test
+++ b/mysql-test/suite/innodb/t/mem_pressure.test
@@ -9,17 +9,27 @@
 --echo #
 
 set @save_dbug=@@debug_dbug;
-
-CREATE TABLE t1 (t TEXT) ENGINE=InnoDB;
-INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR)) FROM seq_1_to_1000;
-
+set @save_limit=@@GLOBAL.innodb_limit_optimistic_insert_debug;
+SET GLOBAL innodb_max_purge_lag_wait=0;
+
+CREATE TABLE t1 (a INT PRIMARY KEY) ENGINE=InnoDB;
+SET GLOBAL innodb_limit_optimistic_insert_debug=2;
+SET unique_checks=0, foreign_key_checks=0;
+INSERT INTO t1 SELECT * FROM seq_1_to_1000;
+SET GLOBAL innodb_limit_optimistic_insert_debug=@save_limit;
 DROP TABLE t1;
-SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+SET unique_checks=1, foreign_key_checks=1;
+
+SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 
 set debug_dbug="d,trigger_garbage_collection";
 SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
 
-SELECT VARIABLE_VALUE < @dirty_prev AS LESS_DIRTY_IS_GOOD FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+SELECT CAST(VARIABLE_VALUE AS INTEGER) < @dirty_prev AS LESS_DIRTY_IS_GOOD
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 
 let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
 let SEARCH_PATTERN= InnoDB: Memory pressure event freed;
@@ -27,6 +37,4 @@ let SEARCH_PATTERN= InnoDB: Memory pressure event freed;
 
 set debug_dbug=@save_dbug;
 
---echo #
 --echo # End of 10.11 tests
---echo #
```
The test as modified above passed my local test:
```
innodb.mem_pressure 'innodb'             w22 [ 1000 pass ]    139
The servers were restarted 0 times
Spent 5053.123 of 291 seconds executing testcases

Completed: All 42000 tests were successful.
```

storage/innobase/include/buf0buf.h
```diff
@@ -1301,6 +1298,11 @@ class buf_pool_t
   /** Resize from srv_buf_pool_old_size to srv_buf_pool_size. */
   inline void resize();
 
+  /** Collect garbage (release pages from the LRU list) */
+#ifdef __linux__
+  inline void garbage_collect();
```
dr-m, line 1303: I think that Doxygen could complain about the stray comment if not run with `-D__linux__`. Better move it inside the `#ifdef`.

storage/innobase/buf/buf0buf.cc
```diff
@@ -742,6 +742,221 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
...
+
+#ifdef __linux__
+/*
+  based off https://www.kernel.org/doc/html/latest/accounting/psi.html#pressure-interface
+  and https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#memory
+*/
+#include <poll.h>
+#include <fstream>
+
+class mem_pressure
```
dr-m, line 755: Please move the `/**` Doxygen comment right before `class mem_pressure`, and consider including the whole comment there, in case someone is actually generating documentation from the source code and reading it:

```c++
#ifdef __linux__
# include <poll.h>
# include <fstream>

/** Memory Pressure

based on …
*/
class mem_pressure
```

dr-m (approved):
OK to push after addressing these. It was a bit tricky to stabilize the regression test.

PR #2807, opened 2023-10-27
MDEV-32567 : remove thr_alarm from server codebase

vaintroub:
- [x] *The Jira issue number for this PR is: MDEV-32567*

## Description
Refactoring - remove thr_alarm from server codebase.

The thr_alarm.h, alarm() and SIGALRM has not been used for a  very long time. Originally, it seemed to have been used for socket timeouts, but even as far as in MySQL 5.1  it was not used (instead, socket timeout were used in the past, and later, non-blocking reads and writes + poll/select on the waiting socket) . Its existence just obscures the source code, net_real_read/net_read_write, and signal handler  still have some code for "if alarm in use"

## How can this PR be tested?

No need for specific testing, passing buildbot should be enough. There are already tests for different network timeouts.

- [ x] *This is a new feature and the PR is based against the latest MariaDB development branch.*

## PR quality check


grooverdan (commented):
All good after not doing a `shutdown` on the socket activation file descriptors.

grooverdan (approved):

PR #2808, opened 2023-10-28
MDEV-29101 update deprecated mariabackup help options

gulshanpr:
- [x] *The Jira issue number for this PR is: MDEV-29101*

## Description
The patch is for updating the client side `mariabackup --help` options, some of them are deprecated from version 10.5.

## How can this PR be tested?
This PR is tested manually on Windows using `server\bld\mysql-test>perl mysql-test-run.pl mariabackup.deprecated_options`.
I have attached the mtr test and result file along with this PR.
- [x] output of the test was as expected.

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2324,6 +2324,7 @@ void buf_page_free(fil_space_t *space, uint32_t page, mtr_t *mtr,
       mtr->memo_push(block, MTR_MEMO_PAGE_X_FIX);
       rw_lock_x_lock_inline(&block->lock, 0, file, line);
       buf_block_dbg_add_level(block, SYNC_NO_ORDER_CHECK);
+      mtr->set_modified(*block);
 
 #ifdef BTR_CUR_HASH_ADAPT
       if (block->index)
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -875,7 +875,18 @@ static bool buf_flush_page(buf_page_t *bpage, bool lru, fil_space_t *space)
   page_t *frame= bpage->zip.data;
 
   if (status == buf_page_t::FREED)
+  {
+    if (UNIV_LIKELY(space->purpose == FIL_TYPE_TABLESPACE))
+    {
+      const lsn_t lsn= mach_read_from_8(my_assume_aligned<8>
+                                        (FIL_PAGE_LSN + (frame ? frame
+                                                         : block->frame)));
+      ut_ad(lsn >= oldest_modification);
+      if (lsn > log_sys.get_flushed_lsn())
+        log_write_up_to(lsn, true);
+    }
     buf_pool.release_freed_page(&block->page);
+  }
   else
   {
     space->reacquire();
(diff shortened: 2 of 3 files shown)


extra/mariabackup/xtrabackup.cc
```diff
@@ -1267,23 +1267,19 @@ struct my_option xb_client_options[]= {
      GET_STR, REQUIRED_ARG, 0, 0, 0, 0, 0, 0},
 
     {"compress", OPT_XTRA_COMPRESS,
-     "Compress individual backup files using the "
-     "specified compression algorithm. Currently the only supported algorithm "
-     "is 'quicklz'. It is also the default algorithm, i.e. the one used when "
-     "--compress is used without an argument.",
+     "This option was deprecated starting with MariaDB 10.1.31 and 10.2.13 "
```
grooverdan, line 1270: Where did you get these versions from?

If there's a reason, explain it in the commit message, otherwise a non-versioned deprecation notice is ok.

mysql-test/suite/mariabackup/deprecated_options.result
```diff
@@ -0,0 +1,485 @@
...
+of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You can download full text of the license on http://www.gnu.org/licenses/gpl-2.0.txt
+
+Usage: C:\GitHub-local-machine-testing\server\bld\extra\mariabackup\RelWithDebInfo\mariadb-backup.exe [--defaults-file=#] [--backup | --prepare | --copy-back | --move-back] [OPTIONS]
```
grooverdan, line 20: Appreciate the effort made to add this test case. As you see with this line of the test result, and the few lines below, the result is going to depend on the environment where the test case is run.

Also test cases on largely help output can quickly change on any documentation fixes.

For something as simple as a deprecating an option, no test case is needed.

So please remove the test cases.

grooverdan (changes requested):
As this targets 10.5 per the MDEV, you'll need to rebase this to the 10.5 branch.

The way to do this is:
```
# retreive the latest 10.5
git fetch https://github.com/MariaDB/server.git 10.5
# change the one commit of this change onto the fetched 10.5
# force push the change to the remote, which will update this PR
git push -f
```

extra/mariabackup/xtrabackup.cc
gulshanpr: this was in mariabackup knowledgebase [documentation](https://mariadb.com/kb/en/mariabackup-options/#-compress), I took it from there.

mysql-test/suite/mariabackup/deprecated_options.result
gulshanpr: Thanks @grooverdan, I was discussing this issue on channel with Brandon, he said it's a client side change so to prove my changes I need to write test but not an issue I'll remove it.

grooverdan: You've changed the help output. If it was more substantial change in behaviour then a test case would be required.

PR #2809, opened 2023-10-30
MDEV-12404 Index condition pushdown on partitioned tables

DaveGosselin-MariaDB:
Support index condition pushdown on class ha_partition.  Adapts code and test from mysql and includes support for mariadb index ranges.

Index condition pushdown is reflected in EXPLAINs as shown in the accompanying test code.

- [x] *The Jira issue number for this PR is: MDEV-12404*

### DIFF
--- a/mysql-test/suite/parts/t/partition_icp.test
+++ b/mysql-test/suite/parts/t/partition_icp.test
@@ -1,0 +1,63 @@
+#
+# Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023, MariaDB Corporation.
+#
+
+--source include/have_partition.inc
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+
+--echo #
+--echo # test using Index Condition Pushdown for partitioned tables
+--echo #
+
+--echo # Test failure of ICP calls -> don't use ICP (MyISAM does not support
+--echo # ICP on BLOB indexes)
+
+CREATE TABLE t1 (a int PRIMARY KEY, b BLOB, c varchar(16) DEFAULT 'Filler...',
+INDEX (b(4), a)) ENGINE = MyISAM PARTITION BY HASH (a) PARTITIONS 3;
+
+SHOW CREATE TABLE t1;
+
+INSERT INTO t1 (a, b) VALUES (1, 0xdeadbeef), (2, "text filler"),
+(3, 'filler...'), (4, " more filler "), (5, "test text"), (6, "testing...");
+
+let $query= SELECT a, HEX(b) FROM t1 WHERE b >= 'te' and (a % 2);
+eval EXPLAIN $query;
+eval EXPLAIN FORMAT=JSON $query;
+eval $query;
+
+--echo Only MyISAM and InnoDB supports both INDEX and BLOBS...
+ALTER TABLE t1 ENGINE = InnoDB;
+eval EXPLAIN $query;
+eval EXPLAIN FORMAT=JSON $query;
+eval $query;
+
+DROP TABLE t1;
+
+CREATE TABLE t1 (
+  pk INT PRIMARY KEY,
+  a INT,
+  b INT,
+  c INT,
+  filler varchar(100),
+  key(a,b,c)
+) partition by hash(pk) partitions 4;
+
+INSERT INTO t1 SELECT seq, seq, seq, seq, 'hello' FROM seq_1_to_10000;
+EXPLAIN format=json SELECT * from t1 where a < 10 AND (b+1>3);
+
+DROP TABLE t1;
+
+CREATE TABLE t2 (
+  pk INT PRIMARY KEY,
+  a INT,
+  b INT,
+  c INT,
+  filler VARCHAR(100),
+  KEY(a,b,c)
+);
+INSERT INTO t2 SELECT seq, seq, seq, seq, 'hello' FROM seq_1_to_10000;
+EXPLAIN format=json SELECT * FROM t2 WHERE a < 10 AND (b+1>3);
+
+DROP TABLE t2;
--- a/sql/ha_partition.cc
+++ b/sql/ha_partition.cc
@@ -62,6 +62,9 @@
 #include "ddl_log.h"
 
 #include "debug_sync.h"
+#ifndef DBUG_OFF
+#include "sql_test.h"                        // print_where
+#endif
 
 /* First 4 bytes in the .par file is the number of 32-bit words in the file */
 #define PAR_WORD_SIZE 4
@@ -428,6 +431,7 @@ void ha_partition::init_handler_variables()
   part_share= NULL;
   m_new_partitions_share_refs.empty();
   m_part_ids_sorted_by_num_of_records= NULL;
+  m_icp_in_use= false;
   m_partitions_to_open= NULL;
 
   m_range_info= NULL;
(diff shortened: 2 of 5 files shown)


DaveGosselin-MariaDB:
I've introduced a problem when summing over partitioned tables, so I'm investigating.  Separately, I need to update some EXPLAIN output in the tests given this feature.

sql/handler.cc
```diff
@@ -249,12 +249,12 @@ handlerton *ha_default_tmp_handlerton(THD *thd)
 
 /** @brief
   Return the storage engine handlerton for the supplied name
-  
+
   SYNOPSIS
```
spetrunia, line 253: Please avoid making changes like this:  endspace removal, reformatting. 
Rationale:
* The patch becomes harder to review. First few pages of this diff have nothing to do with ICP or partitioning..
* It clutters revision history.  We have to look at history a lot, and will see that this patch touched this code, then will have to look and see that it didn't make any meaningful changes.

If we have to do this kind of code cleanup, let it be a separate commit with a comment clearly saying it only does reformatting.

sql/ha_partition.h
```diff
@@ -1303,18 +1304,7 @@ class ha_partition final :public handler
...
-      The underlying storage engine might support Rowid Filtering. But
-      ha_partition does not forward the needed SE API calls, so the feature
-      will not be used.
-
-      Note: It's the same with IndexConditionPushdown, except for its variant
-      of IndexConditionPushdown+BatchedKeyAccess (that one works). Because of
-      that, we do not clear HA_DO_INDEX_COND_PUSHDOWN here.
-    */
-    return part_flags & ~HA_DO_RANGE_FILTER_PUSHDOWN;
+    return m_file[0]->index_flags(inx, part, all_parts);
```
spetrunia, line 1307: Do I interpret it correctly that this will make ha_partition also support Rowid Filtering? 

This might be good, but then the commit comment needs to mention this and there should be test coverage.

I think it's better to enable Rowid Filtering in a separate patch.  Unless the effort to enable ICP and Rowid Filtering at the same time is much less than enabling them one-by-one.

mysql-test/suite/parts/t/partition_icp.test
```diff
@@ -0,0 +1,63 @@
+#
+# Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
```
spetrunia, line 2: Is this file really copied from MySQL? It doesn't look so, please remove the copyright...

spetrunia (changes requested):

sql/handler.cc
DaveGosselin-MariaDB: My apologies, I will fix my editor.  It's configured to trim trailing whitespace on save, but I will make it stop doing that.

mysql-test/suite/parts/t/partition_icp.test
DaveGosselin-MariaDB: Yes, it is copied from mysql-server sha 67d2e4e but I added some new tests to the end of it.

grooverdan: I'd include this reference and credit to the author/company in the commit message.

mysql-test/suite/parts/t/partition_icp.test
```diff
@@ -0,0 +1,63 @@
...
+# Copyright (c) 2023, MariaDB Corporation.
+#
+
+--source include/have_partition.inc
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+
+--echo #
+--echo # test using Index Condition Pushdown for partitioned tables
```
grooverdan, line 11: a MDEV reference at the beginning of the test make it easier to navigate later.

mysql-test/suite/parts/t/partition_icp.test
```diff
@@ -0,0 +1,63 @@
...
+  a INT,
+  b INT,
+  c INT,
+  filler VARCHAR(100),
+  KEY(a,b,c)
+);
+INSERT INTO t2 SELECT seq, seq, seq, seq, 'hello' FROM seq_1_to_10000;
+EXPLAIN format=json SELECT * FROM t2 WHERE a < 10 AND (b+1>3);
+
+DROP TABLE t2;
```
grooverdan, line 63:
```
--echo #
--echo # End of 11.4 tests
--echo #
```
I assume this is getting rebased to 11.4 now.

sql/ha_partition.cc
```diff
@@ -5193,6 +5197,88 @@ void ha_partition::reset_copy_info()
...
+  @return Remainder of non handled condition
+  @note Only handles full condition or nothing at all. MyISAM and InnoDB
+  both only supports full or nothing.
+*/
+Item *ha_partition::idx_cond_push(uint keyno, Item* idx_cond)
+{
+  uint i;
+  Item *res;
+  DBUG_ENTER("ha_partition::idx_cond_push");
+  DBUG_EXECUTE("where", print_where(idx_cond, "cond", QT_ORDINARY););
```
grooverdan, line 5218: should there a debug test that validates that this output is right?

sql/ha_partition.cc
```diff
@@ -5193,6 +5197,88 @@ void ha_partition::reset_copy_info()
...
+  m_icp_in_use= true;
+
+  for (i= bitmap_get_first_set(&m_part_info->read_partitions);
+       i < m_tot_parts;
+       i= bitmap_get_next_set(&m_part_info->read_partitions, i))
+  {
+    res= m_file[i]->idx_cond_push(keyno, idx_cond);
+    if (res)
+    {
+      uint j;
```
grooverdan, line 5231: this can be moved inside the `for` loop initializer

sql/ha_partition.cc
```diff
@@ -5193,6 +5197,88 @@ void ha_partition::reset_copy_info()
...
+  @param keyno     Key number for the condition
+  @param idx_cond  Item tree of the condition to test
+  @return Remainder of non handled condition
+  @note Only handles full condition or nothing at all. MyISAM and InnoDB
+  both only supports full or nothing.
+*/
+Item *ha_partition::idx_cond_push(uint keyno, Item* idx_cond)
+{
+  uint i;
+  Item *res;
```
grooverdan, line 5216: move inside `for` look initializer

grooverdan (changes requested):
The test failures in main.partition where the SQL result is change because of adding a Index condition are highly suspicious.

The few query plan change in failing test case are probably correct, but should be validated and the re-record the test result. Ensure the original purpose of the test isn't invalidated.

mysql-test/suite/parts/t/partition_icp.test
dr-m: For what it is worth, the MySQL test file was last changed in mysql/mysql-server@f966d5d5617f5805a08942aac297a05693ef6128 and there is no copyright notice. From my time at Oracle, I remember that there was a policy that no copyright notices should be added to public MySQL test files.

DaveGosselin-MariaDB:
> The test failures in main.partition where the SQL result is change because of adding a Index condition are highly suspicious.

Completely agree (please see my earlier comment https://github.com/MariaDB/server/pull/2809#issuecomment-1785476299).  I'm working through the changes to see what I could've missed.
 
> The few query plan change in failing test case are probably correct, but should be validated and the re-record the test result. Ensure the original purpose of the test isn't invalidated.

Agreed.

PR #2810, opened 2023-10-30
MDEV-29101 updated help options for `mariabackup`

gulshanpr:
- [x] *The Jira issue number for this PR is: MDEV-29101*

## Description
This PR is based on #2808, closed due to making blank commit sincere apologies for that.
This patch is for updating the client side `mariabackup --help` options, some of them are deprecated from version 10.5 .

## How can this PR be tested?
This PR is tested manually on Windows using `server\bld\mysql-test>perl mysql-test-run.pl mariabackup.deprecated_options`.
by executing `mariabackup --help` in test.

## PR quality check


extra/mariabackup/xtrabackup.cc
```diff
@@ -1194,23 +1194,19 @@ struct my_option xb_client_options[]= {
...
-     "--compress is used without an argument.",
+     "This option was deprecated starting with MariaDB 10.1.31 and 10.2.13 "
+	 "as it relies on the no longer maintained QuickLZ library.",
      (G_PTR *) &xtrabackup_compress_alg, (G_PTR *) &xtrabackup_compress_alg, 0,
      GET_STR, OPT_ARG, 0, 0, 0, 0, 0, 0},
 
     {"compress-threads", OPT_XTRA_COMPRESS_THREADS,
-     "Number of threads for parallel data compression. The default value is "
-     "1.",
+     "This option was deprecated with --compress ",
```
bnestere, line 1203: Hi Gulshan!

Thanks for the patch! I have a couple quick notes:

1.  I don't think we want to get rid of the existing description text; I'd imagine we should just extend it with the deprecation message so we don't lose anything.

 2. I think the end part mentioning "with --compress" is a little ambiguous too. I'd suggest omitting it altogether, or replacing it with the "as it relies on the no longer maintained QuickLZ library." to be explicit (and the same goes for `--chunk-size` and `decompress`).

gulshanpr: @bnestere thanks for you suggestions, i will make the changes.

grooverdan (approved):

bnestere:
@grooverdan If the 'effective' deprecation date of this is 10.1.31, would it make sense for this patch to go into 10.4, rather than 10.5? (though note the JIRA says 10.5)

grooverdan:
> @grooverdan If the 'effective' deprecation date of this is 10.1.31, would it make sense for this patch to go into 10.4, rather than 10.5? (though note the JIRA says 10.5)

Good idea. Rebased now.

bnestere (approved):
Looks good. Thanks @gulshanpr!

PR #2812, opened 2023-10-31
MDEV-31116 SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2

DaveGosselin-MariaDB:
Verify that a pointer is set before dereferencing it.  Adds automated test that will crash without this fix.

### DIFF
--- a/mysql-test/main/order_by_innodb.result
+++ b/mysql-test/main/order_by_innodb.result
@@ -296,3 +296,13 @@ a	b	c
 6	2	26
 6	3	36
 drop table t1;
+#
+#  MDEV-31116: SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2
+#
+CREATE TABLE t1 (a BINARY (2),b BINARY (1),KEY(a)) ENGINE=innodb;
+INSERT INTO t1 select 'ab', NULL from seq_1_to_14;
+SELECT * FROM t1 WHERE a IN (SELECT a FROM t1 WHERE a >'') ORDER BY a LIMIT 1;
+a	b
+ab	NULL
+DROP TABLE t1;
+# End of 11.0 tests
--- a/mysql-test/main/order_by_innodb.test
+++ b/mysql-test/main/order_by_innodb.test
@@ -250,3 +250,12 @@ explain select * from t1 force index(r) order by a,b limit 20;
 explain select * from t1 force index(r) order by a desc,b limit 20;
         select * from t1 force index(r) order by a desc,b limit 20;
 drop table t1;
+
+--echo #
+--echo #  MDEV-31116: SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2
+--echo #
+CREATE TABLE t1 (a BINARY (2),b BINARY (1),KEY(a)) ENGINE=innodb;
+INSERT INTO t1 select 'ab', NULL from seq_1_to_14;
+SELECT * FROM t1 WHERE a IN (SELECT a FROM t1 WHERE a >'') ORDER BY a LIMIT 1;
+DROP TABLE t1;
+--echo # End of 11.0 tests
--- a/sql/sql_select.cc
+++ b/sql/sql_select.cc
@@ -26478,7 +26478,7 @@ test_if_skip_sort_order(JOIN_TAB *tab,ORDER *order,ha_rows select_limit,
          !table->is_clustering_key(best_key)))
       goto use_filesort;
 
-    if (table->opt_range_keys.is_set(best_key) && best_key != ref_key)
+    if (select && table->opt_range_keys.is_set(best_key) && best_key != ref_key)
     {
       key_map tmp_map;
       tmp_map.clear_all();       // Force the creation of quick select


mysql-test/main/select.test
```diff
@@ -4816,3 +4816,13 @@ DROP TABLE t1;
...
+
+--echo #
+--echo # Bug MDEV-31116: SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2
+--echo #
+SET sql_select_limit=1;
+CREATE TABLE t (a BINARY (2),b BINARY (1),KEY(a));
+INSERT INTO t (a) VALUES (''),(''),(''),(''),(''),(''),(''),(''),(''),(''),
+  (''),(''),(''),('');
+SELECT count(*) FROM t WHERE a IN (SELECT a FROM t WHERE a >'') ORDER BY a;
+DROP TABLE t;
```
grooverdan, line 4828: Convention is when test append to the end of a test case. A marker is made so it be merged easier

```
--echo End of 11.0 tests
```

grooverdan (commented):
As this MDEV has a fix version of 11.0 this should be the base and target version.

This can be achieved by:
```
git fetch git@github.com:MariaDB/server.git 11.0
git push -f
```

Then edit the title of this PR where it has 11.1 as the version and change that to 11.0

mysql-test/main/select.test
DaveGosselin-MariaDB: Thank you, I will add this.

grooverdan (approved):
Result makes sense. (though the `a > ''` being true did confuse me for a bit -> resolved because its fixed length structure so always has a value like '\0\0').

Code change around null pointer de-reference seems narrowly focused without side effects.

Awaiting second review by someone who knows better however.

spetrunia (changes requested):
On the commit comment:

```
MDEV-31116 SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2 

Verify that a pointer is set before dereferencing it.  Adds automated
test that will crash without this fix.
```
is not very informative.
How about:
```
MDEV-31116 SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2 

test_if_skip_sort_order() should check if 'select' (=tab->select)
is not NULL before calling select->test_quick_select(...) etc. 

The check was erroneously removed by:
1c88ac60cf5 Simple cleanup of removing QQ comments from sql_select.cc
```

mysql-test/main/select.test
```diff
@@ -4816,3 +4816,15 @@ DROP TABLE t1;
 DROP VIEW v1, v2;
 
 --echo End of 10.0 tests
+
+--echo #
+--echo # Bug MDEV-31116: SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2
+--echo #
+SET sql_select_limit=1;
+CREATE TABLE t (a BINARY (2),b BINARY (1),KEY(a));
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Thank you, I will make these changes and remember to do the same in the future.

spetrunia (changes requested):
Ok, getting closer but there yet.

It's not a good idea to put this this test into select.test: 
* select.test is too generic. Probably that test shouldn't be used at all.
* select.test variants : jcl6, pkeycache -  are not relevant for this bugfix.

This testcase uses innodb, while select.test doesn't.  Generally, we avoid adding `--include have_innodb.inc` to existing tests.   we create  xxx_innodb.test if necessary.

The testcase prints \0  into .result.  This is not prohibited but not very user-friendly and  is not necessary for this test.

Here's a patch adding adjusted testcase :
```diff
diff --git a/mysql-test/main/order_by_innodb.result b/mysql-test/main/order_by_innodb.result
index ad4acad3319..79938715539 100644
--- a/mysql-test/main/order_by_innodb.result
+++ b/mysql-test/main/order_by_innodb.result
@@ -296,3 +296,13 @@ a  b       c
 6      2       26
 6      3       36
 drop table t1;
+#
+#  MDEV-31116: SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2
+#
+CREATE TABLE t1 (a BINARY (2),b BINARY (1),KEY(a)) ENGINE=innodb;
+INSERT INTO t1 select 'ab', NULL from seq_1_to_14;
+SELECT * FROM t1 WHERE a IN (SELECT a FROM t1 WHERE a >'') ORDER BY a LIMIT 1;
+a      b
+ab     NULL
+DROP TABLE t1;
+# End of 11.0 tests
diff --git a/mysql-test/main/order_by_innodb.test b/mysql-test/main/order_by_innodb.test
index acce96c7603..9ad9cc20337 100644
--- a/mysql-test/main/order_by_innodb.test
+++ b/mysql-test/main/order_by_innodb.test
@@ -250,3 +250,12 @@ explain select * from t1 force index(r) order by a,b limit 20;
 explain select * from t1 force index(r) order by a desc,b limit 20;
         select * from t1 force index(r) order by a desc,b limit 20;
 drop table t1;
+
+--echo #
+--echo #  MDEV-31116: SIGSEGV in test_if_skip_sort_order|JOIN::optimize_stage2
+--echo #
+CREATE TABLE t1 (a BINARY (2),b BINARY (1),KEY(a)) ENGINE=innodb;
+INSERT INTO t1 select 'ab', NULL from seq_1_to_14;
+SELECT * FROM t1 WHERE a IN (SELECT a FROM t1 WHERE a >'') ORDER BY a LIMIT 1;
+DROP TABLE t1;
+--echo # End of 11.0 tests
```
Please apply it and remove the test in main/select*.   After that everything will be fine

PR #2818, opened 2023-11-06
MDEV-32686: crash information (to be more concise and) (was include distro information)

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32686*

## Description

To make crash information resolve easier to a distribution lets include this information in the crash output.

3 lines of /etc/os-release is sufficient to identify distro and version in most cases (sampled from
https://github.com/which-distro/os-release). Samples otherwise can be long, and include SUPPORT_URL/BUG_REPORT_URL to distros which in the case of crashes, aren't as easily resolvable by distros. Those normally aren't in the first 3 lines.

FreeBSD also includes the file
https://man.freebsd.org/cgi/man.cgi?query=os-release&sektion=&manpath=freebsd-release-ports

Its reasonable to support non-Windows filesystems could support the path /etc/os-release.

We place this information before the core infomation, example output:

    The manual page at https://mariadb.com/kb/en/how-to-produce-a-full-stack-trace-for-mariadbd/ contains
    information that should help you find out what is causing the crash.

    Distribution information:
    NAME="Fedora Linux"
    VERSION="39 (Workstation Edition)"
    ID=fedora

    Writing a core file...

To support Apple/FreeBSD kernel information its 'len' variables was renamed to avoid conflicts and the shared 'buff' is used.

Small corrections that should have been fixed in MDEV-30613.
* open doesn't take a MY_F argument.
* casting to uchar isn't needed

Also close files as soon as finished.

## How can this PR be tested?

Start `mariadbd`, and then `kill -SEGV $PID`

- [*] *This is a minor feature request that aids the in the resolution of bugs and has been backported for the convenience of the bug resolver.

## PR quality check


sql/signal_handler.cc
```diff
@@ -58,49 +58,70 @@ extern const char *optimizer_switch_names[];
 
 static inline void output_core_info()
 {
-  /* proc is optional on some BSDs so it can't hurt to look */
-#if defined(HAVE_READLINK) && !defined(__APPLE__) && !defined(__FreeBSD__)
+#ifndef __WIN__
```
vaintroub, line 61: Better  do not use __WIN__ , it is [deprecated in higher versions](https://github.com/mariadb/server/commit/3d6eb7afcfb4#diff-f07460fa16c227caac9e71893e5e08cf2eea9cba2226677a285f3f37361dac3dR30). Use _WIN32 instead.

sql/signal_handler.cc
```diff
@@ -58,49 +58,70 @@ extern const char *optimizer_switch_names[];
...
   char buff[PATH_MAX];
   ssize_t len;
   int fd;
+  if ((fd= open("/etc/os-release", O_RDONLY)) >= 0)
```
an3l, line 65: I would check with  ` >=3 ` since `0,1,2` are related to the processes `/dev/tty` file descriptors.
If used, need to be changed on other places too.

an3l: Also I would add some handling for  error when opening, like `no file found`.

sql/signal_handler.cc
```diff
@@ -58,49 +58,70 @@ extern const char *optimizer_switch_names[];
...
+  {
+    len= read(fd, buff, sizeof(buff));
```
an3l, line 67: `len` is `ssize_t` so we should check if `-1` is returned on `read` and write some message.

sql/signal_handler.cc
```diff
@@ -58,49 +58,70 @@ extern const char *optimizer_switch_names[];
...
+    close(fd);
```
an3l, line 68: I would add 
```
if (close(fd) < 0)
{
  error
}
```
only if read is successfull.

sql/signal_handler.cc
```diff
@@ -58,49 +58,70 @@ extern const char *optimizer_switch_names[];
...
+    /* Take first 3 lines of this file, enough for distro and version only */
+    char *endline= (char *) memchr(buff, '\n', len);
```
an3l, line 70: There is code repeatiton,instead I would write like so:
```C
  len_orig= read(fd, buff, sizeof(buff));
  len= len_orig;
  for (size_t num_lines= 0; num_lines <= 2 & len>0; num_lines++, endline++)
  {
    if (num_lines == 0)
      endline= (char *) memchr(buff, '\n', len_orig);
    else
      endline= (char *) memchr(endline, '\n', len_orig);
    len= endline - buff;
  }
```

sql/signal_handler.cc
```diff
@@ -58,49 +58,70 @@ extern const char *optimizer_switch_names[];
...
+#endif
+  /* proc is optional on some BSDs so it can't hurt to look */
+#if defined(HAVE_READLINK) && !defined(__APPLE__) && !defined(__FreeBSD__)
   if ((len= readlink("/proc/self/cwd", buff, sizeof(buff))) >= 0)
   {
     my_safe_printf_stderr("Writing a core file...\nWorking directory at %.*s\n",
                           (int) len, buff);
   }
-  if ((fd= open("/proc/self/limits", O_RDONLY, MYF(0))) >= 0)
+  if ((fd= open("/proc/self/limits", O_RDONLY)) >= 0)
```
an3l, line 93: Same `>=3`

sql/signal_handler.cc
```diff
@@ -58,49 +58,70 @@ extern const char *optimizer_switch_names[];
...
   {
     my_safe_printf_stderr("Resource Limits:\n");
-    while ((len= read(fd, (uchar*)buff, sizeof(buff))) > 0)
+    while ((len= read(fd, buff, sizeof(buff))) > 0)
     {
       my_write_stderr(buff, len);
     }
     close(fd);
```
an3l, line 100: For each close I would add check if successful.

an3l (changes requested):
Hi @grooverdan,
thanks, I left some suggested changes.
In addition to commit message,maybe to rephrase and put on the beginning?
```
Its reasonable to support non-Windows filesystems could support
the path /etc/os-release.
```
Use `variable`
```
To support Apple/FreeBSD kernel information its 'len' variables was
renamed to avoid conflicts and the shared 'buff' is used.
```
Also I would write function like `open()`
And please add why casting is not needed (because of implicit conversion,right).

sql/signal_handler.cc
grooverdan: Ack. I thought you cleaned all these up. Sorry some have crept back in and I foolishly copied them. Thanks for the reminder.

sql/signal_handler.cc
grooverdan: Not necessary tty or stdin/out/error.

```
$ strace -fe trace=file,write  -o /tmp/x.strace  sql/mariadbd --no-defaults --datadir=/tmp/${PWD##*/}-datadir --socket=/tmp/${PWD##*/}.sock --plugin-dir=${PWD}/mysql-test/var/plugins/ --verbose --skip-networking 0>&1 1>&-
..

mariadbd 100386  dan    0r   CHR                1,3      0t0       4 /dev/null
mariadbd 100386  dan    1uW  REG               0,35 12582912    1305 /tmp/build-mariadb-server-10.4-datadir/ibdata1
mariadbd 100386  dan    2u   CHR             136,12      0t0      15 /dev/pts/12
```

Now, don't want to be writing to stdout right.

Or

```
$ strace -fe trace=file,write  -o /tmp/x.strace  sql/mariadbd --no-defaults --datadir=/tmp/${PWD##*/}-datadir --socket=/tmp/${PWD##*/}.sock --plugin-dir=${PWD}/mysql-test/var/plugins/ --verbose --skip-networking 0>&1 1>&- 2>&-

mariadbd 100447  dan    0r   CHR                1,3      0t0       4 /dev/null
mariadbd 100447  dan    1uW  REG               0,35 12582912    1305 /tmp/build-mariadb-server-10.4-datadir/ibdata1
mariadbd 100447  dan    2u   REG               0,35     2947    1551 /tmp/#1551 (deleted)
```

Wonder what the stderr is?

Either way, opening a file returns a valid >=0 file descriptor. We don't test the relation to 3 elsewhere.

On error reporting, its not really an error if the file isn't there. We just have a little less information to show the user.

sql/signal_handler.cc
grooverdan: we only close a file if a read is successful? nope - we close a file if it opened successfully, read errors will be handled. Like above, didn't want to put errors on information gathering to confuse with the original signal causing error.

sql/signal_handler.cc
grooverdan: Right track. I'll do something like this. Generally if you doing an `if condition` on a loop variable inside the loop, one of the two is incorrect. In this example initialising `endline =buff` before the loop simplifies away the `if` condition.

sql/signal_handler.cc
grooverdan: On `man 2 close` "See NOTES for a discussion of why close() should not be retried after an error".  So there's no meaningful action to take if unsuccessful.

vuvova:
I generally don't like the idea of adding more and more code into this code path. SIGSEGV handler is invoked when something is already wrong, memory is likely corrupted and doing *anything* at this point is risky, so we should limit this code to a minimum and only do things we absolutely cannot do outside.

Meaning, stack trace is, probably, ok, but ulimit or distro information is not. Find some other way of having it in the log. E.g systemd could put it there or systemd could invoke an "after-crash helper" that will do it or SIGSEGV handler can exec "after-crash helper", whatever, there are many safe ways of doing it.

grooverdan:
Ok, can I include the second/third commits cleaning up old out of date junk comments?

vuvova:
[MDEV-32686: minimise crash information](https://github.com/MariaDB/server/pull/2818/commits/08b81502d77749b511a1fd14b111577e4af6d4cf) is ok.

As for two others — let's still think how can we print OS global info out of sigsegv handler. OS version and distro, core location — that's all global, doesn't need to be collected from mariadbd.

ulimits are per process, so, likely, has to be done from mariadbd after all. Reducing the number of ulimit rows printed is good, thanks. Why did you keep nice/realtime priorities and msgqueue size, are they important?

How can the OS-global info be printed?

One options — collect all that info on startup, unconditionally. It'll likely take a neglectable amount of time compared to the total startup time anyway. Save it preprocessed in a static buffer and print it with one `write()` on sigsegv.

grooverdan:
> [MDEV-32686: minimise crash information](https://github.com/MariaDB/server/pull/2818/commits/08b81502d77749b511a1fd14b111577e4af6d4cf) is ok.
> 
> As for two others — let's still think how can we print OS global info out of sigsegv handler. OS version and distro, core location — that's all global, doesn't need to be collected from mariadbd.

ack, Options in MDEV.

> ulimits are per process, so, likely, has to be done from mariadbd after all. Reducing the number of ulimit rows printed is good, thanks. Why did you keep nice/realtime priorities and msgqueue size, are they important?

Not particularly, the simple reduction mechanism was reducing "unlimited" resource limits. I was avoiding a complicated filter so there few lines got though the crude filter. Even detecting 0 limits means other 0 limites like core file size would be filtered, so that wasn't done.

> How can the OS-global info be printed?
> 
> One options — collect all that info on startup, unconditionally. It'll likely take a neglectable amount of time compared to the total startup time anyway. Save it preprocessed in a static buffer and print it with one `write()` on sigsegv.

I also needed to consider selinux rules etc. Lets see how the state with improved packaging looks.

PR #2820, opened 2023-11-07
MDEV-13626 Merge InnoDB test cases from MySQL 5.7

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-13626*

## Description
This imports and adapts a number of MySQL 5.7 test cases that are applicable to MariaDB.

Some tests for old bug fixes are not that relevant because the code has been refactored since then (especially starting with MariaDB Server 10.6), and the tests would not reproduce the original bug if the fix was reverted.

In the test `innodb_fts.opt`, there are many duplicate  `MATCH` ranks, which would make the results nondeterministic. The test was stabilized by changing some `LIMIT` clauses or by adding sorted_result in those cases where the purpose of a test was to show that no sorting took place in the server.

In the test `innodb_fts.phrase`, MySQL 5.7 would generate `FTS_DOC_ID` that are 1 larger than in MariaDB. In `innodb_fts.index_table` the difference is 2. This is because in MariaDB, `fts_get_next_doc_id()` post-increments `cache->next_doc_id`, while MySQL 5.7 pre-increments it.

The test `innodb.import_update_stats` is for now disabled on Valgrind because it would reproduce a bug [MDEV-32725](https://jira.mariadb.org/browse/MDEV-32725).

## How can this PR be tested?
This is tested on any CI system that runs `mtr`. There are no code changes.

## PR quality check


mysql-test/suite/innodb_fts/t/phrase.test
```diff
@@ -0,0 +1,39 @@
+-- source include/have_innodb.inc
+
```
dr-m, line 2: We got [this ./mtr --ps-protocol --embedded failure](https://buildbot.mariadb.org/#/builders/536/builds/11170/steps/7/logs/stdio) that starts like this:
```diff
@@ -42,8 +42,8 @@
 Sort_merge_passes	0
 Sort_priority_queue_sorts	0
 Sort_range	0
-Sort_rows	2
-Sort_scan	1
+Sort_rows	4
+Sort_scan	2
 No sorting for this query even if MATCH is part of an expression
 FLUSH STATUS;
 SELECT title, MATCH(title, text) AGAINST ('database') AS score
```
I can reproduce this with `./mtr --ps-protocol --embedded` as well as `./mtr --ps-protocol` but not with `./mtr --embedded`. Maybe we just need `disable_ps_protocol` or `disable_ps2_protocol` somewhere in this test.

dr-m: Adding either `disable_ps2_protocol` or `disable_ps_protocol` would not fix this.

dr-m: I worked around this by adding the following:
```
if ($PS_PROTOCOL) { skip Counter differences for --ps-protocol; }
```

Thirunarayanan (approved):
Patch looks OK to me

mysql-test/suite/innodb/t/innodb-lock-inherit-read_commited.test
```diff
@@ -0,0 +1,114 @@
...
+--source ../../innodb/include/wait_all_purged.inc
+
+SET debug_sync = 'now SIGNAL con2_go WAIT_FOR con2_insert_done';
+SET debug_sync = 'now SIGNAL con1_go WAIT_FOR con1_insert_done';
+
+SET debug_sync = 'now SIGNAL con1_finish';
+SET debug_sync = 'now SIGNAL con2_finish';
+
+--connection con1
+--reap
```
dr-m, line 91: This test occasionally fails on several builders. I will have to check this deeper. It has failed in both iterations of the loop:
```diff
@@ -46,6 +46,8 @@
 SET debug_sync = 'now SIGNAL con1_finish';
 SET debug_sync = 'now SIGNAL con2_finish';
 connection con1;
+Warnings:
+Warning	1639	debug sync point wait timed out
 disconnect con1;
 connection con2;
 disconnect con2;
```
```diff
@@ -104,6 +104,8 @@
 SET debug_sync = 'now SIGNAL con1_finish';
 SET debug_sync = 'now SIGNAL con2_finish';
 connection con1;
+Warnings:
+Warning	1639	debug sync point wait timed out
 disconnect con1;
 connection con2;
 disconnect con2;
```

dr-m: I was able to repeat the `DEBUG_SYNC` hang on my system when running 17×10 tests. After I applied the following cleanup, the test did not hang on my system:
```diff
diff --git a/mysql-test/suite/innodb/t/innodb-lock-inherit-read_commited.test b/mysql-test/suite/innodb/t/innodb-lock-inherit-read_commited.test
index 1c887911069..d9c933fa9b5 100644
--- a/mysql-test/suite/innodb/t/innodb-lock-inherit-read_commited.test
+++ b/mysql-test/suite/innodb/t/innodb-lock-inherit-read_commited.test
@@ -63,8 +63,8 @@ SET debug_sync = 'now WAIT_FOR con1_locks_done';
 
 SET debug_sync = 'lock_wait_suspend_thread_enter SIGNAL con2_blocked
 WAIT_FOR con2_go';
-SET debug_sync = 'ha_commit_trans_after_acquire_commit_lock SIGNAL
-con2_insert_done WAIT_FOR con2_finish';
+SET debug_sync = 'ha_commit_trans_after_acquire_commit_lock
+WAIT_FOR con2_finish';
 SET debug_sync = 'ib_after_row_insert SIGNAL con2_insert_done';
 
 --send
@@ -85,11 +85,12 @@ SET debug_sync = 'now SIGNAL con2_go WAIT_FOR con2_insert_done';
 SET debug_sync = 'now SIGNAL con1_go WAIT_FOR con1_insert_done';
 
 SET debug_sync = 'now SIGNAL con1_finish';
-SET debug_sync = 'now SIGNAL con2_finish';
 
 --connection con1
 --reap
 --disconnect con1
+--connection default
+SET debug_sync = 'now SIGNAL con2_finish';
 
 --connection con2
 --error 0,ER_LOCK_WAIT_TIMEOUT
```
Only one other added test `innodb.innodb-ac-non-locking-select` makes use of `DEBUG_SYNC`, in a much simpler pattern. I successfully ran 68×30=2040 repetitions of it on my system.

PR #2821, opened 2023-11-07
MDEV-28509 Dereferenced null pointer of type 'struct JOIN_TAB' in add_key_field

DaveGosselin-MariaDB:
Do not attempt to merge keys to a non-existent join table; verify that the REGINFO's join_tab field is not NULL.

The newly added tests will crash without this fix, but with this fix they produce same results as when run on MySQL 8.2.


mysql-test/suite/parts/t/optimizer.test
```diff
@@ -47,3 +47,14 @@ SHOW status LIKE 'handler_read_key';
 --enable_ps2_protocol
 
 DROP TABLE t1, t2;
+
+--echo #
+--echo # Bug#28509: Access null pointer during add_key_field call
```
spetrunia, line 52: Please use  `MDEV-28509`.  That's what people grep for.   `Bug#nnnnn` notation was used for bugs at MySQL Ab, people may think you're refererring to https://bugs.mysql.com/bug.php?id=28509

sql/sql_select.cc
```diff
@@ -25007,6 +25008,25 @@ int setup_order(THD *thd, Ref_ptr_array ref_pointer_array, TABLE_LIST *tables,
...
+  OnScopeExit(std::function<void()> work)
+    : _work(work)
+  {}
+
+  ~OnScopeExit()
+  {
+    _work();
+  }
+};
+
```
vuvova, line 25029: No, don't do that. You can use `SCOPE_EXIT` from `include/scope.h`

DaveGosselin-MariaDB: I missed `scope_exit`, thank you; it does exactly what I need 👍

vuvova: uppercase. Like
```c++
SCOPE_EXIT([order] () {
   ...
});
```

PR #2822, opened 2023-11-08
MDEV-22230 : Unexpected ER_ERROR_ON_RENAME upon DROP non-existing FOREIGN KEY

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-22230*

## Description
mysql_prepare_alter_table(): Alter table should check whether foreign key exists when it expected to exists and report the error in early stage

dict_foreign_parse_drop_constraints(): Don't throw error if the foreign key constraints doesn't exist when if exists is given in the statement.

## How can this PR be tested?
Added the test case to check the problematic scenario

## PR quality check


sql/sql_table.cc
```diff
@@ -9048,6 +9048,27 @@ mysql_prepare_alter_table(THD *thd, TABLE *table,
         goto err;
       case Alter_drop::FOREIGN_KEY:
         // Leave the DROP FOREIGN KEY names in the alter_info->drop_list.
+	/* If we are dropping a foreign key and expect it to exists,
```
dr-m, line 9051: Grammar: to exist

Maybe even better would be to use the SQL syntax: "If this is DROP FOREIGN KEY without IF EXIST,"

storage/innobase/dict/dict0dict.cc
```diff
@@ -4645,6 +4646,7 @@ dict_foreign_parse_drop_constraints(
 		ptr1 = dict_accept(cs, ptr1, "EXISTS", &success);
 		if (success) {
 			ptr = ptr1;
+			is_exists = true;
```
dr-m, line 4649: `if_exists` would be a more descriptive name, because the syntax is `IF EXISTS`.

storage/innobase/dict/dict0dict.cc
```diff
@@ -4659,10 +4661,11 @@ dict_foreign_parse_drop_constraints(
 	(*constraints_to_drop)[*n] = id;
 	(*n)++;
 
-	if (std::find_if(table->foreign_set.begin(),
-			 table->foreign_set.end(),
-			 dict_foreign_matches_id(id))
-	    == table->foreign_set.end()) {
+	if (!is_exists
+	    && std::find_if(table->foreign_set.begin(),
```
dr-m, line 4665: It would be safer to do `(*constraints_to_drop)[(*n)++] = id` only if the constraint was found in `table->foreign_set`. I see that `row_delete_constraint()` does not really care if the constraint exists, but we should keep in mind that constraint names of different tables may conflict. That is, the foreign key constraint name would be `database_name/constraint_name`. If you do `ALTER TABLE database_name.table_name DROP FOREIGN KEY IF EXISTS constraint_name`, with the current patch it looks like it could actually drop a constraint by that name even if it belongs to `database_name.other_table_name`.

dr-m (changes requested):
Can this be merged directly to 10.5, or would we need a separate 10.5 version?

mysql-test/main/type_ranges.test
```diff
@@ -76,9 +76,11 @@ alter short drop default,
 DROP INDEX utiny,
 DROP INDEX ushort,
 DROP PRIMARY KEY,
-DROP FOREIGN KEY any_name,
```
dr-m, line 79: Can we add `IF EXISTS` here? I did not notice any test where we would do `DROP FOREIGN KEY IF EXISTS` on a non-existing name.

dr-m (changes requested):
Please add a test where we try both `DROP FOREIGN KEY` and `DROP FOREIGN KEY IF EXISTS` that would match a table in another database:
```sql
CREATE DATABASE best;
CREATE TABLE best.t1 … FOREIGN KEY foo …;
CREATE TABLE t1  … FOREIGN KEY foo …;
ALTER TABLE t1 DROP FOREIGN KEY foo;
--error
ALTER TABLE t1 DROP FOREIGN KEY foo;
ALTER TABLE t1 DROP FOREIGN KEY IF EXISTS foo;
SHOW CREATE TABLE best.t1;
SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_FOREIGN;
DROP TABLE best.t1;
```

mysql-test/main/type_ranges.result
```diff
@@ -144,8 +144,10 @@ alter short drop default,
 DROP INDEX utiny,
 DROP INDEX ushort,
 DROP PRIMARY KEY,
-DROP FOREIGN KEY any_name,
+DROP FOREIGN KEY IF EXISTS any_name,
 ADD INDEX (auto);
+Warnings:
+Note	1091	Can't DROP FOREIGN KEY `any_name`; check that it exists
```
dr-m, line 150: The "Warnings" for the "Note" seems a bit inappropriate here, but this logic was not changed by this patch.

Just out of curiosity: What would `ALTER TABLE t1 DROP COLUMN IF EXISTS non_existent_column` report?

sql/sql_table.cc
```diff
@@ -9056,6 +9056,27 @@ mysql_prepare_alter_table(THD *thd, TABLE *table,
...
+          FOREIGN_KEY_INFO *f_key;
+          table->file->get_foreign_key_list(thd, &fk_child_key_list);
+          List_iterator<FOREIGN_KEY_INFO> fk_key_it(fk_child_key_list);
+          while ((f_key= fk_key_it++))
+          {
+            if (my_strcasecmp(system_charset_info, f_key->foreign_id->str,
+                              drop->name) == 0)
+              break;
+          }
+          if (!f_key)
```
dr-m, line 9073: We can avoid a redundant condition here:
```c++
          while (FOREIGN_KEY_INFO *f_key= fk_key_it++)
            if (!my_strcasecmp(system_charset_info, f_key->foreign_id->str,
                               drop->name))
              goto fk_found;
          my_error(ER_CANT_DROP_FIELD_OR_KEY, MYF(0), drop->type_name(),
                   drop->name);
          goto err;
        fk_found:
          break;
```

dr-m (approved):

mysql-test/main/type_ranges.result
Thirunarayanan: ALTER TABLE t2 DROP FOREIGN KEY IF EXISTS fk_1;
Warnings:
Note	1091	Can't DROP FOREIGN KEY `fk_1`; check that it exists

dr-m:
This was closed in 5bb31bc88273ae00464afc489827dc536136955f, which is identical with aa1f7ebe7247e158b9bafc1829c74b7ff9c46198.

PR #2824, opened 2023-11-09
MDEV-32452 InnoDB system tablespace is not shrunk on slow shutdown

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-32452*

## Description
- Truncate the system tablespace during slow shutdown process.

## How can this PR be tested?
Added the mtr test case.

## PR quality check


storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3552,6 +3552,9 @@ void fsp_system_tablespace_truncate()
 {
   uint32_t last_used_extent= 0;
   fil_space_t *space= fil_system.sys_space;
+
+  if (!space) return;
```
dr-m, line 3556: Why do we need this condition? The only reason that I can think of would be that the server failed to start up for whatever reason. What would be the preceding error log messages in that case? What is the value of `srv_was_started`?

Wouldn’t it be safer to call this function in `innobase_end()` right before `innodb_shutdown()`?

mysql-test/suite/innodb/r/sys_truncate_shutdown_debug.result
```diff
@@ -0,0 +1,39 @@
...
+f3 INT NOT NULL, INDEX(f1),
+INDEX(f2), INDEX(f3))ENGINE=InnoDB;
+BEGIN;
+INSERT INTO t1 SELECT seq, seq, seq FROM seq_1_to_16384;
+INSERT INTO t1 SELECT seq, seq, seq FROM seq_1_to_16384;
+COMMIT;
+DROP TABLE t1;
+SELECT NAME, FILE_SIZE FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESPACES WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	537919488
```
dr-m, line 20: This is occasionally reporting a different size. At least 540016640 has been observed.

Could it be that we need a fresh InnoDB bootstrap? Also, the normal bootstrap, which is executing some SQL, could be nondeterministic. Could we let InnoDB create a new data directory, by specifying `--innodb-data-home-dir` in this test? See the test `innodb.log_file` for an example.

mysql-test/suite/innodb/t/sys_truncate_shutdown_debug.test
```diff
@@ -0,0 +1,50 @@
...
+INSERT INTO t1 SELECT seq, seq, seq FROM seq_1_to_16384;
+COMMIT;
+DROP TABLE t1;
+SELECT NAME, FILE_SIZE FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESPACES WHERE SPACE = 0;
+
+SET GLOBAL INNODB_FAST_SHUTDOWN=0;
+SET GLOBAL DEBUG_DBUG="+d,shrink_buffer_pool_full";
+--source include/restart_mysqld.inc
+
+--let SEARCH_PATTERN= \[Warning\] InnoDB: Cannot shrink the system tablespace
```
dr-m, line 29: Should it be `\\[Warning\\]`?

mysql-test/suite/innodb/t/sys_truncate_shutdown_debug.test
```diff
@@ -0,0 +1,50 @@
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+--source include/not_embedded.inc
+--source include/have_debug.inc
+--source include/not_windows.inc
```
dr-m, line 5: Why are we excluding Windows?

dr-m (changes requested):
This is a simple change, but the devil is in the detail.

mysql-test/suite/innodb/r/sys_truncate_shutdown_debug.result
dr-m: What was the fix? To add another slow shutdown before the test?

dr-m (approved):

PR #2826, opened 2023-11-09
MDEV-29913  Assertion `thd->stmt_arena != thd->progress.arena' failed in thd_progress_init upon bulk load

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-29913*

## Description
row_merge_bulk_t::write_to_index(): Don't report the progress for bulk load operation

## How can this PR be tested?
Wrote test case to test scenario. Running ./mtr innodb.insert_into_empty test case

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/insert_into_empty.test
+++ b/mysql-test/suite/innodb/t/insert_into_empty.test
@@ -495,3 +495,18 @@ DROP TABLE t1;
 CREATE TABLE t (a CHAR CHARACTER SET utf8) ENGINE=InnoDB ROW_FORMAT=REDUNDANT;
 INSERT t SELECT left(seq,1) FROM seq_1_to_43691;
 DROP TABLE t;
+
+--echo #
+--echo #  MDEV-29913  Assertion `thd->stmt_arena != thd->progress.arena'
+--echo #		failed in thd_progress_init upon bulk load
+--echo #
+CREATE TABLE t1(
+	pk int, f1 varchar(4) DEFAULT 'foo', KEY(pk)) ENGINE=InnoDB;
+SET AUTOCOMMIT=0;
+--disable_ps2_protocol
+SELECT * INTO OUTFILE 'load.data' FROM seq_1_to_200;
+--enable_ps2_protocol
+LOAD DATA INFILE 'load.data' INTO TABLE t1(pk);
+DROP TABLE t1;
+let $datadir= `select @@datadir`;
+remove_file $datadir/test/load.data;
--- a/storage/innobase/row/row0merge.cc
+++ b/storage/innobase/row/row0merge.cc
@@ -3545,17 +3545,6 @@ row_merge_sort(
 	of file marker).  Thus, it must be at least one block. */
 	ut_ad(file->offset > 0);
 
-	/* These thd_progress* calls will crash on sol10-64 when innodb_plugin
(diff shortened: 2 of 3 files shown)


storage/innobase/row/row0merge.cc
```diff
@@ -3551,7 +3551,7 @@ row_merge_sort(
 	*/
 #ifndef __sun__
 	/* Progress report only for "normal" indexes. */
-	if (dup && !(dup->index->type & DICT_FTS)) {
+	if (update_progress) {
```
dr-m, line 3554: This is partly repurposing the parameter `update_progress`. That may be fine, but shouldn’t we retain the condition on `dup->index->type` or move it to those callers that currently passes `update_progress=true`? Or add debug assertions at the start of this function:
```c++
ut_ad(!update_progress || dup);
ut_ad(!update_progress || !dup || !(dup->index->type & DICT_FTS));
```

mysql-test/suite/innodb/t/insert_into_empty.test
```diff
@@ -495,3 +495,18 @@ DROP TABLE t1;
...
+CREATE TABLE t1(
+	pk int, f1 varchar(4) DEFAULT 'foo', KEY(pk)) ENGINE=InnoDB;
+SET AUTOCOMMIT=0;
+--disable_ps2_protocol
+SELECT * INTO OUTFILE 'load.data' FROM seq_1_to_200;
+--enable_ps2_protocol
+LOAD DATA INFILE 'load.data' INTO TABLE t1(pk);
+DROP TABLE t1;
+let $datadir= `select @@datadir`;
+remove_file $datadir/test/load.data;
```
dr-m, line 512: Add the following as the last line:
```sql
-- echo # End of 10.11 tests
```

dr-m (commented):
Can you check how the `thd_progress_init()` and friends are being covered in tests, or otherwise visible to users? It is something specific to MariaDB, added in 8984bef2e431db42761d3f267b950293cfd864de and fc31e3114b2538d152194d17ff0f0439db565634, and disabled on Solaris in fa10a65dc293c9222687b33379cebf240cc4bf29. This is subject to overhaul; see [MDEV-9260](https://jira.mariadb.org/browse/MDEV-9260).

dr-m:
Thank you. I think that this should be applied to 10.4 already, if you confirmed that this progress reporting that had been added in fc31e3114b2538d152194d17ff0f0439db565634 actually has no user visible impact under any circumstances.

Related to this, the global variables that were added in fc2df3c637d51c4ab32132e2d8f63dbe7e80f056 near the start of `row0log.cc` look like they will produce bogus results if there are multiple `ALTER TABLE` statements being executed concurrently. They can also form a performance bottleneck. I think that we should do something about that as well.

Thirunarayanan:
`commit 7081feeac9b7442380b2cb134c215b3d74f9a96c (10.4-MDEV-22230)
Date:   Tue Nov 28 13:52:22 2023 +0530

    MDEV-29913 Assertion `thd->stmt_arena != thd->progress.arena' failed in thd_progress_init upon bulk load
    
    - Commit fc31e3114b2538d152194d17ff0f0439db565634(MDEV-8179) doesn't
    report the progress of inplace alter completely. It just does only
    in row_merge_sort(). Removing the progress report function completely
`

PR #2827, opened 2023-11-10
MDEV-28367: BACKUP LOCKS on table to be accessible to those with database LOCK TABLES privileges

an3l:
- Allow database level access via `LOCK TABLES` to execute statement `BACKUP [un]LOCK <object>`
- `BACKUP UNLOCK` works only with `RELOAD` privilege.
      In case there is `LOCK TABLES` privilege without `RELOAD` privilege,
      we check if backup lock is taken before.
      If it is not we raise an error of missing `RELOAD` privilege.
- We had to remove any error/warnings from calling functions because `thd->get_stmt_da()->m_status` will be set to error and will break `my_ok()`.
- Added missing test coverage of `RELOAD` privilege to `main.grant.test`

Reviewer: <daniel@mariadb.org>

- [ x] *This is a new feature and the PR is based against the latest MariaDB development branch.*

## PR quality check


mysql-test/main/backup_locks.test
```diff
@@ -188,6 +188,79 @@ DROP TABLE t3;
...
+create database db1;
+create table db1.t1(t int);
+create user user1@localhost;
+select user,host from mysql.user where user='user1';
+--error ER_DBACCESS_DENIED_ERROR
+--connect (con1, localhost, user1, ,db1)
+
+# Global privileges should be granted to all schemas, not individual DB
+--error ER_WRONG_USAGE
+grant reload on db1.* to user1@localhost;
```
grooverdan, line 207: You shouldn't need to set granting of `reload` to a DB.

mysql-test/main/backup_locks.test
```diff
@@ -188,6 +188,79 @@ DROP TABLE t3;
...
+grant reload on *.* to user1@localhost;
+show grants for user1@localhost;
+--error ER_DBACCESS_DENIED_ERROR
+--connect (con1, localhost, user1, ,db1)
```
grooverdan, line 212: This also isn't testing the functionality you've added.

sql/sql_parse.cc
```diff
@@ -5153,9 +5153,36 @@ mysql_execute_command(THD *thd)
...
-      goto error;
-    /* first table is set for lock. For unlock the list is empty */
+    if (check_global_access(thd, RELOAD_ACL, true))
+    {
+      /* In case there is no global privilege, check DB privilege for BACK LOCK.*/
+      if (first_table)
+      {
+        if (check_single_table_access(thd, LOCK_TABLES_ACL, first_table, true))
+        {
+#ifndef NO_EMBEDDED_ACCESS_CHECKS
```
grooverdan, line 5163: this `#ifdef` could apply to the entire inner of the other `if` condition right? `check_single_table_access` isn't defined under `ifndef NO_EMBEDDED_ACCESS_CHECKS`.

sql/sql_parse.cc
```diff
@@ -5153,9 +5153,36 @@ mysql_execute_command(THD *thd)
...
+        }
+      }
+      else // BACKUP UNLOCK
+      {
+        if (!thd->mdl_backup_lock)
+        {
+#ifndef NO_EMBEDDED_ACCESS_CHECKS
+          char command[30];
+          get_privilege_desc(command, sizeof(command), RELOAD_ACL);
+          my_error(ER_SPECIFIC_ACCESS_DENIED_ERROR, MYF(0), command);
```
grooverdan, line 5178: so this is a denied message for everyone, if there isn't a `backup lock` in place saying reload is needed?

sql/sql_parse.cc
```diff
@@ -5153,9 +5153,36 @@ mysql_execute_command(THD *thd)
...
+#endif
+          goto error;
+        }
+      }
+    }
+    /* There is reload privilege, first table is set for lock.
```
grooverdan, line 5184: multi line comment format.

sql/sql_parse.cc
```diff
@@ -5153,9 +5153,36 @@ mysql_execute_command(THD *thd)
...
+          goto error;
+        }
+      }
+      else // BACKUP UNLOCK
+      {
+        if (!thd->mdl_backup_lock)
+        {
+#ifndef NO_EMBEDDED_ACCESS_CHECKS
+          char command[30];
+          get_privilege_desc(command, sizeof(command), RELOAD_ACL);
```
grooverdan, line 5177: Would be both `RELOAD_ACL|LOCK_TABLES_ACL`. `command` may need to be bigger.

What error under `RELOAD_ACL` happens when you `BACKUP UNLOCK` without a table `BACKUP LOCK`ed? This should be the same.

grooverdan (changes requested):
Good. Thanks. Mainly don't test functionality you're not added.

Finer aspects of `unlock` need to be checked.

mysql-test/main/backup_locks.test
aheacon: There is no test that is covering this specific scenario and even more specifically, there is no test for `backup locks` that is showing how to use this privilege properly. Maybe should be added in other commit, but I think it is worth to be covered. The same goes to line `210`.

mysql-test/main/backup_locks.test
aheacon: As above comment for line `206`.

sql/sql_parse.cc
aheacon: I guess so, yes.

an3l: I guess so, yes.

mysql-test/main/backup_locks.test
an3l: There is no test that is covering this specific scenario and even more specifically, there is no test for backup locks that is showing how to use this privilege properly. Maybe should be added in other commit, but I think it is worth to be covered. The same goes to line 210.

mysql-test/main/backup_locks.test
an3l: Same as for line 206.

sql/sql_parse.cc
aheacon: For `backup unlock`.

an3l: For `backup unlock`

sql/sql_parse.cc
an3l: There will be no error. Basically when you call `backup lock` it will first call `unlock` and process further.
Do you want to add that specific test case too?

sql/sql_parse.cc
an3l: Does it make sense to wrap `#ifndef NO_EMBEDDED_ACCESS_CHECKS` in line `5158`, so for `embedded` basically it should only check global privilege,not database levels?

grooverdan: wrap outside `check_global_access` which returns always true under `NO_EMBEDDED_ACCESS_CHECKS`. You've tried compiling in this mode?

sql/sql_parse.cc
grooverdan: rough thought while rather sleepy says yes, if your code implements it a case to show its behaviour is good.

mysql-test/main/backup_locks.test
grooverdan: sure, put them in `main.grant` however.

sql/sql_parse.cc
grooverdan: Unlocking without existing backup lock:

with ACL_RELOAD -  no error or warning

new behaviour - with LOCK_TABLES

Existing:
```
MariaDB [test]> backup unlock
    -> ;
ERROR 1227 (42000): Access denied; you need (at least one of) the RELOAD privilege(s) for this operation
```

New depends which DB.  `mdl_backup_lock`/`mdl_context`/`mdl_context.owner` won't give the DB.  `thd->db` as the current DB is messy.
 
So document behaviour in the commit message/comment that we know there's an unconsistency with the `backup unlock` in the case where no RELOAD privs, and no table is locked, because we can't easily determine if `lock tables` can apply.

sql/sql_parse.cc
an3l: Yes, now I did. Wrapping outside `check_global_access` (no call for this function) doesn't work and raises an `DBUG_ASSERT(0)` in `Protocol::end_statement()`, so leaving as suggested before.

sql/sql_parse.cc
```diff
@@ -5153,9 +5153,41 @@ mysql_execute_command(THD *thd)
...
+        {
+          char command[30];
+          get_privilege_desc(command, sizeof(command), RELOAD_ACL|LOCK_TABLES_ACL);
+          my_error(ER_SPECIFIC_ACCESS_DENIED_ERROR, MYF(0), command);
+          goto error;
+        }
+      }
+      else // BACKUP UNLOCK
+      {
+        if (!thd->mdl_backup_lock)
```
grooverdan, line 5174: Suggested code comment:

We test mdl_backup_lock here because if a user could obtain a lock it would be silly to error and say you can't BACKUP UNLOCK (because its obvious you did a BACKUP LOCK). As BACKUP UNLOCK doesn't have a database reference, and there's no way we can check if the BACKUP LOCK privilege is missing. Testing thd->db would involve faking a TABLE_LIST structure which because of the depth of inspection in check_single_table_access make the faking likely to cause crashes or unintended effects. The outcome of this is if a user does an BACKUP UNLOCK without a BACKUP LOCKED table they may be a ER_SPECIFIC_ACCESS_DENIED error even though they do have the privilege. Its a bit different to what happens if the user has RELOAD_ACL where the error is silently ignored.

sql/sql_parse.cc
```diff
@@ -5153,9 +5153,41 @@ mysql_execute_command(THD *thd)
...
+          char command[30];
+          get_privilege_desc(command, sizeof(command), RELOAD_ACL);
+          my_error(ER_SPECIFIC_ACCESS_DENIED_ERROR, MYF(0), command);
+          goto error;
+        }
+      }
+#else
       goto error;
-    /* first table is set for lock. For unlock the list is empty */
+#endif
```
grooverdan, line 5185: The `#ifndef` is moving to the outer, and without access checks, there shouldn't be a `goto error`, because what would the error be.

mysql-test/main/backup_locks.test
grooverdan: like previous moved test result. Put grant tests in `grant.test`.

mysql-test/main/backup_locks.test
```diff
@@ -188,6 +188,85 @@ DROP TABLE t3;
...
+--source include/have_metadata_lock_info.inc
+create database db1;
+create table db1.t1(t int);
+create user user1@localhost;
+select user,host from mysql.user where user='user1';
+--replace_result $MASTER_MYSOCK MASTER_SOCKET $MASTER_MYPORT MASTER_PORT
+--error ER_DBACCESS_DENIED_ERROR
+--connect (con1, localhost, user1, ,db1)
+
+# Global privileges should be granted to all schemas, not individual DB
```
grooverdan, line 205: this comment isn't relevent with the code moved.

grooverdan (changes requested):

sql/sql_parse.cc
grooverdan: ok.
 Remove the `goto error;` condition in the else branch, does that work?

sql/sql_parse.cc
grooverdan: `RELOAD_ACL|LOCK_TABLES_ACL` for the error message.

grooverdan (approved):

mysql-test/main/grant.result
```diff
@@ -1951,6 +1951,11 @@ GRANT EXECUTE                 ON mysqltest_db1.* TO mysqltest_u1@localhost;
 GRANT FILE                    ON *.* TO mysqltest_u1@localhost;
 GRANT CREATE USER             ON *.* TO mysqltest_u1@localhost;
 GRANT PROCESS                 ON *.* TO mysqltest_u1@localhost;
+GRANT RELOAD                  ON mysqltest_db1.* TO mysqltest_u1@localhost;
+ERROR HY000: Incorrect usage of DB GRANT and GLOBAL PRIVILEGES
+connect(localhost,mysqltest_u1,,db1,16000,/home/anel/GitHub/mariadb/server/build/10.4/mysql-test/var/tmp/mysqld.1.sock);
```
grooverdan, line 1956: /home/anel --- fix this output.

LinuxJedi (commented):
One implementation detail comment.

I checked and those tests don't run on embedded, so they should be good.

sql/sql_parse.cc
```diff
@@ -5153,9 +5153,55 @@ mysql_execute_command(THD *thd)
...
+    {
+#ifndef NO_EMBEDDED_ACCESS_CHECKS
+      /*
+        In case there is no global privilege, check DB privilege for LOCK TABLES.
+      */
+      if (first_table) // BACKUP LOCK
+      {
+        if (check_single_table_access(thd, LOCK_TABLES_ACL, first_table, true))
+        {
+          char command[30];
```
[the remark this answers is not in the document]
an3l: Yes, it is enough in this case and could be less (~20), but there is assert on the lower limit size that I didn't want to change.

PR #2829, opened 2023-11-10
MDEV-32757 innodb_undo_log_truncate=ON is not crash safe

dr-m:
- [x] *The Jira issue number for this PR is: [MDEV-32757](https://jira.mariadb.org/browse/MDEV-32757)*

## Description
During undo log truncation, `trx_purge_truncate_history()` was prematurely removing dirty pages from the `buf_pool.flush_list`. This is not safe to do before `mtr_t::commit_shrink()`, where it is actually being done. If dirty pages are marked as clean before the size change of the undo tablespace has been durably written, the `buf_flush_page_cleaner()` thread could advance the InnoDB log checkpoint. On crash recovery, this would cause previously committed transactions within the undo tablespace to appear as incomplete, requiring a rollback.

## How can this PR be tested?
```bash
./mtr --parallel=auto --repeat=100 innodb.undo_truncate_recover{,,,,,,,,,,,,,,,,,,}
```
Actually, this is how the bug was found on 10.5, thanks to the 10.5 specific bug [MDEV-32681](https://jira.mariadb.org/browse/MDEV-32681) that causes InnoDB to be forcibly killed before `mtr_t::commit_shrink()` could be executed. The test normally is killed after that spot.

I think that it is unfeasible to write a deterministic regression test for this. We do not have any `DEBUG_SYNC` control of the `buf_flush_page_cleaner` thread or the InnoDB log checkpoints. To reproduce this bug, we’d need a log checkpoint to occur at an unfortunate moment.

The more robust rollback logic in 5a8b6ff7d9805403ed8c7bb9025e9e8b85eccfcf was tested with the copy of the corrupted data directory `data.tar.xz` in [MDEV-32681](https://jira.mariadb.org/browse/MDEV-32681).

## PR quality check


dr-m:
10.5 version: #2830

storage/innobase/trx/trx0purge.cc
```diff
@@ -707,26 +707,20 @@ TRANSACTIONAL_TARGET void trx_purge_truncate_history()
         ut_ad(!bpage->is_io_fixed());
         ut_ad(bpage->id().space() == space_id);
 
-        if (bpage->oldest_modification() > 2)
-        {
+        if (bpage->oldest_modification() > 1 &&
+            !mtr.have_x_latch(*reinterpret_cast<buf_block_t*>(bpage)))
```
dr-m, line 711: I actually intended to retain the `> 2` to be consistent with the outer `if` inside the `for` body, but I forgot to save that change before creating this pull request. Either value is fine (the value 2 is specific to dirty pages of the temporary tablespace), but using the same value could allow the compiler to reuse a register for the common subexpression. In the 10.5 version #2830 we use the value 1 in both.

mleich1:
origin/10.6-MDEV-32757 5a8b6ff7d9805403ed8c7bb9025e9e8b85eccfcf 2023-11-10T12:27:02+02:00
behaved well in RQG testing.

Thirunarayanan (approved):

mleich1:
origin/10.5-MDEV-32757 64601d745903bb3b0487d294cece2958eb312f28 2023-11-10T11:54:20+02:00
behaved well in RQG testing

PR #2831, opened 2023-11-10
MDEV-28613 LeakSanitizer caused by I_S query using LIMIT ROWS EXAMINED

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-28613*

## Description
Problem:
========
- InnoDB fails to free the allocated buffer of stored cursor when information schema query is interrupted.

Solution:
=========
- In case of error handling, information schema query should free the allocated buffer to store the cursor.

## How can this PR be tested?
Added the test case

## PR quality check


mysql-test/suite/innodb/r/innodb_stats_fetch.result
```diff
@@ -150,3 +150,8 @@ max_data_length	0
 index_length	16384
 DROP TABLE test_ps_fetch;
 set @@use_stat_tables = @save_use_stat_tables;
+#
+# MDEV-28613 LeakSanitizer caused by I_S query using LIMIT ROWS EXAMINED
+#
+SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_INDEXES LIMIT ROWS EXAMINED 5;
+ERROR HY000: Unknown error
```
dr-m, line 157: Can we get an outcome that is similar to the following (tested on 10.5 c638051d808a7b29b10e231c9eae97d5909d17f6)?
```
SELECT * FROM seq_1_to_100 limit rows examined 5;
seq
1
2
Warnings:
Warning	1931	Query execution was interrupted. The query exceeded LIMIT ROWS EXAMINED 5. The query result may be incomplete
```
Please make the test case cover all tables whose handling was revised.

storage/innobase/handler/i_s.cc
```diff
@@ -5055,8 +5055,11 @@ i_s_sys_tables_fill_table(
 		mutex_exit(&dict_sys.mutex);
 
 		if (!err_msg) {
-			i_s_dict_fill_sys_tables(thd, table_rec,
-						 tables->table);
+			if (int err = i_s_dict_fill_sys_tables(
+				thd, table_rec, tables->table)) {
+				ut_free(pcur.old_rec_buf);
+				DBUG_RETURN(err);
```
dr-m, line 5061: This is missing `mem_heap_free(heap);`. Maybe, declare `int err` in a wider scope, move the release of `dict_sys.mutex` before `mtr_t::commit()` at the end of the function, and add `goto func_exit;` for the error handling.

storage/innobase/handler/i_s.cc
```diff
@@ -6823,10 +6861,13 @@ i_s_sys_tablespaces_fill_table(
...
+	if (int err = i_s_dict_fill_sys_tablespaces(
 		thd, fil_system.temp_space->id,
 		fil_system.temp_space->name,
-		fil_system.temp_space->flags, tables->table);
+		fil_system.temp_space->flags, tables->table)) {
+		ut_free(pcur.old_rec_buf);
+		DBUG_RETURN(err);
+	}
 
 	DBUG_RETURN(0);
```
dr-m, line 6872: Is the `ut_free()` really needed? We do not invoke it in the normal case. When would the cursor position be saved? Could the `ut_free()` be needed also in the normal case?

The `if` around `int err` could be removed here. We could either just `DBUG_RETURN(i_s_dict_fill_sys_tablespaces(…))` or execute the `if` body unconditionally.

dr-m (changes requested):
In `INFORMATION_SCHEMA.INNODB_SYS_TABLES` there seems to be an additional leak.

Please try to avoid code duplication for the error handling. `goto` is not that evil.

mysql-test/suite/innodb/t/innodb_stats_fetch.test
```diff
@@ -81,3 +81,14 @@ FROM information_schema.tables WHERE table_name = 'test_ps_fetch';
 
 DROP TABLE test_ps_fetch;
 set @@use_stat_tables = @save_use_stat_tables;
+
+--echo #
+--echo # MDEV-28613 LeakSanitizer caused by I_S query using LIMIT ROWS EXAMINED
+--echo #
+SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_INDEXES LIMIT ROWS EXAMINED 5;
+SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_TABLES LIMIT ROWS EXAMINED 5;
+SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESTATS LIMIT ROWS EXAMINED 5;
```
dr-m, line 90: Did you mention all these tables in an `.opt` file? `./mtr --embedded` fails:
```
mysqltest: At line 90: query 'SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESTATS LIMIT ROWS EXAMINED 5' failed: 1109: Unknown table 'INNODB_SYS_TABLESTATS' in information_schema
```

storage/innobase/handler/i_s.cc
```diff
@@ -5765,9 +5792,17 @@ i_s_sys_columns_fill_table(
...
 		if (!err_msg) {
-			i_s_dict_fill_sys_columns(thd, table_id, col_name,
-						 &column_rec, nth_v_col,
-						 tables->table);
+			err = i_s_dict_fill_sys_columns(
+				thd, table_id, col_name,
+				&column_rec, nth_v_col,
+				tables->table);
+			if (err) {
+				ut_free(pcur.old_rec_buf);
```
dr-m, line 5800: I think that the `ut_free()` had better be invoked after `func_exit:`, to fix yet another memory leak.

storage/innobase/handler/i_s.cc
```diff
@@ -5954,8 +5990,15 @@ i_s_sys_virtual_fill_table(
...
-			i_s_dict_fill_sys_virtual(thd, table_id, pos, base_pos,
-						  tables->table);
+			if (int err = i_s_dict_fill_sys_virtual(
+				thd, table_id, pos, base_pos,
+				tables->table)) {
+				ut_free(pcur.old_rec_buf);
+				if (thd_kill_level(thd)) {
+					err = 0;
+				}
+				DBUG_RETURN(err);
```
dr-m, line 6000: Add a `goto func_exit` and move the `ut_free()` call there.

dr-m (changes requested):
I see that we forget to free the `pcur.old_rec_buf` in many functions. It would be better to use a function and a simpler pattern for error handling, like this:
```c++
static int i_s_sys_error_handling(int err, THD *thd)
{
  return thd_kill_level(thd) ? 0 : err;
}
```
and invoke it like this:
```c++
if (err) {
	err = i_s_error_handling(err, thd);
	goto func_exit;
}
```

mysql-test/suite/innodb/r/innodb_stats_fetch.result
```diff
@@ -150,3 +150,30 @@ max_data_length	0
...
+Level	Warning
+Code	1931
+Message	Query execution was interrupted. The query exceeded LIMIT ROWS EXAMINED 5. The query result may be incomplete
+SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESTATS LIMIT ROWS EXAMINED 5;
+Warnings:
+Level	Warning
+Code	1931
+Message	Query execution was interrupted. The query exceeded LIMIT ROWS EXAMINED 5. The query result may be incomplete
+SELECT SPACE FROM INFORMATION_SCHEMA.INNODB_SYS_TABLESPACES LIMIT ROWS EXAMINED 5;
+SPACE	1
```
dr-m, line 172: These results could potentially be nondeterministic. Should be disable results for the queries, to be on the safe side?

mysql-test/suite/innodb/t/innodb_stats_fetch.opt
```diff
@@ -0,0 +1,2 @@
+--innodb_sys_tablestats
+--innodb_sys_tablespaces
```
dr-m, line 2: Should all `INFORMATION_SCHEMA.INNODB_SYS` plugins be explicitly listed?

dr-m (approved):
Thank you, this looks OK, except I’d play it safe and add `--disable_result` and `--enable_result` around each `SELECT…LIMIT ROWS EXAMINED` statement.

Thirunarayanan:
`commit 84e0c027e03f016da9a7c9dd9ae69fa1303d1ae2 (bb-10.4-MDEV-28613)
Date:   Mon Nov 20 14:58:28 2023 +0530

    MDEV-28613 LeakSanitizer caused by I_S query using LIMIT ROWS EXAMINED
    
    Problem:
    ========
    - InnoDB fails to free the allocated buffer of stored cursor
    when information schema query is interrupted.
    
    Solution:
    =========
    - In case of error handling, information schema query should free
    the allocated buffer to store the cursor.
`

Pushed this PR in 10.4 and above is the commit information.

PR #2833, opened 2023-11-13
MDEV-32790: Output result in show create table for mysql_json type should be longtext

an3l:

## PR quality check

### DIFF
--- a/plugin/type_mysql_json/type.cc
+++ b/plugin/type_mysql_json/type.cc
@@ -62,7 +62,7 @@ public:
   bool parse_mysql(String *dest, const char *data, size_t length) const;
   bool send(Protocol *protocol) { return Field::send(protocol); }
   void sql_type(String &s) const
-  { s.set_ascii(STRING_WITH_LEN("json /* MySQL 5.7 */")); }
+  { s.set_ascii(STRING_WITH_LEN("mysql_json /* JSON from MySQL 5.7 */")); }
   /* this will make ALTER TABLE to consider it different from built-in field */
   Compression_method *compression_method() const { return (Compression_method*)1; }
 };
--- a/mysql-test/main/mysql_json_table_recreate.result
+++ b/mysql-test/main/mysql_json_table_recreate.result
@@ -169,3 +169,47 @@ Total_Number_of_Tests	Succesful_Tests	String_is_valid_JSON
 drop table tempty;
 drop table mysql_json_test;
 drop table mysql_json_test_big;
+#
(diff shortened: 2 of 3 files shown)


plugin/type_mysql_json/type.cc
```diff
@@ -62,7 +62,7 @@ class Field_mysql_json: public Field_blob
   bool parse_mysql(String *dest, const char *data, size_t length) const;
   bool send(Protocol *protocol) { return Field::send(protocol); }
   void sql_type(String &s) const
-  { s.set_ascii(STRING_WITH_LEN("json /* MySQL 5.7 */")); }
+  { s.set_ascii(STRING_WITH_LEN("longtext /* MySQL 5.7 */")); }
```
vuvova, line 65: Better `/* JSON from MySQL 5.7 */`
because `longtext` is the same in MariaDB and in MySQL 5.7, so it's unclear why `SHOW CREATE` will print this comment.

an3l: Thanks, have updated.

cvicentiu (changes requested):
Hi @an3l!

I'm ok with the changes in `plugin/type/type_mysql_json.cc`.

I don't agree with the test case however. We shouldn't allow `create table t1(j mysql_json);` in the first place.

To test the change, you should run `show create table` on a table that was created in mysql.

You can use the tempty table from before. Just copy the data files again.

```
--copy_file std_data/mysql_json/tempty.frm $MYSQLD_DATADIR/test/tempty.frm
--copy_file std_data/mysql_json/tempty.MYI $MYSQLD_DATADIR/test/tempty.MYI
--copy_file std_data/mysql_json/tempty.MYD $MYSQLD_DATADIR/test/tempty.MYD
```

an3l:
HI @cvicentiu , 
your suggestion is done in PR[ #2839](https://github.com/MariaDB/server/pull/2839).
But I would also suggest to merge current PR in current form and after review of PR #2839 to update the test accordingly

an3l:
Hi @cvicentiu, 
in order to address this bug we have to create the table with `mysql_json` field type.
Using MySQL files from `std_data` will raise  `ER_TABLE_NEEDS_REBUILD` and that's why we couldn't even catch this bug before.
So I'm suggesting to merge this PR to have logical sequence of changes in git history. 
After merge, I will update PR #2839 accordingly.

cvicentiu:
Feel free to merge this after my last suggestion is implemented.

PR #2837, opened 2023-11-13
MDEV-32795: ALTER SEQUENCE IF NOT EXISTS non_existing_seq Errors rather than note

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32795*

## Description

…
Like all IF NOT EXISTS syntax, a Note should be generated.

The original commit of Seqeuences cleared the IF NOT EXISTS part in the sql/sql_yacc.yy with lex->create_info.init(). Without this bit set there was no way it could do anything other than error.

To remedy this removal, the sql_yacc.yy components have been minimised as they where all set at the beginning of the ALTER. This way the opt_if_not_exists correctly set the IF_EXISTS flag.

In MDEV-13005 (bb4dd70e7c65) the error code changed, requiring ER_UNKNOWN_SEQUENCES to be handled in the function No_such_table_error_handler::handle_condition.

## How can this PR be tested?

had a mtr test case, just had the wrong result expected.

## PR quality check


mariadb-RuchaDeodhar (approved):
looks good ! thanks !

PR #2838, opened 2023-11-14
MDEV-32242	innodb.doublewrite test case always gets skipped

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-32242*

## Description
- Rewrote the doublewrite test case and removes the usage of innodb_buf_flush_list variable, added search pattern to make sure that InnoDB uses doublewrite buffer. Replaced all kill_mysqld.inc with shutdown_mysqld.inc and zero shutdown timeout
## How can this PR be tested?
Run innodb.doublewrite test case in mtr

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/doublewrite.test
+++ b/mysql-test/suite/innodb/t/doublewrite.test
@@ -1,17 +1,11 @@
+
 --echo #
---echo # Bug #17335427 INNODB CAN NOT USE THE DOUBLEWRITE BUFFER PROPERLY
---echo # Bug #18144349 INNODB CANNOT USE THE DOUBLEWRITE BUFFER FOR THE FIRST
---echo # PAGE OF SYSTEM TABLESPACE
+--echo # MDEV-32242 innodb.doublewrite test case always is skipped
 --echo #
 
 --source include/innodb_page_size.inc
---source include/have_debug.inc
 --source include/not_embedded.inc
-# This test is slow on buildbot.
---source include/big_test.inc
 
-# Slow shutdown and restart to make sure ibuf merge is finished
-SET GLOBAL innodb_fast_shutdown = 0;
 --disable_query_log
 call mtr.add_suppression("InnoDB: Data file .* uses page size .* but the innodb_page_size start-up parameter is");
 call mtr.add_suppression("InnoDB: adjusting FSP_SPACE_FLAGS");
@@ -22,21 +16,13 @@ call mtr.add_suppression("Plugin 'InnoDB' (init function returned error|registra
 call mtr.add_suppression("InnoDB: A bad Space ID was found in datafile");
 call mtr.add_suppression("InnoDB: Checksum mismatch in datafile: ");
 call mtr.add_suppression("InnoDB: Inconsistent tablespace ID in .*t1\\.ibd");
+call mtr.add_suppression("\\[Warning\\] Found 1 prepared XA transactions");
 --enable_query_log
---source include/restart_mysqld.inc
 
 let INNODB_PAGE_SIZE=`select @@innodb_page_size`;
 let MYSQLD_DATADIR=`select @@datadir`;
 let ALGO=`select @@innodb_checksum_algorithm`;
 let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
-
-show variables like 'innodb_doublewrite';
-show variables like 'innodb_fil_make_page_dirty_debug';
-show variables like 'innodb_saved_page_number_debug';
-
-connect (stop_purge,localhost,root,,);
-START TRANSACTION WITH CONSISTENT SNAPSHOT;
-connection default;
 
 create table t1 (f1 int primary key, f2 blob) engine=innodb;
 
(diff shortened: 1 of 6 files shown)


mysql-test/suite/innodb/t/doublewrite.test
```diff
@@ -48,130 +48,32 @@ insert into t1 values(4, repeat('-',12));
...
+set global innodb_log_checkpoint_now = 1;
 
 --echo # Make the first page dirty for table t1
 set global innodb_saved_page_number_debug = 0;
 set global innodb_fil_make_page_dirty_debug = @space_id;
 
---echo # Ensure that dirty pages of table t1 are flushed.
-set global innodb_buf_flush_list_now = 1;
+SET GLOBAL innodb_max_dirty_pages_pct_lwm=0.0;
+SET GLOBAL innodb_max_dirty_pages_pct=0.0;
```
dr-m, line 71: Are both `innodb_log_checkpoint_now` and the additional buffer pool flushing via `innodb_max_dirty_page_pct` really needed?

mysql-test/suite/innodb/t/doublewrite.test
```diff
@@ -48,130 +48,32 @@ insert into t1 values(4, repeat('-',12));
...
-
---echo # Make the first page dirty for table t1
-set global innodb_saved_page_number_debug = 0;
-set global innodb_fil_make_page_dirty_debug = @space_id;
-
---echo # Ensure that dirty pages of table t1 are flushed.
-set global innodb_buf_flush_list_now = 1;
-
---source include/no_checkpoint_end.inc
+sleep 1;
```
dr-m, line 74: What are we waiting for during this `sleep 1`? That the buffer pool is flushed?

mysql-test/suite/innodb/t/doublewrite.test
```diff
@@ -48,130 +48,32 @@ insert into t1 values(4, repeat('-',12));
...
+SET GLOBAL innodb_max_dirty_pages_pct=0.0;
 
---let CLEANUP_IF_CHECKPOINT=drop table t1;
---source ../include/no_checkpoint_end.inc
 disconnect stop_purge;
-
---echo # Make the first page (page_no=0) of the user tablespace
---echo # full of zeroes.
---echo #
---echo # MDEV-11623: Use old FSP_SPACE_FLAGS in the doublewrite buffer.
```
dr-m, line 79: The commit message fails to mention why this section of the test was removed. The commit message only says "what" but not "why".

Is this test meaningless only for 10.6, due to 86dc7b4d4cfe15a2d37f8b5f60c4fce5dba9491d?

mysql-test/suite/innodb/t/doublewrite.test
```diff
@@ -206,16 +110,18 @@ unlock tables;
...
 
---echo # Ensure that dirty pages of table t1 are flushed.
-set global innodb_buf_flush_list_now = 1;
+SET GLOBAL innodb_max_dirty_pages_pct_lwm=0.0;
+SET GLOBAL innodb_max_dirty_pages_pct=0.0;
 
---source include/no_checkpoint_end.inc
+sleep 1;
+let $shutdown_timeout=0;
+--source include/shutdown_mysqld.inc
```
dr-m, line 124: It is unclear to me why the write of page 1 of `t1` would be observed after the server is killed. Why would this `sleep 1` not typically wait for a checkpoint to be executed? In that case, crash recovery should parse an empty undo log.

dr-m (changes requested):
I’d like to see a run of the 10.6 version of this on buildbot as well. Please push it to a branch name that starts with bb-.

To make the test as close to the 10.6 version as possible, please create the table in the 10.5 version with explicit `STATS_PERSISTENT=0`.

mysql-test/suite/innodb/t/doublewrite.test
Thirunarayanan: we skip the page if we have flag mismatch during `recv_dblwr_t::find_page()` ?
`
commit dbba1bb1c3522090f9a3bc54c107d684bfaed4ab (origin/st-10.5-MDEV-32144, st-10.5-MDEV-32144)
Date:   Thu Oct 19 13:11:20 2023 +0530

    MDEV-31851   After crash recovery, undo tablespace fails to open
    
    recv_recovery_from_checkpoint_start(): InnoDB should add the
    redo log block header + trailer size while checking the log
    sequence number in log file with log sequence number in the
    system tablespace first page.
`

dr-m: Please add the explanation to this commit message, so that it can be easily found in the history without looking up this pull request.

mysql-test/suite/innodb/t/doublewrite.opt
```diff
@@ -0,0 +1 @@
+--innodb_flush_sync=0
```
dr-m, line 1: This parameter will make the test extremely slow. When I ran 20 concurrent instances of the test, only about 3.2 execution cores were occupied on the average. The test execution time would be 25 to 27 seconds. OK, I see that on https://buildbot.mariadb.net/buildbot/builders/fulltest-debug-big-only/builds/13772 (10.4) the execution time is similar, but I think that my system is generally much faster than that builder.

dr-m: If we really need this parameter, can we please use `SET GLOBAL innodb_flush_sync=0` at the exact places where it is needed?

dr-m (changes requested):
Unfortunately, the changed test does not seem to run anywhere on https://buildbot.mariadb.org. There is an [amd64-ubuntu-2004-bigtest](https://buildbot.mariadb.org/#/builders/585) builder, but it does not run any debug instrumented tests. Ideally, this test would need to run fast enough so that it does not need to be marked as big. I am afraid that this test will require debug injection; it may not feasible to remove the dependency on that. At best, we could create a non-debug version of this test that would occasionally fail if a bug is present

One more issue is that the test fails to catch the following bug, which was fixed in a44869d8426f7008c686dcdef2919998a472dbed:
```diff
diff --git a/storage/innobase/log/log0recv.cc b/storage/innobase/log/log0recv.cc
index cad85b4ceca..67826919eab 100644
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -4794,7 +4794,7 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
     if (page_get_page_no(page) != page_id.page_no() ||
         page_get_space_id(page) != page_id.space())
       continue;
-    if (page_id.page_no() == 0)
+    if (true)
     {
       uint32_t flags= mach_read_from_4(
         FSP_HEADER_OFFSET + FSP_SPACE_FLAGS + page);
```
Can you modify the test so that this error will be caught?

PR #2839, opened 2023-11-14
MDEV-32235: mysql_json cannot be used on newly created table

an3l:

## PR quality check

### DIFF
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -3657,6 +3657,19 @@ mysql_prepare_create_table(THD *thd, HA_CREATE_INFO *create_info,
     if (sql_field->vcol_info)
       sql_field->flags&= ~NOT_NULL_FLAG;
 
+    if (sql_field->real_field_type() == MYSQL_TYPE_BLOB &&
+        thd->lex->sql_command == SQLCOM_CREATE_TABLE)
+    {
+      if (!strcmp(sql_field->type_handler()->name().ptr(), "MYSQL_JSON"))
+      {
+        my_printf_error(ER_CANT_CREATE_TABLE,
+                        "Cannot create table %`s.%`s: "
+                        "Run mariadb-upgrade, "
+                        "to upgrade table with mysql_json type.", MYF(0),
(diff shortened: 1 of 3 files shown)


sql/sql_table.cc
```diff
@@ -3657,6 +3657,18 @@ mysql_prepare_create_table(THD *thd, HA_CREATE_INFO *create_info,
     if (sql_field->vcol_info)
       sql_field->flags&= ~NOT_NULL_FLAG;
 
+    if (sql_field->real_field_type() == MYSQL_TYPE_BLOB &&
+        thd->lex->sql_command == SQLCOM_CREATE_TABLE)
```
cvicentiu, line 3661: Isn't SQLCOM_CREATE_TABLE always set here?

Why do we need to check for real_field_type? Isn't the type_handler() check enough?

sql/sql_table.cc
```diff
@@ -3657,6 +3657,18 @@ mysql_prepare_create_table(THD *thd, HA_CREATE_INFO *create_info,
...
+    {
+      if (!strcmp(sql_field->type_handler()->name().ptr(), "MYSQL_JSON"))
+      {
+        my_printf_error(ER_UNKNOWN_ERROR, "Error parsing MySQL JSON format,"
```
cvicentiu, line 3665: I suggest a better error message.

"mysql_json" is only meant to be used for upgrading from MySQL to MariaDB. Tables can not be created with "mysql_json" datatype.

Alternatively, how about use ER_CANT_CREATE_TABLE?

cvicentiu (changes requested):

sql/sql_table.cc
an3l: Yes both checks are needed, because function `mysql_prepare_create_table` is used for `alter` too.

sql/sql_table.cc
an3l: Agree, will apply this error, like
```
ERROR HY000: Cannot create table `test`.`t1`: Run mariadb-upgrade, to upgrade table with mysql_json type.
```

PR #2840, opened 2023-11-14
MDEV-26231:mysql_upgrade attempts to remove plugins which it failed to install

an3l:
When allocating array for installed `MYSQL_JSON` plugin, size is set.
When uninstalling the plugin, check string instead of size.

- 
## PR quality check


an3l:
HI @cvicentiu , yes I have updated now to catch error for query plugin and properly skip uninstall phase if there is no plugin installed at all.
Results:
- Simulate wrong query:
```C
if(!run_query("INSTALL SONAME 'type_mysql_json1'", &ds_result, TRUE))
```
Check clients
```bash
# No plugin installed
MariaDB [test]> show create table t;
ERROR 4161 (HY000): Unknown data type: 'MYSQL_JSON'

# mariadb_upgrade
$ sudo ./client/mysql_upgrade -uroot -S /tmp/mysql.sock --force
Phase 2/7: Installing used storage engines... Skipped
installing plugin for MYSQL_JSON data type
Failed to install the plugin 'type_mysql_json' for MYSQL_JSON data type
FATAL ERROR: Upgrade failed
```

PR #2842, opened 2023-11-15
MDEV-35991 - Add `RUN_ABI_CHECK` to CMake cache

tgross35:
## Description

Allow disabling the ABI check at build time by setting a new CMake flag:

    -DRUN_ABI_CHECK=NO

This option will default to the current default setting

---

This is nice for debugging when you don't want to regenerate `.pp` files for every rebuild. There may already be a way to do this that I am not aware of.

## How can this PR be tested?

Make a change to a `.pp` file (e.g. `plugin_encryption.h.pp`) and try building with and without this option.

## PR quality check

### DIFF
--- a/cmake/abi_check.cmake
+++ b/cmake/abi_check.cmake
@@ -23,10 +23,13 @@
 # (Solaris) sed or diff might act differently from GNU, so we run only 
 # on systems we can trust.
 IF(APPLE OR CMAKE_SYSTEM_NAME MATCHES "Linux")
- SET(RUN_ABI_CHECK 1)
+ SET(RUN_ABI_CHECK_DEFAULT 1)
 ELSE()
- SET(RUN_ABI_CHECK 0)
+ SET(RUN_ABI_CHECK_DEFAULT 0)
 ENDIF()
+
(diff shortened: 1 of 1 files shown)


grooverdan:
can you elaborate on "nice for debugging" and why `.pp` regeneration isn't desirable?

this should be exposed as `OPTION()`

tgross35:
> can you elaborate on "nice for debugging" and why `.pp` regeneration isn't desirable?

I said debugging but meant developing - when experimenting with something like the plugin API that does affect the `pp` files, I have appreciated being able to disable checks easily if it's not something I actually plan to commit. Instead of updating the `.pp` files each time.

Unless, is there a better workflow?

> this should be exposed as `OPTION()`

I can update this.

vuvova:
this option makes sense, I'd say. but frankly, personally I would've simply run
```
make abi_update && make
```

vuvova:
This somehow fell between cracks and meanwhile one of the developers was also annoyed by the abi check and implemented `-DWITHOUT_ABI_CHECK` in b337e14440ba.

So, closing. Sorry for this, you were, indeed, first.

tgross35:
No worries, glad it made it in!

PR #2844, opened 2023-11-15
MDEV-32628: Cryptic ERROR message & inconsistent behavior on incorrect SHOW BINLOG EVENTS FROM ...

bnestere:
Calling SHOW BINLOG EVENTS FROM <offset> with an invalid offset
writes error messages into the server log about invalid reads. The
read errors that occur from this command should only be relayed back
to the user though, and not written into the server log. This is
because they are read-only and have no impact on server operation,
and the client only need be informed to correct the parameter.

This patch fixes this by omitting binary log read errors from the
server when the invocation happens from SHOW BINLOG EVENTS.
Additionally, redundant error messages are omitted when calling the
string based read_log_event from the IO_Cache based read_log_event,
as the later already will report the error of the former.

This PR is separated into two initial commits:
1. The regression to show that various error messages occur in the log
2. The patch which oppresses these error messages

### DIFF
--- a/sql/log_event.h
+++ b/sql/log_event.h
@@ -1322,7 +1322,8 @@ public:
   static Log_event* read_log_event(IO_CACHE* file,
                                    const Format_description_log_event
                                    *description_event,
-                                   my_bool crc_check);
+                                   my_bool crc_check,
+                                   my_bool print_errors= 1);
 
   /**
     Reads an event from a binlog or relay log. Used by the dump thread
@@ -1465,7 +1466,8 @@ public:
   static Log_event* read_log_event(const char* buf, uint event_len,
 				   const char **error,
                                    const Format_description_log_event
-                                   *description_event, my_bool crc_check);
+                                   *description_event, my_bool crc_check,
+                                   my_bool print_errors= 1);
   /**
     Returns the human readable name of the given event type.
   */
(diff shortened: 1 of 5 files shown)


andrelkin (approved):
Thanks! Looks good.

sql/log_event.h
```diff
@@ -1322,7 +1322,8 @@ class Log_event
   static Log_event* read_log_event(IO_CACHE* file,
                                    const Format_description_log_event
                                    *description_event,
-                                   my_bool crc_check);
+                                   my_bool crc_check,
+                                   my_bool print_errors= 1);
```
knielsen, line 1326: Any reason to have a default of "1" here (not "true")?

knielsen (approved):
Looks good, thanks Brandon!

sql/log_event.h
bnestere: I never noticed before, but I suppose when I define and declare a bool in the same statement, I use `1` and `0`, but when the type isn't with the variable, I use `true` and `false` (as seeing in the function call I write `false`). Seems more concise that way to me, but I suppose that's up to preference, and it isn't in our coding guidelines.

Do you have a strong preference that it should be `true`/`false`?

knielsen: No strong preference, as you like. Just noticed that you were using 1 here and false the other place. Not important.

PR #2847, opened 2023-11-16
MDEV-32820 Race condition between trx_purge_free_segment() and trx_undo_create()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32820*

## Description
`trx_purge_free_segment()`: If `fseg_free_step_not_header()` needs to be called multiple times, acquire an exclusive latch on the rollback segment header page after restarting the mini-transaction so that the rest of this function cannot execute concurrently with `trx_undo_create()` on the same rollback segment.

This fixes a regression that was introduced in commit c14a39431b211017e6809bb79c4079b38ffc3dff (MDEV-30753).

## How can this PR be tested?
The normal regression test suite should exercise this. The bug was found via a one-time failure of the test `encryption.create_or_replace` in 23651e27c672aa57d0dfba1251a4fc0abc6c95d6 (which uses `innodb_undo_tablespaces=3` by default). I think that the bug should be repeatable with any setting of `innodb_undo_tablespaces=0`, but it is a very rare finding (only one occurrence in our CI system, not reproduced elsewhere). The assertion `f >= FREED` that failed was introduced in 10.6 aaef2e1d8c843d1e40b1ce0c5199c3abb3c5da28; I did not check if an equivalent assertion is present in 10.5 at all.

## PR quality check

### DIFF
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -365,8 +365,10 @@ void trx_purge_free_segment(mtr_t &mtr, trx_rseg_t* rseg, fil_addr_t hdr_addr)
   while (!fseg_free_step_not_header(TRX_UNDO_SEG_HDR + TRX_UNDO_FSEG_HEADER +
                                     block->frame, &mtr))
   {
-    block->fix();
-    const page_id_t id{block->page.id()};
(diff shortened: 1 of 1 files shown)


vlad-lesin (approved):
rseg_hdr latching to avoid trx_undo_create() and trx_purge_truncate_rseg_history() racing looks reasonable to me.

PR #2848, opened 2023-11-16
MDEV-32820 Race condition between trx_purge_free_segment() and trx_undo_create()

dr-m:
This is the 10.6 version of #2847.

- [x] *The Jira issue number for this PR is: MDEV-32820*

## Description
`trx_purge_free_segment()`: If `fseg_free_step_not_header()` needs to be called multiple times, acquire an exclusive latch on the rollback segment header page after restarting the mini-transaction so that the rest of this function cannot execute concurrently with `trx_undo_create()` on the same rollback segment.

This fixes a regression that was introduced in commit c14a39431b211017e6809bb79c4079b38ffc3dff (MDEV-30753).

## How can this PR be tested?
The normal regression test suite should exercise this. The bug was found via a one-time failure of the test `encryption.create_or_replace` in 23651e27c672aa57d0dfba1251a4fc0abc6c95d6 (which uses `innodb_undo_tablespaces=3` by default). I think that the bug should be repeatable with any setting of `innodb_undo_tablespaces=0`, but it is a very rare finding (only one occurrence in our CI system, not reproduced elsewhere).

## PR quality check

### DIFF
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -349,14 +349,21 @@ trx_purge_add_undo_to_history(const trx_t* trx, trx_undo_t*& undo, mtr_t* mtr)
 }
 
 /** Free an undo log segment.
-@param block     rollback segment header page
+@param rseg_hdr  rollback segment header page
+@param block     undo segment header page
 @param mtr       mini-transaction */
-static void trx_purge_free_segment(buf_block_t *block, mtr_t &mtr)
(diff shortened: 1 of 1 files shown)


vlad-lesin (approved):

dr-m:
I am sorry for the inaccurate "closed without merge" status. I had accidentally deleted the branch before it had been pushed to 10.6. This was in fact closed by 9c5600adde671e84f03aa7b289091e63d4988e70.

PR #2850, opened 2023-11-17
MDEV-23401  main suite test case prints extra row for metadata_lock_info query

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-23401*

## Description
- Added the parameter stats_persistent=0 for InnoDB engine.
- Before printing metadata_lock_info query, make sure that InnoDB does complete purging.

## How can this PR be tested?
mysql-test-run.pl --suite=main should be sufficient for testing the test case changes

## PR quality check

### DIFF
--- a/mysql-test/main/mdl_sync.test
+++ b/mysql-test/main/mdl_sync.test
@@ -3249,6 +3249,7 @@ connection con2;
 SET DEBUG_SYNC= 'now WAIT_FOR table_opened';
 --echo # Check that FLUSH must wait to get the GRL
 --echo # and let DROP PROCEDURE continue
+--source ../suite/innodb/include/wait_all_purged.inc
 SELECT LOCK_MODE, LOCK_TYPE, TABLE_SCHEMA, TABLE_NAME FROM information_schema.metadata_lock_info;
 SET DEBUG_SYNC= 'mdl_acquire_lock_wait SIGNAL grlwait';
 --send FLUSH TABLES WITH READ LOCK
@@ -3274,7 +3275,7 @@ SET DEBUG_SYNC= 'RESET';
 --echo #
 
 create table t1 (a int) engine=myisam;
-create table t2 (a int) engine=innodb;
+create table t2 (a int) stats_persistent=0, engine=innodb;
 insert into t1 values (1);
 insert into t2 values (1);
 
--- a/mysql-test/main/backup_interaction.result
+++ b/mysql-test/main/backup_interaction.result
@@ -95,7 +95,7 @@ drop table t1;
 #
 # BACKUP STAGE performs implicit commits
 #
-create table t1(a int) engine=InnoDB;
+create table t1(a int) stats_persistent=0, engine=InnoDB;
 begin;
 insert into t1 values(1);
 select lock_mode from information_schema.metadata_lock_info;
@@ -197,8 +197,9 @@ drop table t1;
 # CHECK: RO transaction under BACKUP STAGE is a potential deadlock
 # OTOH we most probably allow them under FTWRL as well
 #
-CREATE TABLE t1 (col1 INT) ENGINE = InnoDB;
+CREATE TABLE t1 (col1 INT)stats_persistent=0, ENGINE = InnoDB;
 insert into t1 values (1);
+InnoDB		0 transactions not purged
 backup stage start;
 backup stage block_commit;
 begin;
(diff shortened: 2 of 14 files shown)


mysql-test/main/mdl_sync.test
```diff
@@ -3249,6 +3249,7 @@ connection con2;
 SET DEBUG_SYNC= 'now WAIT_FOR table_opened';
 --echo # Check that FLUSH must wait to get the GRL
 --echo # and let DROP PROCEDURE continue
+--source ../suite/innodb/include/wait_all_purged.inc
```
dr-m, line 3252: This can make the test extremely slow if you do not first `SET GLOBAL innodb_purge_rseg_truncate_frequency=1;` like the include file says. Remember to save and restore the setting when you override it.

I do not think we need this in 10.4. Before ea37b144094a0c2ebfc6774047fd473c1b2a8658, the purge of committed InnoDB transaction history would only acquire MDL when the table contains indexed virtual columns. As far as I can tell, at this point no InnoDB tables have been created in this test. The only InnoDB table in this test would be `t2`, created some 25 lines later.

dr-m (changes requested):
The changes to `stats_persistent` should be applicable to 10.6 only, because of 9608773f75e2ca21491ef6825c3616cdc96d1ca5. The purge related changes should apply to 10.5, not 10.4. As far as I can tell, none of the changed tests contain InnoDB tables with indexed virtual columns, and therefore the purge of history in 10.4 should not acquire any MDL on them.

I think that two new MDEVs need to be filed for the 10.5+ (purge) and 10.6+ related changes. The original 10.4 MDEV-23401 bug report is about a failed write to `mysql.slow_log` due to `EBADF` (bad file descriptor), which these changes should do nothing about. Therefore, we must not associate these testcase-only changes with [MDEV-23401](https://jira.mariadb.org/browse/MDEV-23401).

dr-m:
I cleaned up the test a little. For some reason, a comparison between string and integer would fail, so we will need an explicit cast.
```diff
diff --git a/mysql-test/suite/innodb/t/mem_pressure.test b/mysql-test/suite/innodb/t/mem_pressure.test
index 2d140fb22ad..844a7bce485 100644
--- a/mysql-test/suite/innodb/t/mem_pressure.test
+++ b/mysql-test/suite/innodb/t/mem_pressure.test
@@ -11,15 +11,20 @@
 set @save_dbug=@@debug_dbug;
 
 CREATE TABLE t1 (t TEXT) ENGINE=InnoDB;
-INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR)) FROM seq_1_to_1000;
+INSERT INTO t1 SELECT CONCAT(REPEAT('junk text', 500), CAST(seq AS CHAR))
+FROM seq_1_to_1000;
 
 DROP TABLE t1;
-SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @dirty_prev
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 
 set debug_dbug="d,trigger_garbage_collection";
 SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
 
-SELECT VARIABLE_VALUE < @dirty_prev AS LESS_DIRTY_IS_GOOD FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+SELECT CAST(VARIABLE_VALUE AS INTEGER) < @dirty_prev AS LESS_DIRTY_IS_GOOD
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 
 let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
 let SEARCH_PATTERN= InnoDB: Memory pressure event freed;
```
I have no idea why the integer return value of `get_one_variable()` is being converted to a string. I cleaned up the test a little. The test still occasionally fails; I reproduced with `rr` a case where `buf_pool.flush_list.count` was not changed during the execution of `buf_pool_t::garbage_collect()`. I’ll try to fix that shortly.

Thirunarayanan:
These changes are addressed in [PR#2874](https://github.com/MariaDB/server/pull/2874). so closing this pull request.

PR #2852, opened 2023-11-17
MDEV-25089 : Assertion `error.len > 0' failed in wsrep_status_t galer…

janlindstrom:
…a::ReplicatorSMM::handle_apply_error(galera::TrxHandleSlave&, const wsrep_buf_t&, const string&)

- [x] *The Jira issue number for this PR is: MDEV-25089*

## Description

Problem is that Galera starts TOI (total order isolation) i.e. it sends query to all nodes. Later it is discovered that used engine or other feature is not supported by Galera. Because TOI is executed parallelly in all nodes appliers could execute given TOI and ignore the error and
start inconsistency voting causing node to leave from cluster or we might have a crash as reported.

For example SEQUENCE engine does not support GEOMETRY data type causing either inconsistency between nodes (because some errors are ignored on applier) or crash.

Fixed my adding new function wsrep_check_support to check can Galera support provided CREATE TABLE/SEQUENCE before TOI is started and if not clear error message is provided to the user.

Currently, not supported cases:

* CREATE TABLE ... AS SELECT when streaming replication is used
* CREATE TABLE ... WITH SYSTEM VERSIONING AS SELECT
* CREATE TABLE ... ENGINE=SEQUENCE
* CREATE SEQUENCE ... ENGINE!=InnoDB
* ALTER TABLE t ... ENGINE!=InnoDB where table t is SEQUENCE

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/disabled.def
+++ b/mysql-test/suite/galera/disabled.def
@@ -22,4 +22,4 @@ galera_as_slave_replay : MDEV-32780 galera_as_slave_replay: assertion in the wsr
 galera_slave_replay : MDEV-32780 galera_as_slave_replay: assertion in the wsrep::transaction::before_rollback()
 galera_bf_lock_wait : MDEV-32781 galera_bf_lock_wait test failed
 galera_sst_mysqldump_with_key : MDEV-32782 galera_sst_mysqldump_with_key test failed
-mdev-31285 : MDEV-25089 Assertion `error.len > 0' failed in galera::ReplicatorSMM::handle_apply_error()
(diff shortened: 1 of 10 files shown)


sysprg:
Thanks, the fixes have been merged with the head revision: https://github.com/MariaDB/server/commit/daaa16a47f67a89e57c4679e54e6b7376b1bfa1d + https://github.com/MariaDB/server/commit/0304dbc3275fd6492dc5a1250d5c5e0aacd6890b

PR #2855, opened 2023-11-18
MDEV-32611: Test suite is missing dump option `delete-master-logs`

gulshanpr:
- [x] *The Jira issue number for this PR is: MDEV-32611*

## Description
This patch is for updating the testcase in `rpl_myslqdump_slave.test` for missing mysqldump `delete-master-logs` option

## How can this PR be tested?

by executing `mysql-test-run.pl rpl_mysqldump_slave.test` desire output can be obtained.
an additional note: I added this test at the very end, because adding it in the middle at it's correct place. it changes pre existing result, so i didn't. on suggestion i can do it. 

## PR quality check

### DIFF
--- a/mysql-test/main/rpl_mysqldump_slave.result
+++ b/mysql-test/main/rpl_mysqldump_slave.result
@@ -65,4 +65,13 @@ SET GLOBAL gtid_slave_pos='0-2-1003';
 -- CHANGE MASTER TO MASTER_LOG_FILE='slave-bin.000001', MASTER_LOG_POS=BINLOG_START;
 CHANGE MASTER TO MASTER_USE_GTID=slave_pos;
 SET GLOBAL gtid_slave_pos='0-2-1003';
+connection master;
+CREATE TABLE t (
+id int
+);
+insert into t values (1);
+insert into t values (2);
+drop table t;
+-- CHANGE MASTER TO MASTER_LOG_FILE='master-bin.000002', MASTER_LOG_POS=BINLOG_START;
+-- SET GLOBAL gtid_slave_pos='0-1-1005';
 include/rpl_end.inc
--- a/mysql-test/main/rpl_mysqldump_slave.test
+++ b/mysql-test/main/rpl_mysqldump_slave.test
@@ -83,6 +83,46 @@ DROP TABLE t2;
 --replace_regex /MASTER_LOG_POS=[0-9]+/MASTER_LOG_POS=BINLOG_START/
 --exec $MYSQL_DUMP_SLAVE --compact --master-data --single-transaction --gtid test
 
+#
+# MDEV-32611 Added test for mysqldump --delete-master-logs option.
+# This options is alias of 
+# get binlogs: show master status -> flush logs -> purge binary logs to <new_binlog>
+# sequence and this test is derived using the same pattern.
+#
 
+connection master;
+
+CREATE TABLE t (
+    id int
+);
+
+insert into t values (1);
+insert into t values (2);
+
+drop table t;
+
+--let $predump_binlog_filename= query_get_value(SHOW MASTER STATUS, File, 1)
+
+# Execute mysqldump with delete-master-logs option
+--replace_regex /MASTER_LOG_POS=[0-9]+/MASTER_LOG_POS=BINLOG_START/
+--exec $MYSQL_DUMP --compact --no-create-info --no-data --delete-master-logs test
+
+--let $postdump_binlog_filename= query_get_value(SHOW MASTER STATUS, File, 1)
+
+--let $postdump_first_binary_log_filename= query_get_value(SHOW BINARY LOGS, Log_name, 1)
+
+if ($predump_binlog_filename == $postdump_binlog_filename)
+{
+  --echo # predump_binlog_filename: $predump_binlog_filename
+  --echo # postdump_binlog_filename: $postdump_binlog_filename
+  --die Master state didn't change after mariadb-dump with --delete-master-logs.
+}
+
+if ($postdump_first_binary_log_filename != $postdump_binlog_filename)
+{
+  --echo # postdump_first_binary_log_filename: $postdump_first_binary_log_filename
+  --echo # postdump_binlog_filename: $postdump_binlog_filename
+  --die Master binlog wasn't deleted after mariadb-dump with --delete-master-logs.
+}
 
 --source include/rpl_end.inc


mysql-test/main/rpl_mysqldump_slave.test
```diff
@@ -86,3 +86,52 @@ DROP TABLE t2;
...
+  --echo # Master state changed after mariadb-dump with --delete-master-logs
+}
+
+if ($predump_binlog_pos != $postdump_binlog_pos)
+{
+  --echo # predump_binlog_pos: $predump_binlog_pos
+  --echo # postdump_binlog_pos: $postdump_binlog_pos
+  --echo # Master state changed after mariadb-dump with --delete-master-logs
+}
+
```
bnestere, line 136: I think we should add another assertion to ensure that the actual log deletion happened, in addition to the log state change verification. To do so, after the `$MYSQL_DUMP`, we should save the log name of the first row of `SHOW BINARY LOGS` (with something like `--let $postdump_first_binary_log_file= query_get_value(SHOW BINARY LOGS, Log_name, 1)`), which should be the first row present on the master, and compare it with the log name of `SHOW MASTER STATUS`, and ensure they are the same (and `--die` if they are different).

bnestere (changes requested):
Hi @gulshanpr !

Thanks for working on this! I've added a few notes as comments. Let me know if you have any questions about them.

mysql-test/main/rpl_mysqldump_slave.test
```diff
@@ -86,3 +86,52 @@ DROP TABLE t2;
...
 --source include/rpl_end.inc
+
+#
+# MDEV-32611 Added test for mysqldump --delete-master-logs option
+#
+
+connection master;
+use test;
+
+--source include/have_binlog_format_row.inc
```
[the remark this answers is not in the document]
gulshanpr: if that's the issue, can't I use `--source include/have_log_bin.inc`, as you told me in [zulip](https://mariadb.zulipchat.com/#narrow/stream/118760-new-members/topic/MDEV-32611/near/402743054).
or as i have included this `--source include/have_binlog_format_row.inc` at the very end, will this still create a problem for the above test?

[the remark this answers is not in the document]
bnestere: Good connection! `have_log_bin.inc` is actually already included in this file though :) Looking through the include chain of `--source include/master-slave.inc` (seen at `<src_dir>/mysql-test/include/master-slave.inc`), you'll see it calls `--source include/rpl_init.inc`, which in-turn calls `--source include/have_log_bin.inc`. 

To elaborate a little more on how this works, when you run `./mtr main.rpl_mysqldump_slave`, you'll see 3 separate tests be run, each re-running the entire test with a different binlog format. By including `--source include/have_binlog_format_<format>.inc`, the test will _only_ run under those hard-coded binlog formats (note you can combine them). So its presence would reduce the test space.

mysql-test/main/rpl_mysqldump_slave.test
gulshanpr: compared `--let $postdump_first_binary_log_filename= query_get_value(SHOW BINARY LOGS, Log_name, 1)` with pre-existing `$postdump_binlog_filename`

mysql-test/main/rpl_mysqldump_slave.test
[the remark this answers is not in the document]
gulshanpr: 1. 
>we'll have to modify the output of our --exec $MYSQL_DUMP command to mask the binary log position. See my comment on that line for the details.

I guess you were talking about this?
`--replace_regex /MASTER_LOG_POS=[0-9]+/MASTER_LOG_POS=BINLOG_START/`

2. removed `--source include/have_binlog_format_row.inc` so that test can run for all three log format. didn't additionally added `have_bin_log.inc` cause it already there through `master-slave.inc`.

gulshanpr:
Apology for the minor issue, that cause you work little more.

Additionally, 
If I do not insert any value into the table or don't even create a table, `delete-master-logs` still works.

mysql-test/main/rpl_mysqldump_slave.test
```diff
@@ -83,6 +83,52 @@ DROP TABLE t2;
...
+  --echo # predump_binlog_pos: $predump_binlog_pos
+  --echo # postdump_binlog_pos: $postdump_binlog_pos
+  --die Master state didn't change after mariadb-dump with --delete-master-logs.
+}
+
+if ($postdump_first_binary_log_filename != $postdump_binlog_filename)
+{
+  --echo # postdump_first_binary_log_filename: $postdump_first_binary_log_filename
+  --echo # postdump_binlog_filename: $postdump_binlog_filename
+  --die Master state didn't change after mariadb-dump with --delete-master-logs.
```
bnestere, line 129: This assertion is to test that the old logs were deleted rather than the master log state changing, so this message should be changed to be clear about that.

bnestere (commented):
Hi Gulshan! Looks good, I only have last one very minor note to address.

Additionally, can you squash the commits together, and format the commit message as

```
MDEV-<JIRA MDEV #>: <JIRA title description>

<Your description here>
```

I'd also like @an3l to take a look, as he noticed there was a missing test in the first place.

Thanks again for the contribution, @gulshanpr !

mysql-test/main/rpl_mysqldump_slave.test
gulshanpr: Sure, doing it now.

bnestere:
> Apology for the minor issue, that cause you to work little more.
> 
> Additionally, If I do not insert any value into the table or don't even create a table, `delete-master-logs` still works.

Nothing to apologize for :) I'm grateful for your effort!

To the test passing without the table/transactions, good observation. When we spoke on Zulip, I didn't realize that you were going to extend the `rpl_mysqldump_slave.test` test, which earlier already created/dropped a table which had already modified the binary log. Conversely, if you created a new test from scratch without having any previous statements written into the binary log, I'd imagine the test would fail with the `pos` assertion being stagnant, as the initial size of the rotated log should be the same as the initial. Feel free to experiment with it :) You can use the tool `mariadb-binlog` to look at the actual contents of the saved binary logs in `<build_dir>/mysql-test/var/mysqld.1/data/master-bin.000001` (or whichever number binary log you want to examine).

mysql-test/main/rpl_mysqldump_slave.test
```diff
@@ -83,6 +83,52 @@ DROP TABLE t2;
...
+--let $postdump_first_binary_log_filename= query_get_value(SHOW BINARY LOGS, Log_name, 1)
+
+if ($predump_binlog_filename == $postdump_binlog_filename)
+{
+  --echo # predump_binlog_filename: $predump_binlog_filename
+  --echo # postdump_binlog_filename: $postdump_binlog_filename
+  --die Master state didn't change after mariadb-dump with --delete-master-logs.
+}
+
+if ($predump_binlog_pos == $postdump_binlog_pos)
```
an3l, line 118: I think we should remove everything related to position, there is no point in checking the position if files are different.

mysql-test/main/rpl_mysqldump_slave.test
```diff
@@ -83,6 +83,52 @@ DROP TABLE t2;
...
+
+connection master;
 
+CREATE TABLE t (
+    id int
+);
+
+insert into t values (1);
+insert into t values (2);
+
```
an3l, line 98: I would move `drop table t` here.

bnestere:
Hi @an3l ! Thanks for doing a review!

To

>That said, we have to update the [documentation](https://mariadb.com/kb/en/mariadb-dump/) , that states:
> This option automatically enables --master-data.
> I couldn't find that opt_delete_master_logs sets opt_master_data in client/mysqldump.c, but that should be next step

I see it in the `get_options()` function:
```
  if (opt_delete_master_logs && !opt_master_data)
    opt_master_data= MYSQL_OPT_MASTER_DATA_COMMENTED_SQL;
```

Though note that the `flush logs` step that happens when enabling `--delete-master-logs` isn't mentioned in the documentation, and should be (though I can update that).


And with
> Please update in commit message that the options is alias to following sequence
> get binlogs: show master status -> flush logs -> purge binary logs to <new_binlog>

Good idea! I'd also like to see this added to the echo/comment around like `87` in the test to help describe what behavior the test is actually validating.

gulshanpr:
Hii, @an3l, @bnestere
I have made the suggested changes and added
>Please update in commit message that the options is alias to following sequence
>get binlogs: show master status -> flush logs -> purge binary logs to <new_binlog>

into commit message as well as echo/comment at the start of test.

mysql-test/main/rpl_mysqldump_slave.test
```diff
@@ -83,6 +83,46 @@ DROP TABLE t2;
...
+--let $predump_binlog_filename= query_get_value(SHOW MASTER STATUS, File, 1)
+
+# Execute mysqldump with delete-master-logs option
+--replace_regex /MASTER_LOG_POS=[0-9]+/MASTER_LOG_POS=BINLOG_START/
+--exec $MYSQL_DUMP --compact --no-create-info --no-data --delete-master-logs test
+
+--let $postdump_binlog_filename= query_get_value(SHOW MASTER STATUS, File, 1)
+
+--let $postdump_first_binary_log_filename= query_get_value(SHOW BINARY LOGS, Log_name, 1)
```
an3l, line 112: I think one of this is redundant (`show master status` or `show binary logs`).

an3l (approved):

mysql-test/main/rpl_mysqldump_slave.result
```diff
@@ -65,4 +65,13 @@ SET GLOBAL gtid_slave_pos='0-2-1003';
...
 CHANGE MASTER TO MASTER_USE_GTID=slave_pos;
 SET GLOBAL gtid_slave_pos='0-2-1003';
+connection master;
+CREATE TABLE t (
+id int
+);
+insert into t values (1);
+insert into t values (2);
+drop table t;
+-- CHANGE MASTER TO MASTER_LOG_FILE='master-bin.000002', MASTER_LOG_POS=BINLOG_START;
```
an3l, line 75: @bnestere  regarding
```
  if (opt_delete_master_logs && !opt_master_data)
    opt_master_data= MYSQL_OPT_MASTER_DATA_COMMENTED_SQL;
```
according to the documentation, the result is good, but according to above code, result should be commented, so that's why I asked where where `opt_master_data` is set to `#define MYSQL_OPT_MASTER_DATA_EFFECTIVE_SQL 1` ?

mysql-test/main/rpl_mysqldump_slave.test
bnestere: They check different aspects of the behavior. The check with `show master status` verifies `flush logs` occurred, and the check with `show binary logs` verifies `purge binary logs` occurred.

mysql-test/main/rpl_mysqldump_slave.result
bnestere: I'm not sure what you mean here. The result is commented (prefixed with `--`). Though I do think the documentation is ambiguous, as it doesn't specify the `opt_master_data` value, so we should extend that to say 
> This option automatically enables --master-data=2.

an3l: I was thinking that the result shouldn't be commented, by enabling `opt_master_data` set to 1, but that make sense agree, we will update the documentation than. Thanks @bnestere  and @gulshanpr .

an3l:
Thanks @gulshanpr merged to 10.4.

gulshanpr:
Thank You, @an3l and @bnestere for merging and helping me write code throughout this issue. 
I've made many small mistakes, I'll take care not to repeat it.

PR #2856, opened 2023-11-20
MDEV-28971 SEQUENCEs do not work with streaming replication

sciascid:
Return an error if user attempts to use SEQUENCEs in combination with streaming replication in a Galera cluster. This is currently not supported.


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/9ab7dbc3bec4e7d6083c03aa7cbabbd1d695a131

PR #2857, opened 2023-11-20
MDEV-32374 log_sys.lsn_lock is a performance hog

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32374*

## Description
The `log_sys.lsn_lock` that was introduced in commit a635c40648519fd6c3729c9657872a16a0a20821 had better be located in the same cache line with `log_sys.latch` so that `log_t::append_prepare()` needs to modify only two first cache lines where `log_sys` is stored.

`log_t::lsn_lock`: On Linux, change the type from `pthread_mutex_t` to something that may be as small as 32 bits, to pack more data members in the same cache line. On Microsoft Windows, `CRITICAL_SECTION` works better.

`log_t::check_flush_or_checkpoint_`: Renamed to `need_checkpoint`. There is no need to pause all writer threads in `log_free_check()` when we only need to write `log_sys.buf` to `ib_logfile0`. That will be done in `mtr_t::commit()`.

`log_t::append_prepare_wait()`: Make the member function non-static to simplify the call interface, and add a parameter for the LSN.

`log_t::append_prepare()`: Invoke append_prepare_wait() at most once. Only `set_check_for_checkpoint()` if a log checkpoint needs to be written. If the log buffer needs to be written, we will take care of it ourselves later in our caller. This will reduce interference with `log_free_check()` in other threads.

`mtr_t::commit()`: Call `log_write_up_to()` if needed.

`log_t::get_write_target()`: Return a `log_write_up_to()` target to `mtr_t::commit()`.

`buf_flush_ahead()`: If we are in furious flushing, call `log_sys.set_check_for_checkpoint()` so that all writers will wait in `log_free_check()` until the checkpoint is done. Otherwise, the test `innodb.insert_into_empty` could occasionally report an error "Crash recovery is broken".

`log_check_margins()`: Replaced by `log_free_check()`.

`log_flush_margin()`: Removed. This is part of `mtr_t::commit()` and other operations that write log.

`log_t::create()`, `log_t::attach()`: Guarantee that `buf_free < max_buf_free` will always hold on PMEM, to satisfy an assumption of `log_t::get_write_target()`.

`log_write_up_to()`: Assert `lsn!=0`. Such calls are not incorrect, but it is cheaper to test that single unlikely condition in `mtr_t::commit()` rather than test several conditions in `log_write_up_to()`.

`innodb_drop_database()`, `unlock_and_close_files()`: Check the LSN before calling `log_write_up_to()`.

`ha_innobase::commit_inplace_alter_table()`: Remove redundant calls to `log_write_up_to()` after calling `unlock_and_close_files()`.

## How can this PR be tested?
This is a low-level change that is rather well exercised by the existing tests in the regression test suite, except that our CI systems mostly use the "fake PMEM" by default. @mleich1 tested this with RQG, specifically disabling the use of PMEM.

This is a performance fix. I tested it with a very skewed microbenchmark: Sysbench `oltp_update_non_index` on 256 tables, 10 rows per table, 256 threads, 10 GiB buffer pool (large enough for the workload), 4 GiB log file, and the minimum `--innodb-log-buffer-size=2m`, running on Linux `/dev/shm` but with `cmake -DCMAKE_DISABLE_FIND_PACKAGE_PMEM=1`. I also tested a more reasonable `--innodb-log-buffer-size=128m`. In either configuration, the modified code did not perform worse than the original.

Steve Shaw tested b58e1cef7f2f5294c393a3709819091213d3a0d5 with HammerDB and reported a 5% improvement in throughput at lower virtual user counts (up to the number of hardware threads). In that version, the `log_write_up_to()` call in `mtr_t::commit()` was unconditional; @vaintroub suggested to add the condition.

## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0bulk.cc
+++ b/storage/innobase/btr/btr0bulk.cc
@@ -969,10 +969,10 @@ BtrBulk::pageCommit(
 /** Log free check */
 inline void BtrBulk::logFreeCheck()
 {
-	if (log_sys.check_flush_or_checkpoint()) {
+	if (log_sys.check_for_checkpoint()) {
 		release();
 
-		log_check_margins();
+		log_free_check();
 
 		latch();
 	}
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -2124,6 +2124,8 @@ ATTRIBUTE_COLD void buf_flush_ahead(lsn_t lsn, bool furious)
       limit= lsn;
       buf_pool.page_cleaner_set_idle(false);
       pthread_cond_signal(&buf_pool.do_flush_list);
+      if (furious)
+        log_sys.set_check_for_checkpoint();
     }
     mysql_mutex_unlock(&buf_pool.flush_list_mutex);
   }
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -3078,7 +3078,7 @@ ATTRIBUTE_NOINLINE ATTRIBUTE_COLD void mtr_t::name_write()
 and write out FILE_MODIFY if needed, and write FILE_CHECKPOINT.
 @param lsn  checkpoint LSN
 @return current LSN */
-lsn_t fil_names_clear(lsn_t lsn)
+ATTRIBUTE_COLD lsn_t fil_names_clear(lsn_t lsn)
 {
 	mtr_t	mtr;
 
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -1575,7 +1575,8 @@ static void innodb_drop_database(handlerton*, char *path)
       ibuf_delete_for_discarded_space(id);
 
     /* Any changes must be persisted before we return. */
-    log_write_up_to(mtr.commit_lsn(), true);
+    if (mtr.commit_lsn())
+      log_write_up_to(mtr.commit_lsn(), true);
   }
 
   my_free(namebuf);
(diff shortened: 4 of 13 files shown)


vaintroub:
Looks OK to me, but of course I have to trust your expertise for the most of the patch :)

One thing - only a small part of the patch here is obviously improving lsn_lock, reducing its size on Linux, and packing log_sys members together, to share the cache line.
There could be a better explanation, inside a patch comment, what the rest of the patch has to do "log_sys.lsn_lock is a performance hog". Does it do less under the lock, and how?

dr-m:
During a test of 4d2b6ab9834603639882c0bbd639a91ff7d6279d, @mleich1 produced a recovery failure due to an incorrect tablespace ID:
```
2023-11-20 19:53:32 0 [ERROR] InnoDB: Space id and page no stored in the page, read in are [page id: space=50315264, page number=2], should be [page id: space=318, page number=2]
```
This looks like a mystery problem in the environment, and in no way related to this change. In the `rr replay` of the server process that was killed right before the failed recovery attempt, everything was written correctly to this page, with the same `FIL_PAGE_LSN` that was present during recovery. In the file system, this page is corrupted. Not only the tablespace ID is wrong (as noted above), but also at the end of the 4096-byte page we had NUL bytes up to the end. The expected 32 bits of LSN and checksum were all 0.

vaintroub (approved):

dr-m:
> Looks OK to me, but of course I have to trust your expertise for the most of the patch :)
> 
> One thing - only a small part of the patch here is obviously improving lsn_lock, reducing its size on Linux, and packing log_sys members together, to share the cache line. There could be a better explanation, inside a patch comment, what the rest of the patch has to do "log_sys.lsn_lock is a performance hog". Does it do less under the lock, and how?

You are right, the bug title and description could have been clarified. It has been a long time since I worked on this. I think that the only thing that we do less under the `log_sys.lsn_lock` is that we access one cache line less. A bigger improvement is that there no longer is a flag that asks every thread to try writing the log buffer to the file. Instead, each thread does that on its own in `mtr_t::commit()` if `log_sys.get_write_target()` returned nonzero.

PR #2858, opened 2023-11-20
MDEV-32781 galera_bf_lock_wait test failed

sciascid:
This test happens to fail if it runs after test
galera_inject_bf_long_wait.
And the reason is that galera_bf_lock_wait greps for message "BF lock wait long" in the error log, and expects that grep matches no lines. Whereas galera_inject_bf_long_wait intentionally causes the message to appear in the log. The fix consists in using assert_grep.inc with option assert_only_after, such that galera_bf_lock_wait is limited to grep only those lines that appeared in the log after it started to execute.


sysprg:
Thanks, the fix has been added to the main branch: https://github.com/MariaDB/server/commit/5c4c1844bf8b60dcee9fdeacd4a39705d40a4515

PR #2860, opened 2023-11-20
MDEV-32068 Some calls to buf_read_ahead_linear() seem to be useless

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32068*

## Description
The linear read-ahead (enabled by nonzero `innodb_read_ahead_threshold`) works best if index leaf pages or undo log pages have been allocated on adjacent page numbers. The read-ahead is assumed not to be helpful in other types of page accesses, such as non-leaf index pages.

We will make the callers of `buf_page_get_low()` responsible for invoking one or both of `buf_page_make_young_if_needed()` and `buf_read_ahead_linear()`. We will invoke them in those callers of `buf_page_get_gen()` or `buf_page_get()` where it makes sense: the access is not one-time-on-startup and the page and not going to be freed soon.

We will not invoke linear read-ahead in functions that would essentially allocate or free pages. Freshly allocated pages are expected to be initialized by `buf_page_create()` and not read from a data file. Likewise, freeing pages should not involve accessing any sibling pages, except for freeing singly-linked lists of BLOB pages.

## How can this PR be tested?
These is a performance oriented change that should be somewhat exercised by the regression test suite. For evaluating the performance impact, we need tests where the active working set is larger than the buffer pool. @hgxl64 already ran some tests and did not observe any difference. The experiment 6415d851c8b81497df871a3c8cb1cbacea1ea504 could be worth testing separately.

This a performance improvement, and the 10.6 release series is the earliest one where most performance fixes have been applied.

## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0btr.cc
+++ b/storage/innobase/btr/btr0btr.cc
@@ -216,10 +216,11 @@ ATTRIBUTE_COLD void btr_decryption_failed(const dict_index_t &index)
 @param[in]	merge	whether change buffer merge should be attempted
 @param[in,out]	mtr	mini-transaction
 @param[out]	err	error code
+@param[out]	first	set if this is a first-time access to the page
 @return block */
 buf_block_t *btr_block_get(const dict_index_t &index,
                            uint32_t page, rw_lock_type_t mode, bool merge,
-                           mtr_t *mtr, dberr_t *err)
+                           mtr_t *mtr, dberr_t *err, bool *first)
 {
   ut_ad(mode != RW_NO_LATCH);
   dberr_t local_err;
@@ -242,6 +243,8 @@ buf_block_t *btr_block_get(const dict_index_t &index,
       *err= DB_PAGE_CORRUPTED;
       block= nullptr;
     }
+    else if (!buf_page_make_young_if_needed(&block->page) && first)
+      *first= true;
   }
   else if (*err == DB_DECRYPTION_FAILED)
     btr_decryption_failed(index);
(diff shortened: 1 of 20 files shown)


vlad-lesin (approved):
LGTM

dr-m:
I tested the performance with an I/O bound workload that would grow the InnoDB system tablespace `ibdata1` to 2.1GiB and the `.ibd` files to 221MiB: Sysbench `oltp_update_index`, 5 tables of 10,000 rows each, 256 concurrent connections, 60 seconds, with a tiny 100MiB buffer pool and log file size, running on Linux ext4fs on a fast NVMe drive. This was with the default `innodb_read_ahead_threshold=56`.
| revision | throughput/tps | history list length | slow shutdown time/s |
|:--------:| --------------:| -------------------:| --------------------:|
|10.6 768a736174d6caf09df43e84b0c1b9ec52f1a301|112655.00|6770647|192|
|with de971d32f39ea0e1fb5e019f68b0743c7fb382d1|113104.43|6792178|186|
|with 6415d851c8b81497df871a3c8cb1cbacea1ea504|112257.87|6740597|186|

I repeated the experiment with `innodb_log_file_size=2g` (and `innodb_buffer_pool_size=100m`) to reduce the relative amount of checkpoint flushing:
| revision | throughput/tps | history list length | slow shutdown time/s |
|:--------:| --------------:| -------------------:| --------------------:|
|10.6 768a736174d6caf09df43e84b0c1b9ec52f1a301|115940.60|6963789|198|
|with de971d32f39ea0e1fb5e019f68b0743c7fb382d1|116459.46|6991899|199|
|with 6415d851c8b81497df871a3c8cb1cbacea1ea504|113055.56|6787135|197|

Based on the average throughput, the linear read-ahead of undo log pages appears to be somewhat useful and we must reject the experiment 6415d851c8b81497df871a3c8cb1cbacea1ea504.

PR #2861, opened 2023-11-21
MDEV-24507: Server Crash using UDF in WHERE clause of VIEW

SeanCAdams:
- [x] *The Jira issue number for this PR is: MDEV-24507*

## Description
These changes are submitted under the BSD 3-clause License.

When the UDF does not have a _deinit function the server crashes in udf_handler::cleanup (sql/item_func.cc:3467). When the UDF has both an _init and a _deinit function but _init does not allocate memory for initid->ptr the server crashes in udf_handler::cleanup (sql/item_func.cc:3467). When the UDF has both an _init and a _deinit function and allocates/deallocates  memory for initid->ptr the server crashes in the memory deallocation of the _deinit function.

The sequence of events seen are:
  1. A UDF, U, is created for the query.
  2. The UDF _init function is called using U->initid.
  3. U is cloned for the sub-query using the [default|implicit] copy constructor, resulting in V.
  4. The UDF _init function is called using V->initid.  U->initid and V->initid are the same value.
  5. The UDF function is called.
  6. The UDF _deinit function is called using U->initid.  If any memory was allocated for initid->ptr it is deallocated here.
  7. udf_handler::cleanup deletes the U->buffers String array.
  8. The UDF _deinit function is called using V->initid.  If any memory was allocated for initid->ptr it was previously deallocated and _deinit crashes the server.
  9. udf_handler::cleanup deletes the V->buffers String array. V->buffers was the same values as U->buffers which was already deallocated.  The server crashes.

The solution is to create a[n explicit] copy constructor for udf_handler which sets not_original to true.  Later, not_original is set back to false (0) after udf_handler::fix_fields has set up a new value for initid->ptr.

## How can this PR be tested?

Automated tests added.

- [x] *This is a bug fix and the PR is based against the maintained branch that ships with our product.*

## PR quality check


grooverdan:
Very nice. Thank you very much.

Can you rebase to 10.4 where the problem is still present?
```
git fetch git@github.com:MariaDB/server.git 10.4
# adjust test case wrt to 10.4 boundary and re-record/
git commit --amend 
git push -f
``

Don't worry about the branch name being incorrect. Then edit the title of this PR and change the target branch to 10.4.

LinuxJedi (approved):
Nice work!

LinuxJedi:
Rebased on 10.5 because 10.4 is near EoL and the final release is for security / blocker fixes only.

PR #2864, opened 2023-11-23
MDEV-24670 memory pressure - eventfd rather than pipe

grooverdan:
After discovering eventfds for #2807 it simplifies the current implementation a little here too.

Like #2807 we don't get to read from the eventfd in the quit case as the m_abort flag is set.

- [x] *The Jira issue number for this PR is: MDEV-24670*

## Description

Eventfds have a simpler interface and are one file descriptor rather than two.

Reuse the patten of the accepting socket connections by testing for abort after a poll returns. This way the same event descriptor can be used for Quit
and debugging trigger.

Also correct the registration of mem pressure file descriptors.

## How can this PR be tested?

mem_pressure test already.

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -744,6 +744,7 @@ bool buf_page_is_corrupted(bool check_lsn, const byte *read_buf,
 
 #ifdef __linux__
 #include <poll.h>
+#include <sys/eventfd.h>
 #include <fstream>
 
 /** Memory Pressure
@@ -752,14 +753,16 @@ based off https://www.kernel.org/doc/html/latest/accounting/psi.html#pressure-in
 and https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#memory */
 class mem_pressure
 {
-  /* triggers + pipe */
+  /* triggers + eventfd */
   struct pollfd m_fds[3];
   nfds_t m_num_fds;
-  int m_pipe[2]= {-1, -1};
+  int m_event_fd= -1;
+  Atomic_relaxed<bool> m_abort= false;
 
   std::thread m_thd;
-  /* mem pressure garbarge collection resticted to interval */
+  /* mem pressure garbage collection restricted to interval */
   static constexpr ulonglong max_interval_us= 60*1000000;
+
 public:
   mem_pressure() : m_num_fds(0) {}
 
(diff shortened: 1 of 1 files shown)


storage/innobase/buf/buf0buf.cc
```diff
@@ -752,14 +753,16 @@ based off https://www.kernel.org/doc/html/latest/accounting/psi.html#pressure-in
...
   struct pollfd m_fds[3];
   nfds_t m_num_fds;
-  int m_pipe[2]= {-1, -1};
+  int m_event_fd= -1;
 
   std::thread m_thd;
-  /* mem pressure garbarge collection resticted to interval */
+  /* mem pressure garbage collection restricted to interval */
   static constexpr ulonglong max_interval_us= 60*1000000;
+  static bool m_abort;
```
dr-m, line 764: If this were declared non-`static` we would not need a separate initialization. Maybe, to silence any ThreadSanitizer warning, this should be `Atomic_relaxed<bool>`.

storage/innobase/buf/buf0buf.cc
```diff
@@ -921,23 +923,17 @@ void mem_pressure::pressure_routine(mem_pressure *m)
...
-            goto shutdown;
-#ifdef UNIV_DEBUG
-          case 'G':
-            buf_pool.garbage_collect();
-#endif
-          }
+        uint32 u;
+        /* we haven't aborted, so this must be a debug trigger */
+        if (read(p.fd, &u, sizeof(u)) >=0)
+          buf_pool.garbage_collect();
```
dr-m, line 932: We wrote `uint64_t` but are reading only `uint32`. Could this be a single-byte read and write?

dr-m (commented):
Looks OK with a couple cleanups.

storage/innobase/buf/buf0buf.cc
grooverdan: eventfd prescribes 64 bit

storage/innobase/buf/buf0buf.cc
grooverdan: missed the obvious non-static implementation, thanks.

storage/innobase/buf/buf0buf.cc
```diff
@@ -921,23 +922,17 @@ void mem_pressure::pressure_routine(mem_pressure *m)
...
-          case 'Q':
-            goto shutdown;
-#ifdef UNIV_DEBUG
-          case 'G':
-            buf_pool.garbage_collect();
-#endif
-          }
+        uint64_t u;
+        /* we haven't aborted, so this must be a debug trigger */
+        if (read(p.fd, &u, sizeof(u)) >=0)
```
dr-m, line 930: We had written 8 bytes, but here (depending on the kernel-side implementation) we could theoretically read that in multiple requests, so a single debug event could theoretically trigger multiple garbage collections. Not a big deal.

dr-m (approved):

storage/innobase/buf/buf0buf.cc
grooverdan: the way its describes seems rather strict on 8 bytes. But who doesn't like a bit of debug fuzzing :-)

PR #2865, opened 2023-11-23
MDEV-31817 SIGSEGV after btr_page_get_father_block() returns nullptr on corrupted data

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-31817*

## Description
A stress test that makes use of `ROW_FORMAT=COMPRESSED` tables caused a crash in `btr_attach_half_pages()` due to a null pointer that had been returned by `btr_page_get_father_block()`, signifying that something is corrupted. The root cause of that corruption is unknown at the moment, but in the spirit of 0b47c126e31cddda1e94588799599e138400bcf8 we should avoid crashes when the database is corrupted.

`btr_attach_half_pages()`, `btr_lift_page_up()`, `btr_compress()`: Return `DB_CORRUPTION` if `btr_page_get_father_block()` returns `nullptr`.

## How can this PR be tested?
Rerun the stress test that was described in [MDEV-31817](https://jira.mariadb.org/browse/MDEV-31817), preferrably in the same environment. The added error handling code is unreachable in the regression test suite.

It is not feasible to fix this in older releases where 0b47c126e31cddda1e94588799599e138400bcf8 is absent. InnoDB would easily crash in many other places.

## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0btr.cc
+++ b/storage/innobase/btr/btr0btr.cc
@@ -873,7 +873,8 @@ static rec_offs *btr_page_get_parent(rec_offs *offsets, mem_heap_t *heap,
 /************************************************************//**
 Returns the upper level node pointer to a page. It is assumed that mtr holds
 an x-latch on the tree.
-@return rec_get_offsets() of the node pointer record */
+@return rec_get_offsets() of the node pointer record
+@retval nullptr on corruption */
 static
 rec_offs*
 btr_page_get_father_block(
@@ -2541,6 +2542,11 @@ btr_attach_half_pages(
 		offsets = btr_page_get_father_block(nullptr, heap, mtr,
 						    &cursor);
 
+		if (UNIV_UNLIKELY(!offsets)) {
+			mem_heap_free(heap);
+			return DB_CORRUPTION;
+		}
+
 		/* Replace the address of the old child node (= page) with the
 		address of the new lower half */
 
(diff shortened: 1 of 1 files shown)


dr-m:
Just to clarify, 3dc8ab782d7489a2e5a9f6bdc3295a3a835ed1e6 will not be part of the final fix. It is only there for testing purposes.

storage/innobase/btr/btr0btr.cc
```diff
@@ -844,14 +844,20 @@ static rec_offs *btr_page_get_parent(rec_offs *offsets, mem_heap_t *heap,
...
           return nullptr;
+        }
         offsets= rec_get_offsets(cursor->page_cur.rec, index, offsets, 0,
                                  ULINT_UNDEFINED, &heap);
         p= btr_node_ptr_get_child_page_no(cursor->page_cur.rec, offsets);
         if (p != page_no)
         {
           if (btr_page_get_level(block->page.frame) == level)
+          {
+            sql_print_information("InnoDB: btr_page_get_parent 2");
```
Thirunarayanan, line 858: I think it shouldn't be part of the final patch.

storage/innobase/btr/btr0btr.cc
```diff
@@ -867,13 +873,15 @@ static rec_offs *btr_page_get_parent(rec_offs *offsets, mem_heap_t *heap,
         return offsets;
       }
 
+  sql_print_information("InnoDB: btr_page_get_parent 3");
```
Thirunarayanan, line 876: same goes for here

Thirunarayanan (approved):

storage/innobase/btr/btr0btr.cc
dr-m: You are right, the entire commit 3dc8ab782d7489a2e5a9f6bdc3295a3a835ed1e6 was there for testing purposes only (and successful at that; see [MDEV-32174](https://jira.mariadb.org/browse/MDEV-32174). It will be dropped in the final version.

PR #2866, opened 2023-11-23
MDEV-32371 Deadlock between buf_page_get_zip() and buf_pool_t::corrupted_evict()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32371*

## Description
`buf_page_get_zip()`: Do not wait for the page latch while holding `hash_lock`. If the latch is not available, ensure that any concurrent `buf_pool_t::corrupted_evict()` will be able to acquire the `hash_lock`, and then retry the lookup. If the page was corrupted and evicted, we will finally `goto must_read_page`, retry the read once more, and then report an error.

## How can this PR be tested?
I think that this is best tested together with [MDEV-31817](https://jira.mariadb.org/browse/MDEV-31817) #2865, which is included for the purpose of testing. The workload must use `ROW_FORMAT=COMPRESSED` tables, and we might want to use `CMAKE_BUILD_TYPE=RelWithDebInfo`.

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2235,14 +2235,21 @@ lookup:
 
       if (discard_attempted || !bpage->frame)
       {
-        /* Even when we are holding a hash_lock, it should be
-        acceptable to wait for a page S-latch here, because
-        buf_page_t::read_complete() will not wait for buf_pool.mutex,
-        and because S-latch would not conflict with a U-latch
-        that would be protecting buf_page_t::write_complete(). */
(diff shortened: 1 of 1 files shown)


dr-m:
While we do have `SET GLOBAL DEBUG_DBUG='+d,intermittent_read_failure';` it does not look like this fault injection has been implemented in `buf_page_t::read_complete()`, which we would need in order to trigger this bug.

However, if a concurrent read of the same block is in progress, the changed code path should be exercised. To test this, a tiny `innodb_buffer_pool_size` would be beneficial.

PR #2867, opened 2023-11-27
MDEV-29092 FOREIGN_KEY_CHECKS does not prevent non-copy alter from creating invalid FK structure

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-29092*

## Description
Problem:
========
- InnoDB should have two keys on the same column for the self referencing foreign key relation.

Solution:
=========
- Allow self referential foreign key relation to work with one key.
## How can this PR be tested?
- Added test case in innodb suite. So normal mysql_test_run.pl can run it

## PR quality check


mysql-test/suite/innodb/r/innodb-index-online-fk.result
```diff
@@ -654,3 +654,45 @@ t2	CREATE TABLE `t2` (
...
+t1	CREATE TABLE `t1` (
+  `f1` int(11) DEFAULT NULL,
+  KEY `f1` (`f1`),
+  CONSTRAINT `t1_ibfk_1` FOREIGN KEY (`f1`) REFERENCES `t1` (`f1`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
+DROP TABLE t1;
+CREATE TABLE t1(f1 INT, KEY(f1), KEY(f1),
+FOREIGN KEY(f1) REFERENCES t1(f1))ENGINE=InnoDB;
+Warnings:
+Note	1831	Duplicate index `f1_2`. This is deprecated and will be disallowed in a future release
```
dr-m, line 674: Let’s avoid adding the duplicated index on `f1`.

dr-m (commented):
Thank you, this is a nice and simple fix. I think that we must test this thoroughly, possibly together with #2868, on both 10.4 and 10.6.

PR #2868, opened 2023-11-27
MDEV-31441 BLOB corruption on UPDATE of PRIMARY KEY with FOREIGN KEY

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-31441*

## Description
`row_upd_clust_rec_by_insert()`: If we are resuming from a lock wait, reset the `disowned` flag of the BLOB pointers in `entry` that we copied from `rec` on which we had invoked `btr_cur_disown_inherited_fields()` before the lock wait started. In this way, the inserted record with the updated `PRIMARY KEY` value will have the BLOB ownership associated with itself, like it is supposed to be.

Note: If the lock wait had been aborted, then rollback would have invoked `btr_cur_unmark_extern_fields()` and no corruption would be possible.

## How can this PR be tested?
The test `innodb.foreign_key` was extended. It would be good to additionally test this with RQG with a broader selection of input. I do not think that it should make any difference if there is 1 or multiple BLOBs attached to the record.

The patch can be cherry-picked to 10.6 or later versions without conflicts. I think that this would be good to test in 10.6 or later as well, because the locking subsystem was refactored in 10.6.

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/foreign_key.test
+++ b/mysql-test/suite/innodb/t/foreign_key.test
@@ -732,7 +732,9 @@ SELECT a FROM t1 FORCE INDEX(a);
 # the "goto rollback_to_savept" in row_mysql_handle_errors() is reverted.
 SELECT * FROM t1;
 # Allow purge to continue by closing the read view.
-disconnect con1;
+connection con1;
+COMMIT;
+connection default;
 
 # Wait for purge. With the fix reverted, the server would crash here.
 --source include/wait_all_purged.inc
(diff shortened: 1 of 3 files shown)


storage/innobase/row/row0upd.cc
```diff
@@ -2782,6 +2801,7 @@ row_upd_clust_rec_by_insert(
 			ut_ad(!trx_undo_roll_ptr_is_insert(
 			              row_get_rec_roll_ptr(rec, index,
 							   offsets)));
+			row_upd_reown_inherited_fields(entry);
```
vlad-lesin, line 2804: The comment above says: "Skip delete marking clustered index and disowning its blobs." I would change it to "re-owning its blobs".

mysql-test/suite/innodb/t/foreign_key.test
```diff
@@ -954,7 +956,34 @@ ALTER TABLE t1 MODIFY id INT unsigned AUTO_INCREMENT;
...
+
+SET @blob = REPEAT('A', @@innodb_page_size / 2);
+INSERT INTO t1 SET pk=1, t=@blob;
+INSERT INTO t2 SET pk=1;
+--connection con1
+BEGIN;
+DELETE FROM t2;
+--connection default
+--send
+UPDATE t1 SET pk=12;
```
vlad-lesin, line 974: I would add a comment that the UPDATE is blocked on checking foreign keys reference constraints.

vlad-lesin (approved):
The fix looks good to me. I would add some comments, which would clarify the case, but the comments does not influence the code correctness, so it's up to @dr-m.

storage/innobase/row/row0upd.cc
dr-m: We will skip changing `rec`, but we must mark BLOBs in the `entry` as "owned", like they were on the previous call of `row_upd_clust_rec_by_insert()` that had returned `DB_LOCK_WAIT`. I will clarify the comment.

PR #2869, opened 2023-11-27
MDEV-32269 InnoDB after ALTER TABLE…IMPORT TABLESPACE may not be crash safe

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32269*

## Description
`mtr_t::commit()`: If `IMPORT TABLESPACE` is first-time-dirtying blocks, acquire both `log_sys.mutex` and `log_sys.flush_order_mutex` to assign a valid `m_commit_lsn` so that the block will be inserted into the correct position of `buf_pool.flush_list`.

## How can this PR be tested?
This fixes occasional debug assertion failures when running the regression test suite.

## PR quality check


vlad-lesin (approved):
The fix looks good to me.

PR #2880, opened 2023-11-29
mallinfo2: include malloc header even if mallinfo undetected

grooverdan:

## Description

It may be the case that for some reason, -Werror deprecated for instance, that mallinfo isn't detected. In this case the malloc.h headers won't be included which defines the mallinfo2 function and its structure.

Re-organise so that either function pulls in the header.

## How can this PR be tested?

remove from generated files

```
include/config.h:#define HAVE_MALLINFO 1
include/my_config.h:#define HAVE_MALLINFO 1
```

## PR quality check

### DIFF
--- a/sql/sql_test.cc
+++ b/sql/sql_test.cc
@@ -27,10 +27,12 @@
 #include "my_json_writer.h"
 #include <hash.h>
 #include <thr_alarm.h>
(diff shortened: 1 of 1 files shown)


LinuxJedi (approved):
Seems simple and sane to me.

PR #2882, opened 2023-11-29
MDEV-10653: SHOW SLAVE STATUS Can Deadlock an Errored Slave

bnestere:
A replication parallel worker thread can deadlock with another
connection running SHOW SLAVE STATUS. That is, if the replication
worker thread is in do_gco_wait() and is killed, it will already
hold the LOCK_parallel_entry, and during error reporting, try to
grab the err_lock. SHOW SLAVE STATUS, however, grabs these locks in
reverse order. It will initially grab the err_lock, and then try to
grab LOCK_parallel_entry. This leads to a deadlock when both threads
have grabbed their first lock without the second.

This patch implements the MDEV-31894 proposed fix to optimize the
workers_idle() check to compare the last in-use relay log’s
queued_count==dequeued_count for idleness. This removes the need for
workers_idle() to grab LOCK_parallel_entry, as these values are
atomically updated.

This PR is organized into two commits:
1. A regression test to show the deadlock
2. The patch which fixes the test

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
+++ b/mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
@@ -1,0 +1,121 @@
+#
+#   Verify that SHOW SLAVE STATUS will not cause deadlocks on the replica.
+# A deadlock has been seen in do_gco_wait if the thread is killed, as it will
+# hold the LOCK_parallel_entry, and during error reporting, try to grab the
+# err_lock. Prior to MDEV-10653, SHOW SLAVE STATUS would grab these locks in
+# the reverse order, as calling workers_idle() used to grab LOCK_parallel_entry
+# with the err_lock already grabbed (though the MDEV-10653 patch changed the
+# workles_idle() implementation to remove the need for locking the
+# parallel_entry).
+#
+# References:
+#   MDEV-10653: SHOW SLAVE STATUS Can Deadlock an Errored Slave
+#
+
+--source include/master-slave.inc
+--source include/have_innodb.inc
+--source include/have_debug.inc
+--source include/have_binlog_format_row.inc
+
+--echo #
+--echo # Initialize test data
+--connection master
+create table t1 (a int) engine=innodb;
+insert into t1 values (1);
+--source include/save_master_gtid.inc
+
+--connection slave
+--source include/sync_with_master_gtid.inc
+--source include/stop_slave.inc
+
+call mtr.add_suppression("Connection was killed");
+call mtr.add_suppression("Commit failed due to failure of an earlier commit on which this one depends");
+
+set @save_parallel_threads= @@global.slave_parallel_threads;
+set @save_parallel_mode= @@global.slave_parallel_mode;
+set @save_transaction_retries= @@global.slave_transaction_retries;
+set @save_innodb_lock_wait_timeout= @@global.innodb_lock_wait_timeout;
+
+set @@global.slave_parallel_threads= 2;
+set @@global.slave_parallel_mode= CONSERVATIVE;
+set @@global.slave_transaction_retries= 0;
+set @@global.innodb_lock_wait_timeout= 10;
+
+--echo # Grabbing lock on innodb row to force future replication transaction to wait (and eventually timeout)
+BEGIN;
+select * from t1 where a=1 for update;
+
+--connection master
+
+set @old_dbug= @@session.debug_dbug;
+set @@session.debug_dbug="+d,binlog_force_commit_id";
+
+
+# GCO 1
+SET @commit_id= 10000;
+# T1
+update t1 set a=2 where a=1;
+
+# GCO 2
+SET @commit_id= 10001;
+# T2
+insert into t1 values (3);
+
+set @@session.debug_dbug= @old_dbug;
+
+--connection slave
+start slave;
+
+--echo # Waiting for first transaction to start (and be held at innodb row lock)..
+--let $wait_condition= SELECT count(*)=1 FROM information_schema.processlist WHERE state LIKE 'Update_rows_log_event::find_row(%)' and  command LIKE 'Slave_worker';
+--source include/wait_condition.inc
+
+--echo # Waiting for next transaction to start and hold at do_gco_wait()..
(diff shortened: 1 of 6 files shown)


mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
```diff
@@ -0,0 +1,123 @@
...
+# T2
+insert into t1 values (3);
+
+set @@session.debug_dbug= @old_dbug;
+
+--connection slave
+start slave;
+
+--echo # Waiting for first transaction to start (and be held at innodb row lock)..
+--let $wait_condition= SELECT count(*)=1 FROM information_schema.processlist WHERE state LIKE 'Update_rows_log_event::find_row(-1)' and  command LIKE 'Slave_worker';
```
bnestere, line 70: _Noting that this should be changed to a wildcard, per @knielsen 's MDEV-32168 fixup_

mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
```diff
@@ -0,0 +1,123 @@
...
+
+--let $t2_tid= `SELECT ID FROM INFORMATION_SCHEMA.PROCESSLIST WHERE STATE LIKE 'Waiting for prior transaction to start commit%'`
+--replace_result $t2_tid "<TID of worker in do_gco_wait>"
+--eval kill $t2_tid
+--let $wait_condition= SELECT count(*)=1 FROM information_schema.processlist WHERE command LIKE 'Killed';
+--source include/wait_condition.inc
+
+set debug_sync="now signal sss_continue";
+--let $wait_condition= SELECT count(*)=0 FROM information_schema.processlist WHERE state LIKE 'debug sync point%' and  command LIKE 'Slave_worker';
+--source include/wait_condition.inc
```
knielsen, line 92: I don't think you need this wait?
The --reap will wait for SHOW SLAVE STATUS to complete, which includes waking up from debug_sync.

mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
```diff
@@ -0,0 +1,123 @@
...
+
+set @save_parallel_threads= @@global.slave_parallel_threads;
+set @save_parallel_mode= @@global.slave_parallel_mode;
+set @save_transaction_retries= @@global.slave_transaction_retries;
+set @save_innodb_lock_wait_timeout= @@global.innodb_lock_wait_timeout;
+
+set @@global.slave_parallel_threads= 2;
+set @@global.slave_parallel_mode= CONSERVATIVE;
+set @@global.slave_transaction_retries= 0;
+set @@global.innodb_lock_wait_timeout= 2;
```
knielsen, line 42: I worry that 2 seconds could be too little in extreme cases of high load during the test.
I'd set it a bit higher (say 10 seconds) and just release the row lock later so the timeout doesn't happen.

mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
```diff
@@ -0,0 +1,123 @@
...
+
+set debug_sync="now signal sss_continue";
+--let $wait_condition= SELECT count(*)=0 FROM information_schema.processlist WHERE state LIKE 'debug sync point%' and  command LIKE 'Slave_worker';
+--source include/wait_condition.inc
+
+--connection slave1
+--echo # Waiting for SHOW SLAVE STATUS to complete..
+# Remove SHOW SLAVE STATUS output
+--replace_regex /.*//
+--reap
```
knielsen, line 98: Simpler to just use --disable_result_log / --enable_result_log around the query (not sure if you need it around the --send or --reap or both, you can TIAS).

mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
```diff
@@ -0,0 +1,123 @@
...
+--echo # ..done
+
+--connection slave
+--let $slave_sql_errno= 1927
+--source include/wait_for_slave_sql_error.inc
+ROLLBACK;
```
knielsen, line 104: Why not put the ROLLBACK before the wait_for_slave_sql_error.inc?
You already checked that the other worker is in killed state, so releasing the row lock here should avoid the test needing to wait for the row lock timeout.

sql/rpl_parallel.h
```diff
@@ -362,6 +362,7 @@ struct rpl_parallel {
   HASH domain_hash;
   rpl_parallel_entry *current;
   bool sql_thread_stopping;
+  Relay_log_info *rli;
```
knielsen, line 365: Don't add this member just to have rli accessible from workers_idle(). Instead just pass the rli as an argument.
As an alternative, moving the workers_idle() into the Relay_log_info class may make more sense now, as you like.

knielsen (changes requested):
Thanks, Brandon. Looks good after fixing small comments written inline.

mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
bnestere: Nice catch! Originally I had the test use `debug_sync` with the GCO2 thread as well, and realized I didn't need it, yet kept this.

mysql-test/suite/rpl/t/rpl_deadlock_show_slave_status.test
bnestere: I had tried that originally, but it didn't work :( Though I didn't try it around both, I'll give that a shot.

sql/rpl_parallel.h
bnestere: Yeah I went back and forth between these thoughts, and another alternative as a standalone/static function. I see pros/cons to all approaches, as having `rpl_parallel` be a direct member of the `Relay_log_info` feels awkward to then have to pass the `rli` _back_ to `rli.parallel.workers_idle(rli)`. I think moving the check into `Relay_log_info` makes the next most sense to me (though it feels to me as the concern of the `rpl_parallel` struct).

Though now another thought comes to mind as a static function of `rpl_parallel`, which I think I am happy with all around :)

knielsen (approved):
Thanks, Brandon!

knielsen:
Looks good now Brandon!
Will you push it?

PR #112, opened 2015-10-23
MDEV-8981: Analyze stmt - cycles can overflow

grooverdan:
A 64bit counter can overflow within the time of a query
so lets take it that the measurement is the small value
rather than an order 1e12 millisecond query.

$ ./mtr analyze_format_json
Logging: ./mtr  analyze_format_json
vardir: /home/dan/software_projects/mariadb-server/mysql-test/var
Checking leftover processes...
Removing old var directory...
Creating var directory '/home/dan/software_projects/mariadb-server/mysql-test/var'...
Checking supported features...
MariaDB Version 10.1.8-MariaDB
 - SSL connections supported
Collecting tests...
Installing system database...

==============================================================================

TEST                                      RESULT   TIME (ms) or COMMENT
--------------------------------------------------------------------------

worker[1] Using MTR_BUILD_THREAD 300, with reserved ports 16000..16019
main.analyze_format_json                 [ pass ]     79
--------------------------------------------------------------------------
The servers were restarted 0 times
Spent 0.079 of 14 seconds executing testcases

Completed: All 1 tests were successful.

PR #2884, opened 2023-11-30
MDEV-32910 cycles can overflow and result in large explain json=X mesurements

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32910*

## Description

Originally fixed with MDEV-8981 however regressed with MDEV-32910.

Its like déjà vu I'm fixing it again #112.

## How can this PR be tested?

Run enough queries and you'll see it. cycles go fast and overflow happen.

## PR quality check

### DIFF
--- a/sql/sql_analyze_stmt.h
+++ b/sql/sql_analyze_stmt.h
@@ -76,6 +76,8 @@ protected:
   {
     ulonglong end= measure();
     cycles += end - last_start;
+    if (unlikely(end < last_start))
+      cycles += ULONGLONG_MAX;
 
     process_gap_time_tracker(thd, end);
     if (my_gap_tracker)


sql/sql_analyze_stmt.h
```diff
@@ -76,6 +76,8 @@ class Exec_time_tracker
   {
     ulonglong end= measure();
     cycles += end - last_start;
+    if (unlikely(end < last_start))
+      cycles += ULONGLONG_MAX;
```
LinuxJedi, line 80: I don't fully understand the intent of this code. Can you please elaborate? This looks to me to be just `cycles--`.

In addition, I don't understand the scenario where `end < last_start` would happen that this would help. `my_timer_cycles()` on Intel I believe this should happen if the computer has been switched on > 100 years, or the execution wasn't pinned to a single CPU / core (in which case, it would be a junk measure anyway).

If it is to fix `my_timer_microseconds()`, the real fix would be to use a monotonic time source (it also wouldn't be a cycles measure, but I'm ignoring that for now :)

PR #2886, opened 2023-11-30
MDEV-7850: make Thread-id available in Gtid

bnestere:
This patch augements Gtid_log_event with the user thread-id. In particular that compensates for the loss of this info in Rows_log_events.

Gtid_log_event::thread_id gets visible in mysqlbinlog output like

  #231025 16:21:45 server id 1  end_log_pos 537 CRC32 0x1cf1d963  GTID 0-1-2 ddl thread_id=10

as 64 bit unsigned integer.

While the size of Gtid event has grown by 8-9 bytes replication from OLD <-> NEW is not affected by it.

This work was started by the late Sujatha Sivakumar. Brandon Nesterenko took it over, reviewed initial patches and extended the work.


andrelkin (approved):
Looks good. Thanks for your hefty part in the whole work, Brandon!

PR #2887, opened 2023-11-30
MDEV-27806 GTIDs diverge in Galera cluster after CTAS

sciascid:
Add OPTION_GTID_BEGIN to applying side thd. This is needed to avoid intermediate commits when CREATE TABLE AS SELECT is applied, causing one more GTID to be consumed with respect to executing node.


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/61daac54d62491fac9b699bccd639494da38a72f

PR #2890, opened 2023-12-01
MDEV-32923: drop errmsg-utf8.txt from packaging

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32923*

## Description

This file isn't used by client or server. Its
functionality is purely in the generation of *.sys files.

As such omit from packaging and save ~1MB.

## How can this PR be tested?

check it doesn't appear in packaging.

## PR quality check


illuusio:
How about `extra/comp_err.c` didn't dig too far where it's used but it's hard coded to have `errmsg-utf8.txt` or `sql/share/insert_translations_into_errmsg.py`. I didn't really get why it's mentioned here: `storage/innobase/handler/ha_innodb.cc`

grooverdan:
comp_err is used in buildtime only to generate the sys files from errmsg-utf8.txt

innodb ha_* references in comments are about looking up the value, technically the value is in the sys files as managed by a higher level function however for developer convenience the errmsg-utf8.txt is the place to lookup/edit errors.

LinuxJedi (approved):
Agreed, I cannot see a good reason to include this in anything but source packages.

PR #2896, opened 2023-12-05
MDEV-32936: Detect do ARM processor have NEON intrinsics or not

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-32936*

## Description
This is mainly to have support for [Debian armel toolchain](https://wiki.debian.org/ArmEabiPort). This toolchain is for mainly for ARMv5/6 with no had float support. It does not have NEON support neither but it can be emulated with `-mfpu=neon -mfloat-abi=softfp'.

This PR overcomes problem:
```
/usr/lib/gcc/arm-linux-gnueabi/13/include/arm_neon.h:31:2: error: #error "NEON intrinsics not available with the soft-float ABI.  Please use -mfloat-abi=softfp or -mfloat-abi=hard"
```
by adding those flags that are needed to emulate NEON and hardware floating point. They are only applied to specific file which is compiled with emulation on. All the rest does not have new flags.

This PR is not needed if MariaDB is build as 32-bits version on ARM (which most of the CPU:s should be)

As I know this is little bit niche group of users whom wants to use MariaDB for example in Raspberry Pi 1 or Raspberry pi Zero (W) but having MariaDB client in those systems won't hurt anyone.

## How can this PR be tested?
PR is only tested with ARM Linux distros on ARM 64-bit machine (docker image is the same that is used in Salsa CI and has correct armel:

```
docker run -it --rm registry.salsa.debian.org/salsa-ci-team/pipeline/arm32v5/base:sid bash
apt update
apt install git-buildpackage
... fetch git repo and get correct branch ...
cd server
mk-build-deps -r -i debian/control -t "apt-get -y -o Debug::pkgProblemResolver=yes --no-install-recommends"
mkdir build
cd build
cmake ..
cd mysys
make
```
Afterwards everything else could be build also but if one just wants to check if this is working then `mysys` dir is enough.

## PR quality check

### DIFF
--- a/mysys/CMakeLists.txt
+++ b/mysys/CMakeLists.txt
@@ -89,6 +89,11 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   IF(CMAKE_COMPILER_IS_GNUCC)
     include(CheckCXXSourceCompiles)
 
+    # Check if NEON intrinsics are available and does not
+    # print error 'NEON intrinsics not available with the soft-float ABI.
+    # Please use -mfloat-abi=softfp or -mfloat-abi=hard'
+    CHECK_INCLUDE_FILES(arm_neon.h HAVE_ARMV8_NEON)
+
     CHECK_CXX_SOURCE_COMPILES("
     #define CRC32CX(crc, value) __asm__(\"crc32cx %w[c], %w[c], %x[v]\":[c]\"+r\"(crc):[v]\"r\"(value))
     asm(\".arch_extension crc\");
@@ -135,9 +140,21 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
       IF(HAVE_ARMV8_CRYPTO)
         ADD_DEFINITIONS(-DHAVE_ARMV8_CRYPTO)
       ENDIF()
+      IF(NOT HAVE_ARMV8_NEON)
+        set(CMAKE_REQUIRED_FLAGS "-mfpu=neon -mfloat-abi=softfp")
+        CHECK_INCLUDE_FILES(arm_neon.h HAVE_ARMV8_NEON_SOFTFP)
+        set(CMAKE_REQUIRED_FLAGS "")
+      ENDIF()
       SET(MYSYS_SOURCES ${MYSYS_SOURCES} crc32/crc32_arm64.c)
+
+      IF(HAVE_ARMV8_NEON_SOFTFP)
+        SET(CRC32_ARM64_FLAGS "-mfpu=neon -mfloat-abi=softfp -march=armv8-a+crc+crypto")
+      ELSE()
+        SET(CRC32_ARM64_FLAGS "-march=armv8-a+crc+crypto")
+      ENDIF()
+
       SET_SOURCE_FILES_PROPERTIES(crc32/crc32_arm64.c PROPERTIES
-        COMPILE_FLAGS "-march=armv8-a+crc+crypto")
+        COMPILE_FLAGS ${CRC32_ARM64_FLAGS})
     ENDIF()
   ENDIF()
 ENDIF()


mysys/CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   IF(CMAKE_COMPILER_IS_GNUCC)
     include(CheckCXXSourceCompiles)
 
+    CHECK_CXX_SOURCE_COMPILES("
+    // print error 'NEON intrinsics not available with the soft-float ABI.
+    // Please use -mfloat-abi=softfp or -mfloat-abi=hard'
+    #include <arm_neon.h>
+    int main(){}" HAVE_ARMV8_NEON)
```
dr-m, line 99: As far as I understand, the intention of this pull request is to fix builds that target ARMv5 or ARMv6. On my Raspberry Pi 2, `${CMAKE_SYSTEM_PROCESSOR}` evaluates to `armv7l`. I would assume it to be something similar on the targeted environments. It definitely should not reach this branch of the CMake code. We currently do not seem to have any CMake code that is specific to 32-bit ARM architectures.

I see that NEON is optional in ARMv8-R. But, which compilation problem on ARMv8 a.k.a. Aarch64 is this solving? I would guess that an ARMv8 build environment would have this header file available, and we’d have to perform a run-time check, in case the code is executing on a processor where NEON is not available. Only the file `mysys/crc32/crc32_arm64.c` (for ARMv8, not 32-bit ARMs) should depend on this.

dr-m (changes requested):
I think that it would be helpful to see some build logs where something is failing.

ottok:
Pipeline https://salsa.debian.org/mariadb-team/mariadb-server/-/pipelines/611348 with this patch passed on all ARM versions!

mysys/CMakeLists.txt
grooverdan: `HWCAP_NEON` (bit 12) is defined in the 32bit userspace, but in arm64 bit 12 is reused for a different capability.

Also these bits mean different things on 32bit userspace:
```
  #ifndef HWCAP_CRC32
  # define HWCAP_CRC32 (1 << 7)
  #endif
  
  #ifndef HWCAP_PMULL
  # define HWCAP_PMULL (1 << 4)
  #endif
```

So as the build system is building crc32/crc32_arm64.c on 32 bit, its going to be broken.

Until Debian start doing a cross-compulation (documented - https://wiki.debian.org/CrossCompiling and [tested](http://crossqa.debian.net/build/mariadb_1:10.11.2-3_armhf_20230423012238.log)) rather than salsa hacky, the packages generated, and the tests in the same hacked environment, aren't worth the BTUs of power that made them. Setting an env  BUILD_ARCH: 'armhf' ([ref](https://salsa.debian.org/salsa-ci-team/pipeline/raw/master/salsa-ci.yml)) is insufficient.

Setting env HOST_ARCH=armhf (or armel) appears to be the necessary bit to trigger [cross compulation](https://salsa.debian.org/salsa-ci-team/pipeline/raw/master/salsa-ci.yml).

illuusio: As this is mainly soft-float problem which NEON solves.  I didn't know that event ARMv8 NEON can be optional I though it's mandatory. I've try to follow that cross compile to produce same binaries.

mysys/CMakeLists.txt
```diff
@@ -138,9 +144,23 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
       IF(HAVE_ARMV8_CRYPTO)
         ADD_DEFINITIONS(-DHAVE_ARMV8_CRYPTO)
       ENDIF()
+      IF(NOT HAVE_ARMV8_NEON)
+        set(CMAKE_REQUIRED_FLAGS "-mfpu=neon -mfloat-abi=softfp")
+
+        CHECK_CXX_SOURCE_COMPILES("
+        #include <arm_neon.h>
+        int main(){}" HAVE_ARMV8_NEON_SOFTFP)
```
grooverdan, line 152: `SET(CMAKE_REQUIRED_FLAGS)` required here now that the test is finished.

mysys/CMakeLists.txt
grooverdan: This branch of the CMakeLists.txt is under a `CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64"` condition, it shouldn't be reached on armv[67] as all armv[67] are 32 bit. Otto's pipeline shows crc32_arm64.cc compiled on armhf and armel and even has `SYSTEM processor: aarch64` in the libmariadb side. This is clearly wrong. The tests passing on a 64bit hardware with a userspace don't prove its right.

So problem 1 - ensure 32 bit builds don't hit this branch.

on https://buildd.debian.org/status/package.php?p=mariadb - it appears that the real build doesn't compile this, so the salsa build is different from the real. So small relief - though ' SYSTEM processor: armv8l' is in both the armhf and armel output.

Problem 2 -  ARMv8 without neon.

The `arm_neon.h` header in the mysys/crc32_arm64.c seems there to purely define `vmull_p64` as `return (poly128_t) __builtin_arm_crypto_vmullp64 ((uint64_t) __a, (uint64_t) __b);` this appears to be a [pmull](https://developer.arm.com/documentation/ddi0597/2023-12/SIMD-FP-Instructions/VMULL--integer-and-polynomial---Vector-Multiply-Long--integer-and-polynomial--?lang=en) feature. This is already checked.

There is no neon features used directly.

option 1 - remove the header and use the `__builtin_arm_crypto_vmullp64` directly.

option 2 - i think is what you've done, try to make it compile with the header. regardless of if the neon features are used.

I tried https://salsa.debian.org/grooverdan/mariadb-server/-/pipelines/623885 yesterday.


So for solving Problem 2, I'm ok with this patch, with the one reset of `SET(CMAKE_REQUIRED_FLAGS)` below, and a much more focused commit message on the problem 2 thing that its solving, including why we even need to play with NEON at all. A comment in the CMakeList.txt around this would also be good.

grooverdan: branch back port also, its still a bug fix.

illuusio: What would be lowest branch to use? 10.11 or 10.4?

mysys/CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   IF(CMAKE_COMPILER_IS_GNUCC)
     include(CheckCXXSourceCompiles)
 
+    CHECK_CXX_SOURCE_COMPILES("
```
vaintroub, line 95: Please use CHECK_INCLUDE_FILES https://cmake.org/cmake/help/latest/module/CheckIncludeFiles.html

grooverdan: Its checking for a compile error in the header. `check_symbol_exists(__ARM_FP arm_neon.h HAVE_ARMV8_NEON_SOFTFP)` maybe.

mysys/CMakeLists.txt
grooverdan: 10.5 where mysys/crc32/crc32_arm64.c was added.

mysys/CMakeLists.txt
vaintroub: CHECK_INCLUDE_FILES checks for compile error, same thing

mysys/CMakeLists.txt
illuusio: DO you mean it should be turned off?

mysys/CMakeLists.txt
```diff
@@ -89,6 +89,11 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   IF(CMAKE_COMPILER_IS_GNUCC)
     include(CheckCXXSourceCompiles)
 
+    # Check if NEON intrinsics are available and does not
+    # print error 'NEON intrinsics not available with the soft-float ABI.
+    # Please use -mfloat-abi=softfp or -mfloat-abi=hard'
+    CHECK_INCLUDE_FILES(arm_neon.h HAVE_ARMV8_NEON)
```
FooBarrior, line 95: What happens if the builder has Neon and and the Debian machine wil not have it? Is it a valid configuration for Debian? Can't Neon extension be softly turned off on the processor?

illuusio: Actually very good question. I assume that if it's build with NEON code and run in machine that does not have it it will crash hardly (or GCC can handle it). As far as I know I don't you can't turn NEON on from CPU it's kind you turn off math unit in x86. If you compile your own kernel you can opt-it out.
If you are talking about Debian `armel` then it should work if machine has or not as it's emulated. But if you have ARMv8-a and run it in ARMV8-r without NEON then as said you are probably out of luck.

mysys/CMakeLists.txt
illuusio: Now it use that approach and it looks lot cleaner. Although people seems to hate  `CHECK_INCLUDE_FILES

mysys/CMakeLists.txt
grooverdan: Yes. I did. As you made it. Good.

mysys/CMakeLists.txt
illuusio: Now rebased

mysys/CMakeLists.txt
illuusio: I've tested this on Debian Salsa also and seems to work as expected. With or without armel

vaintroub: I do not know why people would hate `CHECK_INCLUDE_FILES`. Rather, a raw `CHECK_SOURCE_COMPILES` with an empty  main() usually a sign of misunderstanding. While it might be used for either a check that is too complex, or a test for compiler bug, but it is very seldom. Much more often, compact, and task-specific CHECK_INCLUDE_FILES, CHECK_FUNCTION_EXISTS, CHECK_SYMBOL_EXISTS would do the job.

ottok:
From: https://salsa.debian.org/mariadb-team/mariadb-server/-/merge_requests/32
 > Confirming once again that the ARM build works with Tuukka's patch: https://salsa.debian.org/mariadb-team/mariadb-server/-/pipelines/671545
> ..and fails without it: https://salsa.debian.org/mariadb-team/mariadb-server/-/pipelines/672006

grooverdan:
Don't see a compilation error. Looks like it just got killed off.
Still see "-- SYSTEM processor: aarch64"

From !73 - https://salsa.debian.org/otto/mariadb-server/-/jobs/5665474/raw cross build of armel and armhf worked.

ottok:
> Don't see a compilation error. Looks like it just got killed off. Still see "-- SYSTEM processor: aarch64"
> 
> From !73 - https://salsa.debian.org/otto/mariadb-server/-/jobs/5665474/raw cross build of armel and armhf worked.

This PR#2896 is *not* being used in [!73](https://salsa.debian.org/mariadb-team/mariadb-server/-/merge_requests/73) for cross-compilation, that is a separate thing to related to this PR.

grooverdan:
> > Don't see a compilation error. Looks like it just got killed off. Still see "-- SYSTEM processor: aarch64"
> > From !73 - https://salsa.debian.org/otto/mariadb-server/-/jobs/5665474/raw cross build of armel and armhf worked.
> 
> This PR#2896 is _not_ being used in [!73](https://salsa.debian.org/mariadb-team/mariadb-server/-/merge_requests/73) for cross-compilation, that is a separate thing to related to this PR.

Yes, which is why I think a) not needed to solve armhf/el. b) the chroot 32 bit arm builds are worth the BTUs it generates.

illuusio:
Do we need this is if Debian uses correct 32 bit system then this is not needed..

grooverdan:
per MDEV - not an arm32 fix. Might have uses for Arm8R if anyone reports compile troubles there.

PR #2900, opened 2023-12-05
MDEV-32953: main.rpl_mysqldump_slave Fails with "Master binlog wasn’t…

bnestere:
… deleted" Assertion

Because mariadb-dump --delete-master-logs immediately purges logs after flushing, it is possible the binlog dump thread would still be using the old log when the purge executes, disallowing the file from being deleted.

This patch institutes a work-around in the test as follows:
 1) temporarily stop the slave so there is no chance the old binlog is still being referenced.
 2) set master_use_gtid=Slave_pos so the slave can still appear up-to-date on the master after the master flushes/purges its logs (while the slave is offline). Otherwise (i.e. if using binlog file/pos), the slave would point to a purged log file, and receive an error immediately upon connecting to the master.

Outside of this test work-around, I'm also trying to decide if the underlying issue is a bug or works as expected.. it seems inconvenient for users to have to implement such workarounds in their routines.

### DIFF
--- a/mysql-test/main/rpl_mysqldump_slave.result
+++ b/mysql-test/main/rpl_mysqldump_slave.result
@@ -72,6 +72,13 @@ id int
 insert into t values (1);
 insert into t values (2);
 drop table t;
+connection slave;
+include/stop_slave.inc
+change master to master_use_gtid=slave_pos;
+connection master;
+# Ensuring the binlog dump thread is killed on primary...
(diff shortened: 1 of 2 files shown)


andrelkin (approved):
Looks good.

knielsen (commented):
Agree, looks good.
(I actually have the similar patch pending in a local tree to fix MDEV-29816).

PR #2901, opened 2023-12-06
MDEV-32919: Cannot select particular field from IS.tables in case table needs upgrade from MySQL 5.7

an3l:
- When selecting * fields from IS.tables, open method using full table scan is called `fill_schema_table_by_open()`.
- When select specific field from IS.tables, open method from frm is used `fill_schema_table_from_frm()`.
- Call `get_schema_tables_record` for `MYSQL_JSON` data type in case plugin is installed and table needs upgrade.

## PR quality check


vuvova (changes requested):
No, let's not have a special case for MYSQL_JSON data type, errors must be handled uniformly on all code paths otherwise we'll have the same bug again sooner or later.

PR #2902, opened 2023-12-06
Update Amazon S3 test for URL endpoints

LinuxJedi:
If you are using a region other than us-east-1, the region needs to be part of the domain when accessing S3. This pathc updates the Amazon test so that any Amazon S3 domain is detected as connecting to Amazon S3.

## How can this PR be tested?

This can be tested by running the Aria S3 engine against Amazon using the region as part of the domain. For example `s3-us-east-2.amazonaws.com` or `s3.us-east-2.amazonaws.com`.

## PR quality check


montywi (approved):
Patch ok, but you have a type on the commit comment:
pathc -> patch

PR #2904, opened 2023-12-06
MDEV-28619 Server crash in Window_funcs_sort::setup

DaveGosselin-MariaDB:
Fixes a crash where a query referencing an undefined WINDOW is nested within the IN clause of another query.  Such queries may have the form:
  SELECT ... IN (SELECT ... UNION SELECT ... ORDER BY LAST_VALUE (...) OVER w);
Further examples are in the tests associated with this commit.  The error seen now with this fix is consistent with MySQL 8.2 though not identically worded; MariaDB renders the error as "Window specification with name 'X' is not defined" while MySQL 8.2 renders it as "Window name 'x' is not defined").

Item_window_func::resolve_window_name is idempotent with respect to resolving a window name to a window specification: if the name has been resolved to the specification, then it does not attempt to re-resolve; otherwise, it attempts to resolve the name and if it doesn't resolve it, then it will return a failure.

setup_windows now leverages resolve_window_name to ensure that a window function has a corresponding window spec.  Typically that check would've been completed before this point in execution by setup_order (which eventually invokes fix_fields), but in this case we have an instance of Item_in_subselect which relies on running the nested query back through the execution engine. That nested executor doesn't reparse the query and instead expects that all references have been resolved.

### DIFF
--- a/mysql-test/main/win.result
+++ b/mysql-test/main/win.result
@@ -4390,3 +4390,23 @@ row_number() OVER (order by a)
 2
 3
 drop table t1;
+#
+# MDEV-28619: Server crash in Window_funcs_sort::setup
+#
+SELECT 1 IN (SELECT 1 a UNION SELECT 5 ORDER BY LAST_VALUE (a) OVER XXX);
+ERROR HY000: Window specification with name 'XXX' is not defined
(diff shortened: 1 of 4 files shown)


DaveGosselin-MariaDB:
Closing this PR per a conversation between myself, Sanja, and Igor.

PR #2906, opened 2023-12-07
MDEV-32884 Improve S3 options comaptibility

LinuxJedi:
The previous commit for MDEV-32884 fixed the s3_protocol_version option, which was previous only using "Auto", no matter what it was set to. This patch does several things to keep the old behaviour whilst correcting for new behaviour and laying the groundwork for the future. This includes:

* `Original` now means v2 protocol, which it would have been due to the option not working, so upgrades will stil work.
* A new `Legacy` option has been added to mean v1 protocol.
* Options `Path` and `Domain` have been added, these will be the only two options apart from `Auto` in a future release, and are more aligned with what this variable means.
* Fixed the s3.debug test so that it works with v2 protocol.
* Fixed the s3.amazon test so that it works with region subdomains.
* Added additional modes to the s3.amazon test.
* Added s3.not_amazon test for the remaining modes.

This replaces PR #2902.

- [x] *The Jira issue number for this PR is: MDEV-32884*

## How can this PR be tested?

This has been tested against two regions of AWS and MinIO. Both will need to be tried for all tests to run.

## PR quality check


LinuxJedi:
This has been reviewed and approved by @montywi outside of GitHub.

PR #2907, opened 2023-12-07
MDEV-32738: Don't roll back high-prio txn waiting on a lock in InnoDB

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-32738*

## Description
DML transactions on FK-child tables also get table locks on FK-parent tables.
If there is a DML transaction holding such a lock, and a TOI transaction starts, the latter BF-aborts the former and puts itself into a waiting state.
If at this moment another DML transaction on FK-child table starts, it doesn't check that the transaction waiting on a parent table lock is TOI, and it erroneously BF-aborts the waiting TOI transaction.

The fix: don't roll back high-priority transaction waiting on a lock in InnoDB, instead roll back an incoming DML transaction.

## How can this PR be tested?

MTR test is provided.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/MDEV-32738.result
+++ b/mysql-test/suite/galera/r/MDEV-32738.result
@@ -1,0 +1,32 @@
+connection node_2;
+connection node_1;
+connect con1_1,127.0.0.1,root,,test,$NODE_MYPORT_1;
+connect con1_2,127.0.0.1,root,,test,$NODE_MYPORT_1;
+CREATE TABLE t1(c1 INT PRIMARY KEY) ENGINE=InnoDB;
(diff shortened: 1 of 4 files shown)


dr-m:
Can you please post a separate pull request for 10.6? The lock creation and wait logic is different there.

denis-protivensky:
Okay, I will work on a separate fix for 10.6.

denis-protivensky:
After porting the test case with little changes, 10.6 experiences a hang but for a different reason. I will open a follow-up ticket to fix 10.6 separately.

sysprg:
Thanks, the fix merged with the head revision: https://github.com/MariaDB/server/commit/b7718a1c1cf40f4b15263c27861c3f712cf8a62a

PR #2908, opened 2023-12-07
MDEV-32964: Expect DB_INTERRUPTED from wsrep_row_upd_check_foreign_constraints

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-32964*

## Description
DB_INTERRUPTED should be a valid return code expected from wsrep_row_upd_check_foreign_constraints call.

## How can this PR be tested?

Manually tested with pstress to ensure the issue doesn't reproduce anymore.
No MTR is provided as it's pretty hard to build such a combination of load that would make the issue surface with the given return code.

## PR quality check

### DIFF
--- a/storage/innobase/row/row0upd.cc
+++ b/storage/innobase/row/row0upd.cc
@@ -2486,18 +2486,18 @@ row_upd_sec_index_entry(
 				case DB_NO_REFERENCED_ROW:
 					err = DB_SUCCESS;
 					break;
-				case DB_LOCK_WAIT:
+				case DB_INTERRUPTED:
 				case DB_DEADLOCK:
 				case DB_LOCK_WAIT_TIMEOUT:
 					WSREP_DEBUG("Foreign key check fail: "
 						"%s on table %s index %s query %s",
-						ut_strerr(err), index->name(), index->table->name.m_name,
+						ut_strerr(err), index->table->name.m_name, index->name(),
 						wsrep_thd_query(trx->mysql_thd));
 					break;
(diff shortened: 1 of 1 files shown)


storage/innobase/row/row0upd.cc
```diff
@@ -2506,7 +2507,7 @@ row_upd_sec_index_entry(
 		}
 
 #ifdef WITH_WSREP
-		ut_ad(err == DB_SUCCESS || err == DB_LOCK_WAIT
+		ut_ad(err == DB_SUCCESS || err == DB_LOCK_WAIT || err == DB_INTERRUPTED
 		      || err == DB_DEADLOCK || err == DB_LOCK_WAIT_TIMEOUT);
```
dr-m, line 2511: I do not think that `DB_LOCK_WAIT` is possible here. It should have been resolved to one of the other errors by this point.

denis-protivensky: I see, okay. Let me work on this issue more.

dr-m (changes requested):
I would really like to see a test case for covering this functionality. cd79f102110a3543bc78ef4bec4dbeadaf1f1a83 added a test case that is exercising this area. You would only need to ensure a lock conflict on a secondary index, and add a `KILL` statement to the test.

sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/ddd8a90812e8ab7675b989696023e0a9966b6409

PR #2909, opened 2023-12-07
Add not_amazon tests to S3 engine

LinuxJedi:
Additional tests missed of an earlier pull request.

## How can this PR be tested?

Run the S3 engine tests against MinIO.

## PR quality check

### DIFF
--- a/mysql-test/suite/s3/not_amazon.test
+++ b/mysql-test/suite/s3/not_amazon.test
@@ -1,8 +1,8 @@
 --source include/have_s3.inc
 
-if (`SELECT @@s3_host_name NOT LIKE "%.amazonaws.%"`)
+if (`SELECT @@s3_host_name LIKE "%.amazonaws.%"`)
 {
-  skip Not connected to AWS;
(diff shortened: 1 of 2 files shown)


mysql-test/suite/s3/not_amazon.test
```diff
@@ -0,0 +1,35 @@
+--source include/have_s3.inc
+
+if (`SELECT @@s3_host_name LIKE "%.amazonaws.com"`)
```
dlenski, line 3: Not all Amazon S3 endpoints’ hostnames match `%.amazonaws.com`, in particular those in the AWS China regions. See https://docs.aws.amazon.com/general/latest/gr/s3.html for examples.

LinuxJedi: I was not aware of that one. So. "%.amazonaws.com%" would match?

dlenski: That seems like it would _currently_ be sufficient, yes!

PR #2911, opened 2023-12-08
MDEV-32973: SHOW TABLES LIKE shows temporary tables with non-matching names

an3l:

## PR quality check

### DIFF
--- a/mysql-test/main/information_schema_temp_table.test
+++ b/mysql-test/main/information_schema_temp_table.test
@@ -278,3 +278,45 @@ drop table seq2;
 show  full tables;
 drop table seq3;
 show  full tables;
+
+--echo #
+--echo # MDEV-32973: SHOW TABLES LIKE shows temporary tables with
+--echo #             non-matching names
+--echo #
+
+CREATE DATABASE db;
+USE db;
+CREATE TABLE realtable (a INT);
+CREATE TEMPORARY TABLE temptable (a INT);
+SHOW TABLES LIKE 'faketable';
+SHOW TABLES LIKE 'temptable';
+SHOW TABLES LIKE '%pt%';
+SHOW TABLES LIKE '%lt%';
+select table_name from information_schema.tables where
+  table_schema='db' and table_name='temptable';
+select table_name from information_schema.tables where
+  table_schema='db' and table_name='faketable';
+drop database db;
+
+create database `t\_p`;
+create temporary table `t\_p`.tmp (a int);
+select table_schema, table_name from information_schema.tables where table_schema='t\\_p';
+drop database `t\_p`;
+
+create database db;
+use db;
+create temporary table `t\_p` (a int);
+select table_schema, table_name from information_schema.tables where table_schema='db' and table_name='t\\_p';
+select table_schema, table_name from information_schema.tables where table_name='t\\_p';
+drop table `t\_p`;
+
+-- source include/not_windows.inc
+CREATE TEMPORARY TABLE t2 (a INT);
+select table_name from information_schema.tables where
+  table_schema='db' and table_name='T2';
+SHOW TABLES LIKE 'T%';
+drop database db;
+
+--echo #
+--echo # End of 11.2 tests
+--echo #
--- a/mysql-test/main/information_schema_temp_table_lowercase.test
+++ b/mysql-test/main/information_schema_temp_table_lowercase.test
@@ -1,0 +1,27 @@
+--echo #
+--echo # MDEV-32973: SHOW TABLES LIKE shows temporary tables with
+--echo #             non-matching names
+--echo #
+
+# Compare table names in a case-insensitive manner
+create temporary table t1 (a int);
+select table_name from information_schema.tables where
+  table_schema='test' and table_name='T1';
+SHOW TABLES LIKE 'T%';
+
+create database db;
+use db;
+create temporary table `t\_p` (a int);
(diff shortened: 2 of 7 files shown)


mysql-test/main/information_schema_temp_table.test
```diff
@@ -278,3 +278,19 @@ drop table seq2;
...
+--echo #
+--echo # MDEV-32973: SHOW TABLES LIKE shows temporary tables with
+--echo #             non-matching names
+--echo #
+
+CREATE DATABASE db;
+USE db;
+CREATE TABLE realtable (a INT);
+CREATE TEMPORARY TABLE temptable (a INT);
+SHOW TABLES LIKE 'faketable';
```
vuvova, line 291: add:
```sql
SHOW TABLES LIKE '%pt%';
```
and also you can try with `--lower-case-table-names=1` (in a separate test file)
```sql
create table t1 (a int);
select table_name from information_schema.tables where table_schema='test' and table_name='T1'
```

vuvova (changes requested):

mysql-test/main/information_schema_temp_table_lowercase.test
```diff
@@ -0,0 +1,14 @@
+--echo #
+--echo # MDEV-32973: SHOW TABLES LIKE shows temporary tables with
+--echo #             non-matching names
+--echo #
+
+# Compare table names in a case-insensitive manner
+create temporary table t1 (a int);
+select table_name from information_schema.tables where
+  table_schema='test' and table_name='T1';
```
vuvova, line 9: eh, really?
```
MariaDB [test]> create temporary table t1 (a int);
Query OK, 0 rows affected (0.002 sec)

MariaDB [test]> show create table T1;
ERROR 1146 (42S02): Table 'test.T1' doesn't exist
```
Why do you think table names should be compared case-insensitively?
See https://mariadb.com/kb/en/identifier-case-sensitivity/

vuvova (changes requested):

sql/sql_show.cc
```diff
@@ -5358,9 +5358,20 @@ int get_all_tables(THD *thd, TABLE_LIST *tables, COND *cond)
       }
 
       TABLE *tmp_tbl= share_temp->all_tmp_tables.front();
+      const char *lookup_table= plan->lookup_field_vals.table_value.str;
+      LEX_CSTRING *table_name= &tmp_tbl->s->table_name;
+      if (lookup_table)
+      {
+        if (lower_case_table_names)
+        {
+          if (wild_case_compare(system_charset_info, table_name->str, lookup_table))
```
vuvova, line 5367: what will the following test case show?
```sql
create temporary table tmp (a int);
create temporary table t_p (a int);
select table_name from information_schema.tables where
  table_schema='test' and table_name='t_p;
```

an3l: Updated test case,thanks.

mysql-test/main/information_schema_temp_table.test
```diff
@@ -278,3 +278,36 @@ drop table seq2;
...
+select table_name from information_schema.tables where
+  table_schema='db' and table_name='temptable';
+select table_name from information_schema.tables where
+  table_schema='db' and table_name='faketable';
+CREATE TEMPORARY TABLE t2 (a INT);
+select table_name from information_schema.tables where
+  table_schema='db' and table_name='T2';
+SHOW TABLES LIKE 'T%';
+create  table tmp (a int);
+create  table t_p (a int);
```
vuvova, line 304: this should've been _create **temporary** table_.
but anyway, it won't show the problem, because the server apparently rechecks the condition later and even if your code allows unrelated tables in, the server will filter them out later.

So, remove this test case completely (with `tmp` and `t_p`) and try this:
```sql
create temporary table `t\_p` (a int);
select table_name from information_schema.tables where
  table_schema='db' and table_name='t\\_p';
drop table `t\_p`;
```

Try with both temporary and non-temporary table to see the difference in behavior

vuvova (changes requested):

mysql-test/main/information_schema_temp_table.test
an3l: Hi @vuvova,
I debugged and I don't see way how to:
1. catch difference for temporary tables between 
- a)
```sql
create temporary table `t\_p` (a int); select table_name from information_schema.tables where table_schema='db' and table_name='t\\_p'; 
# Result should be `t\_p`
```
- b)
```sql
CREATE TEMPORARY TABLE t2 (a INT);
select table_name from information_schema.tables where
  table_schema='db' and table_name='T2';
```
Both queries are currently failing in `wild_compare()` and both are respecting following conditions:
```
  if (!plan->lookup_field_vals.wild_table_value && schema_table_idx == SCH_TABLES)
```

2. compare table names for the regular tables.
    Regular tables are using `make_db_list()` and `make_table_name_list()`, but  I don't see when and where is `wild_compare()` checked (have debugged couldn't find stack trace for this debug point, but we know it works).
 
Can you give some hint as  direction to check please?

vuvova: sure.
```sql
create database `t\_p`;
create temporary table `t\_p`.tmp (a int);
select table_schema, table_name from information_schema.tables where
  table_schema='t\\_p';
```
_Disclaimer: untested_

vuvova:
fixed in 11.2 with 7fe764b1095

PR #2914, opened 2023-12-10
Feature that confirms with the User before performing the DELETE operation

Luko429:

## Description
TODO: fill description here
This feature within the command line interface confirms with the User before performing the DELETE operation.  It simply asks the user whether they are sure they want to proceed.  If yes then the DELETE operation is performed, otherwise the CLI doesn't perform the operation.  

## How can this PR be tested?
To test this PR, use the CLI and try to perform a DELETE operation on an existing database table.  You should be prompted with a question on whether you want to proceed with the operation.  A 'y' or 'Y' answer should perform the DELETE as usual and a 'n' or 'N' should bring the terminal back to the waiting state.  

## PR quality check


LinuxJedi:
Hi @Luko429,

Thanks for this contribution. I am going to need to reject this, I will walk you through why and how to get something reviewed in the future.

This feature appears to change the behaviour of MariaDB client so that you are prompted for a DELETE statement. There are several issues with this. First of all, this client can (and often is) used in a non-interactive mode. This change would break things, as can be seen by the regression suite failing. Also, there are other statements that could be considered just as dangerous. It is better to use correct permissions and the `--safe-updates` option if this is a concern.

In future, so that we do not have to straight-out reject this after it has been coded, you could create a ticket on our Jira system or contact us on the MariaDB developers mailing list or Zulip to see if it is a feature we would accept.

As for the PR itself. It looks like you have pulled in some extra things, such as submodule changes and changes that were subsequently removed in other commits. These would need cleaning up first, and the commits squashed to a single commit unless there is a good reason for multiple commits.

I hope this helps, if you have any questions, please feel free to reach out to us.

LukeRecInd:
Thanks for the review @LinuxJedi. I will look to get on the mailing list and ask questions about any future features I work on and if I have an idea I'll submit it on Jira. If anything this was a good introduction to learning about the Mariadb server project.  

Thanks again for the feedback

PR #2915, opened 2023-12-10
MDEV-32983 cosmetic improvement on path separator near ib_buffer_pool

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-32983*

## Description

Use correct path separator in calculating the ib_buffer_pool.

Reported thanks to Github user @celestinoxp.

Closes: https://github.com/ApacheFriends/xampp-build/issues/33

## How can this PR be tested?

## PR quality check


vaintroub:
there is no bug, and the message was accurate. It might be better not to mix up slashes inside a single path, so I do not mind the patch, but a comment should mention "cosmetics" or something like this, rather than "incorrect".

storage/innobase/buf/buf0dump.cc
```diff
@@ -180,7 +180,7 @@ static void buf_dump_generate_path(char *path, size_t path_size)
 	char	buf[FN_REFLEN];
 
 	mysql_mutex_lock(&LOCK_global_system_variables);
-	snprintf(buf, sizeof buf, "%s/%s", get_buf_dump_dir(),
+	snprintf(buf, sizeof buf, "%s%c%s", get_buf_dump_dir(), FN_LIBCHAR,
 		 srv_buf_dump_filename);
```
dr-m, line 184: The following should not induce any run-time overhead:
```c++
	snprintf(buf, sizeof buf, "%s" FN_ROOTDIR "%s", get_buf_dump_dir(),
		 srv_buf_dump_filename);
```
I think that we should adjust also the `format = '%s/%s'` later in the function. That path name is being passed to `buf_dump_status()` where it would be visible to the user.

dr-m (changes requested):
There also is a `"%s/%s%s"` in `ha_innobase::info_low()`, but the string that is generated by that format is only being passed to the file system, not to any user-facing diagnostic message.

dr-m (approved):

PR #2922, opened 2023-12-12
MDEV-18322 Assertion "wrong page type" on instant ALTER TABLE

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-18322*

## Description
`innobase_instant_try()`: Invoke `btr_set_instant()` in the same mini-transaction that has successfully inserted the metadata record. In this way, if inserting the metadata record fails before any undo log record was written for it, the index root page will remain consistent.
## How can this PR be tested?
The regression test was extended with a test case for reproducing this bug.
## PR quality check

### DIFF
--- a/storage/innobase/handler/handler0alter.cc
+++ b/storage/innobase/handler/handler0alter.cc
@@ -6130,10 +6130,6 @@ empty_table:
 			goto err_exit;
 		}
 
-		btr_set_instant(root, *index, &mtr);
-		mtr.commit();
-		mtr.start();
-		index->set_modified(mtr);
 		err = row_ins_clust_index_entry_low(
 			BTR_NO_LOCKING_FLAG, BTR_MODIFY_TREE, index,
 			index->n_uniq, entry, 0, thr);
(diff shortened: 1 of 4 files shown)


storage/innobase/handler/handler0alter.cc
```diff
@@ -6119,20 +6119,20 @@ static bool innobase_instant_try(
...
-		btr_set_instant(root, *index, &mtr);
-		mtr.commit();
-		mtr.start();
-		index->set_modified(mtr);
 		err = row_ins_clust_index_entry_low(
 			BTR_NO_LOCKING_FLAG, BTR_MODIFY_TREE, index,
 			index->n_uniq, entry, 0, thr);
+		if (err == DB_SUCCESS) {
+			btr_set_instant(root, *index, &mtr);
+		}
```
dr-m, line 6135: Sorry, I realized (after testing this on 10.6) that the patch is incorrect. We must pass `mtr` to `row_ins_clust_index_entry_low()`.

dr-m: Addressed in 54b1ff7531058eca238e939eaa1c9a397a672205. I am working on a separate 10.6 version of this fix.

Thirunarayanan (approved):

PR #2924, opened 2023-12-12
MDEV-32968  InnoDB fails to restore tablespace first page from doublewrite buffer when page is empty

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-32968*

## Description
InnoDB fails to find the space id from the page0 of
the tablespace. In that case, InnoDB can use
doublewrite buffer to recover the page0 and write
into the file.    
buf_dblwr_t::init_or_load_pages(): Loads only the pages
which are valid.(page lsn >= checkpoint). To do that,
InnoDB has to open the redo log before system
tablespace, read the latest checkpoint information.
    
 recv_dblwr_t::find_first_page():
    1) Iterate the doublewrite buffer pages and find the 0th page
    2) Read the tablespace flags, space id from the 0th page.
    3) Read the 1st, 2nd and 3rd page from tablespace file and
    compare the space id with the space id which is stored
    in doublewrite buffer.
    4) If it matches then we can write into the file.
    5) Return space which matches the pages from the file.
    
 SysTablespace::read_lsn_and_check_flags(): Remove the
  retry logic for validating the first page. After
  restoring the first page from doublewrite buffer,
  assign tablespace flags by reading the first page.
    
  recv_recovery_read_max_checkpoint(): Reads the maximum
  checkpoint information from log file
    
  recv_recovery_from_checkpoint_start(): Avoid reading
  the checkpoint header information from log file
   
  Datafile::validate_first_page(): Throw error in case
  of first page validation fails.
## How can this PR be tested?

./mtr innodb.doublewrite should test the added function

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/doublewrite.test
+++ b/mysql-test/suite/innodb/t/doublewrite.test
@@ -17,6 +17,7 @@ call mtr.add_suppression("InnoDB: A bad Space ID was found in datafile");
 call mtr.add_suppression("InnoDB: Checksum mismatch in datafile: ");
 call mtr.add_suppression("InnoDB: Inconsistent tablespace ID in .*t1\\.ibd");
 call mtr.add_suppression("\\[Warning\\] Found 1 prepared XA transactions");
+call mtr.add_suppression("InnoDB: Header page consists of zero bytes in datafile:");
 --enable_query_log
 
 let INNODB_PAGE_SIZE=`select @@innodb_page_size`;
@@ -24,7 +25,7 @@ let MYSQLD_DATADIR=`select @@datadir`;
 let ALGO=`select @@innodb_checksum_algorithm`;
 let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
 
-create table t1 (f1 int primary key, f2 blob) engine=innodb;
+create table t1 (f1 int primary key, f2 blob) stats_persistent=0, engine=innodb;
 
 start transaction;
 insert into t1 values(1, repeat('#',12));
@@ -38,7 +39,7 @@ commit work;
 SET GLOBAL innodb_fast_shutdown = 0;
 let $shutdown_timeout=;
 --source include/restart_mysqld.inc
-
+--source ../include/no_checkpoint_start.inc
 connect (dml,localhost,root,,);
 XA START 'x';
 insert into t1 values (6, repeat('%', @@innodb_page_size/2));
@@ -50,8 +51,8 @@ connection default;
 flush table t1 for export;
 
 let $restart_parameters=;
-let $shutdown_timeout=0;
---source include/shutdown_mysqld.inc
+--let CLEANUP_IF_CHECKPOINT=drop table t1, unexpected_checkpoint;
+--source ../include/no_checkpoint_end.inc
 
 perl;
 use IO::Handle;
@@ -115,6 +116,35 @@ EOF
 let SEARCH_PATTERN=InnoDB: Restoring page \[page id: space=[1-9][0-9]*, page number=0\] of datafile;
 --source include/search_pattern_in_file.inc
 let SEARCH_PATTERN=InnoDB: Recovered page \[page id: space=[1-9][0-9]*, page number=3\];
+--source include/search_pattern_in_file.inc
+XA ROLLBACK 'x';
+check table t1;
+select f1, f2 from t1;
+
+--source ../include/no_checkpoint_start.inc
+connect (dml,localhost,root,,);
+XA START 'x';
+insert into t1 values (6, repeat('%', @@innodb_page_size/2));
+XA END 'x';
+XA PREPARE 'x';
+disconnect dml;
+connection default;
+
+flush table t1 for export;
+
+let $restart_parameters=;
+--source ../include/no_checkpoint_end.inc
+
+# Zero out the first page in file and try to recover from dblwr
+perl;
+use IO::Handle;
+open(FILE, "+<", "$ENV{'MYSQLD_DATADIR'}test/t1.ibd") or die;
+syswrite(FILE, chr(0) x $ENV{INNODB_PAGE_SIZE});
+close FILE;
+EOF
+
+--source include/start_mysqld.inc
+let SEARCH_PATTERN=InnoDB: Restoring page \[page id: space=[1-9][0-9]*, page number=0\] of datafile;
 --source include/search_pattern_in_file.inc
 XA ROLLBACK 'x';
 check table t1;
--- a/storage/innobase/fsp/fsp0file.cc
+++ b/storage/innobase/fsp/fsp0file.cc
@@ -475,9 +475,16 @@ Datafile::validate_for_recovery()
 
 		err = find_space_id();
 		if (err != DB_SUCCESS || m_space_id == 0) {
-			ib::error() << "Datafile '" << m_filepath << "' is"
-				" corrupted. Cannot determine the space ID from"
-				" the first 64 pages.";
+
+			m_space_id = recv_sys.dblwr.find_first_page(
+				m_filepath, m_handle);
+
+			if (m_space_id) goto free_first_page;
+
+			sql_print_error(
+				"InnoDB: Datafile '%s' is corrupted."
+				" Cannot determine the space ID from"
+				" the first 64 pages.", m_filepath);
 			return(err);
 		}
 
@@ -485,7 +492,7 @@ Datafile::validate_for_recovery()
 			m_space_id, m_filepath, m_handle)) {
 			return(DB_CORRUPTION);
 		}
-
+free_first_page:
 		/* Free the previously read first page and then re-validate. */
 		free_first_page();
 		err = validate_first_page(0);
@@ -531,9 +538,9 @@ Datafile::validate_first_page(lsn_t* flush_lsn)
 
 	if (error_txt != NULL) {
 err_exit:
-		ib::info() << error_txt << " in datafile: " << m_filepath
-			<< ", Space ID:" << m_space_id  << ", Flags: "
-			<< m_flags;
+		sql_print_error("InnoDB: %s in datafile: %s, Space ID: %zu, "
+				"Flags: %zu", error_txt, m_filepath,
+				m_space_id, m_flags);
 		m_is_valid = false;
 		free_first_page();
 		return(DB_CORRUPTION);
--- a/storage/innobase/fsp/fsp0sysspace.cc
+++ b/storage/innobase/fsp/fsp0sysspace.cc
@@ -574,7 +574,7 @@ SysTablespace::read_lsn_and_check_flags(lsn_t* flushed_lsn)
 	}
 
 	err = it->read_first_page(
-		m_ignore_read_only ?  false : srv_read_only_mode);
+		m_ignore_read_only && srv_read_only_mode);
 
 	if (err != DB_SUCCESS) {
 		return(err);
@@ -588,20 +588,17 @@ SysTablespace::read_lsn_and_check_flags(lsn_t* flushed_lsn)
 
 	/* Check the contents of the first page of the
 	first datafile. */
-	for (int retry = 0; retry < 2; ++retry) {
+	err = it->validate_first_page(flushed_lsn);
 
-		err = it->validate_first_page(flushed_lsn);
-
-		if (err != DB_SUCCESS
-		    && (retry == 1
-			|| recv_sys.dblwr.restore_first_page(
+	if (err != DB_SUCCESS) {
+		if (recv_sys.dblwr.restore_first_page(
 				it->m_space_id, it->m_filepath,
-				it->handle()))) {
-
+				it->handle())) {
 			it->close();
-
 			return(err);
 		}
+		err = it->read_first_page(
+			m_ignore_read_only && srv_read_only_mode);
 	}
 
 	/* Make sure the tablespace space ID matches the
--- a/storage/innobase/include/buf0dblwr.h
+++ b/storage/innobase/include/buf0dblwr.h
@@ -103,7 +103,8 @@ public:
   If we are upgrading from a version before MySQL 4.1, then this
   function performs the necessary update operations to support
   innodb_file_per_table. If we are in a crash recovery, this function
-  loads the pages from double write buffer into memory.
+  loads the pages from double write buffer which are not older than
+  the checkpoint into memory.
   @param file File handle
   @param path Path name of file
   @return DB_SUCCESS or error code */
--- a/storage/innobase/include/log0recv.h
+++ b/storage/innobase/include/log0recv.h
@@ -49,6 +49,11 @@ recv_find_max_checkpoint(ulint* max_field)
 ATTRIBUTE_COLD void recv_recover_page(fil_space_t* space, buf_page_t* bpage)
 	MY_ATTRIBUTE((nonnull));
 
+/** Read the latest checkpoint information from log file
+and store it in log_sys.next_checkpoint and recv_sys.mlog_checkpoint_lsn
+@return error code or DB_SUCCESS */
+dberr_t recv_recovery_read_max_checkpoint();
+
 /** Start recovering from a redo log checkpoint.
 @param[in]	flush_lsn	FIL_PAGE_FILE_FLUSH_LSN
 of first system tablespace page
@@ -141,7 +146,18 @@ struct recv_dblwr_t
   @param file      tablespace file handle
   @return whether the operation failed */
   bool restore_first_page(
-         ulint space_id, const char *name, os_file_t file);
+         ulint space_id, const char *name, pfs_os_file_t file);
+
+  /** Restore the first page of the given tablespace from
+  doublewrite buffer.
+  1) Find the page which has page_no as 0
+  2) Read first 3 pages from tablespace file
+  3) Compare the space_ids from the pages with page0 which
+  was retrieved from doublewrite buffer
+  @param name tablespace filepath
+  @param file tablespace file handle
+  @return space_id or 0 in case of error */
+  uint32_t find_first_page(const char *name, pfs_os_file_t file);
 
   typedef std::deque<byte*, ut_allocator<byte*> > list;
 
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3415,6 +3415,46 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
 	return DB_SUCCESS;
 }
 
+dberr_t recv_recovery_read_max_checkpoint()
+{
+  ut_ad(srv_operation <= SRV_OPERATION_EXPORT_RESTORED ||
+        srv_operation == SRV_OPERATION_RESTORE ||
+        srv_operation == SRV_OPERATION_RESTORE_EXPORT);
+  ut_d(mysql_mutex_lock(&buf_pool.mutex));
+  ut_ad(UT_LIST_GET_LEN(buf_pool.LRU) == 0);
+  ut_ad(UT_LIST_GET_LEN(buf_pool.unzip_LRU) == 0);
+  ut_d(mysql_mutex_unlock(&buf_pool.mutex));
+
+  if (srv_force_recovery >= SRV_FORCE_NO_LOG_REDO)
+  {
+    sql_print_information("InnoDB: innodb_force_recovery=6 skips redo log apply");
+    return DB_SUCCESS;
+  }
+
+  mysql_mutex_lock(&log_sys.mutex);
+  ulint max_cp;
+  dberr_t err= recv_find_max_checkpoint(&max_cp);
+
+  if (err != DB_SUCCESS)
+    recv_sys.recovered_lsn= log_sys.get_lsn();
+  else
+  {
+    byte* buf= log_sys.checkpoint_buf;
+    err= log_sys.log.read(max_cp, {buf, OS_FILE_LOG_BLOCK_SIZE});
+    if (err == DB_SUCCESS)
+    {
+      log_sys.next_checkpoint_no=
+        mach_read_from_8(buf + LOG_CHECKPOINT_NO);
+      log_sys.next_checkpoint_lsn=
+        mach_read_from_8(buf + LOG_CHECKPOINT_LSN);
+      recv_sys.mlog_checkpoint_lsn=
+        mach_read_from_8(buf + LOG_CHECKPOINT_END_LSN);
+    }
+  }
+  mysql_mutex_unlock(&log_sys.mutex);
+  return err;
+}
+
 /** Start recovering from a redo log checkpoint.
 @param[in]	flush_lsn	FIL_PAGE_FILE_FLUSH_LSN
 of first system tablespace page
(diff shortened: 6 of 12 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -3845,6 +3845,49 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+        goto next_page;
+      if (buf_page_is_corrupted(false, read_page, flags)
+          || space_id != page_get_space_id(read_page))
+	goto next_page;
+    }
+    space= space_id;
+    break;
+  }
+
+  if (space && !restore_first_page(space, name, file))
```
dr-m, line 3886: Side note: `find_page()` in `restore_first_page()` will not check that the `FIL_PAGE_LSN` is at least the log checkpoint LSN. I think that this should be fixed in `buf_dblwr_t::init_or_load_pages()` (skip pages whose `FIL_PAGE_LSN` is too old).

storage/innobase/log/log0recv.cc
```diff
@@ -3845,6 +3845,49 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
 }
 
+uint32_t recv_dblwr_t::find_first_page(const char* name, os_file_t file)
+{
+  uint32_t space= 0;
+  os_offset_t file_size= os_file_get_size(file);
+  if (file_size == (os_offset_t) -1) return space;
+  for (auto it= pages.begin(); it != pages.end();)
+  {
+    byte* page= *it;
```
dr-m, line 3855:
```c++
  for (const page_t *page : pages)
```

storage/innobase/log/log0recv.cc
```diff
@@ -3845,6 +3845,49 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+    uint32_t space_id= page_get_space_id(page);
+    if (page_get_page_no(page) > 0 || space_id == 0)
+    {
+next_page:
+      it++;
+      continue;
+    }
```
dr-m, line 3862: The body can be just `continue;`.

dr-m (commented):
This looks mostly OK, but I would like some cleanup. It looks like we may try to use outdated pages from the doublewrite buffer.

storage/innobase/log/log0recv.cc
```diff
@@ -3845,6 +3872,46 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+    doublewrite buffer page. */
+    for (ulint j= 1; j <= 3; j++)
+    {
+      if (os_file_read(IORequestRead, file, read_page,
+                       j * page_size, page_size) != DB_SUCCESS)
+        goto next_page;
+      if (buf_page_is_corrupted(false, read_page, flags) ||
+          memcmp(read_page + FIL_PAGE_SPACE_ID,
+                 page + FIL_PAGE_SPACE_ID, 4))
+	goto next_page;
```
dr-m, line 3904: We should execute the cheapest check first. Also, it would be good to validate the page number:
```c++
      if (page_get_page_no(read_page) != j ||
          memcmp(read_page + FIL_PAGE_SPACE_ID,
                 page + FIL_PAGE_SPACE_ID, 4) ||
                 buf_page_is_corrupted(false, read_page, flags))
	goto next_page;
```

storage/innobase/log/log0recv.cc
```diff
@@ -3845,6 +3872,46 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+next_page:
+      continue;
+    uint32_t flags= mach_read_from_4(
+      FSP_HEADER_OFFSET + FSP_SPACE_FLAGS + page);
+    page_id_t page_id(space_id, 0);
+    size_t page_size= fil_space_t::physical_size(flags);
+    if (file_size < 4 * page_size) goto next_page;
+    byte *read_page= static_cast<byte*>(
+      aligned_malloc(page_size, page_size));
+    /** Read 3 pages from the file and match the space id
```
dr-m, line 3893: Doxygen does not look for comments inside a function body. Start the comment with a normal `/*`.

storage/innobase/include/log0recv.h
```diff
@@ -143,6 +147,15 @@ struct recv_dblwr_t
...
          ulint space_id, const char *name, os_file_t file);
 
+  /** Restore the first page of the given tablespace from
+  doublewrite buffer.
+  1) Find the page which has page_no as 0
+  2) Read first 3 pages from tablespace file
+  3) Compare the space_ids from the pages with page0 which
+  was retrieved from doublewrite buffer
+  @return space_id or 0 in case of error */
+  uint32_t find_first_page(const char* name, os_file_t file);
```
dr-m, line 157: Neither `name` nor `file` are documented with `@param`. Shouldn’t the type be `pfs_os_file_t`? The parameters are in opposite order of those of `init_or_load_pages()`.

storage/innobase/log/log0recv.cc
```diff
@@ -3845,6 +3872,46 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+      if (buf_page_is_corrupted(false, read_page, flags) ||
+          memcmp(read_page + FIL_PAGE_SPACE_ID,
+                 page + FIL_PAGE_SPACE_ID, 4))
+	goto next_page;
+    }
+    space= space_id;
+    break;
+  }
+
+  if (space && !restore_first_page(space, name, file))
```
dr-m, line 3910: If we replace the `break` statement with the `if` we can remove a condition as well as the local variable `space`:
```c++
    if (!restore_first_page(space_id, name, file))
      return space_id;
    break;
  }
  return 0;
```

storage/innobase/log/log0recv.cc
```diff
@@ -3415,19 +3415,64 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
 	return DB_SUCCESS;
 }
 
+dberr_t
+recv_recovery_read_max_checkpoint()
```
dr-m, line 3419: Let’s avoid unnecessary line breaks.

storage/innobase/log/log0recv.cc
```diff
@@ -3415,19 +3415,64 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
...
+{
+  ulint		max_cp_field;
+  dberr_t	err = DB_SUCCESS;
```
dr-m, line 3422: TAB is not allowed in the common formatting style.

storage/innobase/log/log0recv.cc
```diff
@@ -3415,19 +3415,64 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
...
+    return err;
+  }
+
+  log_sys.next_checkpoint_no= mach_read_from_8(
+				buf + LOG_CHECKPOINT_NO);
+  log_sys.next_checkpoint_lsn= mach_read_from_8(
+				buf + LOG_CHECKPOINT_LSN);
+  recv_sys.mlog_checkpoint_lsn= mach_read_from_8(
+				buf + LOG_CHECKPOINT_END_LSN);
+  mysql_mutex_unlock(&log_sys.mutex);
```
dr-m, line 3463: If we have to split lines, we should avoid it right after `(`. Instead, split the line after the assignment operator.

Can we avoid duplicating the `mysql_mutex_unlock()` calls?
```c++
  mysql_mutex_lock(&log_sys.mutex);
  ulint max_cp;
  dberr_t err= recv_find_max_checkpoint(&max_cp);

  if (err != DB_SUCCESS)
    recv_sys.recovered_lsn= log_sys.get_lsn();
  else
  {
    byte *buf= log_sys.checkpoint_buf;
    err= log_sys.log.read(max_cp, {buf, OS_FILE_LOG_BLOCK_SIZE});
    if (err == DB_SUCCESS)
    {
      log_sys.next_checkpoint_no= mach_read_from_8(buf + LOG_CHECKPOINT_NO);
      log_sys.next_checkpoint_lsn= mach_read_from_8(buf + LOG_CHECKPOINT_LSN);
      recv_sys.mlog_checkpoint_lsn=
        mach_read_from_8(buf + LOG_CHECKPOINT_END_LSN);
    }
  }

  mysql_mutex_unlock(&log_sys.mutex);
  return err;

File: storage/innobase/srv/srv0start.cc [hunk 11]
```diff
@@ -1339,47 +1301,80 @@ dberr_t srv_start(bool create_new_db)
...
+			return(srv_init_abort(err));
+		}
+
+		/* Suppress the message about
+		crash recovery. */
+		flushed_lsn = log_sys.get_lsn();
+		goto file_checked;
 	}
 
 file_checked:
```
dr-m, line 1380: The comment after this line needs to be modified. The log file has already been opened.

dr-m (changes requested):
Thank you. I see that some rather extensive refactoring in `srv_start()` is unavoidable. I think that we will need a 10.11 version of this as well, and both the 10.5 version (possibly ported to 10.6) and 10.11 version must be extensively tested.

storage/innobase/log/log0recv.cc
```diff
@@ -3415,19 +3415,64 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
...
+  ulint		max_cp_field;
+  dberr_t	err = DB_SUCCESS;
+
+  ut_ad(srv_operation <= SRV_OPERATION_EXPORT_RESTORED
+        || srv_operation == SRV_OPERATION_RESTORE
+        || srv_operation == SRV_OPERATION_RESTORE_EXPORT);
+  ut_d(mysql_mutex_lock(&buf_pool.flush_list_mutex));
+  ut_ad(UT_LIST_GET_LEN(buf_pool.LRU) == 0);
+  ut_ad(UT_LIST_GET_LEN(buf_pool.unzip_LRU) == 0);
+  ut_d(mysql_mutex_unlock(&buf_pool.flush_list_mutex));
```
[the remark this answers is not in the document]
Thirunarayanan: No.

storage/innobase/fsp/fsp0file.cc
```diff
@@ -531,7 +537,7 @@ Datafile::validate_first_page(lsn_t* flush_lsn)
 
 	if (error_txt != NULL) {
 err_exit:
-		ib::info() << error_txt << " in datafile: " << m_filepath
+		ib::error() << error_txt << " in datafile: " << m_filepath
 			<< ", Space ID:" << m_space_id  << ", Flags: "
 			<< m_flags;
```
dr-m, line 542: We could (and IMO should) use `sql_print_error()` here.

storage/innobase/include/buf0dblwr.h
```diff
@@ -103,7 +103,8 @@ class buf_dblwr_t
   If we are upgrading from a version before MySQL 4.1, then this
   function performs the necessary update operations to support
   innodb_file_per_table. If we are in a crash recovery, this function
-  loads the pages from double write buffer into memory.
+  loads the pages from double write buffer which are larger than
+  checkpoint lsn into memory.
```
dr-m, line 107: s/larger than checkpoint lsn/not older than the checkpoint/

storage/innobase/include/log0recv.h
```diff
@@ -49,6 +49,10 @@ recv_find_max_checkpoint(ulint* max_field)
 ATTRIBUTE_COLD void recv_recover_page(fil_space_t* space, buf_page_t* bpage)
 	MY_ATTRIBUTE((nonnull));
 
+/** Read the latest checkpoint information from log file
+and store it in log_sys.next_checkpoint and recv_sys.mlog_checkpoint_lsn */
+dberr_t recv_recovery_read_max_checkpoint();
```
dr-m, line 54: Missing `@retval` comments

storage/innobase/log/log0recv.cc
```diff
@@ -3415,19 +3415,55 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
...
+    sql_print_information("innodb_force_recovery=6 skips redo log apply");
+    return DB_SUCCESS;
+  }
+
+  mysql_mutex_lock(&log_sys.mutex);
+  ulint max_cp;
+  dberr_t err= recv_find_max_checkpoint(&max_cp);
+
+  if (err != DB_SUCCESS)
+     recv_sys.recovered_lsn= log_sys.get_lsn();
```
dr-m, line 3439: This line is indented 1 space too deep.

storage/innobase/log/log0recv.cc
```diff
@@ -3415,19 +3415,55 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
...
+        || srv_operation == SRV_OPERATION_RESTORE
+        || srv_operation == SRV_OPERATION_RESTORE_EXPORT);
+  ut_d(mysql_mutex_lock(&buf_pool.mutex));
+  ut_ad(UT_LIST_GET_LEN(buf_pool.LRU) == 0);
+  ut_ad(UT_LIST_GET_LEN(buf_pool.unzip_LRU) == 0);
+  ut_d(mysql_mutex_unlock(&buf_pool.mutex));
+
+  if (srv_force_recovery >= SRV_FORCE_NO_LOG_REDO)
+  {
+    sql_print_information("innodb_force_recovery=6 skips redo log apply");
```
dr-m, line 3430: Missing `InnoDB:` prefix.

storage/innobase/log/log0recv.cc
```diff
@@ -3445,27 +3481,11 @@ recv_recovery_from_checkpoint_start(lsn_t flush_lsn)
...
-
-	buf = log_sys.checkpoint_buf;
-	if ((err= log_sys.log.read(max_cp_field, {buf, OS_FILE_LOG_BLOCK_SIZE}))) {
-		mysql_mutex_unlock(&log_sys.mutex);
-		return(err);
-	}
-
-	checkpoint_lsn = mach_read_from_8(buf + LOG_CHECKPOINT_LSN);
-	checkpoint_no = mach_read_from_8(buf + LOG_CHECKPOINT_NO);
+	ib_uint64_t checkpoint_no= log_sys.next_checkpoint_no;
```
dr-m, line 3485: We could just use the standard type `uint64_t`.

storage/innobase/log/log0recv.cc
```diff
@@ -3845,9 +3863,51 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+    uint32_t flags= mach_read_from_4(
+      FSP_HEADER_OFFSET + FSP_SPACE_FLAGS + page);
+    page_id_t page_id(space_id, 0);
+    size_t page_size= fil_space_t::physical_size(flags);
+    if (file_size < 4 * page_size) goto next_page;
+    byte *read_page= static_cast<byte*>(
+      aligned_malloc(page_size, page_size));
+    /* Read 3 pages from the file and match the space id
+    with the space id which is stored in
+    doublewrite buffer page. */
```
dr-m, line 3886: Can we allocate a buffer of 3 pages and invoke `os_file_read()` only once on that larger buffer?

storage/innobase/srv/srv0start.cc
```diff
@@ -1343,51 +1305,84 @@ dberr_t srv_start(bool create_new_db)
...
+	err = srv_sys_space.open_or_create(
+		false, create_new_db, &sum_of_new_sizes, &flushed_lsn);
 
-		log_sys.log.create();
+	switch (err) {
+	case DB_SUCCESS:
+		break;
+	case DB_CANNOT_OPEN_FILE:
+		ib::error()
+			<< "Could not open or create the system tablespace. If"
```
dr-m, line 1342: Let’s use `sql_print_error()` when there is no complex `operator<<` involved.

storage/innobase/srv/srv0start.cc
```diff
@@ -1343,51 +1305,84 @@ dberr_t srv_start(bool create_new_db)
...
+		/* Other errors might come from Datafile::validate_first_page() */
+		return(srv_init_abort(err));
+	}
 
-		if (!log_set_capacity(srv_log_file_size_requested)) {
-			return(srv_init_abort(DB_ERROR));
+	if (!create_new_db && create_new_log) {
+		if (flushed_lsn < lsn_t(1000)) {
+			ib::error()
+				<< "Cannot create log file because"
```
dr-m, line 1360: `sql_print_error`

storage/innobase/fsp/fsp0file.cc
```diff
@@ -475,6 +475,12 @@ Datafile::validate_for_recovery()
...
 		if (err != DB_SUCCESS || m_space_id == 0) {
+
+			m_space_id = recv_sys.dblwr.find_first_page(
+					m_filepath, m_handle);
+
+			if (m_space_id) goto free_first_page;
+
 			ib::error() << "Datafile '" << m_filepath << "' is"
 				" corrupted. Cannot determine the space ID from"
 				" the first 64 pages.";
```
dr-m, line 486: Invoking `sql_print_error` would save a number of calls and some heap allocation.

storage/innobase/fsp/fsp0file.cc
```diff
@@ -475,6 +475,12 @@ Datafile::validate_for_recovery()
 
 		err = find_space_id();
 		if (err != DB_SUCCESS || m_space_id == 0) {
+
+			m_space_id = recv_sys.dblwr.find_first_page(
+					m_filepath, m_handle);
```
dr-m, line 480: The second line is indented too deep.

dr-m (changes requested):
Thank you, this is almost good to go.

storage/innobase/fsp/fsp0file.cc
```diff
@@ -531,9 +538,9 @@ Datafile::validate_first_page(lsn_t* flush_lsn)
 
 	if (error_txt != NULL) {
 err_exit:
-		ib::info() << error_txt << " in datafile: " << m_filepath
-			<< ", Space ID:" << m_space_id  << ", Flags: "
-			<< m_flags;
+		sql_print_error("InnoDB: %s in datafile: %s, Space ID: %d, "
+				"Flags: %d", error_txt, m_filepath,
+				m_space_id, m_flags);
```
dr-m, line 543: `m_space_id` and `m_flags` are not `int`. They require `ULINTPF` or `"%zu"` in this version.

storage/innobase/fsp/fsp0sysspace.cc
```diff
@@ -588,20 +588,17 @@ SysTablespace::read_lsn_and_check_flags(lsn_t* flushed_lsn)
...
 				it->m_space_id, it->m_filepath,
-				it->handle()))) {
-
+				it->handle())) {
 			it->close();
-
 			return(err);
 		}
+		err = it->read_first_page(
+			m_ignore_read_only ? false : srv_read_only_mode);
```
dr-m, line 601: Can we write `m_ignore_read_only && srv_read_only_mode`?

dr-m (changes requested):

mysql-test/suite/innodb/t/doublewrite.test
```diff
@@ -119,6 +120,35 @@ let SEARCH_PATTERN=InnoDB: Recovered page \[page id: space=[1-9][0-9]*, page num
...
+XA END 'x';
+XA PREPARE 'x';
+disconnect dml;
+connection default;
+
+flush table t1 for export;
+
+let $restart_parameters=;
+let $shutdown_timeout=0;
+--source include/shutdown_mysqld.inc
```
dr-m, line 136: Please specify `CREATE TABLE t1 … STATS_PERSISTENT=0;` so that we will not get any unexpected log checkpoints here.

Use `no_checkpoint_start.inc` and `no_checkpoint_end.inc` around the critical section, and specify a non-existing table in the cleanup action so that the test will fail in an obvious way if an unexpected checkpoint occurs.

storage/innobase/fsp/fsp0sysspace.cc
dr-m: You did not address this comment everywhere.

storage/innobase/include/log0recv.h
```diff
@@ -141,7 +146,18 @@ struct recv_dblwr_t
...
+  /** Restore the first page of the given tablespace from
+  doublewrite buffer.
+  1) Find the page which has page_no as 0
+  2) Read first 3 pages from tablespace file
+  3) Compare the space_ids from the pages with page0 which
+  was retrieved from doublewrite buffer
+  @param name tablespace filepath
+  @param file tablespace file handle
+  @return space_id or 0 in case of error */
+  uint32_t find_first_page(const char* name, pfs_os_file_t file);
```
dr-m, line 160: Formatting nitpick: in the non-legacy formatting style, there should be a space before `*` and not after it.

storage/innobase/log/log0recv.cc
```diff
@@ -3845,9 +3863,52 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
   return result;
 }
 
+uint32_t recv_dblwr_t::find_first_page(const char* name, pfs_os_file_t file)
```
dr-m, line 3866: formatting

storage/innobase/log/log0recv.cc
```diff
@@ -3845,9 +3863,52 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+{
+  uint32_t space= 0;
+  os_offset_t file_size= os_file_get_size(file);
+  if (file_size == (os_offset_t) -1) return space;
```
dr-m, line 3870: I think that it would be clearer to break the line before `return`, and maybe even change it to `return 0` to make it stand out more. A `return 0` or `return space` should make no difference to the generated code, thanks to common subexpression elimination at a high-level stage in every compiler.

Alternatively, we could negate the condition and put the entire `for` loop inside `if (file_size != os_offset_t(-1))`.

storage/innobase/log/log0recv.cc
```diff
@@ -3845,9 +3863,52 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+    for (ulint j= 0; j <= 2; j++)
+    {
+      byte *cur_page= read_page + j * page_size;
+      if (page_get_page_no(cur_page) != j + 1 ||
+          memcmp(cur_page + FIL_PAGE_SPACE_ID,
+                 page + FIL_PAGE_SPACE_ID, 4) ||
+          buf_page_is_corrupted(false, cur_page, flags))
+	goto next_page;
+    }
+    space= space_id;
```
dr-m, line 3899: `space` seems to be an unused variable. Maybe in an earlier version of this function, there was a `return space` at the end.

storage/innobase/log/log0recv.cc
```diff
@@ -3845,9 +3863,52 @@ byte *recv_dblwr_t::find_page(const page_id_t page_id,
...
+    if (page_get_page_no(page) > 0 || space_id == 0)
+next_page:
+      continue;
+    uint32_t flags= mach_read_from_4(
+      FSP_HEADER_OFFSET + FSP_SPACE_FLAGS + page);
+    page_id_t page_id(space_id, 0);
+    size_t page_size= fil_space_t::physical_size(flags);
+    if (file_size < 4 * page_size) goto next_page;
+    byte *read_page= static_cast<byte*>(
+      aligned_malloc(3 * page_size, page_size));
```
dr-m, line 3883: Formatting: break the line between the `if` condition and the body, unless it is part of a very short block.

Also, do not use the legacy InnoDB style line breaks after `(`. Break lines after `,` or similar, but never after `(`, to be compatible with the default settings of the GNU Emacs `cc-mode`.

dr-m (commented):
Much better, but I noticed some room for improvement, mainly in the test.

dr-m (approved):

PR #2928, opened 2023-12-13
MDEV-30908 Create INVOKING_USER() function

christian7007:
- [x] *The Jira issue number for this PR is: MDEV-30908*

## Description
Add `INVOKING_USER()` function to easily retrieve the user and host invoking a stored procedure when `SQL SECURITY DEFINER` is set. The current alternative, `USER()`, will return the host used by the client instead of the value of the `host` column of the `mysql.user` table (as `CURRENT_USER()` does) making it harder to know exactly which row from `mysql.user` the user was authenticated against

Below there is an example of the output of this function compared with the existing alternatives, if we create `mysql.user_function_demo()` like:

```
CREATE PROCEDURE mysql.user_function_demo() SQL SECURITY DEFINER BEGIN
SELECT CURRENT_USER(), USER(), INVOKING_USER();
END
```

use the command below to connect to the server:

```
$ mariadb -u test_user -h 127.0.0.1   # force connection to be over TCP, not Unix socket
```

and then execute the procedure:

```
CURRENT_USER()	USER()	                   INVOKING_USER()
root@localhost	test_user@127.0.0.1:12345  test_user@%
```

## How can this PR be tested?

This PR adds a new MTR test for testing this new function.

## PR quality check

## Copyright
Amazon Web Services, Inc.


mysql-test/main/invoking_user.test
```diff
@@ -0,0 +1,31 @@
...
+#
+# Check that INVOKING_USER() returns the right user and host
+# matching the content of `mysql.user` table for the user
+# executing the stored procedure.
+
+-- source include/not_embedded.inc
+
+--echo # Create a stored procedure to call the different ways to retrieve the user
+--delimiter /
+CREATE PROCEDURE mysql.user_function_demo() SQL SECURITY DEFINER BEGIN
```
grooverdan, line 11: Could just concatenate the 3 functions with a `;` separate and return that. The select from mysql.user isn't particularly adding to the test case.

If this was removed, small simplification - don't need `begin/end` or delimiter if its just single statement.

Not need to put into the `mysql` database. Default `test` database serves the same function.

grooverdan: Note a VIEW is another way the function could be exposed and a test around that would be good.

Can you check if a procedure calling a select on the view also returns the right results?

mysql-test/main/invoking_user.test
```diff
@@ -0,0 +1,31 @@
...
+connect (test_user_con,localhost,test_user,,);
+--echo # Check the function outside of a stored procedure
+SELECT CURRENT_USER(), USER(), INVOKING_USER();
+--echo # Check the function inside of a stored procedure
+CALL mysql.user_function_demo();
+
+--echo # Cleanup
+--connection default
+DROP USER 'test_user'@'%';
+DROP PROCEDURE mysql.user_function_demo;
```
grooverdan, line 31: Convention is to end with:
`--echo # End of 11.4 tests`

sql/lex.h
```diff
@@ -317,8 +317,9 @@ SYMBOL symbols[] = {
   { "INT8",		SYM(BIGINT)},
   { "INTEGER",		SYM(INT_SYM)},
   { "INTERVAL",		SYM(INTERVAL_SYM)},
-  { "INVISIBLE",        SYM(INVISIBLE_SYM)},
   { "INTO",		SYM(INTO)},
+  { "INVISIBLE",        SYM(INVISIBLE_SYM)},
```
grooverdan, line 321: lets leave the misording of INVISIBLE to a different problem. Someone fixing earlier and merging up may end up with a double-invisible.

grooverdan (changes requested):
small changes. Good to see it happen finally.

I assume its [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ORA_INVOKING_USER.html) that the function name was inspired by. I can't find a postgres reference equivalent. Have you looked elsewhere in other implementations?

You could add `ORA_INVOKING_USER` as an alias to improve oracle mode compatibility (and not worry about oracle mode being set for the function to be available).

A test case without the optional braces would be good too.

christian7007:
Thanks for the feedback @grooverdan! I just updated the PR addressing it.

christian7007:
It seems [buildbot/amd64-ubuntu-2004-debug](https://buildbot.mariadb.org/#builders/369/builds/15909) failed with some unrelated error. Is it possible to re-run it?

vuvova:
Thoughts:
* the standard doesn't have `INVOKING_USER`, it has `CURRENT_USER` and `SESSION_USER`. In MariaDB and MySQL, `SESSION_USER` always was an alias for `USER`, but perhaps it's time to change it to a more standard-compliant behavior? Keep `USER` and is and make `SESSION_USER` return what `CURRENT_USER` returns without routines or views? With an `old_mode` value to restore historical behavior, if needed.
* No matter what the function is called, currently in the PR it returns the session user. But if one stored routine invokes another stored routine — then strictly speaking the **invoking user** is the definer of the routine one level up the stack, not the session user.
* And if the invoking user is the definer of the routine one level up the stack, then perhaps there should be a way of getting an invoker any number of levels up the stack? Or is it over-engineering?

A couple of comments about the code
* a function invocation always uses parentheses, they aren't optional.
* there's no need to allow `ORA_INVOKING_USER` outside of `sql_mode=ORACLE`

christian7007:
Hey @vuvova, thanks for the feedback!

> the standard doesn't have INVOKING_USER, it has CURRENT_USER and SESSION_USER. In MariaDB and MySQL, SESSION_USER always was an alias for USER, but perhaps it's time to change it to a more standard-compliant behavior? Keep USER and is and make SESSION_USER return what CURRENT_USER returns without routines or views? With an old_mode value to restore historical behavior, if needed.

I think this will cover our use case and it will simplify the solution (the other two points you mentioned wouldn't be necessary if we go with this). For us it's usually enough to just know the real user logged in so we can do permission checks properly. Not sure if you or maybe @grooverdan have in mind any use case where actually knowing intermediate invokers in a stack of stored procedures calls will be required. 

> No matter what the function is called, currently in the PR it returns the session user. But if one stored routine invokes another stored routine — then strictly speaking the invoking user is the definer of the routine one level up the stack, not the session user.

I guess if we change the behaviour of `SESSION_USER` instead of adding `INVOKING_USER` this won't be required as the `SESSION_USER` will technically be the same even using nested calls to procedures whit different definers. But I agree if we decide to proceed with I`NVOKING_USER` this is something we should take into account.

> And if the invoking user is the definer of the routine one level up the stack, then perhaps there should be a way of getting an invoker any number of levels up the stack? Or is it over-engineering?

I think this is a bit over-engineering yep, If we decide to go with the `INVOKING_USER`, I think we should only return the one calling the specific procedure executing the call, and if for some reason someone is calling multiple procedures and needs to know the `INVOKING_USER` from a call lower in the sack they can pass it as argument.

Let me know which alternative works better for you and I'll update the PR accordingly, I lean more towards updating `SESSION_USER` behaviour as it will be much more simpler and it will still cover our use case, but if you think it's worth it to go with `INVOKING_USER` I can also work on this and take into account the nested calls.

Regarding the code comments, I will remove the optional parameters and work one the `ORA_INVOKING_USER` depending on the final approach we decide (I assume `ORA_INVOKING_USER` won't be needed if we go with the `SESSION_USER` approach).

dlenski:
@vuvova wrote:

> * the standard doesn't have `INVOKING_USER`, it has `CURRENT_USER` and `SESSION_USER`. In MariaDB and MySQL, `SESSION_USER` always was an alias for `USER`, but perhaps it's time to change it to a more standard-compliant behavior?
> Keep `USER` and is and make `SESSION_USER` return what `CURRENT_USER` returns without routines or views? With an `old_mode` value to restore historical behavior, if needed.

I am all for not creating a new non-standard name here :+1:

So replacing `SESSION_USER()` , instead of creating `INVOKING_USER()`, sounds good.

(Keep in mind, though, that even though the _names_ may be defined by ANSI SQL standards, there's absolutely no consistency of implementation or semantics across different SQL databases. I'm just mentioning this to forestall future objections.)

christian7007:
I have published https://github.com/christian7007/server/commit/4c9ec8155eb9d72edc4a30560ff941d70e634eb5 with the modifications to use `SESSION_USER()` instead.

@vuvova @grooverdan let me know your thoughts and if you agree on switching to that approach I will add the `OLD_MODE`, close this PR and open a new one with the `SESSION_USER()` changes.

vuvova:
I agree that SQL standard doesn't specify the semantics. Still I think it is reasonable to assume that both standard `SESSION_USER` and `CURRENT_USER` functions should return values from the same domain, that their return values should be comparable, should have the same semantics (whatever it is). As of now they don't.

christian7007:
Sounds good to me. I'm traveling for Christmas but as soon as I'm back I'll open a new PR for this addressing all the feedback you provide in the commit, and adding the `OLD_MODE` option.

christian7007:
@vuvova @grooverdan I've open a new PR with the changes to `SESSION_USER()`.

PR #2930, opened 2023-12-13
MDEV-33009 Server hangs for a long time with innodb_undo_log_truncate=ON

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33009*
## Description
`trx_purge_truncate_history()`: When we need to wait for a page latch, acquire and release `buf_pool.mutex` to ensure that any thread that may be holding `buf_pool.mutex` and waiting for `buf_pool.flush_list_mutex` will be able to proceed.

This fixes up commit a0f02f743848fb6ef6d2a51960f77e1e38da1682 (MDEV-32757).
## How can this PR be tested?
This is best tested by @hgxl64 in the Ubuntu 18.04 environment where the original anomaly was observed. The existing regression tests `innodb.undo_truncate` and `innodb.undo_truncate_recover` does cover this code.

## Basing the PR against the correct MariaDB version
- [ ] *This is a new feature and the PR is based against the latest MariaDB development branch.*
- [x] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*
## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [x] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #2931, opened 2023-12-13
MDEV-33009 Server hangs for a long time with innodb_undo_log_truncate=ON

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33009*

## Description
`trx_purge_truncate_history()`: When we need to wait for a page latch, acquire and release `buf_pool.mutex` to ensure that any thread that may be holding `buf_pool.mutex` and waiting for `buf_pool.flush_list_mutex` will be able to proceed.

This fixes up commit 5dbe7a8c9aa88b7ed17311c2f1df651c9da7783b (MDEV-32757).
## How can this PR be tested?
This is best tested by @hgxl64 in the Ubuntu 18.04 environment where the original anomaly was observed. The existing regression tests `innodb.undo_truncate` and `innodb.undo_truncate_recover` do cover this code.
- [x] This is a 10.6 version of #2930.
## PR quality check

### DIFF
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -687,8 +687,8 @@ not_free:
     mini-transaction commit and the server was killed, then
     discarding the to-be-trimmed pages without flushing would
     break crash recovery. */
-    mysql_mutex_lock(&buf_pool.flush_list_mutex);
   rescan:
+    mysql_mutex_lock(&buf_pool.flush_list_mutex);
     for (buf_page_t *bpage= UT_LIST_GET_LAST(buf_pool.flush_list); bpage; )
     {
       ut_ad(bpage->oldest_modification());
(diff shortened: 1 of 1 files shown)


vlad-lesin (approved):
LGTM

dr-m:
Sorry, this still has some freezes. I posted to #2930 an idea that is rather obvious if it works.

storage/innobase/trx/trx0purge.cc
```diff
@@ -707,7 +707,16 @@ TRANSACTIONAL_TARGET void trx_purge_truncate_history()
...
+          {
+            /* The functions buf_pool_t::release_freed_page() or
+            buf_do_flush_list_batch() may be right now holding
+            buf_pool.mutex and waiting to acquire
+            buf_pool.flush_list_mutex. Ensure that they can proceed,
+            to avoid extreme waits. */
+            mysql_mutex_lock(&buf_pool.mutex);
+            mysql_mutex_unlock(&buf_pool.mutex);
             bpage->lock.x_lock();
+          }
```
dr-m, line 719: This is an attempt to improve over 06280f233565fc030a406350a6f9b60e9c563d81 so that the undo tablespace truncation would not cause any prolonged blocking of the workload. In the first attempted run, we did not get any undo tablespace truncation triggered during the workload. During the shutdown, it happened, and took extremely long time (while ignoring `innodb_fast_shutdown=2`). I suspect that adding this wait could seriously increase the probability of `goto rescan`. That is, the linear scan of `buf_pool.flush_list` may degrade into a quadratic scan. A longer benchmark run is needed to determine if this is an improvement or something to be avoided.

dr-m: The newest update 98287bd2d6877969514469ebfbf27f9951b99657 is the same as 06280f233565fc030a406350a6f9b60e9c563d81, just rebased on the current 10.6 branch.

vlad-lesin:
The commit 98287bd looks good.

PR #2935, opened 2023-12-13
MDEV-26923 Check all invalid config options

tonychen2001:
* [x] _The Jira issue number for this PR is: MDEV-26923_

## Description

This change makes the effort to reduce the number of server restarts required due to passing invalid options.

Previously, the behavior was to error out on the first invalid option encountered. With this change, a best effort approach is made so that all invalid options processed will be printed before exiting.

There is a caveat. The options are processed many times at varying stages of server startup because the server is not aware of all valid options immediately (e.g. plugins have to be loaded first before the server knows what are the available plugin options). So, there are some options that the server can determine are invalid "early" on, and there are some options that the server cannot determine are invalid until "later" on. For example, the server can determine an option such as `--a` is an ambiguous option very early on but an option such as `--this-does-not-match-any-option` cannot be labelled as invalid until the server is aware of all available options.

Thus, it is possible that the server will still fail before printing out all "invalid" options. You can see this by passing `--a --obvious-invalid-option`.

## How can this PR be tested?

MTR test main.mysqld_option_err has been updated with a new test which passes

* [ ] _This is a new feature and the PR is based against the latest MariaDB development branch._
* [x] _This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced._

## PR quality check

* [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.

# Copyright

### DIFF
--- a/mysql-test/main/mysqld_option_err.test
+++ b/mysql-test/main/mysqld_option_err.test
@@ -46,6 +46,18 @@ mkdir $MYSQLTEST_VARDIR/tmp/mysqld_option_err;
 --error 7
 --exec $MYSQLD_BOOTSTRAP_CMD --skip-networking --datadir=$MYSQLTEST_VARDIR/tmp/mysqld_option_err --skip-grant-tables --plugin-dir=$MYSQLTEST_VARDIR/plugins --plugin-load=example=ha_example.so --plugin-example-enum-var=noexist >>$MYSQLTEST_VARDIR/tmp/mysqld_option_err/mysqltest.log 2>&1
 
+--echo Test to see if multiple unknown options will be displayed in the error output
+# Remove the noise to make the test robust
+--replace_regex /^((?!nonexistent).)*$// /.*unknown/unknown/
+--error 7
+--exec $MYSQLD_BOOTSTRAP_CMD --skip-networking --datadir=$MYSQLTEST_VARDIR/tmp/mysqld_option_err --skip-grant-tables --nonexistentoption --alsononexistent --nonexistentvariable=1 2>&1
+
+--echo Test to see if multiple ambiguous options and invalid arguments will be displayed in the error output
+# Remove the noise to make the test robust
+--replace_regex /^((?!('sql_mode'|'--character'|'--bootstrap')).)*$// /.*Error while setting value/Error while setting value/ /.*ambiguous option/ambiguous option/ /.*option '--bootstrap'/option '--bootstrap'/
+--error 1
+--exec $MYSQLD_BOOTSTRAP_CMD --skip-networking --datadir=$MYSQLTEST_VARDIR/tmp/mysqld_option_err --skip-grant-tables --getopt-prefix-matching --sql-mode=invalid_value --character --bootstrap=partstoob 2>&1
+
 #
 # Test that an wrong option with --help --verbose gives an error
 #
--- a/mysys/my_getopt.c
+++ b/mysys/my_getopt.c
@@ -133,6 +133,8 @@ double getopt_ulonglong2double(ulonglong v)
   return u.dbl;
 }
 
+#define SET_HO_ERROR_AND_CONTINUE(e) { ho_error= (e); continue; }
+
 /**
   Handle command line options.
   Sort options.
@@ -202,7 +204,7 @@ int handle_options(int *argc, char ***argv, const struct my_option *longopts,
   const char *UNINIT_VAR(prev_found);
   const struct my_option *optp;
   void *value;
-  int error, i;
+  int ho_error= 0, error, i;
   my_bool is_cmdline_arg= 1;
   DBUG_ENTER("handle_options");
 
@@ -216,7 +218,7 @@ int handle_options(int *argc, char ***argv, const struct my_option *longopts,
 
   is_cmdline_arg= !is_file_marker(**argv);
 
-  for (pos= *argv, pos_end=pos+ *argc; pos != pos_end ; pos++)
+  for (pos= *argv, pos_end=pos+ *argc; pos < pos_end ; pos++)
   {
     char **first= pos;
     char *cur_arg= *pos;
@@ -305,7 +307,7 @@ int handle_options(int *argc, char ***argv, const struct my_option *longopts,
                                                my_progname, special_opt_prefix[i],
                                                opt_str, special_opt_prefix[i],
                                                prev_found);
-		    DBUG_RETURN(EXIT_AMBIGUOUS_OPTION);
+		    SET_HO_ERROR_AND_CONTINUE(EXIT_AMBIGUOUS_OPTION)
 		  }
 		  switch (i) {
 		  case OPT_SKIP:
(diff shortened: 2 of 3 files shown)


LinuxJedi:
@tonychen2001 has officially requested to have his CLA signature revoked. This message is a temporary confirmation of that we consider it revoked for this PR onwards until we can get CLA assistant's data updated.

Tony: Please link to this PR in other PR until we can get this fixed so that others are aware.

mysql-test/main/mysqld_option_err.test
```diff
@@ -46,6 +46,11 @@ mkdir $MYSQLTEST_VARDIR/tmp/mysqld_option_err;
 --error 7
 --exec $MYSQLD_BOOTSTRAP_CMD --skip-networking --datadir=$MYSQLTEST_VARDIR/tmp/mysqld_option_err --skip-grant-tables --plugin-dir=$MYSQLTEST_VARDIR/plugins --plugin-load=example=ha_example.so --plugin-example-enum-var=noexist >>$MYSQLTEST_VARDIR/tmp/mysqld_option_err/mysqltest.log 2>&1
 
+--echo Test to see if multiple invalid options will be displayed in the error output
+--replace_regex /^((?!nonexistent).)*$// /.*unknown/unknown/ 
+--error 2
+--exec $MYSQLD_BOOTSTRAP_CMD --skip-networking --datadir=$MYSQLTEST_VARDIR/tmp/mysqld_option_err --skip-grant-tables --nonexistentoption --alsononexistent --nonexistentvariable=1 2>&1
+
```
vuvova, line 53: what about `--sql-mode=invalid_value` ?

vuvova: what about ambiguous options, like `--character` ?

vuvova: generally, invalid arguments, like `--bootstrap=partstoob` or `--table-cache=enabled` ?

tonychen2001: Trying to start mariadbd with `--sql-mode=invalid_value --character --bootstrap=partstoob --table-cache=enabled`:
```
./build/sql/mariadbd --datadir=./data --user=mariadb --socket=/tmp/mysqld.sock --sql-mode=invalid_value --character --bootstrap=partstoob --table-open-cache=enabled
2023-12-18 16:22:20 0 [ERROR] ./build/sql/mariadbd: Error while setting value 'invalid_value' to 'sql_mode'
2023-12-18 16:22:20 0 [ERROR] ./build/sql/mariadbd: ambiguous option '--character' (character-set-client-handshake, character_sets_dir)
2023-12-18 16:22:20 0 [ERROR] ./build/sql/mariadbd: option '--bootstrap' cannot take an argument
2023-12-18 16:22:20 0 [Warning] option 'table_open_cache': unsigned value 0 adjusted to 10
```

I noticed that `--table-cache=enabled` does not prevent the server from starting up (even without my changes).

tonychen2001: One thing to point out (I mentioned it in the commit body and you also commented on this in [MDEV-26922](https://jira.mariadb.org/browse/MDEV-26922)) is that some options can be determined by the server as invalid "early" on and some options cannot be determined by the server as invalid until "later" on.

For example, ambiguous options like --character and invalid arguments like --bootstrap=partstoob can be determined as invalid by the server immediately.
On the other hand, the invalidity of unknown options like --nonexistentoption cannot be determined until other actions have taken place such as loading of all plugins.

So with these changes, it's still possible that the server does not print all invalid options. For example, `--nonexistent` is not printed here:
```
./build/sql/mariadbd --datadir=./data --user=mariadb --socket=/tmp/mysqld.sock --sql-mode=invalid_value --character --bootstrap=partstoob --nonexistentoption
2023-12-18 16:34:04 0 [ERROR] ./build/sql/mariadbd: Error while setting value 'invalid_value' to 'sql_mode'
2023-12-18 16:34:04 0 [ERROR] ./build/sql/mariadbd: ambiguous option '--character' (character-set-client-handshake, character_sets_dir)
2023-12-18 16:34:04 0 [ERROR] ./build/sql/mariadbd: option '--bootstrap' cannot take an argument
```

I went with this approach because it stays in line as much as possible with the current behavior. Also, if we want the server to display all invalid options at the same time, this means that sometimes the server will error out later than it could have.

tonychen2001: I can add another test case to `mysqld_option_err.test` to verify the behaviour on `--sql-mode=invalid_value --character --bootstrap=partstoob`.

Please let me know your thoughts on my above comment. Thanks!

vuvova: Yes, the test would be great, that was my point — to add a test for `--sql-mode=invalid_value --character --bootstrap=partstoob`. `handle_options()` returns an error in many different places, would be good to cover if not all of them, then at least most.

About "early" and "late" — I would think the current behavior is good enough. It's not a feature where a user needs 100% correctness, but just a helper to report as many errors as possible. You cannot report all errors correctly in all cases anyway — e.g. a missing value for an option will make my_getopt to swallow next option as a value and then everything will be out of sync, to resolve that you'd need to know user's intentions (like in `--aa --bb --cc` — are they options without values? or `--aa` with a value of `--bb` and `--cc` without? or `--aa` without a value and `--bb` with?). So, being "good enough", I think, should be the goal here.

tonychen2001: Agreed.

Test for `--sql-mode=invalid_value --character --bootstrap=partstoob` is added.

LinuxJedi (approved):
I think this looks good to me. I leave the final decision to @vuvova. Likely it would be 10.5 this would need to land in now due to 10.4 being near EoL and the final release is typically reserved for critical / secutiry fixes.

mysys/my_getopt.c
```diff
@@ -139,6 +139,19 @@ double getopt_ulonglong2double(ulonglong v)
...
+  variable/option.  The error code is determined by the first invalid
+  configuration parsed.
+*/
+static void set_ho_error(int *ho_error, int err)
+{
+  if (!*ho_error && err)
+    *ho_error= err;
+}
+
+#define SET_HO_ERROR_AND_CONTINUE(e) { set_ho_error(&ho_error, (e)); continue; }
```
vuvova, line 153: it'd be simpler without `set_ho_error()` function, you don't use it outside of the macro anyway. And it'd be even simpler if you remove the check for `err!=0` (it cannot happen anyway, one doesn't `continue` without an error), and if you return the last error, not the first — it doesn't really matter what error to return, right?

Then it'll be just
```c
#define SET_HO_ERROR_AND_CONTINUE(e) { ho_error=(e); continue; }
```
10x less code :)

vuvova (approved):
it's ok, but I'd suggest to simplify the implementation a bit, see the comment inline

mysys/my_getopt.c
tonychen2001: Agreed, it doesn't matter if we return the first or the last error code.

I removed the checks and `set_ho_error()`.

tonychen2001:
I rebased this PR onto `10.5`

PR #2937, opened 2023-12-14
sql/sql_test.cc include header correction

grooverdan:

## Description

While 1fec50120fb4b08ea76cc79cf65905a0d5027e84 (#2880 was right, when merged to 10.5, the MacOS include header
(9a3cbc05415dd862852dffbd368527319107f460)
was included under the #ifdef HAVE_MALLINFO{,2} which isn't the case on MacOS.

sys/malloc.h possibly has use under OpenBSD from MDEV-4190 patch so keep it separate. C89 doesn't mention the need however and malloc only needs stdlib.h.

## How can this PR be tested?

Debug builds cover this.

## PR quality check

### DIFF
--- a/sql/sql_test.cc
+++ b/sql/sql_test.cc
@@ -27,12 +27,12 @@
 #include "my_json_writer.h"
 #include <hash.h>
 #include <thr_alarm.h>
-#if defined(HAVE_MALLINFO) || defined(HAVE_MALLINFO2)
-#if defined(HAVE_MALLOC_H)
+#if (defined(HAVE_MALLINFO) || defined(HAVE_MALLINFO2)) && defined(HAVE_MALLOC_H)
 #include <malloc.h>
-#elif defined(HAVE_SYS_MALLOC_H)
-#include <sys/malloc.h>
 #endif
+
+#if defined(HAVE_SYS_MALLOC_H)
+#include <sys/malloc.h>
 #endif
 
 #ifdef HAVE_EVENT_SCHEDULER


dmitryshulga (changes requested):
Proposed patch doesn't fix the issue.
Why you don't borrow the same code snippet as in 10.6 and upper (see below) ?

#if defined(HAVE_MALLINFO) && defined(HAVE_MALLOC_H)
#include <malloc.h>
#elif defined(HAVE_MALLINFO) && defined(HAVE_SYS_MALLOC_H)
#include <sys/malloc.h>
#elif defined(HAVE_MALLOC_ZONE)
#include <malloc/malloc.h>

grooverdan:
> Proposed patch doesn't fix the issue.

https://github.com/MariaDB/server/pull/2938/files#diff-e5bce400387ad6db934dff6498b2eb9ff98f61237bf68977679367ef41a079cb does in 10.5 which is the main fix.

> Why you don't borrow the same code snippet as in 10.6 and upper (see below) ?
> 
> #if defined(HAVE_MALLINFO) && defined(HAVE_MALLOC_H) #include <malloc.h> #elif defined(HAVE_MALLINFO) && defined(HAVE_SYS_MALLOC_H) #include <sys/malloc.h> #elif defined(HAVE_MALLOC_ZONE) #include <malloc/malloc.h>

Because its incorrect. Both `mallinfo` and `mallinfo2` need malloc.h. `sys/malloc.h` isn't needed for these, might be needed on OpenBSD, I have doubt, but haven't got a test environment to prove its not.

dmitryshulga:
> > Why you don't borrow the same code snippet as in 10.6 and upper (see below) ?
> > #if defined(HAVE_MALLINFO) && defined(HAVE_MALLOC_H) #include <malloc.h> #elif defined(HAVE_MALLINFO) && defined(HAVE_SYS_MALLOC_H) #include <sys/malloc.h> #elif defined(HAVE_MALLOC_ZONE) #include <malloc/malloc.h>
> 
> Because its incorrect. Both `mallinfo` and `mallinfo2` need malloc.h. `sys/malloc.h` isn't needed for these, might be needed on OpenBSD, I have doubt, but haven't got a test environment to prove its not.
> 
> Its only a a produce of backward compatible glibc that ensures the existence of mallinfo2 is there when mallinfo is available, otherwise it would include the wrong header. This is the assumption that [1fec501](https://github.com/MariaDB/server/commit/1fec50120fb4b08ea76cc79cf65905a0d5027e84) was avoiding, it just got merged up incorrectly.

If we speak about correctness, the original block of the code is rather incorrect than right since it mixes tests for presence of library calls and header files, and as a consequence such approach makes implicit dependency between presence of header files and library api, that generally is not correct.

Correct way would be to use only macros HAVE_MALLOC_H, HAVE_SYS_MALLOC_H, HAVE_MALLOC_MALLOC_H (the last macro definition could be used for checking presence of the file malloc/malloc.h if such directive be added into configure.cmake)

To summary, proposed patch doesn't solve the original issue with successful build of mariadb server on macos, instead it adds some kludge that at least doesn't make the code better.

So, once again I suggest using the code from 10.6. At least it works.

grooverdan:
1fec50120fb4b08ea76cc79cf65905a0d5027e84 was a fix, for devs that upgrade their glibc version. Its now merged to 10.6.

#2952 is the minimal fix.

PR #2939, opened 2023-12-15
Fix unaligned access in my_dir

labath:
Accessing data through an unaligned pointer is undefined behavior and can cause crashes certain architectures (e.g. arm). Use a union to force the compiler to allocate storage suitably aligned for a `struct dirent`.

unaligned access

No functional change (apart from not crashing when accessing an unaligned pointer)

No side effects.

## How can this PR be tested?

> If the changes are not amenable to automated testing, please explain why not and carefully describe how to test manually.

Compile with undefined behavior sanitizer, check that it does not crash. Automated testing is difficult, since it's not even guaranteed that variable will be actually unaligned.

## PR quality check

### DIFF
--- a/mysys/my_lib.c
+++ b/mysys/my_lib.c
@@ -115,7 +115,10 @@ MY_DIR	*my_dir(const char *path, myf MyFlags)
   DIR		*dirp;
   struct dirent *dp;
   char		tmp_path[FN_REFLEN + 2], *tmp_file;
-  char	dirent_tmp[sizeof(struct dirent)+_POSIX_PATH_MAX+1];
+  union {
+    char storage[sizeof(struct dirent)+_POSIX_PATH_MAX+1];
+    struct dirent d;
+  } dirent_tmp;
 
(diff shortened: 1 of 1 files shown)


dr-m:
The oldest maintained branch is 10.4. Can you please rebase this?

labath:
> The oldest maintained branch is 10.4. Can you please rebase this?

I'm afraid my employer does not allow me to sign the CLA, so I'll have to withdraw this pull request. I'll file a bug instead. :)

FWIW, 10.3 was cited in the pull request template as the oldest maintained branch. I guess the template needs an update (?)

vuvova:
You don't have to sign a CLA, it's just one of the two options. You can contribute under the new BDS (3-clause) license. See

https://mariadb.com/kb/en/mariadb-contributor-agreement-faq/#when-should-i-use-mca-and-when-should-i-use-bsd-new

labath:
Thank you for that. Let me just confirm that with my employer.

PR #2941, opened 2023-12-18
Update my_print_defaults man page with mariadbd option

iangilfillan:
The `--mariadbd` option was added in 10.11.3, this commit adds the option
to the man page as well.


illuusio:
Good addition to man page. Have you tested output?

iangilfillan:
Yes, output looks fine

grooverdan (approved):

PR #2949, opened 2023-12-19
MDEV-33053 InnoDB LRU flushing does not run before running out of buffer pool

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33053*

## Description
When using `innodb_buffer_pool_size=2000g` and `innodb_log_file_size=128g` it could happen that InnoDB will be stalled for less than a second while the buffer pool is running out of free pages. Setting a smaller `innodb_max_dirty_pages_pct` or (nonzero) `innodb_max_dirty_pages_pct_lwm` will not necessarily keep the `buf_flush_page_cleaner()` thread busy.

`buf_pool_t::page_cleaner_wakeup()`: Change the parameter `bool for_LRU` to an integer, to distinguish immediate wakeup and a "soft" wakeup (no need to interrupt a 1-second sleep).

`buf_LRU_get_free_block()`: After `buf_LRU_get_free_only()` succeeded, invoke `buf_pool_t::page_cleaner_wakeup()` if the number of free blocks is below `innodb_lru_scan_depth`. If it is below half that, wake up the page cleaner immediately, not letting it wait for the 1-second interval to elapse.

## How can this PR be tested?
This is best tested in the environment where this problem was reported. I suspect that the workload is rather read-heavy.

It is unlikely that this change will break anything. I think that the normal RQG stress tests (which include the use of tiny buffer pools) should cover this reasonably well.

The `for_LRU` parameter in `buf_pool_t::page_cleaner_wakeup()` was added in a55b951e6082a4ce9a1f2ed5ee176ea7dbbaf1f2 (10.6).

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -1797,6 +1797,28 @@ ulint buf_flush_LRU(ulint max_n, bool evict)
     buf_pool.try_LRU_scan= true;
     pthread_cond_broadcast(&buf_pool.done_free);
   }
+  else if (!pages && !buf_pool.try_LRU_scan &&
+           buf_pool.LRU_warned.test_and_set(std::memory_order_acquire))
+  {
+    /* For example, with the minimum innodb_buffer_pool_size=5M and
+    the default innodb_page_size=16k there are only a little over 316
+    pages in the buffer pool. The buffer pool can easily be exhausted
+    by a workload of some dozen concurrent connections. The system could
+    reach a deadlock like the following:
+
+    (1) Many threads are waiting in buf_LRU_get_free_block()
+    for buf_pool.done_free.
+    (2) Some threads are waiting for a page latch which is held by
+    another thread that is waiting in buf_LRU_get_free_block().
+    (3) This thread is the only one that could make progress, but
+    we fail to do so because all the pages that we scanned are
+    buffer-fixed or latched by some thread. */
+    sql_print_warning("InnoDB: Could not free any blocks in the buffer pool!"
+                      " %zu blocks are in use and %zu free."
+                      " Consider increasing innodb_buffer_pool_size.",
+                      UT_LIST_GET_LEN(buf_pool.LRU),
+                      UT_LIST_GET_LEN(buf_pool.free));
+  }
 
   return pages;
 }
@@ -2287,6 +2309,16 @@ func_exit:
 	goto func_exit;
 }
 
+TPOOL_SUPPRESS_TSAN
+bool buf_pool_t::need_LRU_eviction() const
+{
+  /* try_LRU_scan==false means that buf_LRU_get_free_block() is waiting
+  for buf_flush_page_cleaner() to evict some blocks */
+  return UNIV_UNLIKELY(!try_LRU_scan ||
+                       (UT_LIST_GET_LEN(LRU) > BUF_LRU_MIN_LEN &&
+                        UT_LIST_GET_LEN(free) < srv_LRU_scan_depth / 2));
+}
+
 /** page_cleaner thread tasked with flushing dirty pages from the buffer
 pools. As of now we'll have only one coordinator. */
 static void buf_flush_page_cleaner()
@@ -2319,21 +2351,24 @@ static void buf_flush_page_cleaner()
     }
 
     mysql_mutex_lock(&buf_pool.flush_list_mutex);
-    if (buf_pool.ran_out())
-      goto no_wait;
-    else if (srv_shutdown_state > SRV_SHUTDOWN_INITIATED)
-      break;
+    if (!buf_pool.need_LRU_eviction())
+    {
+      if (srv_shutdown_state > SRV_SHUTDOWN_INITIATED)
+        break;
 
-    if (buf_pool.page_cleaner_idle() &&
-        (!UT_LIST_GET_LEN(buf_pool.flush_list) ||
-         srv_max_dirty_pages_pct_lwm == 0.0))
-      /* We are idle; wait for buf_pool.page_cleaner_wakeup() */
-      my_cond_wait(&buf_pool.do_flush_list,
-                   &buf_pool.flush_list_mutex.m_mutex);
-    else
-      my_cond_timedwait(&buf_pool.do_flush_list,
-                        &buf_pool.flush_list_mutex.m_mutex, &abstime);
-  no_wait:
+      if (buf_pool.page_cleaner_idle() &&
+          (!UT_LIST_GET_LEN(buf_pool.flush_list) ||
+           srv_max_dirty_pages_pct_lwm == 0.0))
+      {
+        buf_pool.LRU_warned.clear(std::memory_order_release);
+        /* We are idle; wait for buf_pool.page_cleaner_wakeup() */
+        my_cond_wait(&buf_pool.do_flush_list,
+                     &buf_pool.flush_list_mutex.m_mutex);
+      }
+      else
+        my_cond_timedwait(&buf_pool.do_flush_list,
+                          &buf_pool.flush_list_mutex.m_mutex, &abstime);
+    }
     set_timespec(abstime, 1);
 
     lsn_limit= buf_flush_sync_lsn;
@@ -2365,7 +2400,7 @@ static void buf_flush_page_cleaner()
       }
       while (false);
 
-      if (!buf_pool.ran_out())
+      if (!buf_pool.need_LRU_eviction())
         continue;
       mysql_mutex_lock(&buf_pool.flush_list_mutex);
       oldest_lsn= buf_pool.get_oldest_modification(0);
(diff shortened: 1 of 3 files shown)


vaintroub:
Please create an enum, instead  "int forLRU", otherwise one needs to guess what those 0,1,-1 values mean. With bool, meaning of forLRU was understandable, but now it is not.

storage/innobase/buf/buf0lru.cc
```diff
@@ -417,9 +417,20 @@ buf_block_t *buf_LRU_get_free_block(bool have_mutex)
...
+		const ulint scan_depth = srv_LRU_scan_depth;
+
 		if (!have_mutex) {
 			mysql_mutex_unlock(&buf_pool.mutex);
 		}
+
+		if (UNIV_UNLIKELY(available < scan_depth)) {
+			mysql_mutex_lock(&buf_pool.flush_list_mutex);
+			buf_pool.page_cleaner_wakeup(available < scan_depth / 2
+						     ? -1 : 1);
```
vaintroub, line 430: enum rather than -1:1

storage/innobase/buf/buf0lru.cc
```diff
@@ -449,7 +460,7 @@ buf_block_t *buf_LRU_get_free_block(bool have_mutex)
 		mysql_mutex_lock(&buf_pool.flush_list_mutex);
 		const auto n_flush = buf_pool.n_flush();
 		if (!buf_pool.try_LRU_scan) {
-			buf_pool.page_cleaner_wakeup(true);
+			buf_pool.page_cleaner_wakeup(-1);
```
vaintroub, line 463: enum value, not -1

storage/innobase/include/buf0buf.h
```diff
@@ -1781,7 +1781,7 @@ class buf_pool_t
 
   /** Wake up the page cleaner if needed.
   @param for_LRU  whether to wake up for LRU eviction */
-  void page_cleaner_wakeup(bool for_LRU= false);
+  void page_cleaner_wakeup(int for_LRU= 0);
```
vaintroub, line 1784: enum here, too

storage/innobase/buf/buf0flu.cc
```diff
@@ -112,11 +112,17 @@ static void buf_flush_validate_skip()
...
+    /* Many ISA provide N and Z flags, which we desire to use for
+    efficient checking of the wakeup reason. */
+    static_assert(WAKE_NOW_LRU < 0, "efficiency");
+    static_assert(WAKE_IDLE == 0, "efficiency");
+    static_assert(WAKE_IDLE_LRU > 0, "efficiency");
+
+    if (reason < WAKE_IDLE)
       /* Ensure that the page cleaner is not in a timed wait. */
       pthread_cond_signal(&do_flush_list);
     return;
```
dr-m, line 128: For the record, retaining the equivalent condition `reason != WAKE_IDLE` would result in roughly the same maximum latency in my 160-thread `oltp_read_only` test on 50M buffer pool, 1.2G data, 100M log, run on RAM disk. The maximum latency would be about 48ms.

That is, this change of the parameter from `bool` to `enum` might not be useful.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2394,7 +2428,7 @@ static void buf_flush_page_cleaner()
       if (oldest_lsn >= soft_lsn_limit)
         buf_flush_async_lsn= soft_lsn_limit= 0;
     }
-    else if (buf_pool.ran_out())
+    else if (buf_pool.need_LRU_eviction())
```
mariadb-DebarunBanerjee, line 2431: Not related to this patch but just thought of mentioning as it could affect. 

The code below has unconditional loop over flush list. This could be expensive when the buffer pool is large and expected number of dirty pages in the system is very large. Since this patch would lead to more number of LRU flush it could start impacting.

mariadb-DebarunBanerjee (commented):
Hi Marko,
Thanks for implementing the final patch. The high level concern is possible regression due to more frequent LRU flush but let's depend on our performance tests for that.

Some comments for your consideration.

Regards,
Deb

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: The code I am referring to. Looks like some clean-up by unconditionally looping over flush list. 
```c
/* Remove clean blocks from buf_pool.flush_list before the LRU scan. */
      for (buf_page_t *p= UT_LIST_GET_FIRST(buf_pool.flush_list); p; )
      {
        const lsn_t lsn{p->oldest_modification()};
        ut_ad(lsn > 2 || lsn == 1);
        buf_page_t *n= UT_LIST_GET_NEXT(list, p);
        if (lsn <= 1)
          buf_pool.delete_from_flush_list(p);
        p= n;
      }
```

storage/innobase/buf/buf0lru.cc
```diff
@@ -417,9 +417,19 @@ buf_block_t *buf_LRU_get_free_block(bool have_mutex)
...
 got_block:
+		const ulint available = UT_LIST_GET_LEN(buf_pool.free);
+		const ulint scan_depth = srv_LRU_scan_depth;
+
 		if (!have_mutex) {
 			mysql_mutex_unlock(&buf_pool.mutex);
 		}
+
+		if (UNIV_UNLIKELY(available < scan_depth)) {
+			mysql_mutex_lock(&buf_pool.flush_list_mutex);
```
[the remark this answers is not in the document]
dr-m: At the moment, `buf_pool.page_cleaner_status` is a non-atomic `unsigned` variable that is protected by `buf_pool.flush_list_mutex`. It looks like we could change it to `Atomic_relaxed<unsigned>` without too much trouble. All `+=` and `-=` have to be replaced with `=` so that they will use relaxed loads and stores instead of using atomic instructions, such as the 80486 `lock xadd`.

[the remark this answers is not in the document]
dr-m: Changing the variable to `Atomic_relaxed<unsigned>` is not trivial. It may result in some overhead in code like this:
```c++
      buf_pool.page_cleaner_set_idle(false);
      buf_pool.n_flush_inc();
```
With a regular `unsigned` variable, the compiler is allowed to generate code that only does one load and one store for the above. For `Atomic_relaxed<unsigned>`, it would have to emit instructions that access the memory bus several times. I think that this is too risky to do without careful analysis of the generated code.

storage/innobase/buf/buf0lru.cc
```diff
@@ -417,9 +417,19 @@ buf_block_t *buf_LRU_get_free_block(bool have_mutex)
...
+			buf_pool.page_cleaner_wakeup(true);
```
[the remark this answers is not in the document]
dr-m: I think that for now, it is better to just add a condition `!buf_pool.page_cleaner_active()` before both calls to `buf_pool.page_cleaner_wakeup(true)` in `buf_LRU_get_free_block()`.

storage/innobase/buf/buf0flu.cc
dr-m: That scan was added in 9d1466522ea92963ac6ca16b597392714280c9f1 in order to prevent a race condition at the end of crash recovery. This is also related to 5b53342a6a59ae1141b2a46467954fc3836f8e80. Ever since 22b62edaedddb1cabd5b855cdd39a5e90a5695a2 we do not remove blocks from `buf_pool.flush_list` on `buf_page_write_complete()`, but only update the `oldest_modification` to the special value 1 that denotes that the page is actually clean. In this way, we significantly reduced contention on the `buf_pool.flush_list_mutex`.

storage/innobase/buf/buf0flu.cc
```diff
@@ -1797,6 +1798,30 @@ ulint buf_flush_LRU(ulint max_n, bool evict)
...
+
+    (1) Many threads are waiting in buf_LRU_get_free_block()
+    for buf_pool.done_free.
+    (2) Some threads are waiting for a page latch which is held by
+    another thread that is waiting in buf_LRU_get_free_block().
+    (3) This thread is the only one that could make progress, but
+    we fail to do so because all the pages that we scanned are
+    buffer-fixed or latched by some thread. */
+
+    buf_lru_free_blocks_error_printed= true;
```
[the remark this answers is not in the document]
dr-m: I’ll change this to `std::atomic_flag`. The safest place for `buf_pool.LRU_warned.clear()` is right before the `my_cond_wait()` in `buf_flush_page_cleaner()`.

storage/innobase/buf/buf0lru.cc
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Agree. I think we can evaluate more in future.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: ok, thanks for explaining. I will go through the optimization sometime.

mariadb-DebarunBanerjee (approved):
Looks good to me. Let's hope the tests go well.

vaintroub (approved):

storage/innobase/buf/buf0flu.cc
dr-m: The loop that you identified looks redundant and does now cause a significant regression. #3078 was filed for the removal.

PR #2953, opened 2023-12-20
MDEV-33049 Assertion `marked_for_write_or_computed()' failed in bool Field_new_decimal::store_value(const my_decimal*, int*)

SongLibing:
Analysis
========
When rpl applier is unpacking a before row image, Field::reset() will be
called before setting a field to null if null bit of the field is set in
the row image. For Field_new_decimal::reset(), it calls
Field_new_decimal::store_value() to reset the value. store_value() asserts
that the field is in the write_set bitmap since it thinks the field is
updating.

But that is not true for the row image generated in FULL_NODUP
mode. In the mode, the before image includes all fields and the after
image includes only updated fields.

Fix
===
In the case unpacking binlog row images, the assertion is meaningless.
So the unpacking field is marked in write_set temporarily to avoid the
assertion failure.

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check


sql/field.cc
```diff
@@ -75,6 +75,14 @@ bool Field::marked_for_read() const
 
 bool Field::marked_for_write_or_computed() const
 {
+  /*
+    Field_new_decimal::store_value() will be called if the field is null when
+    binlog applier is unpacking a row image. In that case, write_set should not
+    be checked.
+  */
+  if (get_thd() && get_thd()->lex->is_stmt_row_injection())
+    return true;
```
andrelkin, line 84: I think I got the idea of this hunk, let me share with what I understood.
At this point the rbr applier  executes a stack of 
```
Field::marked_for_write_or_computed
Field_new_decimal::store_value
Field_new_decimal::store_value
Field_new_decimal::reset
unpack_field
unpack_row
Rows_log_event::unpack_current_row
Rows_log_event::write_row
Write_rows_log_event::do_exec_row
Rows_log_event::do_apply_event
``` 
`Field_new_decimal::reset` does not need any bitmap as it just wipes of any old content. And it looks to be me that the above hunk assumes `Field::marked_for_write_or_computed` is indeed invoked necessarily from the `unpack_row` stack.
But can `Field_new_decimal::store_value -> Field::marked_for_write_or_computed`  be also reached later, that is after the resetting, e.g [here](https://github.com/MariaDB/server/blob/11.4/sql/rpl_record.cc#L232)?

SongLibing: No, for applier row event Field::unpack() is called to store the value if it is not null. Field::store_value() is called only for reset() and it happens only for Field_decimal and Field_new_decimal, other type of fields don't call ::store_value() in reset().

mysql-test/suite/rpl/t/rpl_unpack_new_decimal_crash.test
```diff
@@ -0,0 +1,39 @@
...
+--source include/rpl_sync.inc
+
+FLUSH BINARY LOGS;
+UPDATE t1 SET c2 = "c2_new";
+
+# BINLOG statement is same to applier, it should work well.
+--exec $MYSQL_BINLOG --start-position=$_master_pos $datadir$_master_file > $MYSQLTEST_VARDIR/tmp/binlog.sql
+--exec $MYSQL < $MYSQLTEST_VARDIR/tmp/binlog.sql
+
+# c2's value should be "c2"
```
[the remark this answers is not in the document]
SongLibing: fixed

sql/field.cc
andrelkin: Thanks for correcting me, LiBing!
After thinking more and brief discussion with Bar I think it's safer to tackle assert in the replication code,
which I propose to in [a review commit](https://github.com/MariaDB/server/blob/review-11.4-MDEV-33049/sql/rpl_record.cc#L205) on top of yours.
Please check that out.

andrelkin: @FooBarrior, let me also invite you as this bug is certainly about ` 3a42f2869e80` changes. 
Maybe you'll have the best of solutions :-)! 
Thank you at once!

SongLibing: Hi Andrei, fixed as your suggestion.

andrelkin (approved):
Thanks LiBing, looks really good!

mysql-test/suite/rpl/t/rpl_unpack_new_decimal_crash.test
```diff
@@ -0,0 +1,40 @@
+################################################################################
+# MDEV-33049 Assertion `marked_for_write_or_computed()' failed in bool
+#            Field_new_decimal::store_value(const my_decimal*, int*)
+#
+# In FULL_NODUP mode, the before image has all fields and the after image has
+# only updated fields. Crash happened when slave was unpacking a row event
+# if the new decimal field is null in the before image.
+################################################################################
+--source include/have_binlog_format_row.inc
+--source include/master-slave.inc
```
FooBarrior, line 10: Adding a new test adds a notable extra execution time. Also [de]initializing master-slave turns to be extra slow. If possible, please add this test somewhere in the existing test file related to the row-based replication.  `binlog_row_image` can be configured for a test case (inside a test file) just fine.

The good place would be one of `rpl/t/rpl_row_*.test` or `rpl/include/rpl_row_*.test` files

FooBarrior (changes requested):
Hi all and thank you for inviting me here, @andrelkin.

I went though your solutions and thought for some time. 

Yes, the second solution with tackling down the assert is better. I would also want us to commit more into the maintainability of our solutions. For the long time we are accruing these bitmap overriding rules and make it more and more complicated. Each such override adds two use-cases for the bitmap usage. If one override is met in the call chain after another one, then the complexities multiply before adding.

I think we should reduce amount of these cases. Here, in particular, I think it should be safe to call `Field::reset` for read. So `Field::reset` could instead contain this assertion. Or even better, just call the functions that don't have these assertion.

Take a look at the patch **bd1414417** made for this review.

I actually think that it should be the correct interface resolution: `reset` should never call `store_value`, which triggers the assertion. Even better would be to create a set of `store_*` functions that are fine to be called for read (like for selects, etc). But that's a different task.

Also, there are more questions to this `reset` function, i.e. why would it do _any_ stores, as its only purpose is to, basically, make the field ready for write. Nothing more than resetting the blob pointers shouldn't be done in fact. I'd add marking as uninitialized for memory sanitizers. And then it should return just void - its return result is never checked anyway (Sic!). Also another refactoring, of course. Just created [MDEV-33135](https://jira.mariadb.org/browse/MDEV-33135).

As for this PR -- accept the whatever solution you prefer, @SongLibing. Just please:
* Use existing test file if possible. Or make a more general one (for example, for all FULL_NODUP tests, but then I'd better make some existsng MINIMAL or NOBLOB test and transform it to run in different row format combinations)
* If you choose your current solution, please add a note in [MDEV-33135](https://jira.mariadb.org/browse/MDEV-33135) about that it should be removed as part of the refactoring TODO.

FooBarrior:
Maybe you may add an assertion for `marked_for_read()` BTW

mysql-test/suite/rpl/t/rpl_unpack_new_decimal_crash.test
SongLibing: Hi FooBarrior,
Thanks for your suggestions. I moved the test into rpl_row_img_eng_full_nodup.test.

LinuxJedi:
Hi @SongLibing, can you please squash these two commits into one? This means we can merge via rebase easily.

SongLibing:
> Hi @SongLibing, can you please squash these two commits into one? This means we can merge via rebase easily.
Hi @LinuxJedi I rebased the pull request branch.  There is no merge commit now.

LinuxJedi (approved):

PR #2955, opened 2023-12-20
MDEV-33067 Support creating snapshot based on SCN

poempeng:
What's SCN
==========
SCN is a sequence number which was generated while committing transaction, and it should always be increased. With SCN, the snapshot for MVCC can only use a version number to compare with SCN.
To make code simple, we use trx->no as SCN of transaction. During startup, the current trx id and no is initialized by max id stored int trx_sys page, while max_trx_id is multiple of 2 and max_scn is not, so we can distinguish transaction id and SCN stored in record field. For example, if id on startup is 101, then max_scn is 101 and max_trx_id is 102. the increasing step of trx id/scn is 2.

Get SCN
=======
While committing transaction, trx->no (aka SCN) is stored in undo header of transaction.
Before undo log for insert is immediately purged after committing transaction. In this commit, it'll keep insert undo not cleaned, so that user thread can get chance to read scn from undo header. The purge thread is resposible for removing all undo log. In each clust record, it stores roll_ptr which points to undo record in undo space. In order to get undo header address from undo log, we changed the format of undo record by adding extra 9 bytes in header of undo record:
-- 2 bytes: point to next record
-- 1 bytes: 0xFF, indicate it's new version of undo record (added) -- 2 bytes: low 2 bytes of trx id, for verification (added) -- 4 bytes: page number of undo header (added)
-- 2 bytes: offset of undo header inside undo hdr page (added) With this change we can find the undo hdr page, and get SCN number from it. To avoid too frequent reading of undo log, it implementted a scn map which stores trx_id->SCN, SCN_Map is a lock free structure. and it has two map:
- normal map: add id->scn to it while commtting transaction
- random map: add id->scn to it after extracting a scn from undo log while reading

WriteBack SCN
=============
While reader got record and check visibility, it'll read in scn of the record, the cursor will be stored in buf_block_t.
A group of background threads take cursor from buf_block_t and process them in background.

Snapshot
========
Snapshot now has three fields to implement MVCC:
* For fast checking, up_limit_id and low_limit_id is still preserved, but it's not always precise and just for fast checking.
* Max SCN value as its version number, all scn bigger than/equal to it is not visible.

Undo Purge
==========
We don't prevent purging undo log, even when related SCN is not written back to record.

New Variable
============
1. innodb_use_scn: use scn or not, default value is OFF, this variable is read only, all undo logs should be purged before changing it to different value.
2. innodb_cleanout_threads: the number of threads to write back scn.

Limitations:
============
1. Check table ... extended is not supported.
2. Undo log optimization of inserting into empty table is not supported.

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check


storage/innobase/trx/trx0trx.cc
```diff
@@ -725,9 +747,48 @@ dberr_t trx_lists_init_at_db_start()
 		return err;
 	}
 
+  if (srv_operation == SRV_OPERATION_NORMAL) {
+    if (trx_sys.is_start_from_scn() && !srv_use_scn) {
+      if (!trx_sys.is_undo_empty() || !purge_sys.purge_queue.empty()) {
+        ib::error()
+            << "Can't start with innodb_use_scn=OFF while undo is not empty.";
```
dr-m, line 754: We should prefer `sql_print_error()` for simple error messages.

storage/innobase/trx/trx0trx.cc
```diff
@@ -725,9 +747,48 @@ dberr_t trx_lists_init_at_db_start()
 		return err;
 	}
 
+  if (srv_operation == SRV_OPERATION_NORMAL) {
+    if (trx_sys.is_start_from_scn() && !srv_use_scn) {
+      if (!trx_sys.is_undo_empty() || !purge_sys.purge_queue.empty()) {
```
dr-m, line 752: Is the check `purge_sys.purge_queue.empty()` really needed?

storage/innobase/trx/trx0trx.cc
```diff
@@ -725,9 +747,48 @@ dberr_t trx_lists_init_at_db_start()
 		return err;
 	}
 
+  if (srv_operation == SRV_OPERATION_NORMAL) {
```
dr-m, line 750: How will this work with `mariadb-backup --prepare --export`?

mysql-test/suite/innodb/t/scn_on_and_off.test
```diff
@@ -0,0 +1,54 @@
...
+create table t1(id int, c1 int) engine=innodb;
+insert into t1 value(1,1);
+insert into t1 value(2,1);
+insert into t1 value(3,1);
+update t1 set c1=2 where id=1;
+update t1 set c1=2 where id=2;
+check table t1;
+--source include/wait_all_purged.inc
+
+--source include/shutdown_mysqld.inc
```
dr-m, line 16: Replace the first `--source` with
```sql
SET GLOBAL innodb_fast_shutdown=0;
```

mysql-test/suite/innodb/t/scn_on_and_off.test
```diff
@@ -0,0 +1,54 @@
...
+--let $restart_parameters= --innodb-use-scn=ON
+--source include/start_mysqld.inc
+show variables like '%use_scn%';
+
+insert into t1 values(4,1);
+update t1 set c1=2 where id=3;
+update t1 set c1=4 where id=3;
+check table t1;
```
dr-m, line 24: What would happen on `check table t1 extended`?

storage/innobase/srv/srv0srv.cc
```diff
@@ -1713,3 +1713,22 @@ void srv_purge_shutdown()
     srv_shutdown_purge_tasks();
   }
 }
+
+uint32_t srv_cleanout_threads = 8;
+my_bool srv_use_scn = FALSE;
```
dr-m, line 1718: The default values should already set via the data structure in `ha_innodb.cc`. We can let these variables live in the BSS. I’d like the variables for any new parameters to be named in the same way as the configuration parameter, e.g., `innodb_use_scn`.

storage/innobase/srv/srv0start.cc
```diff
@@ -1403,6 +1404,8 @@ dberr_t srv_start(bool create_new_db)
 		return(srv_init_abort(err));
 	}
 
+	scn_mgr = new SCN_Mgr();
```
dr-m, line 1407: Can we avoid the pointer indirection and declare `scn_mgr` as a singleton object, like we do for `dict_sys`, `lock_sys`, `buf_pool` and many others?

storage/innobase/dict/dict0load.cc
```diff
@@ -1897,6 +1897,7 @@ dict_load_index_low(
 	(*index)->page = mach_read_from_4(field);
 	ut_ad((*index)->page);
 	(*index)->merge_threshold = merge_threshold & ((1U << 6) - 1);
+	(*index)->trx_scn = 0;
```
dr-m, line 1900: `dict_mem_index_create()` will invoke `mem_heap_zalloc()` to create a zero-initialized object. I do not think that the caller of `dict_process_sys_indexes_rec()` would refer to the field `trx_scn`. At least I do not see any change to `i_s_sys_indexes_fill_table()` in this pull request.

storage/innobase/fil/fil0fil.cc
```diff
@@ -3366,3 +3366,19 @@ fil_space_t *fil_space_t::prev_in_unflushed_spaces()
 }
 
 #endif
+
+/** Returns the size of the space in pages. The tablespace must be cached
+in the memory cache.
+@param[in]      space_id        Tablespace ID
+@return space size, 0 if space not found */
+ulint fil_space_get_size(uint32_t space_id)
```
dr-m, line 3374: Why is this function being added? How is the return value being used? The return value may already be stale by the time the function returns.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15252,6 +15252,16 @@ ha_innobase::check(
 		m_prebuilt->need_to_access_clustered =
 			!!(check_opt->flags & T_EXTEND);
 
+		if (m_prebuilt->need_to_access_clustered && srv_use_scn) {
+			/* trx id on secondary index page can't be used to check visibility while scn is used, so disable it and print warning */
```
dr-m, line 15256: How would MVCC on secondary indexes work if `innodb_use_scn=ON`? The commit message does not mention anything about it.

storage/innobase/read/read0read.cc
```diff
@@ -262,4 +307,552 @@ void trx_sys_t::clone_oldest_view(ReadViewBase *view) const
...
+}
+
+Scn_Map::~Scn_Map() { delete[] m_elems; }
+
+void run_cleanout_task(void *arg);
+
+SCN_Mgr::CleanoutWorker::CleanoutWorker(uint32_t id)
+{
+  m_id= id;
+  m_pages= new CleanoutArray(CLEANOUT_ARRAY_MAX_SIZE);
```
dr-m, line 428: Can we avoid the pointer indirection?

storage/innobase/read/read0read.cc
```diff
@@ -262,4 +307,552 @@ void trx_sys_t::clone_oldest_view(ReadViewBase *view) const
...
+  m_thd= innobase_create_background_thd("SCN cleanout worker");
+  m_task= new tpool::waitable_task(run_cleanout_task, &m_id);
+}
+
+SCN_Mgr::CleanoutWorker::~CleanoutWorker()
+{
+  delete m_pages;
+  destroy_background_thd(m_thd);
+  delete m_task;
```
dr-m, line 437: Can we avoid the pointer indirection for `m_task`?

storage/innobase/read/read0read.cc
```diff
@@ -262,4 +307,552 @@ void trx_sys_t::clone_oldest_view(ReadViewBase *view) const
...
+  while ((m_pages->get(page_id, table_id)))
+  {
+    pages.insert({page_id, table_id});
+  }
+}
+
+SCN_Mgr::SCN_Mgr()
+{
+  m_scn_map= new Scn_Map();
+  m_random_map= new Scn_Map();
```
dr-m, line 461: Can we avoid the pointer indirection?

storage/innobase/read/read0read.cc
```diff
@@ -262,4 +307,552 @@ void trx_sys_t::clone_oldest_view(ReadViewBase *view) const
...
+  m_startup_id= 0;
+  m_startup_scn= 0;
+  m_abort= false;
+  m_max_cleanout_threads= srv_cleanout_threads;
```
dr-m, line 465: Can we do without `m_max_cleanout_threads`? One variable name `innodb_cleanout_threads` would be more `grep` friendly.

storage/innobase/read/read0read.cc
```diff
@@ -262,4 +307,552 @@ void trx_sys_t::clone_oldest_view(ReadViewBase *view) const
...
+SCN_Mgr::~SCN_Mgr()
+{
+  if (m_scn_map)
+  {
+    delete m_scn_map;
+  }
+  if (m_random_map)
+  {
+    delete m_random_map;
+  }
```
dr-m, line 493: `delete` of a null pointer is allowed already in C++98.

storage/innobase/read/read0read.cc
```diff
@@ -262,4 +307,552 @@ void trx_sys_t::clone_oldest_view(ReadViewBase *view) const
...
+      fil_space_t *space= fil_space_get(page_id.space());
+      ulint zip_size= 0;
+      ut_a(space);
+      zip_size= space->zip_size();
+      mtr_t mtr;
+      buf_block_t *block= nullptr;
+      mtr_start(&mtr);
+      mtr.set_in_scn_cleanout();
+      block= buf_page_get_gen(page_id, zip_size, RW_X_LATCH, nullptr,
+                              BUF_GET_IF_IN_POOL, &mtr, nullptr);
```
dr-m, line 778: `BUF_GET_IF_IN_POOL` does not require `zip_size`.

storage/innobase/read/read0read.cc
```diff
@@ -262,4 +307,552 @@ void trx_sys_t::clone_oldest_view(ReadViewBase *view) const
...
+  }
+  m_abort= false;
+
+  m_cleanout_task_timer.reset(
+      srv_thread_pool->create_timer(::cleanout_task_monitor, nullptr));
+  m_cleanout_task_timer->set_time(0, 1000);
+
+  m_view_task_timer.reset(
+      srv_thread_pool->create_timer(run_view_task, nullptr));
+  m_view_task_timer->set_time(0, 1000);
```
dr-m, line 833: What are these for? The commit message does not mention any periodic task or timer.

Could any of this be executed by the `purge_coordinator_task`?

storage/innobase/row/row0log.cc
```diff
@@ -1812,7 +1812,7 @@ row_log_table_apply_delete(
 		ut_ad(len == DATA_ROLL_PTR_LEN);
 
 		if (memcmp(mrec_trx_id, rec_trx_id,
-			   DATA_TRX_ID_LEN + DATA_ROLL_PTR_LEN)) {
+		    DATA_TRX_ID_LEN + DATA_ROLL_PTR_LEN)) {
```
dr-m, line 1815: The indentation was correct before this change.

dr-m (changes requested):
As far as I understand, this change introduces a Boolean file format option `innodb_use_scn`. If we implemented this, we should also test how older versions would react to undo logs that exist in this format. The latest change to any InnoDB file format was in the MariaDB Server 10.8 short term release. You’d have to test how MariaDB Server 10.11 (the latest long-term support release) would react.

Changing the option on server restart is only expected to work if the undo logs are logically empty. I see that this rule is being enforced in `trx_lists_init_at_db_start()`.

I think that the parameter needs to be covered in `--suite=mariabackup` as well. Did all test suites pass when running `./mtr --mysqld=--loose-innodb-use-scn=on`? I would like to see a successful CI run that enables this option by default. All builders (including ASAN, MSAN, UBSAN) need to pass all tests in both modes. What is the test suite execution time with both values of the parameter?

Has this been tested with Galera or with replication? Do all nodes have to use the same value of the parameter? How does the SCN affect checks for implicit locks?

I would like to know how the performance impact of this change was tested, so that those tests can be repeated independently and the results analyzed. Do you have any build scripts, benchmark scripts, information on the test environment, and results, comparing to the parent commit as well as comparing the two values of the new configuration parameter `innodb_use_scn`?

I am worried that the added conditions in low-level code could incur some performance penalty. Could we have a `cmake` option for enabling this feature, similar to `PLUGIN_PERFSCHEMA`, `WITH_WSREP`, or `WITH_INNODB_AHI`?

storage/innobase/page/page0cur.cc
```diff
@@ -2298,6 +2298,9 @@ page_cur_delete_rec(
 	consistency checks would fail for the dummy block that is being
 	used during IMPORT TABLESPACE. */
 	block->modify_clock++;
+  if (mtr != nullptr) {
```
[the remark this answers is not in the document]
poempeng: I will remove the unnecessary condition.

mysql-test/suite/gcol/t/innodb_virtual_debug_purge.test
```diff
@@ -104,6 +104,7 @@ INSERT INTO t1(a, b) VALUES (5, 5);
 SET DEBUG_SYNC= 'inplace_after_index_build SIGNAL uncommitted WAIT_FOR purged';
 --error ER_ALTER_OPERATION_NOT_SUPPORTED_REASON
 ALTER TABLE t1 ADD COLUMN c INT GENERATED ALWAYS AS(a+b), ADD INDEX idx (c), ALGORITHM=INPLACE, LOCK=NONE;
+sleep 2;
 send ALTER TABLE t1 ADD COLUMN c INT GENERATED ALWAYS AS(a+b), ADD INDEX idx (c), ALGORITHM=INPLACE, LOCK=SHARED;
```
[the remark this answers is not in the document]
poempeng: The alter statement will create an readview, and the value of `ReadViewBase::m_low_limit_no` is obtained from `SCN_Mgr::m_safe_limit_no`, which is refreshed every second. Here `sleep` is used to wait `SCN_Mgr::m_safe_limit_no` to be refreshed, otherwise the subsequent `wait_all_purged.inc` will be blocked by this alter statement.

storage/innobase/trx/trx0trx.cc
poempeng: Well, the check `purge_sys.purge_queue.empty()` is not needed here.

storage/innobase/trx/trx0trx.cc
poempeng: For mariabackup, we always use the scn status in the trx sys page

mysql-test/suite/innodb/t/scn_on_and_off.test
poempeng: Check table ... extended is not supported(`extended` is ignored and a warning is reported).

mysql-test/suite/innodb/t/scn_on_and_off.test
poempeng: OK

storage/innobase/srv/srv0srv.cc
poempeng: OK

storage/innobase/srv/srv0start.cc
poempeng: Sure.

storage/innobase/fil/fil0fil.cc
poempeng: This function is used in trx_undo_hdr_get_scn to perform a rough check. If the page exceeds the size of the space, it means that the history of the transaction has been purged.

storage/innobase/handler/ha_innodb.cc
poempeng: The MVCC  on secondary index remains the same as before. Firstly call `ReadViewBase::sees` with the `max_trx_id` obtained from the page. If the result is false, the next step is to examine the clustered index.

storage/innobase/read/read0read.cc
poempeng: Sure.

storage/innobase/read/read0read.cc
poempeng: OK.

storage/innobase/read/read0read.cc
poempeng: To optimize the performance of MVCC, we need to write back SCN to TRX_ID fields of rows that touched by reading. These write back operations are done in background tasks. The timer `m_view_task_timer` is used to traverse `trx_sys_t::rw_trx_hash ` periodically and refresh `SCN_Mgr::m_safe_limit_no` and `SCN_Mgr::m_min_active_id`.
The work of these two timers can be done in `purge_coordinator_task`, but it may make the code become complex.

storage/innobase/include/dict0mem.h
```diff
@@ -1430,6 +1430,9 @@ struct dict_index_t {
   {
     online_log= reinterpret_cast<row_log_t*>(this);
   }
+
+  /** SCN of the index */
+	trx_id_t trx_scn;
```
dr-m, line 1435: The comment should explain the exact purpose of the field. As far as I can tell, this field is not being restored from any persistent data structures. Modifications appear to be scheduled via a call to `trx_t::add_scn_index()` in `row_create_index_for_mysql()`, that is, during DDL operations only.

What is the relationship of `dict_index_t::trx_scn` and `dict_index_t::trx_id`? The latter field is not being restored on database startup, because all transactions will be committed or rolled back at server startup and therefore all indexes after crash recovery will be accessible (unless they are marked as corrupted). As far as I could tell (by adding `__attribute((deprecated))` to the declarations), the only place where these fields matter is `row_merge_is_index_usable()`, whose original purpose was to prevent old read views from using indexes that had been created (or tables rebuilt) after the read view creation. Can we do with just `dict_index_t::trx_id`?

storage/innobase/include/trx0sys.h
```diff
@@ -856,7 +858,7 @@ class trx_sys_t
     The smallest number not yet assigned as a transaction id or transaction
     number. Accessed and updated with atomic operations.
   */
-  alignas(CPU_LEVEL1_DCACHE_LINESIZE) Atomic_counter<trx_id_t> m_max_trx_id;
+  alignas(CPU_LEVEL1_DCACHE_LINESIZE) std::atomic<trx_id_t> m_max_trx_id;
```
dr-m, line 861: I think that `Atomic_relaxed` would be a better choice here. By default, `std::atomic` would use `std::memory_order_seq_cst`, which would makes stores more expensive even on processors that implement Total Store Ordering. On processors that employ a weak memory model there will be an increased cost for loads as well. See https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html.

storage/innobase/btr/btr0cur.cc
```diff
@@ -4402,6 +4431,19 @@ btr_cur_pessimistic_delete(
...
+		if (!index->online_log)
+		{
+			scn_mgr.batch_write(block, mtr);
+		}
+		else
+		{
+			block->clear_cursor();
+		}
+	}
+#endif
```
dr-m, line 4445: This code is being duplicated in many places. I think that it would be better to write a separate function for it.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15212,6 +15212,15 @@ ha_innobase::check(
...
 			!!(check_opt->flags & T_EXTEND);
 
+#ifdef WITH_INNODB_SCN
+		if (m_prebuilt->need_to_access_clustered && innodb_use_scn) {
+		  /* trx id on secondary index page can't be used to accurately determine
+		  visibility while scn is used, so disable it and print a warning */
+		  m_prebuilt->need_to_access_clustered = FALSE;
+		  sql_print_warning("InnoDB: check table ... extended is not supported by scn");
+		}
+#endif
```
dr-m, line 15222: What is the exact problem, and would it also affect `lock_sec_rec_some_has_impl()` or `row_vers_impl_x_locked_low()`?

If I understood it correctly, `innodb_use_scn=true` causes the `DB_TRX_ID` in clustered index records to be something else than the last transaction identifier that modified the record. This could potentially break some logic in `trx_undo_prev_version_build()`.

I did not find any change to the way how `PAGE_MAX_TRX_ID` is updated on secondary index pages. The `CHECK TABLE…EXTENDED` code is simply traversing the entire available history of clustered index record versions to find out whether a secondary index record is orphan or belongs to some available record version. Why would that not work when `innodb_use_scn=true`?

storage/innobase/trx/trx0trx.cc
```diff
@@ -1416,6 +1496,42 @@ TRANSACTIONAL_INLINE inline void trx_t::commit_in_memory(const mtr_t *mtr)
...
+        if (scn == TRX_ID_MAX)
+        {
+          scn= trx_sys.get_new_trx_scn();
+          trx_sys.refresh_rw_trx_hash_scn_version();
+        }
+
+        for (auto &item : this->scn_indexes)
+        {
+          dict_index_t *index= dict_index_get_by_id(
+              item.first, item.second, this->dict_operation_lock_mode);
```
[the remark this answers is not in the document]
dr-m: Sorry, I later realized that the `trx_t::scn_indexes` is only used during DDL operations, via `row_create_index_for_mysql()`. I wonder if we can simplify that and change the way how `dict_index_t::trx_id` is assigned earlier during DDL.

dr-m:
This does show some potential in performance tests. However, I think that we need to examine other ways to enhance performance, such as [MDEV-35702](https://jira.mariadb.org/browse/MDEV-35702), before changing the file format.

PR #2957, opened 2023-12-21
Fix a case of `unused-but-set-variable`

tgross35:

## Description

The `unused-but-set-variable` warning is raised on MacOS from the `posix_fadvise` standin macro, since offset is often otherwise unused. Add a cast to absorb this warning.

## How can this PR be tested?

Compile files such as `ds_tmpfile.cc` with `-Wunused-but-set-variable`

## PR quality check


grooverdan:
seems odd that offset is special. Note the other macro:
```
storage/innobase/row/row0merge.cc:# define posix_fadvise(fd, offset, len, advice) /* nothing */
```

Please target 10.4

tgross35:
Agreed, it is weird that this shows up and can't be flag disabled. I think it's just the one variable that doesn't have a use otherwise - `fd` and `len` are used to access the file and `advice` expands to nothing, but it seems like we only use `offset` for this purpose.

Seems like use in `row0merge.cc` is mostly gated but the one from `common.h` isn't?

Changed the base, still doesn't build fully but at least it doesn't get stuck on this.

LinuxJedi:
Hi @tgross35, unfortunately this has missed the boat for 10.4. It will need rebasing against 10.5 for us to approve it. Would you like to do it? Or should we do it?

tgross35:
Sure, I reapplied it.

LinuxJedi (approved):
Many thanks, looks good to me.

PR #2958, opened 2023-12-21
Convert core instances of `sprintf` to `snprintf`

tgross35:

## Description

## How can this PR be tested?

Compile with `-Wdeprecated-declarations` (I believe it is enabled by default for the relevant files) and use a version of `libc` that marks `sprintf` deprecated.

This is based against 11.4, I am not sure if it is better to base against an older version.

## PR quality check


storage/innobase/srv/srv0start.cc
```diff
@@ -1264,13 +1264,13 @@ dberr_t srv_start(bool create_new_db)
 
 	if (!srv_read_only_mode) {
 		if (srv_innodb_status) {
-
+			ulong file_name_len = strlen(fil_path_to_mysql_datadir)
```
vaintroub, line 1267: the correct datatype returned by `strlen` is `size_t`, not `ulong`.  Using longs is discourages by the coding standard, and also this line made pull request fail on Windows builders.

tgross35: Thanks, I made these type adjustments.

tgross35:
Seems like I get different failures every time CI runs. There must be a case where either I did the buffer size incorrectly, or we are actually doing an OOB write.

grooverdan:
@tgross35  could just be some tests are flakey, can you resolve the conflict and see how it goes.

tgross35:
I'm having trouble debugging this locally, I am going to split it up into parts to make it easier to debug via CI

- https://github.com/MariaDB/server/pull/3075
- https://github.com/MariaDB/server/pull/3076

tgross35:
@LinuxJedi I think that #3076 and #3077 should be good for a look, seems like the problem is coming from one of the other sections.

dlenski:
Note that the long-open #2601 is overlapping with this PR in its goals, even if not in the precise code that it touches.

tgross35:
> Note that the long-open #2601 is overlapping with this PR in its goals, even if not in the precise code that it touches.

Thanks for the link, indeed that overlaps quite a bit. My focus was mostly on the changes that are blocking builds (due to deprecation warnings that apparently can't be overridden in some cases, or else I haven't figured out how), but there are a lot more cases around.

I almost wonder if it would make more sense to switch to a infallible wrapper, since the cursor pointer pattern is pretty common but kind of ugly to work with.

```c
/** A cursor for writing to a buffer that is expected to never fail (asserts on failure) */
struct cursor {
    char *pos;
    char *end;
}

/** Write a format string to the cursor, wrapping snprintf internally */
void cursor_write(struct cursor *curs, const char * format, ...) {
    size_t remaining = cursor_remaining(curs);
    int res = snprintf(curs->pos, remaining, format, ...);
    DBUG_ASSERT(res >= 0 && res <= remaining);
    curs->pos += res;
}

/** Write a string to the cursor and return the start of the appended string (strcat) */
char *cursor_write_str(struct cursor *curs, const char *s) {
    size_t len = strlen(s);
    char *ret = curs->pos;
    DBUG_ASSERT(len + 1 <= cursor_remaining(curs));
    memcpy(pos, s, len);
    curs->pos += len;
    *curs->pos = '\0';
    return ret;
}

/** Write a single character to the string */
void cursor_write_char(struct cursor *curs, char c) {
    DBUG_ASSERT(cursor_remaining(curs) >= 1);
    *curs->pos = c;
    ++(curs->pos);
}

/** Return the remaining space in this cursor. Will assert if in an invalid state. */
size_t cursor_remaining(const struct cursor *curs) {
    DBUG_ASSERT(curs->end >= curs->pos && "cursor is in an invalid state");
    return (size_t)(curs->end - curs->pos);
}
```

dlenski:
@tgross35 wrote:

> I almost wonder if it would make more sense to switch to a infallible wrapper, since the cursor pointer pattern is pretty common but kind of ugly to work with.

Yes, almost every C project of any size and sophistication uses _some kind of wrapper_ for mutable text buffers that keep track of their size, since it's extremely error-prone to code this by hand.

I'm one of the developers of [OpenConnect (the multi-protocol VPN client)](https://gitlab.com/openconnect/openconnect), and it has had [just such a wrapper](https://gitlab.com/openconnect/openconnect/blob/master/textbuf.c) since before I was involved in the project.

Somehow MySQL/MariaDB has managed to exist for nearly 30 years without adopting such a mechanism in any consistent way…

tgross35:
Closing this out since the other PRs exist

PR #2959, opened 2023-12-21
MDEV-31772 Innodb_rows_* missing

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-31772*

## Description
In commit 0dab74ff3fbebfcaf899ad439b17c8077f66d78a (MDEV-28539) a number of counters were removed. Due to popular demand, we are reinstating some of the counters, in a way that incurs less performance penalty.

We reinstate the following global status variables:
innodb_rows_read
innodb_rows_deleted
innodb_rows_inserted
innodb_rows_updated

These counters will also include whatever was previously covered by innodb_system_rows_* counters.

We also add the INFORMATION_SCHEMA.INNODB_METRICS counters dml_reads, dml_inserts, dml_delete, dml_updates, but not their dml_system_* counterparts.

ha_innobase::external_lock(): Update the global export_vars.rows from the local ha_innobase::rows counters at the end of each SQL statement.

## How can this PR be tested?
The regression test suite is extended, but we would also want to test the performance impact of this.

I would also like to know if we really have to reinstate the `information_schema.innodb_metrics` counters. I would like to deprecate and remove that interface, and just export everything as global status variables.

## PR quality check


storage/innobase/handler/ha_innodb.cc
```diff
@@ -16114,6 +16134,13 @@ ha_innobase::external_lock(
 		DBUG_RETURN(0);
 	} else {
 		DEBUG_SYNC_C("ha_innobase_end_statement");
+		export_vars.rows.mutex.wr_lock();
+		export_vars.rows.deleted += rows.deleted;
+		export_vars.rows.inserted += rows.inserted;
+		export_vars.rows.read += rows.read;
+		export_vars.rows.updated += rows.updated;
+		export_vars.rows.mutex.wr_unlock();
+		new (&rows) typeof(rows) {};
```
dr-m, line 16143: MSVC did not like the `typeof`, so I guess we have to use a named `struct` type for `rows`. The constructor call is for resetting the local `rows` counters after they have been added to the global ones.

FooBarrior: can `rows= {};` work out?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -16114,6 +16134,13 @@ ha_innobase::external_lock(
 		DBUG_RETURN(0);
 	} else {
 		DEBUG_SYNC_C("ha_innobase_end_statement");
+		export_vars.rows.mutex.wr_lock();
```
FooBarrior, line 16137: You may have problems with the `lookup_handler` stats by putting this into `ha_external_unlock` after @vuvova's 3bb5c6b. Try this test:
```
create table t(id int, s date, e date, period for p(s,e), primary key(id, p without overlaps));

insert t values(1, '2020-01-01', '2021-01-01');
update t for portion of p from '2020-02-02' to '2020-03-03' set id = 2;
--source suite/rpl/include/rpl_innodb_rows_counters.inc
flush table t;
--source suite/rpl/include/rpl_innodb_rows_counters.inc
```

FooBarrior: If I'm right and you don't want to wait for a fix I could suggest you to store the counters only in main handler for a table:
```
ha_innobase *main_handler= static_cast<ha_innobase*>table->file;
main_handler->row.read++;
```
etc.

FooBarrior (changes requested):

storage/innobase/handler/ha_innodb.cc
FooBarrior: On the second thought, let's make a correct behavior right here.
Add these two lines in the end of `ha_external_lock`:
```c++
       mysql_audit_external_lock(thd, table_share, lock_type);
   }
 
+  if (lookup_handler != this)
+      lookup_handler->ha_external_unlock(table->in_use);
+
   if (MYSQL_HANDLER_RDLOCK_DONE_ENABLED() ||
       MYSQL_HANDLER_WRLOCK_DONE_ENABLED() ||
       MYSQL_HANDLER_UNLOCK_DONE_ENABLED())

```

dr-m: I tried your suggestion. I see no impact on the counters: they are 6,2,1,4 (master) and 5,2,1,4 (replica) with or without applying the code change. Can you check if those numbers make any sense? Also, with your suggested code change, I am observing an assertion failure:
```
#10 0x00007f66990533a2 in __assert_fail (assertion=0x563f66529230 "!n_mysql_tables_in_use", file=0x563f6663b778 "/mariadb/10.11/storage/innobase/trx/trx0trx.cc", line=377, function=0x563f66529247 "void trx_t::free()") at ./assert/assert.c:101
#11 0x0000563f675d9088 in trx_t::free (this=this@entry=0x7f668f3feb80) at /mariadb/10.11/storage/innobase/trx/trx0trx.cc:377
#12 0x0000563f67340e12 in innobase_close_connection (hton=<optimized out>, thd=0x7f6654000d58) at /mariadb/10.11/storage/innobase/handler/ha_innodb.cc:4951
#13 0x0000563f66e1e972 in ha_close_connection (thd=thd@entry=0x7f6654000d58) at /mariadb/10.11/sql/handler.cc:914
#14 0x0000563f66f84b88 in THD::free_connection (this=this@entry=0x7f6654000d58) at /mariadb/10.11/sql/sql_class.cc:1616
#15 0x0000563f66d8aa11 in unlink_thd (thd=0x7f6654000d58) at /mariadb/10.11/sql/mysqld.cc:2851
```
My entire patch is as follows:
```diff
diff --git a/mysql-test/suite/rpl/t/rpl_innodb_bug68220.test b/mysql-test/suite/rpl/t/rpl_innodb_bug68220.test
index 967ac2cf164..567cf4ad106 100644
--- a/mysql-test/suite/rpl/t/rpl_innodb_bug68220.test
+++ b/mysql-test/suite/rpl/t/rpl_innodb_bug68220.test
@@ -22,10 +22,10 @@ select variable_value into @slave_rows_inserted from information_schema.global_s
 --connection master
 CREATE DATABASE testdb;
 USE testdb;
-CREATE TABLE testdb.t1 (i int NOT NULL PRIMARY KEY) ENGINE=InnoDB;
+CREATE TABLE t1 (i int NOT NULL PRIMARY KEY) ENGINE=InnoDB;
 
 # insert a row and show counters on master and slave
-INSERT INTO testdb.t1 VALUES (1);
+INSERT INTO t1 VALUES (1);
 --source suite/rpl/include/rpl_innodb_rows_counters.inc
 
 # update the row and show counters on master and slave
@@ -37,8 +37,16 @@ UPDATE t1 SET i=2 WHERE i=1;
 DELETE FROM t1 WHERE i=2;
 --source suite/rpl/include/rpl_innodb_rows_counters.inc
 
+CREATE TABLE t
+(id INT, s DATE, e DATE, PERIOD FOR p(s,e),
+ PRIMARY KEY(id, p WITHOUT OVERLAPS)) ENGINE=InnoDB;
+INSERT INTO t VALUES(1, '2020-01-01', '2021-01-01');
+UPDATE t FOR PORTION OF p FROM '2020-02-02' TO '2020-03-03' SET id = 2;
+--source suite/rpl/include/rpl_innodb_rows_counters.inc
+FLUSH TABLE t;
+--source suite/rpl/include/rpl_innodb_rows_counters.inc
+
 # clean the test
-DROP TABLE t1;
 DROP DATABASE testdb;
 --sync_slave_with_master
 
diff --git a/sql/handler.cc b/sql/handler.cc
index 83ae4117725..785e7e2797d 100644
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -7159,6 +7159,9 @@ int handler::ha_external_lock(THD *thd, int lock_type)
       mysql_audit_external_lock(thd, table_share, lock_type);
   }
 
+  if (lookup_handler != this && lock_type == F_UNLCK)
+    lookup_handler->ha_external_unlock(table->in_use);
+
   if (MYSQL_HANDLER_RDLOCK_DONE_ENABLED() ||
       MYSQL_HANDLER_WRLOCK_DONE_ENABLED() ||
       MYSQL_HANDLER_UNLOCK_DONE_ENABLED())
diff --git a/storage/innobase/handler/ha_innodb.cc b/storage/innobase/handler/ha_innodb.cc
index 67ce245f48c..aed122832af 100644
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -16140,7 +16140,7 @@ ha_innobase::external_lock(
 		export_vars.rows.read += rows.read;
 		export_vars.rows.updated += rows.updated;
 		export_vars.rows.mutex.wr_unlock();
-		new (&rows) typeof(rows) {};
+		rows = {};
 	}
 
 	/* MySQL is releasing a table lock */
```

dr-m: I applied the test case patch to 10.6 bdaa6bac05b2c0cfb8a974a798775a88e124e2de and got the same counter differences: 6,2,1,4 on the master and 5,2,1,4 on the replica.

FooBarrior: I was wrong about the nature of `ha_reset` it's called for every DML in the end.  So there is no problem with Serg's solution.

FooBarrior (approved):

dr-m:
I ran a simple performance test: Sysbench `oltp_update_non_index`, 5×10000 rows, 256 concurrent connections, 60 seconds, CPU bound, RAM disk.
| revision | average throughput/tps | run1/tps | run2/tps | run3/tps |
| -------- | ----------------------:| --------:| --------:| --------:|
| 09049fe496eea1c19cd3ce80a788fa4b75d9609e (10.11 baseline) | 197918.43 | 197257.75 | 198060.00 | 198437.55 |
| 5284d6174e8771402c5db98f5227998b62e99dad | 196706.34 | 196827.54 | 196510.00 | 196781.49 |

I am not happy about the 0.6% performance loss. I will experiment with sharding the `export_vars.rows`.

dr-m:
I tried a few other ideas:
* splitting `export_vars.rows` into 128 shards
* updating the totals only on `ha_innobase::close()` (`flush tables`) and not at the end of each statement
* not updating the `innodb_rows_read` counter:
```diff
diff --git a/storage/innobase/handler/ha_innodb.cc b/storage/innobase/handler/ha_innodb.cc
index b3b082ac449..9eddcdee2d2 100644
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -9079,7 +9079,7 @@ ha_innobase::index_read(
 	case DB_SUCCESS:
 		error = 0;
 		table->status = 0;
-		rows.read++;
+		//rows.read++;
 		break;
 
 	case DB_RECORD_NOT_FOUND:
@@ -9335,7 +9335,7 @@ ha_innobase::general_fetch(
 	case DB_SUCCESS:
 		error = 0;
 		table->status = 0;
-		rows.read++;
+		//rows.read++;
 		break;
 	case DB_RECORD_NOT_FOUND:
 		error = HA_ERR_END_OF_FILE;
```
| variant | average throughput/tps | run1/tps | run2/tps | run3/tps |
| -------- | ----------------------:| --------:| --------:| --------:|
| bdf65893ddb2ce0552a20e53e2358350521d2d35 (10.11 baseline) | 236735.67 | 238188.90 | 234919.27 | 237098.83 |
| 1 shard (c8fa41cac0d15f670094322464a15821dca87348 a.k.a. rebased 5284d6174e8771402c5db98f5227998b62e99dad) | 234799.73 | 235391.10 | 234028.00 | 234980.08 |
| 1 shard, updated on `ha_innobase::close()` | 235422.28 | 235362.99 | 235095.15 | 235808.70 |
| 1 shard, updated on `ha_innobase::close()`, no `innodb_rows_read` | 235282.07 | 235984.95 | 233536.71 | 236324.56 |
| 128 shards (bdba02ddd17871f8f5fb6beade375692e90cf8ce) | 236448.81 | 236424.07 | 235775.24 | 237147.11 |
| 128 shards, no `innodb_rows_read` | 237621.26 | 237419.82 | 237657.59 | 237786.36 |
| 128 shards, updated on `ha_innobase::close()` (dffef77b5bec3929baf271184cfee2035fa8f257) | 233779.64 | 233934.62 | 233604.62 | 233799.69 |
| 128 shards, updated on `ha_innobase::close()`, no `innodb_rows_read` | 234488.34 | 235796.12 | 232811.35 | 234857.56 |

My benchmark is not reliable, because there is rather large variance for the numbers, The last 2 rows indicate a worse regression contrary to my expectations. For the baseline, I think I had observed one run that exceeded 240 ktps. The lowest that I got for it was 232404.93 tps.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -6290,6 +6294,17 @@ ha_innobase::close()
 {
 	DBUG_ENTER("ha_innobase::close");
 
+	auto& slice = export_vars.row_slices
+		[get_rnd_value()
```
FooBarrior, line 6298: Should be worse on NUMA: now a node should fetch `array_elements(export_vars.row_slices)` cache lines!
If a transaction has more that one table, it will be sensed. I'd suggest to use getcpu for x86 instead if your glibc is fresh enough, at least for a benchmark purpose. Then we could improve the patch to support the rest of the platforms (sadly, arm64 is [still missing](https://elixir.bootlin.com/linux/v6.7/C/ident/__vdso_getcpu), however the [patch is ready](https://elixir.bootlin.com/linux/v6.7/C/ident/__vdso_getcpu) for as long as 3 years.

storage/innobase/handler/ha_innodb.h
```diff
@@ -499,6 +499,10 @@ class ha_innobase final : public handler
 	this is set in external_lock function */
 	THD*			m_user_thd;
 
+private:
+	/** Row operation counters */
+	struct { size_t deleted= 0, inserted= 0, read= 0, updated= 0; } rows;
```
FooBarrior, line 504: I just noticed it now - the stats are global, not per-table. No need to store this per handler, you can put it in `trx_t`. Then, all the all the tables' stats in the transaction will be updated with a single write. Perhaps, I didnt't consider a performance as an issue here, so didn't notice.

FooBarrior (commented):
I think you rather need a microbenchmark here. It's clear that _some_ time is consumed by the global stats update (not by rows.read++ -- it's 1ns most), so we'll see the result > 0, but what's more important, we'll be able to distinguish your approaches by the time consumed. The important measure can be a relation of stats update time to the total/average transaction time.

storage/innobase/handler/ha_innodb.cc
dr-m: Oh, you are right, `get_rnd_value()` primarily invokes `my_timer_cycles()`, which already by itself could be an expensive operation. If that returns 0, only then it will attempt an imitation of `getcpu()` by computing something from `pthread_self()`. In fact, the least significant 7 bits of that address may have no chance of uniquely identifying a thread. The address might be aligned to `CPU_LEVEL1_DCACHE_LINESIZE`, which could be 64 or 128 or an even larger power of 2.

storage/innobase/handler/ha_innodb.h
dr-m: Good idea. In that way, we can simply update the (possibly sharded) global statistics in `trx_t::commit()`.

dr-m:
e9ccbc275796f24b97be09395af1f6378e636842 implements @FooBarrior’s suggestion of updating transaction-level counters.

I was thinking if there would be an even better way, to also make the counters available in `ANALYZE FORMAT=JSON` (`trace_engine_stats()`, 6e484c3bd99a6709e3554f32a99167042ea88496). The challenges are:
1. What if the thread pool migrates a transaction between threads while it is waiting for a lock? Is the call to `handler::ha_handler_stats_reset()` in `TABLE::init()` even correct with respect to the thread pool?
2. How to traverse all thread-local row operation counts to update the global counters? At `trx_t::commit_cleanup()`, or is there some way to invoke this less often? Again, this would seem to be a bottleneck due to cache line contention on the global counters.

Perhaps it is better to revise e9ccbc275796f24b97be09395af1f6378e636842 so that `trx_t::commit_cleanup()` will update nothing, and instead the counters will be updated by traversing `trx_pools`.

dr-m:
With e9ccbc275796f24b97be09395af1f6378e636842 there was a clear performance regression. Let us see if eee6f9a0237f21c36cbc1370341b73a78bc5de20 is any better. It will avoid adding `innodb_rows_read`. The modification counters will only be updated in the slow code path of `trx_t::commit_in_memory()`. Thus, the only performance penalty for read-only transactions would be the zeroing of the counters in `trx_init()`.

FooBarrior (commented):
This patch has more changes than just the counters update. Did you try to measure its performance just without any counters?

storage/innobase/include/trx0trx.h
```diff
@@ -756,6 +757,10 @@ struct trx_t : ilist_node<>
   Transitions to COMMITTED are protected by trx_t::mutex. */
   Atomic_relaxed<trx_state_t> state;
 
+  enum row_op_names { ROW_INSERT= 0, ROW_DELETE, ROW_UPDATE };
+  /** Row operation counters */
+  alignas(CPU_LEVEL1_DCACHE_LINESIZE) size_t row_ops[3];
```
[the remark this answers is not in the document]
dr-m: `trx_t::state` will be read by other threads. It could change much less often than the added counter array.

dr-m:
If we were to resume this work, this would have to be refactored in the style of #4252.

PR #2960, opened 2023-12-21
Replace incorrect message `mariadb-safe` with correct `mariadbd-safe`

Sophist-UK:
## Description

This patch corrects some message text that tells the user (incorrectly) to use `mariadb-safe` instead of `mysqld_safe` because the future release file is `mariadbd-safe` not `mariadb-safe`.

## How can this PR be tested?

No testing is needed as there are zero functional changes in this PR.

## PR quality check


grooverdan (approved):

grooverdan:
Thanks!

PR #2961, opened 2023-12-21
[MDEV-28162] Replace PFS_atomic with std::atomic<T>

keeper:
- [x] *The Jira issue number for this PR is: MDEV-28162*

## Description
PFS_atomic class contains wrappers around my_atomic_* operations, which
are macros to GNU atomic operations (`__atomic_*`). Due to different
implementations of compilers, clang may encounter errors when compiling
on x86_32 architecture.

The following functions are replaced with C++ std::atomic type in
performance schema code base:
```
  - PFS_atomic::store_*()
      -> my_atomic_store*
        -> __atomic_store_n()
    => std::atomic<T>::store()

  - PFS_atomic::load_*()
      -> my_atomic_load*
        -> __atomic_load_n()
    => std::atomic<T>::load()

  - PFS_atomic::add_*()
      -> my_atomic_add*
        -> __atomic_fetch_add()
    => std::atomic<T>::fetch_add()

  - PFS_atomic::cas_*()
    -> my_atomic_cas*
      -> __atomic_compare_exchange_n()
    => std::atomic<T>::compare_exchange_strong()
```

and PFS_atomic class could be dropped completely.

`std::atomic` was added in C++11 (see https://en.cppreference.com/w/cpp/atomic/atomic), and [MariaDB server C++ source files explicitly require C++11](https://github.com/MariaDB/server/blob/10.4/CMakeLists.txt#L84-L92).

Note that in the wrapper memory order passed to original GNU atomic
extensions are hard-coded as `__ATOMIC_SEQ_CST`, which is equivalent to
`std::memory_order_seq_cst` in C++, and is the default parameter for
std::atomic_* functions.

## How can this PR be tested?
* [x] Builds with no errors
* [x] Unit test `pfs*` all passed
* [x] No regression in MTR test `perfschema` suite

## PR quality check

## Copyright
Services.


storage/perfschema/pfs_instr_class.cc
```diff
@@ -476,7 +476,8 @@ void PFS_table_share::refresh_setup_object_flags(PFS_thread *thread)
 int init_file_class(uint file_class_sizing)
 {
   int result= 0;
-  file_class_dirty_count= file_class_allocated_count= 0;
+  file_class_dirty_count.store(0);
+  file_class_allocated_count.store(0);
```
FooBarrior, line 480: these two can be of relaxed memory order

storage/perfschema/pfs_setup_actor.cc
```diff
@@ -175,15 +177,15 @@ int insert_setup_actor(const String *user, const String *host, const String *rol
   if (unlikely(pins == NULL))
     return HA_ERR_OUT_OF_MEM;
 
-  static uint PFS_ALIGNED setup_actor_monotonic_index= 0;
+  static std::atomic<uint> setup_actor_monotonic_index(0);
```
FooBarrior, line 180: please return PFS_ALIGNED back. It aligns the variable to the L1 cache line size, and this variable is static, and besides inside the function, so its layout is not clear (to me at least), and it can be false-shared with some other global variables.

storage/perfschema/pfs_setup_object.cc
```diff
@@ -175,15 +177,15 @@ int insert_setup_object(enum_object_type object_type, const String *schema,
   if (unlikely(pins == NULL))
     return HA_ERR_OUT_OF_MEM;
 
-  static uint PFS_ALIGNED setup_object_monotonic_index= 0;
+  static std::atomic<uint> setup_object_monotonic_index(0);
```
FooBarrior, line 180: PFS_ALIGNED is needed as well.

storage/perfschema/pfs_atomic.h
```diff
@@ -1,141 +0,0 @@
-/* Copyright (c) 2009, 2011, Oracle and/or its affiliates. All rights reserved.
```
FooBarrior, line 1: I think that unused shortcuts from my_atomic.h can also be removed (but the header itself should remain as a meta-header)

FooBarrior (changes requested):
Thank you @keeper! Looks fine so far. 

Please update the code according to my requests.

@dr-m  also mentions `Atomic_relaxed` in MDEV-28162, however I'm unsure if it's safe to apply. I think this task can be closed after the cleanup is made to my_atomic.h and the change is made to the same performance quality, as the code was before (consider my alignment fix requests). for improving the performance a separate task can be filed

storage/perfschema/pfs_setup_object.cc
keeper: I also dropped `PFS_ALIGNED` in other files such as `pfs_instr.cc` and `pfs_diget.cc` (static or not), presumably I should also add them back?

storage/perfschema/pfs_atomic.h
keeper: I also looked into this a bit when working on this jira, and found that it would affect other components (`my_open.c` and `my_fopen.c` for example). I surely would like to do that but perhaps in another PR?

dr-m:
> @dr-m also mentions `Atomic_relaxed` in MDEV-28162, however I'm unsure if it's safe to apply.

Each case needs to be considered carefully. My understanding is that unless there is some expected dependency or consistency between data fields, the relaxed memory ordering should be sufficient. This may be applicable to most or all of `performance_schema`, but I did not look deeper at it.

When multiple data fields need to be somehow consistent with each other, I think that [the release-acquire ordering](https://en.cppreference.com/w/cpp/atomic/memory_order#Release-Acquire_ordering) should work in a well designed system.

Based on the load/store examples in https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html the default sequential consistent memory order is expensive on any ISA, including those that implement total store ordering.

storage/perfschema/pfs_setup_object.cc
dr-m: If original the intention of the `PFS_ALIGNED` around `digest_monotonic_index` was to prevent false sharing of cache lines, then also the following global variable that is expected to be part of the false sharing should be declared as `PFS_ALIGNED`. It could be better to declare a `struct` around it, so that the `struct` itself will be padded to a multiple of the `PFS_ALIGNED` size. It would not seem to me that `digest_full` could be part of false sharing; it is often being changed together with that field. Something like the following might work:
```c++
struct
{
  PFS_ALIGNED Atomic_counter<uint32_t> monotonic_index;
  bool full;
} digest;
```
Note: Global or `static` variables are zero-initialized by default. An initializer like `bool digest_full=false;` is redundant.

keeper: Hey all, sorry for the late response. Since the PR is a bit stale so I rebased my changes to 10.5. In 10.5 these atomic counters are already wrapped in a structure (`struct PFS_cacheline_uint32` for example) so it can be simply done by replacing the member with atomic counter part (`uint32 m_u32;` to `std::atomic<uint32> m_u32;`)

storage/perfschema/pfs_account.h
```diff
@@ -62,22 +63,22 @@ struct PFS_ALIGNED PFS_account : PFS_connection_slice
...
   {
-    PFS_atomic::add_32(& m_refcount, 1);
+    m_refcount.fetch_add(1);
   }
 
   inline void dec_refcount(void)
   {
-    PFS_atomic::add_32(& m_refcount, -1);
+    m_refcount.fetch_sub(1);
   }
```
dr-m, line 82: Would it make sense to use [release-acquire ordering](https://en.cppreference.com/w/cpp/atomic/memory_order#Release-Acquire_ordering) at least in these operations, instead of continuing the use of the more expensive sequential consistent memory order? (For stores, according to https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html using a weaker ordering would seem to always avoid some overhead.)

FooBarrior: I think since it's a refactoring patch, we shouldn't demand ordering improvement in this PR/MDEV. However yes, pfs may have some improvements after some well-thinking. But before improving the performance schema performance, I'd fix the bugs.

Here's your favorite one: https://github.com/MariaDB/server/blob/d821fd7fab2723b84f774aaa4882acb86b35769d/storage/perfschema/pfs_setup_actor.cc#L88

Also don't think it's a good place to demand a bugfix

storage/perfschema/pfs_digest.cc
```diff
@@ -75,7 +75,7 @@ int init_digest(const PFS_global_param *param)
   */
   digest_max= param->m_digest_sizing;
   digest_lost= 0;
-  PFS_atomic::store_u32(& digest_monotonic_index.m_u32, 1);
+  digest_monotonic_index.m_u32.store(1);
```
FooBarrior, line 78: > static PFS_ALIGNED PFS_cacheline_uint32 digest_monotonic_index;
why would someone make both alignas(cache line) and wrap it into an char array-aligned struct (`PFS_cacheline_uint32`)? @dr-m do you have any idea, or is it just an overingineering coming from the mysql port?

FooBarrior (approved):
I have no more changes to request. It's no doubt, that PFS requires more work on improving its atomics work: bug fixing, even more refactoring, operformance improvements, but this particular patch looks complete to me and ok to push.

storage/perfschema/pfs_account.h
dr-m: @FooBarrior Can you file bugs for those suggestions and note the MDEV numbers here? I agree that we should not extend the scope of this pull request.

FooBarrior:
@dr-m As we discussed I have created separate tasks for fixing the races and improving performance in PFS:
* [MDEV-34465](https://jira.mariadb.org/browse/MDEV-34465)
* [MDEV-34464](https://jira.mariadb.org/browse/MDEV-34464)

PR #2962, opened 2023-12-22
MDEV-33098 The test innodb.doublewrite_debug occasionally fails to start up InnoDB

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-33098*

## Description
- innodb.doublewrite_debug should avoid the checkpoint before killing the server. So used debug sync and innodb_flush_sync to avoid the checkpoint completely.

## How can this PR be tested?
./mtr innodb.doublewrite_debug

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/doublewrite_debug.test
+++ b/mysql-test/suite/innodb/t/doublewrite_debug.test
@@ -44,7 +44,9 @@ commit work;
 # Slow shutdown and restart to make sure ibuf merge is finished
 SET GLOBAL innodb_fast_shutdown = 0;
 let $shutdown_timeout=;
+let $restart_parameters="--debug_dbug=+d,ib_log_checkpoint_avoid_hard --innodb_flush_sync=0";
 --source include/restart_mysqld.inc
+--source ../include/no_checkpoint_start.inc
 begin;
 insert into t1 values (6, repeat('%', 400));
 
@@ -57,8 +59,20 @@ set global innodb_saved_page_number_debug = 1;
 set global innodb_fil_make_page_dirty_debug = 0;
 
 set global innodb_buf_flush_list_now = 1;
-let $shutdown_timeout=0;
---source include/shutdown_mysqld.inc
+
+--let CLEANUP_IF_CHECKPOINT=drop table t1, unexpected_checkpoint;
+# Occasionally, a checkpoint would occur on the MSAN builder.
+# We do not know the reason, because the failure can only be reproduced if there is
+# enough load in that environment.
+# Therefore, we allow the test to be skipped when run on MSAN.
+# In other environments, we want the test to fail if a checkpoint occurs,
+# so that we would catch it if it starts to happen regularly.
+if (`select count(*) from information_schema.system_variables where variable_name='have_sanitizer' and global_value like "MSAN%"`)
+{
+--let CLEANUP_IF_CHECKPOINT=drop table t1;
+}
+
+--source ../include/no_checkpoint_end.inc
 
 --echo # Make the 1st page (page_no=0) and 2nd page (page_no=1)
 --echo # of the system tablespace all zeroes.
@@ -74,6 +88,7 @@ print FILE chr(0) x ($ENV{'INNODB_PAGE_SIZE'});
 close FILE;
 EOF
 
+let $restart_parameters=;
 --source include/start_mysqld.inc
 
 let SEARCH_PATTERN=InnoDB: Restoring page \[page id: space=0, page number=0\] of datafile;
@@ -90,7 +105,9 @@ select f1, f2 from t1;
 --echo # Test Begin: Test if recovery works if 1st page of
 --echo # system tablespace is corrupted and 2nd page as corrupted.
 
-set global innodb_log_checkpoint_now = 1;
+let $restart_parameters="--debug_dbug=+d,ib_log_checkpoint_avoid_hard --innodb_flush_sync=0";
+--source include/restart_mysqld.inc
+--source ../include/no_checkpoint_start.inc
 begin;
 insert into t1 values (6, repeat('%', 400));
 
@@ -103,8 +120,8 @@ set global innodb_saved_page_number_debug = 1;
 set global innodb_fil_make_page_dirty_debug = 0;
 
 set global innodb_buf_flush_list_now = 1;
-let $shutdown_timeout=0;
---source include/shutdown_mysqld.inc
+
+--source ../include/no_checkpoint_end.inc
 
 --echo # Corrupt the 1st page (page_no=0) and 2nd page of the system tablespace.
 perl;
@@ -119,6 +136,7 @@ print FILE chr(0) x ($ENV{'INNODB_PAGE_SIZE'}/2);
 close FILE;
 EOF
 
+let $restart_parameters=;
 --source include/start_mysqld.inc
 
 let SEARCH_PATTERN=InnoDB: Restoring page \[page id: space=0, page number=0\] of datafile;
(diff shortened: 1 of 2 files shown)


dr-m (changes requested):
I think that we’d better also use `no_checkpoint_start.inc` and `no_checkpoint_end.inc` so that in case an unexpected log checkpoint is triggered, the test will fail in a more obvious way. (Set `CLEANUP_IF_CHECKPOINT` to something that will cause a syntax error, so that the test will not be silently skipped but fail in that case.)

mysql-test/suite/innodb/t/doublewrite_debug.test
```diff
@@ -44,7 +44,10 @@ commit work;
 # Slow shutdown and restart to make sure ibuf merge is finished
 SET GLOBAL innodb_fast_shutdown = 0;
 let $shutdown_timeout=;
+let $restart_parameters="--debug_dbug=+d,ib_log_checkpoint_avoid_hard";
 --source include/restart_mysqld.inc
+--source ../include/no_checkpoint_start.inc
+set global innodb_flush_sync=0;
```
dr-m, line 50: I think that we should swap these. The "critical section" for checkpoint is the transaction that follows these lines.

mysql-test/suite/innodb/t/doublewrite_debug.test
```diff
@@ -57,8 +60,9 @@ set global innodb_saved_page_number_debug = 1;
 set global innodb_fil_make_page_dirty_debug = 0;
 
 set global innodb_buf_flush_list_now = 1;
-let $shutdown_timeout=0;
---source include/shutdown_mysqld.inc
+
+--let CLEANUP_IF_CHECKPOINT=drop table t1;
+--source ../include/no_checkpoint_end.inc
```
dr-m, line 65: I think that we want the test to fail rather than be skipped when a checkpoint occurs. Please define a statement that will cause an error, such as
```sql
--let CLEANUP_IF_CHECKPOINT=drop table t1,unexpected_checkpoint;
```

mysql-test/suite/innodb/t/doublewrite_debug.test
```diff
@@ -103,8 +112,9 @@ set global innodb_saved_page_number_debug = 1;
 set global innodb_fil_make_page_dirty_debug = 0;
 
 set global innodb_buf_flush_list_now = 1;
-let $shutdown_timeout=0;
---source include/shutdown_mysqld.inc
+
+--let CLEANUP_IF_CHECKPOINT=drop table t1;
```
dr-m, line 116: We already defined `CLEANUP_IF_CHECKPOINT`. It is an environment variable in the client process; it will not be forgotten during a server restart.

mysql-test/suite/innodb/t/doublewrite_debug.test
```diff
@@ -90,7 +95,11 @@ select f1, f2 from t1;
 --echo # Test Begin: Test if recovery works if 1st page of
 --echo # system tablespace is corrupted and 2nd page as corrupted.
 
-set global innodb_log_checkpoint_now = 1;
+let $restart_parameters="--debug_dbug=+d,ib_log_checkpoint_avoid_hard";
+--source include/restart_mysqld.inc
+--source ../include/no_checkpoint_start.inc
+
+set global innodb_flush_sync=0;
```
dr-m, line 102: Swap these two to avoid an improbable test failure.

mysql-test/suite/innodb/t/doublewrite_debug.test
```diff
@@ -145,7 +156,7 @@ WHERE engine = 'innodb'
 AND support IN ('YES', 'DEFAULT', 'ENABLED');
 
 --let $ibp=--innodb-log-group-home-dir=$bugdir --innodb-data-home-dir=$bugdir
---let $ibd=$ibp --innodb-undo-tablespaces=0
+--let $ibp=$ibp --innodb-undo-tablespaces=0
```
dr-m, line 159: On merge to 11.0, we must revert the change to `doublewrite_debug.combinations` that had been made in the previous merge 8c8bce05d22e0e97e844fb2518e23aa04a544cd9.

dr-m (approved):
OK to push after addressing some minor comments.

I also tested the base revision with the following patch:
```diff
diff --git a/mysql-test/suite/innodb/t/doublewrite_debug.test b/mysql-test/suite/innodb/t/doublewrite_debug.test
index 86809cc43c0..083fad60ba0 100644
--- a/mysql-test/suite/innodb/t/doublewrite_debug.test
+++ b/mysql-test/suite/innodb/t/doublewrite_debug.test
@@ -57,7 +57,7 @@ set global innodb_saved_page_number_debug = 1;
 set global innodb_fil_make_page_dirty_debug = 0;
 
 set global innodb_buf_flush_list_now = 1;
-let $shutdown_timeout=0;
+#let $shutdown_timeout=0;
 --source include/shutdown_mysqld.inc
 
 --echo # Make the 1st page (page_no=0) and 2nd page (page_no=1)
```
This will cause failures:
```
CURRENT_TEST: innodb.doublewrite_debug
mysqltest: At line 86: query 'select f1, f2 from t1' failed: 1286: Unknown storage engine 'InnoDB'
```
The server error log contains the following:
```
2023-12-22 15:29:32 0 [Note] InnoDB: Header page consists of zero bytes in datafile: ./ibdata1, Space ID:0, Flags: 0
2023-12-22 15:29:32 0 [ERROR] InnoDB: Corrupted page [page id: space=0, page number=0] of datafile './ibdata1' could not be found in the doublewrite buffer.
2023-12-22 15:29:32 0 [ERROR] InnoDB: Plugin initialization aborted at srv0start.cc[1304] with error Data structure corruption
```
So, this test patch (which you provided) reproduced the problem that we observe very often in 10.11 and sometimes in 10.6, and we believe that this pull request should address the problem.

We are not supposed to attempt any doublewrite buffer recovery if there is no redo log to be applied. The error handling of this is something that we agreed to address in #2924. With its current revision 0e6d87b5fa6c0f756a9e9c8596a5848ecd89d5c3 I am observing similar messages, so that pull request will need to be revised.

dr-m:
Unfortunately, the test case is failing on a mandatory builder:
```
innodb.doublewrite_debug '64k,innodb,strict_full_crc32' w49 [ fail ]
        Test ended at 2023-12-22 14:58:26
CURRENT_TEST: innodb.doublewrite_debug
mysqltest: In included file "/home/buildbot/amd64-debian-11-msan/build/mysql-test/var/49/log/check.txt": 
included from /home/buildbot/amd64-debian-11-msan/build/mysql-test/suite/innodb/t/../include/no_checkpoint_end.inc at line 38:
included from /home/buildbot/amd64-debian-11-msan/build/mysql-test/suite/innodb/t/doublewrite_debug.test at line 116:
At line 2: query 'drop table t1,unexpected_checkpoint' failed: 1051: Unknown table 'test.unexpected_checkpoint'
```

mysql-test/suite/innodb/t/doublewrite_debug.test
```diff
@@ -57,8 +59,15 @@ set global innodb_saved_page_number_debug = 1;
...
---source include/shutdown_mysqld.inc
+
+--let CLEANUP_IF_CHECKPOINT=drop table t1, unexpected_checkpoint;
+
+if (`select count(*) from information_schema.system_variables where variable_name='have_sanitizer' and global_value like "MSAN%"`)
+{
+--let CLEANUP_IF_CHECKPOINT=drop table t1;
+}
+
+--source ../include/no_checkpoint_end.inc
```
dr-m, line 70: I think that this requires a `#` comment that explains why we are allowing the test to be skipped on the MSAN builder:
```
# Occasionally, a checkpoint would occur on the MSAN builder.
# We do not know the reason, because the failure can only be reproduced if there is
# enough load in that environment.
# Therefore, we allow the test to be skipped when run on MSAN.
# In other environments, we want the test to fail if a checkpoint occurs,
# so that we would catch it if it starts to happen regularly.
```
I got [a skip message](https://buildbot.mariadb.org/#/builders/572/builds/5341/steps/8/logs/stdio) after repeating the MSAN test a few times:
```
innodb.doublewrite_debug '4k,innodb,strict_crc32' w34 [ skipped ]  Extra checkpoint 1 after 67204 (0:10=67204,0:11=67284)
```

PR #2963, opened 2023-12-22
MDEV-32892 IO Thread Reports False Error When Stopped During Connecting to Primary

hsser:
- [x] *The Jira issue number for this PR is: MDEV-32892*

## Description
This PR addresses an issue where the replication system incorrectly logs a connection error (Errno: 2013) when the slave IO thread is manually stopped by the user during the connection handshake process. The problem arises because the system treats a user-initiated stop as a connection error.

The fix involves modifying the `connect_to_master `function to check the `mi->abort_slave` flag, which is set true when a user issues a stop command for the slave IO thread. The function now bypasses error logging if `mi->abort_slave` is true, thereby avoiding misleading error messages for user-initiated actions. 

## How can this PR be tested?

This PR can be tested by starting and then quickly stopping the slave IO thread, and then verifying that no connection error is logged in this case. 

## PR quality check

### DIFF
--- a/sql/slave.cc
+++ b/sql/slave.cc
@@ -7390,7 +7390,7 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
                              mi->port, 0, client_flag) == 0))
   {
     /* Don't repeat last error */
-    if ((int)mysql_errno(mysql) != last_errno)
+    if ((int)mysql_errno(mysql) != last_errno && !mi->abort_slave)
     {
       last_errno=mysql_errno(mysql);
       suppress_warnings= 0;


bnestere (changes requested):
Hi @hsser !

Thanks for the contribution! I have a couple requests for extensions:

 1)  Can you add an MTR test to show the absence of the error with your patch? You'll likely need to use the `DEBUG_SYNC` mechanism to hold the IO thread at the correct place when executing the `STOP SLAVE`; I recently pushed a similar patch which you can reference: [1407f9996](https://github.com/MariaDB/server/commit/1407f99963481d2c59c109642a6ed81a1ae201b4).

 2) The MDEV mentions that this happens in the `get_master_version_and_clock()` function as well, if the replica is stopped in this process. Can you fix that function as well (and include a test)?

bnestere:
PR has been superseded by [PR-3323|https://github.com/MariaDB/server/pull/3323].

PR #2965, opened 2023-12-22
[MDEV-32652] New sql_mode to prevent BEGIN inside a transaction

PhysicsTing:
* [x] _The Jira issue number for this PR is: MDEV-32652_

## Description

Currently a 'START TRANSACTION'/'BEGIN' statement implicitly commits the existing transaction and starts a new one. This conflicts with ANSI SQL standard (2023):

```plaintext
  If an SQL-transaction is currently active, then an exception condition
  is raised: invalid transaction state — active SQL-transaction (25001).
```

Thus change behavior to emit ER_CANT_DO_THIS_DURING_AN_TRANSACTION when such statement is issued inside an active transaction.

Before:

```plaintext
  > START TRANSACTION;
  > START TRANSACTION; ->This will work and trigger an implicitly commit
```

Now:

```plaintext
  > START TRANSACTION;
  > START TRANSACTION;
  ERROR 25000: You are not allowed to execute this command in a transaction
```

To maintain backward compatibility, this is NOT enabled by default until sql_mode is set with NO_NEW_TRANS_IN_TRANS. This flag will also be set automatically when sql_mode is set with ANSI group, because the new feature aligns with the tenet of ANSI group.

Note 1:

This will be reflected as error 25000, which is different than 25001 referred by SQL standard. This is because 25001 in MariaDB is reserved for another error when modifying isolation level of a transaction.

Note 2:

This does not error when there is a implicit multi-statement transaction (aka. autocommit=off) because doing so is essentially disabling this statement when autocommit=off and that severely breaks compatibility. See code comment for more details.

## How can this PR be tested?

All existing test cases in main, sys_vars and func_1 test suite passed. Two new test cases added into main/sql_mode and main/sql_mode_transaction for the new feature.

## Backward compatibility

To maintain backward compatibility, this is NOT enabled by default until sql_mode is set with NO_NEW_TRANS_IN_TRANS. This flag will also be set automatically when sql_mode is set with ANSI group, because the new feature aligns with the tenet of ANSI group.

## PR quality check

## Copyright


sql/event_db_repository.cc
```diff
@@ -125,7 +125,8 @@ const TABLE_FIELD_TYPE event_table_fields[ET_FIELD_COUNT] =
     "'STRICT_ALL_TABLES','NO_ZERO_IN_DATE','NO_ZERO_DATE','INVALID_DATES',"
     "'ERROR_FOR_DIVISION_BY_ZERO','TRADITIONAL','NO_AUTO_CREATE_USER',"
     "'HIGH_NOT_PRECEDENCE','NO_ENGINE_SUBSTITUTION','PAD_CHAR_TO_FULL_LENGTH',"
-    "'EMPTY_STRING_IS_NULL','SIMULTANEOUS_ASSIGNMENT')") },
+    "'EMPTY_STRING_IS_NULL','SIMULTANEOUS_ASSIGNMENT',"
+    "'TIME_ROUND_FRACTIONAL','NO_NEW_TRANS_IN_TRANS')") },
```
dlenski, line 129: Discussed this with @PhysicsTing.

He is adding `TIME_ROUND_FRACTIONAL` here as well, because it was apparently omitted/forgotten in 4447a02cf13a49876001a40ca7db8fdedb731fd5.

vuvova: No, it wasn't, that would've broken too many tests. Here: https://github.com/MariaDB/server/commit/4447a02cf13a49876001a40ca7db8fdedb731fd5#diff-d4c0aa3fa92fea6f90d7f43915d55b441a5a741134c37e8b3f7909c7ebb9abdfR3356

dlenski: > No, it wasn't, that would've broken too many tests. Here: [4447a02#diff-d4c0aa3fa92fea6f90d7f43915d55b441a5a741134c37e8b3f7909c7ebb9abdfR3356](https://github.com/MariaDB/server/commit/4447a02cf13a49876001a40ca7db8fdedb731fd5#diff-d4c0aa3fa92fea6f90d7f43915d55b441a5a741134c37e8b3f7909c7ebb9abdfR3356)

@vuvova, you're showing that it was indeed added in `sys_vars.cc`. But it _wasn't_ added in _this_ file.

vuvova: Right, that's why I deleted my comment within minutes after posting it. Sorry for confusion. Indeed, `Table_check_intact::check` only requires that `set` type in `TABLE_FIELD_TYPE` is a **prefix** of what's actually in the table, so missing new value at the end of the `set` didn't trigger an error.

PhysicsTing:
### Checks ``buildbot/amd64-debian-11-debug-ps-embedded``:

Test failed for ``innodb_gis.rtree_purge`` with error ``Test case timeout after 900 seconds``, but it passed on automatic re-try. This is the only failed test. Not sure why thic check is marked as failed.
```
innodb_gis.rtree_purge '32k,innodb'      w9 [ retry-pass ]  480054
```

### Checks ``buildbot/amd64-fedora-38-last-N-failed``:
```
At line 15: query 'select 1' failed with wrong errno <Unknown> (2026): 'TLS/SSL error: shutdown while in init', instead of  (0)...
```
This error is not in the list of expected errors defined in wait_until_connected_again.inc. I do not believe this is relevant to my change. Reason is more like the test needs better timing control. I have seen this test failure in other pull requests as well.

Please let me know if it makes sense, and help approve if so. Thanks!

dr-m:
> ### Checks `buildbot/amd64-debian-11-debug-ps-embedded`:
> Test failed for `innodb_gis.rtree_purge` with error `Test case timeout after 900 seconds`, but it passed on automatic re-try. This is the only failed test. Not sure why this check is marked as failed.

[MDEV-15284](https://jira.mariadb.org/browse/MDEV-15284) and some other open bugs demonstrate that the locking in InnoDB `SPATIAL INDEX` is broken from the beginning. You can ignore that.

> This error is not in the list of expected errors defined in line of "wait_until_connected_again.inc". I do not believe this is relevant to my change. Reason is more like the newly added test `galera_gtid_consistency` (added 1 month ago) needs better timing control. I have seen this test failure in other pull requests as well.

Galera tests are notably unstable, partly due to race conditions in the tests or in the code, partly due to some not-accounted-for error codes in case of disconnect. This one might be a simple case of [MDEV-30587](https://jira.mariadb.org/browse/MDEV-30587), which is currently assigned to @grooverdan.

ottok:
Thanks @dr-m for reviewing! Could you please clarify your feedback a bit and state what (if anything) you request to be changed or as next steps in general for this change?

dr-m:
> Thanks @dr-m for reviewing! Could you please clarify your feedback a bit and state what (if anything) you request to be changed or as next steps in general for this change?

@ottok My comments were not meant to be a review, and this ticket has not been assigned to my review. I was only trying to explain why some test failures are unrelated to these changes. Recently, there has been some effort to fix frequently failing tests; I believe that the `last-N-failed` builders are related to that.

My area of expertise is the InnoDB storage engine, and this pull request is not changing any code there.

Somewhat related to this, in [MDEV-24813](https://jira.mariadb.org/browse/MDEV-24813) I found out that `LOCK TABLES` can implicitly commit a transaction. That is partly why you need to write something funny to acquire an exclusive table lock in InnoDB:
```sql
SET autocommit = 0;
START TRANSACTION;
LOCK TABLE t1 WRITE;
```
It could be interesting to add some tests for `LOCK TABLES`. It might be not necessary to change any related code, just to document what would happen.

vuvova:
I'd prefer to avoid adding stuff to sql_mode. It's really for cases when we expect that some users forever will prefer one behavior and others — another. This is not the case here. I suggest to add it to the old_mode instead, which will disappear after a specific deprecation period and we won't have to keep the old behavior forever.

vuvova:
see https://jira.mariadb.org/browse/MDEV-32652

PR #2966, opened 2023-12-26
Maintenance: update vendored minizip code in CONNECT module to match zlib v1.3

jayaddison:
- [ ] *The Jira issue number for this PR is: N/A (pending)*

## Description
The [CONNECT module](https://mariadb.com/kb/en/connect/) includes support for reading data from `zip` format files, and the code that supports this functionality originates from the contrib `minizip` code that is packaged in the [`zlib` library](https://github.com/madler/zlib/).

This pull request refreshes the `minizip` code in the CONNECT storage module, applying changes up-to the version included in [version 1.3 of `zlib`](https://github.com/madler/zlib/commit/09155eaa2f9270dc4ed1fa13e2b4b2613e6e4851).

#### Method Notes
There seemed to be two paths to performing the maintenance here.  Either:

  * Copy-paste the `minizip` code from `zlib` v1.3 over the top of the relevant files in `server.git`, and then apply adjustments as-needed.
  OR
  * Work through the `zlib.git` commit history, applying all relevant changes to `minizip` to `server.git`.

The latter is more work but has the benefit that it highlights merge conflicts individually, which helps to understand the areas where the two codebases have diverged (as they have, to a small extent).

For reference, the key commands used to achieve this were:

```sh
$ git cherry-pick -x 94575859cf7f657f0f31aff4c50761fe3f182699...v1.3 --allow-empty  # pick the commit range, and don't pause for empty commits
$ for i in $(git status --porcelain | grep "^DU" | sed -e "s/^DU//"); do git rm $i; done; for i in $(git status --porcelain | grep "^UU" | grep -v "storage/connect" | sed -e "s/^UU//"); do git checkout 11.4 $i; done; for i in $(git status --porcelain | grep "^A" | sed -e "s/^A//"); do git reset $i; done; git clean -df ; git status;
```

The commit https://github.com/madler/zlib/commit/94575859cf7f657f0f31aff4c50761fe3f182699 is the closest-match I could identify in `zlib.git` to the version of `minizip` vendored into the CONNECT module in this repository.

The second command, approximately translated is: 'ignore any files deleted in our branch, or added in the zlib branch, and ignore most updated files _except_ for modifications to `storage/connect` related code in our branch, and show the resulting git index state'.  I used that to handle most of the conflicting commits from `zlib` near-automatically.

The `-x` option during cherry-picking is worth noting too; it includes the original `zlib` git SHA in the commit message.  That should make it easier to compare the commits I'm offering to the originals that they reference in `zlib.git` -- and in particular, help to spot any differences in how I resolved the conflicts.

Speaking of which: a few merge conflicts that may be non-obvious:

  * I preferred to continue to use the `static` modifier keyword for function declarations instead of _local_ that zlib uses - my understanding is that they're equivalent because zlib has defined local-as-static since Y2011, but given that that's not the preferred style in `server.git`, then it seemed best to prefer using the `static` modifier.
  * I chose _not_ to remove some `__attribute__(unused)` annotations, despite upstream zlib adding scoped references to the affected variables.  These may be safe to remove (and if an associated linter is clever enough, perhaps it will suggest that).
  * The `EXTERN` keyword is _not_ added to the `zipOpen3` function, nor is it declared in the `zip.h` header (this diverges from upstream zlib, where that function was added to `zip.h`).
  * I _did_ include a reference to HaikuOS in a macro define-check; not because I have any expectations about HaikuOS support in MariaDB server, but simply to keep the diff between the CONNECT module's `minizip` and upstream `zlib`'s `minizip` as minimal and/or explainable as possible.
  * There's use of implicit-scoping (absence of curly-brackets) in a [`for`...`if` construction in `zlib`](https://github.com/madler/zlib/blob/09155eaa2f9270dc4ed1fa13e2b4b2613e6e4851/contrib/minizip/zip.c#L492-L493) where the CONNECT module code has diverged and adds curly brackets (good style, in my opinion).  I've retained the CONNECT module style (curly braces added) in this case.

## How can this PR be tested?
So far I have confirmed that the updated code compiles using the provided Debian packaging on `bookworm` (using `dpkg-buildpackage`).

However: the automated tests seemed to get stuck / paused, and **did not** complete (the last item of output was `Installing system database...`).

- [x] *This is a maintenance pull request, and is neither a feature nor a specific bugfix.

## PR quality check

### DIFF
--- a/storage/connect/ioapi.c
+++ b/storage/connect/ioapi.c
@@ -14,7 +14,7 @@
         #define _CRT_SECURE_NO_WARNINGS
 #endif
 
-#if defined(IOAPI_NO_64)
+#if defined(IOAPI_NO_64) || defined(MINIZIP_FOPEN_NO_64)
 #define FOPEN_FUNC(filename, mode) fopen(filename, mode)
 #define FTELLO_FUNC(stream) ftello(stream)
 #define FSEEKO_FUNC(stream, offset, origin) fseeko(stream, offset, origin)
--- a/storage/connect/ioapi.c
+++ b/storage/connect/ioapi.c
@@ -28,8 +28,7 @@
 #include "ioapi.h"
 #include "my_attribute.h"
 
-voidpf call_zopen64 (const zlib_filefunc64_32_def* pfilefunc,const void*filename,int mode)
-{
+voidpf call_zopen64 (const zlib_filefunc64_32_def* pfilefunc, const void*filename, int mode) {
     if (pfilefunc->zfile_func64.zopen64_file != NULL)
         return (*(pfilefunc->zfile_func64.zopen64_file)) (pfilefunc->zfile_func64.opaque,filename,mode);
     else
@@ -38,8 +37,7 @@ voidpf call_zopen64 (const zlib_filefunc64_32_def* pfilefunc,const void*filename
     }
 }
 
-long call_zseek64 (const zlib_filefunc64_32_def* pfilefunc,voidpf filestream, ZPOS64_T offset, int origin)
-{
+long call_zseek64 (const zlib_filefunc64_32_def* pfilefunc,voidpf filestream, ZPOS64_T offset, int origin) {
     if (pfilefunc->zfile_func64.zseek64_file != NULL)
         return (*(pfilefunc->zfile_func64.zseek64_file)) (pfilefunc->zfile_func64.opaque,filestream,offset,origin);
     else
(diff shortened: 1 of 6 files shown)


jayaddison:
Self-review nitpick: somehow I've accidentally melded two commits -- https://github.com/madler/zlib/commit/bf2578be2d0566dc3ee47655d56b62063fe6bae6 and https://github.com/madler/zlib/commit/c4aa356742e03af6670c043905cd2c96f515ad72 -- into one commit dd1e3b1023ba5c79baa51484de6f23b82e15d43a here.

It's pedantic, but I'd like to unpick thoses and force-push the branch with a correction.

vuvova:
Thanks for the help! We've updated minizip in CONNECT to match the bundled zlib version (1.3) in ca90b61fcafeab31fe10b20437e9e94a043f2190

We'll keep it updated from now on.

PR #2967, opened 2023-12-27
MDEV-33071 : MDL deadlock loop information

northhurricane:
- [x] *The Jira issue number for this PR is: MDEV-33071*

## Description
This feature provide MDL deadlock loop printing.  Store the lock infomation when MDL deadlock 
is detected and print these infomation to logfile when user turn on the print switch.
This can help tracing back the connection by thread id to check how the deadlock happened.

## How can this PR be tested?
Provide test case mysql-test/main/print_mdl_deadlock.test to verify this feature

## PR quality check


sql/mdl.cc
```diff
@@ -304,6 +357,148 @@ bool Deadlock_detection_visitor::enter_node(MDL_context *node)
...
+    but we need to attempt a snapshot on the pointer values to avoid using NULL                   
+    values. The pointer to thd->query however, doesn't point to static memory                     
+    and has to be protected by LOCK_thd_query or risk pointing to                                 
+    uninitialized memory.                                                                      
+  */
+  const char *proc_info = thd->proc_info;
+
+  len = snprintf(header, sizeof(header),
+                 "MySQL thread id %lu, OS thread handle %lu, query id %lu",
+                 (ulong)thd->thread_id, (ulong)thd->real_id, (ulong)thd->query_id);
```
vaintroub, line 379: The ulong cast is not a good idea. Especially with thd->read_id, which on some OS is a struct. It is also not a good idea with the other members that are 64bit. ulong is , as we know, not always 64 bit (not even if OS is 64bit). ulong  is the least portable datatype, so maybe you avoid in places it is unnecessary? luckily, "long long" appears to be better portable, and there are %llu and %lld for it.

northhurricane: Yes, there are potential overflows for different operating system platforms. We re-writed it according to the portable format.

vaintroub: Not only that, THD::real_id is pthread_t, which is not required to be a number, but can be a struct, and in this case you can't cast it even to ulonglong.   https://stackoverflow.com/questions/1759794/how-to-print-pthread-t discusses exactly this problem.  I do know that we came across this problem in the past, although I can't recall the OS we had this problem (a traditional Unix,  maybe Solaris,  it might be less relevant now)

THD::os_thread_id is  more meaningful and can be output not problem, although it is only defined on Windows and Linux

northhurricane: Initially I referenced the code from functions like 'MYSQL_QUERY_LOG::write' like below
    if (my_b_printf(&log_file,
                    "# Thread_id: %lu  Schema: %s  QC_hit: %s\n"
                    "# Query_time: %s  Lock_time: %s  Rows_sent: %lu  Rows_examined: %lu\n"
                    "# Rows_affected: %lu  Bytes_sent: %lu\n",
                    (ulong) thd->thread_id, thd->get_db(),
                    ((thd->query_plan_flags & QPLAN_QC) ? "Yes" : "No"),
                    query_time_buff, lock_time_buff,
                    (ulong) thd->get_sent_row_count(),
                    (ulong) thd->get_examined_row_count(),
                    (ulong) thd->get_affected_rows(),
                    (ulong) (thd->status_var.bytes_sent - thd->bytes_sent_old)))
Codes here are not bracketed with macro and maybe this writing is platform independ.
So thread any way to print it in common? Or maybe we encapsulate a function the thread_id to string print?

northhurricane: Sorry, I found the my mistake. I don't found the thd_security_context function in marriadb maybe for my mind-mess. So I copied it from MySQL code. But the function thd_security_context already exists. So I think call this marriadb implement will resolve this problem.

northhurricane: Ohhh, in function thd_get_error_context_description, the real_id and thread_id are also casted to ulong type.😓

sql/mdl.cc
```diff
@@ -304,6 +357,87 @@ bool Deadlock_detection_visitor::enter_node(MDL_context *node)
   return m_found_deadlock;
 }
 
+#if MARIA_PLUGIN_INTERFACE_VERSION < 0x0200
```
vaintroub, line 360: Why?

sql/mdl.cc
```diff
@@ -304,6 +357,87 @@ bool Deadlock_detection_visitor::enter_node(MDL_context *node)
...
+  }
+}
+
+void Deadlock_detection_visitor::notify() {
+  char svr_error[256];
+  if (m_found_deadlock) {
+    if (m_is_reach_max_search_depth) {
+      sprintf(svr_error, "MDL deadlock occurs. %s",
+               "TOO DEEP OR LONG SEARCH IN THE LOCK TABLE WAITS-FOR GRAPH");
+      sql_print_error(svr_error);
```
vaintroub, line 399: That's a constant string, you do not need svr_error,  you can just do 
`sql_print_error("MDL deadlock occurs. TOO DEEP  OR LONG SEARCH")`
I recommend "too deep",and omit mentioning of "long", because it it more accurate. Maybe also avoid ALLCAPS without necessity.

sql/mdl.cc
```diff
@@ -304,6 +357,87 @@ bool Deadlock_detection_visitor::enter_node(MDL_context *node)
...
+      return;
+    }
+    std::stringstream ss;
+    ss << "MDL deadlock occurs.Start print MDL deadlock information."; 
+    
+    char print_strs[1024];
+    size_t print_strs_size = sizeof(print_strs);
+    snprintf(print_strs, print_strs_size,
+             "\n**************MDL DeadLock Loop Info Start*************");
+    ss << (std::string)(const char*)print_strs;
```
vaintroub, line 409: you do not need to cast to std::string, there is an overload for operator `<<` for the type `const  char*`
,see https://en.cppreference.com/w/cpp/io/basic_ostream/operator_ltlt2

sql/mdl.cc
```diff
@@ -242,6 +277,21 @@ class Deadlock_detection_visitor: public MDL_wait_for_graph_visitor
           deadlocks are even shorter typically.
   */
   static const uint MAX_SEARCH_DEPTH= 32;
+
+  /* struct of store mdl deadlock information*/
+  struct MDL_deadlock_info {
+    char *thread_info;
```
vaintroub, line 283: I do not think you need length anywhere, you can just store null terminated `char *` string.or you can store std::string.

sql/mdl.cc
```diff
@@ -304,6 +357,87 @@ bool Deadlock_detection_visitor::enter_node(MDL_context *node)
...
+    snprintf(print_strs, print_strs_size,
+             "\n**************MDL DeadLock Loop Info Start*************");
+    ss << (std::string)(const char*)print_strs;
+
+    for (uint i = 0; i < m_n_element; i++) {
+      if (!m_deadlock_loop_infos[m_n_element - i - 1]) {
+        continue;
+      }
+      snprintf(print_strs, print_strs_size, "\n=== LOOP INFO [%u]  ===", i);
+      ss << (std::string)(const char*)print_strs;
```
vaintroub, line 416: you do not need std::string cast here, and in other places.

sql/mdl.cc
```diff
@@ -1091,6 +1225,16 @@ uint MDL_ticket::get_deadlock_weight() const
          DEADLOCK_WEIGHT_DDL : DEADLOCK_WEIGHT_DML;
 }
 
+void MDL_ticket::get_wait_info(char *output, size_t *output_length) {
+  char buffer[256];
```
vaintroub, line 1229: You do not need `buffer`, can write directly to `output`, if you pass the size of the buffer.

vaintroub: Actually, maybe you just return std::string from it?

sql/mdl.cc
northhurricane: this copied from sql_class.cc
the code is

=========
#if MARIA_PLUGIN_INTERFACE_VERSION < 0x0200
/**                                                                                                 
  TODO: This function is for API compatibility, remove it eventually.                               
  All engines should switch to use thd_get_error_context_description()                              
  plugin service function.                                                                          
*/
extern "C"
char *thd_security_context(THD *thd,
                           char *buffer, unsigned int length,
                           unsigned int max_query_len)
{
  return thd_get_error_context_description(thd, buffer, length, max_query_len);
}
#endif
=========

If the macro MARIA_PLUGIN_INTERFACE_VERSION's value change to equal or bigger, this will cause a compile error.

Is it better to remove this macro from this patch?

sql/mdl.cc
northhurricane: Accepted. Use const string, and lower case the words.

gkodinov:
Closing due to inactivity. it's been more than a year since the last review round. Should you decide to continue working on this, please ask me (or any other of MariaDB committers) to re-open the pull request. Or just file a new one and mention this in it.

PR #2968, opened 2023-12-27
MDEV-33064: Sync trx->wsrep state from THD on trx start

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-33064*

## Description

InnoDB transactions may be reused after committed:

- when taken from the transaction pool
- during a DDL operation execution

In this case wsrep flag on trx object is cleared, which may cause wrong
execution logic afterwards (wsrep-related hooks are not run).

Make trx->wsrep flag initialize from THD object only once on InnoDB transaction
start and don't change it throughout the transaction's lifetime.
The flag is reset at commit time as before.

Unconditionally set wsrep=OFF for THD objects that represent InnoDB background
threads.

Make Wsrep_schema::store_view() operate in its own transaction.

Fix streaming replication transactions' fragments rollback to not switch
THD->wsrep value during transaction's execution
(use THD->wsrep_ignore_table as a workaround).

## How can this PR be tested?

MTR test is provided.

## PR quality check

### DIFF
--- a/sql/wsrep_schema.cc
+++ b/sql/wsrep_schema.cc
@@ -155,6 +155,24 @@ private:
   my_bool m_wsrep_on;
 };
 
+class wsrep_ignore_table
+{
+public:
+  wsrep_ignore_table(THD* thd)
+    : m_thd(thd)
+    , m_wsrep_ignore_table(thd->wsrep_ignore_table)
+  {
+    thd->wsrep_ignore_table= true;
+  }
+  ~wsrep_ignore_table()
+  {
+    m_thd->wsrep_ignore_table= m_wsrep_ignore_table;
+  }
+private:
+  THD* m_thd;
+  my_bool m_wsrep_ignore_table;
+};
+
 class thd_server_status
 {
 public:
@@ -739,6 +757,12 @@ int Wsrep_schema::store_view(THD* thd, const Wsrep_view& view)
   Wsrep_schema_impl::binlog_off binlog_off(thd);
   Wsrep_schema_impl::sql_safe_updates sql_safe_updates(thd);
 
+  if (trans_begin(thd, MYSQL_START_TRANS_OPT_READ_WRITE))
+  {
+    WSREP_ERROR("Failed to start transaction for store view");
+    goto out_not_started;
+  }
+
   /*
     Clean up cluster table and members table.
   */
@@ -832,7 +856,22 @@ int Wsrep_schema::store_view(THD* thd, const Wsrep_view& view)
 #endif /* WSREP_SCHEMA_MEMBERS_HISTORY */
   ret= 0;
  out:
+  if (ret)
+  {
+    trans_rollback_stmt(thd);
+    if (!trans_rollback(thd))
+    {
+      close_thread_tables(thd);
+    }
+  }
+  else if (trans_commit(thd))
+  {
+    ret= 1;
+    WSREP_ERROR("Failed to commit transaction for store view");
+  }
+  thd->release_transactional_locks();
 
+out_not_started:
   DBUG_RETURN(ret);
 }
 
@@ -1184,7 +1223,7 @@ int Wsrep_schema::remove_fragments(THD* thd,
   int ret= 0;
 
   WSREP_DEBUG("Removing %zu fragments", fragments.size());
-  Wsrep_schema_impl::wsrep_off  wsrep_off(thd);
+  Wsrep_schema_impl::wsrep_ignore_table wsrep_ignore_table(thd);
   Wsrep_schema_impl::binlog_off binlog_off(thd);
   Wsrep_schema_impl::sql_safe_updates sql_safe_updates(thd);
 
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -2892,10 +2892,6 @@ innobase_trx_init(
 		thd, OPTION_RELAXED_UNIQUE_CHECKS);
 	trx->snapshot_isolation = THDVAR(thd, snapshot_isolation) & 1;
 
-#ifdef WITH_WSREP
-	trx->wsrep = wsrep_on(thd);
-#endif
-
 	DBUG_VOID_RETURN;
 }
 
@@ -4414,9 +4410,6 @@ innobase_commit_low(
 		trx_commit_for_mysql(trx);
 	} else {
 		trx->will_lock = false;
-#ifdef WITH_WSREP
-		trx->wsrep = false;
-#endif /* WITH_WSREP */
 	}
 
 #ifdef WITH_WSREP
@@ -8735,7 +8728,10 @@ func_exit:
 	}
 
 #ifdef WITH_WSREP
-	if (error == DB_SUCCESS && trx->is_wsrep()
+	if (error == DB_SUCCESS &&
+	    /* For sequences, InnoDB transaction may not have been started yet.
+	    Check THD-level wsrep state in that case. */
+	    (trx->is_wsrep() || (!trx_is_started(trx) && wsrep_on(m_user_thd)))
 	    && wsrep_thd_is_local(m_user_thd)
 	    && !wsrep_thd_ignore_table(m_user_thd)) {
 		DBUG_PRINT("wsrep", ("update row key"));
--- a/storage/innobase/trx/trx0trx.cc
+++ b/storage/innobase/trx/trx0trx.cc
@@ -919,6 +919,7 @@ trx_start_low(
 
 #ifdef WITH_WSREP
 	trx->xid.null();
+	trx->wsrep = wsrep_on(trx->mysql_thd);
 #endif /* WITH_WSREP */
 
 	ut_a(ib_vector_is_empty(trx->autoinc_locks));
@@ -1492,6 +1493,8 @@ TRANSACTIONAL_INLINE inline void trx_t::commit_in_memory(const mtr_t *mtr)
     trx_finalize_for_fts(this, undo_no != 0);
 
 #ifdef WITH_WSREP
+  ut_ad(is_wsrep() == wsrep_on(mysql_thd));
+
   /* Serialization history has been written and the transaction is
   committed in memory, which makes this commit ordered. Release commit
   order critical section. */
--- a/mysql-test/suite/galera/r/MDEV-33064.result
+++ b/mysql-test/suite/galera/r/MDEV-33064.result
@@ -1,0 +1,26 @@
+connection node_2;
+connection node_1;
+connect con1,127.0.0.1,root,,test,$NODE_MYPORT_1;
+CREATE TABLE t1(c1 INT PRIMARY KEY) ENGINE=InnoDB;
+CREATE TABLE t1_fk(c1 INT PRIMARY KEY, c2 INT, INDEX (c2), FOREIGN KEY (c2) REFERENCES t1(c1)) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1);
+connection con1;
+SET SESSION wsrep_retry_autocommit = 0;
+SET DEBUG_SYNC = 'ib_after_row_insert SIGNAL may_alter WAIT_FOR bf_abort';
+INSERT INTO t1_fk VALUES (1, 1);
+connection node_1;
+SET DEBUG_SYNC = 'now WAIT_FOR may_alter';
+SET DEBUG_SYNC = 'lock_wait_end WAIT_FOR alter_continue';
+ALTER TABLE t1 ADD COLUMN c2 INT, ALGORITHM=INPLACE;
+connection con1;
+ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+SET DEBUG_SYNC = 'now SIGNAL alter_continue';
+connection node_1;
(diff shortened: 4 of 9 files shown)


storage/innobase/trx/trx0trx.cc
```diff
@@ -918,6 +918,9 @@ trx_start_low(
 
 #ifdef WITH_WSREP
 	trx->xid.null();
+	/* Transactions may be reused after commit, where wsrep state flag
+	gets reset. Sync it again from the corresponding THD object. */
+	trx->wsrep = wsrep_on(trx->mysql_thd);
```
dr-m, line 923: We already do this at a higher level in `innobase_trx_init()`. Why does it need to be duplicated?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1820,6 +1820,9 @@ MYSQL_THD innobase_create_background_thd(const char* name)
 	MYSQL_THD thd= create_background_thd();
 	thd_proc_info(thd, name);
 	THDVAR(thd, background_thread) = true;
+#ifdef WITH_WSREP
+	wsrep_thd_off(thd);
+#endif /* WITH_WSREP */
```
dr-m, line 1825: As far as I can tell, if we initialized this field in `create_background_thd()` (maybe we already do?), we would not need `wsrep_thd_off()` at all.

dr-m (changes requested):

storage/innobase/handler/ha_innodb.cc
denis-protivensky: There's no `create_background_thd()` in 10.4 and 10.5, where this fix is also required. As I added `wsrep_thd_off()` there, I ported it to 10.6 as well for consistency. Also, `create_background_thd()` is used not only in InnoDB, so I wanted to keep the fix local to InnoDB code and don't change higher-level code.

storage/innobase/trx/trx0trx.cc
denis-protivensky: InnoDB transactions may be reused (for example, when they are taken from the pool of transactions or during DDL operation when there may be multiple commit calls using the same `trx_t` object). In this case, wsrep state flag is reset at commit time and is never restored back (`innobase_trx_init()` doesn't seem to be called). The case with DDL was the issue that initially surfaced. The check added in `trx_t::commit_in_memory()` allowed to also catch another case with the pool of transactions.

storage/innobase/handler/ha_innodb.cc
dr-m: Is this bug reproducible in 10.4 or 10.5? The locking during DDL operations was rewritten in 10.6.

I think that we need a minimal fix for 10.6 and later releases. If a fix with `wsrep_thd_off(THD*)` is really needed for 10.4 or 10.5, such a fix should be null-merged. As far as I can tell, e7549917e17d61ced4a99f5e4c168d780c02fab1 introduced `create_background_thd()` specifically for InnoDB. Which other code is using it? What harm would clearing the flag cause for other threads?

denis-protivensky: The bug is not reproducible in 10.4 and 10.5, but I think the fix should be there as well as keeping wsrep flags in THD and trx_t in sync is the correct behavior.
I cannot find another usage of `create_background_thd()` in 10.6, but I think I saw it in more recent versions.
Since the function is in the higher-level server code, its usage may be extended at any time. I cannot reason about the consequences of clearing wsrep flag specifically for background threads, it seems ok, but usually these kind of fixes are done on the case-by-case basis.
Okay, I can remove `wsrep_thd_off()` from 10.6 upwards.

storage/innobase/trx/trx0trx.cc
denis-protivensky: We don't set the flag, we sync it from THD object. If THD object has this flag set, then it should be in sync with `trx_t`.

The initial place for the fix was exactly `trx_start_for_ddl_low()`, but after adding the check in `trx_t::commit_in_memory()` it turned out that such a fix was insufficient. And again, we don't set the flag, we sync it from THD object. I think it's a rather safe change.

denis-protivensky: In case `trx->mysql_thd` is `NULL` the flag will be `false`. So the code that doesn't have THD objects for `trx_t` assigned, will keep running with `trx->wsrep` set to `false` as before.

denis-protivensky: Thank you for the thorough review. I will need to re-explore the issue with this added knowledge.

dr-m: Non-replicated changes should be run in a separate transaction, right?

janlindstrom: There is bug on ::store_view but that is for different MDEV

dr-m: In any case, I am not going to approve something that will duplicate the assignment of `trx_t::wsrep`. The natural place for it would be when the transaction is started.

denis-protivensky: I updated the PR and made `trx->wsrep` to be set in a single place. Other changes were in the handlerton code to use `THD::wsrep` variable rather than `trx->wsrep`, and a fix to `Wsrep_schema::store_view()` to operate in it own transaction like other methods do.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7847,7 +7847,7 @@ ha_innobase::write_row(
 		m_prebuilt->autoinc_error = DB_SUCCESS;
 
 #ifdef WITH_WSREP
-		wsrep_auto_inc_inserted = trx->is_wsrep()
+		wsrep_auto_inc_inserted = wsrep_on(m_user_thd)
 			&& wsrep_drupal_282555_workaround
 			&& table->next_number_field->val_int() == 0;
```
dr-m, line 7852: This seems to introduce a function call every time when there is an `INSERT` of a row into a table that contains an `AUTO_INCREMENT` column.

Can we at least change the order of the arguments, so that we will avoid paying the extra penalty?
```c++
#ifdef WITH_WSREP
		wsrep_auto_inc_inserted = trx->is_wsrep()
		wsrep_auto_inc_inserted = UNIV_UNLIKELY(wsrep_drupal_282555_workaround)
			&& wsrep_on(m_user_thd)
			&& table->next_number_field->val_int() == 0;
#endif
```
The default value of that global parameter is `false`.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -8002,7 +8002,7 @@ ha_innobase::write_row(
 		error, m_prebuilt->table->flags, m_user_thd);
 
 #ifdef WITH_WSREP
-	if (!error_result && trx->is_wsrep()
+	if (!error_result && wsrep_on(m_user_thd)
 	    && !trx->is_bulk_insert()
 	    && wsrep_thd_is_local(m_user_thd)
 	    && !wsrep_thd_ignore_table(m_user_thd)
```
dr-m, line 8008: Can we consolidate more checks into one function call, and can we avoid introducing a mandatory function call for the case where Galera replication is not being used? The `trx->is_wsrep()` was an inline check for `trx->wsrep`.

This `if` expression is also invoking `thd_sql_command()` twice.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -8718,7 +8718,7 @@ ha_innobase::update_row(
 	}
 
 #ifdef WITH_WSREP
-	if (error == DB_SUCCESS && trx->is_wsrep()
+	if (error == DB_SUCCESS && wsrep_on(m_user_thd)
 	    && wsrep_thd_is_local(m_user_thd)
 	    && !wsrep_thd_ignore_table(m_user_thd)) {
```
dr-m, line 8723: This is replacing a check of `trx->wsrep` with function calls, which introduces a performance regression to all successful `UPDATE` operations. Can we fix the way how `trx->wsrep` is initialized?

Also, could the multiple `wsrep_thd_` function calls be consolidated into one?

The same anti-pattern occurs also in `ha_innobase::delete_row()`, affecting `DELETE` operations.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -16027,7 +16027,7 @@ ha_innobase::external_lock(
 	    && thd_sqlcom_can_generate_row_events(thd)) {
 		bool skip = false;
 #ifdef WITH_WSREP
-		skip = trx->is_wsrep() && !wsrep_thd_is_local(thd);
+		skip = wsrep_on(thd) && !wsrep_thd_is_local(thd);
 #endif /* WITH_WSREP */
```
dr-m, line 16031: Can this be consolidated in `thd_sqlcom_can_generate_row_events(thd)` and the local variable `skip` be removed altogether? This code is its only caller.

storage/innobase/trx/trx0trx.cc
```diff
@@ -918,6 +918,7 @@ trx_start_low(
 
 #ifdef WITH_WSREP
 	trx->xid.null();
+	trx->wsrep = wsrep_on(trx->mysql_thd);
```
dr-m, line 921: It would be good to add a comment why we do not trust this field in several places in `ha_innodb.cc`. If the field can’t be trusted in some cases, maybe it shouldn’t be trusted in any case and should be removed altogether, to avoid any inconsistencies? Or the initialization should be fixed.

dr-m (changes requested):

storage/innobase/handler/ha_innodb.cc
janlindstrom: I agree because default value for wsrep_drupal_282555_workaround is false (and naturally I hope it is not used if wsrep_on != ON.

storage/innobase/trx/trx0trx.cc
denis-protivensky: I think I understood what I did wrong. I accidentally reverted back `trx->wsrep = wsrep_on(thd);` assignment in `innobase_trx_init()` which we previously agreed to remove and that made all the tests pass. I will need to work on this PR more.

dr-m: Thank you. This is rather complex, but hopefully with the use of `rr record` and some data watchpoints or breakpoints you will be able to figure it out. I think that we really should try to fix the `trx->wsrep` mechanism, because it should incur much less overhead than the function calls.

sql/wsrep_schema.cc
```diff
@@ -739,6 +739,14 @@ int Wsrep_schema::store_view(THD* thd, const Wsrep_view& view)
   Wsrep_schema_impl::binlog_off binlog_off(thd);
   Wsrep_schema_impl::sql_safe_updates sql_safe_updates(thd);
 
+  bool trans_started= false;
+  if (trans_begin(thd, MYSQL_START_TRANS_OPT_READ_WRITE))
+  {
+    WSREP_ERROR("Failed to start transaction for store view");
+    goto out;
+  }
```
dr-m, line 747: We do not really need the variable `trans_started`. The `goto out` could simply be replaced with `DBUG_RETURN(ret)` or a new `out_not_started:` label added right before the final statement of the function if we prefer `goto out_not_started` to duplicating the `DBUG_RETURN(ret)`.

storage/innobase/trx/trx0trx.cc
dr-m: Can we remove the assignments to `trx->wsrep` from `innobase_trx_init()` and `innobase_commit_low()` and simply let the field be 0 when there is no active transaction?

dr-m (commented):
Thank you, this is much clearer now. But I’d still like to remove any assignment of `trx_t::wsrep` for non-active transactions.

sql/wsrep_schema.cc
denis-protivensky: Fixed.

storage/innobase/trx/trx0trx.cc
denis-protivensky: Removed all all assignments to `trx->wsrep` except the one in `trx_start_low()`.

dr-m:
In `trx_rollback_for_mysql()`, there still is an assignment `trx->wsrep= false`. I think that it should be replaced with `ut_ad(!trx->is_wsrep())`.

After that, the only assignments from/to nonzero would be in `trx_start_low()` and `trx_t::commit_in_memory()`. I think that the assignment in `trx_t::commit_in_memory()` as well as the original zero-initialization of the transaction pool should guarantee that the assertion that I am suggesting would hold in `trx_rollback_for_mysql()`. I see that we already have `ut_ad(!is_wsrep())` in `trx_t::assert_freed()`.

denis-protivensky:
Indeed, for a not started transaction it looks safe to place an assertion on `trx->wsrep` rather than the assignment. Fixed it in a separate commit.

dr-m (approved):

sysprg:
Thanks, the fixes have been merged with the head revision: https://github.com/MariaDB/server/commit/cfbd57dfb7920ed9858d3d12f74f0ba4b6f60c36 and https://github.com/MariaDB/server/commit/f6fcfc1a6a058fd7cac6bf53216ea73f3a04b22d

PR #2969, opened 2023-12-27
MDEV-33064: Sync trx->wsrep state from THD on trx start

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-33064*

## Description

This is a backport of the fix from 10.6 branch. The issue doesn't reproduce on 10.4 and 10.5, but it's good to have it there anyways.

InnoDB transactions may be reused after committed:
- when taken from the transaction pool
- during a DDL operation execution

In this case wsrep flag on trx object is cleared, which may cause wrong execution logic afterwards (wsrep-related hooks are not run).

## How can this PR be tested?

MTR test case is provided.

## PR quality check


dr-m:
16394f1aa1b4097f897b8ab01ea2064726cca059 (mariadb-10.4.34) is expected to be the last release in that series. I think that it suffices to cover this in #2968.

PR #2971, opened 2023-12-30
MDEV-32302 Window functions operate on uninitialized Item_copy objects

Olernov:
During a JOIN execution there may be employed Item_copy_* objects, for example, for tableless queries with aggregate functions. A temporary table is used to accumulate values, and those aggregated values must be copied to the Item_copy_* objects before the window functions processing takes part.
Method copy_fields() does this, but the call of copy_fields() is missing before the start of window functions processing. As a result, values are being requested from uninitialized Item_copy_* objects, which leads to either a wrong result or a crash.

This commit adds the missing call of copy_fields() to AGGR_OP::end_send() before the start of window functions step. Also the commits adds debug assertions on attempts to request a value from uninitialized Item_copy_* objects.

The name sane() for new method of Item_copy is chosen in accordance to similar functionality already implemented for Item_copy_timestamp class

- [x] *The Jira issue number for this PR is: MDEV-32302*

## Description
MDEV-32302 Window functions operate on uninitialized Item_copy objects

## How can this PR be tested?

Test case added.

## PR quality check


spetrunia:
So,

This patch adds `Item_copy::sane()` which initially returns `false`.
Item_copy itself doesn't use sane() or set_sane().  
There are two descendant classes, Item_copy_inet6 and Item_copy_string which do call set_sane() and call sane() to check.

## Point 1 
Other descendants will have sane()=false all the time. 
This doesn't look good. If we want  Item_copy_inet6 and Item_copy_string have sanity checks, they should implement them individually. If they need to share some code, how about factoring it out into a base class and then using multiple inheritance (or containment) to get it in both classes?

##  Point 2
"Sane" is poor wording.  Especially when you have sane()=false at the start of most queries.  Something like `has_value` would be a better name.

## Point 3
There is also Item_copy_timestamp::sane() introduced earlier in ca619ed123ff6caf15f48244a318ad096238fcc6.

Note that two functions have different meanings:
* Item_copy_timestamp has sane()==true all the time. If it doesn't, it's a bug.
*  Item_copy_inet6 and Item_copy_string have sane()==false until copy() is called.

Olernov:
OK, I agree that naming was not appropriate. In the next revision I suggest `is_initialized()` and `set_initialized()`. `Item_copy_timestamp::sane()` is left as it was before but the new checks `is_initialized()` are added, so all `Item_copy` descendants behave in the same way.

sql/item.h
```diff
@@ -6311,49 +6327,52 @@ class Item_copy_timestamp: public Item_copy
     null_value= tmp.is_null();
     m_value= tmp.is_null() ? Timestamp_or_zero_datetime() :
                              Timestamp_or_zero_datetime(tmp);
+#ifndef DBUG_OFF
+  set_initialized();
```
cvicentiu, line 6331: wrong indent.

PR #2974, opened 2024-01-03
MDEV-33160 show_status_array() calls various functions via incompatible pointer

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33160*

## Description
In commit b4ff64568c88ab3ce559e7bd39853d9cbf86704a the signature of mysql_show_var_func was changed, but not all functions of that type were adjusted.

When the server is configured with `cmake -DWITH_ASAN=ON` and compiled with clang, runtime errors would be flagged for invoking functions through an incompatible function pointer.

## How can this PR be tested?
This is covered by the regression test suite. I noticed the problem by running `./mtr --suite=perfschema` and searching the logs. I had applied the following patch in order to ignore all other errors that `clang -fsanitize=undefined` would catch:
```diff
diff --git a/mysql-test/mysql-test-run.pl b/mysql-test/mysql-test-run.pl
index cdd8f0bf6cd..2ad0651afb6 100755
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -4370,8 +4370,7 @@ sub extract_warning_lines ($$) {
      qr/missing DBUG_RETURN/,
      qr/Attempting backtrace/,
      qr/Assertion .* failed/,
-     qr/Sanitizer/,
-     qr/runtime error:/,
+     qr/(Address|Leak)Sanitizer/,
     );
   # These are taken from the include/mtr_warnings.sql global suppression
   # list. They occur delayed, so they can be parsed during shutdown rather
```

## PR quality check

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -7180,8 +7180,8 @@ struct my_option my_long_options[]=
   MYSQL_TO_BE_IMPLEMENTED_OPTION("validate-user-plugins") // NO_EMBEDDED_ACCESS_CHECKS
 };
 
-static int show_queries(THD *thd, SHOW_VAR *var, char *buff,
-                        enum enum_var_type scope)
+static int show_queries(THD *thd, SHOW_VAR *var, void *,
+                        system_status_var *, enum_var_type)
 {
   var->type= SHOW_LONGLONG;
   var->value= &thd->query_id;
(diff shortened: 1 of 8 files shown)


montywi (approved):
ok to push

PR #2976, opened 2024-01-04
MDEV-32745 Config upgrade helper tool

vainiovano:
- [x] *The Jira issue number for this PR is: MDEV-32745*

## Description

This PR implements a config upgrade helper tool as specified in MDEV-32745.

### Implementation

The list of valid options is parsed from `mariadbd --no-defaults --verbose --help` output by a Perl script along with valid enum and set values and stored in an include file. Generating the file is not included in the build system and is currently done manually instead.

The tool itself is currently called `my_convert` and implemented in C. The implementation is mostly based on `my_default.c` and  `my_print_defaults.c` but comes with added option validity checking and file rewriting.

### Unsolved questions

- What to do with plugins? The help output parsing Perl script currently enables all the plugins that are included in the `mariadbd` from my Arch Linux install and that can be enabled, but this list is not complete. Using a full MariaDB build could avoid this issue if all the plugins can be enabled simultaneously. One alternative option would be to read the plugin options from their source code instead.
- Should the tool try giving options to `mariadbd` to check their validity? This might not be entirely trivial if it needs to keep track of which plugins are enabled.

## How can this PR be tested?

The in-place `--update` functionality is tested by `my_convert.test`. The tool's stdout output is not tested automatically as it includes file paths and I'm not sure how to make them reproducible in the tests or to make the tests pass regardless of the selected file path.

## PR quality check


montywi:
Hi!

On Thu, Jan 4, 2024 at 2:11 PM 'Väinö Mäkelä' via Michael Widenius <
***@***.***> wrote:

>
>    - *The Jira issue number for this PR is: MDEV-32745*
>
> Description
>
> This PR implements a config upgrade helper tool as specified in MDEV-32745.
> Implementation
>
> The list of valid options is parsed from mariadbd --no-defaults --verbose
> --help output by a Perl script along with valid enum and set values and
> stored in an include file. Generating the file is not included in the build
> system and is currently done manually instead.
>
This should be automatically generated in the final version


 The tool itself is currently called my_convert and implemented in C. The
implementation is mostly based on my_default.c and my_print_defaults.c but
comes with added option validity checking and file rewriting.
A better name would be
mariadb-upgrade-config-file
(We already have a too called mariadb-conv that could easily be confused
with my_convery.  We should probably rename mariadb-conv to
mariadb-charset-conv)

 Unsolved questions

>
>    - What to do with plugins? The help output parsing Perl script
>    currently enables all the plugins that are included in the mariadbd
>    from my Arch Linux install and that can be enabled, but this list is not
>    complete. Using a full MariaDB build could avoid this issue if all the
>    plugins can be enabled simultaneously. One alternative option would be to
>    read the plugin options from their source code instead.
>    - Should the tool try giving options to mariadbd to check their
>    validity? This might not be entirely trivial if it needs to keep track of
>    which plugins are enabled.
>
>
Agree that that is not easy. What could maybe be done is to first convert
the files, then start mariadbd --help, --verbose which should write any not
understood option to stderr.   If there was any errors, then the tool could
restore the original config files.


>
>
> How can this PR be tested?
>
> The in-place --update functionality is tested by my_convert.test. The
> tool's stdout output is not tested automatically as it includes file paths
> and I'm not sure how to make them reproducible in the tests or to make the
> tests pass regardless of the selected file path.
>
The tool should have --config-file= option, with the default one being the
same as the server.

 mariadb-upgrade-config-file --config-file=/etc/my.cnf

In this case it should modify the given file and any file it includes.

I will check the current PR this week.
As this is a new project, we need in the final commit one commit for any
changes in the MariaDB server or libraries and one commit for all the code
in the tool

Regards,
Monty

vainiovano:
> Agree that that is not easy. What could maybe be done is to first convert the files, then start mariadbd --help, --verbose which should write any not understood option to stderr. If there was any errors, then the tool could restore the original config files.

Is there a way to spawn processes and read their output portably? POSIX-like systems have `posix_spawn`, but Windows will probably need something else.

> The tool should have --config-file= option, with the default one being the same as the server. mariadb-upgrade-config-file --config-file=/etc/my.cnf In this case it should modify the given file and any file it includes.

It should currently process the same files that the server reads by default, and it accepts `--defaults-file` and other `--defaults` options to set what files to process instead.

montywi:
Regrading spawning a process and reading the output:
This is exactly what client/mysqltest.c and client/mysql_upgrade.c.  See run_tool() in the later code.

vainiovano:
> What could maybe be done is to first convert the files, then start mariadbd --help, --verbose which should write any not understood option to stderr.   If there was any errors, then the tool could restore the original config files.

This is now implemented. Should there be some trickery like in `mysql_upgrade.c`'s `find_tool` to try and find the correct mariadbd to use or is it OK to just use the mariadbd from `$PATH`?

vainiovano:
The option list header is now generated automatically, but I'm still not sure what to do with the plugin list.

extra/CMakeLists.txt
```diff
@@ -46,6 +46,19 @@ ADD_CUSTOM_TARGET(GenError
                   DEPENDS
                   ${CMAKE_BINARY_DIR}/include/mysqld_error.h.tmp)
 
+
+ADD_CUSTOM_COMMAND(OUTPUT ${CMAKE_BINARY_DIR}/include/mariadbd_options.h
+  COMMAND perl ${PROJECT_SOURCE_DIR}/scripts/generate_option_list.pl
```
vaintroub, line 51: Perl is not a prerequisite for the build . But CMake is, I'd suggest to take a look at scripting in CMake, it has regular expressions, and not at all weird.

vainiovano: CMake's regexes look a lot more basic than what Perl can do, so I don't think that this is feasible

vaintroub: well, not everything is solved with regular expression. An alternative is a C++ program, not quite as elegant but does not introduce a required build dependency. On a bright side, CMake does actually know which plugins are compiled, so maybe you would not need that hardcoded list of "plugins that might be there"

dr-m: In C++, we can compile and execute Perl compatible regular expressions fairly easily, with no additional dependency. For an example, you might want to refer to `reg_replace()` in `client/mysqltest.cc`.

vainiovano: I just pushed a version written in C++. I don't think that the `--help` parsing can be easily implemented without Perl-style regexes, but they are now used from C++ with PCRE instead.

vainiovano:
@montywi You said earlier that this tool should be ready by FOSDEM. Is there still something you'd like me to do before that?

extra/mariadb_upgrade_config_file.c
```diff
@@ -0,0 +1,1171 @@
...
+     We already process --defaults* options at the beginning in
+     get_defaults_options(). So skip --defaults* options and
+     pass remaining options to handle_options().
+  */
+  org_argv+=args_used-1;
+  argc-=args_used-1;
+
+  /* Check out the args */
+  if (get_options(&argc,&org_argv))
+    cleanup_and_exit(1);
```
dr-m, line 1140:
```
CURRENT_TEST: main.mariadb_upgrade_config_file
=================================================================
==279966==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55e4cf2dd2a0 at pc 0x55e4ce247595 bp 0x7ffd0749a270 sp 0x7ffd0749a268
WRITE of size 8 at 0x55e4cf2dd2a0 thread T0
    #0 0x55e4ce247594 in init_one_value /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/mysys/my_getopt.c:1352:25
    #1 0x55e4ce244bc1 in init_variables /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/mysys/my_getopt.c:1481:7
    #2 0x55e4ce244bc1 in handle_options /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/mysys/my_getopt.c:254:5
    #3 0x55e4ce232084 in get_options /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/extra/mariadb_upgrade_config_file.c:1077:17
    #4 0x55e4ce232084 in main /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/extra/mariadb_upgrade_config_file.c:1139:7
    #5 0x7f3e00200d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: c289da5071a3399de893d2af81d6a30c62646e1e)
    #6 0x7f3e00200e3f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f) (BuildId: c289da5071a3399de893d2af81d6a30c62646e1e)
    #7 0x55e4ce171c34 in _start (/home/buildbot/amd64-ubuntu-2204-clang14-asan/build/extra/mariadb-upgrade-config-file+0xd4c34) (BuildId: 50912b41cfa853dccfab3aba5bc77e7ca677e090)
0x55e4cf2dd2a0 is located 32 bytes to the left of global variable 'opt_backup' defined in '/home/buildbot/amd64-ubuntu-2204-clang14-asan/build/extra/mariadb_upgrade_config_file.c' (0x55e4cf2dd2c0) of size 1
  'opt_backup' is ascii string ''
0x55e4cf2dd2a4 is located 0 bytes to the right of global variable 'opt_edit_mode' defined in '/home/buildbot/amd64-ubuntu-2204-clang14-asan/build/extra/mariadb_upgrade_config_file.c:52:23' (0x55e4cf2dd2a0) of size 4
SUMMARY: AddressSanitizer: global-buffer-overflow /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/mysys/my_getopt.c:1352:25 in init_one_value
```
I don’t immediately have an idea what could be wrong here. I see that it is `my_bool opt_backup`, which should be `char` or `unsigned char`.

I was interested in the ASAN run because the [grid view for this pull request](https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F2976%2Fmerge) showed some memory leaks caught by `SAFEMALLOC` on a different builder. Instead of seeing leak reports, we got this in the `cmake -DWITH_ASAN=ON` build.

extra/mariadb_upgrade_config_file.c
```diff
@@ -0,0 +1,1171 @@
...
+  my_print_help(my_long_options);
+  my_print_default_files(config_file);
+  cleanup_and_exit(0);
+}
+
+
+static my_bool
+get_one_option(const struct my_option *opt __attribute__((unused)),
+	       const char *argument __attribute__((unused)),
+               const char *filename __attribute__((unused)))
```
dr-m, line 1050: If this file had been C++ instead of C, you could simply write the following:
```c++
static my_bool get_one_option(const my_option*, const char*, const char*)
```

extra/mariadb_upgrade_config_file.c
```diff
@@ -0,0 +1,1171 @@
...
+{
+  switch (opt->id) {
+    case 'I':
+    case '?':
+      usage();
```
dr-m, line 1055: The body of the `switch` should be indented by 1 level less. The `case` and `default` labels are supposed to be at the same level as the `switch` keyword.

vainiovano: I copied this part from `my_print_defaults` and didn't bother checking the formatting. Fixed now.

extra/generate_option_list.cc
```diff
@@ -0,0 +1,359 @@
...
+#include <my_sys.h>
+
+#include <algorithm>
+#include <cstdio>
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <vector>
+#define PCRE2_STATIC 1 /* Important on Windows */
+#include "pcre2.h"
```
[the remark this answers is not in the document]
vainiovano: I just threw `PCRE_INCLUDES` into the CMakeLists. Let's see if this helps.

dr-m:
There is still [this failure](https://buildbot.mariadb.org/#/builders/222/builds/26497):
```
2024-01-31 11:10:07 0 [ERROR] mariadbd: Can't open shared library '/usr/local/mysql/lib/plugin/adt_null.so' (errno: 22, cannot open shared object file: No such file or directory)
2024-01-31 11:10:07 0 [ERROR] Couldn't load plugins from 'adt_null.so'.
2024-01-31 11:10:07 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
/bin/sh: line 1: 14582 Segmentation fault      (core dumped) ./generate_option_list /home/buildbot/amd64-rhel-7/build/sql/mariadbd /home/buildbot/amd64-rhel-7/build/include/mariadbd_options.h
```
@vaintroub Do you have any ideas how to overcome this? `adt_null` apparently is some "test" plugin, and not modified recently.

extra/generate_option_list.cc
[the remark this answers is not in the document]
dr-m: It did, and now [it is complaining about the generated code](https://buildbot.mariadb.org/#/builders/234/builds/26141/steps/8/logs/stdio):
```
   188>ClCompile:
         mariadb_upgrade_config_file.c
   188>C:\Buildbot\amd64-windows\build\include\mariadbd_options.h(5,1): error C2059: syntax error: '}' [C:\Buildbot\amd64-windows\build\extra\mariadb-upgrade-config-file.vcxproj]
C:\Buildbot\amd64-windows\build\include\mariadbd_options.h(5,1): error C2059: }; [C:\Buildbot\amd64-windows\build\extra\mariadb-upgrade-config-file.vcxproj]
C:\Buildbot\amd64-windows\build\include\mariadbd_options.h(5,1): error C2059: ^ [C:\Buildbot\amd64-windows\build\extra\mariadb-upgrade-config-file.vcxproj]
```

vainiovano:
> There is still [this failure](https://buildbot.mariadb.org/#/builders/222/builds/26497):
> 
> ```
> 2024-01-31 11:10:07 0 [ERROR] mariadbd: Can't open shared library '/usr/local/mysql/lib/plugin/adt_null.so' (errno: 22, cannot open shared object file: No such file or directory)
> 2024-01-31 11:10:07 0 [ERROR] Couldn't load plugins from 'adt_null.so'.
> 2024-01-31 11:10:07 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
> /bin/sh: line 1: 14582 Segmentation fault      (core dumped) ./generate_option_list /home/buildbot/amd64-rhel-7/build/sql/mariadbd /home/buildbot/amd64-rhel-7/build/include/mariadbd_options.h
> ```
> 
> @vaintroub Do you have any ideas how to overcome this? `adt_null` apparently is some "test" plugin, and not modified recently.

I don't think that adt_null is relevant here as it's crashing my code. For me, `generate_option_list` segfaults randomly when running with asan and doesn't crash otherwise. Is it possible to get some kind of a stack trace or a coredump while using asan?

dr-m:
> Is it possible to get some kind of a stack trace or a coredump while using asan?
`ASAN_OPTIONS=help=1 ./program_compiled_with_asan` should display the help. I’d start with
```sh
ASAN_OPTIONS=log_path=/tmp/generate_option_asan:fast_unwind_on_malloc=0:abort_on_error=1 ./generate_option_list …
```

vainiovano:
> > Is it possible to get some kind of a stack trace or a coredump while using asan?
> > `ASAN_OPTIONS=help=1 ./program_compiled_with_asan` should display the help. I’d start with
> 
> ```shell
> ASAN_OPTIONS=log_path=/tmp/generate_option_asan:fast_unwind_on_malloc=0:abort_on_error=1 ./generate_option_list …
> ```

It seems that my asan build just randomly crashes before main and is not relevant here. I can't reproduce the runtime crash.

dr-m:
I tried GCC 13.2.0 and the following:
```sh
_RR_TRACE_DIR=/dev/shm/rr ASAN_OPTIONS=abort_on_error=1:log_path=/dev/shm/asan:fast_unwind_on_malloc=0 rr record extra/generate_option_list $(pwd)/sql/mariadbd include/mariadbd_options.h
```
But there is no crash (other than the final SIGKILL at exit) in the `rr replay /dev/shm/rr/latest-trace`. Also the following did not report anything extraordinary after `run`:
```sh
ASAN_OPTIONS=detect_leaks=0:abort_on_error=1:log_path=/dev/shm/asan:fast_unwind_on_malloc=0 gdb --args extra/generate_option_list $(pwd)/sql/mariadbd include/mariadbd_options.h
```
The `detect_leaks=0` is needed because LeakSanitizer is incompatible with GDB.

I’ll try `clang-17` next.

dr-m:
`clang-17`’s `cmake -DWITH=UBSAN=ON` is complaining quite a bit, but I do not know if those are relevant. I’ll try GCC again, because it will not complain about such function pointer mismatch.
```
/mariadb/11-config/mysys/hash.c:197:20: runtime error: call to function get_sys_var_length(sys_var const*, unsigned long*, char) through pointer to incorrect function type 'unsigned char *(*)(const unsigned char *, unsigned long *, char)'
/mariadb/11-config/sql/set_var.cc:53: note: get_sys_var_length(sys_var const*, unsigned long*, char) defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /mariadb/11-config/mysys/hash.c:197:20 in 
/mariadb/11-config/strings/strings_def.h:83:25: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /mariadb/11-config/strings/strings_def.h:83:25 in 
/mariadb/11-config/strings/ctype-utf8.c:699:20: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /mariadb/11-config/strings/ctype-utf8.c:699:20 in 
/mariadb/11-config/mysys/mf_keycache.c:5950:11: runtime error: call to function init_simple_key_cache through pointer to incorrect function type 'int (*)(void *, unsigned int, unsigned long, unsigned int, unsigned int, unsigned int)'
/mariadb/11-config/mysys/mf_keycache.c:480: note: init_simple_key_cache defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /mariadb/11-config/mysys/mf_keycache.c:5950:11 in 
2024-01-31 14:31:34 0 [ERROR] mariadbd: Can't open shared library '/usr/local/mysql/lib/plugin/adt_null.so' (errno: 25, cannot open shared object file: No such file or directory)
2024-01-31 14:31:34 0 [ERROR] Couldn't load plugins from 'adt_null.so'.
/mariadb/11-config/mysys/hash.c:809:9: runtime error: call to function add_used_to_locked_mutex through pointer to incorrect function type 'char (*)(void *, void *)'
/mariadb/11-config/mysys/thr_mutex.c:716: note: add_used_to_locked_mutex defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /mariadb/11-config/mysys/hash.c:809:9 in 
/mariadb/11-config/mysys/lf_hash.cc:419:3: runtime error: call to function MDL_lock::lf_hash_initializer(st_lf_hash*, MDL_lock*, MDL_key*) through pointer to incorrect function type 'void (*)(st_lf_hash *, void *, const void *)'
/mariadb/11-config/sql/mdl.cc:677: note: MDL_lock::lf_hash_initializer(st_lf_hash*, MDL_lock*, MDL_key*) defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /mariadb/11-config/mysys/lf_hash.cc:419:3 in 
/mariadb/11-config/mysys/lf_hash.cc:312:12: runtime error: call to function tdc_hash_key(TDC_element const*, unsigned long*, char) through pointer to incorrect function type 'unsigned char *(*)(const unsigned char *, unsigned long *, char)'
/mariadb/11-config/sql/table_cache.cc:589: note: tdc_hash_key(TDC_element const*, unsigned long*, char) defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /mariadb/11-config/mysys/lf_hash.cc:312:12 in 
```

extra/generate_option_list.cc
```diff
@@ -0,0 +1,359 @@
...
+                       "test_versioning;"
+                       "type_mysql_json;"
+                       "type_mysql_timestamp;"
+                       "type_test;"
+                       "wsrep_info"
+                       "\""
+                       " --verbose"
+                       " --help";
+
+  FILE *f= my_popen(command.c_str(), "r");
```
dr-m, line 128: `cmake -DWITH_UBSAN=ON` using GCC 13.2.0 does not give any hints. But, there is a promising [-DWITH_MSAN=ON build](https://buildbot.mariadb.org/#/builders/572/builds/6184):
```
Uninitialized bytes in __interceptor_popen at offset 2 inside [0x717000000000, 805)
==12326==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55a357b8085a in call_mariadbd(char const*) /home/buildbot/amd64-debian-11-msan/build/extra/generate_option_list.cc:128:12
    #1 0x55a357b87f19 in main /home/buildbot/amd64-debian-11-msan/build/extra/generate_option_list.cc:356:20
    #2 0x7f29c62a2d09 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x23d09) (BuildId: a3780b0b8a5bf5876e31d16b0a9d8fc6ba69a1f2)
    #3 0x55a357af8519 in _start (/home/buildbot/amd64-debian-11-msan/build/extra/generate_option_list+0x25519) (BuildId: f17428483dcac03d445c4d10f154c049a08fd510)
  Uninitialized value was stored to memory at
    #0 0x55a357b25db4 in __msan_memmove (/home/buildbot/amd64-debian-11-msan/build/extra/generate_option_list+0x52db4) (BuildId: f17428483dcac03d445c4d10f154c049a08fd510)
    #1 0x7f29c669c3ca in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::append(char const*, unsigned long) (/usr/lib/x86_64-linux-gnu/libc++.so.1+0x5b3ca) (BuildId: bdd6136d41123bda4a991a8ff4594c1c11f3d56d)
  Uninitialized value was stored to memory at
    #0 0x55a357b25db4 in __msan_memmove (/home/buildbot/amd64-debian-11-msan/build/extra/generate_option_list+0x52db4) (BuildId: f17428483dcac03d445c4d10f154c049a08fd510)
    #1 0x7f29c669c818 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::push_back(char) (/usr/lib/x86_64-linux-gnu/libc++.so.1+0x5b818) (BuildId: bdd6136d41123bda4a991a8ff4594c1c11f3d56d)
  Uninitialized value was stored to memory at
    #0 0x55a357b25db4 in __msan_memmove (/home/buildbot/amd64-debian-11-msan/build/extra/generate_option_list+0x52db4) (BuildId: f17428483dcac03d445c4d10f154c049a08fd510)
    #1 0x7f29c669c818 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::push_back(char) (/usr/lib/x86_64-linux-gnu/libc++.so.1+0x5b818) (BuildId: bdd6136d41123bda4a991a8ff4594c1c11f3d56d)
  Uninitialized value was created by an allocation of 'ref.tmp' in the stack frame of function '_Z13call_mariadbdPKc'
    #0 0x55a357b804a0 in call_mariadbd(char const*) /home/buildbot/amd64-debian-11-msan/build/extra/generate_option_list.cc:70
```
I will try to reproduce this.

dr-m: I can reproduce this locally on `clang-15`:
```
cd /dev/shm/11m-config/extra && /dev/shm/11m-config/extra/generate_option_list /dev/shm/11m-config/sql/mariadbd /dev/shm/11m-config/include/mariadbd_options.h
Uninitialized bytes in __interceptor_popen at offset 2 inside [0x717000000000, 783)
==1202847==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55db7b334831 in call_mariadbd(char const*) /mariadb/11-config/extra/generate_option_list.cc:128:12
    #1 0x55db7b33c381 in main /mariadb/11-config/extra/generate_option_list.cc:356:20
    #2 0x7f9ad746e6c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x7f9ad746e784 in __libc_start_main csu/../csu/libc-start.c:360:3
    #4 0x55db7b2a9730 in _start (/dev/shm/11m-config/extra/generate_option_list+0x7d730) (BuildId: 9864fc16f8ed3435)

  Uninitialized value was stored to memory at
    #0 0x55db7b2d7ef8 in __msan_memmove (/dev/shm/11m-config/extra/generate_option_list+0xabef8) (BuildId: 9864fc16f8ed3435)
    #1 0x7f9ad76ff193 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::append(char const*, unsigned long) (/lib/x86_64-linux-gnu/libc++.so.1+0x59193) (BuildId: 9529825510c549922deaf45f996aaf182b1ecb4e)

  Uninitialized value was stored to memory at
    #0 0x55db7b2d7ef8 in __msan_memmove (/dev/shm/11m-config/extra/generate_option_list+0xabef8) (BuildId: 9864fc16f8ed3435)
    #1 0x7f9ad76ff5f2 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::push_back(char) (/lib/x86_64-linux-gnu/libc++.so.1+0x595f2) (BuildId: 9529825510c549922deaf45f996aaf182b1ecb4e)

  Uninitialized value was created by an allocation of 'ref.tmp' in the stack frame of function '_Z13call_mariadbdPKc'
    #0 0x55db7b334470 in call_mariadbd(char const*) /mariadb/11-config/extra/generate_option_list.cc:70
```

dr-m: If I got the shadow byte addresses right, there are 32 uninitialized bytes in the middle of the string:
```
(rr) p (char*)$rdi
$3 = 0x717000000000 "\"/dev/shm/11m-config/sql/mariadbd\" --no-defaults --plugin-maturity=unknown --plugin-load=\"adt_null;auth_0x0100;auth_ed25519;auth_gssapi;auth_pam;auth_test_plugin;cracklib_password_check;debug_key_mana"...
(rr) p *(char*)0x717000000000@782
$4 = "\"/dev/shm/11m-config/sql/mariadbd\" --no-defaults --plugin-maturity=unknown --plugin-load=\"adt_null;auth_0x0100;auth_ed25519;auth_gssapi;auth_pam;auth_test_plugin;cracklib_password_check;debug_key_mana"...
(rr) p *(char*)0x217000000000@782
$5 = "\000\000", '\377' <repeats 32 times>, '\000' <repeats 747 times>
```
That is, the `/dev/shm/11m-config/sql/mariadbd` is claimed to be uninitialized. It was marked as such here:
```
(rr) watch -l *(char*)0x217000000002@8
Hardware watchpoint 4: -location *(char*)0x217000000002@8
(rr) rc
Continuing.

Hardware watchpoint 4: -location *(char*)0x217000000002@8

Old value = "\377\377\377\377\377\377\377\377"
New value = "\000\000\000\000\000\000\000"
__memset_avx2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:244
244	../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S: Tiedostoa tai hakemistoa ei ole.
(rr) bt
#0  __memset_avx2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:244
#1  0x000055e5307ad7df in __msan::MsanAllocate(__sanitizer::StackTrace*, unsigned long, unsigned long, bool) ()
#2  0x000055e5307ad6fa in __msan::msan_malloc(unsigned long, __sanitizer::StackTrace*) ()
#3  0x000055e53080c149 in operator new(unsigned long) ()
#4  0x00007f33b696417c in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::append(char const*, unsigned long) () from /lib/x86_64-linux-gnu/libc++.so.1
#5  0x000055e53080d6e9 in std::__1::operator+[abi:v160006]<char, std::__1::char_traits<char>, std::__1::allocator<char> >(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&&, char const*) (
    __lhs=@0x7ffdf6110d90: {__r_ = {<std::__1::__compressed_pair_elem<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__rep, 0, false>> = {__value_ = {{__l = {{__is_long_ = 1, __cap_ = 24}, __size_ = 34, __data_ = 0x703000000000 "\"/dev/shm/11m-config/sql/mariadbd\""}, __s = {{__is_long_ = 1 '\001', __size_ = 24 '\030'}, __padding_ = 0x7ffdf6110d91 "", __data_ = "\000\000\000\000\000\000\000\"", '\000' <repeats 11 times>, "0p\000"}, __r = {__words = {49, 34, 123351460741120}}}}}, <std::__1::__compressed_pair_elem<std::__1::allocator<char>, 1, true>> = {<std::__1::allocator<char>> = {<std::__1::__non_trivial_if<true, std::__1::allocator<char> >> = {<No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}}, __rhs=0xff <error: Cannot access memory at address 0xff>)
    at /usr/include/c++/v1/string:4374
#6  call_mariadbd (mariadbd_path=<optimized out>) at /mariadb/11-config/extra/generate_option_list.cc:72
#7  0x000055e530815382 in main (argc=3, argv=0x7ffdf6111fd8) at /mariadb/11-config/extra/generate_option_list.cc:356
```
The problem appears to be the `escape_command()` call.

I was also able to reproduce this with `MSAN_OPTIONS=poison_in_dtor=0` ([MDEV-30942](https://jira.mariadb.org/browse/MDEV-30942)), which I suppose is set on buildbot anyway.

dr-m: That code was only copying already declared-as-uninitialized data. I single-stepped to the `memmove` call inside `std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::append(char const*, unsigned long)` and checked the shadow bytes of the source string (`$rsi` minus the shadow byte offset). But I did not really understand where this uninitialized data would come from.
I tried applying the following patch, with no improvement:
```diff
diff --git a/extra/generate_option_list.cc b/extra/generate_option_list.cc
index 1afd17f76cb..c557b76c89b 100644
--- a/extra/generate_option_list.cc
+++ b/extra/generate_option_list.cc
@@ -69,7 +69,8 @@ std::string read_output(FILE *f)
 std::string call_mariadbd(const char *mariadbd_path)
 {
   std::string output;
-  std::string command= escape_command(mariadbd_path) +
+  std::string command= escape_command(mariadbd_path);
+  command +=
                        " --no-defaults"
                        " --plugin-maturity=unknown"
                        " --plugin-load=\""
```
This led to the following output:
```sh
_RR_TRACE_DIR=/dev/shm/rr MSAN_OPTIONS=msan_track_origins=1 rr record /dev/shm/11m-config/extra/generate_option_list /dev/shm/11m-config/sql/mariadbd /dev/shm/11m-config/include/mariadbd_options.h
```
```
rr: Saving execution to trace directory `/dev/shm/rr/generate_option_list-6'.
Uninitialized bytes in __interceptor_popen at offset 2 inside [0x717000000000, 783)
==1207137==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x562466ffb725 in call_mariadbd(char const*) /mariadb/11-config/extra/generate_option_list.cc:129:12
    #1 0x5624670031c1 in main /mariadb/11-config/extra/generate_option_list.cc:357:20
    #2 0x7fc0e9bc06c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x7fc0e9bc0784 in __libc_start_main csu/../csu/libc-start.c:360:3
    #4 0x562466f70730 in _start (/dev/shm/11m-config/extra/generate_option_list+0x7d730) (BuildId: ea4ce4857a1fd6a2)

  Uninitialized value was stored to memory at
    #0 0x562466f9eef8 in __msan_memmove (/dev/shm/11m-config/extra/generate_option_list+0xabef8) (BuildId: ea4ce4857a1fd6a2)
    #1 0x7fc0e9e51193 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::append(char const*, unsigned long) (/lib/x86_64-linux-gnu/libc++.so.1+0x59193) (BuildId: 9529825510c549922deaf45f996aaf182b1ecb4e)

  Uninitialized value was stored to memory at
    #0 0x562466f9eef8 in __msan_memmove (/dev/shm/11m-config/extra/generate_option_list+0xabef8) (BuildId: ea4ce4857a1fd6a2)
    #1 0x7fc0e9e515f2 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>::push_back(char) (/lib/x86_64-linux-gnu/libc++.so.1+0x595f2) (BuildId: 9529825510c549922deaf45f996aaf182b1ecb4e)

  Uninitialized value was created by an allocation of 'command' in the stack frame of function '_Z13call_mariadbdPKc'
    #0 0x562466ffb470 in call_mariadbd(char const*) /mariadb/11-config/extra/generate_option_list.cc:70

SUMMARY: MemorySanitizer: use-of-uninitialized-value /mariadb/11-config/extra/generate_option_list.cc:129:12 in call_mariadbd(char const*)
Exiting
```
It could be safer to use `std::ostringstream`, which we use elsewhere.

vainiovano: > It could be safer to use `std::ostringstream`, which we use elsewhere.

Isn't that just a ([slow](https://lemire.me/blog/2023/10/19/for-processing-strings-streams-in-c-can-be-slow/)) wrapper over strings? I just pushed a version that uses `ostringstream` anyway.

dr-m: Yeah, the problem still remains, but now we got some more detailed hint:
```
cd /dev/shm/11m-config/extra && /dev/shm/11m-config/extra/generate_option_list /dev/shm/11m-config/sql/mariadbd /dev/shm/11m-config/include/mariadbd_options.h
==1215051==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55d417c97f3d in std::__1::basic_ostream<char, std::__1::char_traits<char>>& std::__1::__put_character_sequence[abi:v160006]<char, std::__1::char_traits<char>>(std::__1::basic_ostream<char, std::__1::char_traits<char>>&, char const*, unsigned long) /usr/include/c++/v1/ostream:759:13
    #1 0x55d417c86fc8 in std::__1::basic_ostream<char, std::__1::char_traits<char>>& std::__1::operator<<[abi:v160006]<std::__1::char_traits<char>>(std::__1::basic_ostream<char, std::__1::char_traits<char>>&, char) /usr/include/c++/v1/ostream:827:12
    #2 0x55d417c86fc8 in escape_command(std::__1::basic_ostringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>&, char const*) /mariadb/11-config/extra/generate_option_list.cc:39:7
    #3 0x55d417c87728 in call_mariadbd(char const*) /mariadb/11-config/extra/generate_option_list.cc:73:3
    #4 0x55d417c8f4a1 in main /mariadb/11-config/extra/generate_option_list.cc:357:20
    #5 0x7f9db520f6c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #6 0x7f9db520f784 in __libc_start_main csu/../csu/libc-start.c:360:3
    #7 0x55d417bfc640 in _start (/dev/shm/11m-config/extra/generate_option_list+0x7e640) (BuildId: 5f5e24aa74e03b40)

  Uninitialized value was created by an allocation of '__s' in the stack frame of function '_ZNSt3__124__put_character_sequenceB7v160006IcNS_11char_traitsIcEEEERNS_13basic_ostreamIT_T0_EES7_PKS4_m'
    #0 0x55d417c979e0 in std::__1::basic_ostream<char, std::__1::char_traits<char>>& std::__1::__put_character_sequence[abi:v160006]<char, std::__1::char_traits<char>>(std::__1::basic_ostream<char, std::__1::char_traits<char>>&, char const*, unsigned long) /usr/include/c++/v1/ostream:753

SUMMARY: MemorySanitizer: use-of-uninitialized-value /usr/include/c++/v1/ostream:759:13 in std::__1::basic_ostream<char, std::__1::char_traits<char>>& std::__1::__put_character_sequence[abi:v160006]<char, std::__1::char_traits<char>>(std::__1::basic_ostream<char, std::__1::char_traits<char>>&, char const*, unsigned long)
```
I got a bit better stack trace with
```sh
MSAN_OPTIONS=msan_track_origin=1:abort_on_error=1 _RR_TRACE_DIR=/dev/shm/rr rr record /dev/shm/11m-config/extra/generate_option_list /dev/shm/11m-config/sql/mariadbd /dev/shm/11m-config/include/mariadbd_options.h
rr replay /dev/shm/rr/latest-trace
```
like this:
```
(rr) set print frame-arguments scalar
(rr) bt
#0  std::__1::__put_character_sequence[abi:v160006]<char, std::__1::char_traits<char> >(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, char const*, unsigned long) (__os=..., 
    __str=__str@entry=0x7ffddd28730d "\"", __len=__len@entry=1) at /usr/include/c++/v1/ostream:758
#1  0x0000557cfdc24fc9 in std::__1::operator<<[abi:v160006]<std::__1::char_traits<char> >(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, char) (__os=..., __c=34 '"')
    at /usr/include/c++/v1/ostream:827
#2  escape_command (out=..., command=command@entry=0x7ffddd28a86f "/dev/shm/11m-config/sql/mariadbd") at /mariadb/11-config/extra/generate_option_list.cc:39
#3  0x0000557cfdc25729 in call_mariadbd (mariadbd_path=0x7ffddd28a86f "/dev/shm/11m-config/sql/mariadbd") at /mariadb/11-config/extra/generate_option_list.cc:73
#4  0x0000557cfdc2d4a2 in main (argc=3, argv=0x7ffddd288758) at /mariadb/11-config/extra/generate_option_list.cc:357
```
All MSAN shadow bytes corresponding to `mariadbd_path` are 0, meaning that the string is initialized. Besides, there should be an error when branching on an uninitialized `*command` already. The amd64-centos-7 build also still fails.

Another idea is that we have a dependency on `libfmt`. It might be used here. Or just `my_snprintf` or some variant. It might implement some suitable escaping already.

dr-m: An old-fashioned way might be to determine the length in one call, then allocate a C `char` array, and finally copy the escaped output to the buffer. Too bad that neither `snprintf()` nor `my_snprintf()` support anything that would escape string literals.

vaintroub:
> Member



> > There is still [this failure](https://buildbot.mariadb.org/#/builders/222/builds/26497):
> > ```
> > 2024-01-31 11:10:07 0 [ERROR] mariadbd: Can't open shared library '/usr/local/mysql/lib/plugin/adt_null.so' (errno: 22, cannot open shared object file: No such file or directory)
> > 2024-01-31 11:10:07 0 [ERROR] Couldn't load plugins from 'adt_null.so'.
> > 2024-01-31 11:10:07 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
> > /bin/sh: line 1: 14582 Segmentation fault      (core dumped) ./generate_option_list /home/buildbot/amd64-rhel-7/build/sql/mariadbd /home/buildbot/amd64-rhel-7/build/include/mariadbd_options.h
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @vaintroub Do you have any ideas how to overcome this? `adt_null` apparently is some "test" plugin, and not modified recently.
> 
> I don't think that adt_null is relevant here as it's crashing my code. For me, `generate_option_list` segfaults randomly when running with asan and doesn't crash otherwise. Is it possible to get some kind of a stack trace or a coredump while using asan?


> There is still [this failure](https://buildbot.mariadb.org/#/builders/222/builds/26497):
> 
> ```
> 2024-01-31 11:10:07 0 [ERROR] mariadbd: Can't open shared library '/usr/local/mysql/lib/plugin/adt_null.so' (errno: 22, cannot open shared object file: No such file or directory)
> 2024-01-31 11:10:07 0 [ERROR] Couldn't load plugins from 'adt_null.so'.
> 2024-01-31 11:10:07 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
> /bin/sh: line 1: 14582 Segmentation fault      (core dumped) ./generate_option_list /home/buildbot/amd64-rhel-7/build/sql/mariadbd /home/buildbot/amd64-rhel-7/build/include/mariadbd_options.h
> ```
> 
> @vaintroub Do you have any ideas how to overcome this? `adt_null` apparently is some "test" plugin, and not modified recently.

supply --plugin-dir? 
Side note: This is a bit unfortunate, we have many plugins spread over the filesystems, all in different directories. 
MTR fixes it for itself, in the installation layout all plugins are in the same directory. There can a fix for this, but this required a non-trivial overhaul of the CMake , and probably, MTR too.

extra/generate_option_list.cc
dr-m: I finally figured it out. Sorry, I did not realize that `cmake --build .` (or whatever we use to build) will invoke this program without pointing `LD_LIBRARY_PATH` to a `clang -fsanitize=memory` instrumented `libc++`.

If I specify a suitable `LD_LIBRARY_PATH`, then the invocation will complete without errors. I am really sorry for this time waste for us both. But I guess the easiest way to work around this might be to use C strings and avoid any C++.

The crash on amd64-centos-7 must be something else, and you’d probably need remote access to the environment to debug that.

extra/generate_option_list.cc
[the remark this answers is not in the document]
vaintroub: Sorry, I see it is already $<TARGET_FILE>:mariadbd, was looking into a different builder, appveyor where ninja is used, so the path was fine.

mysql-test/mariadb-test-run.pl
```diff
@@ -2266,6 +2266,14 @@ sub environment_setup {
 		   "$path_client_bindir/my_print_defaults");
   $ENV{'MYSQL_MY_PRINT_DEFAULTS'}= native_path($exe_my_print_defaults);
 
+  # ----------------------------------------------------
+  # mariadb-upgrade-config-file
+  # ----------------------------------------------------
+  my $exe_mariadb_upgrade_config_file=
+    mtr_exe_exists("$bindir/extra$multiconfig/mariadb-upgrade-config-file",
+		   "$path_client_bindir/mariadb-upgrade-config-file");
+  $ENV{'MARIADB_UPGRADE_CONFIG_FILE'}= native_path($exe_mariadb_upgrade_config_file);
```
vaintroub, line 2275: If you require an executable to be built for mtr, you must update the list in `ADD_DEPENDENCIES(minbuild` in the top level CMakeLists.txt with your executable.  minbuild is the minimal target that is MTR-capable, and it is used in the appveyor build

extra/generate_option_list.cc
dr-m: We might also tweak the MSAN builder configuration so that a suitable `LD_LIBRARY_PATH` (pointing only to `libc++` and not the rest) will be set during the build already.

vaintroub:
> > There is still [this failure](https://buildbot.mariadb.org/#/builders/222/builds/26497):
> > ```
> > 2024-01-31 11:10:07 0 [ERROR] mariadbd: Can't open shared library '/usr/local/mysql/lib/plugin/adt_null.so' (errno: 22, cannot open shared object file: No such file or directory)
> > 2024-01-31 11:10:07 0 [ERROR] Couldn't load plugins from 'adt_null.so'.
> > 2024-01-31 11:10:07 0 [Warning] Could not open mysql.plugin table: "Table 'mysql.plugin' doesn't exist". Some options may be missing from the help text
> > /bin/sh: line 1: 14582 Segmentation fault      (core dumped) ./generate_option_list /home/buildbot/amd64-rhel-7/build/sql/mariadbd /home/buildbot/amd64-rhel-7/build/include/mariadbd_options.h
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @vaintroub Do you have any ideas how to overcome this? `adt_null` apparently is some "test" plugin, and not modified recently.
> 
> I don't think that adt_null is relevant here as it's crashing my code. For me, `generate_option_list` segfaults randomly when running with asan and doesn't crash otherwise. Is it possible to get some kind of a stack trace or a coredump while using asan?

Just remove the plugin list. It does not work. I spent some time debugging it on Windows.

First, my_popen does not like quotes around plugin load, and it would not like either  --plugin-load="<item>;<item>" , nor it would like "--plugin-load=item;item" . Second - those quotes do nothing on Windows, command interpreter cmd.exe does not need anything for ";", it is not a special character. Third - without quotes, process starts but dies because it can't load adt_null.dll. No surprises here for me, you want plugin-load for non-existing (in --plugin-dir) shared library. You can see the same on Linux.

Now, I'd say, this hardcoded plugin list is something that would not work for now anyway ,as 1. currently all plugins are scattered around build directory, and there is no --plugin-dir they all could be loaded from, and 2. It is not certain, that any of these dynamic plugins will be built. They are all optional.  I'd propose to just leave that out of the 1st version, it is too much work to get it right.

mysql-test/mariadb-test-run.pl
vaintroub: Oh, I see, it is already done, so ignore the above

extra/mariadb_upgrade_config_file.c
```diff
@@ -0,0 +1,1178 @@
...
+{
+  const char *first= *(const char **)a;
+  const char *second= *(const char **)b;
+  return strcmp(first, second);
+}
+
+
+static my_bool mariadbd_option_exists(const char *option)
+{
+   return bsearch(&option, mariadbd_valid_options,
```
vaintroub, line 241: Can you sort these options before you apply bsearch, in the generator program?

Apparently, on Windows it produces something that is not sorted, neither are the options unique.
This is what I got (after removing --plugin-load-add, for mariadbd --help)
```
static const char *mariadbd_valid_options[]= {
"install",
"install_manual",
"install",
"install_manual",
"remove",
"remove",
"enable_named_pipe",
"standalone",
"allow_suspicious_udfs",
"alter_algorithm",
"analyze_sample_percentage",
"aria_block_size",
"aria_checkpoint_interval",
```
After `allow_suspicious_udf` , it seems sorted,  but before it there is some Windows specific stuff,  random order, with repetitions.

dr-m: Apparently we may [produce an empty array](https://buildbot.mariadb.org/#/builders/234/builds/26156):
```
         C:\Buildbot\amd64-windows\build\extra\mariadb_upgrade_config_file.c(242,12): warning C4034: sizeof returns 0 [C:\Buildbot\amd64-windows\build\extra\mariadb-upgrade-config-file.vcxproj]
       C:\Buildbot\amd64-windows\build\extra\mariadb_upgrade_config_file.c(242,12): warning C4034:            array_elements(mariadbd_valid_options), [C:\Buildbot\amd64-windows\build\extra\mariadb-upgrade-config-file.vcxproj]
       C:\Buildbot\amd64-windows\build\extra\mariadb_upgrade_config_file.c(242,12): warning C4034:            ^ [C:\Buildbot\amd64-windows\build\extra\mariadb-upgrade-config-file.vcxproj]
```

dr-m:
Something is missing for Debian packaging as well, because debhelper is complaining:
```
dh_missing: usr/bin/mariadb-upgrade-config-file exists in debian/tmp but is not installed to anywhere
```
The fix could be as simple as adding the executable file name to the file `debian/mariadb-test.install`.

dr-m:
> > Now, I'd say, this hardcoded plugin list is something that would not work for now anyway ,as a) currently all plugins are scattered around build directory, and there is no single `--plugin-dir` they all could be loaded from, and b) It is not certain, that any of these dynamic plugins will be built. They are all optional. I'd propose to just leave that out of the 1st version, it is too much work to get it right.
> 
> If the upgrade helper didn't know anything about plugins, the tool would lose its ability to check which options are valid and which aren't. It would still be able to do checking for enum and set values, but most of its functionality would be lost.

It may be meaningless to run the tool as part of the default `mtr` suites, because those tests do not require anything to be installed, and therefore the `--plugin-dir` layout would be quite different. Perhaps this could be set up to run as part of special ‘upgrade’ tests, which actually install the ‘real’ packages in various environments. Some but not all of those tests are configured in this repository. `debian/salsa-ci.yml` is an example that only runs on a Debian CI which we do not have direct access to (and it typically only runs only once for each point release).

vainiovano:
> It may be meaningless to run the tool as part of the default `mtr` suites, because those tests do not require anything to be installed, and therefore the `--plugin-dir` layout would be quite different. Perhaps this could be set up to run as part of special ‘upgrade’ tests, which actually install the ‘real’ packages in various environments. Some but not all of those tests are configured in this repository. `debian/salsa-ci.yml` is an example that only runs on a Debian CI which we do not have direct access to (and it typically only runs only once for each point release).

The options are currently read at compile time, not at runtime. If the tool acquired the list of options at runtime instead (by calling `mariadbd` from `$PATH` or whatever), then that would indeed make sense.

extra/generate_option_list.cc
dr-m: There should be a practical work-around to get the MSAN build to work: Add the following right before the definition of the function that calls `my_popen`:
```c++
#if defined(__has_feature)
#  if __has_feature(memory_sanitizer)
     __attribute__((no_sanitize("memory"))) /* we may run with uninstrumented std::string */
#  endif
#endif
```
We have that in `slide_hash()` in `zlib/deflate.c`.

ottok:
> debian/mariadb-client.install could be a better choice

Yes

> Some but not all of those tests are configured in this repository. debian/salsa-ci.yml is an example that only runs on a Debian CI which we do not have direct access to (and it typically only runs only once for each point release).

Anyone can create an account of salsa.debian.org or gitlab.com and run that CI whenever they want. For example Tuukka is doing it on regular basis.

extra/generate_option_list.cc
vainiovano: I tried putting that to both `call_mariadbd` and `escape_command`, but [the MSAN CI is still failing](https://buildbot.mariadb.org/#/builders/572/builds/6217)

dr-m: There are a couple of alternative options:

1. Modify the `CMAKE_CXX_FLAGS` for this target by removing any `-fsanitize=memory -fsanitize-memory-track-origins`. Hopefully @vaintroub can help with that.
2. Change the tool invocation in CMake, by extending the definition of `LD_LIBRARY_PATH` with the contents of another environment variable if available. The invocation could be something like this:
```sh
GENERATOR_LD_LIBRARY_PATH=/path/to_msan_instrumented_libc++ cmake --build .
```
3. (hacky) No source code change, but change the compilation invocation to something like the following:
```sh
LD_LIBRARY_PATH=/path/to_msan_instrumented_libc++ cmake --build .
```
This is risky, because if anything else that is invoked during the build happens to be dynamically linked to `libc++.so` (`clang` could be a good candidate, although on my system it uses `libstdc++` instead), then those commands would fail to start due to an unresolved symbol because the MSAN-instrumented library would depend on extra symbols that are part of each MSAN instrumented executable.

vaintroub: I do not know if 1) will work. It links with mysys, which is instrumented, and one can't remove the flags from mysys.
I'd say - just do not build it with MSAN, and make it an optional thing, also in mtr.

BTW, there are also problems with cross-compilation builds. One can't invoke TARGET_FILE:mariadbd in cross-compilation, that would not start.

dr-m: There is another hacky solution: Call `MEM_MAKE_DEFINED()` to "bless" the string after the non-inline function had copied the `mariadbd_path` to it. In any case, it might make sense to go back to the original `perl` script approach. This executable would be for maintainers only, for manually creating a list of options or option values that need to be replaced. The generated output would then be checked into the repository, a bit like the `flex` and `bison` generated files under `storage/innobase`. Not kosher, but maybe the least pain for something that is not expected to change often.

montywi:
I am now working on integrating this into MariaDB server tree

PR #2979, opened 2024-01-04
Server crashes with innodb-force-recovery=6 and enabling the innodb_truncate_temporary_tablespace_now variable

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33101*

## Description
The issue is introduced by "MDEV-28699: Shrink temporary
tablespaces without restart". SRV_FORCE_NO_LOG_REDO forces
server to read only mode and we don't initialize temporary
tablespace in read only mode.

solution: innodb_truncate_temporary_tablespace_now should 
be no-op in read only mode.

## How can this PR be tested?
./mtr innodb.temp_truncate

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/temp_truncate.test
+++ b/mysql-test/suite/innodb/t/temp_truncate.test
@@ -32,3 +32,20 @@ connection default;
 COMMIT;
 SELECT COUNT(*) FROM t1;
 DROP TABLE t1;
+
+--echo # MDEV-33101 Server crashes when starting the server with
+--echo # innodb-force-recovery=6 and enabling the
+--echo # innodb_truncate_temporary_tablespace_now variable
+
+--let $restart_parameters=--innodb-force-recovery=6
+--source include/restart_mysqld.inc
+SHOW VARIABLES LIKE "innodb_read_only";
+SET GLOBAL innodb_truncate_temporary_tablespace_now=1;
+
+--let $restart_parameters=--innodb-read-only
+--source include/restart_mysqld.inc
+SHOW VARIABLES LIKE "innodb_read_only";
+SET GLOBAL innodb_truncate_temporary_tablespace_now=1;
+
+--let $restart_parameters=
+--source include/restart_mysqld.inc
--- a/mysql-test/suite/innodb/r/temp_truncate.result
+++ b/mysql-test/suite/innodb/r/temp_truncate.result
@@ -36,3 +36,17 @@ SELECT COUNT(*) FROM t1;
 COUNT(*)
 65536
 DROP TABLE t1;
+# MDEV-33101 Server crashes when starting the server with
+# innodb-force-recovery=6 and enabling the
+# innodb_truncate_temporary_tablespace_now variable
+# restart: --innodb-force-recovery=6
+SHOW VARIABLES LIKE "innodb_read_only";
+Variable_name	Value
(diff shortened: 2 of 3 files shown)


mysql-test/suite/innodb/t/temp_truncate.test
```diff
@@ -32,3 +32,20 @@ connection default;
...
+SHOW VARIABLES LIKE "innodb_read_only";
+SET GLOBAL innodb_truncate_temporary_tablespace_now=1;
+
+--let $restart_parameters=--innodb-read-only
+--source include/restart_mysqld.inc
+SHOW VARIABLES LIKE "innodb_read_only";
+SET GLOBAL innodb_truncate_temporary_tablespace_now=1;
+
+--let $restart_parameters=
+--source include/restart_mysqld.inc
```
dr-m, line 51: The test already does one normal restart at the very start, I guess to start up with an empty temporary tablespace. If we cover the `innodb_force_recovery=6` case before that restart, then we can avoid adding this one more ‘cleanup restart’ at the end.

A server restart can easily consume 1 or 2 seconds of additional time on each test run, and possibly more on slower builders. That is why I want to avoid them.

dr-m (commented):
The code change looks OK to me, and hopefully @Thirunarayanan as well.

However, this fix should target the earliest applicable branch. This functionality was added in c507678b207a1a8ce6d99ea28fb947bed1d95795 (11.3 branch). To fix this, I think you need to force-push to this same branch a rebase of the commit to the 11.3 branch, and then maybe update the pull request to say that the target branch is 11.3 and not 11.4.

mysql-test/suite/innodb/t/temp_truncate.test
```diff
@@ -32,3 +32,20 @@ connection default;
...
+
+--let $restart_parameters=--innodb-force-recovery=6
+--source include/restart_mysqld.inc
+SHOW VARIABLES LIKE "innodb_read_only";
+SET GLOBAL innodb_truncate_temporary_tablespace_now=1;
+
+--let $restart_parameters=--innodb-read-only
+--source include/restart_mysqld.inc
+SHOW VARIABLES LIKE "innodb_read_only";
+SET GLOBAL innodb_truncate_temporary_tablespace_now=1;
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: ok, not  a bad idea. Implemented in the new patch (11.3) https://github.com/MariaDB/server/pull/2981.

mysql-test/suite/innodb/t/temp_truncate.test
mariadb-DebarunBanerjee: I prefer to keep the functional tests together but I see the point. Made the changes in https://github.com/MariaDB/server/pull/2981 11.3 branch.

mariadb-DebarunBanerjee:
The new branch contains the changes. Closing this one.
https://github.com/MariaDB/server/pull/2981

mysql-test/suite/innodb/t/temp_truncate.test
dr-m: Yes, this is a deviation from what I would consider good style, but I think that the tests are primarily for computers, not humans. Humans can find the context with `git blame` and `git show`.

PR #2980, opened 2024-01-04
[MDEV-33435] RISC-V: use RDTIME instead of RDCYCLE

aurel32:
- [ ] *The Jira issue number for this PR is: MDEV-33435*

## Description
Starting with Linux 6.6 [1], RDCYCLE is a privileged instruction on RISC-V and can't be used directly from userland. There is a sysctl option to change that as a transition period, but it will eventually disappear.

Use RDTIME instead, which while less accurate has the advantage of being synchronized between CPU (and thus monotonic) and of constant frequency.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=cc4c07c89aada16229084eeb93895c95b7eabaa3

## How can this PR be tested?

Running MariaDB on a riscv64 host running Linux 6.6+ without this change (or even just the testsuite) fails with a SIGILL. After applying this patch, the testsuite passes fully.

## PR quality check

### DIFF
--- a/include/my_rdtsc.h
+++ b/include/my_rdtsc.h
@@ -111,7 +111,7 @@ C_MODE_START
   On AARCH64, we use the generic timer base register. We override clang
   implementation for aarch64 as it access a PMU register which is not
   guaranteed to be active.
-  On RISC-V, we use the rdcycle instruction to read from mcycle register.
+  On RISC-V, we use the rdtime instruction to read from mtime register.
 
   Sadly, we have nothing for the Digital Alpha, MIPS, Motorola m68k,
   HP PA-RISC or other non-mainstream (or obsolete) processors.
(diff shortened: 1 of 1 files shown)


ottok:
I confirm this patch works and should be applied upstream. It was applied in Debian on MariaDB 1:10.11.6-2 in https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/7a8fcf712769b1737e3f5243077aad833f53cced

Before this riscv64 was failing to [build](https://buildd.debian.org/status/package.php?p=mariadb):
image

Now it is passing:
image

Ubuntu riscv64 was passing before, and still is, so no regression on older Linux kernel: https://launchpad.net/~mysql-ubuntu/+archive/ubuntu/mariadb-10.11/+builds?build_text=&build_state=all

illuusio:
Patch seems to work but how about older kernels which still have rdcycle?

grooverdan:
The TiMe and CYcle bits where both enabled for userspace in the [same commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v5.17&id=de1d7b6a51dab546160d252e47baa54adf104d4a) (5.17 + stable backports).

From the links in the bottom of #1981 all kernel/platform references I've found enable TM/CY and IR at the same time. I think its as backward comparable as we'll get without putting SIGILL handers and doing a runtime check.

grooverdan (approved):
Thanks @aurel32.  Much appreciate your diligent effort into updating the comments too.

LinuxJedi (approved):

PR #2983, opened 2024-01-07
[MDEV-33439] Fix build with libxml2 2.12

jtojnar:
- [ ] *The Jira issue number for this PR is: MDEV-33439*

## Description

libxml2 2.12.0 made `xmlGetLastError()` return `const` pointer:

https://gitlab.gnome.org/GNOME/libxml2/-/commit/61034116d0a3c8b295c6137956adc3ae55720711

Clang 16 does not like this:

    error: assigning to 'xmlErrorPtr' (aka '_xmlError *') from 'const xmlError *' (aka 'const _xmlError *') discards qualifiers
    error: cannot initialize a variable of type 'xmlErrorPtr' (aka '_xmlError *') with an rvalue of type 'const xmlError *' (aka 'const _xmlError *')

Let’s update the variables to `const`. For older versions, it will be automatically converted.

But then `xmlResetError(xmlError*)` will not like the `const` pointer:

    error: no matching function for call to 'xmlResetError'
    note: candidate function not viable: 1st argument ('const xmlError *' (aka 'const _xmlError *')) would lose const qualifier

Let’s replace it with `xmlResetLastError()`.

ALso remove `LIBXMLDOC::Xerr` protected member property.
It was introduced in 65b0e5455b547a3d574fa77b34cce23ae3bea0a0 along with the `xmlResetError` calls.
It does not appear to be used for anything.

## How can this PR be tested?

I tested this on NixOS with Clang 16.0.6 and libxml2 2.12.3.

## PR quality check

### DIFF
--- a/storage/connect/libdoc.cpp
+++ b/storage/connect/libdoc.cpp
@@ -93,7 +93,6 @@ class LIBXMLDOC : public XMLDOCUMENT {
   xmlXPathContextPtr Ctxp;
   xmlXPathObjectPtr  Xop;
   xmlXPathObjectPtr  NlXop;
-  xmlErrorPtr        Xerr;
   char              *Buf;                  // Temporary
   bool               Nofreelist;
 }; // end of class LIBXMLDOC
@@ -327,7 +326,6 @@ LIBXMLDOC::LIBXMLDOC(char *nsl, char *nsdf, char *enc, PFBLOCK fp)
   Ctxp = NULL;
   Xop = NULL;
   NlXop = NULL;
-  Xerr = NULL;
   Buf = NULL;
   Nofreelist = false;
   } // end of LIBXMLDOC constructor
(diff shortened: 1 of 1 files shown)


grooverdan (approved):
Looks good. Thanks @jtojnar.

Also note to self/other reviewers that [xmlResetLastError](https://github.com/GNOME/libxml2/commit/2b8c4a151bb6bf91c10541b0593218bca5b3d5ce#diff-f2e6c18de8171e389aeff0b8d25b2755253108a59a13baf7167c2ffa0af36a58) was added 21 years ago.

LinuxJedi:
Hi @jtojnar, all looks good to me, many thanks. One thing that we need to do before merging it is rebase against 10.5 instead of 10.4. The reason is 10.4 is approaching EoL and the final release is for critical / security fixes only.

Would you like to do this rebase? Or shall we do?

PR #2984, opened 2024-01-08
MDEV-32090 Test for null-safe equals in join

DaveGosselin-MariaDB:
This ticket is fixed by MDEV-32555 and this test captures a different use case.

### DIFF
--- a/mysql-test/main/subselect_nulls_innodb.result
+++ b/mysql-test/main/subselect_nulls_innodb.result
@@ -1,0 +1,27 @@
+#
+# MDEV-32090 Index does not handle null-safe equals operator correctly in join
+#
+CREATE TEMPORARY TABLE t1 (
+`id` int(10) unsigned NOT NULL,
+`number` int(10) unsigned DEFAULT 0,
+`name` varchar(47) DEFAULT NULL,
+`street` mediumint(8) unsigned DEFAULT NULL,
+PRIMARY KEY (`id`),
+KEY `streetNumber` (`street`,`number`,`name`)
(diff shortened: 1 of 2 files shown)


mysql-test/main/subselect_nulls_innodb.result
```diff
@@ -0,0 +1,26 @@
+#
+# MDEV-32090 Index does not handle null-safe equals operator correctly in join
+#
+CREATE TEMPORARY TABLE `Building` (
+`id` int(10) unsigned NOT NULL,
```
spetrunia, line 5: Please use table names `t1`, `t2` and so forth.
Please also drop the table at testcase end.
Rationale: 
 * all other tests do so, it's better to follow their path.
 * temporary tables are dropped when server restarts, which makes it hard to debug the SELECT query if it crashes the server (you need to re-issue the CREATE TABLE after the server is restarted)
 * It's not a good practice for each testcase to leave behind its tables (which are not guaranteed to be unique)

mysql-test/main/subselect_nulls_innodb.result
```diff
@@ -0,0 +1,26 @@
+#
+# MDEV-32090 Index does not handle null-safe equals operator correctly in join
```
spetrunia, line 2: I would also add "(testcase only)" a the end of this line as well as at the end of the first line of the commit comment.

spetrunia (changes requested):
Please address the changes and Ok to push after that.

spetrunia (approved):

PR #2991, opened 2024-01-10
MDEV-32841 Provide Innodb async IO statistics

vaintroub:
Provide some statistics about asynchronous IO reads and writes:
 - number of pending operations
 - number of completion callbacks that are currently being executed
 - number of completion callbacks that are currently queued (due to restriction on number of IO threads)
 - total number of IOs finished
 - total time to wait for free IO slot
 - total number of completions that were queued.

Also revert tpool InnoDB perfschema instrumentation (MDEV-31048) That instrumentation of cache mutex did not bring any revelation ( the mutex is taken for a couple of instructions), and made it impossible to use tpool outside of the server (e.g in mariadbimport/dump)

- [x] *The Jira issue number for this PR is: MDEV-32841

## Description
Provide detailed Innodb async IO information
Remove previous attempt on async IO instrumentation MDEV-31048 - that attempt was unsuccessful, and did not reveal any interesting information.

## How can this PR be tested?

Not amenable to automated testing, because number of IOs can't be predicted beforehand. 
Can be tested manually, using sysbench or any other load tool.
Async read is best tested loading large-ish (10-20GB) bufferpool, only in this case of thousands non-throttled async IO there is a wait for  free read slot.
Async write can be tested with any write benchmark.

## PR quality check


storage/innobase/os/os0file.cc
```diff
@@ -152,6 +157,22 @@ class io_slots
...
+  @param[in] stats pointer to the structure to fill
+*/
+void innodb_io_slots_stats(tpool::aio_opcode op, innodb_async_io_stats_t *stats)
+{
+   io_slots *slots= op == tpool::aio_opcode::AIO_PREAD? read_slots : write_slots;
+
+   stats->pending_ops = slots->pending_io_count();
+   stats->slot_wait_time_sec=
+       std::chrono::duration_cast<std::chrono::duration<float>>(
+           slots->wait_time()).count();
```
dr-m, line 172: `stats->slot_wait_time_sec` is `double`, but the `slots->wait_time()` is only `float`.

storage/innobase/os/os0file.cc
```diff
@@ -152,6 +157,22 @@ class io_slots
 static io_slots *read_slots;
 static io_slots *write_slots;
 
+/**
+  Statistics for asynchronous I/O
+  @param[in] op operation type (aio_opcode::AIO_PREAD or aio_opcode::AIO_PWRITE)
+  @param[in] stats pointer to the structure to fill
```
dr-m, line 163: Can we please remove the misleading `[in]`? The second parameter is actually for output, not input.

tpool/tpool.h
```diff
@@ -75,6 +89,7 @@ class task_group
   void set_max_tasks(unsigned int max_concurrent_tasks);
   void execute(task* t);
   void cancel_pending(task *t);
+  void get_stats(group_stats* stats);
```
dr-m, line 92: There should be a space before `*`, not after it.

tpool/tpool_structs.h
```diff
@@ -107,61 +111,65 @@ template<typename T> class cache
...
+   Wait on condition. Instrumented (wait time is recorded),
+   if timed template parameter is true.
+  */
+  void condition_wait(std::unique_lock<std::mutex>& lock)
+  {
+    if (timed)
+    {
+      auto start= std::chrono::high_resolution_clock::now();
+      m_cv.wait(lock);
+      m_wait_time+= std::chrono::high_resolution_clock::now() - start;
```
dr-m, line 124: How often is this code executed? I see that it is enabled for `io_slots::m_cache` but not anything else. This would seem to be executed only in `io_slots::acquire()` when there are no vacant slots, right?

dr-m (changes requested):
Can you extend the regression suite to demonstrate that at least some of these counters are changing in response to the workload?

storage/innobase/os/os0file.cc
vaintroub: Yes? I think the underlying type is float, but casting float to double should be ok. There are only double status variables, no floats.

tpool/tpool_structs.h
vaintroub: yes, and it is rather tricky to get this code executed, normally.  But loading large bufferpool, 10+ GB will, for the read cache the wait time will be about the same as it took bufferpool to load.  I think I did see 0.4 or so seconds for writes, but it required innodb_doublewrite=0, and innodb_write_io_threads = 2, and heavy write benchmark. This won't execute often anyway, with the only exception of unthrottled bufferpool load.

storage/innobase/os/os0file.cc
dr-m: Could we collect it in `double` precision?

tpool/tpool.h
vaintroub: done

vaintroub:
> Can you extend the regression suite to demonstrate that at least some of these counters are changing in response to the workload?

Done, but in general case, it is a little bit tricky. Especially waiting for io_slot. If you load large bufferpool, you'll see it,  but it is all non-deterministic.

mysql-test/suite/innodb/r/page_cleaner.result
```diff
@@ -3,12 +3,20 @@ SET @save_pct_lwm= @@GLOBAL.innodb_max_dirty_pages_pct_lwm;
...
+select variable_value into @innodb_async_writes_total_count_before
+from information_schema.global_status
+where variable_name='innodb_async_writes_total_count';
 CREATE TABLE t ENGINE=InnoDB SELECT * FROM seq_1_to_10000;
 SELECT variable_value>0 FROM information_schema.global_status
 WHERE variable_name = 'INNODB_BUFFER_POOL_PAGES_DIRTY';
 variable_value>0
 1
 SET GLOBAL innodb_max_dirty_pages_pct=0.0;
+select (variable_value > @innodb_async_writes_total_count_before) as innodb_async_writes_grow
```
dr-m, line 15: This will not work reliably: `'10'>'9'` does not hold. Yes, for some reason, `variable_value` is of a string type. Maybe `variable_value - @before > 0` would work, or you should use explicit conversions for both comparands, like this:
```sql
SELECT CAST(VARIABLE_VALUE AS INTEGER) INTO @var /* … */;
```
It took me quite some effort to figure out the causes of some sporadic test failures due to this.

dr-m (changes requested):
This looks OK, but the glitch in the tests needs to be fixed.

mysql-test/suite/innodb/r/page_cleaner.result
vaintroub: Thanks for figuring it out. Fixed now.

dr-m (approved):

PR #2992, opened 2024-01-11
macos: Fix CMAKE_OSX_ARCHITECTURES when not set

grooverdan:

## Description

When CMAKE_OSX_ARCHITECTURES isn't set we end up with "Packaging as: mariadb-10.4.33-osx10.19-x86_64" on arm64 builders.

This is covering up some dated code in cmake/package_name.cmake on determining the MacOS name.

Based on https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html this should set before `PROJECT`.

## How can this PR be tested?

Look at "Packaging as" in MacOS builder.

## PR quality check


grooverdan:
[didn't help](https://buildbot.mariadb.org/#/builders/590/builds/1558/steps/4/logs/stdio)
```
-- Packaging as: mariadb-10.4.33-osx10.19-x86_64
```

vaintroub:
> [didn't help](https://buildbot.mariadb.org/#/builders/590/builds/1558/steps/4/logs/stdio)
> 
> ```
> -- Packaging as: mariadb-10.4.33-osx10.19-x86_64
> ```

The documentation say "This variable should be set as a CACHE entry", https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html

i.e set(CMAKE_OSX_ARCHITECTURES. ... CACHE STRING "CMAKE OSX architectures"), although it might be easier to fix it in in cmake/package_name.cmake , where it assumed 64BIT means x86_64

grooverdan:
Thanks on `CACHE`. Says to set before PROJECT hence its location. So much cmake/package_name.cmake to reorg but I haven't got a good grasp of its requirement beyond a filename, and if anyone actually uses it.

grooverdan:
Right - as "-- Packaging as: mariadb-10.4.33-osx10.19-arm64" now.

PR #2993, opened 2024-01-11
MDEV-21829 Use packed sort keys in Unique objects

cvicentiu:
This is a patch building on top of Varun Gupta's work for MDEV-21829.

I've left Varun's patch untouched, then you can see my modifications to it.

In general, there are many areas of the code that can be improved, particularly in filesort.cc, uniques.cc, `GROUP_CONCAT` and `JSON_ARRAYAGG` implementations when it comes to code duplication. However, I consider those changes to be outside the scope of this patch, provided we want to get this fix in as soon as possible.

My approach for this patch was as follows:
1. I defined an interface contract for Keys_descriptor, a class responsible for storing (creating a binary representation) and comparing different types of keys.
2. I cleaned up as much as possible from Varun's initial patch. Particularly where there was quite a lot of duplicated code. All in all, the code is now more generic.
3. I've left some TODOs in places where I'm not sure how to proceed or have decided to not investigate further.

The patch is currently based on top of the most recent 10.6, as that's where Varun started. I can port it to 10.11 or 11.4, but I would like a review first of the whole approach, or if there are any critical show stoppers in the architecture. I've coded it in such a way that further refining is possible, but I do not want to invest more time, unless I get some rough consensus from you first.

### DIFF
--- a/sql/field.h
+++ b/sql/field.h
@@ -1126,6 +1126,11 @@ public:
   */
   virtual uint32 data_length() { return pack_length(); }
   virtual uint32 sort_length() const { return pack_length(); }
+  /*
+    returns the sort_length for a field without the suffix length bytes
+    for field with binary charset.
+  */
+  virtual uint32 sort_length_without_suffix() const { return pack_length(); }
 
   /*
     sort_suffix_length() return the length bytes needed to store the length
@@ -1485,6 +1490,9 @@ public:
   */
   virtual uint make_packed_sort_key_part(uchar *buff,
                                          const SORT_FIELD_ATTR *sort_field);
+
+  virtual uint make_packed_key_part(uchar *buff,
+                                    const SORT_FIELD_ATTR *sort_field);
 
   virtual void make_send_field(Send_field *);
 
@@ -2272,6 +2280,8 @@ public:
   bool is_packable() const override { return true; }
   uint make_packed_sort_key_part(uchar *buff,
                                  const SORT_FIELD_ATTR *sort_field)override;
+  uint make_packed_key_part(uchar *buff,
+                            const SORT_FIELD_ATTR *sort_field) override;
   uchar* pack_sort_string(uchar *to, const SORT_FIELD_ATTR *sort_field);
 };
 
@@ -4184,7 +4194,11 @@ public:
   {
     return (uint32) field_length + sort_suffix_length();
   }
-  uint32 sort_suffix_length() const override
+  uint32 sort_length_without_suffix() const override
+  {
+    return (uint32) field_length;
+  }
+  virtual uint32 sort_suffix_length() const override
   {
     return (field_charset() == &my_charset_bin ? length_bytes : 0);
   }
@@ -4530,6 +4544,10 @@ public:
   { return (uint32) (packlength); }
   uint row_pack_length() const override { return pack_length_no_ptr(); }
   uint32 sort_length() const override;
+  uint32 sort_length_without_suffix() const override
+  {
+    return (uint32)field_length;
+  }
   uint32 sort_suffix_length() const override;
   uint32 value_length() override { return get_length(); }
   uint32 max_data_length() const override
--- a/sql/item_sum.cc
+++ b/sql/item_sum.cc
@@ -34,6 +34,7 @@
 #include "sp.h"
 #include "sql_parse.h"
 #include "sp_head.h"
+#include "sql_sort.h"
 
 /**
   Calculate the affordable RAM limit for structures like TREE or Unique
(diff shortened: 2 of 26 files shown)


sql/sql_type.h
```diff
@@ -4113,6 +4113,12 @@ class Type_handler
 
   /*
     create a compact size key part for a sort key
+
+    @param to                   buffer to store value of keypart
+    @param item                 item corresponding to the keypart
+    @param sort_field           sort field structure
+    @param tmp_buffer           temporary buffer to store the packed value
```
spetrunia, line 4120: There is no parameter with this name.

sql/sql_type.h
```diff
@@ -4113,6 +4113,12 @@ class Type_handler
 
   /*
     create a compact size key part for a sort key
+
+    @param to                   buffer to store value of keypart
+    @param item                 item corresponding to the keypart
```
spetrunia, line 4118: Do we get the value from here? If yes I think it's better to mention this explicitly.

sql/uniques.h
```diff
@@ -16,20 +16,228 @@
...
+  virtual bool setup_for_field(THD *thd, Field *field);
+
+  uint get_max_key_length() const { return max_length; };
+  Sort_keys *get_keys() { return sort_keys; }
+  SORT_FIELD *get_sortorder() { return sortorder; }
+
+  /*
+    Create a keys record according to this descriptor's storage format.
+
+    The data is taken from the SORT_FIELD. The SORT_FIELD's values (underlying
```
spetrunia, line 76: Un-closed bracket?

sql/uniques.h
```diff
@@ -16,20 +16,228 @@
...
+    table::record[0] must be set before calling this function.
```
spetrunia, line 77: s/table/TABLE/?

sql/uniques.h
```diff
@@ -16,20 +16,228 @@
...
+
+    The record is then to be used within a Unique object to store the result.
+  */
```
spetrunia, line 80: "a Unique object"? Do you mean not this (`*this`) object but some other?

sql/item_sum.cc
```diff
@@ -4491,61 +4441,102 @@ String* Item_func_group_concat::val_str(String* str)
...
-{
-  return skip_nulls() ? 0 : table->s->null_bytes;
+  /*
+    TODO varun: we can introduce a heuristic here, no need to do packing if we
+    have small VARCHAR or CHAR
+  */
+  *total_length= tot_length;
+  /*
+    Unique::size_of_lengt_field is the length bytes to store the packed length
+    for each record inserted in the Unique tree
```
spetrunia, line 4535: There is no `Unique::size_of_lengt_field` ?

sql/uniques.h
```diff
@@ -38,45 +246,118 @@ class Unique :public Sql_alloc
...
+
+  SORT_INFO *get_sort() { return &sort; }
+
+  Unique(Keys_descriptor *desc,
+         size_t max_in_memory_size_arg,
+         uint min_dupl_count_arg);
+  ~Unique();
+  ulong elements_in_tree() { return tree.elements_in_tree; }
+
+  bool unique_add(uchar *ptr)
```
spetrunia, line 342: So, now there are two methods:
```
private:
    bool unique_add(void *ptr, uint key_size);
public:
    bool unique_add(uchar *ptr);
```

Am I the only one who thinks it's confusing?

spetrunia: Actually, they have different meanings.
`bool unique_add(uchar *ptr)` - this one accepts the in some "source format".   It uses `keys_descriptor` to convert it to some internal representation and then `bool unique_add(void *ptr, uint key_size);`  is called to insert that into the RB-tree...

sql/uniques.h
```diff
@@ -38,45 +246,118 @@ class Unique :public Sql_alloc
...
+      return false;
+    return record_size > space_left();
+  }
+
+  /*
+    @brief
+      Add a record to the Unique tree
+    @param
+      ptr                      key value
+      size                     length of the key
```
spetrunia, line 291: The argument is `key_size`, not `size` and this is not a valid doxygen markup.

sql/uniques.h
cvicentiu: As a user of the unique class, you don't have to care conceptually about the unique_add method with two parameters.

It's part of the internal implementation. If you think this would be better named unique_add_internal or add_internal or something else, I could change it to that, but I consider the fact that it's marked private a sufficiently strong of a hint.

When writing this code I took two stances: as a class implementer, the names of private methods are of no concern to the user of said class, at all. I didn't care about any conflicts.
As a user of a class, the code strives to rely strictly on the public methods, not rely on what they internally do. There were cases when Varun's patch required a certain order of methods being called for them to work. I tried to minimize that.

cvicentiu: Also, one more thing: the unique object (albeit not completely clear cut with this current iteration) doesn't actually handle the "internal representation format". That part is delegated to the Keys_descriptor. The unique object just handles storing stuff in the tree, optionally on disk when memory limit exceeded and getting it back out. The other details of fields, packed values, etc. should be handled by the Keys_descriptor.

sql/field.h
```diff
@@ -1126,6 +1126,11 @@ class Field: public Value_source
   */
   virtual uint32 data_length() { return pack_length(); }
   virtual uint32 sort_length() const { return pack_length(); }
+  /*
+    returns the sort_length for a field without the suffix length bytes
+    for field with binary charset.
```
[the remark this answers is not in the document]
cvicentiu: When comparing varchars, we stop after a certain length and the remainder is the suffix. That's something I need to clear up. I'll add some comments.

sql/uniques.h
cvicentiu: This is not a Unique object class, it's a Keys_descriptor. It is indeed passed to the constructor of Unique, but the `Keys_descriptor` doesn't hold a reference to the Unique object, so no, not any `*this` object. This object is just a "binary-data-generator + comparison function for generated data"

cvicentiu: Ok, some leftover comments from Varun's patch make this confusing. I will reword them.

sql/sql_statistics.h
```diff
@@ -144,6 +144,8 @@ double get_column_range_cardinality(Field *field,
                                     uint range_flag);
 bool is_stat_table(const LEX_CSTRING *db, LEX_CSTRING *table);
 bool is_eits_usable(Field* field);
+uint get_offset_to_value(Field *field);
+uchar *get_buffer_end(uchar *to);
```
spetrunia, line 149: These are only used in sql_statistics.cc (once each) so should at least be made static.

Maybe they also should be named differently and/or moved somewhere because it looks to me like these functions assume some storage format but it is not immediately clear where that format is defined, etc.

cvicentiu: Good point. I usually clean these sorts of functions up, I missed this.

PR #2997, opened 2024-01-11
Small cleanup of replication code (log.cc)

an3l:
- Remove unused variable `file` from `MYSQL_BIN_LOG::open()`
- Remove wrong argument in comment
- Remove single/trivial call of function MYSQL_BIN_LOG::init() and remove function
- Remove `delay_close` variable from `MYSQL_BIN_LOG::new_file_impl()` and adapt the logic to `is_relay_log`
- Remove single jump to label `end2` and use code instead; remove label `end2`
- Remove unncesseray assignemt to `close_on_error`

## PR quality check

### DIFF
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -3574,16 +3574,6 @@ void MYSQL_BIN_LOG::cleanup()
 }
 
 
-/* Init binlog-specific vars */
-void MYSQL_BIN_LOG::init(ulong max_size_arg)
-{
-  DBUG_ENTER("MYSQL_BIN_LOG::init");
-  max_size= max_size_arg;
-  DBUG_PRINT("info",("max_size: %lu", max_size));
-  DBUG_VOID_RETURN;
-}
-
-
 void MYSQL_BIN_LOG::init_pthread_objects()
 {
   MYSQL_LOG::init_pthread_objects();
@@ -3775,7 +3765,7 @@ bool MYSQL_BIN_LOG::open(const char *log_name,
     DBUG_RETURN(1);                            /* all warnings issued */
   }
 
-  init(max_size_arg);
+  max_size= max_size_arg;
 
   open_count++;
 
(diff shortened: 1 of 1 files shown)


sql/log.cc
```diff
@@ -5513,8 +5493,7 @@ int MYSQL_BIN_LOG::new_file_impl()
     last_used_log_number--;
   }
 
-end2:
-  if (delay_close)
+  if (!is_relay_log)
```
cvicentiu, line 5496: And this also changes behaviour.

For example, in the previous code:
`is_relay_log == false`
`(error= write_event(&r)) != 0`

In this case, delay_close would remain false. With your patch, the file is always closed. That is not what the code did before. Is that acceptable or not? I do not know, but because of this, this is not a cleanup, rather a logic change.

sql/log.cc
```diff
@@ -3694,7 +3684,6 @@ bool MYSQL_BIN_LOG::open(const char *log_name,
                          bool null_created_arg,
                          bool need_mutex)
 {
-  File file= -1;
```
cvicentiu, line 3697: Ok!

sql/log.cc
```diff
@@ -3571,16 +3571,6 @@ void MYSQL_BIN_LOG::cleanup()
 }
 
 
-/* Init binlog-specific vars */
-void MYSQL_BIN_LOG::init(ulong max_size_arg)
```
cvicentiu, line 3575: Ok!

cvicentiu (commented):
Please double check your changes, some actually change how the code executes.

bnestere (changes requested):
Hi @an3l !

Thanks for the contribution! Though I'm not sure that we should go through with it, as it has some potentially breaking/conflicting changes. See my notes.

sql/log.cc
andrelkin: The API argument should not apply here. E.g mysqlbinlog is an external program and would obviously feel good.

bnestere (approved):

PR #2998, opened 2024-01-11
MDEV-29369: rpl.rpl_semi_sync_shutdown_await_ack fails regularly with…

bnestere:
… Result content mismatch

This test was prone to failures for a few reasons, summarized below:

 1) MDEV-32168 introduced “only_running_threads=1” to
slave_stop.inc, which allowed the stop logic to bypass an attempting-to-reconnect IO thread. That is, the IO thread could realize the master shutdown in `read_event()`, and thereby call into `try_to_reconnect()`. This would leave the IO thread up when the test expected it to be stopped. The fix for this is two-part. First, rather than stop the io thread, wait for it to realize the master shutdown so it can turn itself off automatically. Second, change the test configuration `master_retry_count` to 2 (default 10000) so it waits a reasonable amount of time.

 2) On slow systems (or those running profiling tools, e.g. MSAN),
the waiting-for-ack transaction can complete before the system processes the `SHUTDOWN WAIT FOR ALL SLAVES`. There was shutdown preparation logic in-between the transaction and shutdown itself, which contributes to this problem. This patch also moves this preparation logic before the transaction, so there is less to do in-between the calls.

 3) Changed work-around for MDEV-28141 to use debug_sync instead
of sleep delay, as it was still possible to hit the bug on very slow systems.

Also adjusted for MDEV-32031 fix to change MTR restoration variables to user variables

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
+++ b/mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
@@ -3,7 +3,6 @@
 # replicas before shutting down.
 #
 # Parameters:
-#   server_1_dbug (string)          Debug setting for primary (server 1)
 #   server_2_dbug (string)          Debug setting to simulate delay or error on
 #                                   the first replica (server 2)
 #   server_3_dbug (string)          Debug setting to simulate delay or error on
@@ -32,8 +31,8 @@ while (`SELECT $i <= $slave_last`)
   --connection server_$i
   --sync_with_master
 
-  set global rpl_semi_sync_slave_enabled = 1;
   source include/stop_slave.inc;
+  set global rpl_semi_sync_slave_enabled = 1;
   source include/start_slave.inc;
   show status like 'Rpl_semi_sync_slave_status';
 
@@ -56,8 +55,6 @@ show status like 'Rpl_semi_sync_master_status';
 show status like 'Rpl_semi_sync_master_clients';
 
 --echo #-- Prepare servers to simulate delay or error
---connection server_1
---eval SET @@GLOBAL.debug_dbug= $server_1_dbug
 --connection server_2
 --eval SET @@GLOBAL.debug_dbug= $server_2_dbug
 --connection server_3
@@ -65,6 +62,14 @@ show status like 'Rpl_semi_sync_master_clients';
 
 --echo #--
 --echo #-- Test begins
+
+--connection server_1_con2
+--echo #-- Give enough time after timeout/ack received to query yes_tx/no_tx
+SET @@GLOBAL.debug_dbug= "+d,delay_shutdown_phase_2_after_semisync_wait";
+
+--write_file $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+wait
+EOF
 
 --connection server_1
 --echo #-- Begin semi-sync transaction
@@ -75,14 +80,6 @@ show status like 'Rpl_semi_sync_master_clients';
 let $status_var= Rpl_semi_sync_master_wait_sessions;
 let $status_var_value= 1;
 source include/wait_for_status_var.inc;
-show status like 'Rpl_semi_sync_master_wait_sessions';
-
---write_file $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
-wait
-EOF
-
---echo #-- Give enough time after timeout/ack received to query yes_tx/no_tx
-SET @@GLOBAL.debug_dbug= "+d,delay_shutdown_phase_2_after_semisync_wait";
 
 --echo #-- Begin master shutdown
 --send SHUTDOWN WAIT FOR ALL SLAVES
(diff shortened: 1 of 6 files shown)


mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
```diff
@@ -142,12 +200,11 @@ EOF
...
-let $status_var= Rpl_semi_sync_master_clients;
-let $status_var_value= 0;
-source include/wait_for_status_var.inc;
---eval SET @@GLOBAL.rpl_semi_sync_master_enabled = 0
-show status like 'Rpl_semi_sync_master_status';
+# Because we restarted, we have to store the default values again for later
+# restoration at cleanup
+set @sav_master_timeout= @@global.rpl_semi_sync_master_timeout;
+set @sav_enabled_master= @@GLOBAL.rpl_semi_sync_master_enabled;
+set @sav_master_dbug= @@GLOBAL.debug_dbug;
```
knielsen, line 207: In this case (@user variables lost due to server restart), I think it's better to simply use mysqltest --let $save_xxx variables that preserve their value, and simply use --disable-query-log to avoid a possibly changing default value to get into the .result file.

Or even simpler, if the server restarts, there's no need to restore the values, the restart already resets all variables to default. If the values are changed again after the restart, it's sufficient to save just after restart, avoiding the issue mentioned in the comment here.

mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
```diff
@@ -111,19 +108,80 @@ show status like 'Rpl_semi_sync_master_no_tx';
...
+  --source include/stop_slave_sql.inc
+}
+if ($server_2_expect_row_count)
+{
+  --echo #-- Waiting for IO thread to realize master shutdown and stop
+  --let $slave_io_errno=2003
+  --source include/wait_for_slave_io_error.inc
+  --source include/stop_slave_sql.inc
+  #--source include/stop_slave.inc
+}
```
knielsen, line 163: Here, is it really necessary to wait_for_slave_io_error?
If I understand correctly, you set the retry count low and wait for the retry count to give up and the IO THREAD to fail with reconnect error.
Isn't it possible to just --source include/stop_slave.inc ?
I fixed wait_for_slave_io_to_stop.inc to disregards an error left by the IO thread, since this can happen in a number of scenarios.

In general, I think it should be possible to just --source include/stop_slave_io.inc to stop the IO thread in most normal cases. If there's a special reason that this is not possible in this test, please add a comment explaining why this is so.
If there is a

knielsen (changes requested):
A few suggestions for possible simplifications.

It's important that most replication tests can use simple stop_slave.inc / start_slave.inc , getting replication tests stable is plenty difficult already. That's why I'm careful about if this can be done simpler.

I did not fully follow the details of the problems you described in the patch around stop/start, so there might be some special things going on in this test case that I missed. If so it's probably ok, this change only concerns a single test. But ideally it should be possible to stop and start slave threads easily, even if the master is restarted by the test.

mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
bnestere: We could do `stop_slave.inc` now, but in `wait_for_slave_io_to_stop.inc` which is used, as you mention, your fix includes:

```
# Disabled, as IO errors are left behind when a normal STOP SLAVE interrupts
# the initial communication between the IO thread and the master (MDEV-32892).
#if (!$rpl_allow_error)
#{
#  --let $slave_error_param= Last_IO_Errno
#}
```

so once we re-enable it after MDEV-32892, the test would fail again. We could preempt that by setting `rpl_allow_error` now too, and it would save a few seconds from the test run as we don't have to wait for the disconnection. But I slightly prefer consistent test behavior over that. Though if you have a strong preference to use `stop_slave.inc`, I'm also okay with that.

mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
bnestere: Ah yeah, disable query log indeed takes care of the awkwardness. Thanks!

mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
knielsen: I don't understand, what do you mean with "consistent test behaviour"?

If we want to stop the slave, then using stop_slave.inc is the consistent way, and the stop_slave.inc script needs to be working so that it is able to stop the slave. If the script is not able to do that now or in the future, it needs to be fixed.

If it's expected that the IO thread may or may not have reached an error condition at the point where we stop it, that's fine.

Setting the IO thread to retry connections to a shut-down master until giving up seems to only add unnecessary complexity, I don't see how that adds value to the test?

bnestere: > If it's expected that the IO thread may or may not have reached an error condition at the point where we stop it, that's fine.

The "may or may not" part is the inconsistent behavior. As if something does go wrong, there would be a clear expectation being broken, rather than having to debug through different possible outcomes.

Though note `stop_slave.inc` works fine, no adjustment necessary there.

bnestere: @knielsen I updated the patch to use stop slave, though I kept the stop of the sql and io threads separate to be explicit in the intent that only the io thread can error.

knielsen (approved):
Sure Brandon, I'm fine with the changes, I didn't spend the time to fully follow all the details of the test and changes, but looks fine to me.

bnestere:
Pushed via a different branch into 10.5: 03d1346e7f535285bd84e348ece0f128998e0796

PR #2999, opened 2024-01-11
MDEV-30975 Wrong result with cross Join given join order

DaveGosselin-MariaDB:
For queries with derived tables populated having some side-effect, we will fill such a derived table more than once, but without clearing its rows.  Consequently it will have duplicate rows. An example query exhibiting the problem is
  SELECT STRAIGHT_JOIN c1 FROM t1 JOIN (SELECT @a := 0) x;
Since mysql_derived_fill will, for UNCACHEABLE_DEPENDENT tables, drop all rows and repopulate, we relax the condition within mysql_derived_fill: rather than assume all uncacheable values prevent an early return, we now allow an early return for uncacheable values other than UNCACHEABLE_DEPENDENT.  In general, we only populate derived tables once unless they're dependent tables.

### DIFF
--- a/sql/sql_derived.cc
+++ b/sql/sql_derived.cc
@@ -1201,8 +1201,10 @@ bool mysql_derived_fill(THD *thd, LEX *lex, TABLE_LIST *derived)
                        (derived->alias.str ? derived->alias.str : "<NULL>"),
                        derived->get_unit()));
 
-  if (unit->executed && !unit->uncacheable && !unit->describe &&
-      !derived_is_recursive)
+  /* Only fill derived tables once, unless the derived table is dependent in
+     which case we will delete all of its rows and refill it below. */
+  if (unit->executed && !(unit->uncacheable & UNCACHEABLE_DEPENDENT) &&
+      !unit->describe && !derived_is_recursive)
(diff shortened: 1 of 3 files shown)


sql/sql_derived.cc
```diff
@@ -1259,6 +1261,7 @@ bool mysql_derived_fill(THD *thd, LEX *lex, TABLE_LIST *derived)
   }
   else
   {
+    assert (!unit->executed || (unit->uncacheable & UNCACHEABLE_DEPENDENT));
```
cvicentiu, line 1264: use DBUG_ASSERT()

spetrunia (approved):
Please act on @cvicentiu's input
Then apply this change to fix coding style and make the testcase contain a clue about what is being testsed:

https://gist.github.com/spetrunia/9d0cffde72308ea904c2f372992b6e08

Ok to push after that is addressed.

DaveGosselin-MariaDB:
I created a new PR to merge to 10.5 and I'm closing this PR as a result.  The new PR is https://github.com/MariaDB/server/pull/3062 which I will merge once the buildbot checks complete as it addresses the comments made here.

PR #3002, opened 2024-01-12
MDEV-33095 innodb_flush_method=O_DIRECT creates excessive errors on Solaris

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33095*
## Description
The `directio(3C)` function on Solaris is supported on NFS and UFS while the majority of users should be on ZFS, which is a copy-on-write file system that implements transparent compression and therefore cannot support unbuffered I/O.

Let us remove the call to `directio()` and simply treat `innodb_flush_method=O_DIRECT` in the same way as the previous default value `innodb_flush_method=fsync` on Solaris. Also, let us remove some dead code around calls to `os_file_set_nocache()` on platforms where `<fcntl.h>` does not define `O_DIRECT`.
## How can this PR be tested?
I tested this on Debian GNU/Linux by editing `config.h` and `my_config.h` in the build directory so that they would not define the added preprocessor symbol `HAVE_FCNTL_DIRECT`.

## PR quality check


cmake/os/SunOS.cmake
```diff
@@ -17,6 +17,10 @@ INCLUDE(CheckSymbolExists)
 INCLUDE(CheckCSourceRuns)
 INCLUDE(CheckCSourceCompiles) 
 
+# fcntl(fd, F_SETFL, O_DIRECT) is not supported,
+# and directio(3C) would only work on UFS or NFS, not ZFS.
+SET(HAVE_FCNTL_DIRECT 0 CACHE INTERNAL "")
```
dr-m, line 22: I checked a few other POSIX-like operating systems. OpenBSD does not define `O_DIRECT` at all. FreeBSD and NetBSD allow `O_DIRECT` for both `open` and `fcntl`. For IBM AIX, `O_DIRECT` is documented for `open` but not for `fcntl()`.

@grooverdan Should we do something similar in `cmake/os/AIX.cmake`? And perhaps refactor the code so that we will pass the `O_DIRECT` flag already to `open`, like it was before cf1cf3a0d1fe277d18dae782b164f6c99f81be83?

dr-m: Addressed in 3dd8648073c390e843bfbabe428841fad987442b.
https://buildbot.mariadb.org/#/builders/121/builds/17538 is a successful IBM AIX run of this. In the build log there is no mention of looking for `O_DIRECT`; there is one about `O_NONBLOCK`. In the FreeBSD 14 log https://buildbot.mariadb.org/#/builders/619/builds/654 we can find attempts to look up both symbols:
```
-- Looking for O_NONBLOCK
-- Looking for O_NONBLOCK - found
-- Looking for O_DIRECT
-- Looking for O_DIRECT - found
```
This makes me believe that this approach should work on Solaris as well.

grooverdan (approved):
Looks good. Tested AIX. I note that the range of innodb_flush_method wasn't reduced for the not-HAVE_FCNTL_DIRECT arches  - to keep this is a stable version change. Currently just silently ignored.

I started to look at what an O_DIRECT at open time would look like:
```diff
diff --git a/storage/innobase/os/os0file.cc b/storage/innobase/os/os0file.cc
index 014a9eb632f..f2545c56cc6 100644
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -1126,6 +1126,11 @@ os_file_create_func(
        bool            read_only,
        bool*           success)
 {
+#ifdef O_DIRECT
+       static bool     o_direct_available = true;
+#else
+       constexpr auto  o_direct_available =  false;
+#endif
        bool            on_error_no_exit;
        bool            on_error_silent;
 
@@ -1202,6 +1207,13 @@ os_file_create_func(
                create_flag |= O_SYNC;
 #endif
        }
+#ifdef O_DIRECT
+       /* We disable OS caching (O_DIRECT) only on data files */
+       if (!read_only && o_direct_available
+           && type != OS_LOG_FILE
+           && type != OS_DATA_FILE_NO_O_DIRECT)
+               create_flag |= O_DIRECT;
+#endif
 
        os_file_t       file;
        bool            retry;
@@ -1212,6 +1224,15 @@ os_file_create_func(
                if (file == -1) {
                        const char*     operation;
 
+#ifdef O_DIRECT
+                       if (errno == EINVAL && (create_flag & O_DIRECT)) {
+                               o_direct_available = false;
+                               create_flag &= ~O_DIRECT;
+                               retry = true;
+                               continue;
+                       }
+#endif
+
                        operation = (create_mode == OS_FILE_CREATE
                                     && !read_only) ? "create" : "open";
 
@@ -1230,15 +1251,6 @@ os_file_create_func(
 
        } while (retry);
 
-#ifdef HAVE_FCNTL_DIRECT
-       /* We disable OS caching (O_DIRECT) only on data files */
-       if (!read_only
-           && *success
-           && type != OS_LOG_FILE
-           && type != OS_DATA_FILE_NO_O_DIRECT) {
-               os_file_set_nocache(file, name, mode_str);
-       }
-#endif
 
 #ifndef _WIN32
        if (!read_only
(END)
```

Crude start, depends at what level to cache the O_DIRECT unsupported. I think probably treat file/log separate (which with `type` exposed in the function should be easy enough`), but going to each individual tablespace could be too onerous. Given its one extra syscall on something that should be kept open for a while maybe it doesn't make much difference to cache it. With MDEV-30136 - lets consider it 11.0+ at least.

PR #3003, opened 2024-01-12
MDEV-33053 InnoDB LRU flushing does not run before running out of buffer pool

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33053*

## Description
Along with flushing appropriate number of pages page cleaner thread
has the responsibility of maintaining enough number of free pages
available for user threads to consume. In buf_flush_LRU, the LRU list
is scanned from tail flushing any dirty pages and evicting non-dirty
pages to fill the free list.
    
The page cleaner improvements have shifted the balance in such a way
that the LRU flushing and eviction can be skipped for long until the
buffer pool is out of free pages or there is a need to flush from the
LRU list based on LSN or dirty page percentage. In second case, the LRU
scan can be terminated early if dirty pages are found in LRU tail.
    
The patch has the following changes in page cleaner to address the issue.
    
1. Disable deep sleep without timeout for page cleaner.
2. Attempt LRU flush to free pages before the free list is exhausted.
3. While doing LRU flush followed by flush-list flush, adjust the flush limit
    based on max_io_capacity.

## How can this PR be tested?
No new automated test added.

This is a performance bug where a stall is observed once the free page list becomes empty. It is somewhat hard to have an mtr test to validate the patch.
 
* The status variables Innodb_buffer_pool_pages_free, Innodb_buffer_pool_pages_lru_flushed and Innodb_buffer_pool_pages_lru_freed needs to be observed in intervals while running queries and DML statements. The testcase could create demand for new page read at a moderate rate of about 500 pages per second.

Behaviour Without patch:  The free page count (Innodb_buffer_pool_pages_free) comes down and eventually reaches zero before jumping to  innodb_lru_scan_depth. Innodb_buffer_pool_pages_lru_freed trend also shows sudden jump at this point.

Behaviour with patch: The free pages are constantly added keeping Innodb_buffer_pool_pages_lru_freed consistently increasing over time and maintaining Innodb_buffer_pool_pages_free. 

* The patch needs to be tested with standard performance suite (like sysbench) for possible impact.

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -2319,18 +2319,24 @@ static void buf_flush_page_cleaner()
     }
 
     mysql_mutex_lock(&buf_pool.flush_list_mutex);
-    if (buf_pool.ran_out())
+    if (buf_pool.running_short())
       goto no_wait;
     else if (srv_shutdown_state > SRV_SHUTDOWN_INITIATED)
       break;
 
+    /* Disabling the deep sleep code for page cleaner. We should evaluate the
+    possibility of backing off and sleeping longer up to a maximum value if
+    needed. With deep sleep, the wake up for free pages happen at emergency
+    when it could be too late. It is possible to signal the page cleaner at
+    earlier stage, but it seems better to avoid too much signalling across and
+    synchronizing among user threads when there is no emergency.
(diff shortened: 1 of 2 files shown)


mariadb-DebarunBanerjee:
Need to create a separate branch with MDEV.

PR #3009, opened 2024-01-18
MDEV-33023 Crash in mariadb-backup --prepare --export after --prepare

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33023*

## Description
mariadb-backup with --prepare option could result in empty redo log
file. When --prepare is followed by --prepare --export, we exit early
in srv_start function without opening the ibdata1 tablespace. Later
while trying to read rollback segment header page, we hit the debug
assert which claims that the system space should already have been
opened.

There are two assert cases here.
    
Issue-1: System tablespace object is not there in fil space hash i.e.
srv_sys_space.open_or_create() is not called.
    
Issue-2: The system tablespace data file ibdata1 is not opened i.e.
fil_system.sys_space->open() is not called.

Fix: For empty redo log and restore operation, open system tablespace
before returning.

## How can this PR be tested?
./mtr mariabackup.partial

## PR quality check


storage/innobase/srv/srv0start.cc
```diff
@@ -1305,7 +1334,31 @@ dberr_t srv_start(bool create_new_db)
 		}
 
 		create_new_log = srv_log_file_size == 0;
-		if (!create_new_log) {
+		if (create_new_log) {
+			if (flushed_lsn < lsn_t(1000)) {
```
dr-m, line 1338: FWIW, the minimum LSN before is 8192 or 8192+12=8204. I think that we could simply check for zero here. In 685d958e38b825ad9829be311f26729cccf37c46 [MDEV-14425](https://jira.mariadb.org/browse/MDEV-14425) it was changed to `log_sys.FIRST_LSN=12288` and this code was removed. Did you also test a 10.11 version of this change?

storage/innobase/fil/fil0fil.cc
```diff
@@ -441,7 +441,8 @@ static bool fil_node_open_file(fil_node_t *node)
   ut_ad(fil_is_user_tablespace_id(node->space->id) ||
         srv_operation == SRV_OPERATION_BACKUP ||
         srv_operation == SRV_OPERATION_RESTORE ||
-        srv_operation == SRV_OPERATION_RESTORE_DELTA);
+        srv_operation == SRV_OPERATION_RESTORE_DELTA ||
+        srv_operation == SRV_OPERATION_RESTORE_EXPORT);
```
dr-m, line 445: `SRV_OPERATION_RESTORE_EXPORT` should be as close to `SRV_OPERATION_NORMAL` as possible, with the main exception that `mariadb-backup --prepare --export` will just export a few tablespaces and exit. I am not sure that relaxing this assertion is the right thing to do.

dr-m: The original assertion does not fail for me if I apply the following patch:
```diff
diff --git a/extra/mariabackup/xtrabackup.cc b/extra/mariabackup/xtrabackup.cc
index 5f6ec1b0c0c..e547d9fafa4 100644
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -6214,7 +6214,7 @@ static bool xtrabackup_prepare_func(char** argv)
 		    recv_sys.recovered_lsn, target_lsn);
 		ok = false;
 	}
-#ifdef WITH_WSREP
+#if 0
 	else if (ok) xb_write_galera_info(xtrabackup_incremental);
 #endif
 
```
I think that we should only invoke this once, at the first `--prepare` and not thereafter.

dr-m: Using the following patch:
```diff
diff --git a/mysql-test/suite/mariabackup/partial.test b/mysql-test/suite/mariabackup/partial.test
index 2a744a7f729..aa02f123d41 100644
--- a/mysql-test/suite/mariabackup/partial.test
+++ b/mysql-test/suite/mariabackup/partial.test
@@ -57,8 +57,8 @@ SELECT * FROM t1;
 
 --echo # MDEV-33023 Crash in mariadb-backup --prepare --export after --prepare
 --disable_result_log
-exec $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --target-dir=$targetdir;
-exec $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --export --target-dir=$targetdir;
+exec rr record $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --target-dir=$targetdir;
+exec rr record $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --export --target-dir=$targetdir;
 --enable_result_log
 
 list_files $targetdir/test *.cfg;
```
and the invocation
```sh
_RR_TRACE_DIR=/dev/shm/rr ./mtr mariabackup.partial
rr replay /dev/shm/rr/mariabackup-0
```
I found out that the first `--prepare` run would open the system tablespace file here:
```
Thread 1 hit Hardware watchpoint 1: fil_system.sys_space.chain.start.handle

Old value = {m_file = -1}
New value = {m_file = 10}
bt
fil_node_open_file_low (node=node@entry=0x5643989fc570)
    at /mariadb/10.5/storage/innobase/fil/fil0fil.cc:396
396	    if (success && node->is_open())
(rr) bt
#0  fil_node_open_file_low (node=node@entry=0x5643989fc570)
    at /mariadb/10.5/storage/innobase/fil/fil0fil.cc:396
#1  0x0000564395b98c6c in fil_space_t::open (this=0x5643989fc280, 
    create_new_db=create_new_db@entry=false)
    at /mariadb/10.5/storage/innobase/fil/fil0fil.cc:1189
#2  0x0000564395d67f3e in srv_start (create_new_db=false)
    at /mariadb/10.5/storage/innobase/srv/srv0start.cc:1388
#3  0x00005643955c2bc3 in innodb_init ()
    at /mariadb/10.5/extra/mariabackup/xtrabackup.cc:2291
#4  0x00005643955ca6b7 in xtrabackup_prepare_func (
    argv=argv@entry=0x5643984ecb38)
    at /mariadb/10.5/extra/mariabackup/xtrabackup.cc:6159
#5  0x00005643955ce33e in main_low (argv=0x5643984ecb38)
    at /mariadb/10.5/extra/mariabackup/xtrabackup.cc:7098
#6  0x00005643955ce58e in main (argc=<optimized out>, argv=<optimized out>)
    at /mariadb/10.5/extra/mariabackup/xtrabackup.cc:6882
```
The second run `--prepare --export` never changed the file handle from -1 to anything, that is, it had not opened the system tablespace at the point when `xb_write_galera_info()` had been invoked. I think that the problem is the placement of the `xb_write_galera_info()` call, or the fact that we are not opening the system tablespace at all during `--prepare --export`.

mariadb-DebarunBanerjee: yes, I actually have written already in the "Description" where we skip the open call. Note that the 2 --prepare calls are independent and different execution.

dr-m: I found the place that prevents the system and undo tablespaces from being opened: The first `mariadb-backup --prepare` would delete the `ib_logfile0` or replace it with a dummy zero-length one. The following would allow the test case to pass:
```diff
diff --git a/storage/innobase/srv/srv0start.cc b/storage/innobase/srv/srv0start.cc
index cd9f07845b9..08fb024923e 100644
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -1328,7 +1328,12 @@ dberr_t srv_start(bool create_new_db)
 		bool log_file_found;
 		if (dberr_t err = find_and_check_log_file(log_file_found)) {
 			if (err == DB_NOT_FOUND) {
-				return DB_SUCCESS;
+				err = fil_system.sys_space->open(false)
+					? srv_undo_tablespaces_init(false)
+					: DB_ERROR;
+				if (err == DB_SUCCESS) {
+					return err;
+				}
 			}
 			return srv_init_abort(err);
 		}

```
This code should be different on 10.11 and possibly 11.0, so please test your changes there as well.

dr-m: I was expecting my above patch to `srv_start()` to fix the following test:
```diff
diff --git a/mysql-test/suite/innodb/t/alter_kill.test b/mysql-test/suite/innodb/t/alter_kill.test
index a8a151ab1e2..693dd0b496e 100644
--- a/mysql-test/suite/innodb/t/alter_kill.test
+++ b/mysql-test/suite/innodb/t/alter_kill.test
@@ -117,6 +117,7 @@ syswrite(FILE, $page, $ps)==$ps || die "Unable to write $file\n";
 close(FILE) || die "Unable to close $file";
 EOF
 
+-- remove_file $MYSQLD_DATADIR/ib_logfile0
 -- source include/start_mysqld.inc
 
 INSERT INTO bug16720368 VALUES(9,1);
```
However, InnoDB would successfully start up in this modified test even when that patch is not applied.

storage/innobase/srv/srv0start.cc
mariadb-DebarunBanerjee: 1. I do not intend to include any other functional changes part of this patch. The numbers and code logic remains exactly same. If we intend to backport MDEV-14425 to 10.5, it is better to open and MDEV.

2. This part of the patch (Issue-1:  in description) won't be required from 10.6 onwards. I have checked in 10.6 but not in 10.11.

storage/innobase/fil/fil0fil.cc
mariadb-DebarunBanerjee: I have now modified the patch to open system and undo tablespace earlier. Please have a look.

dr-m:
Thank you, this is much simpler. For the record, I tested a cherry-pick of c55714d953a2481b03d7f0bce3c104b7600cdf4b to 10.11 8c5db7a1872c5644a14bbaddadee277d0b354c23, and I disregarded the code changes (there would be conflicts for `storage/innobase/srv/srv0start.cc` anyway):
```
mariabackup.partial                      [ pass ]   2181
```
I then applied the following patch:
```diff
diff --git a/mysql-test/suite/mariabackup/partial.test b/mysql-test/suite/mariabackup/partial.test
index 85808749b62..08b82e3f22c 100644
--- a/mysql-test/suite/mariabackup/partial.test
+++ b/mysql-test/suite/mariabackup/partial.test
@@ -59,6 +59,7 @@ SELECT * FROM t1;
 --echo # MDEV-33023 Crash in mariadb-backup --prepare --export after --prepare
 --disable_result_log
 exec $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --target-dir=$targetdir;
+remove_file $targetdir/ib_logfile0;
 exec $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --export --target-dir=$targetdir;
 --enable_result_log
 
```
This would (incorrectly) simulate the situation that the backup had been prepared before MariaDB 10.8. What is missing is that we should write the LSN to `FIL_PAGE_FILE_FLUSH_LSN` in the first page of the system tablespace. The test would fail like this:
```
2024-01-24  9:26:44 0 [Note] InnoDB: Completed initialization of buffer pool
2024-01-24  9:26:44 0 [ERROR] InnoDB: File ./ib_logfile0 was not found
[00] 2024-01-24 09:26:44 mariadb-backup: srv_start() returned 11 (Generic error).
```
I also tried to create a zero-length `ib_logfile0`, which is what `mariadb-backup --copy-back` would do in older versions:
```diff
diff --git a/mysql-test/suite/mariabackup/partial.test b/mysql-test/suite/mariabackup/partial.test
index 85808749b62..9f36758eaf6 100644
--- a/mysql-test/suite/mariabackup/partial.test
+++ b/mysql-test/suite/mariabackup/partial.test
@@ -59,6 +59,9 @@ SELECT * FROM t1;
 --echo # MDEV-33023 Crash in mariadb-backup --prepare --export after --prepare
 --disable_result_log
 exec $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --target-dir=$targetdir;
+remove_file $targetdir/ib_logfile0;
+write_file $targetdir/ib_logfile0;
+EOF
 exec $XTRABACKUP --defaults-file=$server_cnf --defaults-group-suffix=.1 --prepare --export --target-dir=$targetdir;
 --enable_result_log
 
```
This would fail in a different way:
```
2024-01-24  9:33:19 0 [Note] InnoDB: Completed initialization of buffer pool
2024-01-24  9:33:19 0 [ERROR] InnoDB: File ./ib_logfile0 is too small
[00] 2024-01-24 09:33:19 mariadb-backup: srv_start() returned 11 (Generic error).
```
I think that we can live with these limitations.

dr-m (approved):

I tested this in 10.6 as well. There will be a conflict when merging to 10.11. We can simply omit the code change there.

PR #3010, opened 2024-01-18
MDEV-33133: MDL conflict handling code should skip BF-aborted trxs

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-33133*

## Description
It's possible that MDL conflict handling code is called more than once for a transaction when:
- it holds more than one conflicting MDL lock
- reschedule_waiters() is executed, which results in repeated attempts to BF-abort already aborted transaction.
In such situations, it might be that BF-aborting logic sees a partially rolled back transaction and erroneously decides on future actions for such a transaction.

The specific situation tested and fixed is when a SR transaction applied in the node gets BF-aborted by a started TOI operation. It's then caught with the server transaction already rolled back, but with no MDL locks yet released. This caused wrong state detection for such a transaction during repeated MDL conflict handling code execution.

## How can this PR be tested?

MTR test is provided.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/MDEV-33133.result
+++ b/mysql-test/suite/galera/r/MDEV-33133.result
@@ -1,0 +1,34 @@
+connection node_2;
+connection node_1;
+connect node_1a,127.0.0.1,root,,test,$NODE_MYPORT_1;
+connection node_1;
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY) ENGINE=InnoDB;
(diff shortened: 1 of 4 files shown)


sysprg:
Thanks, fix has been merged with head revision (as part of https://github.com/MariaDB/server/commit/d5a669b6b645926360c2a36a99d14501fe8c91fa), the problem with mutex has been moved to a separate task

PR #3012, opened 2024-01-18
MDEV-30660 COUNT DISTINCT seems unnecessarily slow when run on a PK

Olernov:
If we have a statement like SELECT AGGFN(DISTINCT c1, c2,..,cn) FROM t1, where AGGFN is an aggregate function like COUNT(), AVG(), SUM(), and there is a unique index on table t1 including all columns (c1, c2,..,cn), the values retrieved from the table are guaranteed to be unique. So we can optimize aggregation and avoid de-duplicating of values, in other words - ignore the DISTINCT clause.
The restrictions are:
  - only one table involved in the join
  - all arguments of the aggregate function are fields (not functions/subqueries)

- [x] *The Jira issue number for this PR is: MDEV-_30660_*

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/main/distinct.test
+++ b/mysql-test/main/distinct.test
@@ -928,4 +928,4 @@ drop table t1;
 
 --echo #
 --echo # end of 10.5 tests
---echo #
+
--- a/mysql-test/main/distinct_notembedded.result
+++ b/mysql-test/main/distinct_notembedded.result
@@ -289,6 +289,31 @@ JS
         "aggregator_type": "simple"
     }
 ]
+# GROUP_CONCAT implements non-standard distinct aggregator
+SELECT GROUP_CONCAT(b) FROM t1;
+GROUP_CONCAT(b)
+1,1,1
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "group_concat(t1.b separator ',')",
+        "aggregator_type": "simple"
+    }
+]
+SELECT GROUP_CONCAT(DISTINCT b) FROM t1;
+GROUP_CONCAT(DISTINCT b)
+1
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "group_concat(distinct t1.b separator ',')",
+        "aggregator_type": "distinct"
+    }
+]
 DROP TABLE t1, t2, t3;
 DROP VIEW v1;
 SET optimizer_trace = @save_optimizer_trace;
--- a/mysql-test/main/distinct_notembedded.result
+++ b/mysql-test/main/distinct_notembedded.result
@@ -1,0 +1,280 @@
+#
+# MDEV-30660 COUNT DISTINCT seems unnecessarily slow when run on a PK
+#
+set @save_optimizer_trace = @@optimizer_trace;
+SET optimizer_trace='enabled=on';
+CREATE TABLE t1 (a INT NOT NULL PRIMARY KEY, b INT NOT NULL);
+INSERT INTO t1 VALUES (1,1), (2,1), (3,1);
+# Optimization is applied (aggregator=simple):
+SELECT COUNT(DISTINCT a) FROM t1;
+COUNT(DISTINCT a)
+3
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "count(distinct t1.a)",
+        "type of aggregator": "simple"
+    }
+]
+SELECT AVG(DISTINCT a), SUM(DISTINCT b) FROM t1;
+AVG(DISTINCT a)	SUM(DISTINCT b)
+2.0000	1
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "avg(distinct t1.a)",
+        "type of aggregator": "simple"
+    },
+    {
+        "function": "sum(distinct t1.b)",
+        "type of aggregator": "distinct"
+    }
+]
+# Only `a` is unique but it's enough to eliminate DISTINCT:
+SELECT COUNT(DISTINCT b, a) FROM t1;
+COUNT(DISTINCT b, a)
+3
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "count(distinct t1.b,t1.a)",
+        "type of aggregator": "simple"
+    }
+]
+SELECT COUNT(DISTINCT a, a + b) FROM t1;
+COUNT(DISTINCT a, a + b)
+3
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "count(distinct t1.a,t1.a + t1.b)",
+        "type of aggregator": "simple"
+    }
+]
+SELECT SUM(DISTINCT a), AVG(DISTINCT a), COUNT(DISTINCT a) FROM t1 WHERE a > 1;
+SUM(DISTINCT a)	AVG(DISTINCT a)	COUNT(DISTINCT a)
+5	2.5000	2
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "sum(distinct t1.a)",
+        "type of aggregator": "simple"
+    },
+    {
+        "function": "avg(distinct t1.a)",
+        "type of aggregator": "simple"
+    },
+    {
+        "function": "count(distinct t1.a)",
+        "type of aggregator": "simple"
+    }
+]
+# Optimization is not applied 'cause function argument is not a field
+# (aggregator=distinct):
+SELECT SUM(DISTINCT a + b) FROM t1;
+SUM(DISTINCT a + b)
+9
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "sum(distinct t1.a + t1.b)",
+        "type of aggregator": "distinct"
+    }
+]
+SELECT COUNT(DISTINCT b) FROM t1;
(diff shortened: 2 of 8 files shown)


mysql-test/main/distinct.test
```diff
@@ -928,4 +928,47 @@ drop table t1;
...
+
+CREATE TABLE t1 (a INT NOT NULL PRIMARY KEY, b INT NOT NULL);
+INSERT INTO t1 VALUES (1,1), (2,1), (3,1);
+
+SELECT COUNT(DISTINCT a) FROM t1;
+SELECT COUNT(DISTINCT (SELECT a)) FROM t1;
+SELECT AVG(DISTINCT a) FROM t1;
+SELECT SUM(DISTINCT a) FROM t1;
+SELECT SUM(DISTINCT a) FROM t1 WHERE a > 1;
+SELECT SUM(DISTINCT a + b) FROM t1;
```
grooverdan, line 944: this case doesn't use the optimization as `func->arguments()[i]->type()` is `Item::FUNC_ITEM`. Same with last `AVG(DISTINCT b + a)`. For a PK(a) this is correct, but should these be listed separately to show the not-selected case.

mysql-test/main/distinct.test
```diff
@@ -928,4 +928,47 @@ drop table t1;
...
+SELECT SUM(DISTINCT b) FROM t1, (SELECT 1) AS t2;
+SELECT COUNT(DISTINCT a) FROM t1, (SELECT 1 UNION SELECT 2) AS t2;
+
+--echo # Unique index on two columns
+CREATE TABLE t3 (a INT NOT NULL, b INT NOT NULL);
+INSERT INTO t3 VALUES (1,1), (2,2), (3,3);
+CREATE UNIQUE INDEX t3_a_b ON t3 (a, b);
+SELECT COUNT(DISTINCT a, b) FROM t3;
+SELECT COUNT(DISTINCT b, a) FROM t3;
+SELECT COUNT(DISTINCT b, a) FROM t3 WHERE a < 3;
```
grooverdan, line 969: Would `SELECT COUNT(DISTINCT b + a) FROM t3;`  be valid for deterministic functions like `+` where arguments are FIELD_ITEM, in the index, or CONST?

sql/sql_select.cc
```diff
@@ -27147,15 +27147,48 @@ static bool setup_sum_funcs(THD *thd, Item_sum **func_ptr)
...
+          all_args_are_fields= false;
+          break;
+        }
+      }
+
+      if (all_args_are_fields &&
+          list_contains_unique_index(join_tab[const_tables].table,
+                                     find_field_in_item_list,
+                                     (void *) &fields_list))
+        need_distinct_aggregator= false;
```
grooverdan, line 27188: Enhancement to handle case:

`SELECT a,COUNT(DISTINCT b) FROM t3 GROUP BY a;`
```
-      if (all_args_are_fields &&
-          list_contains_unique_index(join_tab[const_tables].table,
+      if (all_args_are_fields)
+      {
+         for (ORDER *tmp_group= group_list; tmp_group ; tmp_group= tmp_group->next)
+            if ((*tmp_group->item)->type() == Item::FIELD_ITEM)
+              fields_list.push_back(*tmp_group->item);
+         if (list_contains_unique_index(join_tab[const_tables].table,
                                      find_field_in_item_list,
                                      (void *) &fields_list))
-        need_distinct_aggregator= false;
+            need_distinct_aggregator= false;
+      }
```

sql/sql_select.cc
```diff
@@ -27147,15 +27147,48 @@ static bool setup_sum_funcs(THD *thd, Item_sum **func_ptr)
...
+    }
     if (func->set_aggregator(thd,
-                             need_distinct && func->has_with_distinct() ?
+                             need_distinct_aggregator ?
```
grooverdan, line 27191: Pity there's no ANALYZE FORMAT=JSON or EXPLAIN output to include which aggregator is used. If its stable to add, it would be nice.

sql/sql_select.cc
grooverdan: is there an `else` condition here that can `break` out of the `func`  list if any of the functions don't support a SIMPLE_AGGREGATOR?

mysql-test/main/distinct.test
Olernov: I don't think so, for example we have `UNIQUE KEY(a, b)` and values (1,2) and (2,1). `SELECT COUNT(DISTINCT b + a)` must return 1 but will return 2 if we apply SIMPLE_AGGREGATOR. This could work for something simpler using only one field: `SELECT COUNT(DISTINCT a + <const>)`. At the same time, it will not work for `SELECT COUNT(DISTINCT a + b + <const>)`. I also don't know what functions are safe to use, even the arithmetic division is questionable... 
I'm not sure we should dive deep in such particular cases unless we can generalize them somehow.

grooverdan: ah right. Thanks

sql/sql_select.cc
Olernov: The aggregator is set separately for each function, no need to `break` the loop

sql/sql_select.cc
```diff
@@ -290,8 +290,8 @@ static void update_tmptable_sum_func(Item_sum **func,TABLE *tmp_table);
 static void copy_sum_funcs(Item_sum **func_ptr, Item_sum **end);
 static bool add_ref_to_table_cond(THD *thd, JOIN_TAB *join_tab);
 static bool setup_sum_funcs(THD *thd, Item_sum **func_ptr);
-static bool prepare_sum_aggregators(THD *thd, Item_sum **func_ptr,
-                                    bool need_distinct);
+// static bool prepare_sum_aggregators(THD *thd, Item_sum **func_ptr,
```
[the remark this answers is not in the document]
Olernov: Done

mysql-test/main/distinct.test
Olernov: OK, moved them to a separate part of the test file and marked them with `--echo #`

sql/sql_select.cc
Olernov: Good point with `GROUP BY`! Added it to the next revision

sql/sql_select.cc
Olernov: Good idea, added logging to the optimizer trace. Now we can test not only the correct results but the fact the optimization is applied

grooverdan (approved):
Apart from re-recording a `main.opt_trace` [test case](https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F3012%2Fmerge) looks good to me.

mysql-test/main/distinct_notembedded.test
```diff
@@ -0,0 +1,109 @@
...
+  FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+
+--echo # Can't apply optimization 'cause GROUP BY argument is not a field:
+SELECT COUNT(DISTINCT b) FROM t3 GROUP BY a+b;
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+  FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+
+DROP TABLE t1, t2, t3;
+DROP VIEW trace;
+SET optimizer_trace='enabled=off';
```
spetrunia, line 107: Please set the optimizer_trace back to its original value. 
We have opt_trace run in the CI which runs the entire testsuite with trace enabled. We don't want any statement to set it to disabled.

Olernov: Done

spetrunia:
The commit comment says:
>    If we have a statement
>      SELECT AGGR_FN(DISTINCT c1, c2,..,cn) FROM t1,
>    where AGGR_FN is an aggregate function like COUNT(), AVG(), SUM(),
>    and there is a unique index on table t1 including all columns (c1, c2,..,cn),

>> including all columns

It doesn't actually need to include all columns.  Example: this patch handles this case:
```
create table t1 (pk int primary key, a int);
explain select count(distinct pk, a) from t1;
```

Olernov:
> The commit comment says:
> 
> > If we have a statement
> > SELECT AGGR_FN(DISTINCT c1, c2,..,cn) FROM t1,
> > where AGGR_FN is an aggregate function like COUNT(), AVG(), SUM(),
> > and there is a unique index on table t1 including all columns (c1, c2,..,cn),
> 
> > > including all columns
> 
> It doesn't actually need to include all columns. Example: this patch handles this case:
> 
> ```
> create table t1 (pk int primary key, a int);
> explain select count(distinct pk, a) from t1;
> ```
> 
> (will provide more input shortly, it makes sense to wait until all of it is here)

I agree, fixed this

spetrunia:
@Olernov 
* Please apply this patch providing coding style fixes and improving comments: https://gist.github.com/spetrunia/9524c1ed226dbbd1bcbf0a4e82468745
* Please fix it to work with VIEWs as described: https://jira.mariadb.org/browse/MDEV-30660?focusedCommentId=279719&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-279719
* Please fix the trace printout to not be misleading for GROUP_CONCAT: https://jira.mariadb.org/browse/MDEV-30660?focusedCommentId=279721&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-279721
 I suggest print aggregator_type as  "internal" for GROUP_CONCAT based aggregates.
* Please rename "type of aggregator" to "aggregator_type".

That's all the input I have.

mysql-test/main/distinct_notembedded.result
```diff
@@ -0,0 +1,321 @@
...
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
+1	PRIMARY	t1	index	NULL	PRIMARY	4	NULL	3	Using index
+2	DEPENDENT SUBQUERY	NULL	NULL	NULL	NULL	NULL	NULL	NULL	No tables used
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "count(distinct (/* select#2 */ select t1.a))",
+        "aggregator_type": "distinct"
```
grooverdan, line 130: this would be a simple in the way it got re-written. But maybe not easy to fix. Other tests include comment on why not.

mysql-test/main/distinct_notembedded.result
```diff
@@ -0,0 +1,321 @@
...
+]
+SELECT AVG(DISTINCT t1.a) FROM (SELECT 1 AS a) AS t2, t1, (SELECT 2 AS a) AS t3;
+AVG(DISTINCT t1.a)
+2.0000
+SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.prepare_sum_aggregators')) AS JS
+FROM INFORMATION_SCHEMA.OPTIMIZER_TRACE;
+JS
+[
+    {
+        "function": "avg(distinct t1.a)",
```
grooverdan, line 181: how did this avoid the "more than one table"? but not the test below?

grooverdan (approved):

mysql-test/main/distinct_notembedded.result
Olernov: `t2` and `t3` are `const` tables here, so they do not count. But `t2` in the test below is not `const`, so it prevents the optimization. `EXPLAIN EXTENDED` displays that in both cases (there is no `t2` and `t3` at all in the first statement after the rewrite)

spetrunia:
@Olernov did you see a compile error: https://buildbot.mariadb.org/#/builders/572/builds/6388/steps/7/logs/stdio

spetrunia (approved):
Other than the compilation issue, Ok to push.

Olernov:
@spetrunia , I missed the compilation error, thanks for the heads up. Have just added the fix.
Regarding "Ok to push" - we target 10.6, right? No need to backport to 10.5?

Olernov:
Closing this PR as the corresponding https://github.com/MariaDB/server/pull/3056 has been merged to 10.5.

PR #3016, opened 2024-01-22
MDEV-33283: Binlog Checksum is Zeroed by Zlib if Part of Event Data is Empty

bnestere:
An existing binlog checksum can be overridden to 0 if writing a NULL payload when using Zlib for the computation. That is, calling into Zlib's crc32 with empty data initializes an incremental CRC computation to 0.

This patch changes the Log_event_writer::write_data() to exit immediately if there is nothing to write, thereby bypassing the checksum computation. This follows the pattern of
Log_event_writer::encrypt_and_write(), which also exits immediately if there is no data to write.

See that this fixes the failing `rpl.rpl_sp_effects` test on [s390x](https://buildbot.mariadb.org/#/builders/476/builds/16004)


andrelkin (approved):
Looks good.

PR #3019, opened 2024-01-23
MDEV-33278 : Assertion failure in thd_get_thread_id at lock_wait_wsrep

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-33278*

## Description

Problem is that not all conflicting transactions have THD object e.g. background persistent statistic calculation thread inside InnoDB. Therefore, it must be checked that victim has THD before it's identification is added to victim list as victim's thread identification is later requested using thd_get_thread_id function that requires that we have valid pointer to THD object in trx->mysql_thd.

This fix does not contain mtr-testcase because it would require either a very big transaction from other node so that persistent statistic calculation would be executed to same table or some debug instrumentation to force triggering it.

## How can this PR be tested?

- [ X] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [ X] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [ X] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

### DIFF
--- a/storage/innobase/lock/lock0lock.cc
+++ b/storage/innobase/lock/lock0lock.cc
@@ -976,8 +976,31 @@ func_exit:
     for (lock_t *lock= UT_LIST_GET_FIRST(table->locks); lock;
          lock= UT_LIST_GET_NEXT(un_member.tab_lock.locks, lock))
     {
-      /* if victim has also BF status, but has earlier seqno, we have to wait */
-      if (lock->trx != trx &&
+      /* Victim trx needs to be different from BF trx and it has to have a
+         THD so that we can kill it. Victim might not have THD in two cases:
+
+         (1) An incomplete transaction that was recovered from undo logs
+         on server startup (and not yet rolled back).
+
+         (2) Transaction that is in XA PREPARE state and whose client
+         connection was disconnected.
+
+         Neither of these can complete before lock_wait_wsrep() releases
+         lock_sys.latch.
+
+         (1) trx_t::commit_in_memory() is clearing both
+         trx_t::state and trx_t::is_recovered before it invokes
+         lock_release(trx_t*) (which would be blocked by the exclusive
+         lock_sys.latch that we are holding here). Hence, it is not
+         possible to write a debug assertion to document this scenario.
+
+         (2) If is in XA PREPARE state, it would eventually be rolled
+         back and the lock conflict would be resolved when an XA COMMIT
+         or XA ROLLBACK statement is executed in some other connection.
+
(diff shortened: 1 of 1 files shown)


storage/innobase/lock/lock0lock.cc
```diff
@@ -1009,8 +1012,11 @@ static void lock_wait_wsrep(trx_t *trx)
         lock= lock_rec_get_next(heap_no, lock);
       do
       {
-        /* if victim has also BF status, but has earlier seqno, we have to wait */
-        if (lock->trx != trx &&
+        /* Victim trx needs to be different from BF trx and it has to have a
+           THD so that we can kill it. If victim has also BF status,
+           but has earlier seqno, we have to wait.
+        */
+        if (lock->trx != trx && lock->trx->mysql_thd &&
```
dr-m, line 1019: The commit message fails to explain what the exact scenario is. For the case `!lock->trx->mysql_thd`, I think that we need a debug assertion to document the assumptions, for example, that this is a recovered transaction, or a transaction in `XA PREPARE` state.

dr-m: Based on some additional detail in [MDEV-33278](https://jira.mariadb.org/browse/MDEV-33278), the scenario is that `lock->trx` is a recovered transaction that has not yet been rolled back. Unless it is in `XA PREPARE` state, it would eventually be rolled back and the lock conflict would be resolved. If it is in `XA PREPARE` state, then the conflict would remain blocking this transaction `trx` until an `XA COMMIT` or `XA ROLLBACK` statement is executed in some other connection.

I see that `trx_t::commit_in_memory()` is clearing both `trx_t::state` and `trx_t::is_recovered` before it invokes `lock_release(trx_t*)` (which would be blocked by the exclusive `lock_sys.latch` that we are holding here). Hence, it is not possible to write a debug assertion to document this scenario.

Please clarify the comment and the commit message about this.

dr-m (approved):
Looks good. Please improve the commit message slightly:
>Both of these can't complete before lock_sys.latch acquired is released.

Neither of these can complete before lock_wait_wsrep() releases lock_sys.latch.

>the lock conflict would be resolvedwhen

missing space

sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/b762541dd63df8935134b477ef9473c709941737

PR #3022, opened 2024-01-24
MDEV-33277 In-place migration from MySQL 5.7 causes invalid AUTO_INCREMENT values

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33277*

## Description
MDEV-33277 In-place migration from MySQL 5.7 causes invalid `AUTO_INCREMENT` values

As noted in commit d0ef1aaf618452efea5a2e74f6d86ed0a586b63e, MySQL as well as older versions of MariaDB server would during `ALTER TABLE…IMPORT TABLESPACE` write bogus values to the `PAGE_MAX_TRX_ID` field in the clustered index. In 8777458a6eb73ac1d7d864ebac390ea7039e21c1 this field was repurposed for `PAGE_ROOT_AUTO_INC` in the clustered index root page.

`btr_read_autoinc_with_fallback()`: Add the parameters to `mysql_version`, `max` to indicate the `TABLE_SHARE::mysql_version` of the `.frm` file and the maximum value allowed for the type of the `AUTO_INCREMENT` column. In case the table was originally created in MySQL or an older version of MariaDB, read also the maximum value of the `AUTO_INCREMENT` column from the table and reset the `PAGE_ROOT_AUTO_INC` if it is above the limit.

TODO: Implement `CHECK TABLE…FOR UPGRADE` and the related logic. 2b2b734bf3167e9797f14e2b3da1857f7d33f7e9 is work in progress.
## How can this PR be tested?
This is covered by the added test `innodb.autoinc_import`.

This is applicable to MariaDB Server 10.2 and later, but 10.4 is near its end of life and we should probably avoid this kind of big changes, due to potential regression.
## PR quality check


mysql-test/suite/innodb/t/autoinc_import.test
```diff
@@ -0,0 +1,133 @@
...
+ALTER TABLE t10_1 IMPORT TABLESPACE;
+ALTER TABLE t10_1b IMPORT TABLESPACE;
+
+# Restart, so that the InnoDB tables will be loaded into the data dictionary.
+--let $restart_parameters=--innodb-read-only --read-only
+--source include/restart_mysqld.inc
+
+# FIXME: If we executed CHECK TABLE here, it would upgrade the mysql_version
+# in the .frm files despite --read-only!
+# CHECK TABLE t1, t1b, t1t, t5_7, t5_7b, t10_1, t10_1b;
```
dr-m, line 95: It would be better to mention [MDEV-33308](https://jira.mariadb.org/browse/MDEV-33308) (which I just filed) in the comment.

montywi: Better to apply my check table patch to be part of this one so that you can uncomment the above check tables.

storage/innobase/btr/btr0btr.cc
```diff
@@ -1269,51 +1270,62 @@ btr_read_autoinc(dict_index_t* index)
...
+      value in the PAGE_ROOT_AUTO_INC field. */
+      const dict_col_t *autoinc_col= dict_table_get_nth_col(table, col_no);
+      while (index && index->fields[0].col != autoinc_col)
+        index= dict_table_get_next_index(index);
 
-	ib_uint64_t	autoinc	= block ? page_get_autoinc(block->frame) : 0;
-	const bool	retry	= block && autoinc == 0
-		&& !page_is_empty(block->frame);
-	mtr.commit();
+      if (index)
```
montywi, line 1307: Better to do a commit and return if 'index' is not set.
It would make the code more clear instead of relying on that: need_adjust will be false on the next line.
It is likely that you a compiler would in the future give a warning that 'index' may be null at line 1321.

montywi (commented):
- Please add the code that fixes check table
- We have also to update that the mariadb-upgrade process changes the frm version for tables that are of type innodb, has a mysql-version < 10020 and an auto-increment key. We can do that in another commit

storage/innobase/btr/btr0btr.cc
dr-m: It looks like `dict_load_table_one()` would crash if after the call to `dict_load_indexes()` we’d have an empty list `table->indexes`. The condition `if (index)` can therefore be removed.

storage/innobase/btr/btr0btr.cc
```diff
@@ -1269,51 +1270,62 @@ btr_read_autoinc(dict_index_t* index)
...
+  {
+    autoinc= page_get_autoinc(block->frame);
 
-	mtr_t		mtr;
-	mtr.start();
-	buf_block_t*	block = buf_page_get(
-		page_id_t(index->table->space_id, index->page),
-		index->table->space->zip_size(),
-		RW_S_LATCH, &mtr);
+    if (autoinc > max || mysql_version == 0 || mysql_version < 100210)
```
mariadb-DebarunBanerjee, line 1297: Nice to have the condition (autoinc > max) as well. Perhaps the case when we have already upgraded the version (before this patch) but the value is still junk. 

I also observed that previously we had the check that if the page is not empty and autoinc is still zero. Although I understand that it was introduced for older version,  it might be good to keep that one as well for safety ? Since we don't expect to hit these conditions, we could also want to add some logs.

    auto valid_autoinc = [&]() {
        /* check for max range */
        /* check for non-empty page zero autoinc */
        /* Log warning in case check fails. */
    }

    if (old_version || !valid_autoinc()) {
        ....
    }

storage/innobase/btr/btr0btr.cc
mariadb-DebarunBanerjee: Looks like the index could be NULL in 2 cases.

* Empty index list (not even any cluster index ?)
* Index for the auto increment column is not found.

We don't expect them to happen and possible ways we can handle are ...

1. Check and return Error
2. Assert
3. Remove the checks (should eventually crash if ever happens)
4. Have the checks and skip the evaluation returning zero value (code today)

My preference is same as the list numbering (top to bottom).

storage/innobase/btr/btr0btr.cc
```diff
@@ -1269,51 +1270,62 @@ btr_read_autoinc(dict_index_t* index)
...
-		/* This should be an old data file where
-		PAGE_ROOT_AUTO_INC was initialized to 0.
-		Fall back to reading MAX(autoinc_col).
-		There should be an index on it. */
-		const dict_col_t*	autoinc_col
-			= dict_table_get_nth_col(table, col_no);
-		while (index && index->fields[0].col != autoinc_col) {
-			index = dict_table_get_next_index(index);
-		}
+      const bool need_adjust{autoinc > max || autoinc < max_autoinc};
```
mariadb-DebarunBanerjee, line 1310: May be we can assert here that (max <= max_autoinc).

storage/innobase/btr/btr0btr.cc
dr-m: Sorry, I was mistaken, and we indeed cannot do any further adjustment if `index==nullptr`. This is traversing the list of indexes, trying to find an index that has the `AUTO_INCREMENT` column as the first field. This may indeed fail if the metadata is corrupted in some non-obvious way.

mysql-test/suite/innodb/t/autoinc_import.test
```diff
@@ -0,0 +1,133 @@
...
+
+ALTER TABLE t1 IMPORT TABLESPACE;
+ALTER TABLE t1b IMPORT TABLESPACE;
+ALTER TABLE t1t IMPORT TABLESPACE;
+ALTER TABLE t5_7 IMPORT TABLESPACE;
+ALTER TABLE t5_7b IMPORT TABLESPACE;
+ALTER TABLE t10_1 IMPORT TABLESPACE;
+ALTER TABLE t10_1b IMPORT TABLESPACE;
+
+# Restart, so that the InnoDB tables will be loaded into the data dictionary.
```
mariadb-DebarunBanerjee, line 89: This is an important point. Does it mean that user could get into similar issues if the server is not restarted ?

1. If this is the case, I think we should try to deal with it, right ? May be by forcing a load.
2. We should have the mrt test that can check the behaviour before server restart.

mariadb-DebarunBanerjee (commented):
Hi Marko,
I could understand the issue and your patch. The approach looks good to me.
A few comments.

-Deb

storage/innobase/btr/btr0btr.cc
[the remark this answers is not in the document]
dr-m: You’re right that the check is redundant. Removed in 01c5855448585338211a7e121e2412abef9eee04.

mysql-test/suite/innodb/t/autoinc_import.test
dr-m: Normally, `ALTER TABLE…IMPORT TABLESPACE` should adjust the `AUTO_INCREMENT` value based on the contents of the `.cfg` file. Importing without a `.cfg` file is rather strongly discouraged. Starting with 89445b64fef2c3c9f2cfb9f572dd19b88f3a48df ([MDEV-26131](https://jira.mariadb.org/browse/MDEV-26131)) we only allow the `PRIMARY KEY` index to exist when importing without a `.cfg` file.

storage/innobase/btr/btr0btr.cc
```diff
@@ -1269,51 +1270,62 @@ btr_read_autoinc(dict_index_t* index)
...
-	mtr.start();
-	buf_block_t*	block = buf_page_get(
-		page_id_t(index->table->space_id, index->page),
-		index->table->space->zip_size(),
-		RW_S_LATCH, &mtr);
+    if (autoinc > max || mysql_version == 0 || mysql_version < 100210)
+    {
+      /* Read MAX(autoinc_col), in case this table had originally been
+      created before MariaDB 10.2.4 introduced persistent AUTO_INCREMENT
+      and MariaDB 10.2.10 fixed MDEV-12123, and there could be a garbage
```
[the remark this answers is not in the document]
dr-m: Theoretically, we could remove the second `if` condition in the code that you quoted, to always reset `PAGE_MAX_TRX_ID` in clustered index pages, including when it corresponds to `PAGE_ROOT_AUTO_INC`.

But, that would make us lose the ability to preserve a persistent `AUTO_INCREMENT` value on `ALTER TABLE…IMPORT TABLESPACE` when not using a `.cfg` file. In [MDEV-11658](https://jira.mariadb.org/browse/MDEV-11658) I would want to get rid of `.cfg` files.

Before [MDEV-26137](https://jira.mariadb.org/browse/MDEV-26137) (9b431d714fdf030e372ed6eeaec0619dcffdde6a) we could not use the original `.frm` file when importing InnoDB data files. Only by using the original `.frm` file it would be possible to augment this condition so that we would preserve legitimate persistent `AUTO_INCREMENT` values.

I think that this suggestion would best be implemented near the end of `row_import_for_mysql()` where we can conveniently access the entire index tree via the buffer pool. At that point, we could invoke this function `btr_read_autoinc_with_fallback()` with appropriate parameters. This should be doable after #3037 has been merged.

storage/innobase/btr/btr0btr.cc
dr-m: You are right, we had better invoke `page_set_autoinc()` also when `page_get_autoinc()` returned 0, no matter what the `mysql_version` is. I will add a test for that.

storage/innobase/btr/btr0btr.cc
dr-m: Good idea.

storage/innobase/btr/btr0btr.cc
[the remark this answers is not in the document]
dr-m: Addressed in efb8c05d352fbf27849866ad8ba8d2c5f822c03f.

mysql-test/suite/innodb/t/autoinc_import.test
dr-m: This is part of the revised commit efb8c05d352fbf27849866ad8ba8d2c5f822c03f.

mariadb-DebarunBanerjee (commented):
Hi Marko,
Thanks for addressing the comments. Some minor comments and doubts.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -14822,12 +14821,34 @@ ha_innobase::check(
 	ulint		n_rows_in_table	= ULINT_UNDEFINED;
 	bool		is_ok		= true;
 	dberr_t		ret;
+        uint handler_flags= check_opt->handler_flags;
```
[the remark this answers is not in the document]
dr-m: This is something that @montywi developed. I am not very familiar with the `CHECK TABLE` code outside InnoDB, and until working on this bug I did not even know that it can rewrite the `.frm` file. It seems to me that implemneting a cleaner solution would require a major refactoring of this API, which could require a nontrivial amount of work in every storage engine, including any possibly externally maintained ones. That would not be something that we would want to do in maintenance releases.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -117,6 +120,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 #include "row0ext.h"
 
 #include <limits>
+#include <myisamchk.h>                          // TT_FOR_UPGRADE
```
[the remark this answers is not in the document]
dr-m: This was contributed by @montywi. I think we may want to avoid a larger refactoring in maintenance releases. It is not nice, but I can live with that.

sql/sql_class.cc
```diff
@@ -5665,6 +5665,37 @@ extern "C" void *thd_mdl_context(MYSQL_THD thd)
...
+  @param op            one of check or repair
+  @param msg_type      one of info, warning or error
+  @param print_to_log  <> 0 if we should also print the message to error log.
+*/
+
+extern "C" void
+print_check_msg(THD *thd, const char *db_name, const char *table_name, const char *op,
+                const char *msg_type, const char *message, my_bool print_to_log)
+{
+  char name[NAME_LEN * 2 + 2];
```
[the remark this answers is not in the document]
dr-m: I will add a `DBUG_ASSERT()` for each string.

storage/innobase/row/row0import.cc
```diff
@@ -4333,6 +4335,37 @@ fil_tablespace_iterate(
...
+{
+  if (!table->persistent_autoinc)
+  {
+    ut_ad(!autoinc);
+    return;
+  }
+
+  if (autoinc)
+  {
+    btr_write_autoinc(dict_table_get_first_index(table), autoinc - 1);
```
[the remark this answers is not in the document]
dr-m: In its current form (with the `goto`), the function expands to about 280 bytes of AMD64 code. The lambda function would actually be emitted separately, occupying 128 bytes (with alignment). The caller would be another 256 bytes or more, resulting in a total size of 128+256 = 384 bytes. I also tried to declare the parameters explicitly, as pass-by-value (`auto autoinc_set = [table, autoinc]() {…};`). That helped a little: 230+100 = 340 bytes. I prefer to optimize for size.

This was an `-O2` debug build, but this particular code does not include any debug assertions, so I think that it is a valid comparison. Curiously, an opportunity for tail recursion was missed, twice even:
```asm
   0x00000000015dd36a <+132>:	call   0x15dd284 <operator()() const>
   0x00000000015dd36f <+137>:	add    $0x18,%rsp
   0x00000000015dd373 <+141>:	pop    %rbx
   0x00000000015dd374 <+142>:	pop    %rbp
   0x00000000015dd375 <+143>:	pop    %r12
   0x00000000015dd377 <+145>:	pop    %r13
   0x00000000015dd379 <+147>:	ret
…
   0x00000000015dd3c3 <+221>:	call   0x15dd284 <operator()() const>
   0x00000000015dd3c8 <+226>:	jmp    0x15dd36f <row_import_autoinc(dict_table_t*, row_prebuilt_t*, uint64_t)+137>
```

storage/innobase/handler/ha_innodb.cc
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Agree. Should be fine if we do it some day :) hopefully.

storage/innobase/row/row0import.cc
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Interesting. I will leave it to you. In general I prefer clarity other than performance critical part of code.

storage/innobase/handler/ha_innodb.cc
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: ok, I just wanted to mention it but don't want to hold the review for it. Let me know if you get more information on it.

mariadb-DebarunBanerjee (approved):
Looks good. Please see if the test/scenario for export/import can be checked.

storage/innobase/row/row0import.cc
```diff
@@ -4333,6 +4335,37 @@ fil_tablespace_iterate(
...
+    sql_print_information("InnoDB: %`.*s.%`s autoinc value set to " UINT64PF,
+                          int(table->name.dblen()), table->name.m_name,
+                          table->name.basename(), autoinc);
+  }
+  else if (TABLE *t= prebuilt->m_mysql_table)
+  {
+    if (const Field *ai= t->found_next_number_field)
+    {
+      autoinc= 1 +
+        btr_read_autoinc_with_fallback(table, innodb_col_no(ai),
```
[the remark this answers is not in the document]
dr-m: With a copy of `tt.ibd` and without this fix, I got the following:
```
mysqltest: At line 91: query 'INSERT INTO t0t VALUES(NULL)' failed: 167: Out of range value for column 'id' at row 1
```
For debugging I had temporarily added the following, which shows the out-of-range `TINYINT` value 129:
```
SHOW CREATE TABLE t0t;
Table	Create Table
t0t	CREATE TABLE `t0t` (
  `id` tinyint(4) NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=129 DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
```
In the final test I will use `INSERT` and `SELECT` to avoid any result differences in future versions that might change the output of `SHOW CREATE TABLE`.

dr-m:
Sorry for the many force-pushes. There was some trouble with the use of the `std_data/autoinc*.frm` files in the test.

PR #3024, opened 2024-01-25
MDEV-33301 memlock with systemd still not working

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-33301*

## Description

CapabilityBoundingSet included CAP_IPC_LOCK in MDEV-9095, however it requires that the executable has the capability marked in extended attributes also.

The alternate to this is raising the RLIMIT_MEMLOCK for the service/ process to be able to complete the mlockall system call. This needs to be adjusted to whatever the MariaDB server was going to allocate. Rather than leave the non-obvious mapping of settings and tuning, add the capability so its easier for the user.

We set the capability, if possible, but may never be used depending on user settings. As such in the Debian postinst script, don't complain if this fails.

The CAP_IPC_LOCK also facilitates the mmaping of huge memory pages. (see man mmap).

## How can this PR be tested?

enable --memlock and a innodb_buffer_pool or similar larger than the default memlock limit (8M on my system). Start service. See if `@@locked_in_memory` is ON after the start.

## PR quality check

### DIFF
--- a/cmake/cpack_rpm.cmake
+++ b/cmake/cpack_rpm.cmake
@@ -163,6 +163,7 @@ SET(CPACK_RPM_server_USER_FILELIST
     ${ignored}
     "%config(noreplace) ${INSTALL_SYSCONF2DIR}/*"
     "%config(noreplace) ${INSTALL_SYSCONFDIR}/logrotate.d/mysql"
+    "%caps(cap_ipc_lock=pe) %{_sbindir}/mysqld"
     )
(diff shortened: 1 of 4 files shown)


illuusio:
As RPM and DEB are working and apparmor need upgrading.. I need to test selinux but I'm not very familiar with it.

LinuxJedi (approved):
Looks good to me, I recommend @dr-m as a second set of eyes. Will likely need to go into 10.5.

PR #3025, opened 2024-01-26
MDEV-32401: expression cache heap-use-after-free

DaveGosselin-MariaDB:
Queries selecting a TEXT column by alias, then joining that with a subquery referencing the same alias may invoke a subquery expression cache.  Typically, these caches expect to rely on heap memory for their backing temporary table, but TEXT (and BLOB) types require an ARIA backing; in this case, a crash may result.  Such a query may look like:

SELECT t1.d FROM (SELECT c AS d FROM t0) AS t1 JOIN t0 ON (
  SELECT ORD((SELECT BIT_OR(d))) FROM t0 GROUP BY d
);

This patch checks if there are blob types present before attempting to create the SQL expression cache and if so, it does not, leaving the associated item cache transparent.

The Expression_cache_tmptable implements the Expression_cache using a temporary table. During Expression_cache_tmptable::init, we expect the created temporary table to have the heap handlerton. However, for the crash in this ticket, it is created using the ARIA handlerton instead. This happens because the Create_tmp_table::choose_engine selects the ARIA engine because the temporary table's associated TABLE_SHARE indicates that there are BLOB fields present. This is true because TEXT fields are handled in the same way as BLOB fields in our system; in our grammar, both TEXT and BLOB types rely on the type_handler_blob handler. After the temporary table is created and the engine associated with it, we then invoke move_field on each field associated with the TABLE_SHARE. However, once this is complete and we return to Expression_cache_tmptable::init, we check to see that the heap hton is used as expected and it is not: so we delete the temporary table. Unfortunately, this leaves the pointers inside the Field objects dangling, and we crash later as indicated in the stack traces on this ticket. Non-BLOB types avoid this problem because they end up with a temporary table backed by the heap handlerton.


DaveGosselin-MariaDB:
I ran `mtr` before I made my test changes, so I missed the fact that subselect is invoked/included into other tests.  I'll re-record those and verify that my test causes only additions to those tests, then I will update the branch.

PR #1131, opened 2019-01-25
MDEV-18374: Add SELinux policy to cracklib_password_check packages

GeoffMontee:

PR #2954, opened 2023-12-20
plugin cracklib password check - selinux

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-______*

## Description

Out of tree builds of -DRPM, with selinux can fail as follows:

[627/2102] Generating mariadb-plugin-cracklib-password-check.pp FAILED: plugin/cracklib_password_check/mariadb-plugin-cracklib-password-check.pp build-mariadb-server-11.4/plugin/cracklib_password_check/mariadb-plugin-cracklib-password-check.pp cd build-mariadb-server-11.4/plugin/cracklib_password_check && /usr/bin/checkmodule -M -m gmariadb-server-11.4/plugin/cracklib_password_check/policy/selinux/mariadb-plugin-cracklib-password-check.te -o build-mariadb-server-11.4/plugin/cracklib_password_check/CMakeFiles/mariadb-plugin-cracklib-password-check-pp.dir/mariadb-plugin-cracklib-password-check.mod && /usr/bin/semodule_package -m build-mariadb-server-11.4/plugin/cracklib_password_check/CMakeFiles/mariadb-plugin-cracklib-password-check-pp.dir/mariadb-plugin-cracklib-password-check.mod -o build-mariadb-server-11.4/plugin/cracklib_password_check/mariadb-plugin-cracklib-password-check.pp /usr/bin/checkmodule:  error opening build-mariadb-server-11.4/plugin/cracklib_password_check/CMakeFiles/mariadb-plugin-cracklib-password-check-pp.dir/mariadb-plugin-cracklib-password-check.mod:  No such file or directory [642/2102] Building CXX object storage/rocksdb/CMakeFiles/rocksdb_tools.dir/rocksdb/tools/ldb_cmd.cc.o ninja: build stopped: subcommand failed.

There isn't a directory to put the temporary artifact in.

There's no need for the temporary artifact to be outside the CMAKE_CURRENT_BINARY_DIR so we just use that path.

## How can this PR be tested?

build with cracklib, -DRPM=, out of tree, with ninja.

If the changes are not amenable to automated testing, please explain why not and carefully describe how to test manually.

## Basing the PR against the correct MariaDB version
- [ ] *This is a new feature and the PR is based against the latest MariaDB development branch.*
- [X] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
[...]

PR #2927, opened 2023-12-13
Explicitly recommend target branches in CODING_STANDARDS

robinnewhouse:
## Description
When making commits, the branches to target are not clearly defined in
public documentation, and frequently change. As the development branch
and the earliest maintained branches progress, it should be reflected in
CODING_STANDARDS.md.

All new code of the whole pull request, including one or several files
that are either new files or modified ones, are contributed under the
BSD-new license. I am contributing on behalf of my employer Amazon Web
Services.

## How can this PR be tested?
Documentation change, no testing necessary.

## Basing the PR against the correct MariaDB version
- [x] *This is a new feature and the PR is based against the latest MariaDB development branch.*
- [ ] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [x] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

## Copyright
All new code of the whole pull request, including one or several files
that are either new files or modified ones, are contributed under the 
BSD-new license. I am contributing on behalf of my employer Amazon Web 
Services.

PR #3026, opened 2024-01-26
Fix ninja build for cracklib_password_check

robinnewhouse:
## Description

As was done in dc77111 for `support-files/CMakeLists.txt`
Do not rely on existence of `CMakeFiles/${target}.dir` directory existence
It is not there for custom targets in Ninja build.

This regression was introduced in #1131 which likely copied the pattern from e79e840 before that regression was addressed in dc77111.

## How can this PR be tested?

A passing build can be seen at https://salsa.debian.org/robinnewhouse/mariadb-server-mirror/-/pipelines/630860.

## PR quality check

## Copyright
Services.

### DIFF
--- a/plugin/cracklib_password_check/CMakeLists.txt
+++ b/plugin/cracklib_password_check/CMakeLists.txt
@@ -32,11 +32,12 @@ IF (HAVE_ALLOCA_H AND HAVE_CRACK_H AND HAVE_LIBCRACK AND HAVE_MEMCPY)
     IF(CHECKMODULE AND SEMODULE_PACKAGE)
       FOREACH(pol mariadb-plugin-cracklib-password-check)
         SET(src ${CMAKE_CURRENT_SOURCE_DIR}/policy/selinux/${pol}.te)
-        SET(tmp ${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/${pol}-pp.dir/${pol}.mod)
+        SET(tmp ${CMAKE_CURRENT_BINARY_DIR}/${pol}.mod)
         SET(out ${CMAKE_CURRENT_BINARY_DIR}/${pol}.pp)
         ADD_CUSTOM_COMMAND(OUTPUT ${out}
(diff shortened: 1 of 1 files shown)


grooverdan (approved):

grooverdan:
Ha, had the same in #2954, but you've explained it better an the cleanup looks better. Let's use this.

LinuxJedi (approved):

LinuxJedi:
@robinnewhouse I have approved this, but unfortunately the final 10.4 release before EoL is for critical updates only. Can you please rebase against 10.5 (or I can if you prefer)? We can then merge.

dlenski:
> @robinnewhouse I have approved this, but unfortunately the final 10.4 release before EoL is for critical updates only. Can you please rebase against 10.5 (or I can if you prefer)? We can then merge.

@LinuxJedi, ever time we inquire about this, we get conflicting answers.

In https://github.com/MariaDB/server/pull/2927#issuecomment-1854578097, @vuvova told us _the opposite_: "10.4 is definitely still maintained and is the target for most bugfixes that affect 10.4 bugs"

vuvova:
I said that on Dec 13, 2023. It still holds, to a degree, 10.4 is still maintained and bugs still are fixed there.

But now (after the very recent 10.4 release) the next 10.4 release will be *the very last 10.4 release*, it reaches EOL, there will be no 10.4 releases after the next one. Meaning, any bug introduced into the last release will never be fixed in 10.4.

Another consideration — users hardly start new projects on 10.4 nowadays, pretty much every use of 10.4 is some old application, old project that have used 10.4 for years. Meaning, any bug we have recently found in 10.4 — they didn't hit it so far and very unlikely will hit it in the future.

Combining these two considerations — yes, we still fix bugs in 10.4, but we're trying to be extra conservative avoid introducing new bugs and breaking 10.4 for people who already use it for years.

It is a generic explanation why pushing fixes into a release series just before EOL is discouraged. It may not necessarily apply to this specific PR

PR #3029, opened 2024-01-30
MDEV-30655  IMPORT TABLESPACE fails with column count or index count mismatch

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-30655*

## Description
Currently import operation fails with schema mismatch when cfg file has hidden fts document id and hidden fts document index. To fix this failure, simply add the fts doc id column, indexes in table definition and do import the table. In case of success:
1) update the fts document id in sys columns.
2) update the number of columns in sys tables.
3) insert the new fts index entry in sys indexes table and sys fields.
4) Reload the table with new table definition

## How can this PR be tested?
./mtr innodb.import_hidden_fts imports the hidden fulltext column

## PR quality check


storage/innobase/row/row0import.cc
```diff
@@ -3073,6 +3126,149 @@ static size_t get_buf_size()
...
+    table->n_v_cols, table->flags,
+    table->flags2);
+
+  new_table->id= table->id;
+  new_table->space_id= table->space_id;
+  size_t col_name_len= 0;
+  /* Copy columns from old table to new fts table */
+  for (int new_i= 0;
+       new_i < new_table->n_cols - (DATA_N_SYS_COLS + 1);
+       new_i++)
```
dr-m, line 3175:
```
D:\Buildbot\amd64-windows\build\storage\innobase\row\row0import.cc(3174,14): warning C4018: '<': signed/unsigned mismatch [D:\Buildbot\amd64-windows\build\storage\innobase\innobase.vcxproj]
```

storage/innobase/row/row0import.cc
```diff
@@ -3073,6 +3126,149 @@ static size_t get_buf_size()
...
+
+  unsigned fts_col_ind= unsigned(table->n_cols - DATA_N_SYS_COLS);
+  fts_add_doc_id_column(new_table, new_table->heap);
+  new_table->cols[fts_col_ind].ind=
+    fts_col_ind & dict_index_t::MAX_N_FIELDS;
+  new_table->cols[fts_col_ind].ord_part= 1;
+  dict_table_add_system_columns(new_table, new_table->heap);
+
+  col_name_len= 0;
+  for (int new_i= 0; new_i < new_table->n_v_cols; new_i++)
```
dr-m, line 3193:
```
D:\Buildbot\amd64-windows\build\storage\innobase\row\row0import.cc(3193,28): warning C4018: '<': signed/unsigned mismatch [D:\Buildbot\amd64-windows\build\storage\innobase\innobase.vcxproj]
```

storage/innobase/row/row0import.cc
```diff
@@ -3073,6 +3126,149 @@ static size_t get_buf_size()
...
+  {
+    dict_col_t old_vcol= table->v_cols[new_i].m_col;
+    ulint pos= old_vcol.ind;
+    if (old_vcol.ind <= fts_col_ind) pos++;
+    dict_mem_table_add_v_col(new_table, new_table->heap,
+                             &table->v_col_names[col_name_len],
+                             old_vcol.mtype, old_vcol.prtype,
+                             old_vcol.len, pos,
+                             table->v_cols[new_i].num_base);
+    for (int i= 0; i < table->v_cols[new_i].num_base; i++)
```
dr-m, line 3203:
```
D:\Buildbot\amd64-windows\build\storage\innobase\row\row0import.cc(3203,22): warning C4018: '<': signed/unsigned mismatch [D:\Buildbot\amd64-windows\build\storage\innobase\innobase.vcxproj]
```

dr-m (changes requested):
Just noting some build errors.

storage/innobase/row/row0import.cc
```diff
@@ -4246,6 +4442,173 @@ fil_tablespace_iterate(
...
+{
+  if (cfg.m_missing || !cfg.has_hidden_fts()) return false;
+
+  ulint col_name_len= 0;
+  for (ulint i= 0; i < table->n_cols; i++)
+  {
+     if (!strcmp(&table->col_names[col_name_len],
+                 FTS_DOC_ID_COL_NAME))
+       return false;
+     col_name_len += strlen(&table->col_names[col_name_len]) + 1;
```
dr-m, line 4600: We could replace `col_name_len` with a straight pointer.
```c++
  const char *col_name= table->col_names;
  for (unsigned i= 0; i < table->n_cols; i++, col_name+= strlen(col_name) + 1)
    if (!strcmp(col_name, FTS_DOC_ID_COL_NAME)
       return false;
```
But, better yet, maybe we should just replace the entire function with the following:
```c++
return !cfg.m_missing && cfg.has_hidden_fts() && !table->fts_doc_id_index;
```
The `m_missing` check might better be part of `bool row_import::has_hidden_fts() const`.

storage/innobase/row/row0import.cc
```diff
@@ -196,6 +196,27 @@ struct row_import {
...
+	  return ULINT_UNDEFINED;
+	}
+
+	/** @return Check whether cfg file has FTS_DOC_ID
+	& FTS_DOC_ID_INDEX_NAME*/
+	bool has_hidden_fts()
+	{
+          if (ULINT_UNDEFINED != find_col(FTS_DOC_ID_COL_NAME))
+	      return find_fts_idx_offset() != ULINT_UNDEFINED;
+          return false;
```
dr-m, line 218: What about the data type of the column? And the `FTS_DOC_ID_INDEX`?

storage/innobase/row/row0import.cc
```diff
@@ -3073,6 +3126,149 @@ static size_t get_buf_size()
...
+  /* Add fields for FTS_DOC_ID_INDEX */
+  dict_index_add_col(
+    fts_index, table,
+    &table->cols[table->n_cols - (DATA_N_SYS_COLS + 1)], 0);
+  dict_index_t *clust_index= UT_LIST_GET_FIRST(table->indexes);
+  for (ulint i= 0; i < clust_index->n_uniq; i++)
+    dict_index_add_col(fts_index, table, clust_index->fields[i].col,
+                       clust_index->fields[i].prefix_len);
+  fts_index->search_info= btr_search_info_create(fts_index->heap);
+  fts_index->search_info->ref_count= 0;
```
dr-m, line 3147: This would break the `cmake -DWITH_INNODB_AHI=OFF` build. Please add
```c++
#ifdef BTR_CUR_HASH_ADAPT
#endif
```
around any `search_info` references.

storage/innobase/row/row0import.cc
```diff
@@ -4246,6 +4442,173 @@ fil_tablespace_iterate(
...
+
+  err= que_eval_sql(info,
+                    "PROCEDURE UPDATE_N_COLS () IS\n"
+                    "BEGIN\n"
+                    "UPDATE SYS_TABLES SET N_COLS = :n"
+                    " WHERE ID = :id;\n"
+                    "END;\n", trx);
+  if (err != DB_SUCCESS)
+    my_error(ER_INTERNAL_ERROR, MYF(0),
+             "InnoDB: Updating SYS_TABLES.N_COLS failed");
```
dr-m, line 4580: I think that the user only needs to know that updating the data dictionary failed. Where exactly, it should not matter, because there hardly is anything that a user can do to correct such errors. Can we use just one procedure for everything? And add a `DBUG_EXECUTE_IF` to inject an error, so that we can check that no memory is leaked.

I checked that this transaction would be committed by `prebuilt->trx->commit()` in `row_import_cleanup()`. I did not find a call to a rollback anywhere. It would seem that in case of an error, `row_import_error()` would invoke `row_import_cleanup()`, which would wrongly commit any partially executed changes to the data dictionary tables.

dr-m (changes requested):
I think that we also need a debug instrumented test, to cover the error handling logic.

storage/innobase/row/row0import.cc
```diff
@@ -2140,11 +2193,42 @@ row_import_cleanup(
...
+						fts_table->indexes);
+			while (index) {
+				dict_index_t *next_index =
+				  UT_LIST_GET_NEXT(indexes, index);
+				dict_index_remove_from_cache(
+						fts_table, index);
+				dict_mem_index_free(index);
+				index = next_index;
+			}
+			dict_mem_table_free(fts_table);
```
dr-m, line 2229:
```
==154785==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55e711fa0a8c in dict_mem_index_free(dict_index_t*) /home/buildbot/amd64-debian-11-msan/build/storage/innobase/dict/dict0mem.cc:1082:2
    #1 0x55e7119a725e in row_import_cleanup(row_prebuilt_t*, dberr_t, dict_table_t*) /home/buildbot/amd64-debian-11-msan/build/storage/innobase/row/row0import.cc:2226:5
    #2 0x55e71199ed1e in row_import_for_mysql(dict_table_t*, row_prebuilt_t*) /home/buildbot/amd64-debian-11-msan/build/storage/innobase/row/row0import.cc:4932:9
    #3 0x55e7114ff30d in ha_innobase::discard_or_import_tablespace(char) /home/buildbot/amd64-debian-11-msan/build/storage/innobase/handler/ha_innodb.cc:13417:9
…
  Uninitialized value was created by a heap deallocation
    #0 0x55e70f46a3ec in free (/home/buildbot/amd64-debian-11-msan/build/sql/mariadbd+0x8503ec) (BuildId: fd22fd2a58de3e751344727a72a89bbf6db2a776)
    #1 0x55e711fa08a0 in mem_heap_free(mem_block_info_t*) /home/buildbot/amd64-debian-11-msan/build/storage/innobase/include/mem0mem.inl:419:3
    #2 0x55e711fa08a0 in dict_mem_index_free(dict_index_t*) /home/buildbot/amd64-debian-11-msan/build/storage/innobase/dict/dict0mem.cc:1096:2
```
It’s nice to see that in this special case, MSAN catches the same error as ASAN would:
```
==191432==ERROR: AddressSanitizer: heap-use-after-free on address 0x61600ba70f54 at pc 0x555d6ca84589 bp 0x7f448034aba0 sp 0x7f448034ab98
READ of size 4 at 0x61600ba70f54 thread T10
    #0 0x555d6ca84588 in dict_mem_index_free(dict_index_t*) /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/storage/innobase/dict/dict0mem.cc:1082:2
    #1 0x555d6c727f2e in row_import_cleanup(row_prebuilt_t*, dberr_t, dict_table_t*) /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/storage/innobase/row/row0import.cc:2226:5
    #2 0x555d6c72385a in row_import_for_mysql(dict_table_t*, row_prebuilt_t*) /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/storage/innobase/row/row0import.cc:4932:9
    #3 0x555d6c497271 in ha_innobase::discard_or_import_tablespace(char) /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/storage/innobase/handler/ha_innodb.cc:13417:9
    #4 0x555d6b56daaa in mysql_discard_or_import_tablespace(THD*, TABLE_LIST*, bool) /home/buildbot/amd64-ubuntu-2204-clang14-asan/build/sql/sql_table.cc:5553:35
```

storage/innobase/row/row0import.cc
```diff
@@ -196,6 +196,57 @@ struct row_import {
 
 	dberr_t match_flags(THD *thd) const ;
 
+	ulint find_fts_idx_offset() const
+	{
+	  for (ulint i= 0; i < m_n_indexes; i++)
+	  {
+            const char* index_name= reinterpret_cast<const char*>(
+                                      m_indexes[i].m_name);
```
dr-m, line 204: The formatting does not correspond to any single accepted style.

dr-m (changes requested):

storage/innobase/row/row0import.cc
```diff
@@ -4332,16 +4625,28 @@ row_import_for_mysql(
...
+		if (cfg.need_hidden_fts(table)) {
+			dict_table_t *fts_table =
+				build_fts_hidden_table(
+					table, thd, cfg);
+			if (fts_table == nullptr) {
+				return row_import_error(
+						prebuilt, DB_ERROR);
+			}
+			table = fts_table;
+			cfg.m_table = fts_table;
```
dr-m, line 4644: We do not seem to need the local variable `fts_table`:
```c++
cfg.m_table = table = build_fts_hidden_table(table, thd, cfg);
if (!table) {
	err = DB_ERROR;
	goto import_error;
}
```
Do we even need `table`, or could we use `cfg.m_table` instead?

storage/innobase/row/row0import.cc
```diff
@@ -4279,6 +4527,50 @@ static void row_import_autoinc(dict_table_t *table, row_prebuilt_t *prebuilt,
...
+  pars_info_add_int4_literal(info, "pos", fts_pos);
+  pars_info_add_int4_literal(info, "space", fts_idx->table->space_id);
+  pars_info_add_int4_literal(info, "page_no", fts_idx->page);
+
+  return que_eval_sql(info,
+                      "PROCEDURE ADD_FTS_COL () IS\n"
+                      "BEGIN\n"
+                      "UPDATE SYS_COLUMNS SET\n"
+                      "POS = POS + 1\n"
+                      "WHERE TABLE_ID = :id AND POS >= :pos;\n"
```
dr-m, line 4561: Do we need this? Shouldn’t the hidden `FTS_DOC_ID` (which we are adding) always be the last column?

I can imagine that we might need this because of virtual columns? If that is the case, then we should update `SYS_VIRTUAL` as well.

storage/innobase/row/row0import.cc
```diff
@@ -3075,6 +3178,151 @@ static size_t get_buf_size()
...
+}
+
+/** Append the hidden fts column and fts doc index to the
+existing table
+@param  table  table to be imported
+@param  thd    thread
+@param  cfg    metadata required by import
+@return table which has fts doc id and fts doc id index */
+static dict_table_t *build_fts_hidden_table(
+  dict_table_t *table, THD *thd, row_import &cfg)
```
dr-m, line 3212: The parameter `thd` is unused. The parameter `cfg` needs to be a `const` reference.

storage/innobase/row/row0import.cc
```diff
@@ -3075,6 +3178,151 @@ static size_t get_buf_size()
...
+  for (ulint new_i= 0;
+       new_i < ulint(new_table->n_cols - (DATA_N_SYS_COLS + 1));
+       new_i++)
+  {
+    dict_mem_table_add_col(new_table, new_table->heap,
+                           &table->col_names[col_name_len],
+			   table->cols[new_i].mtype,
+			   table->cols[new_i].prtype,
+			   table->cols[new_i].len);
+    col_name_len+= strlen(&table->col_names[col_name_len]) + 1;
```
dr-m, line 3236: Instead of maintaining an offset `col_name_len` we could maintain a pointer `col_name`:
```c++
col_name+= strlen(col_name) + 1;
```

storage/innobase/row/row0import.cc
```diff
@@ -3075,6 +3178,151 @@ static size_t get_buf_size()
...
+      new_table->v_cols[new_i].base_col[i]= base_col;
+    }
+    col_name_len+= strlen(&table->v_col_names[col_name_len]) + 1;
+  }
+
+  /* Copy indexes from old table to new table */
+  for (dict_index_t *old_index= UT_LIST_GET_FIRST(table->indexes);
+       old_index;)
+  {
+    bool is_clustered= (old_index->type & DICT_CLUSTERED);
```
dr-m, line 3270: Simpler:
```c++
bool is_clustered= true;
for (dict_index_t *old_index= UT_LIST_GET_FIRST(table->indexes); old_index; is_clustered= false;)
```

storage/innobase/row/row0import.cc
```diff
@@ -3075,6 +3178,151 @@ static size_t get_buf_size()
...
+    dict_index_t *new_index= dict_mem_index_create(
+      new_table, old_index->name, old_index->type,
+      is_clustered ? old_index->n_fields + 1 : old_index->n_fields);
```
dr-m, line 3273: `old_index->n_field + is_clustered` would replace a conditional branch with arithmetics.

storage/innobase/row/row0import.cc
```diff
@@ -3075,6 +3178,151 @@ static size_t get_buf_size()
...
+    for (ulint new_i= 0; new_i < new_index->n_fields; new_i++)
+    {
+      if (is_clustered && new_i == old_index->n_fields)
+      {
+        /* Add fts doc id in clustered index */
+        dict_index_add_col(
+          new_index, new_table, &table->cols[fts_col_ind], 0);
+        new_index->fields[new_i].fixed_len= sizeof(doc_id_t);
+        continue;
+      }
```
dr-m, line 3289: Can we move the `if` body after the body of the `for` loop? It would seem to execute at the end of the loop.

storage/innobase/row/row0import.cc
```diff
@@ -3075,6 +3178,151 @@ static size_t get_buf_size()
...
+      else
+      {
+        unsigned ind= field->col->ind;
+        if (ind >= fts_col_ind) ind++;
+	col= &new_table->cols[ind];
+      }
+      dict_index_add_col(new_index, new_table, col,
+                         field->prefix_len);
+      if (new_i < old_index->n_uniq) col->ord_part= 1;
```
dr-m, line 3306: For importing the table, I do not think that `dict_col_t::ord_part` matters.

I think that we’d better free this created `dict_table_t` object at the end of the import. In case the dictionary was modified, we’d better reload the table definition from the dictionary tables.

storage/innobase/row/row0import.cc
```diff
@@ -3075,6 +3178,151 @@ static size_t get_buf_size()
...
+    }
+
+    if (old_index->search_info)
+    {
+      new_index->search_info=
+        btr_search_info_create(new_index->heap);
+      new_index->search_info->ref_count=
+        old_index->search_info->ref_count;
+    }
```
dr-m, line 3315: This should not matter for importing the data file.

dr-m (commented):
Thank you, this is steadily getting better.

You mentioned that you wanted to test what happens when virtual columns are present in the table. I suspect that the update of `SYS_COLUMNS` is only needed when also `SYS_VIRTUAL` needs to be updated.

It would be great if we could load the table definition from the data dictionary already at the start of the IMPORT TABLESPACE workflow. I understand that it is not feasible, because we do not want to create two dictionary transactions and we do not want to hold exclusive locks on the dictionary tables while the time-consuming rewrite of the tablespace is in progress.

What I think we can do better is to always free the table definition "stub" that is created by `build_fts_hidden_table()`, and to reload the `prebuilt->table` from the data dictionary in case `innodb_insert_hidden_fts_col()` was successful.

mysql-test/suite/innodb/t/import_hidden_fts.test
```diff
@@ -0,0 +1,46 @@
...
+ALTER TABLE t1 ADD COLUMN f6 INT NOT NULL;
+ALTER TABLE t1 DROP COLUMN f6;
+ALTER TABLE t1 DROP INDEX f_idx;
+connect(con1,localhost,root,,);
+START TRANSACTION WITH CONSISTENT SNAPSHOT;
+
+connection default;
+DELETE FROM t1 WHERE f1 > 1;
+FLUSH TABLE t1 FOR EXPORT;
+let MYSQLD_DATADIR =`SELECT @@datadir`;
```
dr-m, line 22: This test hung in an `./mtr --embedded` test [somewhere](https://buildbot.mariadb.org/#/builders/236/builds/19097/steps/6/logs/stdio) after the `FLUSH TABLE t1 FOR EXPORT`:
```
innodb.import_hidden_fts 'innodb'        w11 [ fail ]  timeout after 900 seconds
        Test ended at 2024-02-15 21:58:19
Test case timeout after 900 seconds
```
Because there are no stack traces of the right process, I ran this test on my own system:
```sh
./mtr --embedded innodb.import_hidden_fts
```
The test is taking a very long time on my local system as well. I would suggest the following addition to disable the row-level undo logging, because we should not really need it here:
```diff
diff --git a/mysql-test/suite/innodb/t/import_hidden_fts.test b/mysql-test/suite/innodb/t/import_hidden_fts.test
index 39bb6c62ca7..2cdd397b08c 100644
--- a/mysql-test/suite/innodb/t/import_hidden_fts.test
+++ b/mysql-test/suite/innodb/t/import_hidden_fts.test
@@ -8,8 +8,11 @@ CREATE TABLE t1(f1 INT NOT NULL PRIMARY KEY,
 		f3 INT as (f1) VIRTUAL, INDEX(f3),
 		f4 INT as (f1) STORED, INDEX(f4),
 		f5 INT as (f1) VIRTUAL)ENGINE=InnoDB;
-INSERT INTO t1(f1, f2) VALUES(1, "on");
-INSERT INTO t1(f1, f2) SELECT seq, "si" FROM seq_2_to_16384;
+SET STATEMENT unique_checks=0,foreign_key_checks=0 FOR
+START TRANSACTION;
+INSERT INTO t1(f1, f2) VALUES(1, 'on');
+INSERT INTO t1(f1, f2) SELECT seq, 'si' FROM seq_2_to_16384;
+COMMIT;
 ALTER TABLE t1 ADD COLUMN f6 INT NOT NULL;
 ALTER TABLE t1 DROP COLUMN f6;
 ALTER TABLE t1 DROP INDEX f_idx;
```
With this fix, the test would complete in 13.9 seconds instead of 14.5 seconds on my system. That is still a lot. What do we need to insert that many rows for? If I create a table with 16 rows instead of 16384, the test will complete in 0.13 seconds.

I was curious where the time is being spent. At least `rbt_check_ordering()` seems to be popular, and `fts_doc_id_cmp()` is there as well.

storage/innobase/row/row0import.cc
```diff
@@ -2115,14 +2176,30 @@ dberr_t PageConverter::operator()(buf_block_t* block) UNIV_NOTHROW
...
+static void reload_fts_table(row_prebuilt_t *prebuilt,
+                             dict_table_t* table)
+{
+
+  /* Reload the table in case of hidden fts column */
+  const table_id_t id= prebuilt->table->id;
+  prebuilt->table->release();
+  dict_sys.remove(prebuilt->table);
+  prebuilt->table= dict_table_open_on_id(
+                      id, true, DICT_TABLE_OP_NORMAL);
```
dr-m, line 2188: Splitting a line after `(` was only allowed in the archaic InnoDB formatting style. Better split the line after the `=`.

storage/innobase/row/row0import.cc
```diff
@@ -2115,14 +2176,30 @@ dberr_t PageConverter::operator()(buf_block_t* block) UNIV_NOTHROW
...
+  prebuilt->table->space= table->space;
```
dr-m, line 2189: Shouldn’t these fields already be equal?
```c++
  ut_ad(prebuilt->table->space == table->space);
```
It would also be good to add `ut_ad(prebuilt->table != table);` at the start of the function, because otherwise the call to `dict_sys.remove()` could free `table`.

storage/innobase/row/row0import.cc
```diff
@@ -2115,14 +2176,30 @@ dberr_t PageConverter::operator()(buf_block_t* block) UNIV_NOTHROW
...
+@return error code or DB_SUCCESS */
+static
 dberr_t
 row_import_cleanup(
-/*===============*/
-	row_prebuilt_t*	prebuilt,	/*!< in/out: prebuilt from handler */
-	dberr_t		err)		/*!< in: error code */
+	row_prebuilt_t*	prebuilt,
+	dberr_t		err,
+	dict_table_t*	fts_table = nullptr)
```
dr-m, line 2202: If you are reformatting the parameters, you could as well format them in the common style (no TABs, multiple parameters per line).

There is a mismatch between `@param table` and the parameter name `fts_table`. The comment does not really explain what the parameter is for.

storage/innobase/row/row0import.cc
```diff
@@ -2142,11 +2219,46 @@ row_import_cleanup(
...
 			index->page = FIL_NULL;
 		}
+
+		prebuilt->trx->rollback();
+	}
+	else
+	{
+	  DBUG_EXECUTE_IF("ib_import_before_commit_crash", DBUG_SUICIDE(););
+	  prebuilt->trx->commit();
 	}
```
dr-m, line 2229: Please use a consistent formatting style.

storage/innobase/row/row0import.cc
```diff
@@ -4349,16 +4671,27 @@ row_import_for_mysql(
...
 	if (err == DB_SUCCESS) {
 
-		if (dberr_t err = handle_instant_metadata(table, cfg)) {
-			return row_import_error(prebuilt, err);
+		if (cfg.need_hidden_fts(table)) {
+			cfg.m_table = table = build_fts_hidden_table(
+						table, cfg);
+			if (!table) {
+				return row_import_error(
+						prebuilt, DB_ERROR);
```
dr-m, line 4686: This could be replaced with
```c++
err = DB_ERROR;
goto import_error;
```

storage/innobase/row/row0import.cc
```diff
@@ -4524,7 +4857,7 @@ row_import_for_mysql(
 	dict_index_t*	index = dict_table_get_first_index(table);
 
 	if (!dict_index_is_clust(index)) {
-		return row_import_error(prebuilt, DB_CORRUPTION);
+		goto import_error;
```
dr-m, line 4860: Should we not set `err` first?

storage/innobase/row/row0import.cc
```diff
@@ -4459,7 +4792,7 @@ row_import_for_mysql(
 				table_name, ut_strerr(err));
 		}
 
-		return row_import_cleanup(prebuilt, err);
+		return row_import_cleanup(prebuilt, err, table);
```
dr-m, line 4795: `goto import_error;`

storage/innobase/row/row0import.cc
```diff
@@ -4279,6 +4529,77 @@ static void row_import_autoinc(dict_table_t *table, row_prebuilt_t *prebuilt,
...
+    dict_table_get_index_on_name(table, FTS_DOC_ID_INDEX_NAME);
+  if (!fts_idx) return DB_ERROR;
+
+  for (ulint new_i= 0; new_i < table->n_v_cols; new_i++)
+  {
+    ulint pos= dict_create_v_col_pos(
+                 table->v_cols[new_i].v_pos,
+		 table->v_cols[new_i].m_col.ind);
+    dberr_t err= update_vcol_pos(table->id, pos, trx);
+    if (err != DB_SUCCESS)
```
dr-m, line 4576: The idiomatic C++98 way to write this would be as follows:
```c++
if (dberr_t err= update_vcol_pos(table->id, pos, trx))
```
However, why do we invoke `update_vcol_pos()` several times? Does the `WHERE` condition really have to check equality with `pos`?

Is it possible that `pos` are assigned in ascending order in this loop? If yes, there would seem to be a possible duplicate key error  collision on `SYS_VIRTUAL` if there are multiple virtual columns defined on the same base column(s). I see that the test case `innodb.import_hidden_fts` does define two virtual columns that are derived from the same base column, but only one of them is indexed. Could that be the reason why no bug was caught?

Can we add something to ensure that the `pos` are assigned in ascending order? Or compute the exact bound that needs to be passed to `UPDATE SYS_… SET POS=POS+1 WHERE TABLE_ID=:id AND POS>=:old_pos`?

storage/innobase/row/row0import.cc
```diff
@@ -4279,6 +4529,77 @@ static void row_import_autoinc(dict_table_t *table, row_prebuilt_t *prebuilt,
...
+  dict_hdr_get_new_id(NULL, &fts_idx->id, NULL);
+  pars_info_add_ull_literal(info, "idx_id", fts_idx->id);
+  pars_info_add_int4_literal(info, "pos", fts_pos);
+  pars_info_add_int4_literal(info, "space", fts_idx->table->space_id);
+  pars_info_add_int4_literal(info, "page_no", fts_idx->page);
+
+  return que_eval_sql(info,
+                      "PROCEDURE ADD_FTS_COL () IS\n"
+                      "BEGIN\n"
+                      "INSERT INTO SYS_COLUMNS VALUES"
```
dr-m, line 4591: As far as I can tell, we should be able to update `SYS_TABLES` and `SYS_VIRTUAL` here just fine. The `WHERE` conditions for those tables could match 0 rows when there are no virtual columns.

dr-m (approved):

PR #3030, opened 2024-01-30
MDEV-33853 Async rollback prepared transactions during binlog crash recovery

SongLibing:
MDEV-33853 Async rollback prepared transactions during binlog
           crash recovery

Summary
=======
When doing server recovery, the active transactions will be rolled
back by InnoDB background rollback thread automatically. The
prepared transactions will be committed or rolled back accordingly
by binlog recovery. Binlog recovery is done in main thread before
the server can provide service to users. If there is a big
transaction to rollback, the server will not available for a long
time.

This patch provides a way to rollback the prepared transactions
asynchronously. Thus the rollback will not block server startup.

Design
======
- Handler::recover_rollback_by_xid()
  This patch provides a new handler interface to rollback transactions
  in recover phase. InnoDB just set the transaction's state to active.
  Then the transaction will be rolled back by the background rollback
  thread.

- Handler::signal_tc_log_recover_done()
  This function is called after tc log is opened(typically binlog opened)
  has done. When this function is called, all transactions will be rolled
  back have been reverted to ACTIVE state. Thus it starts rollback thread
  to rollback the transactions.

- Background rollback thread
  With this patch, background rollback thread is defered to run until binlog
  recovery is finished. It is started by innobase_tc_log_recovery_done().

- [x] *The Jira issue number for this PR is: MDEV-33853*

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check


storage/innobase/trx/trx0roll.cc
```diff
@@ -776,13 +779,18 @@ void trx_rollback_recovered(bool all)
...
         trx_free_at_shutdown(trx);
-      }
-      else
+      } else {
         trx->free();
+        DBUG_EXECUTE_IF("print_log_after_rollback_on_recovery", {
+          ib::info() << "One trx rollback by recovery_rollback_thread "
+                     << current_thd << "\n";
+          });
+      }
```
dr-m, line 788: I do not see any test coverage of this. I am not a friend of such added printouts; such information can be better collected with `./mtr --rr` and checked in `rr replay`. If this is really needed, then I would suggest `DBUG_PRINT`, which should be disabled by `cmake -DWITH_DBUG_TRACE=OFF`.

storage/innobase/trx/trx0roll.cc
```diff
@@ -796,7 +804,23 @@ void trx_rollback_all_recovered(void*)
...
+			/* trx_rollback_all_recovered is called after finished is tested,
+			it guarantees that all active transactions are handled. */
+			finished = ddl_recovery_done || srv_shutdown_state != SRV_SHUTDOWN_NONE;
+
+			trx_rollback_recovered(true);
+
+			/* Sleep 10ms to wait for MySQL recover to revert some prepared
+			transactions to active state. */
+			std::this_thread::sleep_for(std::chrono::milliseconds(10));
+		} while (!finished);
```
dr-m, line 822: This looks prone to race conditions. I think that it would be better to initiate another rollback in or after `innodb_ddl_recovery_done()`.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17080,6 +17080,52 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+It only set the transaction's status to ACTIVE and persist the status.
+The transaction will be rolled back by background rollback thread.
+
+@param[in] hton InnoDB handlerton
+@param[in] xid X/Open XA transaction identification
+
+@return 0 or error number */
+int innobase_recover_rollback_by_xid(handlerton *hton, XID *xid)
+{
+	DBUG_ASSERT(hton == innodb_hton_ptr);
```
dr-m, line 17096: New functions in InnoDB should be formatted in the common style. No `[in]` after `@param`, and indent by 2 spaces per level instead of TABs.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17080,6 +17080,52 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+	DBUG_EXECUTE_IF("innobase_xa_fail", return XAER_RMFAIL;);
+	DBUG_EXECUTE_IF("crash_before_async_binlog_recovery", DBUG_SUICIDE(););
```
dr-m, line 17098: I do not see any test coverage of this crash injection point.

I would strongly prefer to use `DEBUG_SYNC` for any crash injection. Simply make the server hang on
```sql
SET DEBUG_SYNC='some_label WAIT_FOR ever';
```
and let the test client kill and restart the server. In that way, it is readily compatible with Valgrind.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17080,6 +17080,52 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+		return(XAER_RMFAIL);
+	}
+
+	trx_t *trx = trx_get_trx_by_xid(xid);
+	if (trx == NULL) {
+		return(XAER_RMFAIL);
+	}
+
+  // ddl should not be rolled back through recovery
+  ut_ad(!trx->dict_operation);
```
dr-m, line 17110: More importantly, this is missing
```c++
  ut_ad(trx->is_recovered);
  ut_ad(trx->state == TRX_STATE_PREPARED_RECOVERED);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17080,6 +17080,52 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+	if (wsrep_is_wsrep_xid(&trx->xid)) {
+		trx->xid.null();
+	}
+#endif /* WITH_WSREP */
+
+	trx_rollback_for_mysql(trx, true /* fast_rollback */);
+
+	/* Change the trx state to TRX_STATE_ACTIVE. Background rollback thread
+	will rollback this trx later. */
+	trx->state = TRX_STATE_ACTIVE;
```
dr-m, line 17125: This new state change is not documented in `trx0trx.h`.

I think that it would be good to mention in a comment that only one call of `trx_get_trx_by_xid()` for any given XID can find a transaction object. Subsequent calls by other threads would return `nullptr`. That is what guarantees that no other thread can be modifying the state of the transaction at this point.

What is preventing a race condition between `trx_rollback_all_recovered()` and this assignment? What guarantees that each of these transactions will actually be rolled back? Wouldn’t it be safer to leave the `trx->state` as is, and instead add the transaction to a list that would be processed later? I would suggest to avoid changing `trx_rollback_for_mysql()` at all, and simply invoking `trx_undo_set_state_at_prepare()` here.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17080,6 +17080,52 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+  ut_ad(!trx->dict_operation);
+
+#ifdef WITH_WSREP
+	/* If a wsrep transaction is being rolled back during
+	the recovery, we must clear the xid in order to avoid
+	writing serialisation history for rolled back transaction. */
+	if (wsrep_is_wsrep_xid(&trx->xid)) {
+		trx->xid.null();
+	}
+#endif /* WITH_WSREP */
```
dr-m, line 17119: For other transactions, this is done as part of `trx_get_trx_by_xid_callback()`, protected by `trx->mutex`. There seems to be a potential race condition with this assignment: Multiple concurrent threads could get the same object returned by `trx_get_trx_by_xid()` in case `wsrep_is_wsrep_xid(&trx->xid)` holds. Can this condition ever hold in this code path? Could we simply write the following?
```c++
#ifdef WITH_WSREP
  ut_ad(!wsrep_is_wsrep_xid(&trx->xid));
#endif
```

storage/innobase/handler/ha_innodb.cc
dr-m, line 17096: I would not add an unused parameter `hton`, and I would pass `const XID &xid` because the parameter is not supposed to be `nullptr` nor to be modified by the callee.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1896,10 +1897,8 @@ ulonglong ha_innobase::table_version() const
   return m_prebuilt->trx_id;
 }
 
-#ifdef UNIV_DEBUG
 /** whether the DDL log recovery has been completed */
-static bool ddl_recovery_done;
-#endif
+std::atomic_bool ddl_recovery_done { false };
```
dr-m, line 1901: This would open up possibilities for many race conditions. I do not think that it is a good idea.

Based on other changes around this, I understood that some binlog recovery would be executed before `innodb_ddl_recovery_done()` is invoked. It would seem cleaner to me to complete any lower-level recovery first.

sql/handler.cc
```diff
@@ -2792,7 +2802,11 @@ static my_bool xarecover_handlerton(THD *unused, plugin_ref plugin,
         }
         else if (!info->mem_root)
         {
-          int rc= hton->rollback_by_xid(hton, info->list+i);
+          int rc= 0;
+          if (hton->recover_rollback_by_xid)
+            rc= hton->recover_rollback_by_xid(hton, info->list+i);
+          else
+            rc= hton->rollback_by_xid(hton, info->list+i);
```
dr-m, line 2809: I think that we must assert or check here if `wsrep_is_wsrep_xid(info->list+i)` holds. If yes, then we’d better invoke `hton->rollback_by_xid()` in the old-fashioned way, to avoid any race condition.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
```diff
@@ -0,0 +1,82 @@
...
+INSERT INTO t1 VALUES (0, 1);
+
+--echo #
+--echo # 1. Check DML in prepared state can rollback correctly.
+--echo #
+
+--source include/expect_crash.inc
+SET debug = "d,crash_before_writing_xid";
+--error 2013
+UPDATE t1 SET data = data + 1 WHERE id = 0;
```
dr-m, line 15: It would be better to hold up the server with `DEBUG_SYNC` in a connection and then invoke `source include/kill_mysqld.inc` in `connection default`.

This test does not seem to work with `./mtr --ps-protocol`.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
```diff
@@ -0,0 +1,82 @@
...
+
+# Crash again.
+--source include/expect_crash.inc
+SET debug = "d,crash_dispatch_command_before";
+--error 2013
+SELECT 1;
+
+# Restart the server.
+--let $restart_parameters = --debug=d,print_log_after_rollback_on_recovery
+--source include/start_mysqld.inc
```
dr-m, line 62: Can we test multiple scenarios across a single server restart? I see that this short test is currently taking more than 7 seconds to execute.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
```diff
@@ -0,0 +1,82 @@
...
+
+# Check whether the DML trx have been rollback asynchronously.
+# For the first crash recovery, prepare transaction become
+# active, no log print because rollback not finish.
+# For the second crash recovery, rollback active trx asynchronously.
+--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_text= Server reported: recovery_rollback_thread
+--let $assert_select= recovery_rollback_thread
+--let $assert_count= 2
+--source include/assert_grep.inc
```
dr-m, line 79: I would suggest to use `include/search_pattern_in_file.inc` instead.

storage/innobase/trx/trx0roll.cc
SongLibing: Hi Marko,
Thanks for the quick and thorough review. 

Starting a new rollback thread seems simpler. But there are two problems. 

First, at the time, the original rollback thread may not exit yet. Thus there will be two rollback threads in parallel. I am not sure if that will work well on MariaDB. It doesn't work on MySQL. We had another patch to initialize more than one rollback threads, thus the big transaction's rollback will not block the rollback of small transactions. There are some statistic variables are global. That need to be fix. 

Second, when server shutdown, it will wait until all active transactions are removed from trx_list. However, the server may exit before ddl_recovery_done, thus there is no chance to start a rollback thread to clear active transactions.

storage/innobase/trx/trx0roll.cc
```diff
@@ -267,9 +267,12 @@ dberr_t trx_rollback_for_mysql(trx_t* trx)
 			change will be durable as well. */
 			mtr.commit();
 			ut_ad(mtr.commit_lsn() > 0);
+			if (fast_rollback) {
+				log_write_up_to(mtr.commit_lsn(), true);
+				return(DB_SUCCESS);
+			}
 		}
 		return(trx_rollback_for_mysql_low(trx));
```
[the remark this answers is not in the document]
SongLibing: really good idea.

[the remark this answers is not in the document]
SongLibing: call ha_flush_logs() at ha_recover_complete() instead.

storage/innobase/handler/ha_innodb.cc
SongLibing: added the assertions.

storage/innobase/handler/ha_innodb.cc
SongLibing: Fixed

storage/innobase/handler/ha_innodb.cc
SongLibing: done

storage/innobase/handler/ha_innodb.cc
SongLibing: sure, addressed.

sql/handler.cc
SongLibing: done

storage/innobase/trx/trx0roll.cc
SongLibing: It is used as restart-parameter when restarting the server, then rollback thread will log a message for every finished trx. the test will check existing of the log.

storage/innobase/handler/ha_innodb.cc
SongLibing: right, removed 'crash_before_async_binlog_recovery'.

storage/innobase/handler/ha_innodb.cc
SongLibing: addressed.

storage/innobase/trx/trx0roll.cc
SongLibing: It seems it is safe to create multiple rollback thread on MariaDB. I will try to follow your suggestions.

grooverdan:
@SongLibing are you ok to finish this?

SongLibing:
> @SongLibing are you ok to finish this?

Sorry for the delay. I have been very busy these days. I can probably finish the update in next week.

storage/innobase/trx/trx0roll.cc
SongLibing: As we discussed, This patch will reuse the original recovery rollback thread.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
SongLibing: removed one restart.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
SongLibing: Any reason to use search_pattern_in_file.inc instead of assert_grep.inc? I used assert_grep.inc, because I want the test stops immediately if any error happens.

SongLibing:
Hi @dr-m , The patch is updated, please have a look. @knielsen Marko suggest you to review the patch, please have a look.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
dr-m: Because there are multiple test steps that seem independent of each other, the rest of the test would be unnecessarily ruined because of the first failure. Replication tests are known to fail intermittently.

sql/handler.cc
```diff
@@ -2681,6 +2691,16 @@ static my_bool xarecover_complete_and_count(void *member_arg,
   return false;
 }
 
+static my_bool recover_complete_handlerton(THD *thd, plugin_ref plugin,
+                                void *arg)
```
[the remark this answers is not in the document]
SongLibing: fixed

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
SongLibing: got it. I used a separate log file to avoid what you said.

storage/innobase/trx/trx0roll.cc
SongLibing: renamed the function to innobase_tc_log_recovery_done(). and rename recover_complet() handler interface to signal_tc_log_recovery_done(). Because the function has to be called even binlog is off. background rollback thread has to be started while binlog is off.
HAVE_REPLICATION is for slave code, so it cannot be used here

SongLibing: renamed the function to innobase_tc_log_recovery_done(). and rename recover_complet() handler interface to signal_tc_log_recovery_done(). Because the function has to be called even binlog is off. background rollback thread has to be started while binlog is off.
HAVE_REPLICATION is for slave code, so it cannot be used here

storage/innobase/log/log0log.cc
```diff
@@ -1135,8 +1135,15 @@ ATTRIBUTE_COLD void logs_empty_and_mark_files_at_shutdown()
...
+	/* if trx_rollback_is_active is fase, rollback thread is
+	either not started yet or has rolled back all active
+	transactions. If that is not a normal shutdown(e.g.
+	error happened), it should not wait for all active
+	transactions to be rolled back.
+	*/
 	if (ulint total_trx = srv_was_started && !srv_read_only_mode
 	    && srv_force_recovery < SRV_FORCE_NO_TRX_UNDO
+	    && trx_rollback_is_active
 	    ? trx_sys.any_active_transactions() : 0) {
```
dr-m, line 1147: We already have the setting `innodb_fast_shutdown=3` that will cause the rollback of recovered transactions to be aborted when the shutdown is started. The design invariant of the shutdown loop is that there are no concurrent writes by other threads. If you want some additional special handling around this, I think that it would have to be done in `row_undo_step()` or at a higher level by assigning `srv_fast_shutdown=3`.

storage/innobase/srv/srv0start.cc
```diff
@@ -1860,9 +1855,6 @@ dberr_t srv_start(bool create_new_db)
 		    && !srv_read_only_mode) {
 			/* Drop partially created indexes. */
 			row_merge_drop_temp_indexes();
-			/* Rollback incomplete non-DDL transactions */
-			trx_rollback_is_active = true;
-			srv_thread_pool->submit_task(&rollback_all_recovered_task);
```
dr-m, line 1865: I suspect that this needs to be preserved here in the MariaDB Embedded Server (`libmariadbd`). We may be missing proper test coverage of that, because `./mtr --embedded` tests do not allow the server to be restarted. This would have to be enclosed in `#ifndef HAVE_REPLICATION`.

dr-m (changes requested):
Thank you, this is definitely moving to the right direction, but some details still need to be sorted out. Did you file an MDEV ticket for this?

storage/innobase/log/log0log.cc
SongLibing: I handled this to fix the failure of tc_heuristic_recover.test. the test will abort the server before binlog recovery. At the time, rollback thread is not  started yet. the test will hang for ever, since there are some active transactions which will never be rolled back.

SongLibing: sorry, but I didn't get it. why to assigning srv_fast_shutdown=3 at row_undo_step()?

storage/innobase/srv/srv0start.cc
SongLibing: ok, I will keep the original code in embedded server.

storage/innobase/log/log0log.cc
dr-m: I did not know why this change was added. Now that you mention the test `main.tc_heuristic_recover` as the reason, I see that this test is using fault injection via `debug_dbug="+d,innobase_xa_fail"`. Could this code change be avoided if those parts of the test were changed to also set `innodb_fast_shutdown=3`?

SongLibing: the following test without innobase_xa_fail has same behavior. It is designed to stop the server after binlog recovery  if the server starts with --tc-heuristic-recover option.

SongLibing: I got your concern, checking trx_rollback_is_active here could cause the server shutdown without waiting for user sessions to finished the transaction.

sql/handler.h
```diff
@@ -1405,6 +1405,23 @@ struct handlerton
...
+     on binlog recovery process.
+   */
+   int  (*recover_rollback_by_xid)(const XID &xid);
+   /*
+     It is called after binlog recovery has done commit/rollback of
+     all transactions. It is used together with recover_rollback_by_xid()
+     together to rollback prepared transactions asynchronously.
+   */
+   void (*signal_tc_log_recovery_done)();
    /*
```
knielsen, line 1425: I think this comment could be clarified a bit.

As written, "rollback status should be persisted after this call", it sounds like recover_rollback_by_xid() should make the changes durable (ie. sync the InnoDB WAL). And it is not clear what the purpose of signal_tc_log_recovery_done() is.

But I understand from the patch that it is actually signal_tc_log_recovery_done() that is responsible for making durable the decision inside innodb to rollback all the transactions that recover_rollback_by_xid() was called on. As an optimization to not need an fsync() for every call of recover_rollback_by_xid().

It would be good to explain this in the comment. That recover_rollback_by_xid() marks that the transaction/XID should be rolled back, but does not need to persist this decision to disk yet, and does not need to complete the actual rollback immediately. And that the purpose of signal_tc_log_recovery_done() is to persist this all to disk(). This would help another storage engine author to know what to do to implement these.

knielsen (commented):
I took a look at the patch 510577461b4427, I only looked at the non-InnoDB parts in detail.

I think the patch looks solid, it does all the things already I would expect it to do. I had one suggestion to clarify the comments on the new handlerton calls introduced.

This part of the commit message I did not understand:

    "With this patch, background rollback thread will not exit unless ddl log
      recovery is finished. It guarantees that all recovered prepared transactions
      which reverted to active state will be rolled back by the background thread."

Maybe you can elaborate/explain how this patch relates to the ddl log recovery?

SongLibing:
> I took a look at the patch [5105774](https://github.com/MariaDB/server/commit/510577461b442730fedffc839049687093c3a9e8), I only looked at the non-InnoDB parts in detail.
> 
> I think the patch looks solid, it does all the things already I would expect it to do. I had one suggestion to clarify the comments on the new handlerton calls introduced.
> 
> This part of the commit message I did not understand:
> 
> ```
> "With this patch, background rollback thread will not exit unless ddl log
>   recovery is finished. It guarantees that all recovered prepared transactions
>   which reverted to active state will be rolled back by the background thread."
> ```
> 
> Maybe you can elaborate/explain how this patch relates to the ddl log recovery?

It is not related to ddl log in the latest patch, the comment message has been updated.

sql/handler.h
SongLibing: updated the comment. I found that persist the status change is not really necessary. it will be persisted by the following undo actions. If crash happens before it is persisted, it will be in prepared state and the binlog recover will rollback it again.

storage/innobase/log/log0log.cc
SongLibing: Hi marko, the patch is updated.

sql/handler.h
SongLibing: updated

dr-m (commented):
Thank you.  The InnoDB part looks almost OK to me. I reviewed the generated code when this was cherry-picked to 10.6 bb2e125d076d9475cbe04fc8c24fbaee39d76bda.

I hope that @knielsen can soon comment on the code outside InnoDB.

storage/innobase/trx/trx0roll.cc
```diff
@@ -776,9 +781,12 @@ void trx_rollback_recovered(bool all)
         before trx_sys.deregister_rw(trx). */
         trx_sys.deregister_rw(trx);
         trx_free_at_shutdown(trx);
-      }
-      else
+      } else {
```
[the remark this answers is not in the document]
SongLibing: fixed

storage/innobase/srv/srv0start.cc
```diff
@@ -1858,11 +1853,13 @@ dberr_t srv_start(bool create_new_db)
...
 		    && !srv_read_only_mode) {
-			/* Drop partially created indexes. */
-			row_merge_drop_temp_indexes();
+#ifdef EMBEDDED_LIBRARY
 			/* Rollback incomplete non-DDL transactions */
 			trx_rollback_is_active = true;
 			srv_thread_pool->submit_task(&rollback_all_recovered_task);
+#endif
+			/* Drop partially created indexes. */
+			row_merge_drop_temp_indexes();
```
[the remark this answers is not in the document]
SongLibing: reverted the change.

storage/innobase/include/trx0trx.h
```diff
@@ -715,9 +715,17 @@ struct trx_t : ilist_node<>
...
+  Recovered XA(2PC) followed by rollback_by_xid() :
+  * NOT_STARTED -> PREPARED -> ACTIVE -> COMMITTED -> (freed)
+
+  Recovered XA(2PC) followed by recover_rollback_by_xid() :
+  recover_rollback_by_xid() just sets the state back to ACTIVE,
+  background rollback thread will do the following actions to finish the trx.
+  * NOT_STARTED -> PREPARED -> ACTIVE  -> COMMITTED -> (freed)
+
   Recovered XA followed by XA ROLLBACK:
   * NOT_STARTED -> PREPARED -> ACTIVE -> COMMITTED -> (freed)
```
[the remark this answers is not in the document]
SongLibing: updated.

storage/innobase/log/log0log.cc
```diff
@@ -1135,8 +1135,13 @@ ATTRIBUTE_COLD void logs_empty_and_mark_files_at_shutdown()
 	shutdown, because the InnoDB layer may have committed or
 	prepared transactions and we don't want to lose them. */
 
+	/* it is an abnormal shutdown(e.g.error happened). If trx_rollback_is_active
+  is fase, and there is any recovered active transaction. Thus it should not
+  wait for all active transactions to be rolled back.
+	*/
 	if (ulint total_trx = srv_was_started && !srv_read_only_mode
 	    && srv_force_recovery < SRV_FORCE_NO_TRX_UNDO
+	    && (trx_rollback_is_active || !trx_sys.any_active_transaction_recovered())
```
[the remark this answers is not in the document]
SongLibing: fixed

storage/innobase/include/trx0sys.h
```diff
@@ -1075,6 +1088,8 @@ class trx_sys_t
   /** @return total number of active (non-prepared) transactions */
   size_t any_active_transactions(size_t *prepared= nullptr);
 
+  /** @return true if any active (non-prepared) transactions is recovered */
+  bool any_active_transaction_recovered();
```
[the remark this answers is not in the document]
SongLibing: fixed

storage/innobase/trx/trx0sys.cc
```diff
@@ -362,3 +362,16 @@ size_t trx_sys_t::any_active_transactions(size_t *prepared)
...
+  bool has_recovered = false;
+
+  trx_sys.trx_list.find_first([&](const trx_t &trx) {
+    if (trx.state == TRX_STATE_ACTIVE && trx.is_recovered) {
+      has_recovered = true;
+    }
+    return has_recovered;
+  });
+  return has_recovered;
+}
```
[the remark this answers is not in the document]
SongLibing: trx_disconnect_prepared() is only for prepared transactions, for an active transaction, is_recovered is never changed. So the mutex is not really necessary.

updated the code to, I would make it a inline function.
#ifndef EMBEDDED_LIBRARY
bool trx_sys_t::any_active_transaction_recovered()
{
  return trx_sys.trx_list.find_first([&](const trx_t &trx) {
    return trx.state == TRX_STATE_ACTIVE && trx.is_recovered;
  });
}
#endif

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
```diff
@@ -0,0 +1,93 @@
...
+--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_text= Server reported: Rolled back recovered transaction
+--let $assert_select= Rolled back recovered transaction
+--let $assert_count= 1
+--source include/assert_grep.inc
+
+# The transaction was set to ACTIVE in previous recovery, thus it should not
+# find any prepared thransactions.
+--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_text= Server NOT report: Found 1 prepared transaction
```
andrelkin, line 82: In 'Server NOT report', is 'NOT' a typo, right? Maybe 'NOW' meant to be...

sql/handler.h
```diff
@@ -1405,6 +1405,21 @@ struct handlerton
...
    int  (*rollback_by_xid)(handlerton *hton, XID *xid);
+   /*
+     recover_rollback_by_xid is optional. If set, it will be called instead
+     of rollback_by_xid when transactions should be rollback at server startup.
+
+     This function should just changed the transaction's state from prepared to
+     active, and rollback the transaction asynchronously by a background
+     thread. Thus rolling back a long transaction will not block server startup.
+   */
+   int  (*recover_rollback_by_xid)(const XID &xid);
```
andrelkin, line 1416: I think the argument should be a pointer (also as elsewhere). After all an implementation's method can not understand C++ references.

andrelkin (approved):
Very good work LiBing!

I have two points to raise.

sql/handler.h
SongLibing: sure, updated

sql/handler.cc
```diff
@@ -2793,7 +2818,12 @@ static my_bool xarecover_handlerton(THD *unused, plugin_ref plugin,
         }
         else if (!info->mem_root)
         {
-          int rc= hton->rollback_by_xid(hton, info->list+i);
+          int rc= 0;
+          if (hton->recover_rollback_by_xid &&
+              !IF_WSREP(wsrep_is_wsrep_xid(info->list + i), true))
```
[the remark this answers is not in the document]
SongLibing: right

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
SongLibing: it is NOT. since the transactions has been set to ACTIVE in the first recovery. So it should not find any prepared transaction in the second recovery.

sql/handler.cc
```diff
@@ -2793,7 +2818,12 @@ static my_bool xarecover_handlerton(THD *unused, plugin_ref plugin,
         }
         else if (!info->mem_root)
```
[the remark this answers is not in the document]
SongLibing: after check the logic, I found it is called only when --tc-heuristic-recover is set.  So I think it is better not to change the behavior. So I just reverted this change.

andrelkin:
Looks good. Thanks LiBing!

andrelkin:
Sorry, in the last comment I meant a mere approval.

storage/innobase/include/trx0sys.h
```diff
@@ -418,6 +420,7 @@ class rw_trx_hash_t
             (trx_state_eq(trx, TRX_STATE_ACTIVE) &&
              (!srv_was_started ||
               srv_read_only_mode ||
+              !trx_rollback_is_active ||
```
dr-m, line 423: It seems that we fail to invoke `rollback_all_recovered_task.wait()` in `logs_empty_and_mark_files_at_shutdown()` or `srv_shutdown_threads()`. It seems that this added condition might introduce a resource leak if the rollback is still running during shutdown. I would suggest to replace the following code in `logs_empty_and_mark_files_at_shutdown()`:
```c++
	/* We need these threads to stop early in shutdown. */
	const char* thread_name = srv_fast_shutdown != 2
		&& trx_rollback_is_active
		? "rollback of recovered transactions" : nullptr;
```
with something like this (note: I hope that it will not hang if there was no prior call to `srv_thread_pool->submit_task(&rollback_all_recovered_task)` in case startup was aborted due to some error):
```c++
	if (srv_fast_shutdown != 2) {
		rollback_all_recovered_task.wait();
		ut_ad(!trx_rollback_is_active);
	}
```
This logic will have to be tested with various values of `innodb_force_recovery` and `innodb_fast_shutdown`.

I would also suggest to minimize the references to `trx_rollback_is_active` in other code. Ideally we would only have a few `ut_ad(!trx_rollback_is_active)` where appropriate.

dr-m (commented):
Thank you, the InnoDB part of this looks mostly OK now. I think that we should try to clean up the shutdown logic a bit more, to avoid any potential race condition.

I did not check the changes outside InnoDB, and I did not compile the code or run the test case.

configure.cmake
```diff
@@ -23,6 +23,12 @@ INCLUDE (CheckCCompilerFlag)
 INCLUDE (CheckCSourceRuns)
 INCLUDE (CheckSymbolExists)
 
+OPTION(WITH_ASYNC_BINLOG_ROLLBACK
+  "Rollback prepared transactions asynchronously at binlog recovery" OFF)
+IF(WITH_ASYNC_BINLOG_ROLLBACK)
+  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DASYNC_BINLOG_ROLLBACK")
```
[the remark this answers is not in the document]
SongLibing: Removed the cmake code after moving to 11.6

dr-m (commented):
Unfortunately, the test `mariabackup.slave_provision_nolock` seems to fail quite often due to a server shutdown hang that could be caused by these changes. Please investigate and fix.

sql/handler.cc
```diff
@@ -2646,8 +2646,18 @@ static void xarecover_do_commit_or_rollback(handlerton *hton,
...
+  {
+    rc= hton->commit_by_xid(hton, &x);
+  }
+  else
+  {
+    if (hton->recover_rollback_by_xid)
+      rc= hton->recover_rollback_by_xid(&x);
+    else
+      rc= hton->rollback_by_xid(hton, &x);
+  }
```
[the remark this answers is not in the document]
SongLibing: removed the {}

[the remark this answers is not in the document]
SongLibing: removed the {}

sql/handler.cc
```diff
@@ -2714,6 +2724,21 @@ static my_bool xarecover_complete_and_count(void *member_arg,
...
+
+  if (hton->signal_tc_log_recovery_done)
+    hton->signal_tc_log_recovery_done();
+  return false;
+}
+
+void ha_signal_tc_log_recovery_done()
+{
+  (void) plugin_foreach(NULL, tc_log_recover_done_handlerton,
+                        MYSQL_STORAGE_ENGINE_PLUGIN, 0);
```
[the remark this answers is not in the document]
SongLibing: fixed

sql/handler.cc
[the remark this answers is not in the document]
SongLibing: removed the {}

sql/handler.cc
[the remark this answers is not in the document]
SongLibing: fixed

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17116,6 +17121,74 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+
+  if (trx->rsegs.m_redo.undo) {
+    ut_ad(trx->rsegs.m_redo.undo->rseg == trx->rsegs.m_redo.rseg);
+
+    mtr_t mtr;
+    mtr.start();
+    trx_undo_set_state_at_prepare(trx, trx->rsegs.m_redo.undo, true, &mtr);
+    mtr.commit();
+
+    ut_ad(mtr.commit_lsn() > 0);
```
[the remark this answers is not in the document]
SongLibing: done

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17116,6 +17121,74 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+  /* Change the trx state to TRX_STATE_ACTIVE. Background rollback thread
+     will rollback this trx later. */
+  trx->state= TRX_STATE_ACTIVE;
+  return 0;
+}
+
+void innobase_tc_log_recovery_done()
+{
+  if (srv_force_recovery < SRV_FORCE_NO_TRX_UNDO
+      && !srv_read_only_mode) {
```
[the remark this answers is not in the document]
SongLibing: log_buffer_flush_to_disk(true) should be called  if srv_read_only_mode is false. otherwise it will cause an assertion failure.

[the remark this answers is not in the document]
dr-m: You are right, it should be more like this:
```c++
  if (srv_read_only_mode)
    return;
  /* Make durable any innobase_recover_rollback_by_xid(). */
  log_buffer_flush_to_disk(true);
  if (srv_force_recovery < SRV_FORCE_NO_TRX_UNDO)
  {
```
Even with `innodb_force_recovery=3` or `innodb_force_recovery=4` we expect that some `XA PREPARE` transactions may be changed back to "active" state, to note that they must be rolled back. We may choose not to complete the rollback during this server run, but a subsequent server invocation with a smaller value of `innodb_force_recovery` could perform that.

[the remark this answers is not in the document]
dr-m: Let’s revise the first condition to
```c++
  if (srv_force_recovery >= SRV_FORCE_NO_UNDO_LOG_SCAN || srv_read_only_mode)
    return;
```
in order to avoid any surprises on `innodb_force_recovery=5` or `innodb_force_recovery=6`. In these cases, no `XA PREPARE` transactions can be found, so no changes will need to be made durable either.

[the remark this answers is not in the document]
SongLibing: fixed

dr-m:
I see that this is rebased on top of the currently latest 11.6. There are some failures that I do not observe for the 11.6 branch. I can also reproduce a hang myself by executing the following:
```bash
./mtr --parallel=auto --repeat=10 mariabackup.slave_provision_nolock{,,,,,,,,,,,,,,,,}
```
I checked the stack traces of one hung `mariadbd` process, and I did not see anything obvious inside InnoDB. Outside InnoDB, I am not that familiar with the code. To debug this, I suggest that you use `./mtr --rr` and then `killall -ABRT mariadbd` to get an `rr replay` trace of a hung `mariadbd` process. With that trace, it should be possible to understand what exact sequence of events is leading to the hang.

storage/innobase/srv/srv0start.cc
```diff
@@ -1988,6 +1985,19 @@ void innodb_preshutdown()
 
   if (srv_read_only_mode)
     return;
+
+  #ifndef EMBEDDED_LIBRARY
+  if (srv_force_recovery < SRV_FORCE_NO_TRX_UNDO && srv_was_started &
+      !trx_rollback_is_active && trx_sys.any_active_transaction_recovered()) {
```
dr-m, line 1991: It looks like the bitwise arithmetic `&` should be replaced with the logical short-circuit `&&`.
(Also, the `{` needs to be on its own line.)

storage/innobase/srv/srv0start.cc
```diff
@@ -1988,6 +1985,19 @@ void innodb_preshutdown()
...
+      /* Trx rollback thread may not be started yet, since error happened before
+         innobase_tc_log_recovery_done(), thus it is started here to finish the
+         normal shutdown process.
+      */
+      trx_rollback_is_active = true;
```
dr-m, line 1996: According to [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) there should be no space before the `=`.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
```diff
@@ -0,0 +1,93 @@
...
+
+CREATE TABLE t1 (id int primary key, data int) ENGINE = InnoDB;
+INSERT INTO t1 VALUES (0, 1);
+
+--echo #
+--echo # 1. Check DML in prepared state can rollback correctly.
+--echo #
+
+--echo [ connection con1 ]
+--connect(con1, localhost, root,,)
```
dr-m, line 21: An `echo` before any `connect` or `connection` is redundant and should be removed. The connections are automatically identified in the `.result` files.

dr-m (approved):
I see that you were able to fix the occasional hangs during the test `mariabackup.slave_provision_nolock`.

The InnoDB changes are OK after my remaining comments have been addressed. I did not review the added test file in detail; I trust that @knielsen will do that.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17116,6 +17121,81 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+  /*
+    trx_get_trx_by_xid() sets trx's xid to null. Thus only one call for any
+    given XID can find the transaction. Subsequent calls by other threads
+    would return nullptr. That is what guarantees that no other thread can be
+    modifying the state of the transaction at this point.
+  */
+  trx_t *trx= trx_get_trx_by_xid(xid);
+  if (trx == NULL) {
+    return(XAER_RMFAIL);
+  }
```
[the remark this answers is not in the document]
SongLibing: fixed

storage/innobase/srv/srv0start.cc
SongLibing: fixed

storage/innobase/include/trx0roll.h
```diff
@@ -31,6 +31,8 @@ Created 3/26/1996 Heikki Tuuri
 #include "mtr0mtr.h"
 #include "trx0sys.h"
 
+extern tpool::task_group rollback_all_recovered_group;
+extern tpool::waitable_task rollback_all_recovered_task;
 extern bool		trx_rollback_is_active;
```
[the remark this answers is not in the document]
SongLibing: fixed

storage/innobase/srv/srv0start.cc
```diff
@@ -1988,6 +1985,22 @@ void innodb_preshutdown()
...
+#ifndef EMBEDDED_LIBRARY
+  if (srv_force_recovery < SRV_FORCE_NO_TRX_UNDO && srv_was_started &&
+      !trx_rollback_is_active && trx_sys.any_active_transaction_recovered())
+  {
+    /* Trx rollback thread may not be started yet, since error happened before
+       innobase_tc_log_recovery_done(), thus it is started here to finish the
+       normal shutdown process.
+     */
+    ib::warn() << "Start rollback task at preshutdown to rollback active "
+               << "transactions recovered.";
```
dr-m, line 1998: If `innodb_fast_shutdown>1` we do not want to roll back any transactions. (Sorry, I did not realize this in my review yesterday.) We are missing a condition `srv_fast_shutdown <= 1 &&` here.

The use of `ib::logger` should be limited to those cases that require some special formatting, such as pretty-printing records. Here, we are unnecessarily invoking `operator<<` twice. It would be better to invoke the underlying function directly. I think that this should be `sql_print_information`, not `sql_print_warning`:
```c++
    sql_print_information("InnoDB: Rolling back recovered transactions at shutdown");
```
Even better, we could refactor the inner part of `trx_roll_report_progress()` to a new function (which would return `bool` indicating whether there were any transactions, and invoke it here:
```c++
  if (srv_force_recovery < SRV_FORCE_NO_TRX_UNDO && srv_was_started &&
      srv_fast_shutdown <= 1 && !trx_rollback_is_active &&
      trx_sys.report_to_rollback())
  {
    trx_rollback_is_active= true;
    srv_thread_pool->submit_task(&rollback_all_recovered_task);
  }
```
That common ‘inner’ function `trx_sys_t::report_to_rollback()` would count transactions and rows:
```c++
  // … declare arg and iterate over `rw_trx_hash`
  if (!arg.n_trx)
    return false;
  service_manager_extend_timeout(INNODB_EXTEND_TIMEOUT_INTERVAL,
		 		 "To roll back: " UINT32PF " transactions, "
				 UINT64PF " rows", arg.n_trx, arg.n_rows);
  sql_print_information("InnoDB: To roll back: " UINT32PF " transactions, "
                        UINT64PF " rows", arg.n_trx, arg.n_rows);
  return true;
```
In this way, we would have a consistent type of output.

dr-m (approved):
Thank you for addressing my formatting comments. One more improvement occurred to me.

storage/innobase/srv/srv0start.cc
SongLibing: Hi Marko,  disable_log_notes is set to 1 at unireg_abort(), information logs will be dropped. That is why I use warning here. The warning happens only when error happens. I run the whole test. It doesn't break other tests.

SongLibing: > If innodb_fast_shutdown>1 we do not want to roll back any transactions. (Sorry, I did not realize this in my review yesterday.) We are missing a condition srv_fast_shutdown <= 1 && here.

Looks the code in `logs_empty_and_mark_files_at_shutdown()`, it will wait for all active transactions to either rollback or deregistered.  It seams that is not expected according to the comment of srv_fast_shutdown. 
If you want it to be fixed in this patch. I think it should be fixed in trx_rollback_recovered() like:
```
--- a/storage/innobase/trx/trx0roll.cc
+++ b/storage/innobase/trx/trx0roll.cc
@@ -749,8 +749,8 @@ void trx_rollback_recovered(bool all)
     ut_ad(trx_state_eq(trx, TRX_STATE_ACTIVE));
     ut_d(trx->mutex_unlock());
 
-    if (srv_shutdown_state != SRV_SHUTDOWN_NONE && !srv_undo_sources &&
-        srv_fast_shutdown)
+    if (srv_shutdown_state != SRV_SHUTDOWN_NONE &&
+        (srv_fast_shutdown > 1 || !srv_undo_sources))
       goto discard;
```

dr-m: Thank you. I did not know about `disable_log_notes`. The pre-shutdown was written by @vaintroub in 5e62b6a5e06eb02cbde1e34e95e26f42d87fce02. It feels wrong to have all informational messages suppressed during shutdown. Occasionally there are shutdown hangs, and it would be useful to know based on logs, at which exact point the shutdown is hanging.

Regarding `srv_fast_shutdown`, I meant that there is no point in scheduling a task if we do not expect it to do anything useful. The purpose of `innodb_fast_shutdown=2` is to execute a crash-like shutdown (not writing out dirty pages and updating the log checkpoint). The purpose of `innodb_fast_shutdown=3` is to execute a clean shutdown, but without rolling back any recovered transactions. That condition is being checked at a deeper level, in `row_undo_step()`:
```c++
	if (UNIV_UNLIKELY(!trx->dict_operation
			  && !srv_undo_sources
			  && srv_shutdown_state != SRV_SHUTDOWN_NONE)
	    && (srv_fast_shutdown == 3 || trx == trx_roll_crash_recv_trx)) {
		/* Shutdown has been initiated. */
		trx->error_state = DB_INTERRUPTED;
		return NULL;
	}
```
I just realized that for recovered transactions (`trx == trx_roll_crash_recv_trx`) this does not even depend on `innodb_fast_shutdown=3`. We will always interrupt the rollback of any recovered non-DDL transactions on shutdown. If `innodb_fast_shutdown=3`, we will additionally interrupt the rollback of any active transaction. Typically, shutdown would cause a rollback of all currently active transactions, and normally shutdown would wait until they have been rolled back.

I suspect that sometimes, small transactions may be rolled back nevertheless, because the `srv_shutdown_state` would be changed later. This should be tested with a larger transaction, such as a `DELETE` or `UPDATE` of a large number of rows.

SongLibing: You are right, so we don't need to change the code in trx_rollback_recovered, but we still need to start the rollback task.

mysql-test/suite/binlog/r/binlog_recover_async_rollback_trx.result
```diff
@@ -0,0 +1,64 @@
...
+#    should be started at preshutdown accordingly to rollback or
+#    deregister all recovered active transactions.
+BEGIN;
+INSERT INTO t1 SELECT id + power(2, 2), 1 FROM t1;
+INSERT INTO t1 SELECT id + power(2, 3), 1 FROM t1;
+INSERT INTO t1 SELECT id + power(2, 4), 1 FROM t1;
+INSERT INTO t1 SELECT id + power(2, 5), 1 FROM t1;
+INSERT INTO t1 SELECT id + power(2, 6), 1 FROM t1;
+INSERT INTO t1 SELECT id + power(2, 7), 1 FROM t1;
+COMMIT;
```
dr-m, line 56: For the `SELECT` part of this, explicit record locks will unnecessarily be allocated until [MDEV-24813](https://jira.mariadb.org/browse/MDEV-24813) has been fixed. See 516f7c111b0cc3e9ad193c5a4e8c9171a081a362 and e98a5e166c16f24fef2d4f356b0d82559a835ee4 how to use the sequence engine to work around this deficiency.

storage/innobase/srv/srv0start.cc
```diff
@@ -1988,6 +1985,22 @@ void innodb_preshutdown()
...
+    ib::warn() << "Start rollback task at preshutdown to rollback active "
+                  "transactions recovered.";
+    trx_rollback_is_active= true;
+    srv_thread_pool->submit_task(&rollback_all_recovered_task);
+  }
+#endif
+
   if (!srv_fast_shutdown && srv_operation <= SRV_OPERATION_EXPORT_RESTORED)
     if (srv_force_recovery < SRV_FORCE_NO_TRX_UNDO && srv_was_started)
       while (trx_sys.any_active_transactions())
```
dr-m, line 2006: Should we invoke the above only if the below condition (shutdown with `innodb_fast_shutdown=0`) holds? In that case, would `sql_print_information()` messages be visible?

@vaintroub Is it intentional that `sql_print_information()` messages would be suppressed during a normal shutdown? I think that the message about starting rollback (if it is needed at all) should be informational and not a warning, but apparently the message would be suppressed during a normal shutdown.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
```diff
@@ -0,0 +1,149 @@
...
+
+# Rollback task will not be started at preshutdown if recovery mode is greater
+# to 2.  Active transactions are not expected to rollback.
+--error 1
+--exec $MYSQLD_LAST_CMD --tc-heuristic-recover=ROLLBACK --innodb-force-recovery=3 --log-error=$MYSQLTEST_VARDIR/tmp/log.err
+
+# Rollback task will be started at preshutdown of fast shutdown if force
+# recovery is 2. But the transaction is deregistered instead of rollback.
+--error 1
+--exec $MYSQLD_LAST_CMD --tc-heuristic-recover=ROLLBACK --innodb-fast-shutdown=1 --innodb-force-recovery=2 --log-error=$MYSQLTEST_VARDIR/tmp/log1.err
```
dr-m, line 124: Why would it matter whether the transaction is fully rolled back? What should matter is that the transaction is no longer in `XA PREPARE` state and it is not causing blocking new transactions. (There could be a locking conflict until the transaction has been rolled back, but that conflict should be resolved automatically, unless `innodb_lock_wait_timeout` expires before the rollback completes.)

I think that the only way to observe the effect of the rollback would be to read data in a `READ UNCOMMITTED` transaction. Only with `innodb_fast_shutdown=0` there is a guarantee that all undo logs and transaction history will be cleared.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
```diff
@@ -0,0 +1,149 @@
...
+--error 1
+--exec $MYSQLD_LAST_CMD --tc-heuristic-recover=ROLLBACK --innodb-fast-shutdown=1 --log-error=$MYSQLTEST_VARDIR/tmp/log1.err
+
+--source include/start_mysqld.inc
+
+--let $assert_file= $MYSQLTEST_VARDIR/tmp/log.err
+--let $assert_text= Server NOT report: Start rollback task at preshutdown
+--let $assert_select= Start rollback task at preshutdown
+--let $assert_count= 0
+--source include/assert_grep.inc
```
dr-m, line 137: I would find the following more concise and understandable:
```sql
--let SEARCH_FILE=$MYSQLTEST_VARDIR/log/mysqld.1.err
--let SEARCH_PATTERN=InnoDB: Start rollback task at preshutdown
--source include/shutdown_mysqld.inc
```
However, even better than checking the presence of such a message which does not seem to add much value to a normal user, I think that it could be better to start the server in `innodb_read_only=1` mode and check that the data is still there. You can find some ideas in `mysql-test/suite/innodb/t/innodb_force_recovery_rollback.test` and `mysql-test/suite/innodb/t/read_only_recover_committed.test`.

storage/innobase/srv/srv0start.cc
vaintroub: If there is anything I suppressed in pre-shutdown, it would be by oversight. I certainly do not remember doing anything with messages. The reason it exist is to run purge to the end, in slow shutdown. purge might need some audit plugin, if I remember correctly, so there needs to be a time during shutdown, when users are gone, but plugins are still intact. 

Before me, (and before tpool) the sequencing was done by some innodb_thd_proxy.

storage/innobase/srv/srv0start.cc
SongLibing: it is disabled by `disable_log_notes= 1;` in function unireg_abort() which is a server layer code, it is called only if error happens at startup.

dr-m: That turns out to be an intentional change that was made in affff1aefc81ea0a3cb778a6617504ccd1e905fc. Maybe it was not realized back then that `unireg_abort()` is being invoked as part of a normal shutdown,with no prior errors. In such a case, it would seem to be helpful to not filter out any `sql_print_information()` messages.

In any case, we may ‘sidestep’ this by not adding a message at all, and by starting the `rollback_all_or_recovered_task` right above the `while` loop in the following code snippet.

mysql-test/suite/binlog/r/binlog_recover_async_rollback_trx.result
SongLibing: fine, use sequence instead.

storage/innobase/srv/srv0start.cc
SongLibing: removed the warning.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
SongLibing: The ROLLBACK option just helps to abort the server in this test, there is no PREPARED transaction in this test.  it doesn't rollback any active transactions.

storage/innobase/srv/srv0start.cc
SongLibing: > Should we invoke the above only if the below condition (shutdown with innodb_fast_shutdown=0) holds?  we have to start the rollback task if innodb_fast_shutdown is enabled. because logs_empty_and_mark_files_at_shutdown() will wait even innodb_fast_shutdown is enabled. It depends on rollback thread to deregister all active transactions in case innodb_fast_shutdown is enabled.

mysql-test/suite/binlog/t/binlog_recover_async_rollback_trx.test
SongLibing: yes, removed the warning and the assertions.

storage/innobase/srv/srv0start.cc
```diff
@@ -1988,6 +1985,20 @@ void innodb_preshutdown()
...
+#ifndef EMBEDDED_LIBRARY
+  if (srv_force_recovery < SRV_FORCE_NO_TRX_UNDO && srv_was_started &&
+      !trx_rollback_is_active && trx_sys.any_active_transaction_recovered())
+  {
+    /* Trx rollback thread may not be started yet, since error happened before
+       innobase_tc_log_recovery_done(), thus it is started here to finish the
+       normal shutdown process.
+     */
+    trx_rollback_is_active= true;
+    srv_thread_pool->submit_task(&rollback_all_recovered_task);
```
dr-m, line 1998: Like you pointed out, we do need this for any value of `innodb_fast_shutdown`, because the recovered transactions must be removed from `trx_sys` by the rollback task, even if the rollback itself is aborted. But a message is not needed here. Thank you for removing it.

dr-m (approved):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17116,6 +17121,81 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+    mtr.start();
+    trx_undo_set_state_at_prepare(trx, trx->rsegs.m_redo.undo, true, &mtr);
+    mtr.commit();
+
+    ut_ad(mtr.commit_lsn() > 0);
+  }
+
+  /* Change the trx state to TRX_STATE_ACTIVE. Background rollback thread
+     will rollback this trx later. */
+  trx->state= TRX_STATE_ACTIVE;
```
dr-m, line 17177: In InnoDB, we usually don’t indent the second line of a comment deeper than the first line.

We could mention the actual function name here and omit the kind-of redundant first part:
```c++
  /* trx_rollback_recovered() will roll back this later. */
  trx->state= TRX_STATE_ACTIVE;
```
We could also remove the previous comment about durability and do it here:
```c++
  /* The above state change from XA PREPARE will be made durable in
  innobase_tc_log_recovery_done(), which will also initiate
  trx_rollback_recovered() to roll back this transaction. */
  trx->state= TRX_STATE_ACTIVE;
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17116,6 +17121,81 @@ int innobase_rollback_by_xid(handlerton* hton, XID* xid)
...
+}
+
+void innobase_tc_log_recovery_done()
+{
+  /* Make durable any innobase_recover_rollback_by_xid(). */
+  if (high_level_read_only)
+    return;
+
+  /* Make durable any innobase_recover_rollback_by_xid(). */
+  log_buffer_flush_to_disk(true);
```
dr-m, line 17188: The duplicated first comment should be removed.

dr-m (approved):
I had one final look on the InnoDB part. I expect @knielsen to run the test case in a debugger as part of his review.

storage/innobase/handler/ha_innodb.cc
SongLibing: fixed

storage/innobase/handler/ha_innodb.cc
SongLibing: fixed

knielsen:
> I see that you were able to fix the occasional hangs during the test mariabackup.slave_provision_nolock.

What was the problem, and how was it fixed? Was it a bug in the patch that was fixed, or just an unrelated issue picked up during a rebase?

It's hard to follow what happened since my first review since there are no new commits, just force-pushes (but I will try...)

SongLibing:
> > I see that you were able to fix the occasional hangs during the test mariabackup.slave_provision_nolock.

the test case has a case to start mysqld with --tc-heuristic-recover=ROLLBACK, that is a case the patch didn't handle correctly.  This patch just postpone the start of rollback thread, it is possible that error happens before rollback thread is started. innodb cleanup suppose rollback thread is there, but that is not true for above case. It cause a failure.

sql/handler.h
```diff
@@ -1424,6 +1424,21 @@ struct handlerton
    int  (*recover)(handlerton *hton, XID *xid_list, uint len);
    int  (*commit_by_xid)(handlerton *hton, XID *xid);
    int  (*rollback_by_xid)(handlerton *hton, XID *xid);
+   /*
+     recover_rollback_by_xid is optional. If set, it will be called instead
+     of rollback_by_xid when transactions should be rollback at server startup.
```
knielsen, line 1429: Grammar: "when transactions should be rolled back at server startup"

sql/handler.h
```diff
@@ -1424,6 +1424,21 @@ struct handlerton
...
+
+     This function should just changed the transaction's state from prepared to
+     active, and rollback the transaction asynchronously by a background
+     thread. Thus rolling back a long transaction will not block server startup.
```
knielsen, line 1433: Suggestion for better wording:

"This function should just change the transaction's state from prepared to active before returning. The actual rollback should then happen asynchroneously (eg. in a background thread). This way, rollbacks that take a long time to complete will not block server startup, and the database becomes available sooner to serve user queries."

knielsen (commented):
Thanks Libing Song for your explanation about  --tc-heuristic-recover=ROLLBACK, that makes sense.

Again, I only looked at the (small) part of the test that relates to replication/handler.cc server code. Good catch with the --tc-heuristic-recover issue, otherwise not much changed since my first review and I still think this part of the patch looks solid and the way I'd expect.

I added a couple very minor English grammer suggestions that you can consider.

 \- Kristian.

sql/handler.h
SongLibing: Hi Kristian, thanks for your suggestions. Fixed them all.

SongLibing:
@dr-m I rebased the PR to main.

dr-m:
@SongLibing Thank you. I had updated the branch bb-11.6-MDEV-33853 earlier today in order to facilitate some testing. The two branches only differ by the current `main` head db5d1cde4505fdd04bdb3389b28da004ca8ec579.

PR #3031, opened 2024-01-30

mariadb-DebarunBanerjee:

## Description
If we fail to open a tablespace while looking for FILE_CHECKPOINT, we set the corruption flag. Specifically, if encryption key is missing, we would not be able to open an encrypted tablespace and the flag could be set. We miss checking for this flag and report "Missing FILE_CHECKPOINT".

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/include/no_checkpoint_start.inc
+++ b/mysql-test/suite/innodb/include/no_checkpoint_start.inc
@@ -1,5 +1,28 @@
 # Preparation for using no_checkpoint_end.inc
 
+# no_checkpoint_flush: Set to trigger flushing the dirty pages from buffer pool
+# and checkpoint before the "no checkpoint" block.
+
+if ($no_checkpoint_flush) {
+  --echo
+  --echo # Flush all dirty pages from buffer pool
+  SET @no_checkpoint_save_pct= @@GLOBAL.innodb_max_dirty_pages_pct;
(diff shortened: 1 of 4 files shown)


dr-m (approved):

mysql-test/suite/innodb/include/no_checkpoint_start.inc
```diff
@@ -1,5 +1,28 @@
...
+  SET GLOBAL innodb_max_dirty_pages_pct=0.0;
+
+  let $wait_condition =
+  SELECT variable_value = 0
+  FROM information_schema.global_status
+  WHERE variable_name = 'INNODB_BUFFER_POOL_PAGES_DIRTY';
+  --source include/wait_condition.inc
+
+  SET GLOBAL innodb_max_dirty_pages_pct= @no_checkpoint_save_pct;
+  SET GLOBAL innodb_max_dirty_pages_pct_lwm= @no_checkpoint_save_pct_lwm;
```
dr-m, line 22: I was going to point out that we do not need this, because the server will be killed by `no_checkpoint_end.inc`. But then I realized that we probably want to restore these variables to something nonzero, to avoid eager flushing of the buffer pool during the critical section of the test (where we do not want any checkpoints to occur).

mariadb-DebarunBanerjee: right, also I think it is good practice to reset things back once used.

PR #3032, opened 2024-01-30

mariadb-DebarunBanerjee:

## Description
1. If we fail to open a tablespace while looking for FILE_CHECKPOINT, we set the corruption flag. Specifically, if encryption key is missing, we would not be able to open an encrypted tablespace and the flag could be set. We miss checking for this flag and report "Missing FILE_CHECKPOINT"

2. Based on the dynamic flushes and checkpoint we could return error for any one of t1-t5 encrypted table. The test needs to be corrected.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/encryption/r/innodb-redo-nokeys.result
+++ b/mysql-test/suite/encryption/r/innodb-redo-nokeys.result
@@ -21,6 +21,15 @@ insert into t2 select * from t1;
 insert into t3 select * from t1;
 insert into t4 select * from t1;
 commit;
+
+# Flush all dirty pages from buffer pool
+SET @no_checkpoint_save_pct= @@GLOBAL.innodb_max_dirty_pages_pct;
+SET @no_checkpoint_save_pct_lwm= @@GLOBAL.innodb_max_dirty_pages_pct_lwm;
+SET GLOBAL innodb_max_dirty_pages_pct_lwm=0.0;
+SET GLOBAL innodb_max_dirty_pages_pct=0.0;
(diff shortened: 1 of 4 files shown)


dr-m (changes requested):
https://buildbot.mariadb.org/#/builders/369/builds/17391/steps/7/logs/stdio shows a similar result as the one that I got locally when testing 8dc3a0556c1ff7596eef8d42039d00d10da6a44b:
```sh
./mtr --parallel=auto encryption.innodb-redo-nokeys
```
```
encryption.innodb-redo-nokeys 'cbc,innodb' w1 [ skipped ]  Extra checkpoint 1 after 54540 (0:8=54580,0:7=54540)
encryption.innodb-redo-nokeys 'ctr,innodb' w2 [ skipped ]  Extra checkpoint 1 after 54540 (0:8=54580,0:7=54540)
```
The 10.5 version of the test is passing for me.

Since this test is not specific to debug builds, instead of `SET GLOBAL innodb_buf_flush_list_now = 1;` we may use something similar to what the non-debug tests `innodb.page_cleaner` or `innodb.doublewrite` do to perform a checkpoint. If possible, we should try to avoid adding new server restarts.

This breakage is not caused by your changes, but I think that we need to fix it as part of claiming to fix this test. A frequently skipped test is not useful.

dr-m (approved):

dr-m:
Closed by 466069b184f318aafc0e87cc6b06633d1fa9eb4b.

PR #3033, opened 2024-01-30

mariadb-DebarunBanerjee:

## Description
1. If we fail to open a tablespace while looking for FILE_CHECKPOINT, we set the corruption flag. Specifically, if encryption key is missing, we would not be able to open an encrypted tablespace and the flag could be set. We miss checking for this flag and report "Missing FILE_CHECKPOINT"
  [ERROR] InnoDB: Missing FILE_CHECKPOINT(xxx) at xxx
  [ERROR] InnoDB: Log scan aborted at LSN xxx

2. Based on the dynamic flushes and checkpoint we could return error for any one of t1-t5 encrypted table. The test needs to be corrected.
  NOT FOUND[ERROR] InnoDB: Encryption key is not found for .*test.t1.ibd

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -4120,9 +4120,10 @@ static bool recv_scan_log(bool last_phase)
 
           const lsn_t end{recv_sys.file_checkpoint};
           ut_ad(!end || end == recv_sys.lsn);
+          bool corrupt_fs = recv_sys.is_corrupt_fs();
           mysql_mutex_unlock(&recv_sys.mutex);
 
-          if (!end)
+          if (!end && !corrupt_fs)
           {
             recv_sys.set_corrupt_log();
             sql_print_error("InnoDB: Missing FILE_CHECKPOINT(" LSN_PF
@@ -4602,6 +4603,9 @@ read_only_recovery:
 					LSN_PF, recv_sys.lsn);
                         goto err_exit;
 		}
+		if (recv_sys.is_corrupt_fs()) {
+			goto err_exit;
+		}
 		ut_ad(recv_sys.file_checkpoint);
 		if (rewind) {
 			recv_sys.lsn = log_sys.next_checkpoint_lsn;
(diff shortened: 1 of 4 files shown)


dr-m (changes requested):
Similar to the 10.6 version #3032 this test is always skipped on my system. It is not something that was introduced by these changes, but I think that we’d better reduce the probability of the test being skipped, by ensuring that a checkpoint will be made before the critical section (`no_checkpoint_start.inc`).

storage/innobase/log/log0recv.cc
```diff
@@ -4122,7 +4122,7 @@ static bool recv_scan_log(bool last_phase)
           ut_ad(!end || end == recv_sys.lsn);
           mysql_mutex_unlock(&recv_sys.mutex);
 
-          if (!end)
+          if (!recv_sys.is_corrupt_fs() && !end)
           {
             recv_sys.set_corrupt_log();
```
dr-m, line 4127: `recv_sys.found_corrupt_fs` is supposed to be protected by `recv_sys.mutex` if you look at `recv_sys_t::set_corrupt_fs()`. I think that we’d better not release `recv_sys.mutex` before we have decided whether to set `recv_sys.found_corrupt_log`.

Another thing is that we’d better check something that likely to reside in a register, before reading something from memory, that is, `!end && recv_sys.is_corrupt_fs()`.

storage/innobase/log/log0recv.cc
```diff
@@ -4602,6 +4602,9 @@ dberr_t recv_recovery_from_checkpoint_start()
 					LSN_PF, recv_sys.lsn);
                         goto err_exit;
 		}
+		if (recv_sys.is_corrupt_fs()) {
+			goto err_exit;
+		}
```
dr-m, line 4607: Here we are not holding `recv_sys.mutex`, but it should not really matter much, because this is the only thread that can modify any `recv_sys` members at this point.

dr-m (approved):
OK to push after a minor adjustment.

storage/innobase/log/log0recv.cc
mariadb-DebarunBanerjee: Makes sense. Modified the code.

storage/innobase/log/log0recv.cc
mariadb-DebarunBanerjee: Makes sense. Modified the code.

dr-m:
Closed by 86c2c897432ab3568ce1442992dcded3f2c5b974.

PR #3035, opened 2024-01-30
MDEV-7850: GTID Thread_id Post-Push Fixes

bnestere:
Note this is currently a work-in-progress, pushed for potential
early feedback.

It addresses:
 1) The original thread_id of the master is propagated into
    the thread_id of the slave-binlogged GTID thread_id,
    rather than the thread_id of the applier thread.
 2) When seeing Q_DUMMY, rather than loop through each
    status var byte, skip to the end of the status var
    section.
 3) Added comments to explain the purpose of Q_DUMMY
 4) Simplified Query_log_event::begin_event to always use
    Q_DUMMY, rather than keeping the time zone str as well
 5) Fixed DBUG_ASSERT to be part of the conditoin to read in
    the event thread id, as to not read unowned memory
 6) Removed invalid ASSERTIONS for begin_event()

### DIFF
--- a/mysql-test/suite/galera_sr/r/MDEV-18585.result
+++ b/mysql-test/suite/galera_sr/r/MDEV-18585.result
@@ -12,7 +12,7 @@ SET SESSION wsrep_trx_fragment_size=1;
 INSERT INTO t1 VALUES (5), (6);
 SET SESSION wsrep_trx_fragment_unit=default;
 SET SESSION wsrep_trx_fragment_size=default;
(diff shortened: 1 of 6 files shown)


knielsen (changes requested):
Review on mailing list: https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/message/CHFD7KWOOW7YL4VA6Q4DTZWYET7CP5O2/

Ok from my side after fixing the missing `break` in the switch statement

PR #3037, opened 2024-02-01
MDEV-18288 Transportable Tablespaces leave AUTO_INCREMENT in mismatch…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-18288*

## Description
During import, if cfg file is not specified, we don't update the autoinc
field in innodb dictionary object dict_table_t. The next insert tries to
insert from the starting position of auto increment and fails.
    
It can be observed that the issue is resolved once server is restarted
as the persistent value is read correctly from PAGE_ROOT_AUTO_INC from
index root page. The patch fixes the issue by reading the the auto
increment value directly from PAGE_ROOT_AUTO_INC during import if cfg
file is not specified.

## How can this PR be tested?

./mtr innodb.import_bugs

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/full_crc32_import.result
+++ b/mysql-test/suite/innodb/r/full_crc32_import.result
@@ -48,7 +48,15 @@ t1	CREATE TABLE `t1` (
   `b` blob DEFAULT NULL,
   `c` blob DEFAULT NULL,
   PRIMARY KEY (`a`)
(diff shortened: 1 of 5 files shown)


dr-m (approved):

mariadb-DebarunBanerjee:
pushed to 10.5
commit 66bb229e910bfeff9fd41ee9d8743344e0738025 (HEAD -> 10.5, origin/10.5-MDEV-18288, origin/10.5, 10.5-MDEV-18288)
Author: mariadb-DebarunBanerjee <debarun.banerjee@mariadb.com>
Date:   Thu Feb 1 12:32:02 2024 +0530

PR #3040, opened 2024-02-02
MDEV-12404 Index condition pushdown on partitioned tables

DaveGosselin-MariaDB:
MDEV-12404 Index condition pushdown on partitioned tables

    Support index condition pushdown within partitioned tables.
    - ha_partition will pass the pushed index condition into all of the used
      partitions.
      - We require that all of the partitions to handle the pushed index
        condition in the same way.
    - When using ICP, one may read rows (e.g. call h->index_read_map(buf, ...)
      only to buf= table->record[0], for two reasons:
      * Pushed index condition's Item_field objects point into record[0]
      * InnoDB requires this: it calls offset() which assumes record[0].
      So, when using ICP, ha_partition will read partition records to
      table->record[0] and then will copy record away if it needs it to be
      elsewhere.

### DIFF
--- a/sql/ha_partition.cc
+++ b/sql/ha_partition.cc
@@ -62,6 +62,7 @@
 #include "ddl_log.h"
 
 #include "debug_sync.h"
+#include <functional>
 
 /* First 4 bytes in the .par file is the number of 32-bit words in the file */
 #define PAR_WORD_SIZE 4
@@ -81,6 +82,30 @@
                                         HA_CAN_TABLES_WITHOUT_ROLLBACK)
 
 static const char *ha_par_ext= PAR_EXT;
+
+/*
+  Index Condition Pushdown relies on invoking val_int() on the pushed index
+  condition to see if the condition applies to the current row.  For that
+  to work we need to ensure that table->record[0] is populated correctly
+  because Item instances rely on table->record[0] for item evaluation.  Doing
+  so would complicate the calling code, so instead encapsulate that behind
+  this function to juggle between buffers like m_ordered_rec_buffer, etc and
+  table->record[0] during index condition evaluation (if present).
+*/
+using IndexOperationFunc= std::function<int(uchar* read_buf)>;
+static int read_with_icp(TABLE* table,
+                         bool has_idx_cond,
+                         uchar* record_buf,
+                         size_t record_buf_len,
+                         IndexOperationFunc func)
+{
+  if (!has_idx_cond)
+    return func(record_buf);
+
+  int error= func(table->record[0]);
+  memcpy(record_buf, table->record[0], record_buf_len);
+  return error;
+}
 
 /****************************************************************************
                 MODULE create/delete handler object
@@ -7941,10 +7966,17 @@ int ha_partition::handle_ordered_index_scan(uchar *buf, bool reverse_order)
 
     switch (m_index_scan_type) {
     case partition_index_read:
-      error= file->ha_index_read_map(rec_buf_ptr,
-                                     m_start_key.key,
-                                     m_start_key.keypart_map,
-                                     m_start_key.flag);
+      error=
+        read_with_icp(table,
+                      pushed_idx_cond != nullptr,
+                      rec_buf_ptr,
+                      m_rec_length,
+                      [this, file] (uchar* read_buf) {
+                        return file->ha_index_read_map(read_buf,
+                                                       m_start_key.key,
+                                                       m_start_key.keypart_map,
+                                                       m_start_key.flag);
+                      });
       /* Caller has specified reverse_order */
       break;
     case partition_index_first:
@@ -8399,10 +8431,27 @@ int ha_partition::handle_ordered_next(uchar *buf, bool is_next_same)
     }
   }
   else if (!is_next_same)
-    error= file->ha_index_next(rec_buf);
+  {
+    error= read_with_icp(table,
+                         pushed_idx_cond != nullptr,
+                         rec_buf,
+                         m_rec_length,
+                         [file] (uchar* read_buf) {
+                           return file->ha_index_next(read_buf);
+                         });
+  }
   else
-    error= file->ha_index_next_same(rec_buf, m_start_key.key,
-                                    m_start_key.length);
+  {
+    error= read_with_icp(table,
+                         pushed_idx_cond != nullptr,
+                         rec_buf,
+                         m_rec_length,
+                           [this, file] (uchar* read_buf) {
+                             return file->ha_index_next_same(read_buf,
+                                                             m_start_key.key,
+                                                             m_start_key.length);
+                           });
+  }
 
   if (unlikely(error))
   {
@@ -12227,6 +12276,57 @@ int ha_partition::info_push(uint info_type, void *info)
   DBUG_RETURN(error);
 }
 
+static void cancel_pushed_idx_cond_impl(handler** m_file,
+                                        MY_BITMAP* read_partitions,
+                                        uint limit)
+{
+  for (uint i= bitmap_get_first_set(read_partitions);
+       i < limit;
+       i= bitmap_get_next_set(read_partitions, i))
+  {
+    m_file[i]->cancel_pushed_idx_cond();
+  }
+}
+
+Item* ha_partition::idx_cond_push(uint keyno, Item* idx_cond)
+{
+  DBUG_ASSERT(pushed_idx_cond == nullptr);
+  DBUG_ASSERT(pushed_idx_cond_keyno == MAX_KEY);
+
+  for (uint i= bitmap_get_first_set(&m_part_info->read_partitions);
+       i < m_tot_parts;
+       i= bitmap_get_next_set(&m_part_info->read_partitions, i))
+  {
+    Item* res= m_file[i]->idx_cond_push(keyno, idx_cond);
+    if (!res)  // Returning nullptr indicates success.
+      continue;
+
+    // One of the partitions couldn't accept the pushed condition, or
+    // one of the partitions returned a partial pushed condition that
+    // indicates that it could handle some portion of the pushed index
+    // condition.  At this point, we require all partitions to handle
+    // the pushed condition in the same way; consequently  we need to
+    // cancel the pushed condition for the partitions that succeeded
+    // up to this point.
+    DBUG_ASSERT(i == bitmap_get_first_set(&m_part_info->read_partitions));
+    DBUG_ASSERT(res == idx_cond);
+    if (res != idx_cond)
+      m_file[i]->cancel_pushed_idx_cond();
+    cancel_pushed_idx_cond_impl(m_file, &m_part_info->read_partitions, i);
+    return idx_cond;
+  }
+  pushed_idx_cond= idx_cond;
+  pushed_idx_cond_keyno= keyno;
+  in_range_check_pushed_down = TRUE;
+  return NULL;
+}
+
+void ha_partition::cancel_pushed_idx_cond()
+{
+  cancel_pushed_idx_cond_impl(m_file, &m_part_info->read_partitions,
+                              m_tot_parts);
+  handler::cancel_pushed_idx_cond();
+}
 
 bool
 ha_partition::can_convert_nocopy(const Field &field,
--- a/mysql-test/main/ignored_index.result
+++ b/mysql-test/main/ignored_index.result
@@ -371,7 +371,7 @@ id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
 1	SIMPLE	t1	index	NULL	a	2	NULL	14	Using index
 EXPLAIN SELECT * FROM t1 WHERE a = '04';
 id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
-1	SIMPLE	t1	ref	a	a	2	const	2	Using where
+1	SIMPLE	t1	ref	a	a	2	const	2	Using index condition
 ALTER TABLE t1 ALTER INDEX b IGNORED;
 EXPLAIN SELECT b FROM t1;
 id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
--- a/mysql-test/main/partition_icp.result
+++ b/mysql-test/main/partition_icp.result
@@ -1,0 +1,160 @@
+set @old_handler_icp_attempts=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_attempts');
+set @old_handler_icp_match=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_match');
+CREATE TABLE t1 (a int PRIMARY KEY, b BLOB, c varchar(16) DEFAULT 'Filler...',
+INDEX (b(4), a)) ENGINE = MyISAM PARTITION BY HASH (a) PARTITIONS 3;
+INSERT INTO t1 (a, b) VALUES (1, 0xdeadbeef), (2, "text filler"),
+(3, 'filler...'), (4, " more filler "), (5, "test text"), (6, "testing...");
+SELECT a, HEX(b) FROM t1 WHERE b >= 'te' and (a % 2);
+a	HEX(b)
+5	746573742074657874
+1	DEADBEEF
+set @handler_icp_attempts=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_attempts');
+set @handler_icp_match=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_match');
+select @old_handler_icp_attempts = @handler_icp_attempts;
+@old_handler_icp_attempts = @handler_icp_attempts
+1
+select @old_handler_icp_match = @handler_icp_match;
+@old_handler_icp_match = @handler_icp_match
+1
+CREATE TABLE t2 (a int PRIMARY KEY, b BLOB, c varchar(16) DEFAULT 'Filler...',
+INDEX (b(4), a)) ENGINE = innodb PARTITION BY HASH (a) PARTITIONS 3;
+INSERT INTO t2 (a, b) VALUES (1, 0xdeadbeef), (2, "text filler"),
+(3, 'filler...'), (4, " more filler "), (5, "test text"), (6, "testing...");
+set @@optimizer_switch='index_condition_pushdown=off';
+SELECT a, HEX(b) FROM t2 WHERE b >= 'te' and (a % 2);
+a	HEX(b)
+5	746573742074657874
+1	DEADBEEF
+set @handler_icp_attempts=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_attempts');
+set @handler_icp_match=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_match');
+select @old_handler_icp_attempts = @handler_icp_attempts;
+@old_handler_icp_attempts = @handler_icp_attempts
+1
+select @old_handler_icp_match = @handler_icp_match;
+@old_handler_icp_match = @handler_icp_match
+1
+set @@optimizer_switch='index_condition_pushdown=on';
+SELECT a, HEX(b) FROM t2 WHERE b >= 'te' and (a % 2);
+a	HEX(b)
+5	746573742074657874
+1	DEADBEEF
+set @handler_icp_attempts=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_attempts');
+set @handler_icp_match=(select cast(variable_value as int) from information_schema.session_status where variable_name='handler_icp_match');
+select @old_handler_icp_attempts < @handler_icp_attempts;
+@old_handler_icp_attempts < @handler_icp_attempts
+1
+select @old_handler_icp_match < @handler_icp_match;
+@old_handler_icp_match < @handler_icp_match
+1
+CREATE TABLE t3 (
+pk INT PRIMARY KEY,
+a INT,
+b INT,
+c INT,
+filler varchar(100),
+key(a,b,c)
+) engine=innodb partition by hash(pk) partitions 4;
+INSERT INTO t3 SELECT seq, seq, seq, seq, 'hello' FROM seq_1_to_10000;
+set @@optimizer_switch='index_condition_pushdown=off';
+set @old_handler_icp_attempts = @handler_icp_attempts;
+set @old_handler_icp_match = @handler_icp_match;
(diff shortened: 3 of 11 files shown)


DaveGosselin-MariaDB:
The test was failing because some session level statistics that I used to verify whether index condition pushdown has an effect were not stable.  Somehow my updates have made the test less stable than before, even though it passes locally.  I'll investigate further of course.

sql/ha_partition.cc
```diff
@@ -12227,6 +12227,67 @@ int ha_partition::info_push(uint info_type, void *info)
...
+
+   for (uint i= bitmap_get_first_set(&m_part_info->read_partitions);
+        i < m_tot_parts;
+        i= bitmap_get_next_set(&m_part_info->read_partitions, i))
+   {
+     // We don't support pushdown to myisam because of issues with
+     // handling null selects.  According to our KB, "There is
+     // usually little reason to use it [MyISAM] except for legacy
+     // purposes."  It seems reasonable to exclude it from future
+     // developments, like index condition pushdown.
```
spetrunia, line 12243: This sounds very odd. Please provide more details about "issues with handling null selects"...

spetrunia:
Note that parts.partition_alter1_1_2_innodb fails: https://buildbot.mariadb.org/#/builders/572/builds/6465/steps/8/logs/stdio

DaveGosselin-MariaDB:
> Note that parts.partition_alter1_1_2_innodb fails: https://buildbot.mariadb.org/#/builders/572/builds/6465/steps/8/logs/stdio

I was going to respond by saying "it doesn't fail on my machine," but MTR doesn't run it by default on my system because it requires `--big-test`.  So it is skipped when running locally, which is why I missed it before submitting.  😞

sql/ha_partition.cc
DaveGosselin-MariaDB: Yes, I am using the PR to make a suggestion and to receive feedback on the approach.  (I'm not sure of a better way to share diffs and get feedback, please let me know if there is a better way).

There are some MyISAM tests where we get unexpected null results simply because we support index condition pushdown in this patch.  They are:
 - parts.partition_alter1_1_2_myisam 
 - parts.partition_basic_myisam 
 - parts.partition_basic_symlink_myisam
I don't think it is correct to merely re-record them and add them to the PR.  At the same time, I didn't think it was worth supporting MyISAM because it is legacy and there is little reason to use it, according to our own documentation.  It is fine to support it of course, but I would like to know what to do to address the test failures.  Here's an example failure from `parts.partition_alter1_1_2_myisam`:
```
@@ -182,7 +182,7 @@
 f_charbig = '#NULL#';
 # check null success:    1

-# check null-1 success: 	1
+# check null-1 success: 	0
 UPDATE t1 SET f_int1 = -@max_row
 WHERE f_int1 IS NULL AND f_int2 = -@max_row AND f_char1 = CAST(-@max_row AS CHAR)
 AND f_char2 = CAST(-@max_row AS CHAR) AND f_charbig = '#NULL#';
```
To me, this looks like a wrong result.

spetrunia:
You need to debug it.
my suggestion on how to proceed:
1. isolate a failing example - extract it from that large test file
2. Do side-by side debugging (or executions) of the server before the patch and after the patch.
3. Put breakpoints or print-outs  on A) index read functions made by the problematic query, so that you can see what functions are called and their return value..
4.  Then you'll be able to see which call didn't return a value while it should have.
5. Then look at the pushed index condition check - did it return the right value?
6.

spetrunia:
I see...  Yes, the suggested steps are time-consuming. But unfortunately I can't offer any magic bullet here :-(.

DaveGosselin-MariaDB:
No worries.  I was only looking for a leg-up with respect to optimizer pushdown, I'm happy to spend the time to debug but was hoping that you may have a shortcut.  It's no problem.

DaveGosselin-MariaDB:
The latest patch is working with MyISAM.

spetrunia:
(I have taken the patch from this PR and enabled myisam.  `parts.partition_basic_myisam` has failed. Now debugging it).

There is one matching record in one partition.
Eventually we reach this point where MyISAM is checking the pushed index condition:

```  
#0  _mi_put_key_in_record (info=0x7fffe09950c8, keynr=0, unpack_blobs=0 '\000', record=0x7fffe09f3472 '\245' <repeats 200 times>...) at /home/psergey/dev-git2/11.4-mdev12404/storage/myisam/mi_key.c:355
  #1  0x0000555556b9af29 in mi_unpack_index_tuple (info=0x7fffe09950c8, keynr=0, record=0x7fffe09f3472 '\245' <repeats 200 times>...) at /home/psergey/dev-git2/11.4-mdev12404/storage/myisam/mi_key.c:501
  #2  0x0000555556b9b033 in mi_check_index_tuple_real (info=0x7fffe09950c8, keynr=0, record=0x7fffe09f3472 '\245' <repeats 200 times>...) at /home/psergey/dev-git2/11.4-mdev12404/storage/myisam/mi_key.c:532
  #3  0x0000555556bccffb in mi_check_index_tuple (info=0x7fffe09950c8, keynr=0, record=0x7fffe09f3472 '\245' <repeats 200 times>...) at /home/psergey/dev-git2/11.4-mdev12404/storage/myisam/myisamdef.h:752
  #4  0x0000555556bcd70c in mi_rkey (info=0x7fffe09950c8, buf=0x7fffe09f3472 '\245' <repeats 200 times>..., inx=0, key=0x7fffe001ba80 "\001", keypart_map=1, search_flag=HA_READ_KEY_EXACT) at /home/psergey/dev-gi
  #5  0x0000555556b75441 in ha_myisam::index_read_map (this=0x7fffe09873f0, buf=0x7fffe09f3472 '\245' <repeats 200 times>..., key=0x7fffe001ba80 "\001", keypart_map=1, find_flag=HA_READ_KEY_EXACT) at /home/pserg
  #6  0x0000555556385a24 in handler::ha_index_read_map (this=0x7fffe09873f0, buf=0x7fffe09f3472 '\245' <repeats 200 times>..., key=0x7fffe001ba80 "\001", keypart_map=1, find_flag=HA_READ_KEY_EXACT) at /home/pser
  #7  0x000055555667a261 in ha_partition::handle_ordered_index_scan (this=0x7fffe0986ad0, buf=0x7fffe0988550 <incomplete sequence \374>, reverse_order=false) at /home/psergey/dev-git2/11.4-mdev12404/sql/ha_parti
  #8  0x000055555667437c in ha_partition::common_index_read (this=0x7fffe0986ad0, buf=0x7fffe0988550 <incomplete sequence \374>, have_start_key=true) at /home/psergey/dev-git2/11.4-mdev12404/sql/ha_partition.cc:
  #9  0x0000555556673ee6 in ha_partition::index_read_map (this=0x7fffe0986ad0, buf=0x7fffe0988550 <incomplete sequence \374>, key=0x7fffe001ba80 "\001", keypart_map=1, find_flag=HA_READ_KEY_EXACT) at /home/pserg
  #10 0x00005555563858bd in handler::ha_index_read_map (this=0x7fffe0986ad0, buf=0x7fffe0988550 <incomplete sequence \374>, key=0x7fffe001ba80 "\001", keypart_map=1, find_flag=HA_READ_KEY_EXACT) at /home/psergey
  #11 0x000055555602a329 in join_read_always_key (tab=0x7fffe09f54f0) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_select.cc:24396
  #12 0x0000555556028503 in sub_select (join=0x7fffe001a148, join_tab=0x7fffe09f54f0, end_of_records=false) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_select.cc:23601
  #13 0x00005555560276ae in do_select (join=0x7fffe001a148, procedure=0x0) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_select.cc:23115
  #14 0x0000555555ff44e6 in JOIN::exec_inner (this=0x7fffe001a148) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_select.cc:4988
  #15 0x0000555555ff3531 in JOIN::exec (this=0x7fffe001a148) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_select.cc:4774
  #16 0x0000555555ff4f58 in mysql_select (thd=0x7fffe0000d78, tables=0x7fffe00189d8, fields=@0x7fffe0018208: {<base_list> = {<Sql_alloc> = {<No data fields>}, first = 0x7fffe0018510, last = 0x7fffe0018990, eleme
  #17 0x0000555555fe37c9 in handle_select (thd=0x7fffe0000d78, lex=0x7fffe0005118, result=0x7fffe001a120, setup_tables_done_option=0) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_select.cc:630
  #18 0x0000555555f89480 in execute_sqlcom_select (thd=0x7fffe0000d78, all_tables=0x7fffe00189d8) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_parse.cc:6077
  #19 0x0000555555f813bd in mysql_execute_command (thd=0x7fffe0000d78, is_called_from_prepared_stmt=false) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_parse.cc:3926
  #20 0x0000555555f8e244 in mysql_parse (thd=0x7fffe0000d78, rawbuf=0x7fffe0017e10 "SELECT '# check null-1 success: ' AS \"\",COUNT(*) = 1 AS \"\" FROM t1\nWHERE f_int1 IS NULL AND f_charbig = '#NULL#'", length=
```

`_mi_put_key_in_record()` unpacks the NULL-bit:

```cpp
      record[keyseg->null_pos]&= ~keyseg->null_bit;
```

```
(gdb) p keyseg->null_pos
  $107 = 0
(gdb) p record
  $108 = (uchar *) 0x7fffe09f3472 '\245' <repeats 200 times>...
(gdb) set $inner_rec=record
```
Just remember the pointer.

Continue until we hit the IS NOT NULL check here:
```
(gdb) wher 10
  #0  Field::is_null (this=0x7fffe09867b0, row_offset=0) at /home/psergey/dev-git2/11.4-mdev12404/sql/field.h:1398
  #1  0x0000555555e8c666 in Item_field::is_null (this=0x7fffe0019108) at /home/psergey/dev-git2/11.4-mdev12404/sql/item.h:3708
  #2  0x00005555563dc8b9 in Item_func_isnull::val_int (this=0x7fffe0019228) at /home/psergey/dev-git2/11.4-mdev12404/sql/item_cmpfunc.cc:5719
  #3  0x000055555638feb7 in handler_index_cond_check (h_arg=0x7fffe09873f0) at /home/psergey/dev-git2/11.4-mdev12404/sql/handler.cc:6993
  #4  0x0000555556b9b06c in mi_check_index_tuple_real (info=0x7fffe09950c8, keynr=0, record=0x7fffe09f3472 "\245\245\245\245\245\354\377\377\377", '\245' <repeats 191 times>...) at /home/psergey/dev-git2/11.4-md
  #5  0x0000555556bccffb in mi_check_index_tuple (info=0x7fffe09950c8, keynr=0, record=0x7fffe09f3472 "\245\245\245\245\245\354\377\377\377", '\245' <repeats 191 times>...) at /home/psergey/dev-git2/11.4-mdev124
  #6  0x0000555556bcd70c in mi_rkey (...
```

We are in
```cpp
    inline bool is_null(my_ptrdiff_t row_offset= 0) const
    { 
...
=>    return null_ptr ? null_ptr[row_offset] & null_bit : table->null_row;                                                                                                                                         
```
```
(gdb) print null_ptr
  $113 = (uchar *) 0x7fffe0988550 <incomplete sequence \374>
(gdb) p null_ptr - $inner_rec
  $114 = -438050
```
Note that this location DOES NOT MATCH the location where null-bit was unpacked.

DaveGosselin-MariaDB:
I think you did not pull the latest version of this patch.  `parts.partition_basic_myisam` passes with this code, git sha `a0bc30b199c1f52422bf1829dbded70160165a27`.

spetrunia:
Noting the input to not forget it:
* Need to remove the "accessor" debugging aid functions from this patch.  Should be in a separate patch or not be at all (why accessor funcs when there are watchpoints?)
* `ha_partition::handle_ordered_index_scan_key_not_found()`  and `ha_partition::handle_ordered_next()` also call Storage Engine functions with `buf != table->record[0]`.   Do we need to provide handling for those?
* ( A general discussion about the SE API and whether we make the "don't make calls with buf!=record[0] when ICP is ON" a documented/assert-ed part of the API or do something else...)

DaveGosselin-MariaDB:
> * Need to remove the "accessor" debugging aid functions from this patch.  Should be in a separate patch or not be at all (why accessor funcs when there are watchpoints?)
I will remove them from this patch.  Watch points require that you know the object address ahead of time, which is possible, but inconvenient.  Hooking the call is more convenient because it is one step. Ideally the grammar would only invoke hooks and not do any work, preserving a clean separation.
> * `ha_partition::handle_ordered_index_scan_key_not_found()`  and `ha_partition::handle_ordered_next()` also call Storage Engine functions with `buf != table->record[0]`.   Do we need to provide handling for those?
Unless we have a valid test case, then no.  These could be dead code given our test coverage.  Alternatively, we may be able to come up with queries that exercise this path, but I don't know what they could be.
> * ( A general discussion about the SE API and whether we make the "don't make calls with buf!=record[0] when ICP is ON" a documented/assert-ed part of the API or do something else...)
This is a good idea, let's get consensus on documenting it or adding an assertion.
> * Test coverage for partitioning+ICP using range access. Do we have it?
Yes, I have such a query and will add it to the test cases I introduced in this patch.

spetrunia:
Here's an example which causes reads to wrong record:

```sql
create table t1(pk int primary key, kp1 int, kp2 int, filler int, key(kp1, kp2)) 
  partition by hash(pk) partitions 2; 
insert into t1 select seq, seq/10, seq, seq from seq_1_to_1000;
explain select * from t1 where t1.kp1=10 and mod(t1.kp2,2)=1; 
```
```
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref   | rows | Extra                 |
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
|    1 | SIMPLE      | t1    | ref  | kp1           | kp1  | 5       | const | 10   | Using index condition |
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
1 row in set (0.001 sec)
```

```
select * from t1 where t1.kp1=10 and mod(t1.kp2,3)=1; 
```
```
(gdb) wher
  #0  ha_innobase::index_next_same (this=0x7fff3002dd20, buf=0x7fff30087242 "\361d") at /home/psergey/dev-git2/11.4-mdev12404/storage/innobase/handler/ha_innodb.cc:9352
  #1  0x000055555638f9f2 in handler::ha_index_next_same (this=0x7fff3002dd20, buf=0x7fff30087242 "\361d", key=0x7fff3001a670 "", keylen=5) at /home/psergey/dev-git2/11.4-mdev124
  #2  0x000055555668709f in ha_partition::handle_ordered_next (this=0x7fff3002d400, buf=0x7fff30021a70 "\361d", is_next_same=true) at /home/psergey/dev-git2/11.4-mdev12404/sql/h
  #3  0x000055555667fad3 in ha_partition::index_next_same (this=0x7fff3002d400, buf=0x7fff30021a70 "\361d", key=0x7fff3001a670 "", keylen=5) at /home/psergey/dev-git2/11.4-mdev1
  #4  0x000055555638f9f2 in handler::ha_index_next_same (this=0x7fff3002d400, buf=0x7fff30021a70 "\361d", key=0x7fff3001a670 "", keylen=5) at /home/psergey/dev-git2/11.4-mdev124
  #5  0x0000555556031d52 in join_read_next_same (info=0x7fff3008da90) at /home/psergey/dev-git2/11.4-mdev12404/sql/sql_select.cc:24464
  #6  0x0000555555e3a5fb in READ_RECORD::read_record (this=0x7fff3008da90) at /home/psergey/dev-git2/11.4-mdev12404/sql/records.h:81
```

DaveGosselin-MariaDB:
@spetrunia what is the wrong result?  If I create the same table and inspect the result of the select, I see:
```
MariaDB [test]> create table t1(pk int primary key, kp1 int, kp2 int, filler int, key(kp1, kp2))
    ->   partition by hash(pk) partitions 2;
Query OK, 0 rows affected (0.015 sec)

MariaDB [test]> insert into t1 select seq, seq/10, seq, seq from seq_1_to_1000;
Query OK, 1000 rows affected (0.051 sec)
Records: 1000  Duplicates: 0  Warnings: 0

MariaDB [test]> explain select * from t1 where t1.kp1=10 and mod(t1.kp2,2)=1;
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref   | rows | Extra                 |
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
|    1 | SIMPLE      | t1    | ref  | kp1           | kp1  | 5       | const | 10   | Using index condition |
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
1 row in set (0.001 sec)

MariaDB [test]> select * from t1 where t1.kp1=10 and mod(t1.kp2,2)=1;
+-----+------+------+--------+
| pk  | kp1  | kp2  | filler |
+-----+------+------+--------+
|  95 |   10 |   95 |     95 |
|  97 |   10 |   97 |     97 |
|  99 |   10 |   99 |     99 |
| 101 |   10 |  101 |    101 |
| 103 |   10 |  103 |    103 |
+-----+------+------+--------+
5 rows in set (0.001 sec)
```

This looks correct.  If I create another table without partitions and run the same queries, the results match:

```
MariaDB [test]> create table t2(pk int primary key, kp1 int, kp2 int, filler int, key(kp1, kp2));
Query OK, 0 rows affected (0.012 sec)

MariaDB [test]> insert into t2 select seq, seq/10, seq, seq from seq_1_to_1000;
Query OK, 1000 rows affected (0.064 sec)
Records: 1000  Duplicates: 0  Warnings: 0

MariaDB [test]> select * from t2 where t1.kp1=10 and mod(t1.kp2,2)=1;
ERROR 1054 (42S22): Unknown column 't1.kp1' in 'where clause'
MariaDB [test]> select * from t2 where t2.kp1=10 and mod(t2.kp2,2)=1;
+-----+------+------+--------+
| pk  | kp1  | kp2  | filler |
+-----+------+------+--------+
|  95 |   10 |   95 |     95 |
|  97 |   10 |   97 |     97 |
|  99 |   10 |   99 |     99 |
| 101 |   10 |  101 |    101 |
| 103 |   10 |  103 |    103 |
+-----+------+------+--------+
5 rows in set (0.003 sec)

MariaDB [test]> explain select * from t2 where t2.kp1=10 and mod(t2.kp2,2)=1;
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref   | rows | Extra                 |
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
|    1 | SIMPLE      | t2    | ref  | kp1           | kp1  | 5       | const | 10   | Using index condition |
+------+-------------+-------+------+---------------+------+---------+-------+------+-----------------------+
1 row in set (0.003 sec)
```

Please clarify what you mean by wrong record, thank you.

spetrunia:
I meant this condition:
```
  Thread 13 "mysqld" hit Breakpoint 8, ha_innobase::index_next_same (this=0x7fff3001e8d0, buf=0x7fff3008d27d "\361_") at /home/psergey/dev-git2/11.4-mdev12404/storage/  innobase/handler/ha_innodb.cc:9352
(gdb)  p buf == table->record[0]
  $48 = false
(gdb) p pushed_idx_cond
  $49 = (Item_func_eq *) 0x7fff30017de0
```
here we have buf != table->record[0] ... and ICP...  Didn't we agree that in this case ICP might not be checked correctly?

DaveGosselin-MariaDB:
When we call the underlying storage engine, we need to be sure that we pass table->record[0] for correct ICP evaluation, but `index_next_same` is too early.  Look at `handle_ordered_next` (called by `index_next_same`) which appears to do the right thing and memcpy to `rec_buf` after calls to the underlying engine.  Maybe we need to fixup the calls to `file->ha_index_next` and `file->ha_index_next_same` ?  But what would be the test SQL query to ensure that we do it correctly if a change here is indeed needed?

spetrunia:
> @spetrunia what is the wrong result? If I create the same table and inspect the result of the select, I see:

This is a wrong approach to take. The query returned correct result by accident. Make a testcase larger, make the ICP condition such that it is not lucky and:

```sql
create table t2(pk int primary key, kp1 int, kp2 int, filler int, key(kp1, kp2))
  partition by hash(pk) partitions 2;
insert into t2 select seq, seq/40, seq, seq from seq_1_to_4000;
```

```
select * from t2 where t2.kp1=10 and mod(t2.kp2,2)=1 and t2.kp2+1<401;
+-----+------+------+--------+
| pk  | kp1  | kp2  | filler |
+-----+------+------+--------+
| 381 |   10 |  381 |    381 |
| 383 |   10 |  383 |    383 |
| 385 |   10 |  385 |    385 |
| 387 |   10 |  387 |    387 |
| 389 |   10 |  389 |    389 |
| 391 |   10 |  391 |    391 |
| 393 |   10 |  393 |    393 |
| 395 |   10 |  395 |    395 |
| 397 |   10 |  397 |    397 |
| 399 |   10 |  399 |    399 |
| 401 |   10 |  401 |    401 |
| 403 |   10 |  403 |    403 |
| 405 |   10 |  405 |    405 |
| 407 |   10 |  407 |    407 |
| 409 |   10 |  409 |    409 |
| 411 |   10 |  411 |    411 |
+-----+------+------+--------+

```
```
select * from t2 use index() where t2.kp1=10 and mod(t2.kp2,2)=1 and t2.kp2+1<401;
+-----+------+------+--------+
| pk  | kp1  | kp2  | filler |
+-----+------+------+--------+
| 381 |   10 |  381 |    381 |
| 383 |   10 |  383 |    383 |
| 385 |   10 |  385 |    385 |
| 387 |   10 |  387 |    387 |
| 389 |   10 |  389 |    389 |
| 391 |   10 |  391 |    391 |
| 393 |   10 |  393 |    393 |
| 395 |   10 |  395 |    395 |
| 397 |   10 |  397 |    397 |
| 399 |   10 |  399 |    399 |
+-----+------+------+--------+
10 rows in set (0.013 sec)
```

DaveGosselin-MariaDB:
OK I see now, I didn't realize that it was correct by accident.

spetrunia (approved):

PR #3041, opened 2024-02-02
MDEV-32424 Crash in subselect engine

DaveGosselin-MariaDB:
According to MDEV-29363, recompute the with_subquery() attribute in the pushdown code to avoid writing to freed memory.


grooverdan:
This appears fixed in the server, so closing.

PR #3042, opened 2024-02-04
Fix commit 179424db: No test file or result files should be executable

ottok:
## Description

In commit 179424db the file lowercase_table2.result was made executable for no known reason, most likely just a mistake. Test result files definitely should not be executable.

This is related to #2779. [I recommend using git ci-tool](https://optimizedbyotto.com/post/good-git-commit/) to make commits to have better visibility into what is committed and thus more easily spot if extra changes are about to slip in (such as this mode change).

## PR quality check

## Raise the quality bar requirement

I wish this PR is reviewed and merged by the same author who's commit it is fixing to allow for personalized feedback to happen.


abarkov (approved):
The patch is ok. Thanks.

PR #3043, opened 2024-02-05
MDEV-32034: Second SELECT from view created from INFORMATION_SCHEMA.O…

Olernov:
…PTIMIZER_TRACE gives NULL

Function list_has_optimizer_trace_table() is responsible for checking whether the query includes I_S.OPTIMIZER_TRACE. However, the function is called at an early stage, before references are resolved, resulting in the omission of cases where a SELECT is made from a view constructed from I_S.OPTIMIZER_TRACE.

To address this, an supplementary call to
list_has_optimizer_trace_table() is added at the end of tracing, when all the information is available. That allows to identify cases in which the SELECT is made from a view or a stored procedure containing I_S.OPTIMIZER_TRACE

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check


sql/opt_trace.cc
```diff
@@ -448,12 +442,22 @@ void Opt_trace_context::start(THD *thd, TABLE_LIST *tbl,
...
 {
-  if (current_trace)
+  /*
+    The second call to list_has_optimizer_trace_table() here yiels more accurate
+    results compared to the first one in the Opt_trace_start constructor.
+    At this point, references resolved, enabling us to determine whether
+    the traced query includes a view created from `I_S.OPTIMIZER_TRACE`.
+    If so, the current trace is discarded and the previous one remains.
+  */
+  if (current_trace && !list_has_optimizer_trace_table(tables_list))
```
lottaquestions, line 457: Interesting solution. By running  a few tests I noticed that in the function `Opt_trace_start::init`, the condition that checks `!list_has_optimizer_trace_table(tables_list))` completely prevents the creation of a new `current_trace` if the query does have `I_S.OPTIMIZER_TRACE`, e.g. the query `SELECT JSON_DETAILED(JSON_EXTRACT(trace, '$**.selectivity_for_columns')) AS JS   FROM information_schema.optimizer_trace;` never leads to the creation of a current_trace.

However your solution will allow procedures and functions that contain `information_schema.optimizer_trace` to slip through, and only when `Opt_trace_start` is being destructed do you ensure that if `information_schema.optimizer_trace` was referenced, then the `current_trace` is not saved.

Is it possible to prevent `current_trace` from being created for stored procs or functions or anything else that hides `information_schema.optimizer_trace` access at the outset in `Opt_trace_start::init`instead of when the `Opt_trace_start` is being destructed? Is it possible for example to search the parser tokens for `information_schema.optimizer_trace` as an additional step if the call to `list_has_optimizer_trace_table(tables_list))` returns false within the function `Opt_trace_start::init`.

PR #3046, opened 2024-02-06
[MDEV-33383] fts query calculate ranking crashes

tiandiwonder:
Problem:
========
Run query in customer's instance:
SELECT `tid` FROM `thread_search` WHERE MATCH (`title`,`sub_content`) AGAINST ('****' IN BOOLEAN MODE);

It crashes in fts_query_calculate_ranking():
  ret = rbt_search(word_freq->doc_freqs, &parent, &ranking->doc_id);

  /* It must exist. */
  ut_a(ret == 0);

It means that the doc id should exist in doc_freqs, which contains all doc ids having the same word.

Analysis:
=========
The doc id 1675241735 has been inserted into the rbtree(rbt_node_add()), but after some following inserts(3839877411), it could not be searched by rbt_search(). But It does exist in the rbtree(iterated by rbt_next()). Obviously the rbtree is corrupted

The root cause is that the compare funciton fts_ranking_doc_id_cmp() returns the wrong result. So the rbtree is not working as a sorted binary tree.

It's totally wrong to compare two integers as below, no matter signed or unsigned.
return((int)(rk1->doc_id - rk2->doc_id));

Solution:
=========
The solution is pretty simple: just use > < == to compare two doc ids

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/fts/fts0que.cc
+++ b/storage/innobase/fts/fts0que.cc
@@ -398,7 +398,7 @@ fts_freq_doc_id_cmp(
 	const fts_doc_freq_t*	fq1 = (const fts_doc_freq_t*) p1;
 	const fts_doc_freq_t*	fq2 = (const fts_doc_freq_t*) p2;
 
-	return((int) (fq1->doc_id - fq2->doc_id));
+	return(fts_doc_id_cmp(&fq1->doc_id, &fq2->doc_id));
 }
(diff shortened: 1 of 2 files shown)


dr-m:
This is superceded by #3047.

PR #3047, opened 2024-02-06
MDEV-33383 fts query calculate ranking crashes

tiandiwonder:
Problem:
========
Run query in customer's instance:
SELECT `tid` FROM `thread_search` WHERE MATCH (`title`,`sub_content`) AGAINST ('****' IN BOOLEAN MODE);

It crashes in fts_query_calculate_ranking():
  ret = rbt_search(word_freq->doc_freqs, &parent, &ranking->doc_id);

  /* It must exist. */
  ut_a(ret == 0);

It means that the doc id should exist in doc_freqs, which contains all doc ids having the same word.

Analysis:
=========
The doc id 1675241735 has been inserted into the rbtree(rbt_node_add()), but after some following inserts(3839877411), it could not be searched by rbt_search(). But It does exist in the rbtree(iterated by rbt_next()). Obviously the rbtree is corrupted

The root cause is that the compare funciton fts_ranking_doc_id_cmp() returns the wrong result. So the rbtree is not working as a sorted binary tree.

It's totally wrong to compare two integers as below, no matter signed or unsigned.
return((int)(rk1->doc_id - rk2->doc_id));

Solution:
=========
The solution is pretty simple: just use > < == to compare two doc ids

## Description
TODO: fill description here

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/include/fts0types.inl
+++ b/storage/innobase/include/fts0types.inl
@@ -59,7 +59,7 @@ fts_trx_row_doc_id_cmp(
 	const fts_trx_row_t*	tr1 = (const fts_trx_row_t*) p1;
 	const fts_trx_row_t*	tr2 = (const fts_trx_row_t*) p2;
 
-	return((int)(tr1->doc_id - tr2->doc_id));
+	return(fts_doc_id_cmp(&tr1->doc_id, &tr2->doc_id));
 }
 
 /******************************************************************//**
@@ -75,7 +75,7 @@ fts_ranking_doc_id_cmp(
 	const fts_ranking_t*	rk1 = (const fts_ranking_t*) p1;
 	const fts_ranking_t*	rk2 = (const fts_ranking_t*) p2;
 
-	return((int)(rk1->doc_id - rk2->doc_id));
+	return(fts_doc_id_cmp(&rk1->doc_id, &rk2->doc_id));
 }
 
 /******************************************************************//**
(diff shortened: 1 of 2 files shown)


storage/innobase/include/fts0types.inl
```diff
@@ -75,22 +75,28 @@ fts_ranking_doc_id_cmp(
...
 	const doc_id_t*	up2 = static_cast<const doc_id_t*>(p2);
 
-	return static_cast<int>(*up1 - *up2);
+	if (*up1 < *up2) {
+		return (-1);
+	} else if (*up1 > *up2) {
+		return (1);
+	} else {
+		return (0);
+	}
```
dr-m, line 99: I was curious about the generated code, so I tested the following on https://godbolt.org:
```c++
int cmp(int a) { return a < 0 ? -1 : a > 0; }
int cmp2(unsigned long long a, unsigned long long b) { if (a < b) return -1; if (a > b) return 1; return 0; }
```
Up to GCC 7, for the unary function `cmp` the compiler would generate conditional branches on IA-32, AMD64, ARMv7 and ARMv8. Starting with GCC 8, it would generate some conditional instructions.

For the binary function `cmp2` the code is rather acceptable already on GCC 4.8.5 for AMD64, which is the oldest that we currently support:
```asm
        xor     edx, edx
        cmp     rdi, rsi
        mov     eax, -1
        seta    dl
        cmovnb  eax, edx
        ret
```
On ARMv7 as well as ARMv8, no conditional branches were emitted. On POWER, I guess that there are no conditional ALU instructions, because whichever compiler version I tested, there would be conditional branches.

dr-m (commented):
Thank you for the patch. I would like to experiment a bit, to see how big a change it would be to replace all the `rbt_` with `std::map` and the `ib_vector_sort()` with `std::sort()`.

storage/innobase/include/fts0types.inl
dr-m: I checked this a little deeper. There are quite a few references to `ib_rbt_node_t`, and there is a `fts_cache_t::total_size` that will have to be maintained. It might not be easy to replace this data structure with the C++11 `std::multimap` while retaining that bookkeeping.

What we can easily do is to remove `ib_vector_sort` and `fts_doc_id_cmp` (30dbb1ad36cce9752d597484b68bf91ac61ef1c3). The generated code for my replacement `fts_doc_ids_sort` as well as its callees looked reasonable to me.

dr-m:
I created #3048 with my version of this. I did not add a unit test, because I think that the new comparison predicate (which is equivalent with what you wrote) is obvious. A unit test could have made sense for a comparison predicate that would have used some bitwise arithmetics to implement a saturating conversion to a narrower type.

dr-m:
This was fixed a bit more extensively in #3048. Thank you for providing the data for the unit test!

PR #3051, opened 2024-02-07
MDEV-35993 - 🐛 Fixed configure for Xcode, CMake generate

saltpi:

## Description

Execute command `cmake . -G Xcode`

with error information:
```
CMake Error in sql/CMakeLists.txt:
  The custom command generating

    /Volumes/Code/MariaDB/sql/yy_mariadb.yy

  is attached to multiple targets:

    gen_lex_token
    dist

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".
```

## How can this PR be tested?

Execute command `cmake . -G Xcode`

## PR quality check


vaintroub:
I think it is a backwards dependency. nothing should depend on "dist"

svoj (changes requested):
We should have good commit comment as described in https://github.com/MariaDB/server/blob/main/CODING_STANDARDS.md. It should describe the problem and how it was fixed.

Also please remove merge commits, use rebase instead.

PR #3054, opened 2024-02-08
MDEV-14357: rpl.rpl_domain_id_filter_io_crash failed in buildbot with…

bnestere:
… wrong result

A race condition with the SQL thread, where depending on if it was killed before or after it had executed the fake/generated IGN_GTIDS Gtid_list_log_event, may or may not update gtid_slave_pos with the position of the ignored events. Then, the slave would be restarted while resetting IGNORE_DOMAIN_IDS to be empty, which would result in the slave requesting different starting locations, depending on whether or not gtid_slave_pos was updated. And, because previously ignored events could now be requested and executed (no longer ignored), their presence would fail the test.

This patch fixes this in two ways. First, to use GTID positions for synchronization rather than binlog file positions. Then second, to synchronize the SQL thread’s gtid_slave_pos with the ignored events before killing the SQL thread.

To consistently reproduce the test failure, the following patch can be applied:
```
diff --git a/sql/log_event_server.cc b/sql/log_event_server.cc index f51f5b7deec..de62233acff 100644
--- a/sql/log_event_server.cc
+++ b/sql/log_event_server.cc
@@ -3686,6 +3686,12 @@ Gtid_list_log_event::do_apply_event(rpl_group_info *rgi)
     void *hton= NULL;
     uint32 i;

+    sleep(1);
+    if (rli->sql_driver_thd->killed || rli->abort_slave)
+    {
+      return 0;
+    }
+
```

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_domain_id_filter_io_crash.result
+++ b/mysql-test/suite/rpl/r/rpl_domain_id_filter_io_crash.result
@@ -11,6 +11,7 @@ SELECT * FROM t1;
 i
 1
 connection slave;
+include/save_master_gtid.inc
 connection slave;
 call mtr.add_suppression("Slave I/O: Relay log write failure: could not queue event from master.*");
 # Case 0 : Start slave with IGNORE_DOMAIN_IDS=(), then restart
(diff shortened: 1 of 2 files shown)


knielsen (approved):
Looks fine to me.

andrelkin (approved):

bnestere:
Pushed via a different branch: ee895583126f6991ae885b03c2b832b4dd8ab24c

PR #3058, opened 2024-02-13
MDEV-27952 Improve error log messages with descriptive text already present in the code

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-27952*

## Description

…
This is explicitly for the test case:

CREATE TABLE t (c INT) ENGINE=InnoDB;
ALTER TABLE t DISCARD TABLESPACE;
ALTER TABLE t ADD INDEX (c);
SELECT * FROM t;

That had an error:
  [ERROR] Got error 194 when reading table './test/t'

194 is the code from HA_ERR_TABLESPACE_MISSING.

As these code are at the handler level, based on the existing code the table name is assumed. There is a built in function my_get_err_msg that translates the number to a format with its argument that can be assumed to be a table string.

The error is now:

2024-02-13 19:08:44 3 [ERROR] Tablespace is missing for table './test/t'

## Release Notes

Per title

## How can this PR be tested?

Manual test per MDEV

## PR quality check


grooverdan:
```
+    sql_print_error("Error %M", error, table->s->path.str);
```
Generates:
```
2024-02-13 19:49:55 4 [ERROR] Error 194 "Tablespace is missing for a table"
```
which isn't correct.

mysql-test/suite/archive/archive.result
```diff
@@ -1,5 +1,6 @@
 CALL mtr.add_suppression("Unsafe statement written to the binary log using statement format since BINLOG_FORMAT = STATEMENT");
 call mtr.add_suppression("Table 't1' is marked as crashed and should be repaired");
+call mtr.add_suppression("Table './test/t1' is marked as crashed and should be repaired");
```
dr-m, line 3: Can we edit the existing suppressions instead:
```sql
call mtr.add_suppression("Table '.*t1' is marked as crashed and should be repaired");
```

mysql-test/suite/innodb/t/innodb_corrupt_bit.test
```diff
@@ -8,6 +8,7 @@
 
 --disable_query_log
 call mtr.add_suppression("Flagged corruption of.* in table .* in .*");
+call mtr.add_suppression("Index ./test/corrupt_bit_test_@1s is corrupted");
```
dr-m, line 11: Could we convert the filename-safe table name to the user-specified name `test.corrupt_bit_test_ā` with something like this?
```c++
static void dict_table_open_failed(const table_name_t &name)
{
  my_printf_error(ER_TABLE_CORRUPT,
                  "Table %`.*s.%`s is corrupted."
                  " Please drop the table and recreate.",
                  MYF(ME_ERROR_LOG),
                  int(name.dblen()), name.m_name, name.basename());
}
```
Or would it require a different format variant to be implemented?

montywi: Please do not use &table_name_t ; Please use pointers instead.
If possible, use C syntax cast instead of C++ syntax cast.

montywi: Note that when using a table name for error messages inside InnoDB, you have to use conversion functions to change it from filename format to table name format.
You have to use innobase_convert_to_system_charset() when converting filenames to table names.

mysql-test/suite/archive/archive.test
```diff
@@ -6,6 +6,7 @@
 
 CALL mtr.add_suppression("Unsafe statement written to the binary log using statement format since BINLOG_FORMAT = STATEMENT");
 call mtr.add_suppression("Table 't1' is marked as crashed and should be repaired");
+call mtr.add_suppression("Table './test/t1' is marked as crashed and should be repaired");
```
montywi, line 9: Please ensure that we do not use '/' in the suppression.  (Will not work on windows)

storage/archive/ha_archive.cc
```diff
@@ -287,7 +287,7 @@ static int archive_discover(handlerton *hton, THD* thd, TABLE_SHARE *share)
   if (!(azopen(&frm_stream, az_file, O_RDONLY|O_BINARY)))
   {
     if (errno == EROFS || errno == EACCES)
-      DBUG_RETURN(my_errno= errno);
+      DBUG_RETURN(HA_ERR_TABLE_READONLY);
```
montywi, line 290: You MUST set my_errno here!

storage/archive/ha_archive.cc
```diff
@@ -990,7 +990,7 @@ int ha_archive::write_row(const uchar *buf)
 
   if (!share->archive_write_open && share->init_archive_writer())
   {
-    rc= errno;
+    rc= HA_ERR_CRASHED_ON_USAGE;
```
montywi, line 993: Please don't change errno to HA_ERR_CRASHED_ON_USAGE.
This will hide the original error !

storage/archive/ha_archive.cc
montywi: Please don't change errno to HA_ERR_CRASHED_ON_USAGE.
This will hide the original error !

storage/archive/ha_archive.cc
```diff
@@ -1228,7 +1228,7 @@ int ha_archive::rnd_init(bool scan)
       DBUG_RETURN(HA_ERR_CRASHED_ON_USAGE);
 
   if (init_archive_reader())
-      DBUG_RETURN(errno);
+      DBUG_RETURN(HA_ERR_CRASHED_ON_USAGE);
```
montywi, line 1231: Please don't change errno to HA_ERR_CRASHED_ON_USAGE.
This will hide the original error !

storage/archive/ha_archive.cc
```diff
@@ -1556,7 +1556,7 @@ int ha_archive::optimize(THD* thd, HA_CHECK_OPT* check_opt)
   if (init_archive_reader())
   {
     mysql_mutex_unlock(&share->mutex);
-    DBUG_RETURN(errno);
+    DBUG_RETURN(HA_ERR_CRASHED_ON_USAGE);
```
montywi, line 1559: Please don't change errno to HA_ERR_CRASHED_ON_USAGE.
This will hide the original error !

storage/archive/ha_archive.cc
```diff
@@ -1771,7 +1777,7 @@ int ha_archive::info(uint flag)
   if (flag & HA_STATUS_AUTO)
   {
     if (init_archive_reader())
-      DBUG_RETURN(errno);
+      DBUG_RETURN(HA_ERR_CRASHED_ON_USAGE);
```
montywi, line 1780: Please don't change errno to HA_ERR_CRASHED_ON_USAGE.
This will hide the original error !

storage/archive/ha_archive.cc
montywi: Instead of changing errno, you can instead change handler::print_error() to give more information about the error

PR #3060, opened 2024-02-13
MDEV-33454 release row locks for non-modified rows at XA PREPARE

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33454*
## Description
From the correctness point of view, it should be safe to release all locks on index records that were not modified by the transaction. Doing so should make the locks after `XA PREPARE` fully compatible with what would happen if the server were restarted: InnoDB table IX locks and exclusive record locks would be resurrected based on undo log records.

Concurrently running transactions that are waiting for a lock may invoke `lock_rec_convert_impl_to_expl()` to create an explicit record lock object on behalf of the lock-owning transaction so that they can attaching their waiting lock request on the explicit record lock object. Explicit locks would be released by `trx_t::release_locks()` during commit or rollback.

Any clustered index record whose `DB_TRX_ID` belongs to a transaction that is in active or `XA PREPARE` state will be implicitly locked by that transaction. On `XA PREPARE`, we can release explicit exclusive locks on records whose `DB_TRX_ID` does not match the current transaction identifier.

`lock_rec_unlock_unmodified()`: Release record locks that are not implicitly held by the current transaction.

`lock_release_on_prepare_try()`, `lock_release_on_prepare()`: Invoke `lock_rec_unlock_unmodified()`.

`row_trx_id_offset()`: Declare non-`static`.

`lock_rec_unlock()`: Replaces `lock_rec_unlock_supremum()`.

## Release Notes
Reduce the chances of lock contention in replication.

## How can this PR be tested?
This is best tested by someone who really knows replication and bugs like MDEV-32020 and MDEV-31949.

This is a bug fix that could be applicable also to earlier versions, but the InnoDB locking was heavily refactored in MariaDB Server 10.6, which is the version that we are targeting.

## PR quality check

### DIFF
--- a/storage/innobase/lock/lock0lock.cc
+++ b/storage/innobase/lock/lock0lock.cc
@@ -4182,7 +4182,7 @@ restart:
   ulint count= 1000;
   /* We will not attempt hardware lock elision (memory transaction)
   here. Both lock_rec_dequeue_from_page() and lock_table_dequeue()
-  would likely lead to a memory transaction due to a system call, to
+  would likely lead to a memory transaction abort due to a system call, to
   wake up a waiting transaction. */
   lock_sys.rd_lock(SRW_LOCK_CALL);
   trx->mutex_lock();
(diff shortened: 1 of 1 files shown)


storage/innobase/lock/lock0lock.cc
```diff
@@ -4466,6 +4532,20 @@ void lock_release_on_prepare(trx_t *trx)
     if (lock_release_on_prepare_try(trx))
       return;
 
+  mtr_t mtr;
+  rec_offs offsets_[REC_OFFS_NORMAL_SIZE];
+  rec_offs *offsets= offsets_;
+  mem_heap_t *heap= nullptr;
+  if (false)
```
vlad-lesin, line 4539: Why do we need this? func_exit label is used only at the end of the function. Why just don't move the code to the end of the function?

vlad-lesin (commented):
The code looks good to me except that "if (false)" block in lock_release_on_prepare().

storage/innobase/lock/lock0lock.cc
dr-m: I did not want to change the indentation of the rest of the function to ensure that `LockMutexGuard` is released before the conditional call to `mem_heap_free(heap)`. At the end when we do `goto func_exit` to this `if (false)` block, `LockMutexGuard g` will have been destroyed and the latch released.

vlad-lesin: Maybe we could add comment about LockMutexGuard destructor call before "if (false)" condition. Otherwise it looks confusing.

vlad-lesin (approved):

storage/innobase/lock/lock0lock.cc
dr-m: I could also just add those `{` and `}` and reindent the code block.

PR #3061, opened 2024-02-13
MDEV-33363 CI failure: innodb.import_corrupted: Assertion failed: old…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33363*

## Description
This regression is introduced in MDEV-28708 where the MTR_LOG_NO_REDO mtrs are assigned last_checkpoint_lsn as the LSN. It causes a race with checkpoint in pending state. The concurrent checkpoint writes a checkpoint LSN of larger value after pages with older checkpoint LSN is inserted into the flush list. The next checkpoint sees the reversal in checkpoint sequence and asserts if the pages are not yet flushed.

There could be several ways to solve this issue. Ideally the unlogged mtr should take the latest LSN as opposed to going behind and use the previous checkpoint LSN. It has been the older design and seems good. Also, other than the critical race, using the old checkpoint LSN adds the pages to other end of flush list overriding all existing dirty pages and looks counter intuitive.

## Release Notes
None.

## How can this PR be tested?

./mtr innodb.import_corrupted: This test fails rarely in CI without the fix.
./mtr innodb.import_checkpoint_race: The test code can be found attached in MDEV-33363. The crash can be repeated with this test 50% of the time. The test in rare case could experience debug sync hang and thus is not added part of the patch.

## PR quality check

### DIFF
--- a/storage/innobase/mtr/mtr0mtr.cc
+++ b/storage/innobase/mtr/mtr0mtr.cc
@@ -234,7 +234,14 @@ static void insert_imported(buf_block_t *block)
   if (block->page.oldest_modification() <= 1)
   {
     log_sys.latch.rd_lock(SRW_LOCK_CALL);
-    const lsn_t lsn= log_sys.last_checkpoint_lsn;
+    /* For unlogged mtrs (MTR_LOG_NO_REDO), we use the current system LSN. The
+    mtr that generated the LSN is either already committed or in mtr_t::commit.
+    Shared latch and relaxed atomics should be fine here as it is guaranteed
+    that both the current mtr and the mtr that generated the LSN would have
+    added the dirty pages to flush list before we access the minimum LSN during
+    checkpoint. log_checkpoint_low() acquires exclusive log_sys.latch before
+    commencing. */
+    const lsn_t lsn= log_sys.get_lsn();
     mysql_mutex_lock(&buf_pool.flush_list_mutex);
     buf_pool.insert_into_flush_list
       (buf_pool.prepare_insert_into_flush_list(lsn), block, lsn);


mysql-test/suite/innodb/r/import_checkpoint_race.result
```diff
@@ -0,0 +1,60 @@
...
+connect  purge_control,localhost,root,,;
+START TRANSACTION WITH CONSISTENT SNAPSHOT;
+connection default;
+DELETE FROM t1 WHERE id % 2 = 1;
+FLUSH TABLES t1 FOR EXPORT;
+UNLOCK TABLES;
+connection purge_control;
+COMMIT;
+DROP TABLE t1;
+# First(Current) Checkpoint LNS: X
```
dr-m, line 28: Typo: LSN

storage/innobase/mtr/mtr0mtr.cc
```diff
@@ -234,7 +234,13 @@ static void insert_imported(buf_block_t *block)
...
   {
     log_sys.latch.rd_lock(SRW_LOCK_CALL);
-    const lsn_t lsn= log_sys.last_checkpoint_lsn;
+    /* For unlogged mtrs (MTR_LOG_NO_REDO), we use the current system LSN. The
+    mtr that generated the LSN is either already committed or in mtr_t::commit.
+    Shared latch and relaxed atomics should be fine here as it is guaranteed
+    that both the current mtr and the mtr that generated the LSN would have
+    added the dirty pages to flush list before we access the minimum LSN during
+    checkpoint. */
+    const lsn_t lsn= log_sys.get_lsn();
```
dr-m, line 243: Thank you. This is the actual fix; the rest is `DEBUG_SYNC` instrumentation for the test. In retrospect, this corresponds to what was done in 89a5a8d234832ef9ed5ee814e4db42c636fcde1e to fix a similar bug in 10.5 and 10.6.

The commit might explicitly mention that `log_checkpoint_low()` is covered by an exclusive `log_sys.latch`, which would prevent any concurrent `mtr_t::commit()`.

dr-m (approved):
I see that the regression test needs some more work to be stable in all debug build environments. It would be great if you can improve it. If it is too much trouble, then it is OK to omit the test and the instrumentation.

mysql-test/suite/innodb/r/import_checkpoint_race.result
mariadb-DebarunBanerjee: Removed the test case.

storage/innobase/mtr/mtr0mtr.cc
mariadb-DebarunBanerjee: ok, done.

PR #3063, opened 2024-02-14
MDEV-32061 INTO OUTFILE should work with named pipes

XanC:
*The Jira issue number for this PR is: MDEV-32061*

## Description

This patch allows named pipes (aka fifos) to be used as destinations for INTO OUTFILE statements.

In my case, the time and disk space required to make compressed snapshots of database tables are hugely reduced by this capability.  There are probably many other creative uses of named pipes for INTO OUTFILE.

## Release Notes

Allow INTO OUTFILE to write to named pipes.

https://mariadb.com/kb/en/select-into-outfile/ will need some updates.

## Testing

An automated test for this feature is included in this pull request.

- [x ] *This is a new feature and the PR is based against the latest MariaDB development branch.*

## PR quality check


mysql-test/main/outfile_fifo.test
```diff
@@ -0,0 +1,17 @@
+#
```
grooverdan, line 1: start with:
`--source include/not_windows.inc`

mysql-test/main/outfile_fifo.test
```diff
@@ -0,0 +1,17 @@
...
+# MDEV-32061: Allow INTO OUTFILE to writo to a fifo
```
grooverdan, line 2: typo + convention is to `--echo # MDEV....`

mysql-test/main/outfile_fifo.test
```diff
@@ -0,0 +1,17 @@
...
+#
+--disable_warnings
+DROP TABLE IF EXISTS t1;
+--enable_warnings
```
grooverdan, line 6: unnecessary - test environment ensure its clean. Remove 3 lines.

mysql-test/main/outfile_fifo.test
```diff
@@ -0,0 +1,17 @@
...
+CREATE TABLE t1 (val1 VARCHAR(255), val2 VARCHAR(255));
+INSERT INTO t1 VALUES('hello', 'there');
+SELECT * FROM t1;
+SELECT * INTO OUTFILE '../../tmp/fifo' FROM t1;
+TRUNCATE TABLE t1;
+SELECT SLEEP(.25); #To make sure the background "cat" process has a chance to read from the fifo and write to the text file
```
grooverdan, line 12: Sleep statements are a bit fragile:
concept from `mysql-test/include/wait_until_connected_again.inc`. I haven't looked up the error code.
```
--disable_result_log
--disable_query_log
let $counter= 100;
let $mysql_errno= 9999;
while ($mysql_errno)
{
  --error 0,ER_NO_SUCH_FILE
  LOAD DATA INFILE '../../tmp/fifo_data.txt' INTO TABLE t1;
 
  dec $counter;
  if (!$counter)
  {
    --die Out of time for file to appear
  }
    if ($mysql_errno > 0)
  {
    --sleep 0.1
  }
}
--enable_query_log
--enable_result_log
```

sql/sql_class.cc
```diff
@@ -3258,31 +3260,43 @@ static File create_file(THD *thd, char *path, sql_exchange *exchange,
...
+                               path, O_WRONLY, MYF(MY_WME))) < 0)
+      return file;
   }
-  /* Create the file world readable */
-  if ((file= mysql_file_create(key_select_to_file,
-                               path, 0644, O_WRONLY|O_EXCL, MYF(MY_WME))) < 0)
-    return file;
+  else
+  {
+    if (!access(path, F_OK))
```
grooverdan, line 3272: with `my_stat` done, `access` is redundant right?  Can this be `errno != 0`?

sql/sql_class.cc
```diff
@@ -3258,31 +3260,43 @@ static File create_file(THD *thd, char *path, sql_exchange *exchange,
...
 #endif
+  }
+  
   if (init_io_cache(cache, file, 0L, WRITE_CACHE, 0L, 1, MYF(MY_WME)))
   {
     mysql_file_close(file, MYF(0));
-    /* Delete file on error, it was just created */
-    mysql_file_delete(key_select_to_file, path, MYF(0));
+    /* Delete file on error if it was just created */
+    if (file_was_created_here)
```
grooverdan, line 3293: good.

sql/sql_class.cc
```diff
@@ -3258,31 +3260,43 @@ static File create_file(THD *thd, char *path, sql_exchange *exchange,
...
+      mysql_file_delete(key_select_to_file, path, MYF(0));
     return -1;
   }
   return file;
 }
```
grooverdan, line 3299: don't change unnecessary lines. The convention is 2 lines between functions.

grooverdan (changes requested):
Well done.

Mainly an improvement around the test.

mysql-test/main/outfile_fifo.test
XanC: grooverdan, thanks for all the feedback.  I'll be making changes tomorrow.

This particular one I'm not sure will work, because the `cat` command creates the output file immediately.  It sits empty until the data arrives in the fifo, at which time it starts being written to the file.

We might need to wait until the file is non-empty.  Even that assumes that the file will be written atomically and we won't catch it in progress, which is /probably/ the case.

Even if we find a way to test for non-empty, I'm not sure it's better than just sleeping for plenty of time.

XanC: Maybe we could loop until observing the same non-zero file size twice.  That should be sufficient.

grooverdan: yep. that sounds all right.

mysql-test/main/outfile_fifo.test
XanC: Done.  I also added an ifdef to make it so the actual code change only runs on non-Windows as well.

mysql-test/main/outfile_fifo.test
XanC: Done.

mysql-test/main/outfile_fifo.test
XanC: Done.

mysql-test/main/outfile_fifo.test
XanC: This was really tricky, because I couldn't find any way to get the size of a file, or really to get anything about a file into a variable rather than printing its contents.

The trick ended up being to put " && touch done.txt" into the shell script.  Then I could detect the presence of done.txt.

I wasn't able to find a nice name for error 13, though.  Maybe it exists.

sql/sql_class.cc
XanC: Sorry, that was accidental.  Restored.

mysql-test/main/outfile_fifo-master.sh
```diff
@@ -0,0 +1,5 @@
+if [ -e $MYSQLTEST_VARDIR/tmp/fifo ]; then
+    rm $MYSQLTEST_VARDIR/tmp/fifo
+fi
+mkfifo $MYSQLTEST_VARDIR/tmp/fifo
+cat $MYSQLTEST_VARDIR/tmp/fifo > $MYSQLTEST_VARDIR/tmp/fifo_data.txt && touch $MYSQLTEST_VARDIR/tmp/done.txt &
```
grooverdan, line 5:
```suggestion
if [ -e "$MYSQL_TMP_DIR"/fifo ]; then
    rm "$MYSQL_TMP_DIR"/fifo
fi
if [ -e "$MYSQL_TMP_DIR"/done.txt ]; then
    rm "$MYSQL_TMP_DIR"/done.txt
fi
mkfifo "$MYSQL_TMP_DIR"/fifo
cat "$MYSQL_TMP_DIR"/fifo > "$MYSQL_TMP_DIR"/fifo_data.txt && touch "$MYSQL_TMP_DIR"/done.txt &
```

Using MYSQL_TMP_DIR and removing done.txt

mysql-test/main/outfile_fifo.test
```diff
@@ -0,0 +1,44 @@
+--source include/not_windows.inc
+--echo #
+--echo # MDEV-32061: Allow INTO OUTFILE to writo to a fifo
```
grooverdan, line 3:
```suggestion
--echo # MDEV-32061: Allow INTO OUTFILE to write to a fifo
```

mysql-test/main/outfile_fifo.test
```diff
@@ -0,0 +1,44 @@
...
+SELECT * FROM t1;
+SELECT * INTO OUTFILE '../../tmp/fifo' FROM t1;
+TRUNCATE TABLE t1;
+
+# Let's make sure that `cat` which was started by the .sh script has
+# had a chance to complete.
+
+--disable_result_log
+--disable_query_log
+let $counter= 1000;
```
grooverdan, line 18:
```suggestion
let $counter= 600;
while ($counter)
{
--error 0,1
--file_exists $MYSQL_TMP_DIR/done.txt
  if ($errno)
  {
    dec $counter;
    real_sleep 0.1;
  }
  if (!$errno)
  {
    let $counter= 0;
  }
}
--file_exists $MYSQL_TMP_DIR/done.txt
```
The last file_exists will error if its not there. Otherwise use mysqltest bits to ensure the file exists.

sql/sql_class.cc
```diff
@@ -3258,25 +3261,45 @@ static File create_file(THD *thd, char *path, sql_exchange *exchange,
     return -1;
   }
 
-  if (!access(path, F_OK))
+  file_exists_already= my_stat(path, &file_stat, MYF(0));
+
+/* Probably a fifo could be used under Windows, but we're not
```
grooverdan, line 3266:
```suggestion
/*
  Probably a fifo could be used under Windows, but we're not
  attempting it at the moment.
*/
```
Its odd, but consistent with existing comment format.

grooverdan (changes requested):
good idea with the done.txt. Just a bit more to complete the transition.

grooverdan:
Ok squashed it up and used `--replace_result` for test. Not sure why you added secure-priv - its the default and not needed, so removed.

Sorry the github suggestions weren't as easily apply-able, first time I've used them.

I'm chasing up why the --ps-protocol tests are failing (its executing SELECT queries twice).

XanC:
Thanks very much!

`--replace_result` is definitely the better approach.  Is there anywhere that the different test "commands" are documented?

At one point I was getting a failure to write to a file because of secure-priv, and setting it to the empty string got around the error.  I'm not sure how that would have happened if it were the default as you say, so that's a mystery.  I was definitely chasing my tail for a while trying to get the paths to line up correctly between the SQL commands and the shell/test commands.

Well anyway, thanks for pushing this out.  Would it also be appropriate for me to attempt to apply this change to 10.11?

grooverdan (approved):

grooverdan:
disabled the ps2-protocol around the critical bit that wouldn't execute twice.

> `--replace_result` is definitely the better approach.

> Is there anywhere that the different test "commands" are documented?

I had a look, couldn't find a good reference apart from the client/mysqltest.cc code and the existing test cases. :grimacing:  should be a better answer sorry.

> Well anyway, thanks for pushing this out. Would it also be appropriate for me to attempt to apply this change to 10.11?

Strictly is a new feature so its outside the engineering guidelines to backport. Currently plan is to have a new LTS release out later this year.

grooverdan:
secure-file-priv from the CI did need to be set. Not sure why. After a second review we'll merge this in.

XanC:
Well, at least I'm not crazy!

XanC:
Is there anything else I should be doing with this?

vaintroub:
There is a problem with it, in that those writes are not killable. It is not hard to imagine someone writes more than PIPE_BUF into FIFO, but nobody is reading from it.  It hangs, the connection is not killable, query timeouts do not work, it can't release locks that it might have like global locks (FTWRL), or table locks, user locks, and row locks. It can't finish transaction.

So this writes must be killable, or at least maybe timeout-able. We have nothing like this in mysys at the moment.

Also, this Windows exclusion is  superficial. TBH, currently the our code can't even open named pipe because the name gets translated from  _\\.\pipe\something_   in a way that it loses the third character, the dot 🤦 , thus the name becomes some UNC network name, but this is for another day. Important is that we should not be excluding Windows out of superstition, this is not very sound.

vaintroub:
One way to go could be (On Linux) non-blocking  writes, then wait in a poll() for fd to be writable, with another fd used as signal for the interrupt. another hack (on Linux) is pthread_kill, and look for "killed" indicator on EINTR, not sure how good it works.  Windows has cancellable IO, so given file handle, one can interrupt its IOs.

Note that with regular files we do not have to worry about such stiff, all IOs finish rather quicky

XanC:
@vaintroub, thank you for the feedback.  Killing the write is a situation I had not considered.  I will (as time permits!) experiment with the mechanisms you suggested for making the writes killable.

I'm a bit confused as to what your position is on Windows support.  I didn't originally include Windows support because I don't have much Windows infrastructure for development and testing, and I don't happen to need this feature on Windows.  If your position is not to include this feature until it also works on Windows, that's a very reasonable position.

You go on to give a much better reason that the feature won't work on Windows: currently, the code can't even open the named pipe because of some kind of path munging.

Unless I'm missing what you mean by "superstition", neither of those reasons seem to be superstitious.

Are you willing to merge this feature as a Linux-only feature, or do you require Windows support too?

Thank you!

vaintroub:
What I mean with Windows is that -  if you do not know if something will work or not, just do not add #ifndef _WIN32. If you'll disable it, we'll never know that \\.\pipe gets converted to network share in some my_convert_dir. So please so not add ifdefs and ifndefs, just in case, if you do not know exactly and whether or not you broke something. I understand that maybe, for you adding a test case is challenging. But do not worry, you won't be adding that test case, its ok when it compiles.

vuvova:
Where is it now? to pass review this just needs `#ifndef _WIN32` removed?

gkodinov:
This has been idle for close to two years now, since the last comment by the reviewer. Closing due to inactivity. If you intend to keep working on this, please ask me (or anybody else from the mariadb committers really) to re-open the pull-request. Or, well, just file a new one and reference this one from it.

PR #3065, opened 2024-02-15
MDEV-33464 Crash when innodb_max_undo_log_size is set to innodb_page_size*4294967296

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33464*
## Description
`purge_sys_t::truncating_tablespace()`: Clamp the
`innodb_max_undo_log_size` to the maximum number of pages before converting the result into a 32-bit unsigned integer.

This fixes up commit f8c88d905b44bffe161158309e9acc25ad3691aa (MDEV-33213).

In later major versions, we would use 32-bit unsigned integer here due to commit ca501ffb04246dcaa1f1d433d916d8436e30602e and the code would crash also on 64-bit processors.

## Release Notes
https://mariadb.com/kb/en/innodb-purge/ seems to be covering this area. I am not sure if this bug deserves to be mentioned there.

For 10.11 and later: Setting innodb_max_undo_log_size to something larger than innodb_page_size*4294967295 would 
cause unexpected behavior or crashes when setting innodb_undo_log_truncate=ON and using innodb_undo_tablespaces>1.

For 10.6: The description needs to be qualified with "On 32-bit systems."

## How can this PR be tested?
This case is covered by the test `innodb.undo_truncate`. The modified test would crash on 32-bit systems when the fix is not present.

## PR quality check

### DIFF
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -669,7 +669,9 @@ fil_space_t *purge_sys_t::truncating_tablespace()
   if (space || srv_undo_tablespaces_active < 2 || !srv_undo_log_truncate)
     return space;
 
-  const ulint size= ulint(srv_max_undo_log_size >> srv_page_size_shift);
+  const uint32_t size=
+    uint32_t(std::min(ulonglong{std::numeric_limits<uint32_t>::max()},
+                      srv_max_undo_log_size >> srv_page_size_shift));
   for (ulint i= truncate_undo_space.last, j= i;; )
   {
     if (fil_space_t *s= undo_truncate_try(srv_undo_space_id_start + i, size))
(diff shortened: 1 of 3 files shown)


storage/innobase/trx/trx0purge.cc
```diff
@@ -669,7 +669,9 @@ fil_space_t *purge_sys_t::truncating_tablespace()
   if (space || srv_undo_tablespaces_active < 2 || !srv_undo_log_truncate)
     return space;
 
-  const ulint size= ulint(srv_max_undo_log_size >> srv_page_size_shift);
+  const uint32_t size=
+    uint32_t(std::min((1ULL << 32) - 1,
```
mariadb-DebarunBanerjee, line 673: Can we can use c++11 std::numeric_limits<uint32_t>::max() ?

mariadb-DebarunBanerjee (approved):
looks good to me.

storage/innobase/trx/trx0purge.cc
dr-m: With a `ulonglong{}` widening cast (or an explicitly specified `std::min<ulonglong>`) we can: 53c6c823dc7cafefffdc93c79661cfb146ff8641.

PR #3067, opened 2024-02-15
MDEV-32898 Phantom rows caused by updates of PRIMARY KEY

dr-m:
- [x] *The Jira issue number for this PR is: [MDEV-32898](https://jira.mariadb.org/browse/MDEV-32898), [MDEV-26642](https://jira.mariadb.org/browse/MDEV-26642), [MDEV-26643](https://jira.mariadb.org/browse/MDEV-26643)*
## Description
https://jepsen.io/analyses/mysql-8.0.34 highlights that the transaction isolation levels in the InnoDB storage engine do not correspond to any widely accepted definitions, such as "Generalized Isolation Level Definitions" https://pmg.csail.mit.edu/papers/icde00.pdf (PL-1 = `READ UNCOMMITTED`, PL-2 = `READ COMMITTED`, PL-2.99 = `REPEATABLE READ`, PL-3 = `SERIALIZABLE`). Only `READ UNCOMMITTED` in InnoDB seems to match the above definition.

The issue is that InnoDB does not detect write/write conflicts (Section 4.4.3, Definition 6) in the above.

It appears that as soon as we implement write/write conflict detection (`SET SESSION innodb_snapshot_isolation=ON`), the default isolation level (`SET TRANSACTION ISOLATION LEVEL REPEATABLE READ`) will become Snapshot Isolation (similar to Postgres), as defined in Section 4.2 of [*A Critique of ANSI SQL Isolation Levels*, MSR-TR-95-51, June 1995](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tr-95-51.pdf).

Locking reads inside InnoDB used to read the latest committed version, ignoring what should actually be visible to the transaction. The added test `innodb.lock_isolation` illustrates this. The statement
```sql
UPDATE t SET a=3 WHERE b=2;
```
is executed in a transaction that was started before a read view or a snapshot of the current transaction was created, and committed before the current transaction attempts to execute
```sql
UPDATE t SET b=3;
```
If `SET innodb_snapshot_isolation=ON` is in effect when the second transaction was started, the second transaction will be aborted with the error `ER_CHECKREAD`. By default (`innodb_snapshot_isolation=OFF`), the second transaction would execute inconsistently, displaying an incorrect `SELECT COUNT(*) FROM t` in its read view.

If `innodb_snapshot_isolation=ON`, if an attempt to acquire a lock on a record that does not exist in the current read view is made, an error `DB_RECORD_CHANGED` (`HA_ERR_RECORD_CHANGED`, `ER_CHECKREAD`) will be raised. This error will be treated in the same way as a deadlock: the transaction will be rolled back.

`lock_clust_rec_read_check_and_lock()`: If the current transaction has a read view where the record is not visible and `innodb_snapshot_isolation=ON`, fail before trying to acquire the lock.

`row_sel_build_committed_vers_for_mysql()`: If `innodb_snapshot_isolation=ON`, disable the "semi-consistent read" logic that had been implemented by myself on the directions of Heikki Tuuri in order to address https://bugs.mysql.com/bug.php?id=3300 that was motivated by a customer wanting `UPDATE` to skip locked rows that do not match the `WHERE` condition. It looks like my changes were included in the MySQL 5.1.5 commit ad126d90e019f223470e73e1b2b528f9007c4532; at that time, employees of Innobase Oy (a recent acquisition of Oracle) had lost write access to the repository.

The only reason why we set `innodb_snapshot_isolation=OFF` by default is backward compatibility with applications, such as the one that motivated the implementation of "semi-consistent read" back in 2005. In a later major release, we can default to `innodb_snapshot_isolation=ON`.

Thanks to Peter Alvaro, Kyle Kingsbury and Alexey Gotsman for their work on https://github.com/jepsen-io/ and to Kyle and Alexey for explanations and some testing of this fix.

Thanks to Vladislav Lesin for the initial test for MDEV-26643.

## Release Notes
`SET innodb_snapshot_isolation=ON` (a new configuration parameter) makes it an error if an attempt to acquire a lock on a record that does not exist in the current read view. The error `DB_RECORD_CHANGED` (`HA_ERR_RECORD_CHANGED`, `ER_CHECKREAD`) would be raised. This error will be treated in the same way as a deadlock: the transaction will be rolled back. This will make the default `REPEATABLE READ` isolation level correspond to Snapshot Isolation, and the `SERIALIZABLE` and `READ COMMITTED` isolation levels to behave as expected.

## How can this PR be tested?
The changed behaviour is covered by the test `innodb.lock_isolation`.

This has been subjected to some testing with the Jepsen tool.

This bug is present since the beginning of InnoDB. We are targeting 10.6, because that is where the lock subsystem has already been heavily refactored compared to earlier releases.
## PR quality check


mysql-test/suite/sys_vars/r/slave_transaction_retry_errors.result
```diff
@@ -1,20 +1,20 @@
 select @@global.slave_transaction_retry_errors;
 @@global.slave_transaction_retry_errors
-1158,1159,1160,1161,1205,1213,1429,2013,12701,10,20,5000,400
+1158,1159,1160,1161,1205,1213,1020,1429,2013,12701,10,20,5000,400
```
dr-m, line 3: A similar change is also needed in the following tests:
```
Failing test(s): spider/bugfix.slave_transaction_retry_errors_5digit spider/feature.slave_transaction_retry_errors main.mysqld--help
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -881,6 +881,10 @@ static MYSQL_THDVAR_BOOL(table_locks, PLUGIN_VAR_OPCMDARG,
   /* check_func */ NULL, /* update_func */ NULL,
   /* default */ TRUE);
 
+static MYSQL_THDVAR_BOOL(strict_isolation, PLUGIN_VAR_OPCMDARG,
+  "Use strict transaction isolation.",
+  NULL, NULL, FALSE);
+
```
dr-m, line 887: I think that `innodb_snapshot_isolation` is a more descriptive name, because the aim is that with this fix (enabling write-write conflict resolution),
```sql
SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;
```
behaves like its Postgres counterpart, actually implementing Snapshot Isolation.

mysql-test/suite/sys_vars/r/slave_transaction_retry_errors.result
dr-m: 7899077bcbf8b4b09b5483dd0bbefca8daa5cc92 addresses this.

storage/innobase/handler/ha_innodb.cc
dr-m: 7899077bcbf8b4b09b5483dd0bbefca8daa5cc92 implements this.

vlad-lesin (approved):
There are some minor code review notes, but in common it looks good to me.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15376,7 +15385,7 @@ ha_innobase::check(
 	}
 
 	/* Restore the original isolation level */
-	m_prebuilt->trx->isolation_level = old_isolation_level;
+	m_prebuilt->trx->isolation_level = old_isolation_level & 3;
```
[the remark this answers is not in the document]
dr-m: The redundant bitmasking is needed for silencing `-Wconversion` in GCC 6 and later when assigning to a bit-field. Back when I enabled `-Wconversion` in f224525204dcd5a4f907cf9eccbba82ee7ea9018 I checked with https://godbolt.org that it does not incur any extra overhead. Upon reading that commit message, I realize that I’d better test this on the kvm-fulltest2 builder (which is not run for pull requests) to avoid the need of any fixups.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -4431,7 +4440,7 @@ innobase_start_trx_and_assign_read_view(
 	Do this only if transaction is using REPEATABLE READ isolation
 	level. */
 	trx->isolation_level = innobase_map_isolation_level(
-		thd_get_trx_isolation(thd));
+		thd_get_trx_isolation(thd)) & 3;
```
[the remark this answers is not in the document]
dr-m: That function already returns values between 0 and 3. I just checked, and even `g++-14` would complain in the same way as `g++-6` if I removed the redundant bit-masking:
```
storage/innobase/handler/ha_innodb.cc:4442:60: warning: conversion from ‘uint’ {aka ‘unsigned int’} to ‘unsigned char:2’ may change value [-Wconversion]
 4442 |         trx->isolation_level = innobase_map_isolation_level(
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
 4443 |                 thd_get_trx_isolation(thd));
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~                 
```
Adding the redundant `& 3` to each return value of `innobase_map_isolation_level()` would not silence this warning.

I do not think that GCC has a concept that a function return value fits in something else than 8, 16, 32, … bits. LLVM and clang might have, but their `-Wconversion` complains only about violations of those coarser boundaries (such as assigning a wider value to a bit-field that is at most 8 bits).

PR #3072, opened 2024-02-16
MDEV-32175 References to buf_page_t::frame may cost some performance

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32175*

## Description
This is work in progress, for performance testing. We are reducing the number of accesses to `buf_page_t::frame`, because there is an idea to remove that data member and to replace it with a member function that would perform some pointer arithmetics.

Based on the outcome, we might want to revert changes that cause additional calls to page_offset() or page_align().

## Release Notes
This does not need to be mentioned, because this is a relatively small performance improvement or code cleanup.

## How can this PR be tested?
This is a low-level change that is covered by the regression test suite. It should be tested with RQG as well.

A small part of this depends on #3069.

This is a code cleanup or a performance improvement.
## PR quality check


storage/innobase/btr/btr0pcur.cc
```diff
@@ -242,7 +239,7 @@ static bool btr_pcur_optimistic_latch_leaves(btr_pcur_t *pcur,
 
   uint64_t modify_clock;
   uint32_t left_page_no;
-  const page_t *const page= block->page.frame;
+  const page_t *const page= btr_pcur_get_page(pcur);
   {
     transactional_shared_lock_guard<block_lock> g{block->page.lock};
```
dr-m, line 244: I thought that I would double-check the sizes of some functions to ensure that these changes are not causing some functions to grow. I started comparing the output of the following for a `CMAKE_BUILD_TYPE=RelWithDebInfo`:
```sh
find storage/innobase -type f -name \*.o -print0|xargs -0 wc -c
```
This is not perfect because the file names are not sorted. For the very first file `btr0pcur.o` there would appear to be an increase of size. I went on to compare the output of `objdump -d` for this file, both with and without these changes. `btr_pcur_reset()` and `btr_pcur_copy_stored_position()` are shrinking a little. But `btr_pcur_t::restore_position()` is growing from 1953 to 1973 bytes. The invocation of this function `btr_pcur_optimistic_latch_leaves()` seems to be inlined there. Because the modes `BTR_SEARCH_PREV` and `BTR_MODIFY_PREV` could be rather rare, `btr_pcur_optimistic_latch_leaves()` might better be declared `ATTRIBUTE_NOINLINE`. I did that and confirmed that the size increase is due to the changes to this function; `btr_pcur_t::restore_position()` with the non-inline invocation would be 1300 bytes for both. The patched version adds a load of `srv_page_size`, right at this point. So, we’d better retain some references to `block->page.frame` and not invoke `page_align()`.

I will revise this patch accordingly. The reason why I tried to remove references to `block->page.frame` was that in #3107 I wanted to remove `buf_page_t::frame` and compute that field. When I tested the performance of that change, I observed a regression. I think that we’d better retain this field.

storage/innobase/btr/btr0pcur.cc
```diff
@@ -260,16 +257,19 @@ static bool btr_pcur_optimistic_latch_leaves(btr_pcur_t *pcur,
...
                                      left_page_no), block->zip_size(),
                            mode, nullptr, BUF_GET_POSSIBLY_FREED, mtr);
-    if (!prev ||
-        page_is_comp(prev->page.frame) != page_is_comp(block->page.frame) ||
-        memcmp_aligned<2>(block->page.frame, prev->page.frame, 2) ||
-        memcmp_aligned<2>(block->page.frame + PAGE_HEADER + PAGE_INDEX_ID,
-                          prev->page.frame + PAGE_HEADER + PAGE_INDEX_ID, 8))
+    if (!prev)
+      goto fail;
+    prev_page= prev->page.frame;
```
dr-m, line 268: If I change `btr_pcur_get_page()` and friends to refer directly to `block->page.frame`, then a non-inlined version of this function would be the same size. But, inlined as part of `btr_pcur_t::restore_position` the result would be 20 bytes larger again. Once I reverted the changes related to the variable `prev_page` the size would be the same again.

I must also remove any added use of `page_align()` and `page_offset()` elsewhere.

dr-m: It turns out that we need to add `page` parameters to many functions that currently only take `rec`. I am working on that. It will be a huge change, so I may end up splitting this fix into two parts.

storage/innobase/btr/btr0pcur.cc
dr-m: With an extensive revision (about 18,000 lines of `diff` compared to base 10.6), I am able to reduce the total size of `.o` files by more than 200 KiB. Once I have revised that fix so that there is no `.text` size increase for any object file, I will update this pull request with a greatly reduced version of that.

PR #3075, opened 2024-02-20
Convert instances of `sprintf` to `snprintf` in `sql`

tgross35:

## Description

This is a portion of #2958

## Release Notes

None needed

## How can this PR be tested?

## PR quality check


dlenski:
> This is a portion of https://github.com/MariaDB/server/pull/2958

Per https://github.com/MariaDB/server/pull/2958#issuecomment-1953627824, you've split this up into 3 smaller pieces in order to narrow down the source of the CI failures.

tgross35:
@LinuxJedi since you have been reviewing the other sprintf PRs, would you mind giving some thoughts on https://github.com/MariaDB/server/blob/e776115ebb723f4280573886363c93daa77ed342/include/bufcursor.h / https://github.com/MariaDB/server/blob/e776115ebb723f4280573886363c93daa77ed342/sql/bufcursor.c included as part of this PR? It is meant as an easy API to replace the pattern of `sprintf`ing or `strcpy`ing something and then incrementing a cursor. It just asserts if there isn't enough room.

I added this because that pattern is used frequently (the rest of this PR and sometimes in https://github.com/MariaDB/server/pull/3076/files). Switching directly to `snprintf` in these cases is messy - keeping the math correct while handling errors seems too error prone to use directly, IMO.

LinuxJedi:
@tgross35 what would that provide us over using the `String` class in `sql_string.h`?

tgross35:
> @tgross35 what would that provide us over using the `String` class in `sql_string.h`?

Not enough to be special, I can update it to use that

PR #3076, opened 2024-02-20
Convert instances of `sprintf` in `storage/` to `snprintf`

tgross35:

## Description

This is a portion of #2958

## Release Notes

None needed

## How can this PR be tested?

## PR quality check


storage/innobase/include/ut0ut.h
```diff
@@ -185,7 +188,8 @@ Sprintfs a timestamp to a buffer, 13..14 chars plus terminating NUL. */
 void
 ut_sprintf_timestamp(
 /*=================*/
-	char*	buf); /*!< in: buffer where to sprintf */
+	char*	buf); /*!< in: buffer where to sprintf. Must be at least
+			`UT_BUF_MIN_LEN` long. */
```
dlenski, line 192: This should be `UT_` ***`TS_`*** `BUF_MIN_LEN`

dlenski (changes requested):

dlenski:
@LinuxJedi, @vuvova, as @ottok and I and others at @aws have been saying for a while: it's extremely difficult for external contributors to evaluate the readiness of their PRs when there are large numbers of ongoing CI failures _on the upstream branches_.

This PR is a case in point.

The builds currently failing on this PR…
image

… are _the same as_ the builds failing on the upstream parent commit (https://github.com/MariaDB/server/commits/a6833faf119e287d63dd367341607846505f1dbc):
image

vuvova:
Yes, I know. Making all tests pass is still work in progress

dlenski:
@vuvova wrote:

> Yes, I know. Making all tests pass is still work in progress

Consider disabling the failing ones altogether until you can stabilize them, so as not to confuse and frustrate new external contributors.

tgross35:
The CI does say which tests are required and which aren't, I split this up because required tests were failing and because I can't actually build on my local machine - which is why I started these in the first place https://mariadb.zulipchat.com/#narrow/stream/118759-general/topic/Building.20on.20aarch64.20macos. But yes, it would be nice to see an entire green pipeline :)

Thank you for the thorough review, I'll update in a bit.

storage/innobase/include/ut0ut.h
tgross35: Good catch, updated this

dr-m:
I made some notes in #3613 and experimented with removing a deprecation suppression that prevented Apple XCode from flagging a few `sprintf()`. There is also #2432.

PR #3077, opened 2024-02-20
Convert miscellaneous instances of `sprintf` to `snprintf`

tgross35:

## Description

This is a portion of #2958

## Release Notes

None needed

## How can this PR be tested?

## PR quality check


client/mysqladmin.cc
```diff
@@ -1351,7 +1351,8 @@ static void usage(void)
 
 static int drop_db(MYSQL *mysql, const char *db)
 {
-  char name_buff[FN_REFLEN+20], buf[10];
+  const size_t name_buf_len= FN_REFLEN+20;
```
dlenski, line 1354: As long as you're removing/updating mysterious constants… any idea where the `+20` and the `10` come from, here?

plugin/type_inet/sql_type_inet.cc
```diff
@@ -500,7 +501,7 @@ size_t Inet6::to_string(char *dst, size_t dstsize) const
       //
       // If it is not the last field, append closing ':'.
 
-      p += sprintf(p, "%x", ipv6_words[i]);
+      p += snprintf(p, p_end-p, "%x", ipv6_words[i]);
```
dlenski, line 504: As in https://github.com/MariaDB/server/pull/3075#discussion_r1496403722, I think `p_end-p` is generally harder for a reader to understand than simply using `dstsize` here.

tgross35: `dstsize` wouldn't be correct since `p` moves around, but I didn't realize there is already a `dstsize_available` variable. Updated this to use that and also assert if the buffer is too short.

client/mysqladmin.cc
tgross35: 20 is just enough to fit the ``drop database `...` `` prefix around the string, and I figure 10 is considered enough to fit user input. Added defines for these.

tgross35:
This one should be ready for another look when you get the chance @dlenski

plugin/type_inet/sql_type_inet.cc
```diff
@@ -499,8 +499,10 @@ size_t Inet6::to_string(char *dst, size_t dstsize) const
       // hex-letters without leading zeros (recommended IPv6-format).
       //
       // If it is not the last field, append closing ':'.
-
-      p += sprintf(p, "%x", ipv6_words[i]);
+      dstsize_available= dstend - p;
+      int res= snprintf(p, dstsize_available, "%x", ipv6_words[i]);
+      DBUG_ASSERT(res >= 0);
+      p += res;
```
dlenski, line 505: Should `dstsize_available` be decreased here as well? E.g. `p += res; dstsize_available -= res;`

tgross35: Actually it doesn't have to be set at all since it is only read once per block, I removed line 502 here. The other blocks do not update `dstsize_available` after doing the pointer math.

tgross35: Must have just been flaky, all required tests pass now

LinuxJedi (approved):
I don't see any initial problems with this. Many thanks!

tgross35:
@LinuxJedi thanks, I changed the base to 10.5.

Side question, about how often do syncs among the versions happen?

client/mysqladmin.cc
```diff
@@ -1428,10 +1428,12 @@ static void usage(void)
   version		Get version info from server");
 }
 
+#define SHOW_MSG_LEN (FN_REFLEN + 20)
```
knielsen, line 1431: Why introduce this #define SHOW_MSG_LEN? It only makes it harder to understand the code, that the right size is being passed to snprintf() (as opposed to just sizeof()), and that the buffer is large enough that it will not be truncated and corrupt the command to be contructed.

client/mysqladmin.cc
```diff
@@ -1428,10 +1428,12 @@ static void usage(void)
...
+#define UINPUT_BUF_LEN 10
```
knielsen, line 1432: Why introduce this \#define UINPUT_BUF_LEN? This is unrelated to sprintf/snprintf(), should be done as a separate patch (or why do it at all, the code is simpler without the \#define, it's only used in one place)

mysql-test/lib/My/SafeProcess/safe_process.cc
```diff
@@ -236,7 +238,8 @@ int main(int argc, char* const argv[] )
   sigaction(SIGCHLD, &sa,NULL);
   sigaction(SIGABRT, &sa_abort,NULL);
 
-  sprintf(safe_process_name, "safe_process[%ld]", (long) own_pid);
+  snprintf(safe_process_name, SAFE_PROCESS_NAME_LEN,
+           "safe_process[%ld]", (long) own_pid);
```
knielsen, line 242: Again, don't introduce SAFE_PROCESS_NAME_LEN. It only makes it harder to see that the buffer size is correct for the string being constructed, and that the correct length is passed to sprintf() (as opposed to using sizeof())

knielsen (changes requested):

client/mysqladmin.cc
tgross35: You are right, I was thinking this existed elsewhere too somehow

tgross35: Actually this was by request to remove the magic values https://github.com/MariaDB/server/pull/3077#discussion_r1496437200

knielsen:
Also what about testing? The patch contains no test cases.

The mysqladmin changes have a restriction on the database name of FN_REFLEN. It needs to be tested that database names up to this length work, and longer (eg. FN_REFLEN+1) throw an error higher in the code, so the snprintf() doesn't silently truncate the name and send a corrupted command to the server.

If this is already covered by existing tests, it would be good to mention this in the commit comment.

gkodinov:
I'm sorry that processing this pull request fell through the way it did. Please do not be discouraged by it. We are improving our contribution processing routines. So, please consider taking another task in the future. 

The good news is that this is actually fixed in main: no more sprintf()s in these tools. So, closing the pull request as a result.

PR #3079, opened 2024-02-21
MDEV-33462 Disallow LOCK=NONE operation on statistics table

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-33462*

## Description
- Disallow LOCK=NONE operation for the statistics table. InnoDB can avoid to apply online log for the statistics table during dict_stats_save().

## How can this PR be tested?
./mtr innodb.innodb_stats_fetch

## PR quality check


storage/innobase/handler/handler0alter.cc
```diff
@@ -2815,6 +2815,15 @@ ha_innobase::check_if_supported_inplace_alter(
 		}
 	}
 
+	if (m_prebuilt->table->is_stats_table()) {
+		if (ha_alter_info->online) {
+			ha_alter_info->unsupported_reason =
+				"Online alter is not supported for "
+				"statistics table";
```
dr-m, line 2822: For non-English translations of `ER_ALTER_OPERATION_NOT_SUPPORTED_REASON` it would be better to avoid any hard-coded English text. Let us check what the rest of the English error message is saying:
```
ERROR 0A000: LOCK=NONE is not supported. Reason: _. Try LOCK=SHARED
```
In the above reason string, the only non-redundant part is "statistics table". I think that we should use the table name (`mysql.innodb_table_stats` or `mysql.innodb_index_stats`) as the reason. In that way, the error message works also in other languages.

dr-m (changes requested):
The commit message (as well as the bug report title) should say *why* the change is needed, not only *what* the change is. In its current form, it repeats the *what* part twice.

Please change the bug report title to something that describes the original problem, which was a hang or a crash. That is *why* this change is being made.

dr-m (approved):

PR #3080, opened 2024-02-22
MDEV-33526: Create IS.slave_status table as alias for show replica status command

an3l:

## PR quality check


knielsen:
You need to also create INFORMATION_SCHEMA.slave_status. INFORMATION_SCHEMA.replica_status can be an alias for that.

Use "slave" in the internal names in the code to be consistent with the rest of the code.

In the tests, primarily use IS.slave_status when testing functionality, to be consistent with other tests. (There should also be test(s) to specifically test the availability of the IS.replica_status alias).

mysql-test/suite/rpl/include/check_show_slave_status_and_is_slave_status.inc
```diff
@@ -0,0 +1,22 @@
...
+  let $is_slave_status_fields= `SELECT LTRIM(SUBSTRING('$is_slave_status_fields', LENGTH('$sss_field') + 2))`;
+
+  --echo # Check $sss_field
+  let $sss_field_value= query_get_value($show_query, $sss_field, $num_rows);
+  let $is_field_value= `select $sss_field from information_schema.slave_status`;
+  if (`SELECT strcmp("$sss_field_value", "$is_field_value") != 0`)
+  {
+    die SHOW SLAVE STATUS $sss_field value ($sss_field_value) does not match value of IS.slave_status ($is_field_value);
+  }
+}
```
bnestere, line 22: New line

sql/sql_show.cc
```diff
@@ -8595,6 +8601,231 @@ get_referential_constraints_record(THD *thd, TABLE_LIST *tables,
...
+      !(elements= master_info_index->master_info_hash.records))
+    DBUG_RETURN(1);
+
+  mysql_mutex_lock(&LOCK_active_mi);
+  for (i= 0; i < elements; i++)
+  {
+    mi= (Master_info *) my_hash_element(&master_info_index->
+                                        master_info_hash, i);
+    mysql_mutex_lock(&mi->sleep_lock);
+    mi->users++;                                // Mark used
```
bnestere, line 8812: If we are holding `LOCK_active_mi` for this whole loop, why play with the internal reference counting at all? Which is protected by `LOCK_active_mi`. We probably don't want to mess with it (so we'd neither `user++` nor `mi->release()`) as is the pattern of `show_all_master_info()`. Another inconsistency between this and `show all slaves status`, is the later sorts the results by name. Did you skip that intentionally?

bnestere (changes requested):
Hi @an3l !

Thanks for the patch :) I haven't quite finished reviewing, but there's a few things I've noted to address, and I'll finish tomorrow (I haven't gone through the test results at all yet).

sql/sql_show.cc
```diff
@@ -10403,6 +10695,8 @@ ST_SCHEMA_TABLE schema_tables[]=
    fill_status, make_old_format, 0, 0, -1, 0, 0},
   {"SESSION_VARIABLES", Show::variables_fields_info, 0,
    fill_variables, make_old_format, 0, 0, -1, 0, 0},
+  {"SLAVE_STATUS", Show::slave_status_info, 0,
```
bnestere, line 10698: Oh one more thing - Kristian mentioned we should have both `slave status` and `replica status`; where the later is an alias of the former. It looks like `replica` was scrapped altogether in favor of `slave`, but we need both.

bnestere:
Hey Anel!

Andrei mentioned a similar task, MDEV-32172, which aims to extend the performance_schema with related items. We discussed that `SHOW SLAVE STATUS` has overlap between the ideas of both information_schema (intending to report on metadata), and performance_schema (intending to report on performance statistics). So the information_schema extension should only report on things that are set by `CHANGE MASTER TO` (and have a name to reflect that scope, and also be consistent with the rest of the tables, e.g. `INFORMATION_SCHEMA.replicas`); and then anything run-time should be handled by MDEV-32172 in the performance schema.

sql/handler.h
```diff
@@ -1072,8 +1072,8 @@ enum enum_schema_tables
   SCH_OPTIMIZER_COSTS,
   SCH_OPT_TRACE,
   SCH_PARAMETERS,
-  SCH_PERIODS,
   SCH_PARTITIONS,
+  SCH_PERIODS,
```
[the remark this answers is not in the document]
an3l: Hi @FooBarrior  what is your take about this? Can it be single commit in this PR or to create the new one??

sql/rpl_mi.h
```diff
@@ -100,11 +100,12 @@ class Domain_id_filter
...
     buffer.
 
-    @param thd [IN]                   thread handler
+    @param thd   [IN]                   thread handler
+    @param field [IN]                   field (used to store in IS.replica_status)
 
     @retval void
   */
-  void store_ids(THD *thd);
+  void store_ids(THD *thd, Field *field);
```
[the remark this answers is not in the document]
an3l: Thanks, I have updated in the next commit to use templated ptfms

sql/sql_show.cc
```diff
@@ -69,7 +69,14 @@
 #include "key.h"
 
 #include "lex_symbol.h"
+#include "slave.h"
```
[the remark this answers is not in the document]
an3l: Done

sql/sql_show.cc
```diff
@@ -8595,6 +8601,231 @@ get_referential_constraints_record(THD *thd, TABLE_LIST *tables,
...
+#else
+  table->field[29]->store(mi->ssl ? &msg_ignored: &msg_no, cs);
+#endif
+  table->field[30]->store(mi->ssl_ca, strlen(mi->ssl_ca), cs);
+  table->field[31]->store(mi->ssl_capath, strlen(mi->ssl_capath), cs);
+  table->field[32]->store(mi->ssl_cert, strlen(mi->ssl_cert), cs);
+  table->field[33]->store(mi->ssl_cipher, strlen(mi->ssl_cipher), cs);
+  table->field[34]->store(mi->ssl_key, strlen(mi->ssl_key), cs);
+  // SBM
+  if ((mi->slave_running == MYSQL_SLAVE_RUN_READING) &&
```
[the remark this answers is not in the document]
an3l: Done

sql/sql_show.cc
an3l: I agree, have removed. Well yes, I didn't want to sort, because we will have ability to use `WHERE` clause.

sql/sql_show.cc
an3l: Will do and update MDEV.

LinuxJedi:
Closing this as it has been implemented, as per comment in the Jira.

PR #3081, opened 2024-02-22
MDEV-33011 mariabackup --backup: FATAL ERROR: ... Can't open datafile…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33011*

## Description

    The root cause is the WAL logging of file operation when the actual
    operation fails afterwards. It creates a situation with a log entry for
    a operation that would always fail. I could simulate both the backup
    scenario error and Innodb recovery failure exploiting the weakness.
    
    We are following WAL for file rename operation and once logged the
    operation must eventually complete successfully, or it is a major
    catastrophe. Right now, we fail for rename and handle it as normal error
    and it is the problem.
    
    I created a patch to address RENAME operation to a non existing schema
    where the destination schema directory is missing. The patch checks for
    the missing schema before logging in an attempt to avoid the failure
    after WAL log is written/flushed. I also checked that the schema cannot
    be dropped or there cannot be any race with other rename to the same
    file. This is protected by the MDL lock in SQL today.
    
    The patch should be a good improvement over the current situation
    and solves the issue at hand.

## Release Notes
None

## How can this PR be tested?
./mtr innodb.rename_recovery
./mtr mariabackup.rename_during_backup

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/rename_table.test
+++ b/mysql-test/suite/innodb/t/rename_table.test
@@ -30,16 +30,21 @@ DROP DATABASE abc_def;
 
 DROP DATABASE abc_def2;
 
-call mtr.add_suppression("InnoDB: (Operating system error|The error means|Cannot rename file)");
+call mtr.add_suppression("InnoDB: Cannot rename '.*t1.ibd' to '.*non_existing_db.*' because the target schema directory doesn't exist");
 
 CREATE TABLE t1 (a INT) ENGINE=InnoDB;
+INSERT INTO t1 VALUES(100);
 --replace_result "\\" "/"
 --error ER_ERROR_ON_RENAME
 RENAME TABLE t1 TO non_existing_db.t1;
 
---let SEARCH_PATTERN= \[ERROR\] InnoDB: Cannot rename file '.*t1\.ibd' to '.*non_existing_db
+--let SEARCH_PATTERN= \[ERROR\] InnoDB: Cannot rename '.*t1\.ibd' to '.*non_existing_db
 let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
 --source include/search_pattern_in_file.inc
 
+SET GLOBAL innodb_fast_shutdown=2;
+--source include/restart_mysqld.inc
+
+SELECT * FROM t1;
 # Cleanup
 DROP TABLE t1;
--- a/mysql-test/suite/mariabackup/rename_during_backup.test
+++ b/mysql-test/suite/mariabackup/rename_during_backup.test
@@ -90,4 +90,31 @@ SELECT * from t6;
 DROP TABLE t6;
 rmdir $targetdir;
 
+--echo #
+--echo # MDEV-33011 mariabackup --backup: FATAL ERROR: ... Can't open datafile cool_down/t3
+--echo #
 
+--disable_query_log
+call mtr.add_suppression("InnoDB: Cannot rename '.*t1.ibd' to '.*non_existing_db.*' because the target schema directory doesn't exist");
+--enable_query_log
+
+mkdir $targetdir;
+
+--echo # Simulate zero initialized page to defer tablespace load after rename log is found
+SET @save_dbug = @@SESSION.debug_dbug;
+SET DEBUG_DBUG="+d,checkpoint_after_file_create";
+CREATE TABLE t1(f1 INT NOT NULL)ENGINE=InnoDB;
+INSERT INTO t1 VALUES(1);
+
+--echo # RENAME that fails after redo log entry is written and flushed
+--replace_result "\\" "/"
+--error ER_ERROR_ON_RENAME
+RENAME TABLE t1 TO non_existing_db.t1;
+
+--disable_result_log
+exec $XTRABACKUP --defaults-file=$MYSQLTEST_VARDIR/my.cnf --backup --target-dir=$targetdir;
+exec $XTRABACKUP --prepare --target-dir=$targetdir;
+--enable_result_log
+
+DROP TABLE t1;
+rmdir $targetdir;
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -1959,24 +1959,20 @@ Allocates and builds a file name from a path, a table or tablespace name
 and a suffix. The string must be freed by caller with ut_free().
 @param[in] path NULL or the directory path or the full path and filename.
 @param[in] name NULL if path is full, or Table/Tablespace name
-@param[in] suffix NULL or the file extention to use.
+@param[in] extension NULL or the file extension to use.
 @param[in] trim_name true if the last name on the path should be trimmed.
 @return own: file name */
 char*
-fil_make_filepath(
+fil_make_filepath_low(
 	const char*	path,
 	const char*	name,
-	ib_extention	ext,
+	ib_extention	extension,
 	bool		trim_name)
 {
 	/* The path may contain the basename of the file, if so we do not
 	need the name.  If the path is NULL, we can use the default path,
 	but there needs to be a name. */
 	ut_ad(path != NULL || name != NULL);
-
-	/* If we are going to strip a name off the path, there better be a
-	path and a new name to put back on. */
-	ut_ad(!trim_name || (path != NULL && name != NULL));
 
 	if (path == NULL) {
 		path = fil_path_to_mysql_datadir;
@@ -1985,7 +1981,7 @@ fil_make_filepath(
 	ulint	len		= 0;	/* current length */
 	ulint	path_len	= strlen(path);
 	ulint	name_len	= (name ? strlen(name) : 0);
-	const char* suffix	= dot_ext[ext];
+	const char* suffix	= dot_ext[extension];
 	ulint	suffix_len	= strlen(suffix);
 	ulint	full_len	= path_len + 1 + name_len + suffix_len + 1;
 
@@ -2083,6 +2079,21 @@ fil_rename_tablespace_check(
 	}
 
 	exists = false;
+	auto schema_path= fil_make_dirpath(new_path, NULL, NO_EXT, true);
+	if (schema_path == NULL) {
+		return DB_ERROR;
+	}
+
+	if (os_file_status(schema_path, &exists, &ftype) && !exists) {
+		sql_print_error("InnoDB: Cannot rename '%s' to '%s'"
+				" because the target schema directory doesn't exist.",
+				old_path, new_path);
+		ut_free(schema_path);
+		return DB_ERROR;
+	}
+	ut_free(schema_path);
+	exists = false;
+
 	if (os_file_status(new_path, &exists, &ftype) && !exists) {
 		return DB_SUCCESS;
 	}
--- a/storage/innobase/include/fil0fil.h
+++ b/storage/innobase/include/fil0fil.h
@@ -1618,14 +1618,44 @@ Allocates and builds a file name from a path, a table or tablespace name
 and a suffix. The string must be freed by caller with ut_free().
 @param[in] path NULL or the directory path or the full path and filename.
 @param[in] name NULL if path is full, or Table/Tablespace name
-@param[in] suffix NULL or the file extention to use.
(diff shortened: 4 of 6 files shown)


mysql-test/suite/innodb/t/rename_table.test
```diff
@@ -30,16 +30,21 @@ DROP DATABASE abc_def;
 
 DROP DATABASE abc_def2;
 
-call mtr.add_suppression("InnoDB: (Operating system error|The error means|Cannot rename file)");
+call mtr.add_suppression("InnoDB: Cannot rename '.*/test/t1.ibd' to '.*non_existing_db.*' because the target schema directory doesn't exist");
```
dr-m, line 33: For some reason this suppression does not work on Microsoft Windows. Also the test `mariabackup.rename_during_backup` fails due to a similar reason.

mysql-test/suite/mariabackup/rename_during_backup.test
```diff
@@ -90,4 +90,31 @@ SELECT * from t6;
...
+--echo # MDEV-33011 mariabackup --backup: FATAL ERROR: ... Can't open datafile cool_down/t3
+--echo #
 
+--disable_query_log
+call mtr.add_suppression("InnoDB: Cannot rename '.*/test/t1.ibd' to '.*non_existing_db.*' because the target schema directory doesn't exist");
+--enable_query_log
+
+mkdir $targetdir;
+
+--echo # Simulate zero intialized page to defer tablespace load after rename log is found
```
dr-m, line 103: typo for "initialized"

storage/innobase/fil/fil0fil.cc
```diff
@@ -2084,6 +2093,22 @@ fil_rename_tablespace_check(
...
 	exists = false;
 	if (os_file_status(new_path, &exists, &ftype) && !exists) {
+		auto schema_path= fil_make_filepath(new_path, NULL, NO_EXT,
+						    true, true);
+		if (schema_path == NULL) {
+			return DB_ERROR;
+		}
+		exists= false;
+		if (os_file_status(schema_path, &exists, &ftype) && !exists)
+		{
```
dr-m, line 2103: In the archaic-style formatted code, the `{` is supposed to be on the same line, and there is a space before `=`.

storage/innobase/fil/fil0fil.cc
```diff
@@ -2084,6 +2093,22 @@ fil_rename_tablespace_check(
...
+			ib::error() << "Cannot rename '" << old_path
+				    << "' to '" << new_path
+				    << "' because the target schema directory"
+				    << " doesn't exist.";
```
dr-m, line 2107: I would prefer `sql_print_error("InnoDB:` when we are not displaying page identifiers.

storage/innobase/fil/fil0fil.cc
```diff
@@ -2084,6 +2093,22 @@ fil_rename_tablespace_check(
...
+			ut_free(schema_path);
+			return DB_ERROR;
+		}
+		ut_free(schema_path);
 		return DB_SUCCESS;
```
dr-m, line 2112: We could assign the return value to a variable, to avoid code duplication:
```c++
ut_free(schema_path);
return err;
```

storage/innobase/fil/fil0fil.cc
```diff
@@ -2084,6 +2093,22 @@ fil_rename_tablespace_check(
 
 	exists = false;
 	if (os_file_status(new_path, &exists, &ftype) && !exists) {
+		auto schema_path= fil_make_filepath(new_path, NULL, NO_EXT,
+						    true, true);
+		if (schema_path == NULL) {
+			return DB_ERROR;
+		}
```
dr-m, line 2100: We could avoid deepening the indentation by writing this as follows:
```c++
	if (!os_file_status(new_path, &exists, &ftype) || exists) {
	} else if (auto schema_path = fil_make_filepath(/*…*/)) {
		// ...
		ut_free(schema_path);
		return err;
	} else {
		return DB_ERROR;
	}
```

storage/innobase/include/fil0fil.h
```diff
@@ -1619,13 +1619,15 @@ and a suffix. The string must be freed by caller with ut_free().
...
+@param[in] dir_path true, if the intended path name is directory.
 @return own: file name */
 char*
 fil_make_filepath(
 	const char*	path,
 	const char*	name,
 	ib_extention	suffix,
-	bool		strip_name);
+	bool		strip_name,
+	bool		dir_path = false);
```
dr-m, line 1630: Instead of adding a new Boolean parameter, could we add a `suffix` enumeration value to say that it is a directory, similar to the value `NO_EXT` that is only being used in `SysTablespace::parse_params()`. We might also refactor that code and repurpose `NO_EXT`. In that case, `Datafile::make_filepath()` would have to assert that it is not passing this `suffix` value to `fil_make_filepath()`.

By the way, the parameter `strip_name` is not documented in the function comment.

mysql-test/suite/innodb/t/rename_table.test
mariadb-DebarunBanerjee: Fixed.

storage/innobase/fil/fil0fil.cc
mariadb-DebarunBanerjee: right.

storage/innobase/fil/fil0fil.cc
mariadb-DebarunBanerjee: Now restructured the code a bit based on other comments. Please check. ut_free is done in 2 places but should be ok I think.

storage/innobase/fil/fil0fil.cc
mariadb-DebarunBanerjee: Well, perhaps it would lose the readability a bit this way. With a fresh look, I think the check for schema can be done independently and the last updated patch should solve the concern.

storage/innobase/include/fil0fil.h
mariadb-DebarunBanerjee: We don't want any suffix and using NO_EXT should be good. Adding another blank suffix of different name would be confusing I think.

Actually the function fil_make_filepath() works perfectly with the current set of inputs. The only issue is the debug assert 
```
/* If we are going to strip a name off the path, there better be a
path and a new name to put back on */
ut_ad(!trim_name || (path != NULL && name != NULL))
```
The new boolean parameter is added to preserve the debug assert for all other cases.
`bool		dir_path = false); `
and relaxing the assert when we are creating the directory path. So, I think we should not mix it with the functional parameters that are passed to drive the behaviour of the function.

If we want to simplify (removing the bool parameter) I could see 2 viable approaches.

1. Comment the debug assert that prevent the function to return a directory in debug mode.

OR

2. Have a wrapper function
```
fil_make_file_path(...) {
   Debug Assert
   fil_make_file_path_low()
} 
fil_make_directory_path() {
   fil_make_file_path_low()
}
Rename fil_make_file_path to fil_make_file_path_low and make it static scope.
```
I will upload 2 commits one with each method respectively. Please let me know which one you like. Or you still prefer yet any different approach.

dr-m: I think that having two wrapper functions (`inline` ones) would be best. In this way, there should be no overhead in `CMAKE_BUILD_TYPE=RelWithDebInfo` builds.

mysql-test/suite/innodb/t/rename_table.test
```diff
@@ -30,16 +30,21 @@ DROP DATABASE abc_def;
 
 DROP DATABASE abc_def2;
 
-call mtr.add_suppression("InnoDB: (Operating system error|The error means|Cannot rename file)");
+call mtr.add_suppression("InnoDB: Cannot rename '.*t1.ibd' to '.*non_existing_db.*' because the target schema directory doesn't exist");
 
 CREATE TABLE t1 (a INT) ENGINE=InnoDB;
+INSERT INTO t1 VALUES(100);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Well, not strictly required I think but this way we can at least select the row back from the table after restart. Kind of making sure the table good and functioning.

storage/innobase/include/fil0fil.h
mariadb-DebarunBanerjee: Thanks, updated the patch.

mysql-test/suite/innodb/t/rename_table.test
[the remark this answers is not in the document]
dr-m: Sure, but `SELECT` on an empty table should equally well report an error if one of the accessed pages of the 4-page tablespace was corrupted. The extra `INSERT` does not hurt much, so I am OK to keep it.

storage/innobase/fil/fil0fil.cc
```diff
@@ -1959,11 +1959,11 @@ Allocates and builds a file name from a path, a table or tablespace name
 and a suffix. The string must be freed by caller with ut_free().
 @param[in] path NULL or the directory path or the full path and filename.
 @param[in] name NULL if path is full, or Table/Tablespace name
-@param[in] suffix NULL or the file extention to use.
+@param[in] ext NULL or the file extention to use.
```
dr-m, line 1962: "extension" would be the grammatically correct word. In the function declaration, the parameter is called `suffix`, which is a little confusing. Maybe it should be renamed there as well?

storage/innobase/include/fil0fil.h
```diff
@@ -1619,13 +1619,47 @@ and a suffix. The string must be freed by caller with ut_free().
...
+static inline char*
 fil_make_filepath(
 	const char*	path,
 	const char*	name,
 	ib_extention	suffix,
-	bool		strip_name);
+	bool		trim_name) {
+  /* If we are going to strip a name off the path, there better be a
+  path and a new name to put back on. */
+  ut_ad(!trim_name || (path != NULL && name != NULL));
```
dr-m, line 1645: I would not use the macro `NULL` in new code, but the C++11 `nullptr`. Furthermore, I’d prefer conversion from pointer to Boolean: `!trim_name || (path && name)`.

storage/innobase/include/fil0fil.h
```diff
@@ -1619,13 +1619,47 @@ and a suffix. The string must be freed by caller with ut_free().
...
+@param[in] suffix NULL or the file extention to use.
+@param[in] trim_name true if the last name on the path should be trimmed.
+@return own: file name */
+static inline char*
 fil_make_filepath(
 	const char*	path,
 	const char*	name,
 	ib_extention	suffix,
-	bool		strip_name);
+	bool		trim_name) {
```
dr-m, line 1642: Please insert a line break before any `{` that starts the function body, so that Emacs `narrow-to-defun` and similar functions will continue to work. Only for member functions we deviate from this rule.

Also, I think that new functions would better be declared or defined in the common formatting style, with multiple parameters per line, no TAB and no `[in]` after `@param`.

dr-m (approved):
Thank you. This is good enough for now, after addressing my review comments.

Ideally, we would refactor this code more. We could attempt to create an empty file in place of the requested new file name and `fstat` it to see if it is in the same file system with the old name, to better predict if the renaming would succeed. Alternatively, we could write `FILE_MODIFY` or a new type of record right before attempting the rename, but this could change the recovery logic in an incompatible way, while we expect crash-downgrades to succeed.

We might also tweak the recovery in 10.6+ so that the final renaming could be deferred until we have recovered transactions and can determine if there was a rollback of the DDL operation. Before 10.6, DDL operations are not fully crash safe, not even when they are internal to InnoDB, such as `TRUNCATE TABLE`.

storage/innobase/fil/fil0fil.cc
mariadb-DebarunBanerjee: Done

storage/innobase/include/fil0fil.h
mariadb-DebarunBanerjee: Done

storage/innobase/include/fil0fil.h
mariadb-DebarunBanerjee: Done

PR #3083, opened 2024-02-26
MDEV-19044  Alter table corrupts while applying the modification log

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-19044*

## Description
Problem:
========
- InnoDB reads the length of the variable length field wrongly while applying the modification log of instant table.

Solution:
========
rec_init_offsets_comp_ordinary(): For the temporary instant file record, InnoDB should read the length of the variable length field from the record itself.

## How can this PR be tested?
./mtr innodb.instant_alter_crash test case

## PR quality check


dr-m:
The 10.4 version of this in #3084.

PR #3084, opened 2024-02-26
MDEV-19044  Alter table corrupts while applying the modification log

Thirunarayanan:

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/instant_alter_crash.test
+++ b/mysql-test/suite/innodb/t/instant_alter_crash.test
@@ -202,32 +202,6 @@ SHOW CREATE TABLE t1;
 SHOW CREATE TABLE t2;
 SHOW CREATE TABLE t3;
 DROP TABLE t1,t2,t3;
-
+SET DEBUG_SYNC=RESET;
 --remove_files_wildcard $MYSQLD_DATADIR/test #sql*.frm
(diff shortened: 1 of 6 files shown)


mysql-test/suite/innodb/t/instant_alter_crash.test
```diff
@@ -225,6 +225,33 @@ INSERT IGNORE INTO t1 (f3) VALUES ( 'b' );
...
 
+connection default;
+reap;
+CHECK TABLE t1;
+DROP TABLE t1;
+
+--echo #
+--echo # MDEV-19044 Alter table corrupts while applying the
+--echo #             modification log
+--echo #
```
dr-m, line 236: I think that this test as well as the preceding test for MDEV-26198 would better belong to the test file `innodb.instant_alter_debug`, which is for debug instrumented tests that do not require the server to be restarted. Unlike this test, that test could also run with `./mtr --embedded`.

The purpose of this test file is to cover crash recovery. However, I see that already from its initial revision a4948dafcd7eee65f16d848bdc6562fc49ef8916 this test did require a debug instrumented build. So, the only thing that we are losing here is some `./mtr --embedded` coverage on debug instrumented builds.

dr-m (approved):

PR #3086, opened 2024-02-26
MDEV-33546: Rpl_semi_sync_slave_status is ON When Replication Is Not Configured

bnestere:
If a server has a default configuration (e.g. in a my.cnf file) with
rpl_semi_sync_slave_enabled set, on server start, the corresponding
rpl_semi_sync_slave_status variable will also be ON initially, even
if the slave was never configured/started. This is because the
Repl_semi_sync_slave initialization logic (function init_object())
sets the running status to the enabled value during
init_server_components().

This patch fixes this by removing the statement which sets the
semi-sync slave running status from the initialization logic. An
additional change needed from this is to semi-sync recovery: this
status variable was used as a condition to determine binlog
truncation during server recovery. This patch also switches this
condition to reference the global rpl_semi_sync_slave_enabled
variable. Though note, the semi-sync recovery condition is to be
changed entirely with the MDEV-33424 agenda.

1. The first commit is the regression which shows the bug
2. The second commit contains the code patch and fix for the regression failing test

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_semi_sync_fail_over.result
+++ b/mysql-test/suite/rpl/r/rpl_semi_sync_fail_over.result
@@ -50,6 +50,8 @@ SELECT @@GLOBAL.gtid_current_pos;
 0-1-4
 # restart: --skip-slave-start=1 --rpl-semi-sync-slave-enabled=1
 connection server_1;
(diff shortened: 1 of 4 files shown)


andrelkin (approved):
The fixes look good.

PR #3087, opened 2024-02-28
MDEV-33487 Feature non-blocking binlog

poempeng:
- [x] *The Jira issue number for this PR is: [MDEV-33487](https://jira.mariadb.org/browse/MDEV-33487)*

## Description
During the commit stage of a large transaction, it may cost too much time to write binlog cache and block all subsequent transactions for a long time. One possible solution of this issue is to write the binlog of the large transaction to a temporary file and then rename it to the next new binlog file.
    
We need to reserve some space for format description event, gtid list event, binlog checkpoint event and gtid event. After writing the binlog cache, we lock LOCK_log and write the aforementioned events to the reserved space. The default size of reserved space is 10MB, any unusable space is filled with a skip event(Ignorable_log_event). If any error occurs before acquiring LOCK_log, we fallback to the original commit path. However, if an error occurs after acquiring LOCK_log, we abort the server.

## Release Notes
New variable: non_blocking_binlog_threshold, if the size of binlog cache exceeds this value, try using non-blocking binlog.

## How can this PR be tested?
1. Run a sysbench with write only workload.
2. Set non_blocking_binlog_threshold to a non-default value, commit a large transaction(the size of binlog generated by this transaction should be greater than non_blocking_binlog_threshold).
3. Reset non_blocking_binlog_threshold to default value, commit a large transaction similar to step 2.
4. Compare the QPS changes during the commit operations of step 2 and step 3.

## PR quality check


knielsen (changes requested):
Review on the mailing list:

https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/thread/6OUVSUBUUCU6AHUR4CYGYSSZZUKZ5FIZ/

bnestere:
Hi @poempeng ,

As a review was provided over three weeks ago (to which we have received no response), and there is now a new PR (PR-3402) which competes with this work, and the author is very active, I will be closing this PR in favor of continuing with the new PR.

Thank you very much for your submission.

PR #3089, opened 2024-02-28
MDEV-33551: Semi-sync Wait Point AFTER_COMMIT Slow on Workloads with Heavy Concurrency

bnestere:
When using semi-sync replication with
rpl_semi_sync_master_wait_point=AFTER_COMMIT, the performance of the
primary can significantly reduce compared to AFTER_SYNC's
performance for workloads with many concurrent users executing
transactions. This is because all connections on the primary share
the same cond_wait variable/mutex pair, so any time an ACK is
received from a replica, all waiting connections are awoken to check
if the ACK was for itself, which is done in mutual exclusion.

This patch changes this such that the waiting THD will use its own
local condition variable, and the ACK receiver thread only signals
connections which have been ACKed for wakeup.

Additionally:
 1) At master shutdown (when waiting for slaves), instead of the
    main loop individually waiting for each ACK,
    await_slave_reply() (renamed await_all_slave_replies) now
    awaits ACKs for each transaction within one invocation.

 2) The time when thd::is_awaiting_semi_sync_ack is set is moved
    to at binlogging time, to ensure transactions which have been
    binlogged and queued up to await an ACK are not killed,
    and are still waited on.

 3) Repl_semi_sync_master::commit_trx() no longer loops to await
    its specific ACK. It waits once, and will either fail from
    timeout, or receive its ACK.

 1) The first commit shows the regression via an MTR test and added
     assertion
 2) The second commit is the patch

### DIFF
--- a/sql/semisync_master.cc
+++ b/sql/semisync_master.cc
@@ -68,6 +68,19 @@ static ulonglong timespec_to_usec(const struct timespec *ts)
   return (ulonglong) ts->tv_sec * TIME_MILLION + ts->tv_nsec / TIME_THOUSAND;
 }
 
+int signal_waiting_transaction(THD *waiting_thd, const char *binlog_file,
+                                my_off_t binlog_pos)
+{
+  /*
+    It is possible that the connection thd waiting for an ACK was killed. In
+    such circumstance, the connection thread will nullify the thd member of its
+    Active_tranx node. So before we try to signal, ensure the THD exists.
+  */
+  if (waiting_thd)
+    mysql_cond_signal(&waiting_thd->COND_wakeup_ready);
+  return 0;
+}
+
 /*******************************************************************************
  *
  * <Active_tranx> class : manage all active transaction nodes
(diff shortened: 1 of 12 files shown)


knielsen (changes requested):
Review on the mailing list:
https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/thread/EWZAANRVB7PHGQ5H44SQS3ZS54JXXMET/

sql/semisync_master.cc
```diff
@@ -1362,21 +1400,27 @@ void Repl_semi_sync_master::set_export_stats()
...
+    emptied.
+ */
+  lock();
+  if (get_master_enabled() && is_on() && !m_active_tranxs->is_empty())
+  {
+    if (msg)
+      sql_print_information(msg);
+    create_timeout(&timeout, NULL);
+    mysql_cond_timedwait(&COND_binlog_send, &LOCK_binlog, &timeout);
+  }
```
knielsen, line 1423: This should be a while(), to be robust to spurious condition variable wakeups.

sql/semisync_master.cc
```diff
@@ -979,17 +981,50 @@ int Repl_semi_sync_master::commit_trx(const char* trx_wait_binlog_name,
...
     }
 
+    /*
+      If our THD was killed (rather than awoken from an ACK) notify the
+      Active_tranx cache that we are no longer waiting for the ACK, so nobody
+      signals our COND var invalidly.
+    */
+    if (thd_killed(thd))
+      m_active_tranxs->unlink_thd_as_waiter(trx_wait_binlog_name,
+                                            trx_wait_binlog_pos);
```
knielsen, line 1018: (Just a style comment, up to you what you prefer: I'd use a flag here which assigned in the loop as to whether we were awoken or not, as thd_killed() can change value at any time. Though I think the code works in either case, as thd_killed() brobably doesn't change from true to false, and redundant call of unlink_thd_as_waiter is a no-op).

knielsen (approved):
Looks ok to me now, thanks Brandon.
Just one comment about being robust to spurious wakeup.
One style remark that is up to you.

 \- Kristian.

PR #3090, opened 2024-02-29
MDEV-33542  Inplace algorithm occupies more disk space compared to copy algorithm

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-33542*

## Description
Problem:
========
- In case of large file size, InnoDB adds the new extent even though there are many existing unused pages of the segment. Reason is that in case of larger file size, threshold (1/8 of reserved pages) for adding new extent has been reached frequently.

Solution:
=========
- Try to utilise the unused pages in the segment before adding the new extent in the file segment. To achieve that, threshold can be set to larger size for larger files.

need_for_new_extent(): In case of larger file size, try to use the 4 * FSP_EXTENT_SIZE as threshold to allocate the new extent.

## How can this PR be tested?
 Test case given in description can be run against patch and check the file size.

## PR quality check

### DIFF
--- a/storage/innobase/fsp/fsp0fsp.cc
+++ b/storage/innobase/fsp/fsp0fsp.cc
@@ -271,7 +271,7 @@ inline void xdes_init(const buf_block_t &block, xdes_t *descr, mtr_t *mtr)
 static MY_ATTRIBUTE((nonnull, warn_unused_result))
 dberr_t
 fseg_mark_page_used(fseg_inode_t *seg_inode, buf_block_t *iblock,
-                    ulint page, xdes_t *descr, buf_block_t *xdes, mtr_t *mtr)
+                    uint32_t page, xdes_t *descr, buf_block_t *xdes, mtr_t *mtr)
 {
   ut_ad(fil_page_get_type(iblock->page.frame) == FIL_PAGE_INODE);
   ut_ad(!((page_offset(seg_inode) - FSEG_ARR_OFFSET) % FSEG_INODE_SIZE));
@@ -995,7 +995,7 @@ MY_ATTRIBUTE((nonnull, warn_unused_result))
 @return error code */
 static dberr_t
 fsp_alloc_from_free_frag(buf_block_t *header, buf_block_t *xdes, xdes_t *descr,
-			 ulint bit, mtr_t *mtr)
+			 uint32_t bit, mtr_t *mtr)
 {
   if (UNIV_UNLIKELY(xdes_get_state(descr) != XDES_FREE_FRAG ||
                     !xdes_is_free(descr, bit)))
@@ -1987,29 +1987,42 @@ fseg_alloc_free_page_low(
 		}
 	}
 
-	/* In the big if-else below we look for ret_page and ret_descr */
-	/*-------------------------------------------------------------*/
-	if ((xdes_get_state(descr) == XDES_FSEG)
-	    && mach_read_from_8(descr + XDES_ID) == seg_id
-	    && xdes_is_free(descr, hint % FSP_EXTENT_SIZE)) {
+	const uint32_t extent_size = FSP_EXTENT_SIZE;
+	ret_descr = descr;
+	/* Try to get the page from extent which belongs to segment */
+	if (xdes_get_state(descr) == XDES_FSEG
+	    && mach_read_from_8(descr + XDES_ID) == seg_id) {
+		/* Get the page from the segment extent */
+		if (xdes_is_free(descr, hint % extent_size)) {
 take_hinted_page:
-		/* 1. We can take the hinted page
-		=================================*/
-		ret_descr = descr;
-		ret_page = hint;
-		/* Skip the check for extending the tablespace. If the
-		page hint were not within the size of the tablespace,
-		we would have got (descr == NULL) above and reset the hint. */
-		goto got_hinted_page;
-		/*-----------------------------------------------------------*/
-	} else if (xdes_get_state(descr) == XDES_FREE
-		   && reserved - used < reserved / FSEG_FILLFACTOR
-		   && used >= FSEG_FRAG_LIMIT) {
+			ret_page = hint;
+			goto got_hinted_page;
+		} else if (!xdes_is_full(descr)) {
+			/* Take the page from the same extent as the
+			hinted page (and the extent already belongs to
+			the segment) */
+			ret_page = xdes_find_free(descr, hint % extent_size);
+			if (ret_page == FIL_NULL) {
+				ut_ad(!has_done_reservation);
+				return nullptr;
+			}
+			ret_page += xdes_get_offset(ret_descr);
+			goto alloc_done;
+		}
+	}
 
-		/* 2. We allocate the free extent from space and can take
-		=========================================================
-		the hinted page
-		===============*/
+	/** If the number of unused but reserved pages in a segment is
+	esser than minimum value of 1/8 of reserved pages or
+	4 * FSP_EXTENT_SIZE and there are at least half of extent size
+	used pages, then we allow a new empty extent to be added to
+	the segment in fseg_alloc_free_page_general(). Otherwise, we use
+	unused pages of the segment. */
+	if (used < extent_size / 2 ||
+            reserved - used >= reserved / 8 ||
+            reserved - used >= extent_size * 4) {
+	} else if (xdes_get_state(descr) == XDES_FREE) {
+		/* Allocate the free extent from space and can
+		take the hinted page */
 		ret_descr = fsp_alloc_free_extent(space, hint, &xdes,
 						  mtr, err);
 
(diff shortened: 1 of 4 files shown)


Thirunarayanan:
`CREATE TABLE s (id varchar(255) PRIMARY KEY, d mediumblob NOT NULL)
ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci;
INSERT INTO s SELECT UUID(),repeat('x',255) FROM seq_1_to_1000000;
356515840 /home/thiru/mariadb_parent/server/10.6/bld_test-debug/mysql-test/var/mysqld.1/data//test/s.ibd
DELETE FROM s LIMIT 100000;
INSERT INTO s SELECT UUID(),repeat('x',255) FROM seq_1_to_100000;
390070272 /home/thiru/mariadb_parent/server/10.6/bld_test-debug/mysql-test/var/mysqld.1/data//test/s.ibd
ALTER TABLE s FORCE, ALGORITHM=INPLACE;
352321536 /home/thiru/mariadb_parent/server/10.6/bld_test-debug/mysql-test/var/mysqld.1/data//test/s.ibd
DROP TABLE s;
`

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2043,8 +2058,7 @@ fseg_alloc_free_page_low(
 		goto take_hinted_page;
 		/*-----------------------------------------------------------*/
 	} else if ((direction != FSP_NO_DIR)
-		   && ((reserved - used) < reserved / FSEG_FILLFACTOR)
-		   && (used >= FSEG_FRAG_LIMIT)
+		   && need_for_new_extent(used, reserved)
 		   && (ret_descr = fseg_alloc_free_extent(seg_inode, iblock,
 							  &xdes, space,
 							  mtr, err))) {
```
dr-m, line 2064: This is very similar to the previous `else if` condition. I think that we should refactor this so that the condition `need_for_new_extent()`  needs to be evaluated only once. Here is one idea:
```c++
	} else if (need_for_new_extent(used, reserved)) {
		goto no_new_extent;
	} else if (xdes_get_state(descr) == XDES_FREE) {
		// …
	} else if (direction != FSP_NO_DIR
		   && (ret_descr = fseg_alloc_free_extent(seg_inode, iblock,
							  &xdes, space,
							  mtr, err))) {
		// …
	} else if (UNIV_UNLIKELY(*err != DB_SUCCESS)) {
		return nullptr;
	} else {
	no_new_extent:
		// … the rest of the "else if" chain, the first one being an "if"
	}
```
Even better would be to rewrite part of the `else if` chain as `switch (xdes_get_state(descr))`. While doing that, I would reformat the function away from the archaic InnoDB style.

dr-m (commented):
I think that it could be worth the effort to reformat and refactor the entire function `fseg_alloc_free_page_low()` so that it becomes a little more understandable.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2092,7 +2104,7 @@ fseg_alloc_free_page_low(
 			first = flst_get_first(seg_inode + FSEG_FREE);
 		} else {
 			ut_ad(!has_done_reservation);
-			return(NULL);
+			return nullptr;
 		}
 
 		ret_descr = xdes_lst_get_descriptor(*space, first, mtr, &xdes);
```
dr-m, line 2110: After the code that follows this, we are testing for `ret_page == FIL_NULL` twice. I think that we’d better avoid `goto alloc_done` if the allocation fails. Simply `return nullptr` in those cases.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2041,49 +2072,30 @@ fseg_alloc_free_page_low(
...
 		if (direction == FSP_DOWN) {
 			ret_page += FSP_EXTENT_SIZE - 1;
 		}
 		ut_ad(!has_done_reservation || ret_page != FIL_NULL);
-		/*-----------------------------------------------------------*/
-	} else if (UNIV_UNLIKELY(*err != DB_SUCCESS)) {
-		return nullptr;
-	} else if ((xdes_get_state(descr) == XDES_FSEG)
-		   && mach_read_from_8(descr + XDES_ID) == seg_id
-		   && (!xdes_is_full(descr))) {
```
dr-m, line 2068: OK, this was moved to line 2023, inside the first `if` block that checks for the first 2 conditions.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2041,49 +2072,30 @@ fseg_alloc_free_page_low(
...
-		highest page in it, depending on the direction
-		==============================================*/
+	} else if (direction != FSP_NO_DIR) {
+
+		ret_descr = fseg_alloc_free_extent(seg_inode, iblock,
+						   &xdes, space, mtr, err);
+
+		if (*err != DB_SUCCESS) {
+			return nullptr;
+		}
```
dr-m, line 2082: It would be cheaper to test the return value of the function call:
```c++
if (!ret_descr) {
	ut_ad(*err != DB_SUCCESS);
	return nullptr;
}
```

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2132,32 +2144,29 @@ fseg_alloc_free_page_low(
...
-			return nullptr;
-		} else {
-			ret_page = xdes_get_offset(ret_descr);
-		}
 	}
 
+	/* 4. In worst case, try to allocate a new extent and
+	take its first page */
+	ret_descr = fseg_alloc_free_extent(seg_inode, iblock, &xdes,
+					   space, mtr, err);
```
dr-m, line 2152: If we added `} else {` for this one, we could remove `goto alloc_done` from the end of `if (reserved - used > 0) {`.

dr-m (commented):
The logic looks sound to me, but I think that we can clean this up slightly further, especially regarding the redundant computation of `FSP_EXTENT_SIZE`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -1987,29 +2004,36 @@ fseg_alloc_free_page_low(
...
-	} else if (xdes_get_state(descr) == XDES_FREE
-		   && reserved - used < reserved / FSEG_FILLFACTOR
-		   && used >= FSEG_FRAG_LIMIT) {
-
-		/* 2. We allocate the free extent from space and can take
-		=========================================================
-		the hinted page
-		===============*/
+			ret_page = hint;
+			goto got_hinted_page;
```
dr-m, line 2016: At `got_hinted_page:` we are unnecessarily checking if `ret_descr` is a null pointer. It can’t be if we are coming from here. Because this is the only occurrence of `goto got_hinted_page`, the label can be moved inside the `if` body. Please do that.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -1987,29 +2004,36 @@ fseg_alloc_free_page_low(
...
+				return nullptr;
+			}
+			ret_page += xdes_get_offset(ret_descr);
+			goto alloc_done;
+		}
+	}
+
+	/* Check whether the segment needs the new extent.
+	In that case, allocate the new extent and find the page */
+	if (!need_for_new_extent(used, reserved, extent_size)) {
```
dr-m, line 2033: Now that the function is reduced to a rather simple condition and it is only being invoked here, I think that we could simply write the expression here.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2132,33 +2135,30 @@ fseg_alloc_free_page_low(
...
 		}
 	}
-
+alloc_done:
 	if (ret_page == FIL_NULL) {
 		/* Page could not be allocated */
 
 		ut_ad(!has_done_reservation);
-		return(NULL);
+		return nullptr;
```
dr-m, line 2155: Further down the function body, there still are two occurrences of `return(NULL)`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2132,33 +2135,30 @@ fseg_alloc_free_page_low(
...
 			ut_ad(!has_done_reservation || *err);
 			return nullptr;
 		} else {
 			ret_page = xdes_get_offset(ret_descr);
 		}
 	}
-
+alloc_done:
 	if (ret_page == FIL_NULL) {
 		/* Page could not be allocated */
```
dr-m, line 2152: The label should be placed *after* the `if` body, because each `goto alloc_done` has already ensured that `ret_page != FIL_NULL`.

storage/innobase/row/row0import.cc
```diff
@@ -610,7 +610,7 @@ class AbstractCallback
 
 		if (m_xdes != 0) {
 			const xdes_t*	xdesc = xdes(page_no, m_xdes);
-			ulint		pos = page_no % FSP_EXTENT_SIZE;
+			uint32_t	pos = page_no % FSP_EXTENT_SIZE;
```
dr-m, line 613: We could cache the value of `FSP_EXTENT_SIZE` in a data member `AbstractCallback::extent_size`. But, since the `ALTER TABLE…IMPORT TABLESPACE` functionality is rarely used, I would not expand the scope of this fix with that.

dr-m (approved):
Thank you, this looks OK, with a few minor suggestions to improve.

PR #3093, opened 2024-02-29
MDEV-32637 Implement native UUID7 function

StefanoPetrilli:
- [x] *The Jira issue number for this PR is: MDEV-32637*

## Description
MariaDB provides the UUID method that generates UUID v1, however, UUID v1 generates non-linear UUIDS that will spread data across different linear partitions based on key or hash.

The new UUID7 (https://uuid7.com/) will guarantee that continuous records are time-sorted and a good collision resistance.

The main advantage of the UUID7 function is that it will allow MariaDB to avoid the limitation of using a unique combined primary key for time series records, where a unique ID and a date are required to create key/hash partitions. With UUID7 is possible to only use the key as ID and time reference.

## Release Notes

Adds the function:
`UUIDv7()`: returns a UUIDv7

## How can this PR be tested?

This PR adds automatic tests that partially test the new functions. 

I am not aware of an easy way to change the time retrieved by `my_hrtime()`, hence, I could not write automatic tests to verify that the UUIDs generated are monotonic.
I am open to suggestions on how to accomplish that.

## PR quality check


mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,157 @@
...
+{
+  uint64 now;
+
+  if (my_uuid_v7_inited)
+    return;
+  my_uuid_v7_inited= 1;
+  now= my_interval_timer()/100;
+  milliseq= 0;
+
+  my_rnd_init(&uuid_rand, (uint32) (seed1 + now), (uint32) (now/2+ getpid()));
```
grooverdan, line 68: I'm concerted by `getpid()` that in container will always be 1. Even non-container it could be fairly often be from mariadb started at boot and be in a rather narrow range of pids. I don't think it offers much.

grooverdan, line 68: `my_rnd` is of questionable autenticity as pseudo random generator and [6.9](https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-14.html#unguessability) of the RFC is a "SHOULD utilize a cryptographically secure pseudo-random number generator (CSPRNG)".

`my_random_bytes` is therfore recommended to utiilize a crypto libraries implementation.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,157 @@
...
+  {
+    /*
+      Current time is ahead of last timestamp, as it should be.
+      If we "borrowed time", give it back, just as long as we
+      stay ahead of the previous timestamp.
+    */
+    if (milliseq)
+    {
+      uint64 delta;
+      DBUG_ASSERT((tv > uuid_time) && (milliseq > 0));
```
grooverdan, line 99: `millisec > 0` is implied by the `if` condition.

Is the absolutely no possibility of and overflow while this is here? Its looking like a risky assertion to me, but I could be missing something.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,157 @@
...
+void my_uuid_v7(uchar *to)
+{
+  uint64 tv, unix_ts_ms;
+  uint16 rand_a, version_and_rand_a;
+  uint64 rand_b, variant_and_rand_b;
+
+  DBUG_ASSERT(my_uuid_v7_inited);
+
+  mysql_mutex_lock(&LOCK_uuid_v7_generator);
+  tv= my_hrtime().val / 1000;
```
grooverdan, line 87: put lock after this statement to mimise actions performed under lock.

I'm assuming the lock covers `uuid_time` and `miliseq` right? Document near lock maybe.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,157 @@
...
+      If several requests for UUIDs end up on the same tick
+      or if the system clock is turned *back*.
+      We add milli-seconds to make them different.
+      ( current_timestamp + milliseq * calls_in_this_period )
+      may end up > next_timestamp; this is OK. Nonetheless, we'll
+      try to unwind milliseq when we get a chance to.
+    */
+    while (likely(tv <= uuid_time)) {
+      ++milliseq;
+      ++tv;
```
grooverdan, line 120: this looks like it should be maths rather than a while loop to resolve.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,157 @@
...
+  implements Universal Unique Identifiers version 7, as described in
+  draft-ietf-uuidrev-rfc4122bis-14.
+
+  A UUIDv7 has the following structure:
+
+  Field                     Octet #          Note
+ unix_ts_ms                 0-5     Big-endian unsigned number of
+                                    Unix epoch timestamp in
+                                    milliseconds.
+ version_and_rand_a         6-7     Pseudo-random 12 bits multiplexed
```
grooverdan, line 26: CSPRNG per RFC.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   In the case of rand_a and rand_b 0xFFFF would not be correct as
+   the random component could overflow respectively into the version
+   or variant space.
+   By using 0x3FFF the 2 most significant bits, (the ones actually
+   occupied by version and varaint informations) are granted to be
+   zero before the multiplexing.
+  */
+  rand_a = (uint16)(my_rnd(&uuid_rand)*0x3FFF);
```
grooverdan, line 135: `my_random_bytes`

plugin/type_uuid/item_uuidfunc.h
```diff
@@ -64,4 +64,48 @@ class Item_func_uuid: public Item_func_sys_guid
...
+  {
+    collation.set(DTCollation_numeric());
+    fix_char_length(uuid_len());
+    return FALSE;
+  }
+  bool const_item() const override { return false; }
+  table_map used_tables() const override { return RAND_TABLE_BIT; }
+  LEX_CSTRING func_name_cstring() const override
+  {
+    static LEX_CSTRING name= {STRING_WITH_LEN("sys_guidv7") };
```
grooverdan, line 85: I'm wondering why add `sys_guidv7` is the compatible with something else? `sys_guid` afaik is an oracle function so we just provided the equivalent. I'm not seeing a good case for including the alias for `sys_guidv7`.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,157 @@
+/* Copyright (c) 2024, MariaDB
```
grooverdan, line 1: The CLA provides you joint copyright, you don't just have to list MariaDB (unspecified foundation/corporation)

plugin/type_uuid/mysql-test/type_uuid/func_sys_guidv7_plugin.test
```diff
@@ -0,0 +1,30 @@
+--echo #
+--echo # Start of 10.5 tests
```
grooverdan, line 2: a) 11.5
b) the convention to use "Start of .. tests" is limited. End of tests (with right version) is the main factor for easy of merging.

plugin/type_uuid/plugin.cc
```diff
@@ -199,5 +237,35 @@ maria_declare_plugin(type_uuid)
   NULL,                         // System variables
   "1.0",                        // String version representation
   MariaDB_PLUGIN_MATURITY_STABLE// Maturity(see include/mysql/plugin.h)*/
+},
+{
+  MariaDB_FUNCTION_PLUGIN,      // the plugin type (see include/mysql/plugin.h)
+  &plugin_descriptor_function_uuid_v7, // pointer to type-specific plugin descriptor
+  "uuidv7",                     // plugin name
+  "MariaDB Corporation",        // plugin author
```
grooverdan, line 245: Aren't you the author :-)

plugin/type_uuid/plugin.cc
```diff
@@ -199,5 +237,35 @@ maria_declare_plugin(type_uuid)
...
+  NULL,                         // Status variables
+  NULL,                         // System variables
+  "1.0",                        // String version representation
+  MariaDB_PLUGIN_MATURITY_EXPERIMENTAL// Maturity(see include/mysql/plugin.h)*/
+},
+{
+  MariaDB_FUNCTION_PLUGIN,      // the plugin type (see include/mysql/plugin.h)
+  &plugin_descriptor_function_sys_guid_v7,// pointer to type-specific plugin descriptor
+  "sys_guidv7",                // plugin name
+  "MariaDB Corporation",        // plugin author
```
grooverdan, line 260: omit if removing sys_guidv7. Otherwise give yourself credit.

sql/mysqld.cc
```diff
@@ -4846,6 +4847,7 @@ static int init_server_components()
   }
 
   my_uuid_init((ulong) (my_rnd(&sql_rand))*12345,12345);
+  my_uuid_v7_init((uint32) (my_rnd(&sql_rand))*12345,12345);
```
grooverdan, line 4850: this kind of numering is why I question randomness.

plugin/type_uuid/mysql-test/type_uuid/func_uuidv7_plugin.test
```diff
@@ -0,0 +1,30 @@
...
+  PLUGIN_LICENSE,
+  PLUGIN_MATURITY,
+  PLUGIN_AUTH_VERSION
+FROM INFORMATION_SCHEMA.PLUGINS
+WHERE PLUGIN_TYPE='FUNCTION'
+  AND PLUGIN_NAME IN
+    ('uuidv7')
+ORDER BY PLUGIN_NAME;
+--horizontal_results
+
```
grooverdan, line 27: There should be some tests that actually call `UUIDv7`

plugin/type_uuid/mysql-test/type_uuid/order.test.

While a straight SELECT On the result will result in unpredicaable output something like inserting:
```
CREATE TABLE t1 (a int primary key not null, u UUID DEFAULT UUIDv7(), unique key(u));
insert into t1(a) select seq from seq_1_to_10;
select a from t1 order by u asc;
```
Ensuring uniqueness and the order of generation is ascending.

Could also pull the version/variant from the generated values and ensure they are correct.

plugin/type_uuid/mysql-test/type_uuid/func_uuidv7_plugin.test
```diff
@@ -0,0 +1,30 @@
...
+--echo #
+--echo # End of 10.5 tests
```
grooverdan, line 29: 11.5

plugin/type_uuid/mysql-test/type_uuid/func_uuidv7_plugin.test
grooverdan: Also tests involving replication like mysql-test/suite/rpl/include/rpl_row_UUID.test.

grooverdan (commented):
Well done on the depth of initial implementation.

mysql-test/main/func_misc.test
```diff
@@ -18,12 +18,12 @@ select hex(inet_aton('127.1'));
 select hex(inet_aton('127.1.1'));
 
 select length(uuid()), charset(uuid()), length(unhex(replace(uuid(),_utf8'-',_utf8'')));
+select length(uuidv7()), charset(uuidv7()), length(unhex(replace(uuidv7(),_utf8'-',_utf8'')));
```
dlenski, line 21: One of the key properties of [uuid7](https://uuid7.com/) is time-sortability. 

> UUIDv7 values are time-sortable, which means you can sort them in increasing order based on when they were generated. This makes time-based queries more efficient and intuitive.

- Does this test verify that property of the implementation of the `uuid7` function? 
- Does it do it in as simple and direct a way possible?

I think the answer to both of these is **no**. An incorrect implementation of `uuid7`  implementation that [always returns a fixed hexadecimal string](http://bilbo.hobbiton.org/dilbert/browse.php?y=2016&m=04&d=01) would equally well pass this test, wouldn't it?

dlenski: Consider adding something (here or in another test) that actually verifies the functional properties of the `uuid7` implementation, including key properties like time-sortability.

StefanoPetrilli: Hello Dan, look at the changes in the file `plugin/type_uuid/mysql-test/type_uuid/order.test`. This test has been suggested by Grooverdan and it tests exactly that generated UUIDv7s are time-sortable and unique. In my opinion that test should be enough and no better way of testing this property comes to my mind, but, if you have suggestions I am very open to writing more/better tests :)  

I agree that the line that you commented on is incomplete and redundant with the other test, do you think it would be better to remove it?

plugin/type_uuid/mysql-test/type_uuid/order.test
```diff
@@ -60,3 +60,14 @@ drop table t2;
...
+
+--echo #
+--echo # MDEV-32637 Implement native UUID7 function
+--echo #
+
+# Verify that UUIDv7 are time-ordered and unique
+
+CREATE TABLE t1 (a int primary key not null, u UUID DEFAULT UUIDv7(), unique key(u));
+insert into t1(a) select seq from seq_1_to_10;
+select a from t1 order by u asc;
```
dlenski, line 72: Thanks, now I understand that this part verifies the time-ordering of `UUID7` output. Excellent!

mysql-test/main/func_misc.test
dlenski: > look at the changes in the file plugin/type_uuid/mysql-test/type_uuid/order.test. This test has been suggested by Grooverdan and it tests exactly that generated UUIDv7s are time-sortable and unique.   In my opinion that test should be enough and no better way of testing this property comes to my mind, but, if you have suggestions I am very open to writing more/better tests :)

Thanks. My apologies for overlooking this in my initial review. I agree, this test is great, and very clear.

> Hello Dan, look at the changes in the file plugin/type_uuid/mysql-test/type_uuid/order.test. This test has been suggested by Grooverdan and it tests exactly that generated UUIDv7s are time-sortable and unique. In my opinion that test should be enough and no better way of testing this property comes to my mind, but, if you have suggestions I am very open to writing more/better tests :)

> I agree that the test that you commented on is incomplete and redundant with the other test, do you think it would be better to remove it?

Yes, completely agreed.  `func_misc.test` mostly checks some superficial properties of a laundry list of functions, and I don't think it's likely to catch future regressions or easy to understand in this format.

So I think it's unnecessary to add to it here.

dlenski: Looking good to me in https://github.com/MariaDB/server/pull/3093/commits/46bef2ac01750635ea81a733d7664c853154c8ae :rocket: :+1:

mysql-test/main/rpl_row_UUIDv7.test
```diff
@@ -0,0 +1,88 @@
...
+insert into t2 values(fn1(2));
+
+sync_slave_with_master;
+SHOW CREATE TABLE test.t1;
+
+# Dump the databases and so we can see if they match
+# Another method would be to use SELECT INTO OUTFILE on master,
+# then LOAD DATA INFILE in slave, and use a query to compare.
+# This would have the advantage that it would not assume
+# the system has a 'diff'
```
dlenski, line 62: Out of curiosity, are there platforms where `--diff_files` doesn't work? Windows?

There are a couple of long-existing tests which mention _both_ Windows-specific problems _and_ use `--diff-files`, but I haven't tested these on Windows myself…

- https://github.com/MariaDB/server/blob/11.3/mysql-test/main/loaddata.test#L528, https://github.com/MariaDB/server/blob/11.3/mysql-test/main/loaddata.test#L395
- https://github.com/MariaDB/server/blob/11.3/mysql-test/main/mysqltest.test#L2893, r

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,139 @@
...
+
+  // Regulate the access to uuid_time and milliseconds
+  mysql_mutex_lock(&LOCK_uuid_v7_generator);
+
+  if (likely(tv > uuid_time))
+  {
+    /*
+      Current time is ahead of last timestamp, as it should be.
+      If we "borrowed time", give it back, just as long as we
+      stay ahead of the previous timestamp.
```
dlenski, line 85: If you build up a large enough "debt" in `milliseconds`, the server process could conceivably be stopped and restarted… and then would be generating "past" `uuid7()` values due to `milliseconds` resetting to 0.

Should there perhaps be a maximum value for `milliseconds` (maybe 1000 ms?) beyond which the `uuid7` function will **block** rather than continue adding to the debt?

dlenski (approved):

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,139 @@
...
+      If several requests for UUIDs end up on the same tick
+      or if the system clock is turned *back*.
+      We add milli-seconds to make them different.
+      ( current_timestamp + milliseconds * calls_in_this_period )
+      may end up > next_timestamp; this is OK. Nonetheless, we'll
+      try to unwind milliseconds when we get a chance to.
+    */
+    milliseconds += uuid_time - tv + 1;
+    tv = uuid_time + 1;
+  }
```
dlenski, line 110: If you agree that https://github.com/MariaDB/server/pull/3093/files/46bef2ac01750635ea81a733d7664c853154c8ae#r1515136655 is a concern, I think it could be fixed here with…

```c
if (milliseconds > 1000)
{
   /* We are building up too much "borrowed time", >1000 milliseconds.
    * The UUID7() output could become non-time-sortable if the server
    * process restarts, causing `milliseconds` to reset to 0.
    *
    * Sleep until we have less than HALF of that maximum debt (500
    * milliseconds), in order to avoid repeatedly blocking and sleeping
    * on successive calls.
    */
    my_sleep((milliseconds - 500) * 1000UL);
    … recalculate milliseconds, TV …
}
```

dlenski (approved):
Other than the issue of breaking time-sortability if you build up too much "borrowed time" and then restart the server (`CREATE new_table AS SELECT UUID7(), T.* FROM table_with_1_million_rows T`), this looks really good to me. :raised_hands:

mysys/my_uuid_v7.c
StefanoPetrilli: I agree that restarting the server after having built a lot of borrowed time could break time-sortability.

A solution that I would like more would be implementing the [Replace Left-Most Random Bits with Increased Clock Precision](https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/) variation from the UUID specs.

Encoding 10 more bits of timestamp would allow us to have a resolution of microseconds instead of milliseconds. 

With the current implementation, generating 1 million UUIDv7 results in 1000 seconds of borrowed time. With the additional time precision, instead of borrowing milliseconds, we would borrow microseconds. This means that generating 1 million UUIDv7 would only borrow 1 second.
Even with my solution breaking time sortability would still be possible but it would require the database to shut down right after generating a humungous amount of UUIDv7s.

What I don't like about your solution is that it puts an upper bound to the number of UUIDv7 that can be generated in 1 second and that upper bound might be too low for some use cases. How common is it to generate more than 1000 UUIDs per second on average? If I had to speculate I would say that it is not uncommon.

What do you think? @dlenski

mysql-test/main/rpl_row_UUIDv7.test
dlenski: Got it. I suspect the source comment is obsolete.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,162 @@
...
+      /* We are building up too much borrowed time, >500 milliseconds.
+         The output could become non-time-sortable if the server
+         process restarts, causing `borrowed_microseconds` to reset to 0.
+
+         Sleep until we have HALF of that maximum debt,
+         in order to avoid repeatedly blocking and sleeping
+         on successive calls.
+      */
+      my_sleep(SLEEP_MILLISECONDS);
+      borrowed_microseconds -= SLEEP_MICROSECONDS;
```
dlenski, line 127: Naming nitpick: I would rather see `RATE_LIMIT` named something like `MAX_BORROWED_MICROSECONDS`, and just use `MAX_BORROWED_MICROSECONDS / 2` instead of having a second `#define` which could get out of sync out of time.

dlenski (approved):
Other than my small remaining nitpick about the name(s) of the constant(s), this looks fantastic. Approving! :raised_hands:

LinuxJedi (commented):
Great work! A few minor things to fix and I think we are there.

Later on (not for this Jira ticket or PR) we could add a test for this to check that the UUID timestamp is within a tolerable range of a timestamp (+/- 2 hours?) in the same column by creating a new function to extract the UUID timestamp. If we do it range based then we won't need to worry about monotonic / DST quite as much.

dlenski:
@LinuxJedi wrote:

> we could add a test for this to check that the UUID timestamp is within a tolerable range of a timestamp (+/- 2 hours?) in the same column by creating a new function to extract the UUID timestamp. If we do it range based then we won't need to worry about monotonic / DST quite as much.

Since it's based on Unix-epoch timestamp, shouldn't need to worry about DST :sweat_smile:

plugin/type_uuid/item_uuidfunc.h
```diff
@@ -64,4 +64,34 @@ class Item_func_uuid: public Item_func_sys_guid
   { return get_item_copy<Item_func_uuid>(thd, this); }
 };
 
+class Item_func_uuid_v7: public Item_str_func
+{
```
abarkov, line 68: Same comment with Iten_func_uuid_v4: please derive Item_func_uuid_v7 from Item_func_sys_guid to avoid overriding too many methods.

vuvova (changes requested):
UUID code is pretty good, just doesn't need to reside in mysys.
The function could be greatly simplified by using the approach from https://github.com/MariaDB/server/tree/bb-11.5-bar-MDEV-11339 (after rebasing this PR on top of that branch).

@StefanoPetrilli , I know you're busy now. There's no hurry, you can get back to it after GSoC. Just, please, if you decide not to continue with the PR at all — add a comment and say so.

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,171 @@
...
+
+  uuid_time= tv;
+  mysql_mutex_unlock(&LOCK_uuid_v7_generator);
+
+  mi_int6store(to, tv / 1000);
+
+  /**
+   Map all the possible microseconds values (from 0 to 999) to all the
+   values that can be represented in 12 bits (from 0 to 4095) as
+   described in section 6.2, Method 3 of draft-ietf-uuidrev-rfc4122bis-14.
```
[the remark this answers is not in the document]
StefanoPetrilli: Here, I just changed `draft-ietf-uuidrev-rfc4122bis-14` to `RFC 9562`. Was this what you meant? The section and the method described remained the same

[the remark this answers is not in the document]
vuvova: yes, exactly. thanks

mysys/my_uuid_v7.c
```diff
@@ -0,0 +1,171 @@
...
+    {
+      /* We are building up too much borrowed time, >500 milliseconds.
+         The output could become non-time-sortable if the server
+         process restarts, causing `borrowed_microseconds` to reset to 0.
+
+         Sleep until we have HALF of that maximum debt,
+         in order to avoid repeatedly blocking and sleeping
+         on successive calls.
+      */
+      my_sleep(SLEEP_MILLISECONDS);
```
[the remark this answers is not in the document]
StefanoPetrilli: This is for what is described in the RFC under the subtopic `Counter Rollover Handling`. UUIDv7 is supposed to be not only unique but also monotonic. I agree that sleeping is not a viable solution, and I implemented this differently in the last commit.

vuvova:
you've marked all my comments as resolved without actually changing anything. how comes?

vuvova:
I still don't see any changes :man_shrugging:

vuvova:
Thanks. I've cherry-picked both UUIDv4 and UUIDv7 code into bb-11.7-MDEV-33710-uuid branch to be tested and merged into main

vuvova:
pushed into 11.7, thanks!

PR #3098, opened 2024-03-04
MDEV-33588 buf::Block_hint is a performance hog

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33588*
## Description
The class `buf::Block_hint` that was added in order to fix [MDEV-23693](https://jira.mariadb.org/browse/MDEV-23693) is an unnecessary performance hog that causes unnecessary lookups on `buf_pool.page_hash`.

Originally, I was thinking that this function could be removed as part of rewriting the buffer pool resizing in [MDEV-29445](https://jira.mariadb.org/browse/MDEV-29445), but it turns out that we can do it independently of that. We only need to add a field `btr_pcur_t::old_page_id` to remember the page identifier where the persistent cursor position was stored, and pass this identifier to `buf_page_optimistic_get()`, which will have to perform a `buf_pool.page_hash` lookup anyway.

`buf_page_optimistic_fix()`: Prepare for `buf_page_optimistic_get()`. This basically is a simpler version of `Buf::Block_hint`.

`buf_page_optimistic_get()`: Assume that `buf_page_optimistic_fix()` has been called and the page identifier validated. Should the block be evicted, the `block->modify_clock` will be invalidated; we do not need to check the `block->page.id()` again. It suffices to check the `block->modify_clock` after acquiring the page latch.

`btr_pcur_optimistic_latch_leaves()`: Remove redundant parameters. First, invoke `buf_page_optimistic_get()` on the requested page. If needed, acquire a latch on the left page. Finally, acquire a latch on the target page and recheck the `block->modify_clock`, and that the left page (if any) points to the current page.

`btr_latch_prev()`: Always append the previous page to `mtr_t::m_memo` after the current page, even if we need to wait for the page latch. If the `FIL_PAGE_PREV` of the current page was changed while we were not holding a latch, fetch the previous page again. (The `block->modify_clock` is not incremented during page merges!) Add some consistency checks.

`btr_pcur_move_backward_from_page()`: Simplify the logic, now that the left block will be at the top of `mtr_t::m_memo`.

## Release Notes
Nothing needed; this is a rather small performance improvement.

## How can this PR be tested?
The function `btr_pcur_move_backward_from_page()` is *not covered* by the regression test suite. It is covered by RQG, as we learned back when working on MDEV-30400 and MDEV-29835.

This is a performance bug fix that depends on earlier substantial refactoring that is only available starting with 10.6.

## PR quality check


storage/innobase/btr/btr0pcur.cc
```diff
@@ -214,102 +212,98 @@ btr_pcur_copy_stored_position(
...
 
-      if (!left_block);
-      else if (btr_page_get_next(left_block->page.frame) != id.page_no())
+      if (mode == RW_S_LATCH
+          ? left->page.lock.s_lock_try() : left->page.lock.x_lock_try())
       {
-release_left_block:
-        mtr->release_last_page();
-        return false;
+        mtr->lock_register(savepoint + 1, mtr_memo_type_t(mode << 5));
```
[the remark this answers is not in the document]
dr-m: b56d2eb41733a2ae7daa3547928b8001838bccb3 addresses this (in another branch).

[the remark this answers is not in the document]
dr-m: There is still something wrong with `btr_pcur_move_backward_from_page()`. I hope to get an `rr replay` trace for that soon.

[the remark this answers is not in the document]
dr-m: 9b169a63666f8a850c2d10e224ac9f6a6e39ae1f hopefully addresses this by invoking `btr_latch_prev()` and by improving that function so that the preceding block is always `mtr->at_savepoint(1)`.

storage/innobase/btr/btr0cur.cc
```diff
@@ -955,47 +954,67 @@ int btr_latch_prev(buf_block_t *block, page_id_t page_id, ulint zip_size,
...
-                        PAGE_HEADER + PAGE_INDEX_ID + block->page.frame, 8) ||
-      page_is_comp(prev->page.frame) != page_is_comp(block->page.frame))
+  const page_t *const p= prev->page.frame;
+  if (memcmp_aligned<4>(FIL_PAGE_NEXT + p, FIL_PAGE_OFFSET + page, 4) ||
+      memcmp_aligned<2>(FIL_PAGE_TYPE + p, FIL_PAGE_TYPE + page, 2) ||
+      memcmp_aligned<2>(PAGE_HEADER + PAGE_INDEX_ID + p,
+                        PAGE_HEADER + PAGE_INDEX_ID + page, 8) ||
+      page_is_comp(p) != page_is_comp(page))
   {
     ut_ad("corrupted" == 0); // FIXME: remove this
```
dr-m, line 1019: Matthias reported an assertion failure here, due to the `FIL_PAGE_NEXT` check. In the `rr replay` trace that I analyzed, the cause is that while `btr_latch_prev()` was waiting for the previous page latch in `buf_page_get_gen()`, there was a `btr_page_split_and_insert()` executed in another thread.

I think that we must implement a retry logic in case the `FIL_PAGE_PREV` of our current page changed while we were not holding a page latch.

dr-m: a4389ec605fe2d027ec09850fb736078c250ce3a should address this.

mariadb-DebarunBanerjee (commented):
Hi Marko,

The patch is doing way more than removing buf::Block_hint. In fact it is making a major design change. 

I checked the design and doesn't look very promising. The current design is to buffer fix the page and take the locks previous-current-next in order. This is used for usual scan and repositioning. During reposition the only possible conflict is a concurrent pessimistic operation.

The modified design is doing a try latch on reverse order which is bound to conflict and cause repeated retry (which we nee to do) with any concurrent Query. Historically the reverse scan is also known to have performance issue.

I don't agree with the design change and I think will cause issues for us in future. May be if you can explain the reason for the performance hog,  we can improve it without changing the latch order.

Regards,
Deb

storage/innobase/btr/btr0cur.cc
dr-m: The SX latch mode was implemented in MySQL 5.7 already. I renamed it to U in the `sux_lock`. It is not free from hangs; see f2096478d5750b983f9a9cc4691d20e152dafd4a for a prominent example.

This code is mostly being used in `btr_pcur_move_backward_from_page()` but also in the change buffer (`BTR_MODIFY_PREV`).

storage/innobase/buf/buf0buf.cc
```diff
@@ -2991,39 +2991,24 @@ buf_page_get_gen(
...
   buf_pool_t::hash_chain &chain= buf_pool.page_hash.cell_get(id.fold());
   bool have_u_not_x= false;
 
   {
     transactional_shared_lock_guard<page_hash_latch> g
       {buf_pool.page_hash.lock_get(chain)};
-    if (UNIV_UNLIKELY(id != block->page.id() || !block->page.frame))
+    if (UNIV_UNLIKELY(!buf_pool.is_uncompressed(block) ||
+                      id != block->page.id() || !block->page.frame))
       return false;
```
dr-m, line 3012: I tested this with 79e3a658d905ca42e0ec3102e2334dcd182c8ec9 (#3107) to see if the `transactional_lock_guard` works well:
```sh
perf record -e tx-abort ./mtr innodb.innodb_buffer_pool_resize
perf report
```
In that branch, it is OK; we have much more `xabort()` in the buffer pool lookup functions:
```
Samples: 5K of event 'tx-abort', Event count (approx.): 101716
Overhead  Command   Shared Object  Symbol
  99,98%  mariadbd  mariadbd       [.] buf_page_get_low(page_id_t, unsigned long, unsigned long, buf_block_t*, unsigned long, mtr_t*, dberr_t*, bool)
   0,01%  mariadbd  mariadbd       [.] buf_page_create_low(page_id_t, unsigned long, mtr_t*, buf_block_t*)
   0,00%  mariadbd  mariadbd       [.] buf_page_optimistic_get(rw_lock_type_t, buf_block_t*, page_id_t, unsigned long, mtr_t*)
```
In a longer `sysbench` benchmark, there were more `xabort()` in the optimistic function:
```
Samples: 64K of event 'tx-abort', Event count (approx.): 476326
Overhead  Command   Shared Object  Symbol
  83,71%  mariadbd  mariadbd       [.] buf_page_get_low(page_id_t, unsigned long, unsigned long, buf_block_t*, unsigned long, mtr_t*, dberr_t*, bool)
  15,73%  mariadbd  mariadbd       [.] buf_page_optimistic_get(rw_lock_type_t, buf_block_t*, page_id_t, unsigned long, mtr_t*)
   0,33%  mariadbd  mariadbd       [.] TMLockGuard::TMLockGuard(lock_sys_t::hash_table&, page_id_t)
   0,10%  mariadbd  mariadbd       [.] buf_page_create_low(page_id_t, unsigned long, mtr_t*, buf_block_t*)
   0,08%  mariadbd  mariadbd       [.] buf_page_free(fil_space_t*, unsigned int, mtr_t*)
   0,05%  mariadbd  mariadbd       [.] trx_t::commit_low(mtr_t*)
```
For this branch, the traversal of `buf_pool.chunks` could make the memory transaction too big. I reran the same workload on a merge of 4ac8c4c820ebcff3571a2c67acc4fc41510b2d33 and a4389ec605fe2d027ec09850fb736078c250ce3a, and got this:
```
Samples: 18K of event 'tx-abort', Event count (approx.): 278138
Overhead  Command   Shared Object  Symbol
  99,55%  mariadbd  mariadbd       [.] buf_page_get_low(page_id_t, unsigned long, unsigned long, buf_block_t*, unsigned long, mtr_t*, dberr_t*, bool)
   0,20%  mariadbd  mariadbd       [.] buf_page_init_for_read(unsigned long, page_id_t, unsigned long, bool)
   0,13%  mariadbd  mariadbd       [.] buf_page_create_low(page_id_t, unsigned long, mtr_t*, buf_block_t*)
   0,06%  mariadbd  mariadbd       [.] buf_page_optimistic_get(rw_lock_type_t, buf_block_t*, page_id_t, unsigned long, mtr_t*)
   0,04%  mariadbd  mariadbd       [.] trx_t::commit_low(mtr_t*)
   0,02%  mariadbd  mariadbd       [.] TMLockGuard::TMLockGuard(lock_sys_t::hash_table&, page_id_t)
```
As we can see, the does not appear to be any issue with the `buf_page_optimistic_get()` here either. There are fewer samples, though: The server did not complete the `innodb_fast_shutdown=0`. I had to kill the process because it hung due to running out of buffer pool. With #3107 we would abort the buffer pool resize in that scenario.

dr-m: The removed `Block_hint::buffer_fix_block_if_still_valid()` was also essentially performing a `buf_pool.page_hash` lookup, for acquiring a latch on the hash table slice.

One thing that we can do better is to only buffer-fix the block while holding `buf_pool.page_hash`, and try to acquire the `buf_page_t::lock` after we have released the `buf_pool.page_hash`. The buffer-fix will prevent eviction.

dr-m: With the ccd4b967b4890fe235a638317489e8876855a23e simplification of `buf_page_optimistic_get()`, the throughput numbers for my 60-second 64-thread 64×100k row Sysbench `oltp_update_index` were virtually unchanged between that commit and its parent.

dr-m:
@mariadb-DebarunBanerjee, thank you for the review. c10fd5d9797fa5d7629c0e07fe5da46f5ba5013d addresses some of your concerns. Please re-review.

The retry logic in case `FIL_PAGE_PREV` was changed is addressing something for which I got an `rr replay` trace. Because `buf_block_t::modify_clock` is not being incremented when `FIL_PAGE_PREV` is changed, we’d better validate that field. I originally had 91a5e4f8b3eb65dba5f3be4f1660e1f07bef2b9b included in the revised fix, and in my 60-second, 64-thread Sysbench `oltp_update_index` test the throughput was some 0.5% better than 10.6.

storage/innobase/btr/btr0pcur.cc
```diff
@@ -214,101 +212,68 @@ btr_pcur_copy_stored_position(
...
+    return false;
   }
-};
+
+  buf_block_t *prev= nullptr;
+  if (left_page_no != FIL_NULL &&
+      !(prev= btr_block_get(*pcur->index(), left_page_no, mode, true, mtr)))
+    goto fail;
+
+  mtr->upgrade_buffer_fix(savepoint, mode);
```
dr-m, line 267: @mleich1 provided a core dump where we have the `block->page.zip.fix=FREED+1` inside `mtr_t::upgrade_buffer_fix()`. It seems to me that we must relax the assertion inside that function. After acquiring the latch, we must backtrack if the page had been marked as `FREED` in the data file. Callers must check the status of `mtr_t::upgrade_buffer_fix()` and act accordingly.

dr-m (commented):
As discussed today, I must try to create an `mtr` test case that covers the `btr_pcur_move_backward_from_page()` code path.

storage/innobase/btr/btr0pcur.cc
dr-m: We can simply relax the assertion and check for `block->page.is_freed()` here. Additionally, `buf_page_get_low()` was missing a check of `FREED` after acquiring the page latch.

storage/innobase/buf/buf0buf.cc
```diff
@@ -2845,42 +2839,43 @@ buf_page_get_low(
...
 			return block;
 		case RW_S_LATCH:
 			block->page.lock.s_lock();
-			ut_ad(!block->page.is_read_fixed());
-			if (UNIV_UNLIKELY(block->page.id() != page_id)) {
-				block->page.lock.s_unlock();
-				block->page.lock.x_lock();
-				goto page_id_mismatch;
-			}
-get_latch_valid:
```
mariadb-DebarunBanerjee, line 2870: For my edification, why is the page ID mismatch check is now not needed in the "else" part of the "if" condition ?

```
 if (state >= buf_page_t::UNFIXED
            && allow_ibuf_merge
            && fil_page_get_type(block->page.frame) == FIL_PAGE_INDEX
            && page_is_leaf(block->page.frame)) {

```

storage/innobase/btr/btr0cur.cc
```diff
@@ -946,56 +946,89 @@ static MY_ATTRIBUTE((nonnull))
...
+      /* the page was just read and found to be corrupted */
+      mtr->rollback_to_savepoint(prev_savepoint);
+      return 0;
     }
-    block->page.lock.s_unlock();
   }
   else
   {
-    if (UNIV_LIKELY(prev->page.lock.x_lock_try()))
+    mtr->release_last_page();
```
mariadb-DebarunBanerjee, line 996: If I am correct, we should be releasing the lock on the block/page that is passed to this function.
Can we possibly assert it ? May be release_last_page() can return the page ID that it actually released.

mariadb-DebarunBanerjee (approved):
@dr-m The latest changes look good to me. It changes the design and hence there is some risk but I think it is for good. Let's go ahead.

Added some minor comments.

storage/innobase/btr/btr0pcur.cc
mariadb-DebarunBanerjee: Make sense. Perhaps we should check other callers of upgrade_buffer_fix too.

storage/innobase/buf/buf0buf.cc
dr-m: The function `buf_pool_t::corrupted_evict()` does two things before releasing the exclusive page latch: Invalidate the page ID, and set the `state` to `FREED` while preserving the buffer-fix count. In the `else` branch we detect the latter:
```c++
		if (UNIV_UNLIKELY(state < buf_page_t::UNFIXED)) {
			mtr->release_last_page();
			goto ignore_unfixed;
		}
```
In the `if` branch the condition for page ID mismatch could in fact be redundant; the `goto ignore_block` or `goto ignore_unfixed` for the `state` mismatch should do the same. It is the responsibility of `buf_page_t::read_complete()` to invoke `buf_pool_t::corrupted_evict()` when needed. I will revise the logic and remove the redundant check for page ID mismatch.

storage/innobase/btr/btr0pcur.cc
dr-m: All other callers of `mtr_t::upgrade_buffer_fix()` are in `SPATIAL INDEX` code, specifically in `rtr_latch_leaves()` and `rtr_search_to_nth_level()`. At least in the latter we may invoke `buf_page_get_gen()` with `rw_latch=RW_NO_LATCH`, and we may be accessing a page that is in the process of being read into the buffer pool. `rtr_latch_leaves()` is also being called from `rtr_pcur_getnext_from_path()`, which likewise may invoke `buf_page_get_gen()` with `rw_latch=RW_NO_LATCH`.

It would be tempting to always acquire (and if applicable, release) the page latch in `buf_page_get_low()`. Unfortunately, doing so would obviously result in a deadlock in `btr_latch_prev()`. Outside that function and the page eviction in `PageConverter::operator()`, the only real user of `RW_NO_LATCH` is `SPATIAL INDEX`. I will try to see how to fix that. Yes, `SPATIAL INDEX` are known to be broken already for other reasons, but we should avoid wilingly breaking it further.

storage/innobase/buf/buf0buf.cc
```diff
@@ -2804,83 +2814,78 @@ buf_page_get_low(
...
 		switch (rw_latch) {
 		case RW_NO_LATCH:
+			/* Our caller is responsible for checking
+			block->page.is_freed(), that is, whether
+			buf_pool_t::corrupted_evict() had been called,
+			after acquiring the page latch, unless it is
+			known that the block was already successfully
+			loaded into the buffer pool. */
 			mtr->memo_push(block, MTR_MEMO_BUF_FIX);
 			return block;
```
dr-m, line 2881: Sorry, this comment should not have been added. It turns out that we will already have acquired and released a page latch in case the block was read-fixed. So, at this point the block is guaranteed not to be corrupted, and the callers (mainly `rtr_search_to_nth_level()` and `rtr_pcur_getnext_from_path()`) should actually be safe.

storage/innobase/btr/btr0pcur.cc
dr-m: The `btr_block_get()` call in `btr_pcur_optimistic_latch_leaves()` would trigger debug assertion failures when the page had been freed in the mean time. The problem is that `btr_block_get()` is using `mode=BUF_GET` instead of `BUF_GET_POSSIBLY_FREED`. In non-debug builds there is no issue, because `nullptr` would be returned, which we are dealing with. I will revise this.

storage/innobase/buf/buf0buf.cc
dr-m: Removed in 3a570f75d4e1d5ad2e55189b60fe77437e71959c

storage/innobase/btr/btr0cur.cc
dr-m: Added in 5c450f7e2efb429e44fc17190332f5ccaecd792e

storage/innobase/btr/btr0pcur.cc
dr-m: 3a570f75d4e1d5ad2e55189b60fe77437e71959c replaces the `btr_block_get()`

PR #3101, opened 2024-03-05
MDEV-33592: Use X509v3 for compatibility with libraries

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-33592*

## Description

According to rfc5280, x509v3 was standardised in June 1996.

RusTLS only accepts v3 certificates rejected the v1 default. (ref:
https://github.com/rustls/webpki/issues/29#issuecomment-1453783741)

It seems reasonable that all client libraries can accept v3.

Thanks Austin Bonander for the suggested patch.

## Release Notes

## How can this PR be tested?

covered by main.ssl_autoverify

## PR quality check


LinuxJedi:
I would go as far as to say this is a bug. We should in no point in MariaDB's history, have been issuing V1 certs.

LinuxJedi (approved):
Deleted my previous comment as I see 11.4 was where this code was introduced.

Nice fix. Very likely an accident when this was added.

PR #3106, opened 2024-03-06
MDEV-30676 rpl.parallel_backup* tests sometimes fail

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-30676*

## Description

Raise wait_condition timeout from default (30 seconds) to 60.

## Release Notes

nothing significant.

## How can this PR be tested?

## PR quality check


mysql-test/suite/rpl/t/parallel_backup.test
```diff
@@ -41,6 +41,7 @@ INSERT INTO t1 VALUES (1);
 --source include/start_slave.inc
 
 --connection aux_slave
+--let $wait_timeout= 60
```
bnestere, line 44: Looking at the failure reports in MDEV-30676, it looks like the timed out `wait_condition.inc` actually comes from those from within `parallel_backup_xa.inc`. Specifically, it looks like this happens when `$slave_ooo_error= 1`, and the `innodb_lock_wait_timeout` is too quick. That is, if the innodb lock wait timeout occurs before the test is able to run (around line `53` of parallel_backup_xa.inc):
```
--connection backup_slave
  BACKUP STAGE START;
--send BACKUP STAGE BLOCK_COMMIT
```
we get the reported failures. I'd say to increase that timeout by a couple seconds instead.

bnestere (changes requested):
Hi @grooverdan!

Thanks for the fix! See my comment.

PR #3107, opened 2024-03-06
MDEV-29445 reorganise innodb buffer pool (and remove buffer pool chunks)

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-29445*

## Description
The buffer pool will be mapped in a contiguous memory area that will be aligned and divided to extents of 8 MiB on 64-bit systems and 2 MiB on 32-bit systems.

Within an extent, the first few `innodb_page_size pages` contain `buf_block_t` objects that will cover the page frames in the rest
of the extent. In this way, there is a trivial mapping between page frames and block descriptors and we do not need any lookup tables like `buf_pool.zip_hash` or `buf_pool_t::chunk_t::map`.

We will always allocate the same number of page frames for an extent, even if we do not need all the `buf_block_t` in the last extent in case the `innodb_buffer_pool_size` is not an integer multiple of the of extent size.

`my_large_virtual_alloc()`: A new function, similar to `my_large_malloc()`.
FIXME: On Microsoft Windows, let the caller know if large page allocation was used. In that case, we must disallow buffer pool resizing.

`buf_pool_t::create()`: Only initialize the first page descriptor of each chunk.

`buf_pool_t::lazy_allocate()`: Lazily initialize a previously allocated page descriptor and increase `buf_pool.n_blocks`, which must be below `buf_pool.n_blocks_alloc`.

`innodb_init_param()`: Refactored. We must first validate `innodb_page_size` and then determine the valid bounds of
`innodb_buffer_pool_size`.

## Release Notes
We deprecate and ignore the parameter `innodb_buffer_pool_chunk_size` and let the buffer pool size to be changed in arbitrary 1-megabyte increments, all the way up to `innodb_buffer_pool_size_max`, which must be specified at startup.

If `innodb_buffer_pool_size_max` is not specified, it will default to twice the specified `innodb_buffer_pool_size`.

The minimum `innodb_buffer_pool_size` is 320 pages. At the default `innodb_page_size=16k` this corresponds to 5 MiB. However, now that the `innodb_buffer_pool_size` includes the memory allocated for the block descriptors, the minimum would now be `innodb_buffer_pool_size=6m`.

`Innodb_buffer_pool_resize_status` will be removed. The `SET GLOBAL innodb_buffer_pool_size` operation will block until the buffer pool has been resized or the operation aborted by `KILL`, `SHUTDOWN` or disconnect.

## How can this PR be tested?
This is mostly covered by the regression test suite.

Large-page allocation needs to be tested on Linux and Microsoft Windows. Stress testing with lots of buffer pool resizing during the workload would be helpful.

This is a new feature, but it depends on some refactoring that is currently based on the 10.6 branch.

## PR quality check


vaintroub:
There needs to be some proper mysys exported functionality, for manipulating virtual memory
I'm aware, that Linux is missing functionality for some of it, but it provides something similar.
I think the concept of "commited" memory is universal, rather than Windows-ism, but in case it is unfamiliar, commited memory is something backed by physical memory or page file/ swap ( https://stackoverflow.com/a/25911596/547065 )

Personally, I'd prefer 5 functions, but it can be made 2, if taking VirtualAlloc as inspiration.

- reserve_virtual_memory(size) 
  Reserve virtual memory address range only. Accessing memory is an error, and should crash
  Implementation : VirtualAlloc(MEM_RESERVE), mmap (w. asan/MEM_UNDEFINED)

- commit_virtual_memory(ptr, size) .
  Commit previously reserved region of memory. It is possible to read/write into region, after it, it is now backed by pagefile/swap
  Implementation : VirtualAlloc(MEM_COMMIT), madvise/WILL_NEED, asan/MEM_DEFINED

- allocate_virtual_memory(size, bool large_pages)
   Reserve and commit in one step. This is the only function that works on Windows, with large pages
   Implementation : VirtualAlloc(MEM_COMMIT|MEM_RESERVE) possibly with MEM_LARGE_PAGES, 
   mmap.

-decommit_virtual_memory(ptr,size)
  Decommit virtual memory address range only. Accessing memory is an error again, and should crash 
  Implementation : VirtualFree(MEM_DECOMMIT), madvise(WONT_NEED), asan/MEM_UNDEFINED

-free_virtual_memory(ptr)
  Implementation: VirtualFree(MEM_RELEASE)/munmap

storage/innobase/btr/btr0sea.cc
```diff
@@ -218,8 +218,7 @@ ATTRIBUTE_COLD static void btr_search_lazy_free(dict_index_t *index)
   table->autoinc_mutex.wr_unlock();
 }
 
-/** Disable the adaptive hash search system and empty the index. */
```
lottaquestions, line 221: Just thinking it would be helpful to others reading the code to have the comment, even though AHI is generally disabled in MariaDB

storage/innobase/buf/buf0dump.cc
```diff
@@ -58,7 +58,7 @@ take after being waked up. */
 static volatile bool	buf_dump_should_start;
 static volatile bool	buf_load_should_start;
 
-static bool	buf_load_abort_flag;
+static Atomic_relaxed<bool> buf_load_abort_flag;
```
lottaquestions, line 61: Are there any downsides to relaxed memory ordering for this flag especially with regards to other operations?

storage/innobase/dict/dict0load.cc
```diff
@@ -2479,17 +2479,20 @@ static dict_table_t *dict_load_table_one(const span<const char> &name,
...
-				|| page_get_page_no(block->page.frame)
-				!= page_id.page_no()
-				|| (mach_read_from_2(FIL_PAGE_TYPE
-						    + block->page.frame)
-				    != FIL_PAGE_INDEX
-				    && mach_read_from_2(FIL_PAGE_TYPE
-							+ block->page.frame)
-				    != FIL_PAGE_TYPE_INSTANT);
+			bool corrupted = !block;
+			if (block) {
```
lottaquestions, line 2483: What was the downside of the old code? It used short-circuiting to check if block was valid before attempting operations on it. Unless there was some out of order execution going on, which I believe should not happen for logic operations, it was fine as it was. The new code that uses an if offers a chance for branch misprediction IMHO.

lottaquestions (commented):
Had a few questions mostly for my own understanding.

storage/innobase/buf/buf0dump.cc
dr-m: It depends on the memory ordering: https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html

storage/innobase/btr/btr0sea.cc
dr-m: The AHI was rewritten in #3562.

storage/innobase/dict/dict0load.cc
dr-m: This was covered in #3579.

dr-m:
This was closed in #3826.

PR #3111, opened 2024-03-06
MDEV-32633: Fix Galera cluster <-> native replication interaction

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-32633*

## Description
It's possible to establish Galera multi-cluster setups connected through the native replication when every Galera cluster is configured to have a separate domain ID.
For this setup to work, we need to replace domain ID values in generated GTID events when they are written at transaction commit to the values configured by Wsrep replication.

At the same time, it's possible that the GTID event already contains a correct domain ID if it comes through the native replication from another Galera cluster.
In this case, when such an event is applied either through a native replication slave thread or through Wsrep applier, we write GTID event on transaction start and avoid writing it during transaction commit.

The code contained multiple problems that were fixed:
- applying GTID events didn't work because it's applied without a running server transaction and Wsrep transaction was not started
- GTID event generation on transaction start didn't contain proper "standalone" and "is_transactional" flags that the original applied GTID event contained
- condition determining that GTID event is written on transaction start to avoid writing it on commit relied on the fact that the GTID event is the first found in transaction/statement caches, which wasn't the case and resulted in duplicate GTID events written
- instead of relying on the caches to find a GTID event, a simple check is introduced that follows the exact rules for checking if event is written at transaction start as described above
- the test case is improved to check that exact GTID events are applied after two Galera clusters have synced.

## Release Notes
This fix is 10.4 version-only.

## How can this PR be tested?

Re-enabled previously failing MTR test.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera_3nodes/disabled.def
+++ b/mysql-test/suite/galera_3nodes/disabled.def
@@ -11,7 +11,6 @@
 ##############################################################################
 
 galera_2_cluster : MDEV-32631 galera_2_cluster: before_rollback(): Assertion `0' failed
-galera_gtid_2_cluster : MDEV-32633 galera_gtid_2_cluster: Assertion `thd->wsrep_next_trx_id() != (0x7fffffffffffffffLL * 2ULL + 1)'
 galera_ssl_reload : MDEV-32778 galera_ssl_reload failed with warning message
 galera_ipv6_mariabackup : temporarily disabled at the request of Codership
 galera_pc_bootstrap : temporarily disabled at the request of Codership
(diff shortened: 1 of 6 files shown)


sysprg:
Thanks, the fix has been merged with the head revision, but for version 10.5+, since version 10.4 no longer in active development:
https://github.com/MariaDB/server/commit/0cc9b49751fc86f7942e19fe6ed8f7227f847c02
https://github.com/MariaDB/server/commit/a4838721a252ea5570b5fed9ab56e38e5b234865
https://github.com/MariaDB/server/commit/c21aa486a86d9db29ad16ff279380969443df00f

PR #3112, opened 2024-03-07
MDEV-33540 Avoid writes to TRX_SYS page during mariabackup operations

sciascid:
Fix a scenario where `mariabackup --prepare` fails with assertion
`!m_modifications || !recv_no_log_write` in `mtr_t::commit()`. This
happens if the prepare step of the backup encounters a data directory
which happens to store wsrep xid position in TRX SYS page ( is no
longer the case since 10.3.5). And since MDEV-17458,
`trx_rseg_array_init()` handles this case by copying the xid position
to rollback segments, before clearing the xid from TRX SYS page.
However, this step should be avoided when `trx_rseg_array_init()` is
invoked from mariabackup. The relevant code was surrounded by the
condition `srv_operation == SRV_OPERATION_NORMAL`. An additional check
ensures that we are not trying to copy a xid position which has
already zeroed.

### DIFF
--- a/storage/innobase/trx/trx0rseg.cc
+++ b/storage/innobase/trx/trx0rseg.cc
@@ -207,6 +207,11 @@ bool trx_rseg_read_wsrep_checkpoint(const trx_rsegf_t* rseg_header, XID& xid)
 @return	whether the WSREP XID is present */
 static bool trx_rseg_init_wsrep_xid(const page_t* page, XID& xid)
 {
+	if (memcmp(TRX_SYS + TRX_SYS_WSREP_XID_INFO + page,
+		           field_ref_zero, TRX_SYS_WSREP_XID_LEN) == 0) {
+		return false;
+	}
+
 	if (mach_read_from_4(TRX_SYS + TRX_SYS_WSREP_XID_INFO
 			     + TRX_SYS_WSREP_XID_MAGIC_N_FLD
 			     + page)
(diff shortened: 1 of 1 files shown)


storage/innobase/trx/trx0rseg.cc
```diff
@@ -662,22 +662,26 @@ dberr_t trx_rseg_array_init()
...
+	if (srv_operation == SRV_OPERATION_NORMAL && !wsrep_sys_xid.is_null()) {
 		/* Upgrade from a version prior to 10.3.5,
 		where WSREP XID was stored in TRX_SYS page.
 		If no rollback segment has a WSREP XID set,
 		we must copy the XID found in TRX_SYS page
 		to rollback segments. */
 		mtr.start();
+		const buf_block_t *sys = trx_sysf_get(&mtr);
+		if (memcmp(TRX_SYS + TRX_SYS_WSREP_XID_INFO + sys->frame,
+		           field_ref_zero, TRX_SYS_WSREP_XID_LEN) != 0) {
```
dr-m, line 674: In the body of `if (rseg_id == 0)` earlier, we could check if the `sys` page needs to be updated. In that way we could avoid an extra buffer pool lookup as well as the indentation changes here.
```c++
	if (wsrep_xid_in_sys && srv_operation == SRV_OPERATION_NORMAL
	    && !wsrep_sys_xid.is_null()) {
```

dr-m (commented):
Can you please rename the bug title to something more descriptive? There is no such error message present in 10.4.

dr-m (approved):
Thank you. I assume that you tested this with a backup that was created from MariaDB Server 10.2 or earlier.

PR #3113, opened 2024-03-07
MTR flag to mark tests as incompatible with macOS

DaveGosselin-MariaDB:
Introduces a new MTR include, not_mac.inc, which when included at the top of a test, prevents that test from running on macOS.

There are many tests that don't work with macOS presently and they are not excluded in this commit; however, one is, as an example.  The remaining tests will be fixed or marked as incompatible in future commits.


DaveGosselin-MariaDB:
Evidently there is something missing from this patch.

DaveGosselin-MariaDB:
OK it works now.

mysql-test/main/json_normalize.test
```diff
@@ -1,3 +1,6 @@
+# Test doesn't work on macOS.
```
tonychen2001, line 1: Minor suggestion:
This is redundant as including `not_mac.inc` already implies this.
Perhaps a brief explanation of why this test doesn't work on macOS as a comment or in the commit description would be better.

Apart from that, looks good.

DaveGosselin-MariaDB: Hi Tony, thank you for the suggestion.  I haven't (yet) investigated why this test doesn't work on macOS, but I will.  Just not for this commit, as it's purpose is to demonstrate how to use `not_mac.inc`.

grooverdan (commented):
Is mac generating different results than unix? Is so might need:
```
$ git diff mysql-test/
diff --git a/mysql-test/include/platform.combinations b/mysql-test/include/platform.combinations
index 4f0660b7a40..672f78e1fc5 100644
--- a/mysql-test/include/platform.combinations
+++ b/mysql-test/include/platform.combinations
@@ -4,3 +4,4 @@
 
 [unix]
 
+[mac]
diff --git a/mysql-test/suite.pm b/mysql-test/suite.pm
index f30cc5ec431..4800c1e7dc6 100644
--- a/mysql-test/suite.pm
+++ b/mysql-test/suite.pm
@@ -18,11 +18,13 @@ sub skip_combinations {
 
   # don't run tests for the wrong platform
   if (IS_WINDOWS) {
-    $skip{'include/platform.combinations'} = [ 'aix', 'unix' ];
+    $skip{'include/platform.combinations'} = [ 'aix', 'unix', 'mac' ];
   } elsif (IS_AIX) {
-    $skip{'include/platform.combinations'} = [ 'win', 'unix' ];
-  } else {
-    $skip{'include/platform.combinations'} = [ 'aix', 'win' ];
+    $skip{'include/platform.combinations'} = [ 'win', 'unix', 'mac' ];
+  } elsif (IS_MAC) {
+    $skip{'include/platform.combinations'} = [ 'aix', 'win', 'unix' ];
+  } else { # unix
+    $skip{'include/platform.combinations'} = [ 'aix', 'win', 'mac' ];
   }
 
   if ( $::opt_ps_protocol ) {

```

There are a few platform rdiff tests that would need recording on this:
```
$ find mysql-test/ -name \*rdiff | grep -E 'aix|win|unix'
mysql-test/main/mysqld--help,win.rdiff
mysql-test/main/mysqldump-system,win.rdiff
mysql-test/main/mysqld--help,aix.rdiff
mysql-test/suite/sys_vars/r/sysvars_server_notembedded,aix.rdiff
```

Otherwise including in a lower branch like 10.4 would also some use while tests can fixed/improved.

vuvova:
would be great to see a test that needs `include/not_mac.inc`, with an explanation of what's so special in mac that the test doesn't work there. Not file name case sensitivity, not mac's CrashReporter, but really, just "mac"

DaveGosselin-MariaDB:
> would be great to see a test that needs `include/not_mac.inc`, with an explanation of what's so special in mac that the test doesn't work there. Not file name case sensitivity, not mac's CrashReporter, but really, just "mac"

@vuvova I devised this mechanism as a way for us to get the mac buildbot working without fixing every test first.  At the same time, this would prevent new test failures from creeping in because we could make the mac builder "required" when pushing commits, preventing deliveries from introducing new mac test failures.  What you're asking for seems like a diagnosis of each mac test failure that we know about.  At this point, that is time consuming which meanwhile allows for more failures could appear.  Is such a pragmatic approach not desired, and if so why not?  In short, it would be nice to plug the holes in the hull before bail the water out.

DaveGosselin-MariaDB:
Hi @vuvova.  A [comment](https://jira.mariadb.org/browse/MDEV-33652?focusedCommentId=282661&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-282661) from @abarkov on MDEV-33652 highlights the need for this exclusion mechanism to be included sooner rather than later.  In short, he'd like to wait for another fix to merge to 11.5 so that he can fix this issue in another way.  I have no problem with that, but in the meantime this test will continue to fail on macOS.  If the exclusion mechanism was in place, then we could leave a comment for Alex on his ticket to re-enable this failing test as part of his delivery.

abarkov:
Hi @DaveGosselin-MariaDB , in the meanwhile the test could be fixed not to use upper case letters in the object names.

vuvova:
@DaveGosselin-MariaDB my point was that this particular test, and, I suspect many mac-specific failures could be covered by `include/have_case_insensitive_fs.inc`. That is, the question was what tests fail not just *on case insensitive file system* but on mac because of its inherent *macness* ?

mysql-test/lib/My/Platform.pm
```diff
@@ -70,6 +70,14 @@ BEGIN {
   }
 }
 
+BEGIN {
```
DaveGosselin-MariaDB, line 73: @vuvova I will need at least this change to properly support `--parallel` on macOS because we need to fetch the CPU information differently on that platform than we do on Linux or Windows.  Is there any reason to keep this patch back at this point?

vuvova: Cannot think of any. You've created a separate MDEV for this, I recall and @grooverdan wrote "ok to push into 10.5". I concur, please, feel free to push it, and preferably in 10.5.

PR #3115, opened 2024-03-08
 MDEV-34604 mytop - fix specifying filters in .mytop

PinkFreud:

## Description
The mytop utility supports defining filters in .mytop, but this functionality is currently broken.

Specifying filters (filter_status, filter_user, etc) in the mytop config won't work, because any filter specified here is added to the config hash as a literal string.  This change fixes that - if filter_* is defined in the config and matches an existing filter_* key, then run the value through StringOrRegex() and assign the result to the config hash.

## Release Notes
Update mytop utility to fix parsing of filter_* entries in .mytop config.

## How can this PR be tested?

I don't know that the test suite will actually test this script.

That being said, a manual test will work as follows:
- Add a filter_ entry to the .mytop config, eg: filter_state=/data/
- Run mytop, and see if any queries with state 'Sending data' appear.
- Alternatively, hit capital D in mytop to dump %config hash to screen and look for the filter you added.
   - If it appears as 'filter_state' => qr/data/, the change worked.
   - If it appears as 'filter_state' => '/data/', the config entry is being interpreted as a string, and will never match as a regex.

- [x] *This is a bug fix for a util which was originally from a third party and predates MariaDB, and as such all MariaDB branches contain this bug.*

## PR quality check

### DIFF
--- a/scripts/mytop.sh
+++ b/scripts/mytop.sh
@@ -147,7 +147,12 @@ if (-e $config)
 
             if (/(\S+)\s*=\s*(.*\S)/)
             {
-                $config{lc $1} = $2 if exists $config{lc $1};
+                my ($k, $v) = ($1, $2);
+                if ($k =~ /^filter_/i) {
+                  $config{lc $k} = StringOrRegex($v) if exists $config{lc $k};
+                } else {
+                  $config{lc $k} = $v if exists $config{lc $k};
+                }
             }
         }
         close CFG;


scripts/mytop.sh
```diff
@@ -147,7 +147,12 @@ if (-e $config)
 
             if (/(\S+)\s*=\s*(.*\S)/)
             {
-                $config{lc $1} = $2 if exists $config{lc $1};
+                my ($k, $v) = ($1, $2);
+                if ($k =~ /^filter_/i) {
+                  $config{lc $k} = StringOrRegex($v) if exists $config{lc $k};
```
dlenski, line 152: Perhaps worth adding a comment that references where these configuration keys are handled in the mainloop, and the fact that they need to store regex objects rather than string literals in order to work correctly?

https://github.com/MariaDB/server/blob/43f1f9343efad7d29c32aa9c5649b8b0a2c221d7/scripts/mytop.sh#L529-L573

dlenski (approved):

LinuxJedi (approved):
This looks great to me, many thanks. A couple of things will need to happen before we can merge this:

1. It is a bug fix, so should be rebased against 10.5 instead of 11.5 (10.4 is only taking critical / security fixes for the final version).
2. It needs a Jira ticket created/attached to it, so it can be tracked for documentation purposes.

I'm happy to do both of these things for you if you wish.

PR #3116, opened 2024-03-08
MDEV-33631 Ubuntu/Debian MYSQL_SERVER_SUFFIX is version+suffix on Mar…

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-33631*

## Description

…iaDB packaged versions

The debian/rules removed sufficient version information from the downstream version, however the mariadb upstream has a + separator from the suffix.

The result is compilation with:
    -DMYSQL_SERVER_SUFFIX="-1:10.4.34+maria~deb10" \

Compared to Debian:
    -DMYSQL_SERVER_SUFFIX="-1~deb12u1" \

The impact comes into play with SELECT VERSION():
    10.4.34-MariaDB-1:10.4.34+maria~deb10

A more abbreviated form of the following is much more human/machine parsable:
    10.6.16-MariaDB-1~ubu2004

## Release Notes

Removed duplicate version information in VERSION() for MariaDB debian/ubuntu packages.

## How can this PR be tested?

## PR quality check

Notablely I have assumed a Debian -1 suffix beginning.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -21,7 +21,7 @@ ifneq (,$(filter linux,$(DEB_HOST_ARCH_OS)))
 endif
 
 BUILDDIR := builddir
-DEB_VERSION_REVISION := $(shell echo $(DEB_VERSION) | sed -e 's/^.*-//')
+DEB_VERSION_REVISION := $(shell echo $(DEB_VERSION) | sed -e 's/.*[~-]\(.*\)/\1/')
 RELEASE := $(shell lsb_release -r -s) # Use changelog based DEB_DISTRIBUTION instead?
 TMP:=$(CURDIR)/debian/tmp
 


ottok:
> -DEB_VERSION_REVISION := $(shell echo $(DEB_VERSION) | sed -e 's/^.*-//')
> +DEB_VERSION_REVISION := $(shell echo $(DEB_VERSION) | sed -e
's/^.*~/-1~/')
> RELEASE := $(shell lsb_release -r -s) #

Can you please elaborate a bit, why is adding `-1~` to the result of
`SELECT VERSION()` desired? What is the benefit?

grooverdan:
sed in PR not quite right yet.

Goal is to take a native DEB_VERSION and an autobake.sh version and generate a suffix that doesn't include the full version.

The current behaviour is:

```
$ echo -e '1:10.11.4-1~deb12u1\n1:10.6.18+maria~ubu2204' | sed -e 's/^.*-//'
1~deb12u1
1:10.6.18+maria~ubu2204
```

Native (preserve existing behaviour):
(DEB_VERSION) 1:10.11.4-1~deb12u1

Desired suffix: 1~deb12u1

autobake:
(DEB_VERISON) 1:10.6.18+maria~ubu2204

Desired Suffix: 1~ubu2204

(or does it need a "1" at all).

Test of the best regex I have so far:
```
$ echo -e '1:10.11.4-1~deb12u1\n1:10.6.18+maria~ubu2204' | sed -e 's/\([0-9]*\):.*\(~.*\)/\1\2/'
1~deb12u1
1~ubu2204
```

This is taking the leading 1 from the DEB_VERSION rather than the `1~deb12u1` prefix, but I'm having trouble engineering the right regex to capture both.

ottok:
I can help with the regex but lets first specify the desired end goal.

The Debian string seems ok to me already. The additional '-1' is the Debian
revision that intentionally denotes it from upstream.

The upstream 1:10.6.18+maria~ubu2204 used in all mariadb.org releases and
repositories also seem ok to me. It has the upstream version and a '+maria'
to denote that it is a mariadb.org binary release, and it has ~ubu2204 to
show what distro the build was made for.

To my knowledge this pattern makes all upgrades behave correctly. If
somebody is running 1:10.6.18-1 or 1:10.6.18-0ubuntu.22.04 they will
correctly upgrade if they enable mariadb.org repos to 1:10.6.18+maria~ubu2204.
And later if they upgrade distro they will upgrade to 1:10.6.18+maria~ubu2310
as an example.

What is the desired behavior you are aiming for?

ottok:
Agree, the desired end result for the version query would be:
```
$ podman exec -ti m1011 mariadb -pbob -e 'select version()'
+-----------------------------------------+
| version()                               |
+-----------------------------------------+
| 10.11.6-MariaDB~ubu2204            |
+-----------------------------------------+
```

You don't need the Debian/Ubuntu revision `-1` in this at all.

Just for reference about the Debian package version:
- Debian unstable (inherited to both Debian and Ubuntu if package unchanged): `1:10.11.6-2`
- Ubuntu stable updates example: `1:10.11.6-0ubuntu0.23.10.2`
- Debian stable updates example: `1:10.11.4-1~deb12u1`

grooverdan:
So on regex:
```
$ echo -e '1:10.11.4-1~deb12u1\n1:10.6.18+maria~ubu2204\n1:10.11.6-2\n1:10.11.6-2\n1:10.11.4-1~deb12u1' | sed -e 's/.*[~-]\(.*\)/~\1/'
~deb12u1
~ubu2204
~2
~2
~deb12u1
```

ottok:
I tested this in downstream Debian rules file:
```
diff --git a/debian/rules b/debian/rules
index 45c74f740a5..5a8597113fa 100755
--- a/debian/rules
+++ b/debian/rules
@@ -26,7 +26,11 @@ ifneq (,$(filter linux,$(DEB_HOST_ARCH_OS)))
 endif
 
 BUILDDIR := builddir
+dumpvar = $(info variable `$1` value is >>>$($1)<<<)
+$(call dumpvar,DEB_VERSION)
```

The build log contain:
```
variable `DEB_VERSION` value is >>>1:10.11.7-4<<<
...
    -DMYSQL_SERVER_SUFFIX="-4" \
```

Also tested manually:
```
± echo $DEB_VERSION | sed -e 's/.*[~-]\(.*\)/\1/'
4
± echo $DEB_VERSION | sed -e 's/^.*:\(.*\)\(-\|+\).*/\1/'
10.11.7
± echo 10.11.7 | sed -e 's/^\(.*\)\..*$$/\1/'
10.11.7
```

Not sure why build logs has dash in `-4` while the sed output does not have..

Anyway, your suggested new regex would work in Debian downstream equally well:
```
echo $DEB_VERSION | sed -e 's/.*[~-]\(.*\)/\1/'
4
``

PR #3118, opened 2024-03-08
MDEV-31154 Fatal InnoDB error or assertion `!is_v' failure upon multi…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-31154*

## Description
This is issue is about row ID filtering used with index on virtual
column(s). We hit debug assert and crash while building the record
template in Innodb. The primary reason is that we try to force the code
path to use the ICP path. With ICP, we don't support index with virtual
column and we validate it while index condition is pushed.

Simplify the code for building template to handle both ICP and Row ID
filtering by skipping virtual columns.

## Release Notes
None

## How can this PR be tested?

./mtr main.rowid_filter_innodb

## PR quality check

### DIFF
--- a/mysql-test/main/rowid_filter_innodb.test
+++ b/mysql-test/main/rowid_filter_innodb.test
@@ -1,6 +1,8 @@
 --source include/no_valgrind_without_big.inc
 --source include/have_innodb.inc
 --source include/have_debug.inc
+--source include/have_sequence.inc
+--source include/innodb_stable_estimates.inc
 
 SET SESSION STORAGE_ENGINE='InnoDB';
 
@@ -676,5 +678,78 @@ eval $q;
 eval EXPLAIN EXTENDED $q;
 
 DROP TABLE t1,t2;
+
+--echo #
+--echo # MDEV-31154: Fatal InnoDB error or assertion `!is_v' failure upon multi-update with indexed virtual column
+--echo #
+
+--echo # Test with auto generated Primary Key
+--echo #
+
+SET @save_optimizer_switch= @@optimizer_switch;
+SET optimizer_switch='rowid_filter=on';
+
+CREATE TABLE t0(a int);
+INSERT INTO t0 SELECT seq FROM seq_1_to_20;
+ANALYZE TABLE t0 PERSISTENT FOR ALL;
+
+CREATE TABLE t1 (
+    a int,
+    b int as (a * 2) VIRTUAL,
+    f char(200), /* Filler */
+    key (b),
+    key (a)
+) engine=innodb;
+
+INSERT INTO t1 (a, f) SELECT seq, seq FROM seq_1_to_1000;
+ANALYZE TABLE t1 PERSISTENT FOR ALL;
+
+--echo # Test for type 'ref|filter'
+EXPLAIN SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20;
+SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20;
+
+EXPLAIN SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20 FOR UPDATE;
+SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20 FOR UPDATE;
+
+--echo # Test for type 'range|filter'
+EXPLAIN SELECT count(*) FROM t1 WHERE a<100 and b <100;
+SELECT count(*) FROM t1 WHERE a<100 and b <100;
+
+EXPLAIN SELECT count(*) FROM t1 WHERE a<100 and b <100 FOR UPDATE;
+SELECT count(*) FROM t1 WHERE a<100 and b <100 FOR UPDATE;
+
+--echo # Test with Primary Key
+--echo #
+
+DROP TABLE t1;
+CREATE TABLE t1 (
+    p int PRIMARY KEY AUTO_INCREMENT,
+    a int,
+    b int as (a * 2) VIRTUAL,
+    f char(200), /* Filler */
+    key (b),
+    key (a)
+) engine=innodb;
+
+INSERT INTO t1 (a, f) SELECT seq, seq FROM seq_1_to_1000;
+ANALYZE TABLE t1 PERSISTENT FOR ALL;
+
+--echo # Test for type 'ref|filter'
+EXPLAIN SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20;
+SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20;
+
+EXPLAIN SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20 FOR UPDATE;
+SELECT count(*) from t0,t1 WHERE t0.a=t1.b AND t1.a<20 FOR UPDATE;
+
+--echo # Test for type 'range|filter'
+EXPLAIN SELECT count(*) FROM t1 WHERE a<100 and b <100;
+SELECT count(*) FROM t1 WHERE a<100 and b <100;
+
+EXPLAIN SELECT count(*) FROM t1 WHERE a<100 and b <100 FOR UPDATE;
+SELECT count(*) FROM t1 WHERE a<100 and b <100 FOR UPDATE;
+
+SET optimizer_switch=@save_optimizer_switch;
+
+DROP TABLE t0, t1;
 
 --echo # End of 10.4 tests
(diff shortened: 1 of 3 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -7730,26 +7730,58 @@ ha_innobase::build_template(
...
-				num_v = 0;
-				goto no_icp;
+
+			if (is_v) {
+				if (index_contains) {
+					/* We want to ensure that ICP is not
+					used with virtual columns. Ideally we
+					should assert here ut_ad(!pushed_down)
+					but we are taking forward the legacy to
+					possibly handle unexpected cases. */
```
dr-m, line 7780: Why not add the debug assertion anyway, possibly with a comment that we are looking for counterexamples? Are any regression tests failing if we do that?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7919,7 +7949,7 @@ ha_innobase::build_template(
 	} else {
 no_icp:
 		/* No index condition pushdown */
-		m_prebuilt->idx_cond = NULL;
+		ut_ad(m_prebuilt->idx_cond == nullptr);
```
dr-m, line 7952: Elsewhere in this function, we are checking for null pointers with the unary `!` operator. Doing so would also trim a few bytes off the assertion expression string in the debug executables.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7886,7 +7916,7 @@ ha_innobase::build_template(
 			bool index_contains = index->contains_col_or_prefix(
 				is_v ? num_v : i - num_v, is_v);
```
dr-m, line 7917: Can we write just `i - num_v`?

dr-m (commented):
This is just a quick review, without digging deep into this.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7730,26 +7730,58 @@ ha_innobase::build_template(
...
 		for (ulint i = 0; i < n_fields; i++) {
 			const Field* field = table->field[i];
 			const bool is_v = !field->stored_in_db();
-			if (is_v && skip_virtual) {
-				num_v++;
-				continue;
-			}
+
 			bool index_contains = index->contains_col_or_prefix(
 				is_v ? num_v : i - num_v, is_v);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Not really. If you see the is_v check and continue comes later. So, at this point we could be dealing with a virtual column.
num_v: is the number of virtual columns encountered till now.

storage/innobase/handler/ha_innodb.cc
mariadb-DebarunBanerjee: I just didn't want to introduce regression at this point even in debug mode. There is some handling logic even below that moves to no_icp. So, it needs bigger effort to analyze it. I think it could be some investigation task from optimizer side in future.

storage/innobase/handler/ha_innodb.cc
mariadb-DebarunBanerjee: ok, sure.

storage/innobase/handler/ha_innodb.cc
mariadb-DebarunBanerjee: Replied to the other comment. We cannot ignore virtual columns at this point.

FooBarrior (approved):
Looks good indeed, thank you @mariadb-DebarunBanerjee!

storage/innobase/handler/ha_innodb.cc
mariadb-DebarunBanerjee: As discussed, I added debug assert and looks like there is no failure in test run. If this assert fails ever we would need joint investigation with optimizer team.

mysql-test/main/rowid_filter_innodb.test
```diff
@@ -677,4 +679,77 @@ eval EXPLAIN EXTENDED $q;
...
+
+CREATE TABLE t1 (
+    a int,
+    b int as (a * 2) VIRTUAL,
+    f char(200), /* Filler */
+    key (b),
+    key (a)
+) engine=innodb;
+
+INSERT INTO t1 (a, f) SELECT seq, seq FROM seq_1_to_1000;
```
dr-m, line 704: Do we really need this many rows to reproduce the bug? The test is quite slow:
```
main.rowid_filter_innodb 'innodb'        [ pass ]  17564
```
Before this change, it was almost 2 seconds faster:
```
main.rowid_filter_innodb 'innodb'        [ pass ]  15993
```
This is a test for debug-instrumented builds. If we need many pages for reproducing an issue, we can use `innodb_limit_optimistic_insert_debug`.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7730,26 +7730,56 @@ ha_innobase::build_template(
...
+	Neither ICP nor row ID comparison needs virtual columns and the code is
+	simplified to handle both. It should handle the issues. */
+
+	const bool pushed_down = active_index != MAX_KEY
+				 && active_index == pushed_idx_cond_keyno;
+	const bool rowid_filter = pushed_rowid_filter && rowid_filter_is_active;
+
+	m_prebuilt->idx_cond = pushed_down ? this : nullptr;
+
+	if (pushed_down || rowid_filter) {
```
dr-m, line 7764: The compiler might be smart enough to defer the evaluation of `rowid_filter`, but we might also avoid introducing that variable:
```c++
	m_prebuilt->idx_cond = pushed_down ? this : nullptr;

	if (pushed_down || (pushed_rowid_filter && rowid_filter_is_active)) {
```
Even better, I see that `pushed_rowid_filter && rowid_filter_is_active` was already evaluated earlier in this function. `rowid_filter` is actually duplicating an existing predicate:
```c++
	if (pushed_down || m_prebuilt->pk_filter) {
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7730,26 +7730,56 @@ ha_innobase::build_template(
...
+
+			if (is_v) {
+				if (index_contains) {
+					/* We want to ensure that ICP is not
+					used with virtual columns. */
+					ut_ad(!pushed_down);
+					m_prebuilt->idx_cond = nullptr;
+				}
+				num_v++;
+				continue;
```
dr-m, line 7782: Because this guarantees `!is_v` in the rest of the loop body, we should remove any references to `is_v`, such as `ut_ad(!is_v)` or `ut_ad(!templ->is_virtual)` in the rest of the loop body.

dr-m (approved):
Thank you, this looks OK to me. I only have a couple of minor suggestions.

I would also encourage you to enable the commented-out assertion at the end of the `if (build_template_needs_field_in_icp(…))`  body.

mysql-test/main/rowid_filter_innodb.test
FooBarrior: I tried to reduce it and didn't succeed. All the rowid filter rests use this kind of setup. Also, when I compared the execution time, it was about the same, but different runs have varied in several seconds

mariadb-DebarunBanerjee: @dr-m It is not about Innodb and innodb_limit_optimistic_insert_debug won't help. This is required for generating the right optimizer plan.

storage/innobase/handler/ha_innodb.cc
mariadb-DebarunBanerjee: ok, good point. ut_ad(!is_v) looks too obvious. I have removed it.
Still kept ut_ad(!templ->is_virtual) as it is a derived value.

storage/innobase/handler/ha_innodb.cc
mariadb-DebarunBanerjee: In general I would prefer code clarity and readability over such local optimizations unless in true performance critical path and we know the reason precisely. 

m_prebuilt->pk_filter is a good observation and I think we could replace the variable with it. To keep it consistent I have put the condition as 
if (m_prebuilt->idx_cond  || m_prebuilt->pk_filter)

mysql-test/main/rowid_filter_innodb.test
mariadb-DebarunBanerjee: I am keeping test as it is for now. We may improve in future if we could enforce such plan in optimizer with debug switch or something similar.

mariadb-DebarunBanerjee:
> Thank you, this looks OK to me. I only have a couple of minor suggestions.
> 
> I would also encourage you to enable the commented-out assertion at the end of the `if (build_template_needs_field_in_icp(…))` body.

Yes, something we wish to do in future. Cannot be done immediately because of the issues mentioned in the comment. It would need more analysis and possible refactoring.

PR #3121, opened 2024-03-13
MDEV-32974 : Member fails to join due to old seqno in GTID

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-32974*

## Description
Before MDEV-15158, wsrep xid information was stored in only one place: in the TRX_SYS page. Starting with 10.3, it is not stored there but in the rollback segment header pages, and the latest one is what matters. MDEV-19229 allows the undo tablespaces to be rebuilt when innodb_undo_tablespaces is changed on startup. Previously it was not possible to change that parameter.

These changes caused the fact that rollback segment header pages could contain several wsrep xid's storead and when undo tablespaces were rebuilt there was a effort to restore wsrep xid back to rollback segment header page but because there was several of them the latest wsrep xid was overwritten with more older one.

trx_rseg_read_wsrep_checkpoint
trx_rseg_init_wsrep_xid
	Return true if read xid is wsrep xid, false if not

trx_rseg_mem_restore
	Try to read wsrep xid and if it is found copy it to
	rx_sys.recovered_wsrep_xid if read xid has larger seqno.

## Release Notes

## How can this PR be tested?

Manual testing done using upgrade using Galera mariabackup SST from 10.11 to 11.3.

## PR quality check

### DIFF
--- a/storage/innobase/trx/trx0rseg.cc
+++ b/storage/innobase/trx/trx0rseg.cc
@@ -198,7 +198,7 @@ bool trx_rseg_read_wsrep_checkpoint(const trx_rsegf_t* rseg_header, XID& xid)
 
 	memcpy(xid.data, TRX_RSEG_WSREP_XID_DATA + rseg_header, XIDDATASIZE);
 
-	return true;
+	return (wsrep_is_wsrep_xid(&xid));
 }
 
 /** Read the WSREP XID from the TRX_SYS page (in case of upgrade).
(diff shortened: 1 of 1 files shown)


janlindstrom:
Closing this PR because fix is not necessary before 10.11. If donor and joiner innodb-undo-tablespaces is not exactly the same, joiner startup fails with error message. This is because undo log tablespaces (if exists) are copied on SST.

PR #3125, opened 2024-03-13
MDEV-29955: Set path for zlib library with pkg-config

keeper:
- [x] *The Jira issue number for this PR is: MDEV-29955*

## Description
`FindZLIB` module uses variable `ZLIB_ROOT`[1] to look for libraries. By
setting the variable, `FindZLIB` is able to search the libraries that
installed in a non-system path (/workspace/mylib for example).

And when using `z` in `LINK_LIBRARIES()` CMake tries to lookup the
library in system path by default. It doesn't work if the library isn't
installed in the path, and use ${ZLIB_LIBRARY} which set by FindZLIB
solve the issue.

[1]: https://cmake.org/cmake/help/latest/module/FindZLIB.html#hints

## How can this PR be tested?
* [x] Builds with no errors with Zlib installed in custom path

Manual checks:
1. pkg-config:
```
$ pkg-config --variable=libdir zlib
/workspace/mylib/lib
```

2. Generated CMake build file
```
$ cmake -S . -B build -G Ninja --fresh -DWIHT_ZLIB=system
...
...
$ grep libz.so build/build.ninja
...
...
LINK_LIBRARIES = -Wl,-rpath,/workspace/mylib/lib: ... /workspace/mylib/lib/libz.so  -lm  -ldl
```

## PR quality check

## Copyright
Services.


cmake/zlib.cmake
```diff
@@ -37,6 +37,8 @@ MACRO (MYSQL_CHECK_ZLIB_WITH_COMPRESS)
   IF(WITH_ZLIB STREQUAL "bundled")
     MYSQL_USE_BUNDLED_ZLIB()
   ELSE()
+    FIND_PACKAGE(PkgConfig QUIET)
```
grooverdan, line 40: is, to depend on the finding of pkg-config is the following required:
```
IF(PKG_CONFIG_FOUND)
  PKG_GET_VARIABLE(ZLIB_ROOT zlib prefix)
ENDIF()
```

I'm roughly thinking it is based on https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPkgConfig.cmake#L131 .

Have you tried with pkg-config not there?

cmake/zlib.cmake
```diff
@@ -37,6 +37,8 @@ MACRO (MYSQL_CHECK_ZLIB_WITH_COMPRESS)
...
+    PKG_GET_VARIABLE(ZLIB_ROOT zlib prefix)
     INCLUDE(FindZLIB)
```
grooverdan, line 42: odd, but not your fault, not sure why this isn't `find_package(zlib)`

storage/maria/CMakeLists.txt
```diff
@@ -124,7 +124,7 @@ ENDIF()
 IF (CURL_FOUND)
   INCLUDE_DIRECTORIES(${CURL_INCLUDE_DIRS})
   MYSQL_ADD_PLUGIN(s3 ha_s3.cc ${S3_SOURCES} COMPONENT s3-engine
-    LINK_LIBRARIES ${CURL_LIBRARIES} z STORAGE_ENGINE NOT_EMBEDDED CONFIG s3.cnf)
```
grooverdan, line 127: good. Thank you.

cmake/zlib.cmake
```diff
@@ -37,6 +37,8 @@ MACRO (MYSQL_CHECK_ZLIB_WITH_COMPRESS)
...
     IF(ZLIB_FOUND)
      INCLUDE(CheckFunctionExists)
```
grooverdan, line 44: Is  `SET(CMAKE_REQUIRED_LIBRARIES ${ZLIB_LIBRARIES})`  required below?

cmake/zlib.cmake
keeper: > Have you tried with pkg-config not there?
Interesting, I moved `pkg-config` from `/usr/bin` to somewhere not in $PATH but the command can still find the right path:

```
-- Found ZLIB: /workspace/mylib/lib/libz.so (found version "1.2.13") 
```

I basically follow the convention here:
https://github.com/MariaDB/server/commit/3a33ae86010051a1022c392f2936b5fc9f3ab19e#diff-8e35291051929131abcdfd960c0e4e246317876c5ec2989154a8e810d489c7d6R81-R82, do you prefer me checking the availability here?

cmake/zlib.cmake
keeper: Thank you for pointing out, I can also change the line to use `find_package()` which is more conscience

cmake/zlib.cmake
keeper: Yes - thanks again for pointing this out

cmake/zlib.cmake
grooverdan: I think its fine not that you've tested it. Thanks for the pcre reference implementation.

grooverdan (approved):
`find_package(zlib)`

> Thank you for pointing out, I can also change the line to use find_package() which is more conscience

Optional, if you have time, sure happy to take the improvement (and hope it passes all of CI).

vaintroub (changes requested):

cmake/zlib.cmake
vaintroub: What happens if zlib is not installed. What is the value of ZLIB_ROOT?

storage/maria/CMakeLists.txt
```diff
@@ -124,7 +124,7 @@ ENDIF()
...
+    LINK_LIBRARIES ${CURL_LIBRARIES} ${ZLIB_LIBRARY} STORAGE_ENGINE NOT_EMBEDDED CONFIG s3.cnf)
```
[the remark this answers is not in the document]
keeper: > I would prefer checking availability and I would also prefer to exclude this on Windows
This has been fixed in latest update, thank you for the great advice.

> ZLIB_LIBRARY is not even documented, though appears to work
Totally agree. I also tends to use `ZLIB_LIBRARIES` but it's `ZLIB_LIBRARY` everywhere. To keep the consistency across codes I'll leave it as is per your suggestion.

vaintroub (approved):
Looks good to me, thanks!

PR #3128, opened 2024-03-14
MDEV-33672: Gtid_log_event Construction from File Should Ensure Event Length When Using Extra Flags

bnestere:
A GTID event can have variable length, with a contributing factor to
the variable length being the optional extra flags field. This field
is a bitmap, where each set bit indicates an additional value that
should be appended to the event, e.g. multi-engine transactions
append a number to indicate the number of additional engines a
transaction uses. However, if the extra flags bit is set, and no
additional fields are appended to the event, the server can still
try to read from memory as if it did exist (note that in debug
builds, there is a DBUG_ASSERT condition that will cause the server
to crash).

This patch fixes this to check that the length of the event is
aligned with the expectation set by the extra flags. The is_valid()
check of a Gtid event is extended to ensure that the extra_flags is
consistent with valid values, so after unmarshaling the event, it
can be verified for correctness. If found invalid, the server will
report an error about an invalid event in the binary log.

1. The first commit is the regression to show the issue
2. The second commit contains the fix

Note that there are additional flags related to two phase alter added
in 10.8, which will be fixed in another patch in 10.11.

### DIFF
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -2629,6 +2629,11 @@ Gtid_log_event::Gtid_log_event(const uchar *buf, uint event_len,
   }
   if (flags2 & (FL_PREPARED_XA | FL_COMPLETED_XA))
   {
+    if (event_len < static_cast<uint>(buf - buf_0) + 6)
+    {
+      seq_no= 0;
+      return;
+    }
     xid.formatID= uint4korr(buf);
     buf+= 4;
 
@@ -2637,6 +2642,11 @@ Gtid_log_event::Gtid_log_event(const uchar *buf, uint event_len,
     buf+= 2;
 
     long data_length= xid.bqual_length + xid.gtrid_length;
+    if (event_len < static_cast<uint>(buf - buf_0) + data_length)
+    {
+      seq_no= 0;
+      return;
+    }
     memcpy(xid.data, buf, data_length);
     buf+= data_length;
   }
@@ -2651,8 +2661,11 @@ Gtid_log_event::Gtid_log_event(const uchar *buf, uint event_len,
     */
     if (flags_extra & FL_EXTRA_MULTI_ENGINE)
     {
-      DBUG_ASSERT(static_cast<uint>(buf - buf_0) < event_len);
-
+      if (event_len < static_cast<uint>(buf - buf_0) + 1)
+      {
+        seq_no= 0;
+        return;
+      }
       extra_engines= *buf++;
 
       DBUG_ASSERT(extra_engines > 0);
--- a/sql/log_event.h
+++ b/sql/log_event.h
@@ -3666,7 +3666,16 @@ public:
   {
     return GTID_HEADER_LEN + ((flags2 & FL_GROUP_COMMIT_ID) ? 2 : 0);
   }
-  bool is_valid() const { return seq_no != 0; }
+
+  bool is_valid() const
+  {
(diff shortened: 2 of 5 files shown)


mysql-test/suite/rpl/t/rpl_extra_flags_event_len_inconsistent.test
```diff
@@ -0,0 +1,74 @@
+#
+#   This test enusres that, when a GTID event is constructed by reading its
```
knielsen, line 2: Typo s/enusres/ensures/

mysql-test/suite/rpl/t/rpl_extra_flags_event_len_inconsistent.test
```diff
@@ -0,0 +1,74 @@
...
+# content from a binlog file, the reader (e.g. replica, in this test) cannot
+# read beyond the length of the GTID event, if the extra_flags is not
+# consistent with the format of the event itself.
+#
+#   To test this, we create a transaction with its length > GTID_HEADER_LEN
+# (otherwise, it is zero padded and terminated early). To extend its length,
+# we give the transaction a commit id.
```
knielsen, line 9: "we create a transaction" - that was hard to understand, maybe "we create a GTID event".

sql/log_event.cc
```diff
@@ -2649,10 +2649,9 @@ Gtid_log_event::Gtid_log_event(const uchar *buf, uint event_len,
       extra engines flags presence is identifed by non-zero byte value
       at this point
     */
-    if (flags_extra & FL_EXTRA_MULTI_ENGINE)
+    if ((flags_extra & FL_EXTRA_MULTI_ENGINE) &&
+        (static_cast<uint>(buf - buf_0) < event_len))
```
knielsen, line 2653:
```
 if (flags2 & (FL_PREPARED_XA | FL_COMPLETED_XA))
  {
    xid.formatID= uint4korr(buf);
    buf+= 4;
```

You added a check for FL_EXTRA_MULTI_ENGINE, but why no similar check for the FL_PREPARED_XA / FL_COMPLETED_XA ?

sql/log_event.h
```diff
@@ -3666,7 +3666,19 @@ class Gtid_log_event: public Log_event
...
+  {
+    /*
+      If the GTID event indicates that it will use extra engines, there must be
+      extra engines set
+    */
+    if (flags_extra && (flags_extra & FL_EXTRA_MULTI_ENGINE) && !extra_engines)
+      return false;
+
+    return seq_no != 0;
+  }
```
knielsen, line 3680: The other existing check for length (FL_GROUP_COMMIT_ID) sets seq_no=0 to mark that the event is not valid. Why not do the same for the new check you added? For consistency, and to pay the extra price only in the rare error case, not in the common case of checking is_valid() for correct event?

knielsen (approved):
Generally looks ok.
Some detailed comments added for you to consider.

sql/log_event.cc
bnestere: I only looked at `extra_flags` - nice spot. I'll add a check for those flags too.

sql/log_event.h
bnestere: I missed that the earlier code sets `seq_no=0` to invalidate the event. Agreed that it would be better to have the consistency.

PR #3130, opened 2024-03-16
MDEV-33618: add mariadbd_safe to option groups

andremralves:
- [x] *The Jira issue number for this PR is: MDEV-33618*

## Description
- Add `mariadbd_safe` to be used as an option group in my.cnf
- Update man page.

### DIFF
--- a/man/mariadbd-safe.1
+++ b/man/mariadbd-safe.1
@@ -51,14 +51,14 @@ Options unknown to
 are passed to
 \fBmysqld\fR
 if they are specified on the command line, but ignored if they are specified in the
-[mysqld_safe] or [mariadb_safe]
+[mysqld_safe], [mariadbd-safe] or [mariadbd_safe]
 groups of an option file\&.
 .PP
 \fBmysqld_safe\fR
 reads all options from the
 [mysqld],
 [server],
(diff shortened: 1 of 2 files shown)


grooverdan (approved):

tonychen2001:
Also need to change references of `mariadb_safe` to `mariadbd-safe` in `man/mysqld_safe.1`

LinuxJedi:
> > Also need to change references of `mariadb_safe` to `mariadbd-safe` in `man/mysqld_safe.1`
> 
> I wasn't sure if I should change the man pages because `mariadb_safe` still works, but I think it's better. Thanks.

Yes `mariadb_safe` works, but it is deprecated, so I think removing it from the man is safe. If you could add `mariadbd-safe` as well to what you have done, I think this is in a good place.

andremralves:
> Yes `mariadb_safe` works, but it is deprecated, so I think removing it from the man is safe. If you could add `mariadbd-safe` as well to what you have done, I think this is in a good place.

Hi @LinuxJedi, thanks for the clarifications. I also added `mariadbd-safe` to the man page.

LinuxJedi (approved):
Many thanks for fixing that.

PR #3131, opened 2024-03-17
MDEV-33660: Warns the user if they try to set auto_increment <= max_v…

andremralves:
…alue

- [x] *The Jira issue number for this PR is: MDEV-33660*

## Description
This PR adds a warning for the cases where `ALTER TABLE .. AUTO_INCREMENT = X` is set to a value lower than or equal to the current `max_value` in the table.

```
MariaDB [test]> alter table t1 auto_increment=3;
Query OK, 0 rows affected, 1 warning (0.012 sec)
Records: 0  Duplicates: 0  Warnings: 1

MariaDB [test]> show warnings\G
  Level: Warning
   Code: 167
Message: Can not set AUTO_INCREMENT to 3 which is lower than or equal to the current max value in the table: 5
1 row in set (0.000 sec)

```

Points to review:
Is the error code `HA_ERR_AUTOINC_ERANGE` ok or should I use/create another one?

## Release Notes
Warns the user if they try to set auto_increment to a value lower than or equal max_value.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/main/alter_table_autoinc-5574.result
+++ b/mysql-test/main/alter_table_autoinc-5574.result
@@ -2,6 +2,8 @@ create table t1(a int(10)unsigned not null auto_increment primary key,
 b varchar(255) not null) engine=innodb default charset=utf8;
 insert into t1 values(1,'aaa'),(2,'bbb');
 alter table t1 auto_increment=1;
+Warnings:
+Warning	167	Can not set AUTO_INCREMENT to 1 which is lower than or equal to the current max value in the table: 2
 insert into t1 values(NULL, 'ccc');
 select * from t1;
 a	b
--- a/mysql-test/suite/innodb/r/autoinc_persist.result
+++ b/mysql-test/suite/innodb/r/autoinc_persist.result
@@ -562,6 +562,8 @@ Expect 200
 # This will not change the counter to 150, but to 201, which is the next
 # of current max counter in the table
 ALTER TABLE t3 AUTO_INCREMENT = 150;
+Warnings:
+Warning	167	Can not set AUTO_INCREMENT to 150 which is lower than or equal to the current max value in the table: 200
 SHOW CREATE TABLE t3;
 Table	Create Table
 t3	CREATE TABLE `t3` (
@@ -603,6 +605,8 @@ UPDATE t3 SET a = 200 WHERE a = 150;
 INSERT INTO t3 VALUES(220);
 # This still fails to set to 120, but just 221
 ALTER TABLE t3 AUTO_INCREMENT = 120;
+Warnings:
+Warning	167	Can not set AUTO_INCREMENT to 120 which is lower than or equal to the current max value in the table: 220
 SHOW CREATE TABLE t3;
 Table	Create Table
 t3	CREATE TABLE `t3` (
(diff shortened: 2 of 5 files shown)


andremralves:
Is it necessary to add tests for this kind of change? I'm thinking about a test that checks if a warning was generated, but I couldn't find any example.

grooverdan:
Should be possible to add to mysql-test/main/auto_increment_ranges.inc. This is included by the auto_increment_ranges_innodb and auto_increment_ranges_myisam tests.

From the test results here, tests innodb.innodb-autoinc-44030 vcol.innodb_autoinc_vcol innodb.autoinc_persist main.alter_table_autoinc-5574 also need to be re-recorded.

`mysql-test/mtr --record {testname}` and look at the diff in the result to ensure that it is correct. Then commit amend and force push to the same branch here.

You'll note that aria (storage/maria) and myisam and memory engine (storage/heap) code also need to generate the same warning.

andremralves:
Hi @grooverdan, thanks for your feedback it helped me a lot. I've updated the tests results, but I'm still having a hard time to figure out how to generate the warnings for the other engines like myisam. In innodb I was using the function `push_warning_printf` to generate a new warning while in myisam I thought the equivalent function would be `mi_check_print_warning`, but that doesn't seem to be the case.

grooverdan:
push_warning_printf still can be used in MyISAM/Aria. Its not used frequently but still can be.

andremralves:
Hi, sorry for the late response. I got a little bit stuck on this task and started doing other tasks to become more familiar with the codebase and then come back here.

I felt that the way other engines, like Aria, handles autoincrement is a little bit different from Innodb. While in Innodb there was a piece of code that compares the `max_value` in the table with the new `autoincrement` - and I could just add a warning there. In Aria I couldn't find a similar comparison, I think we just call `maria_write` for each row and update the `auto_increment` as we go. If the current row id is greater than `auto_increment` we update it. That makes adding a warning a little difficult since `maria_write` is called multiple times.

Here is the piece of code I'm referring to:
https://github.com/MariaDB/server/blob/f151c5f389e4d5c5bf877d361d406eb1581ad013/storage/maria/ma_write.c#L290-L296


> push_warning_printf still can be used in MyISAM/Aria. Its not used frequently but still can be.

I tried to use it, but I got some errors. Also, I couldn't find any examples of push_warning_printf being used in C files. Is there some other options?

gkodinov (changes requested):
I'm terribly sorry for the delayed response on this PR! Somehow it's not picked up by the Jira UI and probably because of that it received less attention than it should have!

FWIW, there's a more recent PR against the same problem: https://github.com/MariaDB/server/pull/4254. This fixes the problem at a more correct layer (the server) vs InnoDB. 

I guess, at this point, it's prudent to close this PR and contribute to the other one, since it has already received several rounds of review by senior developers. 

If you agree, please close the PR. 

Again: terribly sorry for the lack of reply! We at MariaDB are constantly improving our processes and I hope that your next contribution will be processed much faster.

gkodinov:
No feedback received on my previous comment. I'll assume the submitter is OK with closing this and will go ahead and close it.

PR #3132, opened 2024-03-18
MDEV-33636: RPM caps is on mariadbd exe

grooverdan:

## Description

Postfix on 51e3f1daf54309d14fe8db438024d88aa110e86a that mariadbd should be the executable name rather than capabilities on a symlink.

## Release Notes

not significant

## How can this PR be tested?

Check RPM builds and that capabilities are added on install.

## PR quality check


illuusio:
Ok I checked this couple times but didn't got it correct :monocle_face: but LGTM now.

vuvova:
Note that MDEV-33636 broke openssl config (and tests that rely on that), so it needs to be fixed or reverted.

grooverdan:
Yep, did see MDEV-33636

PR #3134, opened 2024-03-18
MDEV-33652 compat/oracle.sp-inout fails on macOS

DaveGosselin-MariaDB:
On systems where the database is running from a case-insensitive filesystem (like APFS which is case-insensitive by default), we discard the SP name case information.  Later, if we print or otherwise display this name to the user, it’s shown as all lowercase, in contradiction to our documentation.

When specifying the name of a SP, we remember that name both in its original case and in a lowercase form.  We show the original case to the user whenever it may appear in an error message or show output. We retain the all-lowercase value for internal processing as we do today, so internal SP lookup procedures are unaffected.


DaveGosselin-MariaDB:
Closing this PR as we changed the approach.

PR #3139, opened 2024-03-19
MDEV-249034 fix assersion (prefix_size <= width)

ahmedmadbouly186:
the prefix_size might be larger than the  width , in this case the function should return false because it is impossible for the first [prefix_size] bits to be set to 1 , because prefix_size > number of bits available

- [x] *The Jira issue number for this PR is: MDEV-24931

## Description
1-when apply this query : SELECT * FROM v1 NATURAL JOIN t1; where v1 is view from t1 , and t1 is any table with number of columns > 64 , then the server will  crash.
2- i have added test suit/join/join_natural.test , before the patch it fail , after this patch it pass correctly 
3-no i don't think so , i have read the dependencies and i found that there is only one function depend  on this patch  witch is make_join_statistics in file sql/sql_select.cc , and after following the logic it is clearly that this function need to this patch , it is need to return false from is_prefix in case prefix_size > width

## Release Notes
there is nothing need to be changed , the patch is only inside the logic of single function with take about the dependencies 

## How can this PR be tested?
run the following from the build directory :
mysql-test/mtr --suite=join --mem --parallel=5

## PR quality check

### DIFF
--- a/mysql-test/suite/join/join_natural.test
+++ b/mysql-test/suite/join/join_natural.test
@@ -1,0 +1,46 @@
+--echo #
+--echo # Bug #XXXXXX: INET_ATON() returns signed, not unsigned
+--echo #
+DROP TABLE IF EXISTS ttt1;
+DROP VIEW IF EXISTS vvv1;
+
+
+
+CREATE TABLE ttt1 (
+
+  f01 int, f02 int, f03 int, f04 int, f05 int, f06 int, f07 int, f08 int,
+
+  f09 int, f10 int, f11 int, f12 int, f13 int, f14 int, f15 int, f16 int,
+
+  f17 int, f18 int, f19 int, f20 int, f21 int, f22 int, f23 int, f24 int,
+
+  f25 int, f26 int, f27 int, f28 int, f29 int, f30 int, f31 int, f32 int,
+
+  f33 int, f34 int, f35 int, f36 int, f37 int, f38 int, f39 int, f40 int,
+
+  f41 int, f42 int, f43 int, f44 int, f45 int, f46 int, f47 int, f48 int,
+
+  f49 int, f50 int, f51 int, f52 int, f53 int, f54 int, f55 int, f56 int,
+
+  f57 int, f58 int, f59 int, f60 int, f61 int, f62 int, f63 int, f64 int,
+
+  f65 int);
+
+
+
+CREATE ALGORITHM=TEMPTABLE VIEW vvv1 AS SELECT * FROM ttt1;
+
+INSERT INTO ttt1 VALUES (),();
+
+INSERT INTO ttt1 VALUES (1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+1,1,1,1,1);
+
+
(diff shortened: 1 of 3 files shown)


robinnewhouse:
Hi @ahmedmadbouly186, thanks for the submission.
Is there a reason you use `if` instead of `DBUG_ASSERT`? More for my own curiosity.

Also, have you seen https://github.com/MariaDB/server/pull/2520 which addresses the same issue?

ahmedmadbouly186:
Hi @robinnewhouse  yes i have seen https://github.com/MariaDB/server/pull/2520 , after tracing the code and find why there is frailer i found that this solution will also get assertion and crash in case prefix_size >=128 , because the width is set to 64 " max_num_indexies" and also BITS_PER_ELEMENT=64 " for long long " so  we found that if (prefix_size-width)>=64 then it will give the same assertion and same crash 

my solution is based on the logic that if we have bit stream with size n , then it is impossible to make m bit set to 1 , if m>n "there is not enough bits" so it simply return false

mysql-test/suite/join/join_natural.test
```diff
@@ -0,0 +1,46 @@
+--echo #
+--echo # Bug #XXXXXX: INET_ATON() returns signed, not unsigned
+--echo #
+DROP TABLE IF EXISTS ttt1;
+DROP VIEW IF EXISTS vvv1;
```
grooverdan, line 5: DROP TABLE/VIEW not needed.

mysql-test/suite/join/join_natural.test
```diff
@@ -0,0 +1,46 @@
+--echo #
+--echo # Bug #XXXXXX: INET_ATON() returns signed, not unsigned
```
grooverdan, line 2: MDEV-24931 instead of Bug.

Using an existing test file like #2520 is referred.

mysql-test/suite/join/join_natural.test
```diff
@@ -0,0 +1,46 @@
...
+
+  f65 int);
+
+
+
+CREATE ALGORITHM=TEMPTABLE VIEW vvv1 AS SELECT * FROM ttt1;
+
+INSERT INTO ttt1 VALUES (),();
+
+INSERT INTO ttt1 VALUES (1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
```
grooverdan, line 35: values weren't required in the MDEV.

mysql-test/suite/join/join_natural.test
```diff
@@ -0,0 +1,46 @@
...
+1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+1,1,1,1,1);
+
+
+SELECT * FROM vvv1 NATURAL JOIN ttt1;
+
+DROP VIEW vvv1;
+
+DROP TABLE ttt1;
+
```
grooverdan, line 46: Note there was a second test case in the MDEV.

sql/sql_bitmap.h
```diff
@@ -133,8 +133,12 @@ template <uint width> class Bitmap
   }
   bool is_prefix(uint prefix_size) const
   {
-    DBUG_ASSERT(prefix_size <= width);
-
+    // if prefix_size is larger than width, the function would return false
```
grooverdan, line 136: As @vaintroub  said on the MDEV, this is the wrong place to fix it. It will probably have consequences for other uses of the Bitmap.

The solution is in `generate_derived_keys_for_table` or earlier to prevent a prefix_size of 65 being passed here.

grooverdan (changes requested):

PR #3140, opened 2024-03-20
MDEV-21102: Server crashes in JOIN_CACHE::write_record_data upon EXPL…

spetrunia:
…AIN with subqueries

JOIN_CACHE has a light-weight initialization mode that's targeted at EXPLAINs. In that mode, JOIN_CACHE objects are not able to execute.

Light-weight mode was used whenever the statement was an EXPLAIN. However the EXPLAIN can execute subqueries, provided they enumerate less than @@expensive_subquery_limit rows.

Make sure we use light-weight initialization mode only when the select is more expensive @@expensive_subquery_limit.

Also add an assert into JOIN_CACHE::put_record() which prevents its use if it was initialized for EXPLAIN only.

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/sql/sql_select.cc
+++ b/sql/sql_select.cc
@@ -1876,6 +1876,26 @@ JOIN::init_range_rowid_filters()
 
 int JOIN::init_join_caches()
 {
+  bool init_for_explain= false;
+
+  /*
+    Can we use lightweight initalization mode just for EXPLAINs? We can if
+    we're certain that the optimizer will not execute the subquery.
+    The optimzier will not execute the subquery if it's too expensive. For
+    the exact criteria, see Item_subselect::is_expensive(). 
+    Note that the subquery might be a UNION and we might not yet know if it is
+    expensive.
+    What we do know is that if this SELECT is too expensive, then the whole 
+    subquery will be too expensive as well.
+    So, we can use lightweight initialization (init_for_explain=true) if this
+    SELECT examines more than @@expensive_subquery_limit rows.
+  */
+  if ((select_options & SELECT_DESCRIBE) &&
+      get_examined_rows() >= thd->variables.expensive_subquery_limit)
+  {
+    init_for_explain= true;
+  }
+
   JOIN_TAB *tab;
 
   for (tab= first_linear_tab(this, WITH_BUSH_ROOTS, WITHOUT_CONST_TABLES);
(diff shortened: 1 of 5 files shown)


sql/sql_select.cc
```diff
@@ -1895,7 +1897,27 @@ int JOIN::init_join_caches()
...
+    /*
+      We are about to initialize a join buffer.
+      Can we use lightweight initalization mode just for EXPLAINs? We can if
+      we're certain that the optimizer will not execute the subquery.
+
+      It won't execute it if this join is too expensive.  We follow the logic
+      in Item_subselect::is_expensive() and initalize for explain only if this
+      join examines more than @@expensive_subquery_limit.
+    */
+    if (tab->cache && !checked_if_expensive)
```
Olernov, line 1910: `checked_if_expensive` is set once during the loop and is never reset, so it affects other tables. Is this the idea?

sql/sql_select.cc
```diff
@@ -1895,7 +1897,27 @@ int JOIN::init_join_caches()
     {
       table->prepare_for_keyread(tab->index, table->read_set);
     }
-    if (tab->cache && tab->cache->init(select_options & SELECT_DESCRIBE))
+
+    /*
+      We are about to initialize a join buffer.
+      Can we use lightweight initalization mode just for EXPLAINs? We can if
+      we're certain that the optimizer will not execute the subquery.
```
Olernov, line 1904: How the term `subquery` gets into this context? We are looping through tables of the JOIN, how are subqueries related to them?

sql/sql_select.cc
```diff
@@ -1895,7 +1897,27 @@ int JOIN::init_join_caches()
...
+
+      It won't execute it if this join is too expensive.  We follow the logic
+      in Item_subselect::is_expensive() and initalize for explain only if this
+      join examines more than @@expensive_subquery_limit.
+    */
+    if (tab->cache && !checked_if_expensive)
+    {
+      checked_if_expensive= true;
+      if ((select_options & SELECT_DESCRIBE) &&
+          get_examined_rows() >= thd->variables.expensive_subquery_limit)
```
Olernov, line 1914: Same question as above: how do we come from tables to subqueries here?

sql/sql_select.cc
```diff
@@ -1895,7 +1897,27 @@ int JOIN::init_join_caches()
...
+      {
+        init_for_explain= true;
```
Olernov, line 1916: Same question as above: `init_for_explain` is set here once in the loop and is never reset, so affects other tables and their caches. Is it correct?

Olernov, line 1916: Please move this initialization outside the loop

Olernov:
Ok to push

PR #3141, opened 2024-03-20
MDEV-33044 Loading time zones does not work with alter_algorithm INPLACE

anson1014:
- [x] *The Jira issue number for this PR is: [MDEV-33044](https://jira.mariadb.org/browse/MDEV-33044)*

## Description
`$MYSQL_TZINFO_TO_SQL` works by truncating tables. Truncation is an operation that cannot be done in-place and therefore is fundamentally incompatible with `alter_algorithm='INPLACE'`. As a result, we override the default alter_algorithm setting in `tztime.cc` to `alter_algorithm='COPY'` so that timezones can be loaded regardless of the previously set alter_algorithm.

## Release Notes
This is a minor bug fix and therefore does not require release notes.

## How can this PR be tested?

- added new test to load timezones when `alter_algorithm=INPLACE` has been explicitly set.
- updated result file of existing timezone test.
- builds on numerous platforms and main mtr suite are successfully passing.

## PR quality check

## Copyright
Services.

### DIFF
--- a/mysql-test/main/load_timezones_with_alter_algorithm_inplace.test
+++ b/mysql-test/main/load_timezones_with_alter_algorithm_inplace.test
@@ -1,0 +1,38 @@
+# MDEV-33044 Loading time zones does not work with alter_algorithm INPLACE
+
+set global alter_algorithm=INPLACE;
+
+# Because loading timezones alters the mysql tables,
+# this test will leave mysql in a different state than when it started.
+# Furthermore, checksums on the various mysql.timezone_x tables will fail. 
+
+# Therefore we:
+# 1. Make "backups" of the existing tables by renaming them
+# 2. Make dummy clones of the tables we just backed up
+# 3. Load timezones with alterations made to the dummy clone tables
+# 4. Drop the newly made tables with changes made to them
+# 5. Restore the backed up tables so the checksums will pass
+
+RENAME TABLE mysql.time_zone TO mysql.time_zone_BACKUP;
+RENAME TABLE mysql.time_zone_name TO mysql.time_zone_name_BACKUP;
+RENAME TABLE mysql.time_zone_transition TO mysql.time_zone_transition_BACKUP;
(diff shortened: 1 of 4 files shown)


vuvova:
better save the old value and restore it, like it's done for `SQL_LOG_BIN` in the same file. Or use `SET STATEMENT` that restores the old value automatically.

This is for a (rather hypothetical) case when someone sources the generated SQL into the existing session, e.g. by using the `\s` command of the mariadb command line client.

anson1014:
Hi @vuvova, thank you for the feedback. The changes have been made, kindly please review at your leisure.

mysql-test/main/load_timezones_with_alter_algorithm_inplace.test
```diff
@@ -0,0 +1,36 @@
+set global alter_algorithm=INPLACE;
+
+# Because loading timezones alters the mysql tables,
+# this test will leave mysql in a different state than when it started.
+# Furthermore, checksums on the various mysql.timezone_x tables will fail.
```
grooverdan, line 5: Good docs.

Although there is a couple of instances of `set global x=DEFAULT` as a restore mechanism (most are `session` based restores), its generally better to use user variables (a different one from the maradb-tz-info-to-sql).

Good practice to list a MDEV.

sql/tztime.cc
```diff
@@ -2833,6 +2835,9 @@ main(int argc, char **argv)
       "END IF|\n"
       "\\d ;\n");
 
+  if ((argc == 1 && !opt_leap)){
```
grooverdan, line 2838: nit - extra parentheses unneeded and remove `{` / `}`

grooverdan (approved):
Even with minor defects I'm happy with it.

anson1014:
Hi @grooverdan, nits have been addressed and the requested changes have been made. I've also rebased onto the latest 10.5. Please merge if no other changes are needed.

grooverdan (approved):

PR #3142, opened 2024-03-20
MDEV-33574 Improve mysqlbinlog error message

HugoWenTD:
- [x] *The Jira issue number for this PR is: [MDEV-33574](https://jira.mariadb.org/browse/MDEV-33574)*

## Description
Previously, when running mysqlbinlog without providing a binlog file, it would print the entire help text, which was very verbose and made it difficult to identify the actual issue.

Now change the behavior to print a more concise error message instead:

    "ERROR: Please provide the log file(s). Run with '--help' for usage instructions."

This makes the error output more user-friendly and easier to understand, especially when running the tool in scripts or automated processes.

## Release Notes
What should the release notes say about this change?
- I think this minor change does not need to be included in the release notes. Or simply say "Concise error message when no binlog file is provided. (MDEV-33574)"

## How can this PR be tested?
The change includes error message change only, no MTR needed.

Instead of displaying the previous 200+ line help message, the following example shows concise message will be output when executing mysqlbinlog/mariadb-binlog without specifying a log file:

```
$ ./mariadb-binlog -R  --socket=/var/lib/mysql/mysql.sock --stop-never
ERROR: Please provide the log file(s). Run with '--help' for usage instructions.
```

- [x] *This is an improvement of the error message but the fix versions are marked as 10.4+ in Jira [MDEV-33574](https://jira.mariadb.org/browse/MDEV-33574).*
Please let me know if we should fix it only for 10.5+ and I'll rebase.

## PR quality check

### DIFF
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -2997,7 +2997,8 @@ int main(int argc, char** argv)
   {
     if (!opt_version)
     {
-      usage();
+      error("Please provide the log file(s). Run with '--help' for usage "
+            "instructions.");
       retval= ERROR_STOP;
     }
     goto err;


grooverdan (approved):
Looks good. Nice and simple. I'm happy with 10.4.

LinuxJedi (approved):
Seems simple enough.

@HugoWenTD I know you are probably contributing under BSD, but can you please confirm it, I can't see it here.

Also, this will need to be against 10.5. I can do the rebase if you like.

HugoWenTD:
Hi @LinuxJedi, Yes, I'm contributing under BSD, the copyright information is included in the [commit message](https://github.com/MariaDB/server/pull/3142/commits/75c9f9f858153613fc89bfd42a1ef8c5d1255178). 
I've rebased against 10.5 now. Thank you!

PR #3143, opened 2024-03-21
MDEV-11339 Support UUID v4 generation

StefanoPetrilli:
- [x] *The Jira issue number for this PR is: MDEV-11339

## Description
UUIDv4 generates random UUIDs, which means they are not based on timestamps or other predictable data. If you require UUIDs that are not predictable or reveal any information about the time they were generated, UUIDv4 is a better choice.

## Release Notes
`UUIDv4()`: returns a UUIDv4

## How can this PR be tested?

This PR adds automatic tests to test the new function 

## PR quality check


mysys/my_uuid_v4.c
```diff
@@ -0,0 +1,94 @@
...
+
+int is_random_bits_generation_successful(unsigned char *rand_var, int size)
+{
+  return my_random_bytes(rand_var, size) != MY_AES_OK;
+}
+
+/**
+  Init structures needed for my_uuid_v4
+*/
+void my_uuid_v4_init()
```
grooverdan, line 49: I'm not sure I see a point of this init flag. Is there a good precidence? otherwise remove.

mysys/my_uuid_v4.c
```diff
@@ -0,0 +1,94 @@
...
+#include <errmsg.h>
+
+static my_bool my_uuid_v4_inited= 0;
+
+#define UUID_VERSION                         0x4000
+#define UUID_VERSION_MASK                    0x0FFF
+#define UUID_VARIANT                         0x8000000000000000
+#define UUID_VARIANT_MASK                    0x3FFFFFFFFFFFFFFF
+
+int is_random_bits_generation_successful(unsigned char *rand_var, int size)
```
grooverdan, line 41: static function

mysys/my_uuid_v4.c
```diff
@@ -0,0 +1,94 @@
...
+                                           sizeof(random_c)))
+  {
+    my_printf_error(CR_UNKNOWN_ERROR,
+                "Failed to generate the random section of UUIDv4\n",
+                MYF(ME_FATAL|ME_ERROR_LOG));
+  }
+
+  random_b_and_version= (uint16)
+    ((random_b & UUID_VERSION_MASK) | UUID_VERSION);
+  random_c_and_variant= (uint64) ((random_c & UUID_VARIANT_MASK) | UUID_VARIANT);
```
grooverdan, line 83: is casting here necessary?

mysys/my_uuid_v4.c
```diff
@@ -0,0 +1,94 @@
...
+  if (is_random_bits_generation_successful((unsigned char *) &random_a,
+                                           sizeof(random_a)) ||
+      is_random_bits_generation_successful((unsigned char *) &random_b,
+                                           sizeof(random_b)) ||
+      is_random_bits_generation_successful((unsigned char *) &random_c,
+                                           sizeof(random_c)))
+  {
+    my_printf_error(CR_UNKNOWN_ERROR,
+                "Failed to generate the random section of UUIDv4\n",
+                MYF(ME_FATAL|ME_ERROR_LOG));
```
grooverdan, line 78: FATAL seems a little extreme. Probably return an error bool here and let the application push the error to the SQL layer.

plugin/type_uuid/item_uuidfunc.cc
```diff
@@ -44,3 +44,30 @@ bool Item_func_uuid::val_native(THD *, Native *to)
...
+
+String *Item_func_uuid_v4::val_str(String *str)
+{
+  DBUG_ASSERT(fixed());
+  str->alloc(uuid_len()+1);
+  str->length(uuid_len());
+  str->set_charset(collation.collation);
+
+  uchar buf[MY_UUID_SIZE];
+  my_uuid_v4(buf);
```
grooverdan, line 61: on error - `my_error(.....)`

grooverdan (commented):
nice work.

mysys/my_uuid_v4.c
StefanoPetrilli: The windows build fails without this casting

plugin/type_uuid/item_uuidfunc.h
```diff
@@ -64,4 +64,35 @@ class Item_func_uuid: public Item_func_sys_guid
   { return get_item_copy<Item_func_uuid>(thd, this); }
 };
 
+class Item_func_uuid_v4: public Item_str_func
+{
```
abarkov, line 68: I suggest to derive from Item_func_sys_guid, like Item_func_uuid does.
It will help to avoid duplicating of a few methods:
- fix_length_and_dec
- const_item
- used_tables
- check_vcol_func_processor

mysys/my_uuid_v4.c
```diff
@@ -0,0 +1,75 @@
...
+ random_b_and_version           6-7     Random CSPRNG 16 bits multiplexed
+                                        with the version number.
+ random_c_and_variant           8-15    Random CSPRNG 64 bits multiplexed
+                                        with the variant number.
+*/
+
+#include "mysys_priv.h"
+#include <my_rnd.h>
+#include <m_string.h>
+#include <myisampack.h> /* mi_int2store, mi_int4store */
```
grooverdan, line 31: Technically `int2store` is there globally but I do appreciate this was in the original `my_uuid.c` file. Optional change.

plugin/type_uuid/item_uuidfunc.cc
```diff
@@ -17,6 +17,7 @@
 #include "mariadb.h"
 #include "item_uuidfunc.h"
 #include "sql_type_uuid.h"
+#include "my_uuid_v4.c"
```
grooverdan, line 20: This is causing a failure on the windows build. Include the header or optionally move the uuid_v2 implementation inside this file.

plugin/type_uuid/mysql-test/type_uuid/func_uuidv4_plugin.result
```diff
@@ -0,0 +1,35 @@
+#
+# Start of 11.5 tests
```
grooverdan, line 2: Technically its going to make it into 11.6. The 11.5 preview is already out.

Don't worry about change github PR, but if you could edit the test to assume 11.6 that would be good.

plugin/type_uuid/mysql-test/type_uuid/func_uuidv4_plugin.test
```diff
@@ -0,0 +1,30 @@
+--echo #
+--echo # Start of 11.5 tests
```
grooverdan, line 2: 11.6

plugin/type_uuid/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1335  USA
 
+INCLUDE_DIRECTORIES(${CMAKE_SOURCE_DIR}/mysys)
```
grooverdan, line 16: This was needed to include the `.c` file. So omit?

plugin/type_uuid/item_uuidfunc.cc
```diff
@@ -44,3 +45,36 @@ bool Item_func_uuid::val_native(THD *, Native *to)
...
+  DBUG_ASSERT(fixed());
+  str->alloc(uuid_len()+1);
+  str->length(uuid_len());
+  str->set_charset(collation.collation);
+
+  uchar buf[MY_UUID_SIZE];
+  if (!my_uuid_v4(buf)) {
+    my_error(ER_INTERNAL_ERROR, MYF(0),
+    "Failed to generate a random value for UUIDv4");
+  }
```
grooverdan, line 65: Have you tested to see what it takes to get this? I suspect once a certain amount of system entropy has been consumed but I haven't looked at the openssl implementation.

grooverdan (changes requested):
Thanks for address @abarkov 's review.

Just addressing the windows compile issue, and get ready for a 11.6 and will be ready.

Thanks for the contribution.

plugin/type_uuid/item_uuidfunc.cc
StefanoPetrilli: The docs says:

> If the entropy source fails or is not available, the CSPRNG will enter an error state and refuse to generate random bytes

I generated a big amount of UUIDv4 but I could not get to the point of making it fail on my machine.
I hard-coded it to always fail and checked that the behavior is as expected in case of failure.
If you have any idea on how to test this let me know.

plugin/type_uuid/CMakeLists.txt
grooverdan: `LINK_LIBRARIES mysys` inside MYSQL_ADD_PLUGIN to avoid link error and.

plugin/type_uuid/item_uuidfunc.cc
grooverdan: That's good enough. Just making sure a big UUIDv4 user won't immediately hit a fault.

grooverdan (approved):

grooverdan:
I see last push went from `intXstore` -> `mi_intXstore` and then didn't bigendian fail like [this test failure](https://buildbot.mariadb.org/#/builders/476/builds/17617/steps/7/logs/stdio) - something I could watch out for other BE test fails.

abarkov (commented):
Hello Stefano.
Can you please squash the two commits (the code and the tests) into a single commit?

This is how we usually do in MariaDB: the change and the covering test go in a single commit.

Also, I'm seeing some code and CMake changes in the patch with tests. This is confusing why we need a second commit fixing something that was done just in a previous commit. It's better to have everything done properly right in the first commit.

Thanks.

mysys/my_uuid_v4.c
```diff
@@ -0,0 +1,75 @@
...
+  random_b_and_version= (uint16)
+    ((random_b & UUID_VERSION_MASK) | UUID_VERSION);
+  random_c_and_variant= (uint64)
+    ((random_c & UUID_VARIANT_MASK) | UUID_VARIANT);
+
+  mi_int6store(to, random_a);
+  mi_int2store(to+6, random_b_and_version);
+  mi_int8store(to+8, random_c_and_variant);
+  return 1;
+}
```
abarkov, line 75: I propose some improvements into this code.
- You generate in some cases more random bytes than you later use. For example, in random_a you generate 8 bytes but use only 6 bytes. So 2/8 of the initialized data is thrown away.
- It's not clear why we need to store random bytes into integers first, then move these integers to the final destination. Why not generate random bytes directly in their final destination?
- Why we need three separate calls for my_random_bytes(). Looking inside `mysys_ssl/my_crypt.cc` I think that every call is quite expensive. It's cheaper to have one call initializing 16 bytes

Collecting all together:

The function `is_random_bits_generation_successful()` is not needed.
Let's generate 16 random bytes on the whole buffer, with one test for MY_AES_OK:
```
  if (my_random_bytes(to, 16) != MY_AES_OK))
    return 1;
```

The UUID structure is:

`llllllll-mmmm-Vhhh-vsss-nnnnnnnnnnnn`

So now we need to modify two very small fragments of the output:
- the byte with the version, which is to[6]
- the byte with the variant, which is to[8]


I suggest to include a comment with the UUID structure mentioned above, to make it clearer what's going on. Also, add comments how  many bits of to[6] and to[8] remain random, and how many bits gets replaced to the version and the variant.

Thanks.

plugin/type_uuid/item_uuidfunc.cc
abarkov: Right. The source file should not be included here. Please do the following:
- Add the prototype for the new function my_uuid_v4()  into my_sys.h
- Link the new source file into the mysys library. Another option is just to put `my_uuid_v4()` directly into mysys.c. It's quite small, so possibly it does not need a separate file at all.

mysys/my_uuid_v4.c
abarkov: I don't think we need to use int2store() at all, and therefore don't need to include `myisampack.h`. Please see my proposal for `my_uuid_v4()`.

plugin/type_uuid/item_uuidfunc.cc
abarkov: (in particular, which header files are included)

mysys/my_uuid_v4.c
StefanoPetrilli: Applied your tips, and I think the code now looks much better :)
Thanks for such a detailed review.

abarkov (approved):
The patch looks very good. Thanks for addressing review suggestions!

abarkov:
Hello @StefanoPetrilli.

I'm preparing a branch with UUIDv4() for testing by our QA team.
While working on it I realized there was a wrong inheritance in MariaDB sources.
The UUID() erroneously inherited some VARCHAR features from SYS_GUID(), which should not happen.

I reported this problem as:
https://jira.mariadb.org/browse/MDEV-33827 -  UUID() returns a NULL-able result
which is now fixed in 11.5.

This is the patch:
https://github.com/MariaDB/server/commit/9b02b7c77a7f106f33da579279d36b48a8dcad9d

The class Item_func_uuid now derives from something else than Item_func_sys_guid.

Unfortunately, I had to duplicate these methods responsible for non-deterministic behavior:
```
 bool const_item() const;
 table_map used_tables() const;
 bool check_vcol_func_processor(void *arg);
```
But this change guarantees that the functions UUID(), UUIDv4(), UUIDv7() have exactly the same behavior with all other expressions returning UUID, like table columns and `CAST(expr AS UUID)`.

abarkov:
After fixing [MDEV-33827](https://jira.mariadb.org/browse/MDEV-33827), I cherry-picked your patch.
Then pushed two patches on top of them: they are marked as "Fixup" and "Fixup 2" in their commit comments.

"Fixup" addresses the things we came up during discussions with the team (Daniel Black and Sergey Golubchik):
- We don't need the code from my_uuid_v4.c in the server (at least at this point). So it was moved to plugin/type_uuid/, into a new class UUIDv4.
- A failure in my_random_bytes() should not raise an error. A warning with a fallback should be OK instead.

"Fixup 2" additionally introduces a template Item_func_uuid_vx, to share as much as possible code between UUID() and UUIDv4() and the coming soon UUIDv7().



Please have a look into this branch:
https://github.com/MariaDB/server/commits/bb-11.5-bar-MDEV-11339/

Do you agree with these additional changes?

StefanoPetrilli:
Hello @abarkov,

My only comment is that falling back to UUIDv1 in case of entropy error might not be the most appropriate choice. If I try to generate a UUIDv4 I would rather get an error than a UUIDv1 because they have different properties and is not what I asked for. 

What do you think?

StefanoPetrilli:
Hello @abarkov,

I agree that the probability of an error is close to zero and for this reason I think both versions would be acceptable but personally, I like your last version more.

grooverdan:
Thanks @StefanoPetrilli very much for the contribution and making the base.

I'm closing this now per MDEV the current version is [bb-11.5-bar-MDEV-11339](https://github.com/MariaDB/server/tree/bb-11.5-bar-MDEV-11339) and is undergoing testing for 11.6.

PR #3145, opened 2024-03-21
MDEV-33087 ALTER TABLE...ALGORITHM=COPY should build indexes more efficiently

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-33087*

## Description
- During copy algorithm, InnoDB should use bulk insert operation for row by row insert operation. By doing this, copy algorithm can effectively build indexes.

## How can this PR be tested?
Tested with large data set:

Existing alter copy algorithm:

MariaDB [test]> alter table t1 force, algorithm=copy;
Query OK, 6291456 rows affected (12 min 18.519 sec)    
Records: 6291456  Duplicates: 0  Warnings: 0

With bulk insert (copy algorithm)

MariaDB [test]> alter table t1 force, algorithm=copy;
Query OK, 6291456 rows affected (7 min 30.760 sec)     
Records: 6291456  Duplicates: 0  Warnings: 0

create..select copy algorithm:

MariaDB [test]> create table t2 select * from t1;
Query OK, 6619136 rows affected (7 min 5.126 sec)
Records: 6619136  Duplicates: 0  Warnings: 0

create..select (with bulk insert):

MariaDB [test]> create table t2 select * from t1;
Query OK, 6619136 rows affected (4 min 54.647 sec)
Records: 6619136  Duplicates: 0  Warnings: 0

## PR quality check


storage/innobase/row/row0merge.cc
```diff
@@ -5333,8 +5333,19 @@ dberr_t row_merge_bulk_t::write_to_table(dict_table_t *table, trx_t *trx)
       continue;
 
     dberr_t err= write_to_index(i, trx);
-    if (err != DB_SUCCESS)
+    switch(err)
+    {
```
dr-m, line 5337: This should be written as
```c++
    switch (dberr_t err= write_to_index(i, trx) {
```

dr-m (commented):
This looks mostly OK. I think that we must keep using the old code path if `FOREIGN KEY` constraints need to be enforced.

sql/sql_insert.cc
```diff
@@ -4317,7 +4317,17 @@ bool select_insert::prepare_eof()
   if (info.ignore || info.handle_duplicates != DUP_ERROR)
       if (table->file->ha_table_flags() & HA_DUPLICATE_POS)
         table->file->ha_rnd_end();
-  table->file->extra(HA_EXTRA_END_ALTER_COPY);
+  int create_err= table->file->extra(HA_EXTRA_END_ALTER_COPY);
+  if (error <= 0)
```
dr-m, line 4321: Why is the condition not `if (!error)`? Some overridden `handler:end_bulk_insert()` could return a negative error code.

sql/sql_table.cc
```diff
@@ -12085,9 +12085,17 @@ copy_data_between_tables(THD *thd, TABLE *from, TABLE *to, bool ignore,
...
+  {
+    int alt_error= to->file->extra(HA_EXTRA_END_ALTER_COPY);
+    if (alt_error == HA_ERR_FOUND_DUPP_KEY)
+    {
+      uint key_nr= to->file->get_dup_key(alt_error);
+      if ((int)key_nr >= 0 && key_nr < to->s->keys)
+        print_keydup_error(to, &to->key_info[key_nr], MYF(0));
+    }
+    error= alt_error;
+  }
```
dr-m, line 12098: Again, why not `!error` instead of `error <= 0`? In this case, we could overwrite a negative `error` code with `alt_error=0`.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15741,6 +15741,22 @@ ha_innobase::extra(
...
+		if (m_prebuilt->table->skip_alter_undo) {
+			if (dberr_t err= trx->bulk_insert_apply()) {
+				if (err == DB_DUPLICATE_KEY) {
+					return HA_ERR_FOUND_DUPP_KEY;
+				}
+				return 1;
+			}
+
+			trx->end_bulk_insert(*m_prebuilt->table);
+			trx->bulk_insert = false;
```
dr-m, line 15753: How will this work on a partitioned table? Each partition would be a separate table from the InnoDB point of view.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15741,6 +15741,22 @@ ha_innobase::extra(
...
+			if (!dict_stats_is_persistent_enabled(
+					m_prebuilt->table)) {
+				dict_stats_update_if_needed(
+					m_prebuilt->table, *trx);
+			}
```
dr-m, line 15758: It would be nice to have a comment on why this is needed only if `STATS_PERSISTENT=0`.

storage/innobase/include/row0merge.h
```diff
@@ -416,6 +416,15 @@ row_merge_read_rec(
...
 	MY_ATTRIBUTE((warn_unused_result));
 
+/* Report an InnoDB error to the client by invoking my_error(). */
+ATTRIBUTE_COLD __attribute__((nonnull))
+void
+my_error_innodb(
+/*============*/
+        dberr_t         error,  /*!< in: InnoDB error code */
+        const char*     table,  /*!< in: table name */
+        ulint           flags);  /*!< in: table flags */
```
dr-m, line 426: Let’s avoid the `/*====*/` uglification and the 3-column formatting of function parameters in any new code.

storage/innobase/row/row0ins.cc
```diff
@@ -3393,7 +3400,8 @@ row_ins_index_entry(
 		if (auto t= trx->check_bulk_buffer(index->table)) {
 			/* MDEV-25036 FIXME: check also foreign key
 			constraints */
-			ut_ad(!trx->check_foreigns);
+			ut_ad(index->table->skip_alter_undo
+			      || !trx->check_foreigns);
```
dr-m, line 3404: It would be good to have a comment that explains where the `FOREIGN KEY` constraints will be enforced during `ALTER TABLE…ALGORITHM=COPY`.

storage/innobase/row/row0merge.cc
```diff
@@ -5333,8 +5336,19 @@ dberr_t row_merge_bulk_t::write_to_table(dict_table_t *table, trx_t *trx)
...
```
dr-m, line 5340: There is a missing space and an unwanted line break here:
```c++
switch (err) {
```

dr-m (commented):
Here are some initial comments. I think that [MDEV-33809](https://jira.mariadb.org/browse/MDEV-33809) needs to be fixed before this can be merged.

sql/sql_insert.cc
Thirunarayanan: error is set to -1 when handler encounters HA_ERR_END_OF_FILE. So error variable is > 0 then we encounter error.

storage/innobase/handler/ha_innodb.cc
Thirunarayanan: All the partition will do bulk insert operation.
```
CREATE TABLE t1 ( i INT NOT NULL AUTO_INCREMENT PRIMARY KEY, f INT )
        ENGINE = InnoDB PARTITION BY HASH(i) PARTITIONS 2;
INSERT INTO t1 VALUES (2, 2), (3, 3), (4, 4), (5, 5);
ALTER TABLE t1 ADD PARTITION PARTITIONS 2;
DROP TABLE t1;
```
In the above alter, t1 will get rebuild. All the partition will do bulk insert operation. Each partition
will call `HA_EXTRA_BEGIN_ALTER_COPY` and `HA_EXTRA_BEGIN_ALTER_COPY`

sql/sql_table.cc
```diff
@@ -11746,6 +11746,60 @@ bool mysql_trans_commit_alter_copy_data(THD *thd)
...
+  /* Duplicate key error. */
+  if (unlikely(alter_ctx->fk_error_if_delete_row))
+  {
+    /* We are trying to omit a row from the table which serves
+    as parent in a foreign key. This might have broken
+    referential integrity so emit an error. Note that we
+    can't ignore this error even if we are
+    executing ALTER IGNORE TABLE. IGNORE allows to skip rows, but
+    doesn't allow to break unique or foreign key constraints, */
```
dr-m, line 11777: Is the `ALTER IGNORE TABLE` scenario covered by some test? That test case name had better be mentioned in the commit message. (Mentioning it in a source code comment is a worse idea, because someone could rename or refactor the test and forget to adjust that comment. In the commit message, the comment would naturally refer to the current state of affairs.)

sql/sql_table.cc
```diff
@@ -11746,6 +11746,60 @@ bool mysql_trans_commit_alter_copy_data(THD *thd)
...
+    my_error(ER_FK_CANNOT_DELETE_PARENT, MYF(0),
+             alter_ctx->fk_error_id,
+             alter_ctx->fk_error_table);
+    goto func_exit;
+  }
+  if (ignore)
+    return false;
+  /* Ordinary ALTER TABLE. Report duplicate key error. */
+  uint key_nr= to->file->get_dup_key(error);
+  if ((int)key_nr >= 0)
```
dr-m, line 11787: A less hacky way would seem to be to write
```c++
  uint key_nr= to->file->get_dup_key(error);
  if (key_nr >= MAX_KEY)

File: sql/sql_table.cc [hunk 11]
```diff
@@ -12007,57 +12061,13 @@ copy_data_between_tables(THD *thd, TABLE *from, TABLE *to, bool ignore,
...
-      {
-        /* Not a duplicate key error. */
-	to->file->print_error(error, MYF(0));
-        error= 1;
-	break;
-      }
+      if (copy_data_error_handler(error, ignore, to, thd, alter_ctx))
+        break;
       else
       {
```
dr-m, line 12067: The `else` is redundant. We could just write
```c++
if (copy_data_error_handler(error, ignore, to, thd, alter_ctx)))
  break;
DBUG_ASSERT(ignore);
to->file->restore_auto_increment(prev_insert_id);
delete_count++;
```
Above, I would also replace the comment with an assertion.

Even better might be to rename the function and invert its return value to make it self-explanatory:
```c++
if (!copy_data_error_ignore(error, ignore, to, thd, alter_ctx)))
  break;
```

sql/sql_table.cc
Thirunarayanan: error is set to -1 when handler encounters HA_ERR_END_OF_FILE. So error variable is > 0 then we encounter error.

storage/innobase/handler/handler0alter.cc
```diff
@@ -845,14 +845,14 @@ inline void dict_table_t::rollback_instant(
...
+@param table table name
+@param flags table flags */
+ATTRIBUTE_COLD __attribute__((nonnull))
 void
 my_error_innodb(
-/*============*/
-	dberr_t		error,	/*!< in: InnoDB error code */
-	const char*	table,	/*!< in: table name */
-	ulint		flags)	/*!< in: table flags */
+  dberr_t error, const char *table, ulint flags)
```
dr-m, line 855: I’d declare the parameters in the same line, to be more `grep` friendly.

storage/innobase/trx/trx0rec.cc
```diff
@@ -1871,6 +1871,10 @@ trx_undo_report_row_operation(
 			    *clust_entry, *index, trx)) {
 			return err;
 		}
+
+		if (index->table->skip_alter_undo) {
+			return DB_SUCCESS;
+		}
 	} else {
 		bulk = false;
 	}
```
dr-m, line 1880: I do not quite understand this. Why can’t we just always
```c++
		return m.first->second.bulk_insert_buffered(*clust_entry, *index, trx);
```
when we are executing a bulk insert? Why would we need to write row-level undo log records for any bulk insert?

storage/innobase/row/row0ins.cc
```diff
@@ -2711,6 +2711,20 @@ row_ins_clust_index_entry_low(
 
 	DBUG_EXECUTE_IF("row_ins_row_level", goto skip_bulk_insert;);
 
+	/* Ignore the bulk insert operation for temporary,
+	versioning table. In case of foreign key constraint,
```
dr-m, line 2715: I would rephrase "temporary, versioning table" as "temporary or versioned table".

storage/innobase/row/row0ins.cc
```diff
@@ -2711,6 +2711,20 @@ row_ins_clust_index_entry_low(
...
+	operation if the table contains foreign key relation */
+	if (index->table->skip_alter_undo
+	    && !index->table->is_temporary()
+	    && !index->table->versioned()
+	    && index->table->foreign_set.empty()) {
+		trx_start_if_not_started(trx, true);
+		goto bulk_insert;
+	}
+
 	if (!(flags & BTR_NO_UNDO_LOG_FLAG)
```
dr-m, line 2728: Can we avoid adding a new bunch of unlikely conditions at the start of a busy function (inserting a record into a clustered index)? Can we relax the subsequent `if` statement condition so that `goto bulk_insert` will not be needed? That condition is written in such a way that the least likely or easiest-to-evaluate conditions are checked first.

storage/innobase/row/row0ins.cc
```diff
@@ -2711,21 +2711,44 @@ row_ins_clust_index_entry_low(
...
+		temporary or versioning table.
+		In case of foreign key constraint,
+		InnoDB has to check the constraint validity for
+		each row. But bulk insert operation doesn't
+		check foreign constraint. Skip bulk insert
+		operation if the table contains foreign key
+		relation */
+		if (index->table->skip_alter_undo) {
+			trx_start_if_not_started(trx, true);
+			trx->bulk_insert= true;
```
dr-m, line 2741: I do not see any code that refers to a `FOREIGN KEY` here. I suppose that it is being checked at a higher level.

It could be useful to add a comment that we are only executing this code at the start of an `ALTER TABLE` statement if that is the case, and note how and where subsequent written rows will get processed.

Maybe the new Boolean configuration parameter and its name should be specific to only this code path, for example as follows:
```c++
if (UNIV_LIKELY(!index->table->skip_alter_undo)) {
	trx->bulk_insert = true;
} else if (!innodb_alter_copy_bulk) {
	goto skip_bulk_insert;
} else {
	/* We are writing the first row in ALTER TABLE. */
```

storage/innobase/row/row0merge.cc
```diff
@@ -5333,8 +5336,18 @@ dberr_t row_merge_bulk_t::write_to_table(dict_table_t *table, trx_t *trx)
       continue;
 
     dberr_t err= write_to_index(i, trx);
-    if (err != DB_SUCCESS)
+    switch(err) {
```
dr-m, line 5339: Missing space after `switch`.

sql/sql_table.cc
dr-m: In a6bb0ca6ca2ff12a1110a2bf9c6f5bab84c57f7b this is covered by the addition to the test `innodb.foreign_key`.

dr-m (commented):
Thank you, I think that we are almost there.

mysql-test/suite/innodb/t/foreign_key.test
```diff
@@ -1186,6 +1186,29 @@ DROP TABLE binaries, collections;
...
+) ENGINE=InnoDB;
+INSERT INTO t2 VALUES("MySQL", "MySQL", 1),
+                     ("MariaDB", "MariaDB", 1);
+ALTER TABLE t1 ADD FOREIGN KEY (f1) REFERENCES t2 (f2);
+# MDEV-33927 TODO: change the warning message
+--error ER_FK_CANNOT_DELETE_PARENT
+ALTER TABLE t2 ADD FOREIGN KEY (f2) REFERENCES t2 (f2),
+ADD UNIQUE INDEX(f3);
+drop table t1, t2;
 SET GLOBAL innodb_stats_persistent = @save_stats_persistent;
```
dr-m, line 1212: This is missing
```
--echo # End of 10.11 tests
```
right after the added lines.

mysql-test/suite/innodb/t/foreign_key.test
```diff
@@ -1186,6 +1186,29 @@ DROP TABLE binaries, collections;
...
+INSERT INTO t1 VALUES("MySQL", "InnoDB"), ("MariaDB", "NDB");
+
+CREATE TABLE t2
+(
+f1 VARCHAR(32)BINARY NOT NULL,
+f2 VARCHAR(255)BINARY NOT NULL,
+f3 int, PRIMARY KEY (f1), INDEX(f1), INDEX(f2)
+) ENGINE=InnoDB;
+INSERT INTO t2 VALUES("MySQL", "MySQL", 1),
+                     ("MariaDB", "MariaDB", 1);
```
dr-m, line 1205: In the SQL standard, double quotes `"` are for quoting identifiers (such as table or column names that contain spaces), not for string literals. Let’s try to use the standard syntax where possible: `'MariaDB'`, not `"MariaDB"`.

sql/sql_table.cc
```diff
@@ -11752,6 +11752,60 @@ bool mysql_trans_commit_alter_copy_data(THD *thd)
...
 }
 
+/** Handle the error when copying data from source to target table.
+@param error          error code
+@param ignore         alter ignore statement
+@param to             target table handler
+@param thd            Mysql Thread
+@param alter_ctx      Runtime context for alter statement
+@return false in case of error
+@return true in case of skipping the row and continue alter operation */
```
dr-m, line 11762: `@retval` instead of `@return`

storage/innobase/row/row0ins.cc
```diff
@@ -2703,20 +2703,42 @@ row_ins_clust_index_entry_low(
...
-	    && !trx->check_unique_secondary && !trx->check_foreigns
-	    && !trx->dict_operation
+	if (((!(flags & BTR_NO_UNDO_LOG_FLAG)
+	      && page_is_empty(block->page.frame)
+	      && !entry->is_metadata() && !trx->duplicates
+	      && !trx->check_unique_secondary && !trx->check_foreigns
+	      && !trx->dict_operation
+	      && !index->table->skip_alter_undo
+	      && !index->table->is_active_ddl())
+	      || (index->table->skip_alter_undo
```
dr-m, line 2713: The indentation looks misleading here. The `||` should be indented by 1 less space because it is at a higher level than the rest.

But, also the starting parenthesis that is being closed before the `||` looks possibly misplaced. Do we really need to evaluate `index->table->skip_alter_undo` if the `flags` bit is set, or the page is not empty, or it is not the index root page?

Please try to find the minimum necessary change to this `if` expression.

storage/innobase/row/row0ins.cc
dr-m: I think that a source code comment here about the `FOREIGN KEY` constraint enforcement would be useful. In which function is that check executed?

storage/innobase/row/row0merge.cc
```diff
@@ -5333,8 +5336,18 @@ dberr_t row_merge_bulk_t::write_to_table(dict_table_t *table, trx_t *trx)
...
+    switch (err) {
+    default:
+      if (table->skip_alter_undo)
+        my_error_innodb(err, table->name.m_name, table->flags);
+err_exit:
       return err;
+    case DB_SUCCESS:
+      break;
+    case DB_DUPLICATE_KEY:
+      trx->error_info = index;
```
dr-m, line 5348: We do not want space before `=`.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -19672,6 +19692,10 @@ static MYSQL_SYSVAR_BOOL(force_primary_key,
   "Do not allow creating a table without primary key (off by default)",
   NULL, NULL, FALSE);
 
+static MYSQL_SYSVAR_BOOL(alter_copy_bulk, innodb_alter_copy_bulk,
+  PLUGIN_VAR_NOCMDARG,
+  "Allow bulk insert operation for copy alter operation", NULL, NULL, TRUE);
```
dr-m, line 19697: As far as I can tell, there is no regression test that would enable this setting. It would be useful to cover both normal and partitioned tables, both small ones (merge sorting happens in memory) and larger ones (temporary files will be used). It would also be good to cover `ALTER IGNORE TABLE` and operations that involve `FOREIGN KEY`.

dr-m (commented):
Thank you, the code changes look fine to me, but I think that we need to add some test coverage.

mysql-test/suite/innodb/t/bulk_copy_alter.combinations
```diff
@@ -0,0 +1,5 @@
+[bulk_alter_copy]
+--innodb_alter_copy_bulk=ON
+
+[non_bulk_alter_copy]
+--innodb_alter_copy_bulk=OFF
```
dr-m, line 5: I think that we should use shorter names if we rename the test (and the `.combinations` file) to more self-descriptive `innodb_alter_copy_bulk` or `alter_copy_bulk`.
```
[ON]
--innodb_alter_copy_bulk=ON

[OFF]
--innodb_alter_copy_bulk=OFF
```
Then the test log will display `innodb.alter_copy_bulk,ON` which can intuitively be interpreted as a test of `innodb_alter_copy_bulk=ON`.

storage/innobase/row/row0ins.cc
```diff
@@ -2703,22 +2703,43 @@ row_ins_clust_index_entry_low(
...
+	if (((!(flags & BTR_NO_UNDO_LOG_FLAG)
+	      && page_is_empty(block->page.frame)
+	      && !entry->is_metadata() && !trx->duplicates
+	      && !trx->check_unique_secondary && !trx->check_foreigns
+	      && !trx->dict_operation
+	      && !index->table->is_active_ddl()
+	      && block->page.id().page_no() == index->page)
+	     || (index->table->skip_alter_undo
+		 && !index->table->is_temporary()
+		 && index->table->foreign_set.empty()))
```
dr-m, line 2715: As far as I can tell, `flags == BTR_NO_UNDO_LOG_FLAG | BTR_NO_LOCKING_FLAG` in the case that we are interested in. That is, the condition `index->table->skip_alter_undo` could be replaced with that. But, we can do even better.

If we introduced one more `flags` bit and set it in `row_ins_clust_index_entry()` appropriately, we could replace all these 3 conditions with just one `flags ==` test.

storage/innobase/row/row0ins.cc
```diff
@@ -2703,22 +2703,43 @@ row_ins_clust_index_entry_low(
...
+	      && page_is_empty(block->page.frame)
+	      && !entry->is_metadata() && !trx->duplicates
+	      && !trx->check_unique_secondary && !trx->check_foreigns
+	      && !trx->dict_operation
+	      && !index->table->is_active_ddl()
+	      && block->page.id().page_no() == index->page)
+	     || (flags == (BTR_NO_UNDO_LOG_FLAG | BTR_NO_LOCKING_FLAG)
+		 && !index->table->is_temporary()
+		 && index->table->foreign_set.empty()))
 	    && !index->table->n_rec_locks
```
dr-m, line 2716: In the `flags == BTR_NO_UNDO_LOG_FLAG | BTR_NO_LOCKING_FLAG` case, I think that it would be better to assert that some of the conditions that are being checked in the `!(flags & BTR_NO_UNDO_LOG_FLAG)` hold. Maybe this could be rewritten for clarity as a nested `if`:
```c++
if (!(flags & BTR_NO_UNDO_LOG_FLAG) && … && !index->table->has_spatial_index()
    && !(dict_table_is_partition(index->table || thd_sql_command(…) == SQLCOM_INSERT)) {
	if (!index->table->n_rec_locks && !index->table->versioned() && !index->table->skip_alter_undo) {
			trx->bulk_insert = true;
	}
} else if (flags == (BTR_NO_UNDO_LOG_FLAG | BTR_NO_LOCKING_FLAG)
	   && …) {
	ut_ad(index->table->skip_alter_undo);
	ut_ad(!entry->is_metadata());
	ut_ad(index->table->is_active_ddl());
	if (innodb_alter_copy_bulk) {
		ut_ad(page_is_empty(block->page.frame));
		trx->bulk_insert = true;
	}
}
```
Would this make sense? Add  as many assertions as you think should hold. Do we really need the `index->table->is_active_ddl()` condition for the `!(flags & BTR_NO_UNDO_LOG_FLAG)` case?

sql/sql_table.cc
```diff
@@ -11766,6 +11766,60 @@ bool mysql_trans_commit_alter_copy_data(THD *thd)
...
+copy_data_error_ignore(int &error, bool ignore, TABLE *to,
+                       THD *thd, Alter_table_ctx *alter_ctx)
+{
+  if (to->file->is_fatal_error(error, HA_CHECK_DUP))
+  {
+    /* Not a duplicate key error. */
+    to->file->print_error(error, MYF(0));
+    error = 1;
+func_exit:
+    return false;
```
dr-m, line 11787: I think that it would better to replace each `goto func_exit` with `return false` and to omit this label.

sql/sql_table.cc
```diff
@@ -11766,6 +11766,60 @@ bool mysql_trans_commit_alter_copy_data(THD *thd)
...
+@retval true in case of skipping the row and continue alter operation */
+static bool
+copy_data_error_ignore(int &error, bool ignore, TABLE *to,
+                       THD *thd, Alter_table_ctx *alter_ctx)
+{
+  if (to->file->is_fatal_error(error, HA_CHECK_DUP))
+  {
+    /* Not a duplicate key error. */
+    to->file->print_error(error, MYF(0));
+    error = 1;
```
dr-m, line 11785: Monty does not want space before `=`.

storage/innobase/row/row0ins.cc
```diff
@@ -2708,33 +2708,28 @@ row_ins_clust_index_entry_low(
...
-		if (!index->table->is_temporary()) {
+		if (!index->table->has_spatial_index()
+		    && !index->table->n_rec_locks
+		    && !index->table->versioned()
+		    && !index->table->skip_alter_undo) {
+			trx->bulk_insert= true;
+			if (index->table->is_temporary()) {
+				goto skip_bulk_insert;
+			}
 			err = lock_table(index->table, NULL, LOCK_X, thr);
```
dr-m, line 2722: Can we avoid most of these changes? Especially the `goto skip_bulk_insert` after enabling bulk insert for temporary tables is rather confusing, although it is not logically incorrect.

I think that we should always have `index->table->skip_alter_undo==false` in this branch, so we’d better remove that condition and add `ut_ad(!index->table->skip_alter_undo)`.

Why are we removing the check that the current page must be the root page? This is the DML branch after all.

Can the condition on `!index->table->is_active_ddl()` actually ever hold here? If we are about to execute `ALTER TABLE…ALGORITHM=INPLACE, LOCK=NONE` on an empty table, don’t we handle that specially nowadays? I can only imagine that purge might be able to remove any history and make the table empty after `ha_innobase::prepare_inplace_alter_table()` returned. So, maybe that condition can actually hold and should not be removed.

storage/innobase/row/row0ins.cc
```diff
@@ -2773,6 +2768,30 @@ row_ins_clust_index_entry_low(
...
 		}
+	} else if (flags == (BTR_NO_UNDO_LOG_FLAG | BTR_NO_LOCKING_FLAG)
+		   && !index->table->has_spatial_index()
+		   && !index->table->versioned()
+		   && !index->table->is_temporary()
+		   && !index->table->n_rec_locks) {
+
+		ut_ad(index->table->skip_alter_undo);
+		ut_ad(!entry->is_metadata());
+		ut_ad(!index->table->is_active_ddl());
```
dr-m, line 2779: It might be less confusing to rename `is_active_ddl()` to something like `is_native_online_ddl()`.

storage/innobase/row/row0merge.cc
```diff
@@ -5320,6 +5320,9 @@ dberr_t row_merge_bulk_t::write_to_index(ulint index_no, trx_t *trx)
   else if (index->is_primary() && table->persistent_autoinc)
     btr_write_autoinc(index, table->autoinc - 1);
   err= btr_bulk.finish(err);
+  if (err == DB_SUCCESS && index->is_clust())
+    table->stat_n_rows= (file && file->fd != OS_FILE_CLOSED)
+                        ? file->n_rec : buf.n_tuples;
```
dr-m, line 5325: The second line should be indented only 2 columns deeper than the first line.

storage/innobase/row/row0merge.cc
```diff
@@ -5333,8 +5336,18 @@ dberr_t row_merge_bulk_t::write_to_table(dict_table_t *table, trx_t *trx)
       continue;
 
     dberr_t err= write_to_index(i, trx);
-    if (err != DB_SUCCESS)
+    switch (err) {
+    default:
+      if (table->skip_alter_undo)
+        my_error_innodb(err, table->name.m_name, table->flags);
+err_exit:
       return err;
```
dr-m, line 5344: I don’t think that invoking `goto` instead of `return` will save much here.

dr-m (commented):
This looks OK, except for the condition for enabling bulk insert in `row_ins_clust_index_entry_low()`.

storage/innobase/row/row0merge.cc
dr-m: This comment was not addressed in ecad74fee7f410c4273ff6d7e3a2f09659d689fc.

storage/innobase/row/row0ins.cc
```diff
@@ -2704,37 +2704,35 @@ row_ins_clust_index_entry_low(
...
-		trx->bulk_insert = true;
+		if (!index->table->has_spatial_index()
+		    && !index->table->n_rec_locks
+		    && !index->table->versioned()) {
 
-		if (!index->table->is_temporary()) {
+			ut_ad(!index->table->skip_alter_undo);
+			trx->bulk_insert= true;
+			if (index->table->is_temporary()) {
+				goto skip_bulk_insert;
```
dr-m, line 2723: It would be good to add a comment to explain why we assign `trx->bulk_insert = true` for temporary tables. By the way, this function is formatted in the archaic InnoDB style, so there should be a space before the `=`.

It would also be less confusing if we renamed the `skip_bulk_insert` label into `row_level_insert`.

If there were no need to assign `trx->bulk_insert = true` for temporary tables, then we should check `!is_temporary()` before `!has_spatial_index()`. I see that the assignment was there also prior to this change.

storage/innobase/row/row0ins.cc
```diff
@@ -2774,6 +2772,29 @@ row_ins_clust_index_entry_low(
...
+			start of the alter operation. Consecutive
+			insert operation are buffered in the
+			bulk buffer and doesn't check for constraint
+			validity of foreign key relationship. */
+			trx_start_if_not_started(trx, true);
+			trx->bulk_insert = true;
+			auto m= trx->mod_tables.emplace(index->table, 0);
+			m.first->second.start_bulk_insert(index->table);
+			err= m.first->second.bulk_insert_buffered(
+						*entry, *index, trx);
```
dr-m, line 2795: Here, we are sometimes omitting a space before `=`, sometimes adding it. In the archaic InnoDB formatting style, which this function is using, there should be a space on both sides of `=`.

Only the archaic InnoDB formatting style allows a line break after `(`. The following line should be indented exactly 8 columns (1 TAB) deeper per unbalanced `(`. Here the line is indented too deep.

storage/innobase/row/row0merge.cc
```diff
@@ -5320,6 +5320,9 @@ dberr_t row_merge_bulk_t::write_to_index(ulint index_no, trx_t *trx)
   else if (index->is_primary() && table->persistent_autoinc)
     btr_write_autoinc(index, table->autoinc - 1);
   err= btr_bulk.finish(err);
+  if (err == DB_SUCCESS && index->is_clust())
+    table->stat_n_rows= (file && file->fd != OS_FILE_CLOSED)
+                          ? file->n_rec : buf.n_tuples;
```
dr-m, line 5325: The `?` should be indented 2 spaces deeper than the previous line.

dr-m (approved):

storage/innobase/row/row0ins.cc
dr-m: Unfortunately, it looks like this comment was never addressed, and we introduced the regression [MDEV-34756](https://jira.mariadb.org/browse/MDEV-34756) as a result.

dr-m: There was a misunderstanding that caused a regression that will be fixed in #3464. I think that from now on, it is better to always explain how a request was resolved. I could also have been more explicit in requesting that the `FOREIGN KEY` enforcement code path be covered in the regression test suite.

PR #2778, opened 2023-10-08
WIP: Sync maintainer scripts etc with latest downstream 10.11.5 in Debian

ottok:
## Description

Sync downstream packaging improvements in Debian to upstream. See commit messages for details.

*When merging, keep all commits separate so their commit history remains legible.*

## How can this PR be tested?

Best tested by verifying there are no regressions on Buildbot nor Salsa-CI.

## Basing the PR against the correct MariaDB version

This is based on 10.11 because source was Debian 10.11. I have not checked what upstream branch is the truly most upstream branch to land these changes on - potentially 10.7 as that is first branch with fuly versionless 'mariadb-server' packages.

## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [no] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #3146, opened 2024-03-22
MDEV-33750: Sync configuratons from Debian Salsa

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-33750*

## Description
Sync configurations and changes from Debian Salsa. This is mostly same as PR #2778 as it stalled and closed but these commits can be ironed to be valuable addition.

## Release Notes
These commits does not affect release notes.

## How can this PR be tested?
There is Salsa-CI run [655744](https://salsa.debian.org/illuusio/mariadb-server/-/pipelines/655744) which will have some failing tests.

This could be test manually in Debian Sid with
`debian/autobake-debs.sh`

and then manually installed.

## PR quality check


debian/mariadb-server.postinst
```diff
@@ -123,12 +123,12 @@ EOF
...
-    [ -f "$mysql_datadir/mysql/user.MYD" ] &&
-    ! lsof -nt "$mysql_datadir"/mysql/user.MYD > /dev/null &&
-    [ ! -f "$mysql_datadir/undo_001" ]
+       [ -f "$mysql_datadir/mysql/user.MYD" ] &&
+       ! lsof -nt "$mysql_datadir"/mysql/user.MYD > /dev/null &&
+       [ ! -f "$mysql_datadir/undo_001" ]
     then
-      echo "UPDATE mysql.user SET plugin='unix_socket' WHERE plugin='auth_socket';" |
-      mariadbd --skip-innodb --key_buffer_size=0  --default-storage-engine=MyISAM --bootstrap 2> /dev/null
+      echo "UPDATE mysql.user SET plugin='unix_socket' WHERE plugin='auth_socket';" | \
```
fauust, line 130: The `\` should not be necessary here (but maybe it's your linter that automatically adds it). More on the linter on general comment.

debian/mariadb-server.mariadb.init
```diff
@@ -156,16 +177,22 @@ case "${1:-''}" in
     # Start MariaDB!
     /usr/bin/mariadbd-safe "${@:2}" 2>&1 >/dev/null | $ERR_LOGGER &
 
-    for i in $(seq 1 "${MYSQLD_STARTUP_TIMEOUT:-30}"); do
+    for _ in $(seq 1 "${MYSQLD_STARTUP_TIMEOUT:-30}")
```
fauust, line 179: is `_` var a bash best practice thing? Just curious.
Also, why not removing `seq` as you did on line 224.

fauust (approved):
Overall, this looks good. It's very good that those bash scripts are verified with shellcheck and that you commented when and why you decided to skip some rules.

Some remarks though. For the reviewers, it's easier to have specific 'lint' commits, I think it would be good to have them in the future so we can easily separate cosmetic changes from functional ones.

On the linter aspect, I may be wrong but we do not have bash syntax guidelines and we probably should (ping @LinuxJedi). I personally try to follow https://google.github.io/styleguide/shellguide.html and use https://github.com/mvdan/sh with `-d -i 2 -ci`.

For instance, I personally prefer `if; then` or `for; do` on the same [line](https://google.github.io/styleguide/shellguide.html#s5.4-loops), it makes script smaller and do not impact readability. But that's OK as it is now if we decide that it should be how we write bash loops.

fauust:
@illuusio, @grooverdan why is the CI not running for this PR? Since this PR impact starting and probably upgrading MariaDB on Debian/Ubuntu, you should make sure that install and upgrade (minor and major) tests pass on BB.

debian/additions/debian-start.inc.sh
```diff
@@ -21,26 +21,38 @@ function check_for_crashed_tables() {
...
-
-  LC_ALL=C $MARIADB --skip-column-names --batch -e  '
-      select concat('\''select count(*) into @discard from `'\'',
-                    TABLE_SCHEMA, '\''`.`'\'', TABLE_NAME, '\''`'\'')
-      from information_schema.TABLES where TABLE_SCHEMA<>'\''INFORMATION_SCHEMA'\'' and TABLE_SCHEMA<>'\''PERFORMANCE_SCHEMA'\'' and ( ENGINE='\''MyISAM'\'' or ENGINE='\''Aria'\'' )' | \
-    xargs -i ${MARIADB} --skip-column-names --silent --batch \
-                    --force -e "{}" &>"${tempfile}"
+  # The $MARIADB is intentionally used to expand into a command and arguments
+  # shellcheck disable=SC2086
+  LC_ALL=C echo '
```
grooverdan, line 35: At the risk of being too pedantic - beginning of this function says "trigging myisam-recover" and it doesn't. So minimally a better comment there. Somehow a basic open of the table is sufficient - was there a decision recorded for this (its a big search - I couldn't find it)? [check table {} FAST](https://mariadb.com/kb/en/check-table/) could have been used as SQL.

Also (for later), looking for filesystem based files might be quicker than information_schema. And Aria might auto-recover (especially after 10.6).

`LC_ALL=C` is now on echo (which serves no purpose) rather than `$MARIADB` which only effects output into the email. So I'm not sure there's a purpose in including it at all.

SC2086 is due to `$MARIADB` being unquoted (twice now which could be corrected.

`xargs --no-run-if-empty` would be worth saving a needless execution.

Is xargs `-P` desired to perform concurrent

debian/additions/debian-start.inc.sh
```diff
@@ -72,8 +84,9 @@ function check_root_accounts() {
 
   logger -p daemon.info -i -t"$0" "Checking for insecure root accounts."
 
-  ret=$( echo "SELECT count(*) FROM mysql.user WHERE user='root' and password='' and plugin in ('', 'mysql_native_password', 'mysql_old_password');" | $MARIADB --skip-column-names )
-  if [ "$ret" -ne "0" ]; then
+  ret=$(echo "SELECT count(*) FROM mysql.user WHERE user='root' and password='' and plugin in ('', 'mysql_native_password', 'mysql_old_password');" | $MARIADB --skip-column-names)
```
grooverdan, line 87: `and password_expired='N'`

debian/mariadb-server.mariadb.init
```diff
@@ -156,19 +176,41 @@ case "${1:-''}" in
...
       fi
     else
+      # Try one more time but save error log separately, then spit it out
+      # before logging ends and init script execution ends.
+      if pgrep -ax mariadbd > /dev/null
+      then
+        echo "ERROR: The mariadbd process is running but not responding:"
+        # shellcheck disable=SC2009
+        # Show the mariadbd process and it's parent and next line (if there is a child process)
+        ps faxu | grep mariadbd -C 1
```
grooverdan, line 205: `ps -C mariadbd -o ppid` or something?

debian/mariadb-server.mariadb.init
```diff
@@ -156,19 +176,41 @@ case "${1:-''}" in
...
+      else
+        ERROR_LOG_FILE="$(mktemp).err"
+        echo # ensure newline
+        timeout --kill-after=20 10 /usr/bin/mysqld_safe "${@:2}" --log-error="$ERROR_LOG_FILE"
```
grooverdan, line 209: `mariadbd-safe`

debian/mariadb-server.postinst
```diff
@@ -183,6 +183,8 @@ EOF
     # Clean up old flags before setting new one
     rm -f $mysql_datadir/debian-*.flag
     # Flag data dir to avoid downgrades
+    # @TODO: Rewrite this to use the new upstream /var/lib/mysql_upgrade_info file
```
grooverdan, line 186: `mariadb_upgrade_info` from any version that might fix this.

debian/mariadb-server.postinst
```diff
@@ -191,9 +193,13 @@ EOF
     # handle things.
     # Debian: beware of the bashisms...
     # Debian: can safely run on upgrades with existing databases
+    # Workaround for Debian Bug #1022994: failure to create database when
+    # working with libpam-tmpdir (by setting TMPDIR to empty value).
```
grooverdan, line 197: there's a mariadb bug for this. A long as its referenced in commit message, or otherwise closed with this merge its ok.

debian/rules
```diff
@@ -19,16 +19,19 @@ CFLAGS+=$(CPPFLAGS)
...
+    export DPKG_GENSYMBOLS_CHECK_LEVEL = 4
 endif
 
 BUILDDIR := builddir
 DEB_VERSION_REVISION := $(shell echo $(DEB_VERSION) | sed -e 's/^.*-//')
 DEB_VERSION_VERSION := $(shell echo $(DEB_VERSION) | sed -e 's/^.*:\(.*\)\(-\|+\).*/\1/')
 DEB_VERSION_MAJOR := $(shell echo $(DEB_VERSION_VERSION) | sed -e 's/^\(.*\)\..*$$/\1/')
 RELEASE := $(shell lsb_release -r -s) # Use changelog based DEB_DISTRIBUTION instead?
-TMP:=$(CURDIR)/debian/tmp
+TMP := $(CURDIR)/debian/tmp
```
grooverdan, line 33: Should TMP be an export? Its not used by rules (aka this makefile)?

debian/rules
```diff
@@ -72,14 +80,17 @@ override_dh_auto_configure:
...
 endif
 
 	echo "server:Version=$(DEB_VERSION)" >> debian/substvars
 
-	# As packages does not have major version any more  on package name there is no way as it not set by dpkg
-	# to use this on postinst script. Use sed to determine major version
+	# As packages does not have major version any more in package name there is no
+	# way as it not set by dpkg to use this on postinst script. Use sed to
+	# determine major version instead.
+	# @TODO: Rewrite this to use the new upstream /var/lib/mysql_upgrade_info file
```
grooverdan, line 92: `mariadb_upgrade_info`.

debian/mariadb-server.postrm
```diff
@@ -14,13 +14,22 @@ ${DEBIAN_SCRIPT_TRACE:+ echo "#42#DEBUG# RUNNING $0 $*" 1>&2 }
 
 MYADMIN="/usr/bin/mysqladmin --defaults-file=/etc/mysql/debian.cnf"
 
+MARIADBD_USERS="root"
```
grooverdan, line 17: removing file like - https://salsa.debian.org/mariadb-team/mariadb-server/-/merge_requests/75 ?

debian/tests/upstream
```diff
@@ -24,48 +24,48 @@ echo "using tmpdir: $WORKDIR/tmp"
...
 fi
 
 if [ "$ARCH" = "s390x" ]
 then
-  echo "main.func_regexp_pcre : recursion fails on s390x https://bugs.launchpad.net/ubuntu/+source/mariadb-10.1/+bug/1723947" >> $SKIP_TEST_LST
+  echo "main.func_regexp_pcre : recursion fails on s390x https://bugs.launchpad.net/ubuntu/+source/mariadb-10.1/+bug/1723947" >> "$MTR_SKIP_TEST_LIST"
 elif [ "$ARCH" = "armhf" ] || [ "$ARCH" = "i386" ]
 then
-  echo "main.failed_auth_unixsocket : Test returns wrong exit code on armhf and i386 (but only in debci) https://jira.mariadb.org/browse/MDEV-23933" >> $SKIP_TEST_LST
+  echo "main.failed_auth_unixsocket : Test returns wrong exit code on armhf and i386 (but only in debci) https://jira.mariadb.org/browse/MDEV-23933" >> "$MTR_SKIP_TEST_LIST"
```
grooverdan, line 53: [MDEV-23933](https://jira.mariadb.org/browse/MDEV-23933) is fixed.

debian/tests/upstream
```diff
@@ -24,48 +24,48 @@ echo "using tmpdir: $WORKDIR/tmp"
...
+eatmydata perl -I. ./mysql-test-run.pl \
+    --force --testcase-timeout=120 --suite-timeout=540 --retry=3 \
+    --verbose-restart --max-save-core=1 --max-save-datadir=1 \
+    --parallel=auto --skip-rpl --suite=main \
+    --skip-test-list="$MTR_SKIP_TEST_LIST" \
     --vardir="$WORKDIR/var" --tmpdir="$WORKDIR/tmp" \
-    --parallel=auto --skip-rpl \
-    --force --skip-test-list=$SKIP_TEST_LST \
-    --xml-report=$AUTOPKGTEST_ARTIFACTS/mysql-test-run-junit.xml $@ 2>&1
+    --xml-report="$AUTOPKGTEST_ARTIFACTS/mysql-test-run-junit.xml"
```
grooverdan, line 69: for fun, lets call it `mariadb-test-run-junit.xml`.

debian/mariadb-server.mariadb.init
```diff
@@ -19,9 +19,10 @@ ${DEBIAN_SCRIPT_DEBUG:+ set -v -x}
 
 test -x /usr/sbin/mariadbd || exit 0
 
+# shellcheck source=/dev/null
 . /lib/lsb/init-functions
 
-SELF=$(cd "$(dirname $0)"; pwd -P)/$(basename $0)
+SELF="$(cd "$(dirname "$0")"; pwd -P)/$(basename "$0")"
```
tonychen2001, line 25: This quoting here seems fishy.

The inner quotes should be escaped. For e.g.:
```
"$(cd \"$(dirname \"$0\")\"; pwd -P)/$(basename \"$0\")"
```

fauust: No this is not necessary, but the correct thing to do here would be to use parameter expansion since we are using bash:

Something like the following (not tested):
```bash
SELF="$(cd "${0%/*}"; pwd -P)/${0##*/}"
```

debian/mariadb-server.postinst
illuusio: Both seems to work with BASH the same and `shfmt` say there should not be and `beautysh` does not case neither `shellcheck` so I think I'll remove them.

debian/mariadb-server.mariadb.init
illuusio: For constituency I think as new one also have '_' 
Reason is that new syntax {1..30} does not allow to use variable.. although it could be half minute without variable possibility.

debian/mariadb-server.mariadb.init
illuusio: I'm not exactly sure what is intend to achieve in this one. Would `pgrep` give wanted as it gives parent pid and child pid?

debian/mariadb-server.mariadb.init
illuusio: Ok with this one.. there is also file: `debian/additions/mariadb.conf.d/50-mysqld_safe.cnf` which is also installed..

debian/mariadb-server.postinst
illuusio: Can you point JIRA bug so this as I can't find it

debian/mariadb-server.postrm
illuusio: There is now PR #3210 which imports needed changes from Debian Salsa and reworks them to MariaDB upstream repo. It should get in first and then this should be adapted to changes.

debian/tests/upstream
illuusio: So this can be removed altogether?

debian/mariadb-server.mariadb.init
illuusio: Ok I've tested both:
```
echo "${0%/*}"
echo "$(dirname "${0}")"
echo "$(dirname \"${0}\")"
```
An then run it with `bash test.sh` which gives:
```
test.sh
.
.
```
So @fauust approach is correct when there is directory name but fails when run in same dir. Also original works as expected. This is very difficult way to find `realpath`-command?

debian/mariadb-server.postinst
grooverdan: https://jira.mariadb.org/browse/MDEV-29910

debian/mariadb-server.mariadb.init
grooverdan: it assumes the subprocess it somehow next. werid. sysint stuff so I don't particularly care.

debian/additions/debian-start.inc.sh
grooverdan: reminder ^

debian/mariadb-server.mariadb.init
grooverdan: exe not changed yet.

debian/tests/upstream
grooverdan: yes

debian/mariadb-server.mariadb.init
illuusio: Please PR #3240 for this as it should take care of these ones

debian/rules
illuusio: Probably.. It does not hurt though.

debian/mariadb-server.postinst
ottok: Right, both seem to work:

```
± cat test.sh 
#!/bin/bash
set -e

echo test-with-backslash | \
 cat

echo test-no-backslash |
 cat

± ./test.sh 
test-with-backslash
test-no-backslash
```

debian/mariadb-server.mariadb.init
ottok: > is `_` var a bash best practice thing? Just curious. Also, why not removing `seq` as you did on line 224.

Yes: https://www.shellcheck.net/wiki/SC2034#intentionally-unused-variables

debian/rules
ottok: > Its not used by rules (aka this makefile)?

Yes it is, `$(TMP)` appears 17 times in `debian/rules`.

debian/mariadb-server.mariadb.init
```diff
@@ -176,8 +176,21 @@ case "${1:-''}" in
     # Start MariaDB!
     /usr/bin/mariadbd-safe "${@:2}" 2>&1 >/dev/null | $ERR_LOGGER &
 
-    for _ in $(seq 1 "${MYSQLD_STARTUP_TIMEOUT:-30}")
```
ottok: Why not use [seq](https://manpages.debian.org/unstable/coreutils/seq.1.en.html), why need to stop using it?

ottok:
:+1 Unsurprisingly this PR looks good to me since it is upstreaming changes done in Debian in e.g.
* https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/a3b6f3d7f4b88074c76fe67d909977e74d8ba7a5
* https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/b935704063506b1581b297a80cdfe00deda65640
* https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/f2fe8d8300642c373ce325f19d1100d79adeed45
* https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/92c1748770506c9364fe67b8a3fbc1221d71653b
* https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/9787b43c9601a3e1e497d22140de13122277d5c3

debian/additions/debian-start.inc.sh
grooverdan: Checked with Monty - Aria tables will auto recover. MyISAM won't.
Option 1:
* remove all between set +e / -e as old stuff assuming we never need to handle crashed upgrade from < 10.4, maybe users can handle this if it does happen.
Option 2:
execute as a single SQL string:
```
select concat('check table ', group_concat(CONCAT('`', TABLE_SCHEMA, '`.`', TABLE_NAME, '`')), ' fast') into @check FROM information_schema.TABLES WHERE TABLE_SCHEMA<>"INFORMATION_SCHEMA" AND TABLE_SCHEMA<>'PERFORMANCE_SCHEMA'  AND ENGINE='MyISAM' OR (TABLE_SCHEMA='mysql' and TABLE_NAME='time_zone_leap_second'); execute immediate @check;
``` 

rational:
* check table fast - doesn't need to fully read the entire table (unsure if information_schema will open table)
* MyISAM only is needed
* mysql.time_zone_leap_second include to ensure a single table is there.
* quoting - in case of some table names, isn't fully coverage

debian/mariadb-server.mariadb.init
ottok: Why not use [seq](https://manpages.debian.org/unstable/coreutils/seq.1.en.html), why need to stop using it?

debian/mariadb-server.mariadb.init
dr-m: Wouldn’t it be more efficient or idiomatic to write
```sh
seq 1 "${MYSQLD_STARTUP_TIMEOUT:-30}"|while read
do
```
This would assign each read line to the implicit value `$REPLY` if we cared about the contents.

debian/mariadb-server.mariadb.init
illuusio: I've also added them here as mentioned

grooverdan (approved):

PR #3147, opened 2024-03-22
MDEV-31046: Test is fakeroot is working with /usr/bin/true

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-31046*

## Description
Test if `fakeroot` command with `/usr/bin/true` to see if it's working correctly before trying to start building Debian package.

## Release Notes
Make sure Debian package building environment is working before trying to build actual package

## How can this PR be tested?
This can be tested with:
`debian/autobake-deb.sh`
If it's working as expected then it should work. One can change executed command to non exist one then it should fail building.

## PR quality check

### DIFF
--- a/debian/autobake-deb.sh
+++ b/debian/autobake-deb.sh
@@ -170,7 +170,20 @@ BUILDPACKAGE_DPKGCMD=()
 # Fakeroot test
 if fakeroot true; then
   BUILDPACKAGE_DPKGCMD+=( "fakeroot" "--" )
+
+  echo "Testing if 'fakeroot' command is working with '/usr/bin/true'"
+
+  BUILDPACKAGE_FAKEROOTTEST=("${BUILDPACKAGE_DPKGCMD[@]}")
+  BUILDPACKAGE_FAKEROOTTEST+=("/bin/true")
+  if ! "${BUILDPACKAGE_FAKEROOTTEST[@]}";
+  then
+    echo "Fakeroot not working correctly exiting with error"
+    exit 1
+  fi
 fi
+
+
+
 
 # Use eatmydata is available to build faster with less I/O, skipping fsync()
 # during the entire build process (safe because a build can always be restarted)


debian/autobake-deb.sh
```diff
@@ -170,8 +170,21 @@ BUILDPACKAGE_DPKGCMD=()
 # Fakeroot test
 if fakeroot true; then
   BUILDPACKAGE_DPKGCMD+=( "fakeroot" "--" )
+
+  echo "Testing if 'fakeroot' command is working with '/usr/bin/true'"
```
tonychen2001, line 174: '/usr/bin/true' should be '/bin/true'. Or, adjust line 177 to be '/usr/bin/true'

This log is likely unnecessary. A user does not need to know that the 'fakeroot' command is being tested for existence. Instead, the error output on line 180 could be more specific such as "Failed to execute 'fakeroot -- /bin/true'. Proceeding without 'fakeroot'"

debian/autobake-deb.sh
```diff
@@ -170,8 +170,21 @@ BUILDPACKAGE_DPKGCMD=()
...
+
+  BUILDPACKAGE_FAKEROOTTEST=("${BUILDPACKAGE_DPKGCMD[@]}")
+  BUILDPACKAGE_FAKEROOTTEST+=("/bin/true")
+  if ! "${BUILDPACKAGE_FAKEROOTTEST[@]}";
```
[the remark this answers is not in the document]
illuusio: This was just for the one time that BB was failing but it's not anyhow necessary to know if `fakeroot` working on or not before it's launch with dpkg building. If this is not kind needed I can just close it..

debian/autobake-deb.sh
illuusio: I've update it back to `/bin/true` as it's preset location. Debian is going thru Fedora and others everything in `/usr/bin/` transition so that's why it was `/usr/bin/true`

PR #3148, opened 2024-03-22
MDEV-33515 log_sys.lsn_lock causes excessive context switching

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33515*
## Description
The `log_sys.lsn_lock` is a very contended resource with a small critical section in `log_sys.append_prepare()`. On many processor microarchitectures, replacing the system call based `log_sys.lsn_lock` with a pure spin lock would fare worse during high concurrency workloads, wasting a significant amount of CPU cycles in the spin loop.

On other microarchitectures, we would see a significant amount of time being spent in `native_queued_spin_lock_slowpath()` in the Linux kernel, plus context switching between user and kernel address space. This was pointed out by Steve Shaw from Intel Corporation.

Depending on the workload and the hardware implementation, it may be useful to use a pure spin lock in `log_sys.append_prepare()`. We will introduce a parameter. The statement
```sql
SET GLOBAL INNODB_LOG_SPIN_WAIT_DELAY=50;
```
would enable a spin lock that will execute that many `MY_RELAX_CPU()` operations (such as the x86 `PAUSE` instruction) between successive attempts of acquiring the spin lock. The use of a system call based `log_sys.lsn_lock` (which is the default setting) can be enabled by
```sql
SET GLOBAL INNODB_LOG_SPIN_WAIT_DELAY=0;
```
This patch will also introduce `#ifdef LOG_LATCH_DEBUG` (part of `cmake -DWITH_INNODB_EXTRA_DEBUG=ON`) for more accurate tracking of `log_sys.latch` ownership and reorganize the fields of `log_sys` to improve the locality of reference and to reduce the chances of false sharing.

When a spin lock is being used, it will be maintained in the most significant bit of `log_sys.buf_free`. This is useful, because that is one of the fields that is covered by the lock. For IA-32 or AMD64, we implement the spin lock specially via `log_t::lsn_lock_bts()`, employing the i386 `LOCK BTS` instruction. A straightforward `std::atomic::fetch_or()` would translate into an inefficient loop around `LOCK CMPXCHG`.

`mtr_t::spin_wait_delay`: The value of `innodb_log_spin_wait_delay`.

`mtr_t::finisher`: Pointer to the currently used `mtr_t::finish_write()` implementation. This allows to avoid introducing conditional branches. We no longer invoke `log_sys.is_pmem()` at the mini-transaction level, but we would do that in `log_write_up_to()`.

`mtr_t::finisher_update()`: Update finisher when spin_wait_delay is changed from or to 0 (the spin lock is changed to `log_sys.lsn_lock` or vice versa).

## Release Notes
The settable global parameter `INNODB_LOG_SPIN_WAIT_DELAY` may be set to a nonzero value to replace the use of a `log_sys.lsn_lock` with a pure spin lock (no system calls). The number specifies the number of `MY_RELAX_CPU()` operations (such as the x86 `PAUSE` instruction) between attempts to acquire the spin lock. Using this setting may greatly degrade or improve performance. On Intel Xeon processors using the Emerald Rapids microarchitecture, the value 50 could be useful. The default value 0 should work best on most systems.

## How can this PR be tested?
```sh
./mtr
./mtr --mysqld=--loose-innodb-log-spin-wait-delay=10
```
I have tested this on both "fake PMEM" (`/dev/shm` when `libpmem` has been linked in) and on system call based log writes.

In Sysbench workloads on my Haswell microarchitecture dual Xeon E5-2630 v4, the default value (`log_sys.lsn_lock` backed by `futex`) performs significantly better than enabling the spin locks.

## PR quality check


storage/innobase/mtr/mtr0mtr.cc
```diff
@@ -37,6 +37,24 @@ Created 11/26/1995 Heikki Tuuri
...
 #include "mariadb_stats.h"
+#include "my_cpu.h"
+
+std::pair<lsn_t,mtr_t::page_flush_ahead> (*mtr_t::finisher)(mtr_t *, size_t);
+unsigned mtr_t::spin_wait_delay;
+
+void mtr_t::finisher_update()
+{
+  ut_ad(log_sys.latch_have_wr());
+  finisher=
```
vaintroub, line 48: this is a bit terse because of excessive use of ternary operator.

vaintroub:
I also confirm that non-zero INNODB_LOG_SPIN_WAIT_DELAY  performs far worse, when benchmarks are run on in different settings,  "12th Gen Intel Core i9-12900K" , aka Alder Lake,  server running on P-cores, on Windows.

This is consistent with observation that was made a couple of years ago, when spinlocks were tried for the first time.
But we have to allow people some experimentation. So, OK from me.

vaintroub (approved):
Ok to push. 

Hopefully, at some point we get access to hardware that actually benefits from spinning, otherwise it is a little uneasy to add functionality where we do not see benefit.

storage/innobase/mtr/mtr0mtr.cc
dr-m: Sorry, I am a fan of the ternary operator. At least I put some parentheses around it to make it a little clearer.

PR #3149, opened 2024-03-22
MDEV-33509 Failed to apply write set with flags=(rollback|pa_unsafe)

sciascid:
- [x] *The Jira issue number for this PR is: MDEV-33509*

## Description
Fix function `remove_fragment()` in wsrep_schema so that no error is raised if the fragment to be removed is not found in the wsrep_streaming_log table. This is necessary to handle the case where streaming transaction in idle state is BF aborted. This may result in the case where the rollbacker thread successfully removes the transaction's fragments, followed by the applier's attempt to remove the same fragments. Causing the node to leave the cluster after reporting a "Failed to apply write set" error.

## Release Notes
None

## How can this PR be tested?
A MTR test is provided in with the pull request.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera_sr/r/galera_sr_bf_abort_idle.result
+++ b/mysql-test/suite/galera_sr/r/galera_sr_bf_abort_idle.result
@@ -1,0 +1,33 @@
+connection node_2;
+connection node_1;
+connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1;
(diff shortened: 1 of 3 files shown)


sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/e0c8165487ecfbee1d25625422a91a6352d7e522

PR #3152, opened 2024-03-25
MDEV-33757 Get rid of TrxUndoRsegs code

vlad-lesin:
TrxUndoRsegs is wrapper for vector of trx_rseg_t*. It has two constructors, both initialize the vector with only one element. And they are used to push transactions rseg(the singular) to purge queue. There is no function to add elements to the vector. The default constructor is used only for declaration of NullElement.

The TrxUndoRsegs was introduced in WL#6915 in MySQL 5.7 and. MySQL 5.7 would unnecessarily let the purge of history parse the temporary undo records, and then look up the table (via a global hash table), and only at the point of processing the parsed undo log record determine that the table is a temporary table and the undo record must be thrown away.

In MariaDB 10.2 we have two disjoint sets of rollback segments (128 for persistent, 128 for temporary), and purge does not even see the temporary tables. The only reason why temporary tables are visible to other threads is a SQL layer bug (MDEV-17805).

purge_sys_t::choose_next_log(): merge the relevant part of TrxUndoRsegsIterator::set_next() to the start of purge_sys_t::choose_next_log().

purge_sys_t::rseg_get_next_history_log(): add a tail call of purge_sys_t::choose_next_log() and adjust the callers, to simplify the control flow further.

purge_sys.pq_mutex and purge_sys.purge_queue: make it private by adding some simple accessor function.

trx_purge_cleanse_purge_queue(): make it a member of purge_sys_t to have have access to private purge_sys.pq_mutex and purge_sys.purge_queue.

Thanks Marko Mäkelä for historical overview of TrxUndoRsegs development.

Reviewed by: Marko Mäkelä

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/include/trx0purge.h
+++ b/storage/innobase/include/trx0purge.h
@@ -55,80 +55,79 @@ Run a purge batch.
 @return number of undo log pages handled in the batch */
 ulint trx_purge(ulint n_tasks, ulint history_size);
 
-/** Rollback segements from a given transaction with trx-no
-scheduled for purge. */
-class TrxUndoRsegs {
-private:
-	typedef std::vector<trx_rseg_t*, ut_allocator<trx_rseg_t*> >
-		trx_rsegs_t;
-public:
-	typedef trx_rsegs_t::iterator iterator;
-	typedef trx_rsegs_t::const_iterator const_iterator;
-
-	TrxUndoRsegs() = default;
-
-	/** Constructor */
-	TrxUndoRsegs(trx_rseg_t& rseg)
-		: trx_no(rseg.last_trx_no()), m_rsegs(1, &rseg) {}
-	/** Constructor */
-	TrxUndoRsegs(trx_id_t trx_no, trx_rseg_t& rseg)
-		: trx_no(trx_no), m_rsegs(1, &rseg) {}
-
-	bool operator!=(const TrxUndoRsegs& other) const
-	{ return trx_no != other.trx_no; }
-	bool empty() const { return m_rsegs.empty(); }
-	void erase(iterator& it) { m_rsegs.erase(it); }
-	iterator begin() { return(m_rsegs.begin()); }
-	iterator end() { return(m_rsegs.end()); }
-	const_iterator begin() const { return m_rsegs.begin(); }
-	const_iterator end() const { return m_rsegs.end(); }
-
-	/** Compare two TrxUndoRsegs based on trx_no.
-	@param elem1 first element to compare
-	@param elem2 second element to compare
-	@return true if elem1 > elem2 else false.*/
-	bool operator()(const TrxUndoRsegs& lhs, const TrxUndoRsegs& rhs)
-	{
-		return(lhs.trx_no > rhs.trx_no);
-	}
-
-	/** Copy of trx_rseg_t::last_trx_no() */
-	trx_id_t trx_no= 0;
-private:
-	/** Rollback segments of a transaction, scheduled for purge. */
-	trx_rsegs_t m_rsegs{};
-};
-
-typedef std::priority_queue<
-	TrxUndoRsegs,
-	std::vector<TrxUndoRsegs, ut_allocator<TrxUndoRsegs> >,
-	TrxUndoRsegs>	purge_pq_t;
-
-/** Chooses the rollback segment with the oldest committed transaction */
-struct TrxUndoRsegsIterator {
-	/** Constructor */
-	TrxUndoRsegsIterator();
-	/** Sets the next rseg to purge in purge_sys.
-	Executed in the purge coordinator thread.
-	@retval false when nothing is to be purged
-	@retval true  when purge_sys.rseg->latch was locked */
-	inline bool set_next();
-
-private:
-	// Disable copying
-	TrxUndoRsegsIterator(const TrxUndoRsegsIterator&);
-	TrxUndoRsegsIterator& operator=(const TrxUndoRsegsIterator&);
-
-	/** The current element to process */
-	TrxUndoRsegs			m_rsegs;
-	/** Track the current element in m_rsegs */
-	TrxUndoRsegs::const_iterator	m_iter;
-};
-
 /** The control structure used in the purge operation */
 class purge_sys_t
 {
-  friend TrxUndoRsegsIterator;
+  /** Min-heap based priority queue over fixed size array */
+  class purge_queue
+  {
+    /** Array of indexes in trx_sys.rseg_array. */
+    alignas(CPU_LEVEL1_DCACHE_LINESIZE) byte m_array[TRX_SYS_N_RSEGS];
+    /** Pointer to the end of m_array. */
+    byte *m_end= m_array;
+
+  public:
+    struct trx_rseg_cmp
+    {
+      /** Compare two trx_rseg_t* based on trx_no.
+      @param lhs first index in trx_sys.rseg_array to compare
+      @param rhs second index in trx_sys.rseg_array to compare
+      @return whether lhs>rhs */
+      bool operator()(const byte lhs, const byte rhs)
+      {
+        ut_ad(lhs < TRX_SYS_N_RSEGS);
+        ut_ad(rhs < TRX_SYS_N_RSEGS);
+        /* We can compare without trx_rseg_t::latch, because rseg last
+        commit is always set before pushing rseg to purge queue. */
+        return trx_sys.rseg_array[lhs].last_commit_and_offset >
+               trx_sys.rseg_array[rhs].last_commit_and_offset;
+      }
+    };
+    byte *begin() { return m_array; }
+    byte *end() { return m_end; }
+    const byte *c_begin() const { return m_array; }
+    const byte *c_end() const { return m_end; }
+    size_t size() const
+    {
+      size_t s= c_end() - c_begin();
+      ut_ad(s <= TRX_SYS_N_RSEGS);
+      return s;
+    }
+    bool empty() const { return !size(); }
+    void clear() { m_end= m_array; }
+
+    /** Push index of trx_sys.rseg_array into min-heap.
+    @param i index to push */
+    void push_rseg_index(byte i)
+    {
+      ut_ad(i < TRX_SYS_N_RSEGS);
+      ut_ad(size() + 1 <= TRX_SYS_N_RSEGS);
+      *m_end++= i;
+      std::push_heap(begin(), end(), trx_rseg_cmp());
+    }
+
+    /** Push rseg to priority queue.
+    @param rseg trx_rseg_t pointer to push */
+    void push(const trx_rseg_t *rseg)
+    {
+      ut_ad(rseg >= trx_sys.rseg_array);
+      ut_ad(rseg < trx_sys.rseg_array + TRX_SYS_N_RSEGS);
+      byte i= byte(rseg - trx_sys.rseg_array);
+      push_rseg_index(i);
+    }
+
+    /** Pop rseg from priority queue.
+    @return pointer to popped trx_rseg_t object */
+    trx_rseg_t *pop()
+    {
+      ut_ad(!empty());
+      std::pop_heap(begin(), end(), trx_rseg_cmp());
+      byte i= *--m_end;
+      ut_ad(i < TRX_SYS_N_RSEGS);
+      return &trx_sys.rseg_array[i];
+    }
+  };
+
 public:
   /** latch protecting view, m_enabled */
   alignas(CPU_LEVEL1_DCACHE_LINESIZE) mutable srw_spin_lock latch;
@@ -244,15 +243,27 @@ private:
 					record */
 	uint16_t	hdr_offset;	/*!< Header byte offset on the page */
 
+  /** Binary min-heap of indexes in trx_sys.rseg_array, ordered on
+  rseg_t::last_trx_no(). It is protected by the pq_mutex */
+  purge_queue purge_queue;
 
-	TrxUndoRsegsIterator
-			rseg_iter;	/*!< Iterator to get the next rseg
-					to process */
+  /** Mutex protecting purge_queue */
+  mysql_mutex_t pq_mutex;
+
 public:
-	purge_pq_t	purge_queue;	/*!< Binary min-heap, ordered on
-					TrxUndoRsegs::trx_no. It is protected
-					by the pq_mutex */
-	mysql_mutex_t	pq_mutex;	/*!< Mutex protecting purge_queue */
+  /** Push to purge queue without acquiring pq_mutex.
+  @param rseg rseg to push */
+  void enqueue(trx_rseg_t &rseg)
+  {
+    mysql_mutex_assert_owner(&pq_mutex);
+    purge_queue.push(&rseg);
+  }
+
+  /** Acquare purge_queue_mutex */
+  void queue_lock() { mysql_mutex_lock(&pq_mutex); }
+
+  /** Release purge queue mutex */
+  void queue_unlock() { mysql_mutex_unlock(&pq_mutex); }
 
   /** innodb_undo_log_truncate=ON state;
   only modified by purge_coordinator_callback() */
@@ -332,8 +343,9 @@ private:
 
   /** Update the last not yet purged history log info in rseg when
   we have purged a whole undo log. Advances also purge_trx_no
-  past the purged log. */
-  void rseg_get_next_history_log();
+  past the purged log.
+  @return whether anything is to be purged */
+  bool rseg_get_next_history_log();
 
 public:
   /**
@@ -438,6 +450,11 @@ public:
   @param	already_stopped	True indicates purge threads were
 				already stopped */
   void stop_FTS(const dict_table_t &table, bool already_stopped=false);
+
+  /** Cleanse purge queue to remove the rseg that reside in undo-tablespace
+  marked for truncate.
+  @param space undo tablespace being truncated */
+  void cleanse_purge_queue(const fil_space_t &space);
 };
 
 /** The global data structure coordinating a purge */
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -56,84 +56,6 @@ purge_sys_t	purge_sys;
 my_bool		srv_purge_view_update_only_debug;
 #endif /* UNIV_DEBUG */
 
-/** Sentinel value */
-static const TrxUndoRsegs NullElement;
-
-/** Default constructor */
-TrxUndoRsegsIterator::TrxUndoRsegsIterator()
-	: m_rsegs(NullElement), m_iter(m_rsegs.begin())
-{
-}
-
-/** Sets the next rseg to purge in purge_sys.
-Executed in the purge coordinator thread.
-@retval false when nothing is to be purged
-@retval true  when purge_sys.rseg->latch was locked */
-inline bool TrxUndoRsegsIterator::set_next()
-{
-	ut_ad(!purge_sys.next_stored);
-	mysql_mutex_lock(&purge_sys.pq_mutex);
-
-	/* Only purge consumes events from the priority queue, user
-	threads only produce the events. */
-
-	/* Check if there are more rsegs to process in the
-	current element. */
-	if (m_iter != m_rsegs.end()) {
-		/* We are still processing rollback segment from
-		the same transaction and so expected transaction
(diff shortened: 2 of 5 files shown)


dlenski:
If I'm not mistaken, @vlad-lesin, you work for @mariadb-corporation.

However, the commit in this PR isn't using your @mariadb.com email in the author field. I believe that's why the CLAassistant is showing it as CLA-not-signed. And should probably be corrected to clarify the provenance of the PR.

storage/innobase/include/trx0purge.h
```diff
@@ -55,80 +55,28 @@ Run a purge batch.
...
-	TrxUndoRsegsIterator& operator=(const TrxUndoRsegsIterator&);
-
-	/** The current element to process */
-	TrxUndoRsegs			m_rsegs;
-	/** Track the current element in m_rsegs */
-	TrxUndoRsegs::const_iterator	m_iter;
-};
+    trx_rseg_t *, std::vector<trx_rseg_t *, ut_allocator<trx_rseg_t *>>,
+    trx_rseg_cmp>
+    purge_pq_t;
```
dr-m, line 75: These will point to `trx_sys.rseg_array`, right? What if we make the underlying array `byte[TRX_SYS_N_RSEGS]` and allocate it statically? If `std::priority_queue` needs to maintain the number of elements, we won’t be able to use `std::array` but we could implement something similar on our own. Here is some code from another C++ program that I wrote:
```c++
template <typename T, size_t N>
class vararray
{
public:
  constexpr auto begin() const { return m_arr; }
  constexpr auto begin() { return m_arr; }
  constexpr auto end() const { return begin () + size (); }
  constexpr auto end() { return begin () + size (); }
  constexpr auto operator[](size_t s) const { return m_arr[s]; }
  constexpr void push_back(const T& piece) { m_arr[m_size++]= piece; }
  constexpr size_t size() const { return m_size; }

private:
  T m_arr[N];
  size_t m_size= 0;
};
```
The `trx_rseg_cmp` would then be something like this:
```c++
struct trx_rseg_cmp
{
  /** Compare two trx_rseg_t* based on trx_no.
  @param lhs first pointer to trx_rseg_t to compare
  @param rhs second pointer to trx_rseg_t to compare
  @return whether lhs>rhs */
  bool operator()(const byte lhs, const byte rhs)
  {
    /* We can compare without trx_rseg_t::latch, because rseg last
    commit is always set before pushing rseg to purge queue. */
    return trx_sys.rseg_array[lhs].last_trx_no() >
      trx_sys.rseg_array[rhs].last_trx_no();
  }
};
```

storage/innobase/include/trx0purge.h
```diff
@@ -244,16 +192,25 @@ class purge_sys_t
...
 					by the pq_mutex */
 	mysql_mutex_t	pq_mutex;	/*!< Mutex protecting purge_queue */
 
+public:
+  /* Push to purge queue without acquiring pq_mutex.
+  @param rseg rseg to push */
+  void push_to_purge_queue_no_lock(trx_rseg_t &rseg)
+  {
+    purge_queue.push(&rseg);
+  }
```
dr-m, line 206: I think that it would be better to just have one accessor and invoke `mysql_mutex_assert_owner(&pq_mutex)` inside it.

storage/innobase/include/trx0purge.h
```diff
@@ -438,6 +396,11 @@ class purge_sys_t
   @param	already_stopped	True indicates purge threads were
 				already stopped */
   void stop_FTS(const dict_table_t &table, bool already_stopped=false);
+
+  /** Cleanse purge queue to remove the rseg that reside in undo-tablespace
+  marked for truncate.
+  @param[in] space undo tablespace being truncated */
+  void cleanse_purge_queue(const fil_space_t& space);
```
dr-m, line 403: Let’s use plain `@param` in new code. A reference to `const` can only be an input parameter.

storage/innobase/trx/trx0purge.cc
```diff
@@ -571,42 +493,26 @@ inline dberr_t purge_sys_t::iterator::free_history_rseg(trx_rseg_t &rseg) const
...
-				break;
-			}
-		}
+  /* Remove rseg instances that are in the purge queue before we start
+  truncate of corresponding UNDO truncate. */
+  while (!purge_queue.empty())
+  {
+    purge_elem_list.push_back(purge_sys.purge_queue.top());
+    purge_sys.purge_queue.pop();
+  }
```
dr-m, line 509: This logic can be simplified if we [implemented our own priority queue](https://www.geeksforgeeks.org/how-to-implement-priority-queue-using-heap-or-array/) on top of an array.

Note: b3c9cc6f212c650a18ca191a8889e2af68ffc9d6 or mysql/mysql-server@9d6d1902e091c868bb288e0ccf9f975ccb474db9 had added a "binary heap"; the header `ut0bh.h` that those commits refer to was apparently added later. As far as I understand, it is based on an array. The interesting part are the function `ib_bh_push()` and `ib_bh_pop()`. https://www.geeksforgeeks.org/priority-queue-using-binary-heap/ describes this technique.

This loop could access the underlying elements (indexes into `trx_sys.rseg_array`) directly, and then set the number of elements to 0. That would avoid restructuring the heap in the equivalent of `ib_bh_pop()`.

I realize that we do have to construct something like `purge_elem_list` and can’t easily modify `purge_queue` (this member function should not refer to `purge_sys`!) in place. However, the intermediate copy could be a simple `memcpy()` of the underlying array of up to 128 bytes. We could also filter for `&space` already during this first copying loop, and then just insert the surviving copied elements back to `purge_queue`.

dr-m:
> If I'm not mistaken, @vlad-lesin, you work for @mariadb-corporation.
> 
> However, the commit in this PR isn't using your @mariadb.com email in the author field. I believe that's why the CLAassistant is showing it as CLA-not-signed. And should probably be corrected to clarify the provenance of the PR.

@dlenski, I do not think that it is related to the email address. I use a mariadb.com email address in my commits and I work for MariaDB plc, but I have not signed the CLA. Instead, I have signed something equivalent, and I think @vlad-lesin has as well. Until some months ago, pull requests were only used by external contributors. The failing CLA check does not prevent this PR from being merged.

storage/innobase/include/trx0purge.h
vlad-lesin: I don't fully understand your idea. We need to acquire purge queue mutex outside of `purge_sys_t` in `trx_t::write_serialisation_history()`. That's why we need to separate the accessor and purge queue locking. If you propose just to add `mysql_mutex_assert_owner(&pq_mutex)` in `push_to_purge_queue_no_lock()`, then I am ok with this, but we need to manually lock/unlock the mutex in `trx_rseg_mem_restore()`, what, of course will not impact performance too much, as `trx_rseg_mem_restore()` is invoked only on server start. We could even add template argument in `push_to_purge_queue_no_lock()` function.

dr-m: Yes, that exactly is my idea. I would rename `push_to_purge_queue_no_lock()` to just `enqueue()` and omit the variant that acquires and releases the mutex. To reduce the mutex traffic on startup, we could acquire `purge_sys.pq_mutex` at the start of `trx_rseg_array_init()` and release it at the end. That should be fine from the deadlock prevention point of view, because executing that function is a prerequisite for starting the purge subsystem or any transactions. In other words, `trx_rseg_array_init()` can’t be blocked by any thread that would hold a buffer page latch while waiting for `purge_sys.pq_mutex`.

dr-m: Oh, by the way, since inserting into a priority queue is a rather complex operation, it could make sense to define the function non-`inline`.

vlad-lesin:
buildbot/amd64-centos-7 and buildbot/amd64-ubuntu-2004-debug failed tests don't relate to my changes.

storage/innobase/include/trx0purge.h
```diff
@@ -55,80 +55,78 @@ Run a purge batch.
...
-	/** Track the current element in m_rsegs */
-	TrxUndoRsegs::const_iterator	m_iter;
-};
-
 /** The control structure used in the purge operation */
 class purge_sys_t
 {
-  friend TrxUndoRsegsIterator;
+  /** Min-heap based priority queue over fixed size array */
+  class purge_queue_t
```
dr-m, line 62: According to https://stackoverflow.com/questions/231760/what-does-a-type-followed-by-t-underscore-t-represent the `_t` suffix is reserved by POSIX for the implementation. We have been hit by that in the past; the
```c
#define lock_t ib_lock_t
```
is there to allow the code to be compiled on some system. I would not use the `_t` suffix in new data types.

storage/innobase/include/trx0purge.h
```diff
@@ -55,80 +55,78 @@ Run a purge batch.
...
+    bool empty() const { return !size(); }
+    void clear() { m_end= m_array; }
+
+    /** Push index of trx_sys.rseg_array into min-heap.
+    @param index index to push */
+    void push_index(byte index)
+    {
+      ut_ad(index < TRX_SYS_N_RSEGS);
+      ut_ad(size() + 1 <= TRX_SYS_N_RSEGS);
+      *(m_end++)= index;
```
dr-m, line 105: I think that this function needs to be `private`, or maybe merged with the public `push()`. The parentheses around the pointer post-increment are redundant.

The name `index` already has another meaning in the code base. I would suggest using just `i`, which we generally use for any array index.

storage/innobase/include/trx0purge.h
```diff
@@ -55,80 +55,78 @@ Run a purge batch.
...
+      std::push_heap(begin(), end(), trx_rseg_cmp());
+    }
+
+    /** Push rseg to priority queue.
+    @param rseg trx_rseg_t pointer to push */
+    void push(const trx_rseg_t *rseg)
+    {
+      byte index= byte(rseg - trx_sys.rseg_array);
```
dr-m, line 113: If we want to assert that `rseg` is within `trx_sys.rseg_array`, I think that we need to do it before lossily casting the signed `ptrdiff_t` result to an unsigned 8-byte integer:
```c++
ut_ad(rseg >= trx_sys.rseg_array);
ut_ad(rseg < trx_sys.rseg_array + TRX_SYS_N_RSEGS);
```

storage/innobase/include/trx0purge.h
```diff
@@ -55,80 +55,78 @@ Run a purge batch.
...
+      push_index(index);
+    }
+
+    /** Pop rseg from priority queue.
+    @return pointer to poped trx_rseg_t object */
```
dr-m, line 118: I think that the correct spelling is ‘popped’.

storage/innobase/include/trx0purge.h
```diff
@@ -244,15 +242,23 @@ class purge_sys_t
...
-	mysql_mutex_t	pq_mutex;	/*!< Mutex protecting purge_queue */
+  /** Push to purge queue without acquiring pq_mutex.
+  @param rseg rseg to push */
+  void enqueue(trx_rseg_t &rseg);
+
+  /** Acquare purge_queue_mutex */
+  void lock_purge_queue_mutex() { mysql_mutex_lock(&pq_mutex); }
+
+  /** Release purge queue mutex */
+  void unlock_purge_queue_mutex() { mysql_mutex_unlock(&pq_mutex); }
```
dr-m, line 261: The `_mutex` in these names is a bit redundant, and the original InnoDB naming pattern was `noun_verb`, like `purge_queue_lock()` or maybe `queue_lock()` as the `purge` is already part of the name of the singleton object: `purge_sys.queue_lock()`.

storage/innobase/include/trx0purge.h
```diff
@@ -438,6 +445,11 @@ class purge_sys_t
   @param	already_stopped	True indicates purge threads were
 				already stopped */
   void stop_FTS(const dict_table_t &table, bool already_stopped=false);
+
+  /** Cleanse purge queue to remove the rseg that reside in undo-tablespace
+  marked for truncate.
+  @param space undo tablespace being truncated */
+  void cleanse_purge_queue(const fil_space_t& space);
```
dr-m, line 452: In the common formatting style, there should be a space *before* a reference or pointer marker, not after it. It was the opposite way around in the original InnoDB formatting style.

storage/innobase/trx/trx0purge.cc
```diff
@@ -163,6 +85,15 @@ purge_graph_build()
 	return(fork);
 }
 
+/* Define it non-inline as pushing to priority queue is rather complex
+operation */
+void purge_sys_t::enqueue(trx_rseg_t &rseg)
+{
+  ut_ad(srv_operation == SRV_OPERATION_RESTORE ||
+        mysql_mutex_is_owner(&pq_mutex));
```
dr-m, line 93: We could add explicit
```c++
ATTRIBUTE_NOINLINE /* this is a rather complex operation */
```
I am a bit curious about this relaxation. Why do we need it during `mariadb-backup --prepare`? If the relaxation cannot be avoided, the assertion would have to be enclosed within
```c
#ifdef SAFE_MUTEX
```
There is no such issue with `mysql_mutex_assert_owner`.

storage/innobase/trx/trx0trx.cc
```diff
@@ -1138,15 +1138,23 @@ inline void trx_t::write_serialisation_history(mtr_t *mtr)
     }
     else if (rseg->last_page_no == FIL_NULL)
     {
-      mysql_mutex_lock(&purge_sys.pq_mutex);
+      /* trx_sys.assign_new_trx_no() and
+      purge_sys.push_to_purge_queue_no_lock() must be invoked in the same
```
dr-m, line 1142: The latter was renamed to `purge_sys.enqueue()`.

storage/innobase/trx/trx0purge.cc
```diff
@@ -917,24 +823,51 @@ void purge_sys_t::rseg_get_next_history_log()
...
+    mysql_mutex_unlock(&purge_sys.pq_mutex);
+    return false;
+  }
+  rseg= purge_queue.pop();
+  mysql_mutex_unlock(&purge_sys.pq_mutex);
+
+  /* We assume in purge of externally stored fields that space
+  id is in the range of UNDO tablespace space ids */
+  ut_ad(rseg->space == fil_system.sys_space
+   || srv_is_undo_tablespace(rseg->space->id));
```
dr-m, line 854: The line should be split after `||`, not before it.

storage/innobase/trx/trx0rseg.cc
```diff
@@ -584,7 +584,11 @@ dberr_t trx_rseg_array_init()
 #endif
 	mtr_t mtr;
 	dberr_t err = DB_SUCCESS;
-
+	/* Acquiring purge queue mutex here should be fine from the deadlock
+	prevention point of view, because executing that function is a
+	prerequisite for starting the purge subsystem or any transactions. */
+	if (srv_operation != SRV_OPERATION_RESTORE)
+		purge_sys.lock_purge_queue_mutex();
```
dr-m, line 591: The comment should explain why we are avoiding the mutex acquisition in `mariadb-backup --prepare`. Maybe we should simply call `purge_sys.create()` somewhere in `xtrabackup.cc`?

dr-m (commented):
Thank you, this looks good to me. I only have some coding style comments.

I did not know that `std::push_heap` and `std::pop_heap` existed.

storage/innobase/trx/trx0purge.cc
vlad-lesin: The relaxation is due to the following code:

```
dberr_t trx_lists_init_at_db_start()                                            
{                                                                               
...                                                                             
        if (srv_operation == SRV_OPERATION_RESTORE) {                           
                /* mariabackup --prepare only deals with                        
                the redo log and the data files, not with                       
                transactions or the data dictionary. */                         
                return trx_rseg_array_init();                                   
        }                                                                       
                                                                                
        if (srv_force_recovery >= SRV_FORCE_NO_UNDO_LOG_SCAN) {                 
                return DB_SUCCESS;                                              
        }                                                                       
                                                                                
        purge_sys.create();                                                     
        dberr_t err = trx_rseg_array_init();                                    
...                                                                             
}
```

storage/innobase/include/trx0purge.h
vlad-lesin: > I think that this function needs to be private

But it's used in `purge_sys_t::cleanse_purge_queue()`.

storage/innobase/trx/trx0purge.cc
vlad-lesin: Oh, the relaxation is not neccessary because `trx_rseg_mem_restore()` quits before initializing undo log lists.

vlad-lesin: I see that `ATTRIBUTE_NOINLINE` is used in header files. What transaction unit(s) contain(s) the compiled function's code? All units, which include the corresponding header file?

dlenski:
@dr-m wrote:

> > If I'm not mistaken, @vlad-lesin, you work for @mariadb-corporation.
> > However, the commit in this PR isn't using your @mariadb.com email in the author field. I believe that's why the CLAassistant is showing it as CLA-not-signed. And should probably be corrected to clarify the provenance of the PR.
> 
> @dlenski, I do not think that it is related to the email address. I use a mariadb.com email address in my commits and I work for MariaDB plc, but I have not signed the CLA. Instead, I have signed something equivalent, and I think @vlad-lesin has as well. Until some months ago, pull requests were only used by external contributors. The failing CLA check does not prevent this PR from being merged.

Shouldn't corp/foundation employs be submitting their commits with their `@mariadb.com`/`@mariadb.org` email addresses, regardless?

Among other things, it helps us external contributors to understand which organizations are driving specific contributions.

storage/innobase/include/trx0purge.h
```diff
@@ -55,80 +55,80 @@ Run a purge batch.
...
+    @return pointer to popped trx_rseg_t object */
+    trx_rseg_t *pop()
+    {
+      ut_ad(!empty());
+      byte index= m_array[0];
+      ut_ad(index < TRX_SYS_N_RSEGS);
+      std::pop_heap(begin(), end(), trx_rseg_cmp());
+      --m_end;
+      return &trx_sys.rseg_array[index];
+    }
```
dr-m, line 129: If I understood https://en.cppreference.com/w/cpp/algorithm/pop_heap correctly, this could be written as follows, possibly resulting in slightly improved code:
```c++
      ut_ad(!empty());
      std::pop_heap(begin(), end(), trx_seg_cmp());
      byte i= m_array[--m_end];
      ut_ad(i < TRX_SYS_N_RSEGS);
      return &trx_sys.rseg_array[i];

File: storage/innobase/trx/trx0purge.cc [hunk 16]
```diff
@@ -163,6 +85,13 @@ purge_graph_build()
 	return(fork);
 }
 
+ATTRIBUTE_NOINLINE /* this is a rather complex operation */
+void purge_sys_t::enqueue(trx_rseg_t &rseg)
```
dr-m, line 89: Given that all calls after server startup are unconditional from this same compilation unit from `trx_t::write_serialisation_history()`, it could make sense to omit the `ATTRIBUTE_NOINLINE`. I would suggest that you benchmark it once more.

storage/innobase/trx/trx0rseg.cc
```diff
@@ -584,7 +584,17 @@ dberr_t trx_rseg_array_init()
...
+	files, not with	transactions or the data dictionary, that's why
+	trx_lists_init_at_db_start() does not invoke purge_sys.create() and
+	purge queue mutex stays uninitialized, and trx_rseg_mem_restore() quits
+	before initializing undo log lists. */
+	if (srv_operation != SRV_OPERATION_RESTORE)
+		/* Acquiring purge queue mutex here should be fine from the
+		deadlock prevention point of view, because executing that
+		function is a prerequisite for starting the purge subsystem or
+		any transactions. */
+		purge_sys.queue_lock();
```
dr-m, line 597: Omitting the condition would result in a smaller executable. If there is no contention on the mutex (like there should not be here), unconditionally acquiring and releasing the mutex should be no problem.

dr-m (approved):
Looks OK. At your discretion, consider my final comments.

storage/innobase/trx/trx0rseg.cc
vlad-lesin: But the mutex is not initialized for `srv_operation == SRV_OPERATION_RESTORE`.

PR #3153, opened 2024-03-25
MDEV-33659 Fix crash in kdf() without parameters

andremralves:
- [x] *The Jira issue number for this PR is: MDEV-33659*

## Description
 
- When there is no parameter item_list is NULL.
- This code checks if item_list is NULL before setting arg_count to item_list->elements.

## Release Notes
Fix system crash when KDF() was called without parameters.

This function was added to MariaDB in 11.3 (see https://github.com/MariaDB/server/commit/4f9396b9f80031120f325d54f06d346dea7f1bfb and https://mariadb.com/kb/en/kdf) and did not exist in earlier major versions.

## How can this PR be tested?

```sh
./mtr main.func_kdf
```

## PR quality check

### DIFF
--- a/sql/item_create.cc
+++ b/sql/item_create.cc
@@ -3229,8 +3229,12 @@ Item*
 Create_func_kdf::create_native(THD *thd, const LEX_CSTRING *name,
                                List<Item> *item_list)
 {
-  uint arg_count= item_list->elements;
   Item *a[5];
+  uint arg_count= 0;
+
+  if (item_list != NULL)
+    arg_count= item_list->elements;
+
   for (uint i=0; i < MY_MIN(array_elements(a), arg_count); i++)
     a[i]= item_list->pop();
   switch (arg_count)
--- a/mysql-test/main/func_kdf.result
+++ b/mysql-test/main/func_kdf.result
@@ -154,5 +154,10 @@ NULL
 Warnings:
 Warning	3047	Invalid argument error: 65537 in function kdf.
 #
+# MDEV-33659 Test kdf() without parameters
+#
+select kdf();
+ERROR 42000: Incorrect parameter count in the call to native function 'kdf'
+#
 # End of 11.3 tests
 #
--- a/mysql-test/main/func_kdf.test
+++ b/mysql-test/main/func_kdf.test
@@ -52,6 +52,13 @@ select length(kdf('foo', 'bar', 100, 'pbkdf2_hmac', 65536));
 select length(kdf('foo', 'bar', 100, 'pbkdf2_hmac', 65537));
 
 --echo #
+--echo # MDEV-33659 Test kdf() without parameters
+--echo #
+
+--error ER_WRONG_PARAMCOUNT_TO_NATIVE_FCT
+select kdf();
+
+--echo #
 --echo # End of 11.3 tests
 --echo #
 


grooverdan (approved):
Looks good. Rebased to 11.4 as there are going to be no more 11.3 releases due to changed release mode.

dlenski (approved):
Great catch!

sql/item_create.cc
```diff
@@ -3229,8 +3229,12 @@ Item*
 Create_func_kdf::create_native(THD *thd, const LEX_CSTRING *name,
                                List<Item> *item_list)
 {
-  uint arg_count= item_list->elements;
   Item *a[5];
+  uint arg_count= 0;
+
+  if (item_list != NULL)
+    arg_count= item_list->elements;
```
dlenski, line 3236: Now I wonder if there are any/many other SQL functions with the similar mistake of checking `item_list->elements` before checking `item_list != NULL`.

Did you search for any of them, by any chance?

dlenski:
>  This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.

Perhaps worth mentioning in the description that the `KDF()` SQL function [was newly-added in 11.3.0](https://mariadb.com/kb/en/kdf) and not present in earlier major versions at all.

sql/item_create.cc
grooverdan: I did.

grooverdan:
> > This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.
> 
> Perhaps worth mentioning in the description that the `KDF()` SQL function [was newly-added in 11.3.0](https://mariadb.com/kb/en/kdf) and not present in earlier major versions at all.

The 11.4 target version is that 11.3 is EOL now

dlenski:
@grooverdan wrote:
>  The 11.4 target version is that 11.3 is EOL now

Are there some words missing in this sentence?

grooverdan:
> @grooverdan wrote:
> 
> > The 11.4 target version is that 11.3 is EOL now
> 
> Are there some words missing in this sentence?

Sorry
 that -> because
 
I'm a little confused as to why the introduced version needs to be mentioned.

dlenski:
@grooverdan wrote:

> I'm a little confused as to why the introduced version needs to be mentioned.

When I see that there's a crash-inducing bug in a SQL function *and* that this bug is being fixed _only_ on 11.4+ …

… my first thoughts are:

* How long has this bug existed?
* Why isn't it being fixed on earlier MariaDB major versions?

In order to answer those questions, I have to search around for relevant docs. So then I find [the `KDF()` docs in MariaDB.com knowledge base](https://mariadb.com/kb/en/kdf), which do clarify that this function was _only recently added_, specifically in 11.3.

To simplify the understanding of this change for anyone reviewing the commit and/or PR, how about adding something like this @andremralves?

> Fix system crash when **the SQL function** KDF() was called without parameters.
>
> This function was added to MariaDB in 11.3 (see 4f9396b9f80031120f325d54f06d346dea7f1bfb and https://mariadb.com/kb/en/kdf) and did not exist in earlier major versions.

dlenski (changes requested):
Can you add a bit more information to the commit and PR description to clarify in what version this bug originated? See
https://github.com/MariaDB/server/pull/3153#issuecomment-2032814860

grooverdan:
updated maintaince policy - https://mariadb.org/about/#maintenance-policy - we're not doing 11.3 any more.

andremralves:
Hi @dlenski, I added your suggestions to the release note. Thanks.

LinuxJedi (approved):
Looking good, many thanks!

PR #3158, opened 2024-03-27
MDEV-33779 InnoDB row operations could be faster

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33779*
## Description
We have quite a few assertions
```c++
ut_a(m_prebuilt->trx == thd_to_trx(ha_thd()));
```
in low-level functions.
These had better be debug assertions for performance reasons. It should suffice to check that condition in the less frequently invoked `ha_innobase::change_active_index()`.

`convert_search_mode_to_innobase()`: Return whether the mode is unsupported, and optionally update `ha_innobase::m_last_match_mode`.

`ha_innobase::index_read()`: Only branch on `find_flag` once, and simplify the error handling after invoking `row_search_mvcc()`.

`ha_innobase::rnd_pos()`: Remove an assertion that is duplicating one in `ha_innobase::index_read()`, which we are calling unconditionally.

`ha_innobase::records_in_range()`: Check only once whether `min_key`, `max_key` are null pointers.

`row_sel_convert_mysql_key_to_innobase()`: Declare that all parameters except the conversion buffer pointer `buf` be `nonnull`.

## Release Notes
This is a performance improvement that should not affect compatibility in any way.

## How can this PR be tested?
This code is covered by nearly all `SELECT`, `UPDATE`, `DELETE` operations on InnoDB tables, so it should be well covered by the regression test suite. Some additional stress testing with RQG could be useful.

This is a performance improvement that could apply to older versions as well, but 10.6 is the earliest maintained branch where most InnoDB performance improvements land.
## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -8883,47 +8883,63 @@ ha_innobase::index_end(void)
 	DBUG_RETURN(0);
 }
 
-/*********************************************************************//**
-Converts a search mode flag understood by MySQL to a flag understood
-by InnoDB. */
-page_cur_mode_t
-convert_search_mode_to_innobase(
-/*============================*/
-	ha_rkey_function	find_flag)
+/** Convert a MariaDB search mode to an InnoDB search mode.
+@tparam last_match whether last_match_mode is to be set
+@param find_flag       MariaDB search mode
+@param mode            InnoDB search mode
+@param last_match_mode pointer to ha_innobase::m_last_match_mode
+@return whether the search mode is unsupported */
+template<bool last_match= false>
+static bool convert_search_mode_to_innobase(ha_rkey_function find_flag,
+                                            page_cur_mode_t &mode,
+                                            uint *last_match_mode= nullptr)
 {
-	switch (find_flag) {
-	case HA_READ_KEY_EXACT:
-		/* this does not require the index to be UNIQUE */
-	case HA_READ_KEY_OR_NEXT:
-		return(PAGE_CUR_GE);
-	case HA_READ_AFTER_KEY:
-		return(PAGE_CUR_G);
-	case HA_READ_BEFORE_KEY:
-		return(PAGE_CUR_L);
-	case HA_READ_KEY_OR_PREV:
-	case HA_READ_PREFIX_LAST:
-	case HA_READ_PREFIX_LAST_OR_PREV:
-		return(PAGE_CUR_LE);
-	case HA_READ_MBR_CONTAIN:
-		return(PAGE_CUR_CONTAIN);
-	case HA_READ_MBR_INTERSECT:
-		return(PAGE_CUR_INTERSECT);
-	case HA_READ_MBR_WITHIN:
-		return(PAGE_CUR_WITHIN);
-	case HA_READ_MBR_DISJOINT:
-		return(PAGE_CUR_DISJOINT);
-	case HA_READ_MBR_EQUAL:
-		return(PAGE_CUR_MBR_EQUAL);
-	case HA_READ_PREFIX:
-		return(PAGE_CUR_UNSUPP);
-	/* do not use "default:" in order to produce a gcc warning:
-	enumeration value '...' not handled in switch
-	(if -Wswitch or -Wall is used) */
-	}
+  mode= PAGE_CUR_LE;
+  if (last_match)
+    *last_match_mode= 0;
 
-	my_error(ER_CHECK_NOT_IMPLEMENTED, MYF(0), "this functionality");
+  switch (find_flag) {
+  case HA_READ_KEY_EXACT:
+    /* this does not require the index to be UNIQUE */
+    if (last_match)
+      *last_match_mode= ROW_SEL_EXACT;
+    /* fall through */
+  case HA_READ_KEY_OR_NEXT:
+    mode= PAGE_CUR_GE;
+    return false;
+  case HA_READ_AFTER_KEY:
+    mode= PAGE_CUR_G;
+    return false;
+  case HA_READ_BEFORE_KEY:
+    mode= PAGE_CUR_L;
+    return false;
+  case HA_READ_PREFIX_LAST:
+    if (last_match)
+      *last_match_mode= ROW_SEL_EXACT_PREFIX;
+    /* fall through */
+  case HA_READ_KEY_OR_PREV:
+  case HA_READ_PREFIX_LAST_OR_PREV:
+    return false;
+  case HA_READ_MBR_CONTAIN:
+    mode= PAGE_CUR_CONTAIN;
+    return false;
+  case HA_READ_MBR_INTERSECT:
+    mode= PAGE_CUR_INTERSECT;
+    return false;
+  case HA_READ_MBR_WITHIN:
+    mode= PAGE_CUR_WITHIN;
+    return false;
+  case HA_READ_MBR_DISJOINT:
+    mode= PAGE_CUR_DISJOINT;
+    return false;
+  case HA_READ_MBR_EQUAL:
+    mode= PAGE_CUR_MBR_EQUAL;
+    return false;
+  case HA_READ_PREFIX:
+    break;
+  }
 
-	return(PAGE_CUR_UNSUPP);
+  return true;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -9023,91 +9022,100 @@ ha_innobase::index_read(
...
-
-		match_mode = ROW_SEL_EXACT;
-
-	} else if (find_flag == HA_READ_PREFIX_LAST) {
+	page_cur_mode_t	mode = PAGE_CUR_LE;
+	m_last_match_mode = 0;
 
-		match_mode = ROW_SEL_EXACT_PREFIX;
+	switch (find_flag) {
+	case HA_READ_KEY_EXACT:
```
mariadb-DebarunBanerjee, line 9040: I understand this is for performance but it is leading to code duplication with convert_search_mode_to_innobase().

1. Perhaps inlining the convert_search_mode_to_innobase() could have similar effect ?
2. The function could return a std::pair of cursor_mode and match_mode and match_mode can be ignored in other callers.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -9023,91 +9022,100 @@ ha_innobase::index_read(
...
+	case HA_READ_MBR_DISJOINT:
+		mode = PAGE_CUR_DISJOINT;
+		break;
+	case HA_READ_MBR_EQUAL:
+		mode = PAGE_CUR_MBR_EQUAL;
+		break;
+	case HA_READ_PREFIX:
+		table->status = STATUS_NOT_FOUND;
+		DBUG_RETURN(HA_ERR_UNSUPPORTED);
 	}
```
mariadb-DebarunBanerjee, line 9077: After comparing with  convert_search_mode_to_innobase() I find one difference i.e. when the  flag doesn't match with any of the cases.
```
page_cur_mode_t
convert_search_mode_to_innobase(
/*============================*/
        ha_rkey_function        find_flag)
    ...
        my_error(ER_CHECK_NOT_IMPLEMENTED, MYF(0), "this functionality");
        return(PAGE_CUR_UNSUPP);
```
Can we add it as the default case here to ensure zero functional impact ?

mariadb-DebarunBanerjee (commented):
Looks good. I have some minor comments. I will approve after your reply.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -9013,7 +9012,7 @@ ha_innobase::index_read(
 		build_template(false);
 	}
 
-	if (key_ptr != NULL) {
+	if (key_len) {
```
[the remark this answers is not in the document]
dr-m: This statement is guarding a call to `row_sel_convert_mysql_key_to_innobase()`. Near the start of that function there is an assignment
```c++
	key_end = key_ptr + key_len;
```
In `clang -fsanitize=undefined` (`cmake -DWITH_UBSAN=ON`) builds this would lead to
```
runtime error: applying non-zero offset _ to null pointer
```
However, in GCC (up to 14.0.1) there is no check like this, and even after [MDEV-26272](https://jira.mariadb.org/browse/MDEV-26272) was fixed, there still are many bugs outside InnoDB that prevent us from using `clang -fsanitize=undefined`. There is a check that enforces `__attribute__((nonnull))` in both GCC and clang:
``` 
runtime error: null pointer passed as argument _, which is declared to never be null
```
I would address this concern by adding a missing `__attribute__((nonnull))` to the declaration of `row_sel_convert_mysql_key_to_innobase()`. All parameters are supposed to be nonnull.

storage/innobase/handler/ha_innodb.cc
dr-m: GCC at least since version 3 implements a warning if there is a `switch` on an `enum` variable and some constant is not handled. I tested it by removing one of the `case` labels:
```
/mariadb/10.6/storage/innobase/handler/ha_innodb.cc:9040:16: warning: enumeration value ‘HA_READ_KEY_OR_PREV’ not handled in switch [-Wswitch]
```
This useful compile-time check would be disabled if we added a `default:` label. I think that it is better to list all `enum` values so that whenever someone decides to add one more `enum` values, there will be a warning that will force this code to be adjusted.

storage/innobase/handler/ha_innodb.cc
dr-m: I see that `convert_search_mode_to_innobase()` was missing `static`. I am going to implement it as follows, to avoid any run-time overhead:
```c++
/** Convert a MariaDB search mode to an InnoDB search mode.
@tparam last_match whether last_match_mode is to be set
@param find_flag       MariaDB search mode
@param mode            InnoDB search mode
@param last_match_mode pointer to ha_innobase::m_last_match_mode
@return whether the search mode is unsupported */
template<bool last_match= false>
static bool convert_search_mode_to_innobase(ha_rkey_function find_flag,
                                            page_cur_mode_t &mode,
                                            uint *last_match_mode= nullptr)
```
Furthermore, I am going to deduplicate some conditions in `ha_innobase::records_in_range()`.

dr-m: In fact, that refactoring will shorten both `ha_innobase::index_read()` and `ha_innobase::records_in_range()` by some dozens of bytes. The function (with `last_match=true`) would be inlined in the former, and non-inlined in the latter.

storage/innobase/handler/ha_innodb.cc
dr-m: I am going to retain `convert_search_mode_to_innobase()`, and the `return true` (to indicate the unsupported code path) will be outside the `switch` statement.

storage/innobase/handler/ha_innodb.cc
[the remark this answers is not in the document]
dr-m: The calls in `ha_innobase::records_in_range()` do look like undefined behaviour to me, so I am fixing it. I will try to get the regression tests to pass with the added `__attribute__((nonnull))`.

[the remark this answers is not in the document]
dr-m: The conversion buffer `buf` can be passed as `nullptr`. a6b365af342969507314804ad7c65fdd78afc481 declares the other parameters as `nonnull`.

storage/innobase/handler/ha_innodb.cc
dr-m: a6b365af342969507314804ad7c65fdd78afc481 implements this.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -14302,61 +14285,50 @@ ha_innobase::records_in_range(
...
-	dict_index_copy_types(range_end, index, key->ext_key_parts);
-
-	row_sel_convert_mysql_key_to_innobase(
-		range_start,
-		m_prebuilt->srch_key_val1,
-		m_prebuilt->srch_key_val_len,
-		index,
-		(byte*) (min_key ? min_key->key : (const uchar*) 0),
-		(ulint) (min_key ? min_key->length : 0));
-
```
mariadb-DebarunBanerjee, line 14317: For both min_key and max_key we were calling row_sel_convert_mysql_key_to_innobase with key_ptr=nullptr. While the function row_sel_convert_mysql_key_to_innobase() doesn't seem to handle nullptr and is likely was an issue already.

This is great that the refactoring change improves this part. It also indicates that we may not have any mtr test covering the case when either of min or max key is zero. Can we check if such tests are there or can be added ?

mariadb-DebarunBanerjee (approved):
Thanks @dr-m. The latest changes with the added refactoring look good to me. I have one test suggestion for you but I am approving the changes as the test should not be a blocker for this patch.

storage/innobase/handler/ha_innodb.cc
dr-m: That function definitely did receive `key_ptr=nullptr`, but GCC does not care about the UB of adding something to a null pointer. Only `clang` does, but unfortunately `clang -fsanitize=undefined`  is still unusable due to bugs in code outside InnoDB. I got it flagged in GCC after I added the `__attribute__((nullptr))`.  I wrote about that in the thread about the first comment of your first review. That is what prompted this entire change.

PR #3159, opened 2024-03-28
MDEV-33290: Disable ColumnStore based on boost version (postfix)

grooverdan:
Its important to fail early and only contine with the include after the boost version check succeeds.

- [x] *The Jira issue number for this PR is: MDEV-33290*

## Description

Correct previous mistake in determining the boost version - check, then include columnstore.

## Release Notes

Nothing required.

## How can this PR be tested?

See if it builds correctly on fedora autobake builders

## PR quality check

### DIFF
--- a/storage/columnstore/CMakeLists.txt
+++ b/storage/columnstore/CMakeLists.txt
@@ -13,14 +13,16 @@ endmacro()
 
 IF(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR
 CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64")
-    add_subdirectory(columnstore)
+    cmake_policy(SET CMP0093 NEW)
 
     # https://jira.mariadb.org/browse/MCOL-5611
-    FIND_PACKAGE(Boost 1.80.0 COMPONENTS system filesystem thread regex date_time chrono atomic)
-    IF (Boost_FOUND)
+    FIND_PACKAGE(Boost COMPONENTS system filesystem thread regex date_time chrono atomic)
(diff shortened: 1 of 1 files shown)


storage/columnstore/CMakeLists.txt
```diff
@@ -13,15 +13,17 @@ endmacro()
...
 IF(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR
 CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64")
-    add_subdirectory(columnstore)
+    cmake_policy(SET CMP0093 NEW)
 
     # https://jira.mariadb.org/browse/MCOL-5611
-    FIND_PACKAGE(Boost 1.80.0 COMPONENTS system filesystem thread regex date_time chrono atomic)
-    IF (Boost_FOUND)
+    FIND_PACKAGE(Boost COMPONENTS system filesystem thread regex date_time chrono atomic)
+    IF (Boost_VERSION VERSION_GREATER "1.79.0")
```
lottaquestions, line 20: Nice switch from a specific target version to a range of versions that are not covered. Is there a roadmap to update the codebase to support Boost 1.80.0 and above?

vuvova: this fix is only for 10.5 which uses a very old ColumnStore. Using ColumnStore in 10.5 is generally not recommended. 10.6+ are ok

vuvova:
merged. thanks.

grooverdan:
thanks for the merge. I didn't get to completely check all build paths where right.

PR #3161, opened 2024-03-28
MDEV-32633: Fix Galera cluster <-> native replication interaction

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-32633*

GTID events are applied without a running server transaction, we need to set next transaction ID for Wsrep transaction.

The whole Galera cluster now has a single GTID value (including the server ID throughout the cluster), fix the config accordingly.

Add force restart so that repeated MTR test execution prints consistent GTID values, otherwise they would have been recovered from the previous run.

## Release Notes

The fix is for versions 10.5 upwards.

## How can this PR be tested?

Re-enabled previously failing MTR test.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera_3nodes/disabled.def
+++ b/mysql-test/suite/galera_3nodes/disabled.def
@@ -11,7 +11,6 @@
 ##############################################################################
 
 galera_2_cluster : MDEV-32631 galera_2_cluster: before_rollback(): Assertion `0' failed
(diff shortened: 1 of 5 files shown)


sysprg:
Thanks, fixes merged with head revision:
https://github.com/MariaDB/server/commit/0cc9b49751fc86f7942e19fe6ed8f7227f847c02
https://github.com/MariaDB/server/commit/a4838721a252ea5570b5fed9ab56e38e5b234865
https://github.com/MariaDB/server/commit/c21aa486a86d9db29ad16ff279380969443df00f

PR #3163, opened 2024-03-28
MDEV-20094 InnoDB blob allocation allocates extra extents

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-20094*

## Description
- InnoDB reserves the free extents unnecessarily during blob page allocation even though btr_page_alloc() can handle reserving the extent when the existing ran out of pages to be used.

## PR quality check


tonychen2001:
The tests are all failing on `innodb.check_ibd_filesize` which seems directly related to the changes in this PR.

I guess a `check_ibd_filesize,16k.rdiff` file is needed?

```
innodb.check_ibd_filesize '16k,innodb'   w3 [ fail ]
        Test ended at 2024-03-28 11:55:40
CURRENT_TEST: innodb.check_ibd_filesize
--- /home/buildbot/amd64-ubuntu-2204-debug-ps/build/mysql-test/suite/innodb/r/check_ibd_filesize.result	2024-03-28 11:48:01.000000000 +0000
+++ /home/buildbot/amd64-ubuntu-2204-debug-ps/build/mysql-test/suite/innodb/r/check_ibd_filesize.reject	2024-03-28 11:55:39.955625002 +0000
@@ -10,7 +10,7 @@
 CREATE TABLE t1 (a INT PRIMARY KEY, b BLOB) ENGINE=InnoDB;
 # bytes: 65536
 INSERT INTO t1 SELECT seq,REPEAT('a',30000) FROM seq_1_to_20;
-# bytes: 4194304
+# bytes: 2097152
 DROP TABLE t1;
 CREATE TABLE t1 (a INT PRIMARY KEY, b BLOB)
 ENGINE=InnoDB ROW_FORMAT=COMPRESSED KEY_BLOCK_SIZE=1;
Result content mismatch
```

dr-m:
> I guess a `check_ibd_filesize,16k.rdiff` file is needed?

Because the default is `innodb_page_size=16k`, I think that the `.result` file had better correspond to that variant. Try
```sh
./mtr --record innodb.check_ibd_filesize,16k
```
Shouldn’t this target the earliest maintained version (10.4)? Or has this code diverged between 10.5 and 10.6?

mysql-test/suite/innodb/r/check_ibd_filesize,32k.rdiff
```diff
@@ -1,5 +1,5 @@
---- mysql-test/suite/innodb/r/check_ibd_filesize.result	2022-08-16 17:28:06.462350465 +0530
-+++ mysql-test/suite/innodb/r/check_ibd_filesize.reject	2022-08-16 17:29:25.129637040 +0530
+--- check_ibd_filesize.result	2024-04-03 13:46:12.207412052 +0530
++++ check_ibd_filesize.reject	2024-04-03 13:48:48.297394417 +0530
```
dr-m, line 2: I would omit the time stamps from these lines.

dr-m (approved):
This looks OK to me after cleaning up the `.rdiff` files. The failures of `main.query_cache` in embedded server builds look unrelated to this.

mysql-test/suite/innodb/r/check_ibd_filesize,32k.rdiff
```diff
@@ -1,5 +1,3 @@
---- mysql-test/suite/innodb/r/check_ibd_filesize.result	2022-08-16 17:28:06.462350465 +0530
-+++ mysql-test/suite/innodb/r/check_ibd_filesize.reject	2022-08-16 17:29:25.129637040 +0530
```
dr-m, line 2: You should retain these lines, but only omit the part that starts with a timestamp. Example:
```diff
--- check_ibd_filesize.result
+++ check_ibd_filesize.result
```
If you omit the lines, then there will be additional junk in the output on the Windows builder, which uses a suitably old version of the `patch` utility.

PR #3164, opened 2024-03-28
MDEV-33749 hyphen in table name can cause galera certification failures

sjaakola:
- [x] *The Jira issue number for this PR is: MDEV-33749*

## Description
Table and database names in MariaDB can contain regular ASCII characters and a wide variety of special characters. Galera replication write set contains key information of elements accessed in the write set,this key information may contain database and table names and row level key information.
Key information for galera replication may be appended in several phases of transaction execution. it has now turned out that database and table name parts of the key information may be in different character set encoding, depending on where the respective name is retrieved from: 
if key is retrieved during innodb foreign key checking phase, the name parts would be in "filename" character set encoding. And other references would use the system character set encoding.

## Release Notes
If foreign key constraint was referencing a table where either the table or database name contained special characters, then galera replication would record the reference with incompatible character set encoding, with respect to how the key reference is recorded in direct accesses to the same rows. This could lead to unreliable galera write set certification and cause crashes or cluster hanging in parallel applying or cluster wide conflict resolving.   

## How can this PR be tested?
There is a dedicated mtr test: galera.galera_table_with_hyphen for this issue

## PR quality check


sysprg:
Thanks, fix merged with head revision (https://github.com/MariaDB/server/commit/2fcf2ec229862fe24a97c54db4e62f1ae8f9ed63)

PR #3165, opened 2024-03-28
MDEV-32397 join_read_first, keyread SEGV crash

DaveGosselin-MariaDB:
Queries having the following form may cause a crash
  SELECT t1.a FROM ( SELECT a AS a1 FROM t1 ) dt
  JOIN t1 ON a1 LIKE EXISTS ( SELECT a + RAND () FROM t1 UNION SELECT a FROM t1);
because the table t1 has some cleanup operations performed prematurely during the subselect, before the outer query has finished with the table.

In this particular case, the presence of RAND() makes the subquery uncacheable, necessitating the need to execute the subquery multiple times during join record evaluation.  Each time the subquery runs, it creates its own JOIN structure which has references to the table t1. When the subquery completes, JOIN::cleanup and functions called by it result in ha_end_keyread() being called on table t1.  However, we are not done with table t1 because the upper level `select t1.a from...` query requires table t1 to be open.  To solve this, we make the executor aware of when we're in subqueries like this and delay when we call ha_end_keyread() until the end of the parent query.


DaveGosselin-MariaDB:
I'm closing this PR without delivering.  I realized that the problem in another ticket, MDEV-32403, is inclusive of the problem fixed here.  I discovered what I think is a better fix overall for that issue.  Consequently I'll post my change for that ticket for review instead and mark MDEV-32397 as a duplicate of that ticket.

PR #3129, opened 2024-03-15
Add #if defines around usage of OS_DATA_FILE_NO_O_DIRECT

kalyan02:
## Description

There is a compile error on 11.5 branch in the server.git repo when building mariadbd on a m1 mac.

This has to do with the unguarded usage of `OS_DATA_FILE_NO_O_DIRECT`. Its only defined when `WIN32 || HAVE_FCNTL_DIRECT` which fails on macos (atleast current code paths don't define it and afaik macos doesn't have `O_DIRECT`).

https://github.com/mariadb/server/blob/11.5/storage/innobase/fil/fil0fil.cc#L1415

Am using the default cmake flags in gitlab yml

`-DWITH_SSL=system -DPLUGIN_COLUMNSTORE=NO -DPLUGIN_ROCKSDB=NO -DPLUGIN_S3=NO -DPLUGIN_MROONGA=NO -DPLUGIN_CONNECT=NO -DPLUGIN_TOKUDB=NO -DWITH_WSREP=OFF`

This patch makes it compile.  

```
diff --git a/storage/innobase/fil/fil0fil.cc b/storage/innobase/fil/fil0fil.cc
index 0ce54df6574..75ac2bfb53a 100644
--- a/storage/innobase/fil/fil0fil.cc
+++ b/storage/innobase/fil/fil0fil.cc
@@ -1409,11 +1409,12 @@ ATTRIBUTE_COLD void fil_space_t::reopen_all()
         continue;

       ulint type= OS_DATA_FILE;
-
+#if defined _WIN32 || defined HAVE_FCNTL_DIRECT
       switch (FSP_FLAGS_GET_ZIP_SSIZE(space.flags)) {
       case 1: case 2:
         type= OS_DATA_FILE_NO_O_DIRECT;
       }
+#endif
```

## Release Notes
- Fixes build for Macos/M1

## How can this PR be tested?
- Checkout repository on macOS/m1 and run cmake with default flags

## Basing the PR against the correct MariaDB version
- [ ] *This is a new feature and the PR is based against the latest MariaDB development branch.*
- [x] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [x] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #3166, opened 2024-04-01
Check if OS_DATA_FILE_NO_O_DIRECT is defined before using it

jmcarcell:
## Description
When building on MacOS (clang 15) `OS_DATA_FILE_NO_O_DIRECT` isn't defined:

```
  >> 4543    /mariadb/extra/mariabackup/xtrabackup.cc:2421:39: error: use of undeclared identifier 'OS_DATA_FILE_NO_O_DIRECT'
     4544                                          OS_DATA_FILE_NO_O_DIRECT, false, &ret);
```

This PR changes two files to allow compilation on MacOS. In both cases I just made a similar check to those already being done in `fil0fil.cc`.

## Release Notes
- Check if OS_DATA_FILE_NO_O_DIRECT is defined before using it, fixing compilation on MacOS

## PR quality check


grooverdan:
like #3129, it fixed, just hasn't been fully merged up to 11.5 yet.

Thanks for the consideration and effort making this PR.

PR #3168, opened 2024-04-01
MDEV-33799: mysql_manager_submit Segfault at Startup Still Possible During Recovery

bnestere:
MDEV-26473 fixed a segmentation fault at startup between the handle
manager thread and the binlog background thread, such that the
binlog background thread could be started and submit a job to the
handle manager, before it had initialized. Where MDEV-26473 made it
so the handle manager would initialize before the main thread
started the normal binary logs, it did not account for the recovery
case. That is, there is still a possibility of a segmentation fault
when a server is recovering using the binary logs such that it can
open the binary logs, start the binlog background thread, and submit
a job to the handle manager before it is initialized.

 1. The first commit is the regression with a test case that reproduces the issue
 2. The second commit is the fix


andrelkin (approved):
The path looks good! The piece of code relocation must be safe.

PR #3171, opened 2024-04-02
MDEV-33802 Weird read view after ROLLBACK of another transaction

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33802*
## Description
Even after commit b8a671988954870b7db22e20d1a1409fd40f8e3d there is an anomaly where a locking read could return inconsistent results. If a locking read would have to wait for a record lock, then by the definition of a read view, the modifications made by the current lock holder cannot be visible in the read view. This is because the read view must exclude any transactions that had not been committed at the time when the read view was created.

`lock_rec_convert_impl_to_expl_for_trx()`, `lock_rec_convert_impl_to_expl()`: Return an unsafe-to-dereference pointer to a transaction that holds or held the lock, or `nullptr` if the lock was available.

`lock_clust_rec_modify_check_and_lock()`, `lock_sec_rec_read_check_and_lock()`, `lock_clust_rec_read_check_and_lock()`: Return `DB_RECORD_CHANGED` if `innodb_strict_isolation=ON` and the lock was being held by another transaction.

The test case, which is based on a bug report by Zhuang Liu, covers the function `lock_sec_rec_read_check_and_lock()`.

## Release Notes
See #3067.

## How can this PR be tested?
I think that this is adequately covered by `./mtr innodb.lock_isolation`. Our Random Query Generator based stress testing does not cover this kind of problems. https://github.com/jepsen-io/jepsen would.

This is a bug fix that depends on the earlier fix #3067.

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/lock_isolation.test
+++ b/mysql-test/suite/innodb/t/lock_isolation.test
@@ -79,7 +79,6 @@ COMMIT;
 --connection consistent
 --reap
 SELECT * FROM t;
---disconnect consistent
 
 --connection default
 TRUNCATE TABLE t;
@@ -103,8 +102,48 @@ COMMIT;
 --connection con_weird
 --reap
 COMMIT;
---disconnect con_weird
 
 --connection default
 SELECT * FROM t;
+DROP TABLE t;
+
+--echo #
+--echo # MDEV-33802 Weird read view after ROLLBACK of other transactions
+--echo #
+
+CREATE TABLE t(a INT PRIMARY KEY, b INT UNIQUE) ENGINE=InnoDB;
+INSERT INTO t SET a=1;
+
+BEGIN; INSERT INTO t SET a=2;
+
+--connection consistent
+START TRANSACTION WITH CONSISTENT SNAPSHOT;
+--disable_ps2_protocol
+--error ER_CHECKREAD
+SELECT * FROM t FORCE INDEX (b) FOR UPDATE;
+--enable_ps2_protocol
+
+--connection con_weird
+START TRANSACTION WITH CONSISTENT SNAPSHOT;
+send
+SELECT * FROM t FORCE INDEX (b) FOR UPDATE;
+
+--connection default
+let $wait_condition=
+  select count(*) = 1 from information_schema.processlist
+  where state = 'Sending data'
+  and info LIKE 'SELECT * FROM t %';
+--source include/wait_condition.inc
+ROLLBACK;
+
+--connection con_weird
+--reap
+SELECT * FROM t FORCE INDEX (b) FOR UPDATE;
+--disconnect con_weird
+
+--connection consistent
+SELECT * FROM t FORCE INDEX (b) FOR UPDATE;
+--disconnect consistent
+
+--connection default
 DROP TABLE t;
--- a/mysql-test/suite/innodb/r/lock_isolation.result
+++ b/mysql-test/suite/innodb/r/lock_isolation.result
@@ -82,7 +82,6 @@ SELECT * FROM t;
 a	b
 10	20
 10	20
-disconnect consistent;
 connection default;
 TRUNCATE TABLE t;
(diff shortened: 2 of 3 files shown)


mysql-test/suite/innodb/t/lock_isolation.test
```diff
@@ -103,8 +102,46 @@ COMMIT;
...
+
+CREATE TABLE t(a INT PRIMARY KEY, b INT UNIQUE) ENGINE=InnoDB;
+INSERT INTO t SET a=1;
+
+BEGIN; INSERT INTO t SET a=2;
+
+--connection consistent
+START TRANSACTION WITH CONSISTENT SNAPSHOT;
+--error ER_CHECKREAD
+SELECT * FROM t FORCE INDEX (b) FOR UPDATE;
```
dr-m, line 122:
```
CURRENT_TEST: innodb.lock_isolation
mysqltest: At line 122: query 'SELECT * FROM t FORCE INDEX (b) FOR UPDATE' failed with wrong errno ER_LOCK_WAIT_TIMEOUT (1205): 'Lock wait timeout exceeded; try restarting transaction', instead of ER_CHECKREAD (1020)...
```
https://buildbot.mariadb.org/#/builders/534/builds/16014/steps/7/logs/stdio

dr-m: Even after applying some work-arounds (including `SET innodb_snapshot_isolation=ON` in `connection default`) I also reproduced the following bug in a test that precedes the one for this one:
```diff
CURRENT_TEST: innodb.lock_isolation
--- /mariadb/10.6/mysql-test/suite/innodb/r/lock_isolation.result	2024-04-09 10:28:08.988756438 +0300
+++ /mariadb/10.6/mysql-test/suite/innodb/r/lock_isolation.reject	2024-04-09 10:37:27.459725141 +0300
@@ -102,7 +102,7 @@
 connection default;
 SELECT * FROM t;
 a	b
-10	1
+10	20
 10	20
 DROP TABLE t;
 #

Result length mismatch
```
This is executed in a new transaction on its own, right after the execution of `COMMIT` in `con_weird`. Apparently, the `UPDATE` in `con_weird` can occasionally persistently corrupt the table. This is immediately before the results for this regression test.

After I moved the tests for this bug into a separate file, it passed 120,000 rounds of testing. I was so far unable to reproduce the failure that occurred on the CI system.

dr-m: `START TRANSACTION WITH CONSISTENT SNAPSHOT` does create a new read view in `innobase_start_trx_and_assign_read_view()` just fine. The bug is that the error is being ignored, and a new transaction would be implicitly started when using `./mtr --ps-protocol`. The following works around that problem.
```diff
diff --git a/mysql-test/suite/innodb/t/lock_isolation.test b/mysql-test/suite/innodb/t/lock_isolation.test
index 4c17afde7af..04a80fcae78 100644
--- a/mysql-test/suite/innodb/t/lock_isolation.test
+++ b/mysql-test/suite/innodb/t/lock_isolation.test
@@ -118,8 +118,10 @@ BEGIN; INSERT INTO t SET a=2;
 
 --connection consistent
 START TRANSACTION WITH CONSISTENT SNAPSHOT;
+--disable_ps_protocol
 --error ER_CHECKREAD
 SELECT * FROM t FORCE INDEX (b) FOR UPDATE;
+--enable_ps_protocol
 
 --connection con_weird
 START TRANSACTION WITH CONSISTENT SNAPSHOT;
```

dr-m: `--disable_ps2_protocol` turns out to be a sufficient work-around of [MDEV-33864](https://jira.mariadb.org/browse/MDEV-33864), which I just filed for this.

vlad-lesin (approved):

dr-m:
There is also #3187 for fixing the cursor restoration.

PR #3172, opened 2024-04-02
MDEV-33585 The maximum innodb_log_buffer_size is too large

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33585*
## Description
On Microsoft Windows, `ReadFile()` as well as `WriteFile()` limit the size of the request to `DWORD`, which is 32 bits (at most 4 GiB - 1) also on 64-bit systems.

On FreeBSD, `sysctl debug.iosize_max_clamp` could limit the size of a write request to `INT_MAX`. The size of a read request is always limited to `INT_MAX`. This would allow the request size to be 4095 bytes more than the Linux limit (`0x7ffff000` according to `man 2 read` and `man 2 write`).

On OpenBSD, Solaris and possibly NetBSD, the read request size is limited to `SSIZE_T_MAX`, which would be half the current maximum `innodb_log_buffer_size`. This should be not much of an issue anyway, because on contemporary 64-bit platforms, the virtual addresses are limited to 48 bits.

IBM AIX documentation mentions `OFF_MAX` which would apply when a 64-bit application is running on a 32-bit kernel.

Let us declare `innodb_log_buffer_size` as 32-bit unsigned and make the maximum `0x7ffff000`, to be compatible with the least common denominator (Linux).

`log_t`: Define `buf_size`, `max_buf_free` as 32-bit and `next_checkpoint_no` as `byte`, so that on AMD64 we can fit `log_sys.latch` and `log_sys.log` in the same 64-byte cache line.

`mtr_t::commit_log()`, `mtr_t::commit_logger`: A part of `mtr_t::commit()` split into a separate function, so that we will not unnecessarily invoke `log_sys.get_write_target()` when running on a memory-mapped log file.

`SyncFileIO::execute()`: Assert that the size of a synchronous read or write request is limited to `innodb_log_buffer_size` or `innodb_sort_buffer_size`. The maximum `innodb_sort_buffer_size` already was 64 MiB, which is not a problem.
## Release Notes
The type of the start-up parameter `innodb_log_buffer_size` is changed to `UNSIGNED` and the maximum value to `0x7ffff000` bytes (2 GiB - 4 KiB), to be compatible with the least common denominator (Linux).
## How can this PR be tested?
This should be adequately covered by the regression test suite, when running it without any "fake PMEM" (the data directories are outside `/dev/shm` or the executables are not linked with `libpmem`).

When it comes to log _writes_ (not _reads_), the bug might be present in earlier versions.
## PR quality check


storage/innobase/include/log0log.h
```diff
@@ -33,6 +33,17 @@ Created 12/9/1995 Heikki Tuuri
...
+/* According to Linux "man 2 read" and "man 2 write" this is the
+maximum size of a read or write request, both on 32-bit and 64-bit
+systems.
+
+On FreeBSD, the limit is close to the Linux one, INT_MAX.
+
+On Microsoft Windows, the limit is 4 GiB - 1.
+
+On other systems, the limit typically is up to SSIZE_T_MAX. */
+static constexpr unsigned innodb_log_buffer_size_max= 0x7ffff000;
```
vaintroub, line 45: maybe name it single_io_max_size or something that better fits the comment, and put it into os0file.h, where SyncFileIO is defined? The fact. that log buffer is used for the IO, this can perhaps change in the future, however Linux limitation is  going to remain forever.

vaintroub (approved):

storage/innobase/include/log0log.h
dr-m: 0892e6d028f7b08da938defd9443ecbb45cda612 defines `constexpr unsigned os_file_request_size_max` and aliases it as `log_sys.buf_size_max`.

PR #3173, opened 2024-04-02
MDEV-32453  Bulk insert fails to apply when trigger does insert operation

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-32453*

## Description

Reason:
- InnoDB fails to apply the buffered insert operation if the
after insert trigger does change the same table. This behaviour
leads to empty table for the subsequent insert operation and server
abort.

Solution:
========
- InnoDB should apply buffered insert operation if the after insert
trigger changes the same table.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/insert_into_empty.test
+++ b/mysql-test/suite/innodb/t/insert_into_empty.test
@@ -495,3 +495,16 @@ DROP TABLE t1;
 CREATE TABLE t (a CHAR CHARACTER SET utf8) ENGINE=InnoDB ROW_FORMAT=REDUNDANT;
 INSERT t SELECT left(seq,1) FROM seq_1_to_43691;
 DROP TABLE t;
+
+--echo #
+--echo #  MDEV-32453  Bulk insert fails to apply when trigger
+--echo #                  does insert operation
+--echo #
+CREATE TABLE t(c INT)ENGINE=InnoDB;
+CREATE TRIGGER t2_ai AFTER INSERT ON t FOR EACH ROW SET @a:=(SELECT * FROM t);
+BEGIN;
+--error ER_SUBQUERY_NO_1_ROW
+INSERT INTO t VALUES (1),(1);
+COMMIT;
+DROP TABLE t;
+--echo # End of 10.11 tests
--- a/storage/innobase/include/trx0trx.h
+++ b/storage/innobase/include/trx0trx.h
@@ -1188,10 +1188,16 @@ public:
     return UNIV_UNLIKELY(bulk_insert) ? bulk_insert_apply_low(): DB_SUCCESS;
   }
 
+  /** Do the bulk insert for the buffered insert operation of a table.
+  @param table bulk insert operation
+  @return DB_SUCCESS or error code. */
+  dberr_t bulk_insert_apply_for_table(dict_table_t *table);
 private:
   /** Apply the buffered bulk inserts. */
   dberr_t bulk_insert_apply_low();
 
+  /** Rollback the bulk insert operation for the transaction */
+  void bulk_rollback_low();
   /** Assign a rollback segment for modifying temporary tables.
   @return the assigned rollback segment */
   trx_rseg_t *assign_temp_rseg();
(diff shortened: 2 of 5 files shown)


mysql-test/suite/innodb/t/insert_into_empty.test
```diff
@@ -495,3 +495,16 @@ DROP TABLE t1;
...
+--echo #                  does insert operation
+--echo #
+SET AUTOCOMMIT=0;
+CREATE TABLE t(c INT)ENGINE=InnoDB;
+CREATE TRIGGER t2_ai AFTER INSERT ON t FOR EACH ROW SET @a:=(SELECT * FROM t);
+INSERT INTO t VALUES (1);
+--error ER_SUBQUERY_NO_1_ROW
+INSERT INTO t VALUES (1);
+DROP TABLE t;
+SET AUTOCOMMIT=1;
```
dlenski, line 510: This test is demonstrating the _correct_ operation _after_ the bug has been fixed, and **not** the _incorrect_ operation (→ crash) _before_ the bug has been fixed.

Right?

Perhaps you can clarify this in the description in the commit and PR.

storage/innobase/row/row0merge.cc
```diff
@@ -5376,5 +5366,28 @@ dberr_t trx_t::bulk_insert_apply_low()
...
+  ut_ad(bulk_insert);
+  dberr_t err;
+  for (auto& t : mod_tables)
+    if (t.second.is_bulk_insert())
+      if ((err= t.second.write_bulk(t.first, this)) != DB_SUCCESS)
+        goto err_exit;
+  return DB_SUCCESS;
+err_exit:
+  bulk_rollback_low();
   return err;
```
dr-m, line 5392: I’d reduce the scope of the variable like this:
```c++
      if (dberr_t err= t.second.write_bulk(t.first, this))
      {
        bulk_rollback_low();
        return err;
      }
  return DB_SUCCESS;

File: mysql-test/suite/innodb/t/insert_into_empty.test [hunk 3]
```diff
@@ -495,3 +495,16 @@ DROP TABLE t1;
...
+--echo #
+--echo #  MDEV-32453  Bulk insert fails to apply when trigger
+--echo #                  does insert operation
+--echo #
+SET AUTOCOMMIT=0;
+CREATE TABLE t(c INT)ENGINE=InnoDB;
+CREATE TRIGGER t2_ai AFTER INSERT ON t FOR EACH ROW SET @a:=(SELECT * FROM t);
+INSERT INTO t VALUES (1);
+--error ER_SUBQUERY_NO_1_ROW
+INSERT INTO t VALUES (1);
```
dr-m, line 508: Instead of changing a session variable, I’d use ordinary `BEGIN` and `COMMIT` around the two `INSERT` statements if this bug is repeatable with those as well. Alternatively, maybe this would be reproducible with a two-row `INSERT` in autocommit mode:
```sql
--error ER_SUBQUERY_NO_1_ROW
INSERT INTO t VALUES(1),(1);
```
I would also add `--echo # End of 10.11 tests` to the end of this file.

mysql-test/suite/innodb/t/insert_into_empty.test
dr-m: Normally, if the code fix is reverted, the regression test is expected to crash or fail. If that is not the case here, it should be clarified in the commit message. I did not check that yet.

storage/innobase/include/trx0trx.h
```diff
@@ -1188,10 +1188,17 @@ struct trx_t : ilist_node<>
     return UNIV_UNLIKELY(bulk_insert) ? bulk_insert_apply_low(): DB_SUCCESS;
   }
 
+  /** Do the bulk insert for the buffered insert operation of
+  a table.
```
dr-m, line 1192: There should be no reason to split this line.

dr-m (approved):
This looks OK to me. I am happy to see that `trx_t::check_bulk_buffer()` will avoid the traversal of `trx_t::mod_tables` if bulk insert is not active.

PR #3174, opened 2024-04-02
MDEV-33782 Implement WRITE_TO_BINLOG_ENABLE option for events

andremralves:
- [x] *The Jira issue number for this PR is: MDEV-33782*

## Description

Implement a new option for events, `WRITE_TO_BINLOG_ENABLE`. If this option is set, events will be replicated with `Event:status = ENABLE`, otherwise, the event status will be `DISABLED`. Thanks @andrelkin for the idea.

Here is a table showing some examples:

|  Sql command |  Event Status on Master |  Event Status on Slave |
|---|---|---|
| create event .. ENABLE ..   |  Enabled |  Disabled |
| create event .. DISABLE .. |  Disabled |  Disabled |
| create event .. ENABLE WRITE_TO_BINLOG_ENABLE .. |  Enabled |  Enabled |
| create event .. DISABLE WRITE_TO_BINLOG_ENABLE .. |  Disabled |  Enabled |

Previously, events were replicated with the `SLAVESIDE_DISABLED` status. With this patch they will be set either to `DISABLED` or `ENABLED` --if `WRITE_TO_BINLOG_ENABLE` is defined.

## Release Notes
Implement a new option for events, `WRITE_TO_BINLOG_ENABLE`. If this option is set, events will be replicated with `Event:status = ENABLE`, otherwise, the event status will be `DISABLED`. Replicated events will not be set to `SLAVESIDE_DISABLED` anymore. 

## How can this PR be tested?

```
./mtr rpl.rpl_create_drop_event 
```

It can also be tested with manual tests setting up a replication.

## PR quality check


HugoWenTD (commented):
In my opinion, `DISABLE ON SLAVE` and `ENABLE ON SLAVE` seem to suggest that they are specific options for the replica (formerly known as the slave) only. However, the existing `DISABLE ON SLAVE` command actually disables the feature on both the source (formerly known as the master) and the replica. MySQL actually has a bug report for it. https://bugs.mysql.com/bug.php?id=31643

On the other hand, if based on current behavior, new `ENABLE ON SLAVE` seems to do the right thing as the opposite to `DISABLE ON SLAVE`.

---
By the way, better to add some descriptions in the commit message for example why the change is needed.

sql/event_data_objects.cc
```diff
@@ -1249,6 +1252,8 @@ Event_timed::get_create_event(THD *thd, String *buf)
 
   if (status == Event_parse_data::ENABLED)
     buf->append(STRING_WITH_LEN("ENABLE"));
+  else if (status == Event_parse_data::SLAVESIDE_ENABLED)
+    buf->append(STRING_WITH_LEN("ENABLE"));
```
LinuxJedi, line 1256: Shouldn't this be "ENABLE ON SLAVE"?

LinuxJedi (changes requested):
Nice work!

It appears this is failing a test when embedded server is built, `funcs_1.is_columns_mysql_embedded`, could you please fix this too?

sql/event_data_objects.cc
andremralves: You are right! I missed this one. Thanks.

andremralves:
> It appears this is failing a test when embedded server is built, `funcs_1.is_columns_mysql_embedded`, could you please fix this too?
I'm working on it.

LinuxJedi (approved):
Many thanks for fixing that. I'll try to get someone from the replication team to give it the next stage of review.

andrelkin:
@andremralves salute!
I started reviewing the top-level of this work, to have noticed something interesting I believe.
Please read on that on [Jira](https://jira.mariadb.org/browse/MDEV-33782?focusedCommentId=289509&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-289509) and preferably reply on there too (that seems easier to engage the feature reporter).

Thanks for this work for far!

Andrei

sql/sql_yacc.yy
```diff
@@ -2982,6 +2982,12 @@ opt_ev_status:
             Lex->event_parse_data->status_changed= true;
             $$= 1;
           }
+        | ENABLE_SYM ON SLAVE
```
andrelkin, line 2985: Please read a discussion on [Jira|https://jira.mariadb.org/browse/MDEV-33782?focusedCommentId=289601&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-289601] about how to express in SQL way the desired effect.

This patch does address the FR, but it complicates earlier dubious design decision still.  As the FR requires to achieve on slave  `Event::status` as `ENABLE`  a straightforward way to help it is to encode such intent as a new option which would be merely for logging on master and unlogging on slave. That is it need not to affect/expand `Event::status`.

We could even leave the current name of `SLAVESIDE_ENABLED` but not regard it the way it's done in the current patch.
Also consider `WRITE_TO_BINLOG_ENABLE` (somewhat this naming would follow an existing `NO_WRITE_TO_BINLOG`) instead.

When it's specified ...

sql/sql_yacc.yy
```diff
@@ -2982,6 +2982,12 @@ opt_ev_status:
...
+          {
+            Lex->event_parse_data->status= Event_parse_data::SLAVESIDE_ENABLED;
```
andrelkin, line 2987: `status` would be set on slave straight to `ENABLE` unless the event is CREATE-ed with `DISABLE` status. 
Logics inside `Event_parse_data::check_originator_id()` would need some adjustment, which is to disable the event on slave 
```
status= Event_parse_data::DISABLED
```
without the new **logging** option.

The status of event on any server then would flip between just two `ENABLE` and `DISABLE`.
I think the slave should preserve the new option in its binlog when one is configured on the slave.

andrelkin (changes requested):
Hello Andre!

Thanks for a good effort to satisfy to the feature request. 
I am explaining why we could and should do it better though.

Specifically I think we should treat the new syntax as a sub-clause to control logging/unlogging.
Inevitably we would eradicate the old `Event_parse_data::SLAVESIDE_DISABLED` from `Event::status`.

I hope the measures will make sense for you as well.

Cheers,
Andrei

andremralves:
Hi Andrei,

Thanks for your feedback.

I will read more carefully the discussion on jira and then respond here and there.

andrelkin:
Salve @andremralves!

I wonder about your plan and timetable to complete this PR?
Could you please share with us if any help is need for that.

Happy NY on few calendars :-)!

Andrei

andremralves:
Salve @andrelkin !

I got caught up with other things and honestly forgot about this issue—sorry for the delay! I’ll try to finish it over the weekend, and if I can’t, I’ll update you all on how I plan to proceed.

Thanks for checking in, and happy NY on those calendars too!

Cheers,
André

sql/sql_yacc.yy
andremralves: Hi @andrelkin, you also mentioned `create event .. log_as [DISABLE|ENABLE]` in [Jira](https://jira.mariadb.org/browse/MDEV-33782?focusedCommentId=289562&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-289562), which I believe have the same purpose of `WRITE_TO_BINLOG_ENABLE`, but the latter one is just a more polished idea, right?

So basically, if `WRITE_TO_BINLOG_ENABLE` is set we replicate as ENABLED otherwise DISABLED. I made a table to make it easier to visualize:

|  Options |  Master Status |  Slave Status |
|---|---|---|
| status: ENABLED   |  Enabled |  Disabled |
| status: DISABLED |  Disabled |  Disabled |
| status: ENABLED + WRITE_TO_BINLOG_ENABLE |  Enabled |  Enabled |

Now I see that `Event:status` would be reduced to only ENABLED/DISABLED, but what should we do to `SLAVESIDE_DISABLED`? Remove it? Change it's purpose?

andrelkin: Howdy @andremralves !

To make it clear your option's syntax - just a single word `WRITE_TO_BINLOG_ENABLE` looks fine to me.
The patch therefore only lacks SLAVESIDE_DISABLED removal.

bnestere:
@andremralves, hi! It looks like @andrelkin made one final suggestion ([link](https://github.com/MariaDB/server/pull/3174#discussion_r1951246846)) to your patch before it is ready to go. Are you still planning on working on this?

gkodinov (changes requested):
Hello, this is a preliminary review. I really hope that you're still inclined to work on this with us. 
As a first step, can you pleaser make sure your commit has a message that complies with MariaDB coding guidelines? 

Once we're past the basics (commit message, build bot tests passing etc.) I will solicit some review from the replication developers on your behalf.

andrelkin:
@andremralves, thanks for finding the time and motivation to see this through to completion :-)!
Cc-ing here @gkodinov, @bnestere.

The latest patch follows [the plan](https://github.com/MariaDB/server/pull/3174#discussion_r1938584545) to implement 

|  Options |  Master Status |  Slave Status |
|---|---|---|
| status: ENABLED   |  Enabled |  Disabled |
| status: DISABLED |  Disabled |  Disabled |
| status: ENABLED + WRITE_TO_BINLOG_ENABLE |  Enabled |  Enabled |
| status: DISABLED + WRITE_TO_BINLOG_ENABLE |  Disabled |  Enabled |

which is great, **as well as** it preserves brackward compatibility of  OLD -> NEW replication, [mentioned](https://jira.mariadb.org/browse/MDEV-33782?focusedCommentId=305985&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-305985) by Brandon

| status: X + [DISABLE ON SLAVE](https://mariadb.com/docs/server/reference/sql-statements/data-definition/create/create-event) |  X |  Disabled |

But it also continues to support the legacy DISABLE ON SLAVE on **master**. 
I suggest to deprecate it, as we can't remove it w/o this phase, so that the NEW (this feature aware) master  would throw the warning.

Sure it'd be great to have the table in the commit message.

gkodinov:
there has been no reply to my review comments for 3+ weeks now. Converting to draft. Please re-open when/if you intend to resume working on this.

PR #3176, opened 2024-04-03
MDEV-32397, MDEV-32403 Crashes during join processing.

DaveGosselin-MariaDB:
Queries having the following form may cause a crash
  SELECT t1.a FROM ( SELECT a AS a1 FROM t1 ) dt
  JOIN t1 ON a1 LIKE EXISTS ( SELECT a + RAND () FROM t1 UNION SELECT a FROM t1);
because the table t1 has some JOIN cleanup operations performed prematurely during the subselect.

In this particular case, the presence of RAND() makes the subquery uncacheable, necessitating the need to execute the subquery multiple times during join record evaluation.  Each time the subquery runs, it creates its own JOIN structure which has references to the table t1. When the subquery completes, JOIN::cleanup and functions called by it result in ha_end_keyread() being called on table t1.  However, we are not done with table t1 because the upper level `select t1.a from...` query requires table t1 to be open.  To solve this, we make the executor aware of when we're in subqueries like this and delay JOIN cleanup until the end of the query.


spetrunia (changes requested):
First comment here: 
https://jira.mariadb.org/browse/MDEV-32403?focusedCommentId=284534&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-284534

will follow up in the MDEV.

spetrunia:
... see the suggestion in : https://jira.mariadb.org/browse/MDEV-32403?focusedCommentId=284537&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-284537

mysql-test/main/subselect_like_exists.result
```diff
@@ -0,0 +1,185 @@
...
+c37
+NULL
+NULL
+NULL
+NULL
+NULL
+NULL
+CREATE TABLE t1 ( a TEXT, b bool, index (a)) ;
+Warnings:
+Note	1071	Specified key was too long; max key length is 1000 bytes
```
mariadb-OlegSmirnov: Is there a way to modify the test case to avoid this warning?

DaveGosselin-MariaDB: Yes, it should be possible, I will take a look.

With respect to the original test case, I include it because it gives great confidence to the bug reporter that we (1) fixed the original problem and (2) we are accurate in reducing to a minimal reproducible case.  I'm sorry but I don't see how it makes sense to exclude the original test case, regardless of how it was generated.

mariadb-OlegSmirnov: I am not sure bug reporters verify what we add to our test suite, I believe they just want their bug fixed. Of course, if we are not sure that our minimal case is reduced accurately, it makes sense to add the original one. However, that should to be evaluated what  value it brings to the development process. I think the main purpose of the test suite is to avoid regressions. So what should a developer do if such a fuzzer query fails? It is not available for debugging in its original state. The developer will have to reduce to something debuggable anyway. From such a point of view adding convoluted queries to the test suite does not bring much value.

mariadb-OlegSmirnov: Another idea to consider: such a complicated query may have more than one bug present, so it may start failing due to reasons besides the bug fixed. That can also cause additional confusion. We almost don't use unit tests, so I believe SQL test cases should serve as some replacement for them,  being concise and allowing to isolate the exact problem

DaveGosselin-MariaDB: > So what should a developer do if such a fuzzer query fails? It is not available for debugging in its original state.

It doesn't need to be, because the ticket (in this case) has all of the state needed to reproduce the problem.

> The developer will have to reduce to something debuggable anyway. 

Yes, but I disagree that the original should not be included.  Many of our users are technical and will care.  Additionally, future maintenance developers will not want to go through the effort to confirm that the minimal example matches behavior with the user-reported one.

> such a complicated query may have more than one bug present, so it may start failing due to reasons besides the bug fixed. That can also cause additional confusion.

Agreed, but that doesn't mean we shouldn't exclude the original.  Include the original, plus minimal reproduction cases for each bug (as needed).

sql/sql_lex.cc
```diff
@@ -4993,6 +4993,19 @@ bool st_select_lex::optimize_unflattened_subqueries(bool const_only)
...
 
+      /*
+        If any one SELECT in the subquery has UNCACHEABLE_RAND, then all
+        SELECTs should be marked as uncacheable.
+      */
+      bool has_rand= false;
+      for (SELECT_LEX *sl= un->first_select(); sl && !has_rand;
+           sl= sl->next_select())
+        has_rand= sl->uncacheable & UNCACHEABLE_RAND;
+      if (has_rand) {
```
mariadb-OlegSmirnov: Nit: formatting (brace should start from a new line).

DaveGosselin-MariaDB: Good catch, fixed.

Olernov:
What do you think about my previous suggestion for a more descriptive name for the test file, for example `subselect_union_rand.test`?

DaveGosselin-MariaDB:
> What do you think about my previous suggestion for a more descriptive name for the test file, for example `subselect_union_rand.test`?

Sorry, I forgot to make that change, yes I quite like it and will rename the file.  Thank you!

sql/sql_lex.cc
```diff
@@ -4993,6 +4993,20 @@ bool st_select_lex::optimize_unflattened_subqueries(bool const_only)
...
         subquery_predicate->no_rows_in_result();
 
+      /*
+        If any one SELECT in the subquery has UNCACHEABLE_RAND, then all
+        SELECTs should be marked as uncacheable.
+      */
+      bool has_rand= false;
+      for (SELECT_LEX *sl= un->first_select(); sl && !has_rand;
+           sl= sl->next_select())
+        has_rand= sl->uncacheable & UNCACHEABLE_RAND;
```
spetrunia, line 5081: Can we just take the value of `(un->uncacheable & UNCACHEABLE_RAND)`   ? 
If we can't say so, I'd add an assert that `has_rand` and the unit's value agree.

spetrunia:
Then, this needs to be made on top of 10.11.  That's all the input.

sql/sql_lex.cc
DaveGosselin-MariaDB: We can't take the value of `(un->cacheable & UNCACHEABLE_RAND)` nor assert that they agree.  If I add an assertion like 
```
5001       for (SELECT_LEX *sl= un->first_select(); sl && !has_rand;
5002            sl= sl->next_select())
5003       {
5004         has_rand= sl->uncacheable & UNCACHEABLE_RAND;
5005       }
5006       DBUG_ASSERT(has_rand ==
5007                   static_cast<bool>(un->uncacheable & UNCACHEABLE_RAND));
```
it evaluates `false` under the following test case (10.5, 10.11):
```
create table t1 (a int);
create table t2 select * from t1 where 1 in (select 1 union select 2 union select RAND());
```
I wonder if it's a bug that `(un->cacheable & UNCACHEABLE_RAND)` is not set in this case; here are two cases that surprisingly don't trigger the assertion (10.5 - I haven't yet tried 10.11 with these and RAND()):
```
create table t2 select * from t1 where 1 in (select 1 union select UUID());
create table t2 select * from t1 where 1 in (select 2 union select UUID());
```

DaveGosselin-MariaDB: (the above cases are for 10.5, the original target for this bug)

spetrunia: Indeed, it's a bug in the parser.
There's no problem with this query:
```sql
select * from t1 where 1 in (select 1 union select RAND());
```
while this query does have it:
```sql
select * from t1 where 1 in (select 1 union select 2 union select RAND());
```
adding more parts to the UNION:

```sql
select * from t1 where 1 in (select 1 union select 2 union select RAND() union select 5 union select 7);
```
looks like they would have the same problem, too..

spetrunia: The difference starts here:
```
(gdb) wher 10
  #0  st_select_lex_unit::register_select_chain (this=0x7fff1c017e28, first_sel=0x7fff1c0178c0) at /home/psergey/dev-git2/10.11-look/sql/sql_lex.cc:10117
  #1  0x0000555555f652d2 in LEX::create_unit (this=0x7fff1c005168, first_sel=0x7fff1c0178c0) at /home/psergey/dev-git2/10.11-look/sql/sql_lex.cc:6275
  #2  0x0000555555f74c17 in LEX::parsed_select_expr_start (this=0x7fff1c005168, s1=0x7fff1c0178c0, s2=0x7fff1c018658, unit_type=UNION_TYPE, distinct=true) at /home/psergey/dev-git2/10.11-look/sql/sql_lex.cc:10388
  #3  0x0000555555f74f81 in LEX::add_primary_to_query_expression_body (this=0x7fff1c005168, unit=0x7fff1c017e28, sel=0x7fff1c018658, unit_type=UNION_TYPE, distinct=true,            oracle=false) at /home/psergey/dev-git2/10.11-look/sql/sql_lex.cc:10458
  #4  0x0000555555f75017 in LEX::add_primary_to_query_expression_body (this=0x7fff1c005168, unit=0x7fff1c017e28, sel=0x7fff1c018658, unit_type=UNION_TYPE, distinct=true) at /home/  psergey/dev-git2/10.11-look/sql/sql_lex.cc:10473
  #5  0x00005555562c2136 in MYSQLparse (thd=0x7fff1c000d78) at /home/psergey/dev-git2/10.11-look/sql/sql_yacc.yy:8774
```

in LEX::add_primary_to_query_expression_body() there are two codepaths:
```cpp
    if (!sel1->next_select())
      unit= parsed_select_expr_start(sel1, sel2, unit_type, distinct);
    else      
      unit= parsed_select_expr_cont(unit, sel2, unit_type, distinct, oracle);
```
if-branch is taken for the first two UNION members.  else-branch is for the rest.

`parsed_select_expr_cont()` calls `create_unit()`  (which doesn't really create it for the second member),  which calls

```cpp
  unit->register_select_chain(first_sel);
```
which propagates the `uncacheable` flag from UNION members to the SELECT_LEX_UNIT object.

`LEX::parsed_select_expr_cont()` just doesn't do this.

spetrunia: We should put this into a non-stable release:
```diff
diff --git a/sql/sql_lex.cc b/sql/sql_lex.cc
index 2edb1e26d43..eac495a3081 100644
--- a/sql/sql_lex.cc
+++ b/sql/sql_lex.cc
@@ -10398,6 +10398,7 @@ SELECT_LEX_UNIT *LEX::parsed_select_expr_cont(SELECT_LEX_UNIT *unit,
   }
   last->link_neighbour(sel1);
   sel1->set_master_unit(unit);
+    unit->uncacheable |= sel1->uncacheable;
   sel1->set_linkage_and_distinct(unit_type, distinct);
   unit->pre_last_parse= last;
   return unit;
```

spetrunia (approved):
Ok to push this patch.

Let's address the suggestion for `unit->uncacheable` outside of this patch (as mentioned in the MDEV)

sql/sql_lex.cc
DaveGosselin-MariaDB: Created https://jira.mariadb.org/browse/MDEV-37905 to track this.

PR #3177, opened 2024-04-04
MDEV-30732 : UBSAN: null pointer passed as argument 2, which is decla…

janlindstrom:
…red to never be null in storage/innobase/handler/ha_innodb.cc

- [x] *The Jira issue number for this PR is: MDEV-30732*

## Description

Problem was that row_mysql_read_blob_ref can return NULL in case when blob datatype is used in a key and its real value is NULL. This NULL pointer is then used in memcpy function in wsrep_store_key_val_for_row. However,
memcpy is defined so that argument 2 must not be NULL.

Fixed by adding conditions before memcpy functions so that argument 2 is always non NULL.

## Release Notes

## How can this PR be tested?

## PR quality check


anson1014:
Minor nits: 
- some unnecessary whitespace
- consider removing the brackets for your single line if statements in order to adhere to: https://github.com/MariaDB/server/blob/11.5/CODING_STANDARDS.md

anson1014 (approved):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7263,7 +7266,10 @@ wsrep_store_key_val_for_row(
 				true_len = key_len;
 			}
 
-			memcpy(sorted, blob_data, true_len);
+			if (blob_data) {
+				memcpy(sorted, blob_data, true_len);
+			}
```
dr-m, line 7271: If the pointer is `nullptr`, I think that you’d better assert that the length is 0. In any other case, we would be operating in some garbage-in/garbage-out mode, not initializing all of the `sorted` buffer.

janlindstrom:
10.4 is EOL, I will submit to a later release.

mysql-test/suite/galera/t/MDEV-30732.test
```diff
@@ -0,0 +1,18 @@
...
+CREATE TABLE t (c INT,c2 BLOB NOT NULL,KEY k2 (c2 (1),c)) DEFAULT CHARSET=latin1 ROW_FORMAT=COMPACT;
+INSERT INTO t (c) VALUES (1),(1),(1),(1),(1);
+SELECT * FROM t;
+DROP TABLE t;
+CREATE TABLE t (c INT,c2 VARCHAR(270) NOT NULL,KEY k2 (c2 (1),c)) DEFAULT CHARSET=latin1 ROW_FORMAT=COMPACT;
+INSERT INTO t (c) VALUES (1),(1),(1),(1),(1);
+SELECT * FROM t;
+DROP TABLE t;
+CREATE TABLE t (c INT,c2 CHAR(80) NOT NULL,KEY k2 (c2 (1),c)) DEFAULT CHARSET=latin1 ROW_FORMAT=COMPACT;
+INSERT INTO t (c) VALUES (1),(1),(1),(1),(1);
```
sjaakola, line 14: this table and all tables used in this test, do not have a primary key nor a unique key. wsrep should therefore use a hash of the full row as they replication write set key definition.
It is ok, to test tables without proper key definition, but add also a test case for primary/unique key in the test scenario

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7343,7 +7349,11 @@ wsrep_store_key_val_for_row(
 								cs->mbmaxlen),
 							&error);
 				}
-				memcpy(sorted, src_start, true_len);
+
+				if (src_start) {
+					memcpy(sorted, src_start, true_len);
+				}
+
```
sjaakola, line 7356: same applies to this, as the Marko's comment above, assert is better than circumventing the potential mismatch of buffer pointer and buffer length. 
Here, the src_start refers to function parameter record, i suppose it should not be nullptr in any case

sjaakola (commented):
2 change requests posted

storage/innobase/handler/ha_innodb.cc
janlindstrom: When true_len == 0 blob_data is not always nullptr it may be empty string i.e. "", therefore can't add assertion for nullptr.

mysql-test/suite/galera/t/MDEV-30732.test
janlindstrom: Added tests for primary key.

storage/innobase/handler/ha_innodb.cc
janlindstrom: Added assertion.

PR #3178, opened 2024-04-04
MDEV-33757 Get rid of TrxUndoRsegs code

vlad-lesin:
Post-push fix: purge queue array can't be fixed size, because the elements of the array is the analogue of undo logs, which must be processed in the order of transaction commits, and the array can contain more elements, than trx_sys.rseg_array. Also it's necessary to maintain min-heap property by the trx_no of transaction, which produced the first non-purged undo log in all rsegs. That's why the element of purge queue aray must contain not only trx_sys.rseg_array index, but also trx_no of committed transacion, i.e. the pair (trx_no, trx_sys.rseg_array index), which is encoded as uint64_t((trx_no << 8) | (trx_sys.rseg_array index)).

Reviewed by: Marko Mäkelä

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/include/trx0purge.h
+++ b/storage/innobase/include/trx0purge.h
@@ -58,62 +58,47 @@ ulint trx_purge(ulint n_tasks, ulint history_size);
 /** The control structure used in the purge operation */
 class purge_sys_t
 {
-  /** Min-heap based priority queue over fixed size array */
+  /** Min-heap based priority queue of (trx_no, trx_sys.rseg_array index)
+  pairs, ordered on trx_no. The highest 64-TRX_NO_SHIFT bits of each element is
+  trx_no, the lowest 8 bits is rseg's index in trx_sys.rseg_array. */
   class purge_queue
   {
-    /** Array of indexes in trx_sys.rseg_array. */
-    alignas(CPU_LEVEL1_DCACHE_LINESIZE) byte m_array[TRX_SYS_N_RSEGS];
-    /** Pointer to the end of m_array. */
-    byte *m_end= m_array;
-
   public:
-    struct trx_rseg_cmp
-    {
-      /** Compare two trx_rseg_t* based on trx_no.
-      @param lhs first index in trx_sys.rseg_array to compare
-      @param rhs second index in trx_sys.rseg_array to compare
-      @return whether lhs>rhs */
-      bool operator()(const byte lhs, const byte rhs)
-      {
-        ut_ad(lhs < TRX_SYS_N_RSEGS);
-        ut_ad(rhs < TRX_SYS_N_RSEGS);
-        /* We can compare without trx_rseg_t::latch, because rseg last
-        commit is always set before pushing rseg to purge queue. */
-        return trx_sys.rseg_array[lhs].last_commit_and_offset >
-               trx_sys.rseg_array[rhs].last_commit_and_offset;
-      }
-    };
-    byte *begin() { return m_array; }
-    byte *end() { return m_end; }
-    const byte *c_begin() const { return m_array; }
-    const byte *c_end() const { return m_end; }
-    size_t size() const
-    {
-      size_t s= c_end() - c_begin();
-      ut_ad(s <= TRX_SYS_N_RSEGS);
-      return s;
-    }
-    bool empty() const { return !size(); }
-    void clear() { m_end= m_array; }
+    typedef std::vector<uint64_t, ut_allocator<uint64_t>> container_type;
+    /** Number of bits reseved to shift trx_no in purge queue element */
+    static constexpr unsigned TRX_NO_SHIFT= 8;
 
-    /** Push index of trx_sys.rseg_array into min-heap.
-    @param i index to push */
-    void push_rseg_index(byte i)
+    bool empty() const { return m_array.empty(); }
+    void clear() { m_array.clear(); }
+
+    /** Push (trx_no, trx_sys.rseg_array index) into min-heap.
+    @param trx_no_rseg (trx_no << TRX_NO_SHIFT | (trx_sys.rseg_array index)) */
+    void push_trx_no_rseg(container_type::value_type trx_no_rseg)
     {
-      ut_ad(i < TRX_SYS_N_RSEGS);
-      ut_ad(size() + 1 <= TRX_SYS_N_RSEGS);
-      *m_end++= i;
-      std::push_heap(begin(), end(), trx_rseg_cmp());
+      m_array.push_back(trx_no_rseg);
+      std::push_heap(m_array.begin(), m_array.end(),
+                     std::greater<container_type::value_type>());
     }
 
     /** Push rseg to priority queue.
-    @param rseg trx_rseg_t pointer to push */
-    void push(const trx_rseg_t *rseg)
+    @param trx_no trx_no of committed transaction
+    @param rseg rseg of committed transaction*/
+    void push(trx_id_t trx_no, const trx_rseg_t &rseg)
     {
-      ut_ad(rseg >= trx_sys.rseg_array);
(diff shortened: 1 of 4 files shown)


storage/innobase/include/trx0purge.h
```diff
@@ -58,76 +58,64 @@ ulint trx_purge(ulint n_tasks, ulint history_size);
...
+    @param rseg rseg of committed transaction*/
+    void push(trx_id_t trx_no, const trx_rseg_t &rseg)
     {
-      ut_ad(rseg >= trx_sys.rseg_array);
-      ut_ad(rseg < trx_sys.rseg_array + TRX_SYS_N_RSEGS);
-      byte i= byte(rseg - trx_sys.rseg_array);
-      push_rseg_index(i);
+      ut_ad(&rseg >= trx_sys.rseg_array);
+      ut_ad(&rseg < trx_sys.rseg_array + TRX_SYS_N_RSEGS);
+      push_trx_no_rseg(trx_no << 8 | byte(&rseg - trx_sys.rseg_array));
```
dr-m, line 90: We could also add
```c++
ut_ad(trx_no < 1ULL << 48);
```
or (if the symbol is available without adding `#include "data0type.h"`)
```c++
ut_ad(trx_no < 1ULL << (DATA_TRX_ID_LEN * CHAR_BIT));
```

storage/innobase/include/trx0purge.h
```diff
@@ -58,76 +58,64 @@ ulint trx_purge(ulint n_tasks, ulint history_size);
...
+    }
+
+    /** Extracts rseg from (trx_no, trx_sys.rseg_array index) pair.
+    @param trx_no_rseg (trx_no << 8 | (trx_sys.rseg_array index)
+    @return pointer to rseg in trx_sys.rseg_array */
+    static trx_rseg_t *rseg(container_type::value_type trx_no_rseg) {
+      byte i= trx_no_rseg & ((1ULL << 8) - 1);
```
dr-m, line 97: This would be written simpler just `byte i= byte(trx_no_rseg);`

I would not repeat the `trx_no << 8 | …` in multiple places. Maybe, introduce `static constexpr unsigned TRX_NO_SHIFT= 8` and use it where applicable. It might be more efficient to reserve 16 bits for the 7-bit index: I’d compare the generated code on IA-32.

storage/innobase/include/trx0purge.h
```diff
@@ -243,22 +231,25 @@ class purge_sys_t
 					record */
 	uint16_t	hdr_offset;	/*!< Header byte offset on the page */
 
-  /** Binary min-heap of indexes in trx_sys.rseg_array, ordered on
-  rseg_t::last_trx_no(). It is protected by the pq_mutex */
+  /** Binary min-heap of (trx_no, trx_sys.rseg_array index) pairs, ordered on
+  trx_no. The highest 56 bits of each element is trx_no, the lowerst 8 bits -
+  index in trx_sys.rseg_array. It is protected by the pq_mutex */
   purge_queue purge_queue;
```
dr-m, line 237: I would omit the implementation detail about the `container_type::value_type` layout at this level.

storage/innobase/include/trx0rseg.h
```diff
@@ -170,21 +170,19 @@ struct alignas(CPU_LEVEL1_DCACHE_LINESIZE) trx_rseg_t
...
-  {
-    return static_cast<uint16_t>(last_commit_and_offset & ((1ULL << 16) - 1));
-  }
+  { return static_cast<uint16_t>(last_commit_and_offset >> 48); }
 
   void set_last_commit(uint16_t last_offset, trx_id_t trx_no)
   {
-    last_commit_and_offset= trx_no << 16 | static_cast<uint64_t>(last_offset);
+    last_commit_and_offset= static_cast<uint64_t>(last_offset) << 48 | trx_no;
   }
```
dr-m, line 186: I think that this change in #3152 made sense. I would not revert this.

But I would like to point out that the bit-masking in `last_offset()` is redundant and that we can use a widening cast:
```c++
uint16_t last_offset() const { return uint16_t(last_commit_and_offset); }
void set_last_commit(uint16_t last_offset, trx_id_t trx_no)
{
  last_commit_and_offset= trx_no << 16 | uint64_t{last_offset};
}
```

storage/innobase/trx/trx0purge.cc
```diff
@@ -495,14 +495,13 @@ inline dberr_t purge_sys_t::iterator::free_history_rseg(trx_rseg_t &rseg) const
 
 void purge_sys_t::cleanse_purge_queue(const fil_space_t &space)
 {
-  byte purge_elem_list[TRX_SYS_N_RSEGS];
+  purge_queue::container_type purge_elem_list;
   mysql_mutex_lock(&pq_mutex);
-  std::copy(purge_queue.c_begin(), purge_queue.c_end(), purge_elem_list);
-  byte *purge_list_end = purge_elem_list + purge_queue.size();
+  std::copy(purge_queue.cbegin(), purge_queue.cend(), purge_elem_list.begin());
   purge_queue.clear();
```
dr-m, line 501: Now that we have to use a variable-size heap-allocated container, could we just use the copy constructor?
```c++
purge_queue::container_type purge_elem_list(purge_queue.m_array);
```
I see that the data member is `private`, so this probably would not work directly as is. Could this be returned (with copy elision) by a member function?

Maybe, a cleaner way would be to make this a member function of `purge_queue` itself, just adding `mysql_mutex_assert_owner(&purge_sys.pq_mutex);` in the beginning? Everything except the mutex are private to `purge_queue`. Perhaps in that way, we could also remove the member functions `c_begin()` and `c_end()`?

dr-m (commented):
Thank you for noticing the problem in #3152. It is a bit scary that this was not caught in tests, not even by the MDEV-24402 `CHECK TABLE…EXTENDED`.

The algorithm looks sound to me. I have some minor comments about some details.

storage/innobase/include/trx0purge.h
vlad-lesin: Both 8 and 16 bit shifts are compiled to  `shld` and `shl` instructions for IA-32:
```
0x80ae1259 <trx_t::commit_low(mtr_t*)+1865>     shld   $0x8,%ecx,%ebx           
0x80ae125d <trx_t::commit_low(mtr_t*)+1869>     shl    $0x8,%ecx
```

storage/innobase/include/trx0purge.h
```diff
@@ -243,20 +238,29 @@ class purge_sys_t
...
-  {
+  void enqueue(const trx_rseg_t &rseg) { enqueue(rseg.last_trx_no(), rseg); }
+
+  /** Clone purge queue container.
+  @return purge queue container clone */
+  purge_queue::container_type clone_queue_container() const {
     mysql_mutex_assert_owner(&pq_mutex);
-    purge_queue.push(&rseg);
+    return purge_queue.clone_container();
   }
```
dr-m, line 264: Based on the following test program that I checked on https://godbolt.org, this second level of copy propagation will indeed be elided starting with GCC 4.8.5 `-O1`:
```c++
#include <vector>
typedef std::vector<unsigned> vect;
static vect clone(const vect &v) { return v; }
static vect klone(const vect &v) { return clone(v); }
bool f(const vect &v)
{
  vect m = klone(v);
  return m.empty();
}
```
If I remove the `static` (which encourages the compiler to `inline` the calls), then the function `klone()` does not reduce to a simple tail call to `clone()`, not even when using the most recent compilers (GCC 13 or clang 18).

I compiled cb08831e7249039b279c08f819efe3d085d32523 using GCC 13.2.0 with both `-Og` (debug build) and `-O2`, and the extra copies are being elided just fine.

dr-m (approved):

PR #3179, opened 2024-04-04
Add cpu-limit option for mini-benchmark

tonychen2001:
## Description
Enable configuration of CPU cycle limit for mini-benchmark

As performance improves, the permitted number of CPU cycles to be used
during the benchmark should be lowered to catch performance regressions. An
option `--cpu-limit` is added to better support this.

Additionally, there is a minor fix to the exit behavior of mini-benchmark when an error is encountered.

## Release Notes

None

## How can this PR be tested?

This can be tested by installing MariaDB in a docker container and executing `./support-files/mini-benchmark.sh`

## PR quality check

### DIFF
--- a/support-files/mini-benchmark.sh
+++ b/support-files/mini-benchmark.sh
@@ -310,4 +310,4 @@ case $RESULT in
 esac
 # Record the output into the log file, if requested
 ) 2>&1 | ($LOG && tee "$BENCHMARK_NAME"-"$TIMESTAMP".log)
-
+exit ${PIPESTATUS[0]} # Propagate errors in the sub-shell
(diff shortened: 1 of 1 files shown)


ottok (approved):
LGTM!

CI failure in https://buildbot.mariadb.org/#/builders/534/builds/16104 is for sure completely unrelated.

ottok:
I was under the impression that each PR needs two approvals to merge, but apparently no such rule is in place yet.

ottok:
This was merged on 10.11 as it is the oldest still maintained branch that has the mini-benchmark script. Merging this in a stable release is safe as the script is not part of the actual server and only used for development/testing purposes.

PR #3182, opened 2024-04-05
MDEV-14231  MATCH() AGAINST( IN BOOLEAN MODE), results mismatch

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-14231*

## Description
This patch is taken mysql-5.7.
st_mysql_ftparser_boolean_info(): Add variable position to store the correct offset for the default fts parser.
Added plugin_debug.test, multiple_index.test from mysql-5.7

## How can this PR be tested?
./mtr innodb_fts.plugin_debug innodb_fts.multiple_index

## PR quality check

### DIFF
--- a/include/mysql/plugin_audit.h.pp
+++ b/include/mysql/plugin_audit.h.pp
@@ -543,6 +543,7 @@ typedef struct st_mysql_ftparser_boolean_info
   int weight_adjust;
   char wasign;
   char trunc;
+  int position;
   char prev;
   char *quot;
 } MYSQL_FTPARSER_BOOLEAN_INFO;
--- a/include/mysql/plugin_auth.h.pp
+++ b/include/mysql/plugin_auth.h.pp
@@ -543,6 +543,7 @@ typedef struct st_mysql_ftparser_boolean_info
   int weight_adjust;
   char wasign;
   char trunc;
+  int position;
   char prev;
   char *quot;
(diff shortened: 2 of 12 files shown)


vuvova:
First, this absolutely cannot go into 10.4.
Second, adding the support for word positions is a feature, not a bug fix.
Third, what is the bug you're fixing? Where's the "results mismatch" ?

dr-m:
> First, this absolutely cannot go into 10.4. Second, adding the support for word positions is a feature, not a bug fix. Third, what is the bug you're fixing? Where's the "results mismatch" ?

Word positions are something that were added in MySQL 5.7 but not in MariaDB Server 10.2. I think that [MDEV-24088](https://jira.mariadb.org/browse/MDEV-24088) is somewhat related to this. This bug was filed when some MySQL 5.7 test cases were being considered for importing, in [MDEV-13626](https://jira.mariadb.org/browse/MDEV-13626).

Hopefullly @Thirunarayanan can clarify this, basically explain how and why MariaDB is incompatible with MySQL in this area.

Thirunarayanan:
Yes, I don't think we need to add position in fts default parser. 
Position is being used only inside innodb for proximity search query. Simple parser doesn't support phrase search query at all.
So closing this PR

dr-m:
#3422 superceded this (just adding the tests with some adjustments).

PR #3183, opened 2024-04-06
use ftruncate and open when their 64 version not available

xtexChooser:
- [x] *The Jira issue number for this PR is: MDEV-4658*

## Description
If ftruncate64 and open64 are not available, use ftruncate (libc has a transition which made them the same) and open (with flag O_LARGEFILE, open is the same as open64).
This fixes build issue on Alpine Linux and other musl distros.
Although the above issue is about *BSD, the error is also reproducible on Alpine Linux.

## How can this PR be tested?

Build on Alpine Linux edge or *BSD OSs.

The original code are added in d7143a4160d0, GitHub says it is initially included in 10.0.11 so I rebased this patch to 10.0 branch.

## PR quality check


xtexChooser:
f, I gave up, there are more errors from other files in storage/connect.
I am going to simply disable PLUGIN_CONNECT

PR #3184, opened 2024-04-08
MDEV-32489 Change buffer index fails to delete the records

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-32489*

## Description
When the change buffer records for a page span across multiple change buffer leaf pages or the starting record is at the beginning of a page with a left sibling, ibuf_delete_recs deletes only the records in first page and fails to move to subsequent pages.

Subsequently a slow shutdown hangs trying to delete those left over records.

Fix-A: Position the cursor to an user record in B-tree and exit only when all records are exhausted.

Fix-B: Make sure we call ibuf_delete_recs during slow shutdown for pages with IBUF entries to cleanup any previously left over records.

## Release Notes
None

## How can this PR be tested?
RQG test as done by the filer.

It is possible to repeat the issue with good predictability with somewhat complex mtr test, 4k page size and some code modification to force certain path. Otherwise we need multiple large tablespaces of size more than extent size. The test is updated in the MDEV. While the test can be used to validate the patch it is hard to maintain such test in repository and I am not adding it to our mtr test suite,

## PR quality check

### DIFF
--- a/storage/innobase/ibuf/ibuf0ibuf.cc
+++ b/storage/innobase/ibuf/ibuf0ibuf.cc
@@ -24,6 +24,7 @@ Insert buffer
 Created 7/19/1997 Heikki Tuuri
 *******************************************************/
 
+#include <tuple>
 #include "ibuf0ibuf.h"
 #include "sync0sync.h"
 #include "btr0sea.h"
@@ -2331,9 +2332,11 @@ static void ibuf_delete_recs(const page_id_t page_id)
 loop:
   btr_pcur_t pcur;
   ibuf_mtr_start(&mtr);
-  if (btr_pcur_open(ibuf.index, &tuple, PAGE_CUR_GE, BTR_MODIFY_LEAF,
-                    &pcur, &mtr) != DB_SUCCESS)
+  if (btr_pcur_open_on_user_rec(ibuf.index, &tuple, PAGE_CUR_GE,
+                                BTR_MODIFY_LEAF, &pcur, &mtr) != DB_SUCCESS)
+  {
     goto func_exit;
+  }
   if (!btr_pcur_is_on_user_rec(&pcur))
   {
     ut_ad(btr_pcur_is_after_last_on_page(&pcur));
@@ -2370,7 +2373,8 @@ func_exit:
 
 /** Merge the change buffer to some pages. */
 static void ibuf_read_merge_pages(const uint32_t* space_ids,
-				  const uint32_t* page_nos, ulint n_stored)
+				  const uint32_t* page_nos, ulint n_stored,
+				  bool slow_shutdown_cleanup)
 {
 	for (ulint i = 0; i < n_stored; i++) {
 		const ulint space_id = space_ids[i];
@@ -2394,26 +2398,25 @@ tablespace_deleted:
 		if (UNIV_LIKELY(page_nos[i] < size)) {
 			mtr.start();
 			dberr_t err;
-			buf_block_t *block =
+			/* Load the page and apply change buffers. */
+			std::ignore =
 			buf_page_get_gen(page_id_t(space_id, page_nos[i]),
 					 zip_size, RW_X_LATCH, nullptr,
 					 BUF_GET_POSSIBLY_FREED,
 					 __FILE__, __LINE__, &mtr, &err, true);
-			bool remove = !block
-				|| fil_page_get_type(block->frame)
-				!= FIL_PAGE_INDEX
-				|| !page_is_leaf(block->frame);
 			mtr.commit();
 			if (err == DB_TABLESPACE_DELETED) {
 				goto tablespace_deleted;
 			}
-			if (!remove) {
-				continue;
-			}
 		}
 
-		if (srv_shutdown_state == SRV_SHUTDOWN_NONE
-		    || srv_fast_shutdown) {
+		/* During slow shutdown cleanup, we apply all pending IBUF
+		changes and need to cleanup any left-over IBUF records. There
+		are a few cases when the changes are already discarded and IBUF
+		bitmap is cleaned but we miss to delete the record. Even after
+		the issues are fixed, we need to keep this safety measure for
+		upgraded DBs with such left over records. */
+		if (!slow_shutdown_cleanup) {
 			continue;
 		}
 
@@ -2445,10 +2448,11 @@ tablespace_deleted:
 }
 
 /** Contract the change buffer by reading pages to the buffer pool.
+@param slow_shutdown true, if called during slow shutdown
 @return a lower limit for the combined size in bytes of entries which
 will be merged from ibuf trees to the pages read
 @retval 0 if ibuf.empty */
-ulint ibuf_contract()
+ulint ibuf_contract(bool slow_shutdown)
 {
 	if (UNIV_UNLIKELY(!ibuf.index)) return 0;
 	mtr_t		mtr;
@@ -2492,7 +2496,7 @@ ulint ibuf_contract()
 	ibuf_mtr_commit(&mtr);
 	btr_pcur_close(&pcur);
 
-	ibuf_read_merge_pages(space_ids, page_nos, n_pages);
+	ibuf_read_merge_pages(space_ids, page_nos, n_pages, slow_shutdown);
 
 	return(sum_sizes + 1);
 }
@@ -2574,7 +2578,7 @@ ibuf_merge_space(
 		}
 #endif /* UNIV_DEBUG */
 
-		ibuf_read_merge_pages(spaces, pages, n_pages);
+		ibuf_read_merge_pages(spaces, pages, n_pages, false);
 	}
 
 	return(n_pages);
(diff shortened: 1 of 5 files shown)


storage/innobase/include/srv0start.h
```diff
@@ -120,7 +121,7 @@ extern bool srv_undo_sources;
 
 /** At a shutdown this value climbs from SRV_SHUTDOWN_NONE to
 SRV_SHUTDOWN_CLEANUP and then to SRV_SHUTDOWN_LAST_PHASE, and so on */
-extern	enum srv_shutdown_t	srv_shutdown_state;
+extern std::atomic<enum srv_shutdown_t> srv_shutdown_state;
```
dr-m, line 124: This will break the build on clang:
```
/home/buildbot/amd64-debian-11-msan/build/storage/innobase/fil/fil0crypt.cc:1132:3: error: multiple conversions from switch condition type 'std::atomic<enum srv_shutdown_t>' to an integral or enumeration type
                switch (srv_shutdown_state) {
                ^       ~~~~~~~~~~~~~~~~~~
/usr/lib/llvm-15/bin/../include/c++/v1/atomic:1562:5: note: conversion to enumeration type 'srv_shutdown_t'
```
Furthermore, if we want to change this, I think that we should use `Atomic_relaxed<srv_shutdown_t>` to avoid the implicit default `std::memory_order_seq_cst`, which feels like an overkill to me.

storage/innobase/ibuf/ibuf0ibuf.cc
```diff
@@ -2389,26 +2393,25 @@ static void ibuf_read_merge_pages(const uint32_t* space_ids,
...
 
-		if (srv_shutdown_state == SRV_SHUTDOWN_NONE
-		    || srv_fast_shutdown) {
+		/* During slow shutdown cleanup, we apply all pending IBUF
+		changes and need to cleanup any left-over IBUF records. There
+		are a few cases when the changes are already discarded and IBUF
+		bitmap is cleaned but we miss to delete the record. Even after
+		the issues are fixed, we need to keep this safety measure for
+		upgraded DBs with such left over records. */
+		if (!slow_shutdown_cleanup) {
```
dr-m, line 2414: Which cases is this referring to? Which function is cleaning the bitmap flag without deleting (or delete-marking) the change buffer record in the same mini-transaction?

storage/innobase/ibuf/ibuf0ibuf.cc
```diff
@@ -2367,6 +2367,10 @@ static void ibuf_delete_recs(const page_id_t page_id)
 static void ibuf_read_merge_pages(const uint32_t* space_ids,
 				  const uint32_t* page_nos, ulint n_stored)
 {
+	/* Identify slow shutdown cleanup phase to clean up change buffers. */
+	bool slow_shutdown_cleanup = !srv_fast_shutdown
+				     && srv_shutdown_state >= SRV_SHUTDOWN_CLEANUP;
```
dr-m, line 2372: There only ever was one change buffer. In the archaic formatting style, overflow lines are indented by only 1 extra TAB.

dr-m (commented):
I think that operations that involve multiple change buffer pages should set a delete-mark flag on the change buffer record to mark it as processed. Is that not the case in this scenario? I think that the commit message or a source code comment needs to describe the scenario in more detail, given that it is unfeasible to write a regression test for it.

dr-m (commented):
I think that we can clean up the `srv_shutdown_state` in a separate ticket if desired. We do not actually have to read that variable here. Side note: the state values `SRV_SHUTDOWN_LAST_PHASE` and `SRV_SHUTDOWN_EXIT_THREADS` could be defined in debug builds only.

I think that it is safe to invoke `ibuf_delete_recs()` and let it non-atomically delete all buffered changes for a page, before we invoke `ibuf_reset_bitmap()` or similar. In `ibuf_merge_or_delete_for_page()` for the case that the user page is marked as freed in the tablespace, we are doing this in the opposite order. It would be clearer to retain the same order there as well.

I see that the `slow_shutdown_cleanup` in `ibuf_read_merge_pages()` is refining something that already was implemented in 165564d3c33ae3d677d70644a83afcb744bdbf65. I do not see a problem with that. We do have to be careful about that code; see the commits mentioned in that commit message.

storage/innobase/ibuf/ibuf0ibuf.cc
mariadb-DebarunBanerjee: 1. ibuf_delete_recs - the current bug scenario.
2. I am sure there were other cases where we had left over entries. I am not introducing this "if" condition rather replacing it. If you see a bit down you would see more comments.

storage/innobase/include/srv0start.h
mariadb-DebarunBanerjee: While I think it needs careful thoughts to check if such checks should be relaxed atomics I think it is best to keep this part out of the patch. I am now passing the slow_shutdown parameter down as suggested by other comment.

storage/innobase/ibuf/ibuf0ibuf.cc
mariadb-DebarunBanerjee: I think you prefer "change buffer elements" or "change buffer entries" here. It seems I was not correct assuming it can be intuitively deduced. Either way this lines are now removed and "slow_shutdown_cleanup" is now passed down as function parameter.

storage/innobase/ibuf/ibuf0ibuf.cc
```diff
@@ -2326,9 +2327,8 @@ static void ibuf_delete_recs(const page_id_t page_id)
 loop:
   btr_pcur_t pcur;
   ibuf_mtr_start(&mtr);
-  if (btr_pcur_open(ibuf.index, &tuple, PAGE_CUR_GE, BTR_MODIFY_LEAF,
-                    &pcur, &mtr) != DB_SUCCESS)
-    goto func_exit;
+  btr_pcur_open_on_user_rec(ibuf.index, &tuple, PAGE_CUR_GE,
+                            BTR_MODIFY_LEAF, &pcur, &mtr);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: I had a look of the error handling in 10.6 but it would be a large change to port all part and could impact even non-ibuf part of the code. The interface for opening on user record and placing on next user record (moving to next page) is used in many places (10.5) and extending the usage here (a rarer case) seems safe and inline with 10.5.

storage/innobase/ibuf/ibuf0ibuf.cc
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Adding a bool parameter and passing it down seems a good idea. It also eliminates the dependency on the server shutdown state. I will change this part and upload.

storage/innobase/ibuf/ibuf0ibuf.cc
mariadb-DebarunBanerjee: Specifically for the bug scenario it was the place below. If you want more details, please follow the steps in MDEV to repeat the issue.

ibuf_merge_or_delete_for_page()
```
                if (bitmap_bits && fseg_page_is_free(
                                space, page_id.page_no())) {
                        ibuf_mtr_start(&mtr);
                        mtr.set_named_space(space);
                        ibuf_reset_bitmap(block, page_id, zip_size, &mtr);
                        ibuf_mtr_commit(&mtr);
                        bitmap_bits = 0;
                        if (!block
                            || btr_page_get_index_id(block->frame)
                            != DICT_IBUF_ID_MIN + IBUF_SPACE_ID) {
                                ibuf_delete_recs(page_id);
                        }
                }

```

storage/innobase/ibuf/ibuf0ibuf.cc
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Added minimal error handling for btr_pcur_open_on_user_rec.

storage/innobase/ibuf/ibuf0ibuf.cc
```diff
@@ -24,6 +24,7 @@ Insert buffer
 Created 7/19/1997 Heikki Tuuri
 *******************************************************/
 
+#include <tuple>
```
Thirunarayanan, line 27: Do we need this change?

Thirunarayanan (approved):

storage/innobase/ibuf/ibuf0ibuf.cc
dr-m: I assumed that it was for the `std::ignore`, for overriding `__attribute__((warn_unused_result))`. I think I tested it for some earlier pull request.

mariadb-DebarunBanerjee: @dr-m right, I included it for std::ignore.
https://en.cppreference.com/w/cpp/utility/tuple/ignore

@Thirunarayanan 
I think it should be good practice to include the header.

PR #3185, opened 2024-04-08
MDEV-33855 MSAN use-of-uninitialized-value in rtr_pcur_getnext_from_path()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33855*
## Description
`rtr_pcur_getnext_from_path()`: Remove a debug assertion that suffers from a data race with `buf_LRU_block_free_non_file_page()`.

## Release Notes
No mention is needed, because this is only removing a bogus debug assertion.

## How can this PR be tested?
Repeatedly run `./mtr innodb_gis.rtree_compress` in a MemorySanitizer instrumented build.
## PR quality check

### DIFF
--- a/storage/innobase/gis/gis0sea.cc
+++ b/storage/innobase/gis/gis0sea.cc
@@ -289,10 +289,6 @@ rtr_pcur_getnext_from_path(
 			mtr->rollback_to_savepoint(1);
 		}
 
-		ut_ad((my_latch_mode | 4) == BTR_CONT_MODIFY_TREE
-		      || !page_is_leaf(btr_cur_get_page(btr_cur))
-		      || !btr_cur->page_cur.block->page.lock.have_any());
-
 		const auto block_savepoint = mtr->get_savepoint();
 		block = buf_page_get_gen(
 			page_id_t(index->table->space_id,


storage/innobase/gis/gis0sea.cc
```diff
@@ -278,9 +278,8 @@ rtr_pcur_getnext_from_path(
 				      ulint{RW_S_LATCH}, "");
 			static_assert(ulint{BTR_MODIFY_LEAF} ==
 				      ulint{RW_X_LATCH}, "");
-			rw_latch = (my_latch_mode | 4) == BTR_CONT_MODIFY_TREE
-				? RW_NO_LATCH
-				: rw_lock_type_t(my_latch_mode);
+			rw_latch = rw_lock_type_t(my_latch_mode
+						  & (RW_X_LATCH | RW_S_LATCH));
 		} else {
 			rw_latch = RW_X_LATCH;
```
lottaquestions, line 284: A RW_X_LATCH latch is now being set on both sides of the condition. Is there a reason for the additional RW_S_LATCH in the if part?

dr-m: The `if` part is a bit convoluted, but it can actually assign `rw_latch` to either `RW_X_LATCH` or `RW_S_LATCH`. In 75096c84b44875b5d226a734fffb08578bc21e96 I redefined many `btr_latch_mode` constants in such a way that we can simply convert them to `rw_lock_type_t` or `mtr_memo_type_t` with simple bitmasking. The changed statement should be equivalent to the following, if we assume that the variables are integers:
```c++
rw_latch = my_latch_mode & 3;
```
For example, we convert `BTR_MODIFY_TREE` and `BTR_MODIFY_LEAF` to `RW_X_LATCH` and `BTR_SEARCH_TREE` and `BTR_SEARCH_LEAF` to `RW_S_LATCH`.

Conditional branches can be very expensive on modern superscalar processors. In [MDEV-30567](https://jira.mariadb.org/browse/MDEV-30567) and 66b21ed5403aa4ef18a4c797af353ec3379b152d you can find another example where using bitwise arithmetics might make code harder to read for some, but faster to execute.

lottaquestions: Thank you for explaining the rationale for the change. It makes a lot more sense now :-)

storage/innobase/gis/gis0sea.cc
```diff
@@ -289,10 +288,6 @@ rtr_pcur_getnext_from_path(
 			mtr->rollback_to_savepoint(1);
 		}
 
-		ut_ad((my_latch_mode | 4) == BTR_CONT_MODIFY_TREE
-		      || !page_is_leaf(btr_cur_get_page(btr_cur))
-		      || !btr_cur->page_cur.block->page.lock.have_any());
-
```
mariadb-DebarunBanerjee, line 295: We release all the page latches (mtr->rollback_to_savepoint(1)) when rtr_pcur_getnext_from_path() is called. This looks fine as we are either holding X (pessimistic) or SX(optimistic) latch at index level and have all the possible page IDs cached already.

Only issue is that cursor now points to a possible invalid block "btr_cur->page_cur.block". We eventually set the cursor to the next candidate page after latching it and there doesn't seem to be any real issue other than this debug assert trying to access the block.

The block can be freed even can be reused concurrently and MSAN looks to be catching it when it is indeed freed. We should not access the block/page at this point and It justifies the removal of the assert.

Can you please add a comment here stating the page cursor could point to invalid block at this point and should not be accessed ? A braver approach would be to set the page cursor block to nullptr immediately after rolling back to savepoint and assert here that the page cursor block is nullptr at this point.

storage/innobase/gis/gis0sea.cc
```diff
@@ -278,9 +278,8 @@ rtr_pcur_getnext_from_path(
 				      ulint{RW_S_LATCH}, "");
 			static_assert(ulint{BTR_MODIFY_LEAF} ==
 				      ulint{RW_X_LATCH}, "");
-			rw_latch = (my_latch_mode | 4) == BTR_CONT_MODIFY_TREE
-				? RW_NO_LATCH
-				: rw_lock_type_t(my_latch_mode);
+			rw_latch = rw_lock_type_t(my_latch_mode
+						  & (RW_X_LATCH | RW_S_LATCH));
```
mariadb-DebarunBanerjee, line 282: I think this part of latching change should not be required. Please see the other comment.

mariadb-DebarunBanerjee (commented):
@dr-m Looks like the debug assert is at fault. Please check my comments.

storage/innobase/gis/gis0sea.cc
dr-m: You are right, only the debug assertion seems to be at fault. I successfully tested the following:
```diff
diff --git a/storage/innobase/gis/gis0sea.cc b/storage/innobase/gis/gis0sea.cc
index fb4fcf7bc8d..9aedd4af69c 100644
--- a/storage/innobase/gis/gis0sea.cc
+++ b/storage/innobase/gis/gis0sea.cc
@@ -289,10 +289,6 @@ rtr_pcur_getnext_from_path(
 			mtr->rollback_to_savepoint(1);
 		}
 
-		ut_ad((my_latch_mode | 4) == BTR_CONT_MODIFY_TREE
-		      || !page_is_leaf(btr_cur_get_page(btr_cur))
-		      || !btr_cur->page_cur.block->page.lock.have_any());
-
 		const auto block_savepoint = mtr->get_savepoint();
 		block = buf_page_get_gen(
 			page_id_t(index->table->space_id,
```

mariadb-DebarunBanerjee (approved):

PR #3187, opened 2024-04-09
MDEV-33802 Weird read view after ROLLBACK of other transactions.

vlad-lesin:
In the case when all unique key fields are nullable, there can be several records with all NULL key fields in unique index. When transaction is resumed after waiting on the record with all key fields equal to NULL, and stored in persistent cursor record is deleted, persistent cursor can be restored on the record with all key fields equal to NULL, and such record is wrongly treated as the record with the same unique key as the stored in persistent cursor record, what is wrong.

The fix is to check if all unique fields are null when persistent cursor position is restored and not all comparing fields are equal, and if so, don't treat the record as a record with the same unique key as in the stored in persistent cursor record key.

dict_index_t::nulls_equal was removed, as it was initially developed for never existed in MariaDB "intrinsic tables", and there is no code, which would set it to "true".

Reviewed by Marko Mäkelä.

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0pcur.cc
+++ b/storage/innobase/btr/btr0pcur.cc
@@ -453,8 +453,15 @@ btr_pcur_t::restore_position(ulint restore_latch_mode, const char *file,
 
 			return SAME_ALL;
 		}
-		if (n_matched_fields >= index->n_uniq)
-			ret_val= SAME_UNIQ;
+		if (n_matched_fields >= index->n_uniq
+		    /* Unique indexes can contain "NULL" keys, and if all
+		    unique fields are NULL and not all tuple
+		    fields match to record fields, then treat it as if
+		    restored cursor position points to the record with
+		    not the same unique key. */
+		    && !(index->n_nullable
+			    && dtuple_contains_null(tuple, index->n_uniq)))
+			  ret_val= SAME_UNIQ;
 	}
 
 	mem_heap_free(heap);
--- a/storage/innobase/row/row0ins.cc
+++ b/storage/innobase/row/row0ins.cc
@@ -2038,7 +2038,7 @@ row_ins_dupl_error_with_rec(
 	/* In a unique secondary index we allow equal key values if they
 	contain SQL NULLs */
 
-	if (!dict_index_is_clust(index) && !index->nulls_equal) {
+	if (!dict_index_is_clust(index)) {
 
 		for (i = 0; i < n_unique; i++) {
 			if (dfield_is_null(dtuple_get_nth_field(entry, i))) {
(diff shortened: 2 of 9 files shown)


storage/innobase/row/row0ins.cc
```diff
@@ -2144,13 +2144,11 @@ row_ins_scan_sec_index_for_duplicate(
 	n_unique first fields is NULL, a unique key violation cannot occur,
 	since we define NULL != NULL in this case */
 
-	if (!index->nulls_equal) {
-		for (ulint i = 0; i < n_unique; i++) {
-			if (UNIV_SQL_NULL == dfield_get_len(
-					dtuple_get_nth_field(entry, i))) {
+	for (ulint i = 0; i < n_unique; i++) {
+		if (UNIV_SQL_NULL == dfield_get_len(
+				dtuple_get_nth_field(entry, i))) {
```
dr-m, line 2149: This could be written as `dfield_is_null(dtuple_get_nth_field(entry, i))`. In `btr_pcur_t::restore_position()` but not here we only enter the loop if `index->n_nullable` is nonzero. Maybe, invoke a common function for both?

storage/innobase/btr/btr0pcur.cc
```diff
@@ -453,8 +454,22 @@ btr_pcur_t::restore_position(ulint restore_latch_mode, const char *file,
...
+			if (!index->n_nullable)
+			  ret_val= SAME_UNIQ;
+			else
+			  for (uint i= 0; i < index->n_uniq; ++i)
+			    if (!dfield_is_null(
+				  dtuple_get_nth_field(tuple, i))) {
+			      ret_val= SAME_UNIQ;
+			      break;
+			  }
+		}
```
dr-m, line 472: This is using a mix of the archaic and the common formatting style. Maybe we should simply reformat the entire function. The predicate in the `for` loop can be written as `dfield_is_null(dtuple_get_nth_field(tuple, i))`. We might want to refactor `dtuple_contains_null()` a little, to allow the number of fields to be passed as a parameter. I am not sure if there is a benefit for that function to be defined `inline`.

dr-m (approved):
Thank you, this looks OK to me. Please consider my suggestions.

storage/innobase/btr/btr0pcur.cc
vlad-lesin: If we use `dfield_is_null(dtuple_get_nth_field(tuple, i)` in condition, then the fix becomes quite small, and there is no need to reformat all function.

Besides, the previous condition was wrong. Because we need to check if any field equals to NULL in the tuple, but not all fields, as `NULL!=NULL`. I inserted (1, NULL, 2), record in the test, which was previously (1, NULL, NULL) , and got two (1, NULL, 2) records in the output with the previous fix.

storage/innobase/row/row0ins.cc
vlad-lesin: I added `index->n_nullable` check in the condition. I did not include this condition in `dfield_is_null()` because in this case we would need to pass index into the function, and the only reason for such passing is to quit the function if some member of index is 0 at the very beginning of the function, i.e. if the condition is false, we should not invoke the function at all.

PR #3188, opened 2024-04-09
MDEV-33868 Assertion `trx->bulk_insert' failed in innodb_prepare_commit_versioned

Thirunarayanan:
## Description

- This issue is caused by commit 188c5da72a0057e4572b1d7e4efcd0c39332a839 (MDEV-32453). InnoDB fails to end the bulk insert for the table after applying the bulk insert operation. This leads to assertion during commit process.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/row/row0merge.cc
+++ b/storage/innobase/row/row0merge.cc
@@ -5362,6 +5362,7 @@ void trx_t::bulk_rollback_low()
         low_limit= t.second.get_first();
       delete t.second.bulk_store;
       t.second.bulk_store= nullptr;
+      t.second.end_bulk_insert();
     }
   }
   trx_savept_t bulk_save{low_limit};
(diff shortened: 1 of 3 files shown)


storage/innobase/row/row0merge.cc
```diff
@@ -5370,12 +5370,21 @@ void trx_t::bulk_rollback_low()
...
+
+  dberr_t err= DB_SUCCESS;
+  if (it->second.bulk_store)
+  {
+    err= it->second.write_bulk(table, this);
+    if (err) bulk_rollback_low();
+  }
+  if (it->second.is_bulk_insert())
+    it->second.end_bulk_insert();
   return err;
```
dr-m, line 5388: If I understood it correctly, the last `if` is unnecessary and we can just unconditionally invoke `end_bulk_insert()`, which will execute `first&= ~BULK`. It is better to avoid unnecessary conditions.

Could `trx_t::bulk_rollback_low()` execute `t.second.end_bulk_insert()`? That would seem to be more robust to me. In that case, this could be simplified to the following:
```c++
  auto it= mod_tables.find(table);
  if (it != mod_tables.end() && it->second.bulk_store)
    if (dberr_t err= it->second.write_bulk(table, this))
      return err;
  it->second.end_bulk_insert();
  return DB_SUCCESS;
```

dr-m (commented):
Thank you, this looks OK to me, but I would suggest some further cleanup.

dr-m (approved):

PR #3193, opened 2024-04-11
MDEV-33809 Bulk insert or DDL fails if a BLOB is too long

Thirunarayanan:
## Description
SyncFileIO should do multiple read/write call when length is greater than os_file_request_size_max value.

## How can this PR be tested?
Tested the patch with os_file_request_size_max as 65536. InnoDB does multiple read/write in case of read/write.
This is done in the commit b1fa0f59a635f6333a8a0889c53f663746860e55

## PR quality check

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -320,6 +320,12 @@ private:
   ssize_t m_n;
   /** Offset from where to read/write */
   os_offset_t m_offset;
+
+  /** Do the read/write
+  @param request The IO context and type
+  @param n       Number of bytes to read/write
+  @return the number of bytes read/written or negative value on error */
+  ssize_t execute_low(const IORequest& request, ssize_t n);
 };
 
 #ifndef _WIN32 /* On Microsoft Windows, mandatory locking is used */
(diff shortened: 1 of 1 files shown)


storage/innobase/os/os0file.cc
```diff
@@ -675,30 +681,58 @@ os_file_create_subdirs_if_needed(
...
+		advance(os_file_request_size_max);
+	}
+
+	if (ssize_t n= execute_low(request, m_n)) {
+		if (n < 0) return n;
+		n_bytes += n;
+	}
+
+	return(n_bytes);
+}
```
dr-m, line 734: Let’s avoid the old InnoDB style formatting in new code, and let’s only document the parameters in the function declaration in the header file:
```c++
ssize_t SyncFileIO::execute(const IORequest &request)
{
  ssize_t n_bytes= 0;
  ut_ad(m_n > 0);
/// ...
  return n_bytes;
}
```

storage/innobase/os/os0file.cc
```diff
@@ -675,30 +681,58 @@ os_file_create_subdirs_if_needed(
 @param[in]	request	The IO context and type
 @return the number of bytes read/written or negative value on error */
 ssize_t
-SyncFileIO::execute(const IORequest& request)
+SyncFileIO::execute_low(const IORequest& request, ssize_t n)
```
dr-m, line 684: The added parameter is  not documented in the comment. I don’t see much value in duplicating the comment. There already is one at the member function declaration. Let’s use the common formatting style in functions that are almost completely rewritten:
```c++
ssize_t SyncFileIO::execute_low(const IORequest &request, ssize_t n)
{
  ut_ad(n > 0);
  ut_ad(size_t(n) <= os_file_request_size_max);
  if (request.is_read())
    return IF_WIN(tpool::,)pread(m_fh, m_buf, n, m_offset);
  else
    return IF_WIN(tpool::,)pwrite(m_fh, m_buf, n, m_offset);
}
```

dr-m (approved):
Thank you. I would suggest some minor cleanup of the code, mainly formatting and removing duplicated and partly contradictory function comments.

PR #3194, opened 2024-04-12
MDEV-33431 Latching order violation reported fil_system.sys_space.lat…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33431*

## Description
**Issue:**
The actual order of acquisition of the IBUF pessimistic insert mutex (SYNC_IBUF_PESS_INSERT_MUTEX) and IBUF header page latch (SYNC_IBUF_HEADER) w.r.t space latch (SYNC_FSP) differs from the order defined in sync0types.h. It was not discovered earlier as the path to ibuf_remove_free_page was not covered by the mtr test. Ideal order and one defined in sync0types.h is as follows.
SYNC_IBUF_HEADER -> SYNC_IBUF_PESS_INSERT_MUTEX -> SYNC_FSP

In ibuf_remove_free_page, we acquire space latch earlier and we have the order as follows resulting in the assert with innodb_sync_debug=on. SYNC_FSP -> SYNC_IBUF_HEADER -> SYNC_IBUF_PESS_INSERT_MUTEX

**Fix:**
We do maintain this order in other places and there doesn't seem to be any real issue here. To reduce impact in GA versions, we avoid doing extensive changes in mutex ordering to match the current SYNC_IBUF_PESS_INSERT_MUTEX order. Instead we relax the ordering check for IBUF pessimistic insert mutex using SYNC_NO_ORDER_CHECK.

## Release Notes
None

## How can this PR be tested?
./mtr innodb.change_buffer_free

## PR quality check


mysql-test/suite/innodb/t/change_buffer_free.test
```diff
@@ -0,0 +1,53 @@
+#
+# MDEV 33431: Latching order violation reported fil_system.sys_space.latch and ibuf_pessimistic_insert_mutex
+#
+--source include/have_innodb.inc
+--source include/have_debug.inc
```
dr-m, line 5: While normally `defined ENABLED_DEBUG_SYNC` is equal to `!defined DBUG_OFF`, it is possible to enable them independently. Please add also `have_debug_sync.inc`.

dr-m (approved):
Thank you, this looks OK to me. We both concluded that this looks like a false alarm. I assume that if you omit the change to `sync0debug.cc`, the test will cause a crash.

mysql-test/suite/innodb/t/change_buffer_free.test
mariadb-DebarunBanerjee: done.

PR #3196, opened 2024-04-12
Factor TABLE_LIST creation from add_table_to_list

DaveGosselin-MariaDB:
Ideally our methods and functions should do one thing, do that well, and do only that.  add_table_to_list does far more than adding a table to a list, so this commit factors the TABLE_LIST creation out to a new TABLE_LIST constructor.  It then uses placement new() to create it in the correct memory area (result of thd->calloc). Benefits of this approach:
 1. Method add_table_to_list now returns as early as possible on an error
 2. Fewer side-effects incurred on creating the TABLE_LIST object
 3. TABLE_LIST won't be calloc'd if copy_to_db fails
 4. Local declarations moved closer to their respective first uses
 5. Improved code readability and logical flow: factored a couple of other functions to keep the happy path more to the left, which makes them easier to follow at a glance.

### DIFF
--- a/sql/table.cc
+++ b/sql/table.cc
@@ -49,6 +49,7 @@
 #include "wsrep_schema.h"
 #endif
 #include "log_event.h"           // MAX_TABLE_MAP_ID
+#include "sql_class.h"
 
 /* For MySQL 5.7 virtual fields */
 #define MYSQL57_GENERATED_FIELD 128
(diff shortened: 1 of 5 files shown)


sql/table.cc
```diff
@@ -5850,6 +5851,62 @@ void TABLE::reset_item_list(List<Item> *item_list, uint skip) const
...
+                       LEX_CSTRING _alias_str,
+                       LEX_CSTRING *_alias,
+                       Table_ident *_table,
+                       thr_lock_type _lock_type,
+                       enum_mdl_type _mdl_type,
+                       ulong _table_options,
+                       bool _info_schema,
+                       st_select_lex *_select_lex,
+                       List<Index_hint> *_index_hints_org,
+                       LEX_STRING *_option)
```
vuvova, line 5866: Please, try to stick to the coding style which is used elsewhere. In particular, constructor arguments don't typically start with an underscore.

sql/table.cc
```diff
@@ -5850,6 +5851,62 @@ void TABLE::reset_item_list(List<Item> *item_list, uint skip) const
...
+{
+  auto *ptr= this;
+  ptr->db= _db;
```
vuvova, line 5869: you don't really need this `ptr->` after you've copied the code here. It would look more natural without

sql/table.cc
```diff
@@ -5850,6 +5851,62 @@ void TABLE::reset_item_list(List<Item> *item_list, uint skip) const
...
+
+  ptr->table_name= _table->table;
+  ptr->lock_type= _lock_type;
+  ptr->mdl_type= _mdl_type;
+  ptr->table_options= _table_options;
+  ptr->updating= MY_TEST(table_options & TL_OPTION_UPDATING);
+  /* TODO: remove TL_OPTION_FORCE_INDEX as it looks like it's not used */
+  ptr->force_index= MY_TEST(table_options & TL_OPTION_FORCE_INDEX);
+  ptr->ignore_leaves= MY_TEST(table_options & TL_OPTION_IGNORE_LEAVES);
+  ptr->sequence=      MY_TEST(table_options & TL_OPTION_SEQUENCE);
```
vuvova, line 5890: all these MY_TEST are redundant

vuvova (changes requested):

sql/table.cc
DaveGosselin-MariaDB: Removed.

vuvova (approved):
ok to push

PR #3206, opened 2024-04-15
MDEV-33616 Tests failing on macOS

DaveGosselin-MariaDB:
MDEV-33616 Tests failing on macOS

    These tests rely on THR_KEY_mysys but it is not initialized.  On
    Linux, the corresponding thread variable is null, but on macOS it has a
    nonzero value.  In all cases, initialize the variable explicitly by
    calling MY_INIT and my_end appropriately.

### DIFF
--- a/unittest/mysys/dynstring-t.c
+++ b/unittest/mysys/dynstring-t.c
@@ -26,8 +26,10 @@ static void check(const char *res)
   str1.length= 0;
 }
 
-int main(void)
+int main(int argc, char** argv)
 {
+  MY_INIT(argv[0]);
+
   plan(23);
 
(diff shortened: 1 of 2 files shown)


vuvova:
> We could call MY_INIT, however some memory is leaked on test program exit in that case

did you pair it with `my_end()` at the end?

DaveGosselin-MariaDB:
> > We could call MY_INIT, however some memory is leaked on test program exit in that case
> 
> did you pair it with `my_end()` at the end?

I saw the `MY_INIT` macro standing alone and didn't know about `my_end()`.  I tried that and it works, so I'll deliver that instead of the factoring-based approach.

PR #3207, opened 2024-04-15
MDEV-33616 Tests failing on macOS

DaveGosselin-MariaDB:
compat/oracle.sp-inout now uses lowercase object names for test compatibility with both case-insensitive and case-sensitive filesystems.


abarkov (approved):
Dave, the fix is OK to push. Thank you!

PR #3154, opened 2024-03-25
Update tests to be compatible with OpenSSL 3.2.0

DarrenZhang01:
## Description
As of version 3.2.0, OpenSSL updated the error message in new versions
("https://github.com/openssl/openssl/commit/81b741f68984"). Update the
tests and result files such that they are compatible with both original
and new error messages.

## How can this PR be tested?
By running the `ssl_crl` test.

## Basing the PR against the correct MariaDB version
- [x] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## Copyright
All new code of the whole pull request, including one or several files that are
either new files or modified ones, are contributed under the BSD-new license. I
am contributing on behalf of my employer Amazon Web Services, Inc.

PR #3208, opened 2024-04-15
MDEV-35998 - Refine OpenSSL test names and allow more complete OpenSSL version coverage

DarrenZhang01:
## Description
- Remove "6975" from the OpenSSL test and result file names. "6975" refers to https://jira.mariadb.org/browse/MDEV-6975 about implementing TLS support in general, which is not a bug report. In this case, this is a generic OpenSSL test set, so update the test file names for better clarification.
- Remove the upper bound of the OpenSSL version check to allow the tests to also run with latest OpenSSL versions given that this is a generic set of OpenSSL tests.
- Update the tests and result files such that they are compatible with new error messages for OpenSSL 3.2.0 and later versions ("https://github.com/openssl/openssl/commit/81b741f68984").
- Update the configuration name from "tlsv10" to "sslv3" when SSL cipher used is "SSLv3".

## How can this PR be tested?

by running the updated OpenSSL tests.

## PR quality check

## Copyright

### DIFF
--- a/mysql-test/main/openssl.test
+++ b/mysql-test/main/openssl.test
@@ -1,15 +1,13 @@
+# This is OpenSSL test.
 #
-# MDEV-6975 Implement TLS protocol
-#
-# test SSLv3 and TLSv1.2 ciphers when OpenSSL is restricted to SSLv3 or TLSv1.2
+# Test SSLv3 and TLSv1.2 ciphers when OpenSSL is restricted to SSLv3 or TLSv1.2
+# 
+# Related: MDEV-6975 Implement TLS protocol
 #
 source include/have_ssl_communication.inc;
 source include/require_openssl_client.inc;
 
-# this is OpenSSL test.
-
 create user ssl_sslv3@localhost;
-# grant select on test.* to ssl_sslv3@localhost require cipher "AES128-SHA";
 grant select on test.* to ssl_sslv3@localhost require cipher "AES128-SHA";
 create user ssl_tls12@localhost;
 grant select on test.* to ssl_tls12@localhost require cipher "AES128-SHA256";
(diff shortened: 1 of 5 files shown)


LinuxJedi (commented):
I have a few thoughts on this. Not necessarily related to your changes, but in some ways they are.

First of all, I'm not entirely sure what this test is supposed to do. `SSLv3 ciphers: user is ok with any cipher;` but then uses a TLS cipher?

Secondly, this test is known buggy anyway (MDEV-16902).

Finally I'm not sure I'm comfortable still supporting SSLv3 ciphers as an option.

As for the PR, it looks like the test was for 1.0.1d - 1.1.0, but now it is open for all current versions. Is that correct? I don't know the history of this test to know why 1.1.0 was the cut-off.

ottok:
> Secondly, this test is known buggy anyway (MDEV-16902).

Yes but nothing has been done for [MDEV-16902](https://jira.mariadb.org/browse/MDEV-16902) in since 2018. Fixing the test now is useful still, right?

> Finally I'm not sure I'm comfortable still supporting SSLv3 ciphers as an option.

That can be dropped if you think that supporting older versions is not relevant anymore. The biggest fix here is however cleaning up a OpenSSL 1.1.0 -only test to something that has value and actually runs on any modern system.

> As for the PR, it looks like the test was for 1.0.1d - 1.1.0, but now it is open for all current versions. Is that correct? I don't know the history of this test to know why 1.1.0 was the cut-off.

This was discovered while working on #3154. Commit messages and Jira descriptions from MDEV-6975 are thin and thus rewriting the test to something more sensible now is based on best guess. For sure it was stupid to have a test named `openssl_6975` and MDEV-6975 is not a bug, but Jira to implement OpenSSL in general. Thus the test should probably test OpenSSL in general, right? If you think such a test is not needed, then the test should be deleted.

LinuxJedi (approved):
I dug quite a bit into the background of this one. It was disabled (with other tests) for 1.1.1 due to a cipher issue. I don't know exactly what that issue is, but I'm guessing it was attempting to use AES256-SHA with tlsv1.2, which quite rightly shouldn't work.

This is a good start for fixing the test, but a lot more work is needed. For starters, it needs v1.3 support, different ciphers (ECC at least), etc... Although having said that, I think this is probably testing the SSL library more than it tests our code.

In addition, there are other tests disabled for 1.1.1 onwards that need updating, such as ssl_cipher.test.

This will need a rebase against 10.5 and a thorougher check on all platforms.

mysql-test/main/openssl.test
```diff
@@ -1,15 +1,13 @@
+# This is OpenSSL test.
 #
-# MDEV-6975 Implement TLS protocol
-#
-# test SSLv3 and TLSv1.2 ciphers when OpenSSL is restricted to SSLv3 or TLSv1.2
+# Test SSLv3 and TLSv1.2 ciphers when OpenSSL is restricted to SSLv3 or TLSv1.2
+# 
+# Related: MDEV-6975 Implement TLS protocol
```
vuvova, line 5: we have a standard test header,
```
#
# MDEV-xxx issue subject
#
```
please, don't change that.
This is a test specifically for `--tls-version` command line option of mariadb command line client

mysql-test/suite.pm
```diff
@@ -72,8 +72,8 @@ sub skip_combinations {
   $skip{'include/have_ssl_communication.inc'} =
   $skip{'include/have_ssl_crypto_functs.inc'} = 'Requires SSL' unless $ssl_lib;
 
-  $skip{'main/openssl_6975.test'} = 'no or wrong openssl version'
-    unless $openssl_ver ge "1.0.1d" and $openssl_ver lt "1.1.1";
```
vuvova, line 76: the second condition was added in 74f184aff253 "Fix tests not to fail with OpenSSL 1.1.1 with TLSv1.3". May be your replace fixes that and the test no longer fails with TLSv1.3. If it doesn't ­— sure, thanks for removing the unnecessary limitation.

vuvova (approved):

mysql-test/main/openssl.test
ottok: The Jira title in question is `MDEV-6975 Implement TLS protocol`. This title does not reflect the purpose and contents of this test, so having it would be misleading. The fact that it was there to begin with sent many people debugging down the wrong path. An MDEV makes sense when the test is to prevent a specific bug from happening again. But when testing basic functionality (like if TLS works) choosing one of hundreds TLS related MDEVs as the "title" does not seem purposeful in documenting what the test is about.

mysql-test/suite.pm
ottok: Indeed, this test was only running if MariaDB was built with an old OpenSSL version. This test is now fixed to be generic, and to run on any version of OpenSSL (and we no longer have the situation that new OpenSSL are used untested).

mysql-test/main/openssl.test
vuvova: this is a bad MDEV title, as written by the user. Something I rewrite MDEV title to be better, in this case, unfortunately, I didn't. You can have the test file saying
```
#
# MDEV-6975 Implement TLS protocol
#
# despite the confusing title, MDEV-6975 is about allowing mariadb command line clients to
# restrict TLS to a specific protocol version, that is about implementing the --tls-protocol
# command line option. And this is what the following tests.
#
```

gkodinov (changes requested):
Thank you for your contribution. I hope you're still inclined to work on this with us after all that time. 

To start with, you will most definitely need to rebase this. I'd say the right version is 12.2. 

Other than that I have no further comments, aside from what has already been said by the current reviewers.

ottok:
This is two years old and very unlikely that the original submitter will return to this anymore.

gkodinov:
It's worth a try, Otto :) I will eventually process if further if that's indeed the case.

gkodinov:
There has been no reply to my review questions for 3+ weeks now. Converting to draft due to inactivity. Please re-open when/if you intend to resume working on this.

PR #3209, opened 2024-04-16
MDEV-33919: Remove less standard format directive an-trap

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-33919*

## Description
Few man pages have less standard format directive: `.it 1 an-trap` which specifying a formatting instruction related to indentation (adds tab in man page in this)

There is no traces what an-trap should do and removing it does not affect rendering of man page

## Release Notes
There should be any mention this on release notes

## How can this PR be tested?
One can render these man pages as ascii with
```
groff -man -Tascii man/mysqlbinlog.1
groff -man -Tascii man/mysqldump.1
groff -man -Tascii man/mysqladmin.1
groff -man -Tascii man/mysqlcheck.1
groff -man -Tascii man/myisamchk.1
groff -man -Tascii man/mysqld_safe.1
groff -man -Tascii man/mysql_upgrade.1
```
Diff old and new should be the same

## PR quality check

### DIFF
--- a/man/myisamchk.1
+++ b/man/myisamchk.1
@@ -38,7 +38,7 @@ with partitioned tables is not supported\&.
 .sp
 .\}
 .RS 4
-.it 1 an-trap
+.it 1
 .nr an-no-space-flag 1
 .nr an-break-flag 1
 .br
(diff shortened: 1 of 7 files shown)


anson1014 (approved):

anson1014:
Couldn't find too much online regarding the "an-trap" directive. If there's no functional difference between the output of the two pages I think it is a good idea to just remove it to avoid confusion.

illuusio:
> Couldn't find too much online regarding the "an-trap" directive. If there's not much documentation on why it is there and there is no functional difference between the output I think it is a good idea to just remove it to avoid confusion.

This was kind of my 'motivation'. I could no find any traces why it should be there and it renders the same.

PR #3211, opened 2024-04-17
MDEV-33447: libpmem is not available in RHEL 8

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33447*
## Description
Because the core Red Hat Enterprise Linux 8 does not include `libpmem`, let us hard-code the necessary part in our code.

`pmem_persist()`: Implement for 64-bit x86, ARM, POWER, RISC-V and Loongarch in a way that should be compatible with the https://github.com/pmem/pmdk/ implementation of `pmem_persist()`.

Note: libpmem had not been previously enabled for Loongarch in our Debian packaging.
## Release Notes
The dependency on `libpmem` was removed by implementing `log_t::pmem_persist()` directly.

On 64-bit systems, the CMake parameter `WITH_INNODB_PMEM` can be set to enable or disable the memory-mapped InnoDB log. It is enabled by default on 64-bit x86, ARM and POWER, and disabled on RISC-V or Loongarch, which are not covered by our current CI infrastructure.
## How can this PR be tested?
I reviewed the generated machine code for the inline assembler snippets related to `pmem_persist()` on https://godbolt.org. It should be noted that GCC blindly copies the output to an external assembler without validating the mnemonics, while `clang` includes a built-in assembler. The GCC output can be validated by enabling a separate option to compile to binary code. This is how the instruction encodings for POWER were validated.

I tested this on two Intel implementations of the AMD64 ISA: one that implements `clflushopt` and `clwb` and another that only implements `clflush`.

This should be covered on all 64-bit CI targets that run tests on Linux `/dev/shm`. Unfortunately, there is no RISC-V or Loongarch coverage in our CI yet, but we do set `WITH_INNODB_PMEM=OFF` by default on those systems.

## PR quality check


storage/innobase/log/log0log.cc
```diff
@@ -87,6 +87,98 @@ void log_t::set_capacity()
...
+  pmem_persister= (getauxval(AT_HWCAP) & HWCAP_DCPOP) ? log_cvap : log_cvac;
+}
+# else
+void log_t::pmem_persist(const void *buf, size_t size)
+{
+#  ifdef __ppc64__
+  for (uintptr_t u= uintptr_t(buf) & ~(CPU_LEVEL1_DCACHE_LINESIZE),
+         end= uintptr_t(buf) + size;
+       u < end; u+= CPU_LEVEL1_DCACHE_LINESIZE)
+    __asm__ __volatile__("dcbf 0,%0,6"::r(u):"memory");
```
dr-m: Based on tests on https://godbolt.org and its "compile to binary code" option and some reading of the LLVM revision history I came up with the following:
```c++
#   if __GNUC__ >= 11 || (defined __clang_major__ && __clang_major__ >= 12)
    __asm__ __volatile__("dcbstps 0,%0"::r(u):"memory");
#   else
    __asm__ __volatile__(".long (0x7cc000AC | %0 << 11)" :: "r"(u):"memory");
#   endif
```
I hope that the GCC version check is an adequate proxy for having a recent enough assembler. In clang, the assembler is built-in. The GCC source code itself does not contain either mnemonic `dcbstps` or `phwsync`.

dr-m (commented):
324ff48b8a0526f903b402561dacc26d200497cb addresses this. I was glad to find out that https://godbolt.org supports an option to generate binary code from the GCC output.

vaintroub:
Would not it be better to factor out  the low-level stuff, assembly and pmem_persist, into own source and header file, and also hide the Intel-specific persister_inits there? Conceptually, it is too low level for log0log where it now lives.

Also, for architectures where the functionality is  not currently tested, riscv , I'd add #if 0 or its equivalent, to avoid surprises when someone compiles and tests there for the first time.

dr-m:
> Would not it be better to factor out the low-level stuff, assembly and pmem_persist, into own source and header file, and also hide the Intel-specific persister_inits there? Conceptually, it is too low level for log0log where it now lives.

There is some detection for ARMv8 as well. I do not know about ARMv8, but on AMD64 the `clflush` instruction alone would technically do the trick; it only is an overkill that will unnecessarily evict the cache lines from all caches, which is not nice if there soon will be a subsequent writes to the last log block. The `clflushopt` and `clwb` are optimizations. All 3 instructions are available on our system where a PMEM device has been installed.

The PMEM interface is specific to the log subsystem and nothing else. I think that this deep coupling is the only way to avoid unnecessary overhead.

> Also, for architectures where the functionality is not currently tested, riscv , I'd add #if 0 or its equivalent, to avoid surprises when someone compiles and tests there for the first time.

I think that this can be best addressed by having a CMake option that allows the logic to be disabled. Not exactly `WITH_PMEM`, because its `OFF` meant “enabled or disabled, based on whether `libpmem` is available”, but something else. This would seem to require moving the architecture detection logic from `log0types.h` to `CMakeLists.txt` and invoking the compiler with `-DHAVE_PMEM`.

storage/innobase/log/log0log.cc
```diff
@@ -87,6 +87,117 @@ void log_t::set_capacity()
...
+inline void log_t::pmem_persister_init()
+{
+  uint32_t eax= 0, ebx= 0, ecx= 0, edx= 0;
+  __cpuid_count(7, 0, eax, ebx, ecx, edx);
+  if (ebx & 1U<<24 /* CLWB */)
+    pmem_persister= log_clwb;
+  else if (ebx & 1U<<23 /* CLFLUSHOPT */)
+    pmem_persister= log_clflushopt;
+  else
+    pmem_persister= log_clflush;
```
dr-m, line 132: Initially I had `__cpuid(7, eax, ebx, ecx, edx)` here. It would fail to initialize the `ecx` register to 0, and hence all the 4 registers were 0 after executing the `cpuid` instruction. After I fixed that, I got the expected result (`log_clwb`) on our PMEM-equipped system. My development system only implements `clflush`.

vaintroub:
> > Would not it be better to factor out the low-level stuff, assembly and pmem_persist, into own source and header file, and also hide the Intel-specific persister_inits there? Conceptually, it is too low level for log0log where it now lives.
> The PMEM interface is specific to the log subsystem and nothing else. I think that this deep coupling is the only way to avoid unnecessary overhead.

I do not think overhead of a function call is important, for an interface that is essentially dead now.  It is just so, that software has layers, and inline assembly or  `#ifdef __x86_64__`  or even testing for OS is not appropriate for every level.  There would be no harm if you had log0pmem.cc file, I think. It is still log specific,  but not everything in a single file. We dont put whole innodb into single file, even if it would be tempting, and compiler can do better optimization

dr-m:
b582b04a06b2aa02ccb829902b71061846883129 moves the added code into separate files `cache.h` and `cache.cc`, so that at the high level we are only replacing two `#include <libpmem.h>` with `#include "cache.h"`. On AMD64 and ARMv8, `pmem_persist` is a macro that expands to the function pointer `pmem.persist`.

vaintroub (approved):

PR #3215, opened 2024-04-18
Gtid_log_event thread_id missing in binlog when pseudo_thread_id is 0

bnestere:
If pseudo_thread_id is set to 0 (or implicitly set to 0 because it was attempted to be set to a negative value), then the thread_id attribute of the Gtid_log_event will not be written at all, whereas the Query_log_event still will write thread_id=0.  The underlying problem was that the
FL_EXTRA_THREAD_ID would only be added if thread_id was greater than 0.

This patch fixes this by changing the condition to mark the FL_EXTRA_THREAD_ID when generating a GTID log event, such that it will always be written on the primary, and if on the slave, the rpl_group_info's gtid_ev_flags_extra is checked to see if the flag was present when reading it in.


bnestere:
This was squashed into the main MDEV-7850 commit into 11.5: e4afa610539ae01164485554e2de839bea9de816

PR #3216, opened 2024-04-22
fix build with WITH_EXTRA_CHARSETS=none in cmake

mscdex:

## Description

This fixes cmake-based builds with -DWITH_EXTRA_CHARSETS=none, which results in a CHARSETS list that does _not_ include ucs2, utf16, or utf32.

## Release Notes
"Building with CMake and -DWITH_EXTRA_CHARSETS=none now works"

## How can this PR be tested?

Compiling with cmake with -DWITH_EXTRA_CHARSETS=none.

Erm well... the contributing info says to target the "master branch" but no such thing exists, so I targeted the current default branch (11.5 as of this writing). However, the problem exists in all previous version branches as well (I was trying to compile 10.11).

## PR quality check

### DIFF
--- a/strings/ctype-uca.c
+++ b/strings/ctype-uca.c
@@ -39437,18 +39437,24 @@ my_uca1400_collation_definition_init(MY_CHARSET_LOADER *loader,
     *dst= nopad ? my_charset_utf8mb4_unicode_520_nopad_ci :
                   my_charset_utf8mb4_unicode_520_ci;
     break;
+#ifdef HAVE_CHARSET_ucs2
   case MY_CS_ENCODING_UCS2:
     *dst= nopad ? my_charset_ucs2_unicode_520_nopad_ci :
                   my_charset_ucs2_unicode_520_ci;
     break;
+#endif
+#ifdef HAVE_CHARSET_utf16
   case MY_CS_ENCODING_UTF16:
     *dst= nopad ? my_charset_utf16_unicode_520_nopad_ci :
                   my_charset_utf16_unicode_520_ci;
     break;
+#endif
+#ifdef HAVE_CHARSET_utf32
   case MY_CS_ENCODING_UTF32:
     *dst= nopad ? my_charset_utf32_unicode_520_nopad_ci :
                   my_charset_utf32_unicode_520_ci;
     break;
+#endif
   }
 
   dst->number= id;


LinuxJedi:
Hi mscdex,

> Erm well... the contributing info says to target the "master branch" but no such thing exists, so I targeted the current default branch (11.5 as of this writing). However, the problem exists in all previous version branches as well (I was trying to compile 10.11).

Can you please point to where we mention a master branch so that we can fix this? We definitely don't have one.

It needs to target the oldest supported version of MariaDB because commits are merged up from there to the newer versions. In this case the bug appears to be introduced in 10.10 but 10.11 is the oldest supported version with it, in which case the target should be 10.11.

We can do the rebase for you if you wish, or you can rebase this change against 10.11 in this PR.

mscdex:
> Can you please point to where we mention a master branch so that we can fix this? We definitely don't have one.

I believe it was this line from [here](https://mariadb.org/get-involved/getting-started-for-developers/):

```blockquote
In general all development is done on the master branch first, and later high-priority bug fixes might be backported to stable release branches (10.5, 10.4, 10.3, 10.2, 10.1).
```

> rebase this change against 10.11 in this PR

Done.

grooverdan:
While I'm all for a fix, and it looks good. What's the use case of WITH_EXTRA_CHARSETS=none for you @mscdex ? It looks like someone many years ago went in a very liberal proving of compile options, so I'm slightly surprised they are used.

grooverdan (approved):

mscdex:
> What's the use case of WITH_EXTRA_CHARSETS=none for you @mscdex ?

Just trying to slim down a generated OS image everywhere I can and the extra character sets I don't need as really only utf8mb4 and latin1 will be used.

LinuxJedi:
> > Can you please point to where we mention a master branch so that we can fix this? We definitely don't have one.
> 
> I believe it was this line from [here](https://mariadb.org/get-involved/getting-started-for-developers/):
> 
> ```
> In general all development is done on the master branch first, and later high-priority bug fixes might be backported to stable release branches (10.5, 10.4, 10.3, 10.2, 10.1).
> ```

Many thanks! Not sure who wrote it, but this one is definitely incorrect. I'll get this fixed later today.

LinuxJedi (approved):
Thanks again for the information and for moving it to 10.11 it so quickly.

grooverdan:
@mscdex to merge this we need either "Maintainers are allowed to edit this pull request." on the PR so we can rebase this, or if you rebase this yourself we've got a small window to enable auto-merge before someone else commits to the 10.11 branch.

Thanks for your contribution.

mscdex:
I've rebased.

grooverdan:
thanks. Automerge renabled.

PR #3217, opened 2024-04-23
MDEV-33334  mariabackup does not always write encrypted files to backup location

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-33334*

## Description
Problem:
========
mariabackup --prepare fails to write the pages in encrypted format. This issue happens only for default encrypted table when innodb_encrypt_tables variable is enabled.

Fix:
====
backup process should write the value of innodb_encrypt_tables variable in configuration file. prepare should enable the variable based on configuration file.

## How can this PR be tested?
./mtr mariabackup.encrypted_export

## PR quality check

### DIFF
--- a/extra/mariabackup/encryption_plugin.cc
+++ b/extra/mariabackup/encryption_plugin.cc
@@ -71,7 +71,16 @@ static std::string get_encryption_plugin_from_cnf()
       plugin_load = line + 12;
       // remote \n at the end of string
       plugin_load.resize(plugin_load.size() - 1);
-      break;
+    }
+
+    if (strncmp(line, "innodb_encrypt_tables=", 22) == 0)
+    {
+      if (!strncmp(line + 22, "ON", 2) ||
+          !strncmp(line + 22, "1", 1))
+        srv_encrypt_tables= 1;
+      else if (!strncmp(line + 22, "FORCE", 5) ||
+               !strncmp(line + 22, "2", 1))
+        srv_encrypt_tables= 2;
     }
   }
   fclose(f);
@@ -154,6 +163,17 @@ void encryption_plugin_backup_init(MYSQL *mysql)
 
   mysql_free_result(result);
 
+  result = xb_mysql_query(mysql, "select @@innodb_encrypt_tables", true, true);
+  row = mysql_fetch_row(result);
+  if (!row);
+  else if (const char *r= row[0])
+  {
+    if (!strcmp(r, "ON")) srv_encrypt_tables= 1;
+    else if (!strcmp(r, "FORCE")) srv_encrypt_tables= 2;
+    oss << "innodb_encrypt_tables=" << r << std::endl;
+  }
+
+  mysql_free_result(result);
   encryption_plugin_config = oss.str();
 
   argc = 0;
--- a/mysql-test/suite/mariabackup/encrypted_export.opt
+++ b/mysql-test/suite/mariabackup/encrypted_export.opt
@@ -1,0 +1,6 @@
+--innodb_encrypt_tables=1
+--plugin-load-add=$FILE_KEY_MANAGEMENT_SO
+--loose-file-key-management
+--loose-file-key-management-filekey=FILE:$MTR_SUITE_DIR/filekeys-data.key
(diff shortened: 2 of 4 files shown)


extra/mariabackup/encryption_plugin.cc
```diff
@@ -161,6 +168,14 @@ void encryption_plugin_backup_init(MYSQL *mysql)
     oss << "innodb_encrypt_log=" << row[0] << std::endl;
 
     mysql_free_result(result);
+
+    /* Check whether to encrypt tables */
+    result = xb_mysql_query(mysql, "select @@innodb_encrypt_tables", true, true);
+    row = mysql_fetch_row(result);
+    srv_encrypt_tables = (row != 0 && row[0][0] == '1');
+    oss << "innodb_encrypt_tables=" << row[0] << std::endl;
```
dr-m, line 176: This seems to be interpreting any response starting with `1` as true. When can `mysql_fetch_row` return `nullptr`, or when can it return a pointer to `nullptr`? The check `row != 0` looks redundant, because in the following line we are dereferencing `row` anyway. And we fail to check if `row[0]` is a null pointer.

With some error handling (I am not sure if it is correct or redundant), it could be clearer to write
```c++
if (!row);
else if (const char *r= row[0])
{
  srv_encrypt_tables = !strcmp(r, "1");
  oss << "innodb_encrypt_tables=" << r << std::endl
}
```

extra/mariabackup/encryption_plugin.cc
```diff
@@ -72,7 +72,14 @@ static std::string get_encryption_plugin_from_cnf()
...
       plugin_load.resize(plugin_load.size() - 1);
-      break;
+    }
+
+    if (strncmp(line, "innodb_encrypt_tables=", 21) == 0)
+    {
+      char *val = line + 22;
+      if (strncmp(val, "ON", 2) == 0)
+        srv_encrypt_tables= 1;
+      else srv_encrypt_tables= 0;
```
dr-m, line 82: Is there any Boolean parsing function that we could invoke here? It would seem that `encryption_plugin_backup_init()` is writing `0` or `1`, which also are valid values. Other valid values would be `on`, `off`, `Off`, and I could imagine that `yes`, `no`, `true`, `false` might work as well.

dr-m (commented):
I think that this needs a review from someone who is familiar with the configuration parsing.

Is the bug reproducible in older versions, such as 10.5? If not, why not?

dr-m:
Based on my analysis in [MDEV-33438](https://jira.mariadb.org/browse/MDEV-33438), this should affect all MariaDB versions, starting from 10.4. The fix should be similar.

extra/mariabackup/encryption_plugin.cc
```diff
@@ -71,7 +71,16 @@ static std::string get_encryption_plugin_from_cnf()
...
+    }
+
+    if (strncmp(line, "innodb_encrypt_tables=", 22) == 0)
+    {
+      if (!strncmp(line + 22, "ON", 2)
+          || !strncmp(line + 22, "1", 1))
+        srv_encrypt_tables= 1;
+      else if (!strncmp(line + 22, "FORCE", 5)
+               || !strncmp(line + 22, "2", 1))
+        srv_encrypt_tables= 2;
```
dr-m, line 83: In the common formatting style, lines should be split after binary operators such as `||`. Only Heikki wanted to split before them.

extra/mariabackup/encryption_plugin.cc
```diff
@@ -154,6 +163,17 @@ void encryption_plugin_backup_init(MYSQL *mysql)
...
   mysql_free_result(result);
 
+  result = xb_mysql_query(mysql, "select @@innodb_encrypt_tables", true, true);
+  row = mysql_fetch_row(result);
+  if (!row);
+  else if (const char *r= row[0])
+  {
+    if (!strcmp(r, "ON")) srv_encrypt_tables= 1;
+    else if (!strcmp(r, "FORCE")) srv_encrypt_tables= 2;
+    oss << "innodb_encrypt_tables="<< r << std::endl;
```
dr-m, line 173: There is a missing space after the end of the string literal.

PR #3221, opened 2024-04-24
MDEV-33669 mariabackup --backup hangs

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33669*

## Description
This is a server hang and not an issue with backup. While concurrent DDLs in server gets in hanged state, mariabackup waits for DDLs to finish trying to acquire MDL_BACKUP_BLOCK_DDL.

The server hang is serious in nature and caused by thread pool state being incorrectly set to thread creation pending state while no creation is actually pending. Once a thread pool reaches such state no new thread gets created in the pool.

While it could possibly affect all thread pools in server, the innodb thread pool is the victim in current bug where IO job gets blocked when the pool is stuck with much less number of threads than intended. Available workers are blocked in purge waiting for page lock to be released by IO write (SX lock) causing a complete deadlock.

The issue is caused by the state variable m_thread_creation_pending introduced by MDEV-31095: 9e62ab7aafbf. We check and set the variable early while attempting to create a new thread in pool but fail to reset it if we exit the flow for other reasons like maximum threads reached or get into thread creation throttling path.

Fix: The simple fix is to make sure that the state is reset back in case we don't actually attempt to create the thread.

## Release Notes
To Be Updated

## How can this PR be tested?

It is hard to create an automated test for the issue as it involves good lot of background threads and concurrency among them is not easy to control. 

Currently the issue is repeatable in RQG setup as explained in the MDEV.

## PR quality check

### DIFF
--- a/tpool/tpool_generic.cc
+++ b/tpool/tpool_generic.cc
@@ -722,9 +722,6 @@ static int  throttling_interval_ms(size_t n_threads,size_t concurrency)
 /* Create a new worker.*/
 bool thread_pool_generic::add_thread()
 {
-  if (m_thread_creation_pending.test_and_set())
-    return false;
-
   size_t n_threads = thread_count();
 
   if (n_threads >= m_max_threads)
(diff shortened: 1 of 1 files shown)


dr-m (approved):
I trust your debugging but would leave the final judgement to @vaintroub, who created this code.

vaintroub (approved):
Looks good. Thanks for fixing it!

PR #3222, opened 2024-04-25
MDEV-22855 Assertion `!field->prefix_len || field->fixed_len == field->prefix_len` failed in btr_node_ptr_max_size

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-22855*

## Description
Problem:
========
- InnoDB wrongly calulates the record size in btr_node_ptr_max_size() when prefix index of
the column has to be stored externally.

Fix:
====
- InnoDB should add the maximum field size to record size when the field is a fixed length one.

## How can this PR be tested?
./mtr innodb.innodb-index

# Basing the PR against the correct MariaDB version

## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -1126,7 +1126,7 @@ static ulint btr_node_ptr_max_size(const dict_index_t* index)
 		/* Determine the maximum length of the index field. */
 
 		field_max_size = dict_col_get_fixed_size(col, comp);
-		if (field_max_size) {
+		if (field_max_size && field->fixed_len) {
 			/* dict_index_add_col() should guarantee this */
 			ut_ad(!field->prefix_len
 			      || field->fixed_len == field->prefix_len);
--- a/mysql-test/suite/innodb/r/index_length.result
+++ b/mysql-test/suite/innodb/r/index_length.result
@@ -20,4 +20,12 @@ FLOOR(index_length/@@innodb_page_size)
 2
 disconnect stop_purge;
 DROP TABLE t1;
+#
+# MDEV-22855  Assertion (!field->prefix_len ||
+#        field->fixed_len == field->prefix_len)
+#        failed in btr_node_ptr_max_size
+#
+CREATE TABLE t1(c CHAR(194) CHARACTER SET UTF32, KEY k1(c(193)))ENGINE=InnoDB;
+INSERT INTO t1 SET c='';
+DROP TABLE t1;
 # End of 10.4 tests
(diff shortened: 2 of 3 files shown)


storage/innobase/btr/btr0cur.cc
```diff
@@ -1126,7 +1126,7 @@ static ulint btr_node_ptr_max_size(const dict_index_t* index)
 		/* Determine the maximum length of the index field. */
 
 		field_max_size = dict_col_get_fixed_size(col, comp);
-		if (field_max_size) {
+		if (field_max_size && field->fixed_len) {
```
dr-m, line 1129: I remember that any `CHAR` that is longer than 255 bytes would in fact be stored in the same way as `VARCHAR`. If that is the case, then I think that the problem should be reproducible already with a 64-char prefix of a UTF-32 encoded column. Can you try to find out where exactly we would assign the `field->fixed_length`, and add a comment here that explains when we may have `field->fixed_len==0` for something that actually is a fixed-length field?

mysql-test/suite/innodb/t/innodb-index.test
```diff
@@ -1202,3 +1202,13 @@ call mtr.add_suppression("Plugin 'InnoDB' init function returned error");
...
 
 --enable_query_log
+
+--echo #
+--echo # MDEV-22855 Assertion (!field->prefix_len ||
+--echo #	field->fixed_len == field->prefix_len)
+--echo #	failed in btr_node_ptr_max_size
+--echo #
+CREATE TABLE t1(c CHAR(255) CHARACTER SET UTF32, KEY k1(c(250)))ENGINE=InnoDB;
+INSERT INTO t1 VALUES(100000);
```
dr-m, line 1212: This test is a little obfuscated: we are implicitly converting the numeric constant into a string literal `'100000'`. I would suggest just writing the following, if it reproduces the problem:
```
INSERT INTO t1 SET c='';
```
Also, can you try to replace the 255 and 250 with the minimum needed sizes to trigger the bug?

dr-m (commented):
Thank you. The fix looks OK, but I would like to clarify it a little, because I do not remember where exactly the decision to store long `CHAR` as `VARCHAR` has been implemented.

storage/innobase/btr/btr0cur.cc
Thirunarayanan: In dict_index_add_col():
```
      /* Long fixed-length fields that need external storage are treated as
        variable-length fields, so that the extern flag can be embedded in
        the length word. */

        if (field->fixed_len > DICT_MAX_FIXED_COL_LEN) {
                field->fixed_len = 0;
        }
```
This was added in mysql-5.6, I guess. we need the prefix length to greater than 768 at least to repeat the scenario.

dr-m (approved):

PR #3225, opened 2024-04-26
MDEV-36657: add ${PCRE_LIBRARY_PATH} to link directories 

keeper:
- [x] *The Jira issue number for this PR is: MDEV-29955*

## Description
Commit 3a33ae86 uses pkg-config to find the location of PCRE library.
However when a user installs the library in non-standard location
(/workspace/mylib for example), it's also recommended to use the
'-Wl,-rpath -Wl,LIBDIR' linker flag to point the build system to where
the library installed.

In the revision, it's now compatible when `libpcre2-8.pc` is not available. In this case `PCRE_LIBRARY_DIRS` is empty and it still relies on users to set the path correctly so CMake can find the shared object. If everything is in place, `FIND_LIBRARY()` macro should be able to find the library, and we can use the path to set the link directories.

## How can this PR be tested?
* [x] Builds with no errors with PCRE installed in custom path

Manual checks:
1. pkg-config:
```
$ PKG_CONFIG_PATH="/workspace/lib/lib/pkgconfig:/usr/lib64/pkgconfig" pkg-config --variable=libdir libpcre2-8
/workspace/lib/lib

$ PKG_CONFIG_PATH="/workspace/lib/lib/pkgconfig:/usr/lib64/pkgconfig" pkg-config --variable=libdir libpcre2-posix
/workspace/lib/lib
```

2. Generated CMake build file
```
$ PKG_CONFIG_PATH="/workspace/lib/lib/pkgconfig:/usr/lib64/pkgconfig" cmake -S . -B build/ -G Ninja --fresh -DCMAKE_PREFIX_PATH="/workspace/lib" -DWITH_PCRE=system
...
...
$ grep LINK_LIBRARIES build/build.ninja
...
...
LINK_LIBRARIES = -Wl,-rpath,/workspace/lib/lib ...
```

3. When `libpcre2-8.pc` is available (in custom path):
```
$ PKG_CONFIG_PATH="/workspace/lib/lib/pkgconfig:/usr/lib64/pkgconfig" cmake -S . -B build/ -G Ninja --fresh -DCMAKE_PREFIX_PATH="/workspace/lib" -DWITH_PCRE=system
...
...
-- Checking for module 'libpcre2-8'   
--   Found libpcre2-8, version 10.44            
-- PCRE_LIBRARY_DIRS: /workspace/lib/lib  
```

4. When `libpcre2-8.pc` is not available (in custom path):
```
$ cmake -S . -B build/ -G Ninja --fresh -DCMAKE_PREFIX_PATH="/workspace/lib" -DWITH_PCRE=system
...
...
-- Checking for module 'libpcre2-8'                                                                                            
--   Package 'libpcre2-8', required by 'virtual:world', not found    
-- Looking for pcre2_match_8 in pcre2-8
-- Looking for pcre2_match_8 in pcre2-8 - found 
-- PCRE_LIBRARY_DIRS: /workspace/lib/lib 
```
(The debug message `PCRE_LIBRARY_DIRS: ...` is removed in the PR)

## PR quality check


LinuxJedi (approved):
Looks like a good catch to me.

cmake/pcre.cmake
```diff
@@ -85,6 +85,7 @@ MACRO (CHECK_PCRE)
       PKG_CHECK_MODULES(PCRE libpcre2-8)
       # in case pkg-config or libpcre2-8.pc is not installed:
       CHECK_LIBRARY_EXISTS(pcre2-8 pcre2_match_8 "${PCRE_LIBRARY_DIRS}" HAVE_PCRE2_MATCH_8)
+      link_directories(${PCRE_LIBRARY_DIRS})
```
vuvova, line 88: does it work "in case pkg-config or libpcre2-8.pc is not installed" ?

vuvova: I am not sure it's a big problem — not working "in case pkg-config or libpcre2-8.pc is not installed".

But the comment two lines above explicitly says `CHECK_LIBRARY_EXISTS` is to make the code work in this case. If it doesn't — the comment should be removed or changed and `CHECK_LIBRARY_EXISTS` might need to be removed if it's redundant (but perhaps it's still needed for `build_depends.cmake` — then it should stay and the comment should say that it's for build dependencies)

vuvova (changes requested):

cmake/pcre.cmake
keeper: @vuvova Sorry for the hiatus. I feel we should still keep `CHECK_LIBRARY_EXISTS()` and make it compatible when pc files are not available in edge cases. I've reworked the PR and use `FIND_LIBRARY()` when it's missing and it should work for both case now.

vuvova:
Strange. I've installed pcre2 locally, run cmake with your patch and there were no `-Wl,-rpath` anywhere

cmake/pcre.cmake
```diff
@@ -83,8 +83,17 @@ MACRO (CHECK_PCRE)
...
       PKG_CHECK_MODULES(PCRE libpcre2-8)
-      # in case pkg-config or libpcre2-8.pc is not installed:
-      CHECK_LIBRARY_EXISTS(pcre2-8 pcre2_match_8 "${PCRE_LIBRARY_DIRS}" HAVE_PCRE2_MATCH_8)
+      IF(NOT "${PCRE_LIBRARY_DIRS}" STREQUAL "")
+        SET(HAVE_PCRE2_MATCH_8 1)
+      ELSE()
+        # in case pkg-config or libpcre2-8.pc is not installed:
+        CHECK_LIBRARY_EXISTS(pcre2-8 pcre2_match_8 "" HAVE_PCRE2_MATCH_8)
+        IF(HAVE_PCRE2_MATCH_8)
+          FIND_LIBRARY(PCRE_LINK_LIBRARIES pcre2-8)
```
vuvova, line 92: I don't see how it could work. `-DCMAKE_PREFIX_PATH` only affects `FIND_LIBRARY()` but not `CHECK_LIBRARY_EXISTS()`, so the latter will fail if the the library is not in the standard linker path, and then `HAVE_PCRE2_MATCH_8` will be `NOT-FOUND` and `FIND_LIBRARY()` won't be used at all, whether `-DCMAKE_PREFIX_PATH` is set or not.

I'd simply remove `CHECK_LIBRARY_EXISTS()` and replace it with `FIND_LIBRARY()`

vuvova (changes requested):

vuvova:
@keeper, it's been more than a month since my review. We'll give it one more month though.

keeper:
Again sorry for taking so long. I've made the change and some improvements and rebase to latest main.

When `libpcre2-8.pc` is available (in custom path):
```
-- Looking for include files curses.h, term.h - found
-- Checking for module 'libpcre2-8'
--   Found libpcre2-8, version 10.46
-- Looking for PCRE2regcomp in pcre2-posix
...
```

In `CMakeCache.txt`:
```
PCRE_LIBRARY_DIRS:INTERNAL=/tmp/test-pcre2-install/lib64
CMAKE_PREFIX_PATH:UNINITIALIZED=/tmp/test-pcre2-install
pkgcfg_lib_PCRE_pcre2-8:FILEPATH=/tmp/test-pcre2-install/lib64/libpcre2-8.so
PCRE_CFLAGS:INTERNAL=-I/tmp/test-pcre2-install/include
PCRE_INCLUDEDIR:INTERNAL=/tmp/test-pcre2-install/include
PCRE_INCLUDE_DIRS:INTERNAL=/tmp/test-pcre2-install/include
PCRE_LDFLAGS:INTERNAL=-L/tmp/test-pcre2-install/lib64;-lpcre2-8
PCRE_LIBDIR:INTERNAL=/tmp/test-pcre2-install/lib64
PCRE_LIBRARY_DIRS:INTERNAL=/tmp/test-pcre2-install/lib64
PCRE_PREFIX:INTERNAL=/tmp/test-pcre2-install
PCRE_STATIC_CFLAGS:INTERNAL=-I/tmp/test-pcre2-install/include
PCRE_STATIC_INCLUDE_DIRS:INTERNAL=/tmp/test-pcre2-install/include
PCRE_STATIC_LDFLAGS:INTERNAL=-L/tmp/test-pcre2-install/lib64;-lpcre2-8
PCRE_STATIC_LIBRARY_DIRS:INTERNAL=/tmp/test-pcre2-install/lib64
```

When `libpcre2-8.pc` is not available (in custom path):
```
...
-- Checking for module 'libpcre2-8'
--   Package 'libpcre2-8' not found
-- PCRE: pkg-config not available or failed, using FIND_LIBRARY fallback: /tmp/test-pcre2-install/lib64/libpcre2-8.so
-- Looking for PCRE2regcomp in pcre2-posix
...
```

And binary linkage:
```
# ldd sql/mariadbd
        linux-vdso.so.1 (0x00007ffcdd536000)                                                                                                                                                                                                                                          libpcre2-8.so.0 => /tmp/test-pcre2-install/lib64/libpcre2-8.so.0 (0x0000753c244ad000)
        libcrypt.so.2 => /lib64/libcrypt.so.2 (0x0000753c24471000)
...
```

The test script I use:
```
#!/bin/bash
set -euo pipefail

# STEP 1: Setup paths
WORKDIR=$(pwd)
PCRE2_SRC_DIR="$WORKDIR/pcre2-src"
PCRE2_INSTALL_DIR="/tmp/test-pcre2-install"
MARIADB_SRC_DIR="$WORKDIR/mariadb-server"  # You must patch it beforehand
MARIADB_BUILD_DIR="$MARIADB_SRC_DIR/build"

echo "==> Cleaning old test dirs..."
rm -rf "$PCRE2_SRC_DIR" "$PCRE2_INSTALL_DIR" "$MARIADB_BUILD_DIR"

# STEP 2: Build and install PCRE2 to non-standard path
echo "==> Cloning PCRE2..."
git clone --depth=1 https://github.com/PhilipHazel/pcre2.git "$PCRE2_SRC_DIR"

echo "==> Building PCRE2 to $PCRE2_INSTALL_DIR..."
cmake -S "$PCRE2_SRC_DIR" -B "$PCRE2_SRC_DIR/build" -DCMAKE_INSTALL_PREFIX="$PCRE2_INSTALL_DIR" \
  -DBUILD_SHARED_LIBS=ON \
  -DBUILD_STATIC_LIBS=OFF \
  -DPCRE2_BUILD_PCRE2GREP=OFF \
  -DPCRE2_BUILD_TESTS=OFF

cmake --build "$PCRE2_SRC_DIR/build" --target install
# Uncomment to test fallback logic
# mv "$PCRE2_INSTALL_DIR/lib64/pkgconfig/libpcre2-8.pc" "$PCRE2_INSTALL_DIR/lib64/pkgconfig/libpcre2-8.pc.bak"
# export PKG_CONFIG_PATH=""
# export PKG_CONFIG_SYSTEM_INCLUDE_PATH=""
# export PKG_CONFIG_LIBDIR=/tmp/empty-pkgconfig

# STEP 3: Configure MariaDB with custom CMAKE_PREFIX_PATH
echo "==> Configuring MariaDB..."
CMAKE_EXPORT_COMPILE_COMMANDS=ON cmake -S "$MARIADB_SRC_DIR" -B "$MARIADB_BUILD_DIR" \
  -DWITH_PCRE=system \
  -DCMAKE_PREFIX_PATH="$PCRE2_INSTALL_DIR"

# STEP 4: Check results
echo "==> Verifying CMake results..."
grep -E 'PCRE_LIBRARY_DIRS|PCRE_LINK_LIBRARIES|HAVE_PCRE2_MATCH_8' "$MARIADB_BUILD_DIR/CMakeCache.txt" || true
grep -E 'PCRE_LINK_LIBRARIES|HAVE_PCRE2_MATCH_8' "$MARIADB_BUILD_DIR/CMakeCache.txt" || true
grep -E 'test-pcre2-install' "$MARIADB_BUILD_DIR/CMakeCache.txt"

echo "==> Done. MariaDB configured with PCRE2 from non-standard path."
```

vaintroub:
I do not  know why it should be so  complicated

https://github.com/PCRE2Project/pcre2/blob/master/cmake/pcre2-config.cmake.in says this is how you're supposed to use it
  find_package(PCRE2 CONFIG)

rather than the mind boggling detours via pkg-config.
This comes from the creator of the pcre2,  thus trustworthy.

cmake/pcre.cmake
```diff
@@ -89,10 +89,22 @@ ENDMACRO()
 MACRO (CHECK_PCRE)
   IF (NOT TARGET pcre2 AND NOT PCRE_FOUND)
     IF(WITH_PCRE STREQUAL "system" OR WITH_PCRE STREQUAL "auto")
+      SET(HAVE_PCRE2_MATCH_8 0)
       FIND_PACKAGE(PkgConfig QUIET)
       PKG_CHECK_MODULES(PCRE libpcre2-8)
-      # in case pkg-config or libpcre2-8.pc is not installed:
-      CHECK_LIBRARY_EXISTS(pcre2-8 pcre2_match_8 "${PCRE_LIBRARY_DIRS}" HAVE_PCRE2_MATCH_8)
+      IF(NOT "${PCRE_LIBRARY_DIRS}" STREQUAL "")
+        SET(PCRE_LINK_LIBRARIES pcre2-8)
```
vuvova: Why? `PKG_CHECK_MODULES` has set both `PCRE_LIBRARY_DIRS` and `PCRE_LINK_LIBRARIES`

cmake/pcre.cmake
```diff
@@ -89,10 +89,22 @@ ENDMACRO()
...
+        SET(HAVE_PCRE2_MATCH_8 1)
+      ELSE()
+        # in case pkg-config or libpcre2-8.pc is not installed:
+        FIND_LIBRARY(PCRE_LINK_LIBRARIES pcre2-8)
+        IF(PCRE_LINK_LIBRARIES)
+          MESSAGE(STATUS "PCRE: pkg-config not available or failed, using FIND_LIBRARY fallback: ${PCRE_LINK_LIBRARIES}")
+          GET_FILENAME_COMPONENT(PCRE_LIBRARY_DIRS ${PCRE_LINK_LIBRARIES} DIRECTORY)
```
vuvova, line 103: what's the point of this?

cmake/pcre.cmake
keeper: Correct, I'll remove the line

cmake/pcre.cmake
keeper: `find_library` doesn't set up `PCRE_LIBRARY_DIRS` variable, which is used at [Line 107](https://github.com/MariaDB/server/commit/9a212a7b11ab4dbb19295889d23743441e766cff#diff-8e35291051929131abcdfd960c0e4e246317876c5ec2989154a8e810d489c7d6R107) so I try to set the path here when PCRE is installed in a custom path.

vuvova: but you extract it from the `PCRE_LINK_LIBRARIES`, that is, `PCRE_LINK_LIBRARIES` has the full custom path and the library name. The linker won't use the path from `-L` if the library is specified with the full path.

vaintroub: CMake way is having full path to libraries, not setting LINK_DIRECTORIES. It is unclear why it is needed

gkodinov: @vuvova this is the whole premise of the diff I guess: it needs to add to LINK_DIRECTORIES when the PRCE is in a non-default location:

> However when a user installs the library in non-standard location
(/workspace/mylib for example), it's also recommended to use the
'-Wl,-rpath -Wl,LIBDIR' linker flag to point the build system to where
the library installed.
In the revision, it's now compatible when libpcre2-8.pc is not available. In this case PCRE_LIBRARY_DIRS is empty and it still relies on users to set the path correctly so CMake can find the shared object. If everything is in place, FIND_LIBRARY() macro should be able to find the library, and we can use the path to set the link directories.

vaintroub: What part of CMake recommends to use LIBDIR

gkodinov:
FWIW, this is a real problem on MacOSX (where the pcre2 is in homebrew). I've just checked with this fix and now I can successfully compile with -DWITH_PCRE=system and this diff. Previously only -DWITH_PCRE=bundled worked for me.

vaintroub:
We can't use this logic with every external library. If finding system libraries does not work with homebrew, it needs to be fixed to work with homebrew.

gkodinov:
> We can't use this logic with every external library. If finding system libraries does not work with homebrew, it needs to be fixed to work with homebrew.

It's not about brew specifically, Wlad. It's about libraries installed in non-default locations. brew is just an example of that. FWIW, this is what my brew reports:

```
blackbook:~/dev$pkg-config libpcre2-8 --libs
-L/opt/homebrew/Cellar/pcre2/10.47_1/lib -lpcre2-8
blackbook:~/dev$pkg-config --variable=libdir libpcre2-8         
/opt/homebrew/Cellar/pcre2/10.47_1/lib
```

vaintroub:
> > We can't use this logic with every external library. If finding system libraries does not work with homebrew, it needs to be fixed to work with homebrew.
> 
> It's not about brew specifically, Wlad. It's about libraries installed in non-default locations. brew is just an example of that. FWIW, this is what my brew reports:
> 
> ```
> blackbook:~/dev$pkg-config libpcre2-8 --libs
> -L/opt/homebrew/Cellar/pcre2/10.47_1/lib -lpcre2-8
> blackbook:~/dev$pkg-config --variable=libdir libpcre2-8         
> /opt/homebrew/Cellar/pcre2/10.47_1/lib
> ```

C'mon, those are standard homebrew path  This fix can look somewhat like here https://github.com/vaintroub/sysbench/blob/main/CMakeLists.txt#L8C1-L17C8 , which needs to be put into appropriate place , under cmake/os. Someone removed mac from it and it needs to be restored.

gkodinov:
It's been more than 3 weeks since there was no reply to mine and Wlad's comments. Converting to draft. Please convert it back to open when/if you intend to resume working on this.

PR #3234, opened 2024-04-30
MDEV-33543 Server hang caused by InnoDB change buffer

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-33543*

## Description
Issue: When getting a page (buf_page_get_gen) with no latch option (RW_NO_LATCH), the caller is not expected to follow the B-tree latching order. However in buf_page_get_low we try to acquire shared page latch unconditionally to wait for a page that is being loaded by another thread concurrently. In general it could lead to latch order violation and deadlock.

Currently it affects the change buffer insert path btr_latch_prev() which tries to load the previous page out of order with RW_NO_LATCH and two concurrent inserts into IBUF tree cause deadlock. This problem is introduced in 10.6 by MDEV-27058.

Fix: While trying to latch a page with RW_NO_LATCH, always use the "*lock_try" interface and retry operation on failure after unfixing the page.

## Release Notes
MDEV-27058 had introduced this regression in 10.6.6. which could hang the server under load If innodb_change_buffering is enabled and user has tables with non unique secondary index.

## How can this PR be tested?

It is difficult to maintain an automated mtr test in our regular test suite. The RQG test and the attached code/test patch in MDEV can repeat the issue.

The MDEV has a repeatable test and steps for manual reproduction of the issue.

## PR quality check


storage/innobase/buf/buf0buf.cc
```diff
@@ -2579,7 +2579,16 @@ buf_page_get_low(
...
+		if (rw_latch != RW_NO_LATCH) {
+			block->page.lock.s_lock();
+
+		/* For RW_NO_LATCH, we should not try to acquire S or X latch
+		directly as we could be violating the latching order resulting
+		in deadlock. Instead we try latching the page and retry in case
+		of a failure. */
+		} else if (!block->page.lock.s_lock_try()) {
+			goto wait_for_unzip;
+		}
```
dr-m, line 2591: I would place the comment inside the `else if` body and indent it to the same level with the `goto` statement.

With this change, the `wait_for_unzip` label is no longer specific to the buffer pool management of `ROW_FORMAT=COMPRESSED` pages. Therefore, the label had better be renamed, or a new label be added next to it, such as `wait_for_read`.

dr-m (approved):
Thank you. This looks correct to me. In the commit message, please refer to the exact commit that you think introduced this bug, using a wording such as the following:
```
commit aaef2e1d8c843d1e40b1ce0c5199c3abb3c5da28 (MDEV-27058)
```

I was thinking if there could be a simpler hang scenario:
1. A page is being read into the buffer pool.
1. The page is being requested by another thread `buf_page_get_gen()` with `RW_NO_LATCH`.
1. On read completion, the page turns out to be corrupted, and `buf_pool_t::corrupted_evict()` will be invoked on it.

This patch does not affect that scenario, because in that case, we would notice that the block state was changed to `FREED` and release our S-latch and buffer-fix. However, while thinking about this scenario, I noticed that there is a small race condition with `buf_pool_t::corrupted_evict()`, which would be fixed by releasing the buffer-fix a little later in the error handling code path:
```diff
diff --git a/storage/innobase/buf/buf0buf.cc b/storage/innobase/buf/buf0buf.cc
index dc90160a668..12a11aefc12 100644
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2587,15 +2587,15 @@ buf_page_get_low(
 		block->page.lock.s_unlock();
 
 		if (UNIV_UNLIKELY(state < buf_page_t::UNFIXED)) {
-			block->page.unfix();
 			if (UNIV_UNLIKELY(id == page_id)) {
 				/* The page read was completed, and
 				another thread marked the page as free
 				while we were waiting. */
-				goto ignore_unfixed;
+				goto ignore;
 			}
 
 			ut_ad(id == page_id_t{~0ULL});
+			block->page.unfix();
 
 			if (++retries < BUF_PAGE_READ_MAX_RETRIES) {
 				goto loop;
```
If you agree with this, please include this fix as well.

storage/innobase/buf/buf0buf.cc
mariadb-DebarunBanerjee: Done.

PR #3236, opened 2024-04-30
MDEV-33971 NAME_CONST in WHERE clause replaced by inner item

DaveGosselin-MariaDB:
Improve performance of queries like
  SELECT * FROM t1 WHERE field = NAME_CONST('a', 4);
by, in this example, replacing the WHERE clause with field = 4 in the case of ref access.

The rewrite is done during fix_fields and we disambiguate this case from other cases of NAME_CONST by inspecting where we are in parsing.  We rely on THD::where to accomplish this.  To improve performance there, we change the type of THD::where to be an enumeration, so we can avoid string comparisons during Item_name_const::fix_fields.  Consequently, this patch also changes all usages of THD::where to conform likewise.

### DIFF
--- a/sql/item.cc
+++ b/sql/item.cc
@@ -2184,6 +2184,36 @@ bool Item_name_const::fix_fields(THD *thd, Item **ref)
     my_error(ER_RESERVED_SYNTAX, MYF(0), "NAME_CONST");
     return TRUE;
   }
+
+  /*
+    If we have either of the following:
+      ... WHERE foo=NAME_CONST(...)
+      ... JOIN ... ON foo=NAME_CONST(...)
+    then we have an opportunity to unwrap the NAME_CONST and
+    use the enclosed value directly, replacing NAME_CONST in
+    the parse tree with the value it encloses.
+  */
+  if ((thd->where == THD_WHERE::WHERE_CLAUSE ||
+       thd->where == THD_WHERE::ON_CLAUSE) &&
+      (value_item->type() == FUNC_ITEM ||
+       value_item->type() == CONST_ITEM))
+  {
+    thd->change_item_tree(ref, value_item);
+
+    /*
+      We're replacing NAME_CONST('name', value_item) with value_item.
+      Only a few constants and functions are possible as value_item, see
+      Create_func_name_const::create_2_arg.
+      Set the value_item's coercibility to be the same as NAME_CONST(...)
+      would have (see how it's set a few lines below).
+    */
+    if (value_item->collation.derivation != DERIVATION_NUMERIC)
+      value_item->collation.set(value_item->collation.collation,
+                                DERIVATION_IMPLICIT);
+    return FALSE;
+  }
+  // else, could not unwrap, fall back to default handling below.
+
   if (value_item->collation.derivation == DERIVATION_NUMERIC)
     collation= DTCollation_numeric();
   else
@@ -5528,7 +5558,7 @@ static Item** find_field_in_group_list(Item *find_item, ORDER *group_list)
         is ambiguous.
       */
       my_error(ER_NON_UNIQ_ERROR, MYF(0),
-               find_item->full_name(), current_thd->where);
+               find_item->full_name(), thd_where(current_thd));
       return NULL;
     }
   }
@@ -5613,7 +5643,7 @@ resolve_ref_in_select_and_group(THD *thd, Item_ident *ref, SELECT_LEX *select)
       push_warning_printf(thd, Sql_condition::WARN_LEVEL_WARN,
                           ER_NON_UNIQ_ERROR,
                           ER_THD(thd,ER_NON_UNIQ_ERROR), ref->full_name(),
-                          thd->where);
+                          thd_where(thd));
 
     }
   }
@@ -5947,7 +5977,7 @@ Item_field::fix_outer_field(THD *thd, Field **from_field, Item **reference)
     if (upward_lookup)
     {
       // We can't say exactly what absent table or field
-      my_error(ER_BAD_FIELD_ERROR, MYF(0), full_name(), thd->where);
+      my_error(ER_BAD_FIELD_ERROR, MYF(0), full_name(), thd_where(thd));
     }
     else
     {
@@ -6174,7 +6204,7 @@ bool Item_field::fix_fields(THD *thd, Item **reference)
             {
               /* The column to which we link isn't valid. */
               my_error(ER_BAD_FIELD_ERROR, MYF(0), (*res)->name.str,
-                       thd->where);
+                       thd_where(thd));
               return(1);
             }
 
@@ -6219,7 +6249,7 @@ bool Item_field::fix_fields(THD *thd, Item **reference)
 
       if (unlikely(!select))
       {
-        my_error(ER_BAD_FIELD_ERROR, MYF(0), full_name(), thd->where);
+        my_error(ER_BAD_FIELD_ERROR, MYF(0), full_name(), thd_where(thd));
         goto error;
       }
       if ((ret= fix_outer_field(thd, &from_field, reference)) < 0)
@@ -8163,7 +8193,7 @@ bool Item_ref::fix_fields(THD *thd, Item **reference)
       if (unlikely(!outer_context))
       {
         /* The current reference cannot be resolved in this query. */
-        my_error(ER_BAD_FIELD_ERROR,MYF(0), full_name(), thd->where);
+        my_error(ER_BAD_FIELD_ERROR,MYF(0), full_name(), thd_where(thd));
         goto error;
       }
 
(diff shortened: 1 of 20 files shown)


sql/item.cc
```diff
@@ -2176,6 +2176,11 @@ Item::Type Item_name_const::type() const
 
 bool Item_name_const::fix_fields(THD *thd, Item **ref)
 {
+  if (thd->where == THD_WHERE::WHERE_CLAUSE) {
```
spetrunia, line 2179: Coding style: `{` on separate line

sql/item.cc
```diff
@@ -2176,6 +2176,11 @@ Item::Type Item_name_const::type() const
...
+      value_item->fix_fields_if_needed(thd, &value_item);
+      thd->change_item_tree(ref, value_item);
```
spetrunia, line 2181: This needs error handling. I think one needs t ocall fix_fields_if_needed for both value_item and name_item and check the result.

I'm wondering why this is needed:

```cpp
 thd->change_item_tree(ref, value_item);
```
did you have anything in mind?

spetrunia (changes requested):
Generally, looks like a step in the right direction. 

I'm wondering if the patch could be smaller if there was a  `const char* thd_where_is_where_cond` and then one could compare `thd->where` strings by comparing pointers... but since we've added an enum, let's not remove it.  It has an advantage of listing the possible values of `thd->where` ...

I think in addition to handling WHERE, we need to handle ON expressions. 
Handling other expressions (like HAVING) looks more risky as they may have references to select list values...

The patch needs test cases... I used this to play:
```

create table t1 (a int, b int);
insert into t1 values (1,1),(2,2);

explain format=json
select * from t1 where a=name_const(1,'varname');

explain format=json
select * from t1 left join t1 as t2 on t1.a=name_const(1,'varname') and t1.b=t2.b;

drop table t1;
```

spetrunia:
One other question:

What if Item_const::fix_fields() always did the following, regardless of its location:

* call `value_item->set_name(...)` in the same way it does for itself in  `Item_const::Item_const`. 
* substitute itself with `value_item`

Will this work and if not why?

DaveGosselin-MariaDB:
> Generally, looks like a step in the right direction.

Thanks!  😄

> I'm wondering if the patch could be smaller if there was a `const char* thd_where_is_where_cond` and then one could compare `thd->where` strings by comparing pointers... but since we've added an enum, let's not remove it. It has an advantage of listing the possible values of `thd->where` ...

I had considered string pointers very briefly, but decided against that approach for a number of reasons:  
 - pointer values are hard to distinguish in logs (should these get logged)
 - enumerations are easier to read and (sometimes) remember
 - future maintainers could look at the string pointer comparisons and say "did he really mean to compare the pointer values and not the strings themselves?" which would lead to endless stumbling. 
 - comparing values across different builds would be impossible

Generally, enumerations are more intentional and less ambiguous.

> I think in addition to handling WHERE, we need to handle ON expressions. Handling other expressions (like HAVING) looks more risky as they may have references to select list values...

Would that be appropriate to do now or as a separate MDEV?

sql/item.cc
DaveGosselin-MariaDB: Without the call to `change_item_tree`, the node for `NAME_CONST` remains in the parse tree.  This call replaces the `NAME_CONST` node with the value it wraps.

DaveGosselin-MariaDB:
> One other question:
> 
> What if Item_const::fix_fields() always did the following, regardless of its location:
> 
> * call `value_item->set_name(...)` in the same way it does for itself in  `Item_const::Item_const`.
> * substitute itself with `value_item`
> 
> Will this work (EDIT: quick attempt: no, rpl.rpl_name_const fails) and if not why?

This is an interesting idea, we should explore this further in another MDEV.

spetrunia (changes requested):
Got a question: 

https://jira.mariadb.org/browse/MDEV-33971?focusedCommentId=285058&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-285058

need to sort this out first.

spetrunia:
I don't understand how the last variant is supposed to work? I see this check

```
if (thd->last_sql_command == SQLCOM_SELECT &&
```
what is it supposed to achieve?

spetrunia:
As far as I understand, the issue raised in 
https://jira.mariadb.org/browse/MDEV-33971?focusedCommentId=285058&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-285058

is still not addressed.
Double-checking this.  I take 
```
commit 05e9b4a6955bcbedc237e46fafdccb2b4419d1a6 (HEAD -> 10.6-mdev-33971, origin/10.6-mdev-33971)
Author: Dave Gosselin <dave.gosselin@mariadb.com>
Date:   Fri Apr 26 12:13:31 2024 -0400

    MDEV-33971 NAME_CONST in WHERE clause replaced by inner item
```
I'm running this testcase:

```
create table t1 (
  a varchar(100) collate utf8_unicode_ci,
  b int
);
insert into t1 values ('foo', 1),('bar', 1);
select * from t1;
 
select * from t1  where a= NAME_CONST('var1',_latin1'foo' COLLATE 'latin1_swedish_ci');

drop table t1;
```

I get:
```
mysqltest: At line 8: query 'select * from t1  where a= NAME_CONST('var1',_latin1'foo' COLLATE 'latin1_swedish_ci')' failed: ER_CANT_AGGREGATE_2COLLATIONS (1267): Illegal mix of collations (utf8mb3_unicode_ci,IMPLICIT) and (latin1_swedish_ci,EXPLICIT) for operation '='
```

Trying the same test on the previous commit:
```
commit 2d5cba22a98e54c0c8975d8d7fe7e04981d73c8a (HEAD)
Author: mariadb-DebarunBanerjee <debarun.banerjee@mariadb.com>
Date:   Thu May 16 13:10:44 2024 +0530
```
I get no error.

sql/item.cc
```diff
@@ -2176,6 +2176,26 @@ Item::Type Item_name_const::type() const
...
+   */
+  if ((thd->where == THD_WHERE::WHERE_CLAUSE ||
+       thd->where == THD_WHERE::ON_CLAUSE) &&
+    !value_item->fix_fields_if_needed(thd, &value_item))
+  {
+    thd->change_item_tree(ref, value_item);
+    if (value_item->collation.derivation != DERIVATION_NUMERIC)
+      value_item->collation.set(value_item->collation.collation,
+                                DERIVATION_IMPLICIT);
+    return FALSE;
```
spetrunia, line 2195: The first and obvious question: if we reach this return statement, then
```
    base_flags|= item_base_t::FIXED;
```
is not done for the Item_const object, and item remains unfixed.

sql/item.cc
```diff
@@ -2176,6 +2176,28 @@ Item::Type Item_name_const::type() const
...
+      ... JOIN ... ON foo=NAME_CONST(...)
+    then we have an opportunity to unwrap the NAME_CONST and
+    use the enclosed value directly, replacing NAME_CONST in
+    the parse tree with the value it encloses.
+   */
+  if ((thd->where == THD_WHERE::WHERE_CLAUSE ||
+       thd->where == THD_WHERE::ON_CLAUSE) &&
+      (value_item->type() == FUNC_ITEM ||
+       value_item->type() == CONST_ITEM) &&
+    !value_item->fix_fields_if_needed(thd, &value_item))
```
spetrunia, line 2191: We have a call to `value_item->fix_fields_if_needed`  here, and then again we have it below...  I think having all fix_fields calls first and rewrite second would be cleaner...

spetrunia (changes requested):
... Done the above and added a couple of comments.
What do you think of applying this patch: https://gist.github.com/spetrunia/5236479b5c4d5cd286ff00f914c759a9

After that, the patch looks ok to push.

sql/item.cc
```diff
@@ -2176,6 +2176,28 @@ Item::Type Item_name_const::type() const
...
+    If we have either of the following:
+      ... WHERE foo=NAME_CONST(...)
+      ... JOIN ... ON foo=NAME_CONST(...)
+    then we have an opportunity to unwrap the NAME_CONST and
+    use the enclosed value directly, replacing NAME_CONST in
+    the parse tree with the value it encloses.
+   */
+  if ((thd->where == THD_WHERE::WHERE_CLAUSE ||
+       thd->where == THD_WHERE::ON_CLAUSE) &&
+      (value_item->type() == FUNC_ITEM ||
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: I don't understand your comment.  This check is necessary to unwrap the NAME_CONST item and replace it in the tree with the item that it wraps (please see the tests).

DaveGosselin-MariaDB:
> ... Done the above and added a couple of comments. What do you think of applying this patch: https://gist.github.com/spetrunia/5236479b5c4d5cd286ff00f914c759a9
> 
> After that, the patch looks ok to push.

OK I will look, thank you.

sql/item.cc
DaveGosselin-MariaDB: Yes I agree, this is nicer, thank you.

DaveGosselin-MariaDB:
> ... Done the above and added a couple of comments. What do you think of applying this patch: https://gist.github.com/spetrunia/5236479b5c4d5cd286ff00f914c759a9
> 
> After that, the patch looks ok to push.

Thank you, this is a nice change and I included it.  Once the tests are done then I will submit, thank you for the OK.

PR #3238, opened 2024-05-02
MDEV-33078 SysInfo.pm reports incorrect CPU count on macOS

DaveGosselin-MariaDB:
When running on macOS, MTR will ask the operating system for the core count when --parallel=auto


grooverdan (approved):
Looks good.

I see no reason this couldn't be applied to 10.5, especially as you are probably going to be doing MacOS tests there for a while.

1. rebase this pr
```
git rebase --onto origin/10.5 HEAD^
git push -f
```
(push should be to the same branch name, despite the version disparity).
2. edit PR title and change the destination branch to 10.5
3. enable automerge on the PR
4. it should automerge

PR #3240, opened 2024-05-03
MDEV-33750: Rename variables which are mysql_ to mariadb_ in Debian directory

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-33750*

## Description
This more about branding PR. As in many places when reading Debian scripts MariaDB is  mentioned as MySQL (`mysql_`). As MariaDB is something else than MySQL (besides being mostly compatible upgrade path) it should be proudly be just MariaDB named variables in `debian/` directory and get rid of other naming. There is also couple other changes under `debian/` to brand this more.

## Release Notes
This does not affect to release notes. It's more mental thing.

## How can this PR be tested?
Build branch with `debian/deb-autobake.sh` and then install packages nothing should be changed or check Salsa-CI run
https://salsa.debian.org/illuusio/mariadb-server/-/pipelines/672483
which fails badly (which are be taken care of in other PR) but one can see that it works as expected.

## PR quality check


ottok:
This makes sense, just wondering why target stable 10.11 branch? Perhaps this should go in 11.4 or 11.5 instead?

LinuxJedi:
I agree with @ottok, I don't seen any obvious problems in a quick scan of this PR, but I wonder why 10.11 was chosen instead of either 10.5 or 11.6 (with 11.5 as a sub for now).

illuusio:
Ok I'll rebase this to 11.5. I just thought that rename won't be just big deal but it's no problem to rebase.

illuusio:
Ok now it's rebased to 11.5

PR #3241, opened 2024-05-04
WIP: MDEV-34041 Display additional information for materialized subqueries…

Olernov:
… in EXPLAIN/ANALYZE FORMAT=JSON

- [x] *The Jira issue number for this PR is: MDEV-34041*

## Description
This is a WIP (work-in-progress) PR, that's why the test suite is not updated. The ANALYZE output format is open for discussion, so it makes sense to update existing tests after the output format is approved

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/main/subselect_mat_analyze_json.result
+++ b/mysql-test/main/subselect_mat_analyze_json.result
@@ -1,0 +1,945 @@
+set @save_optimizer_switch=@@optimizer_switch;
+create table t1 (a int);
+create table t2 (b int);
+insert into t1 values (null), (1), (2), (3);
+insert into t2 values (3), (4);
+set @@optimizer_switch = "materialization=on,in_to_exists=off,semijoin=off";
+explain format=json select * from t1 where a in (select b from t2);
+EXPLAIN
+{
+  "query_block": {
+    "select_id": 1,
+    "table": {
+      "table_name": "t1",
+      "access_type": "ALL",
+      "rows": 4,
+      "filtered": 100,
+      "attached_condition": "<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "materialization": {
+          "query_block": {
+            "select_id": 2,
+            "table": {
+              "table_name": "t2",
+              "access_type": "ALL",
+              "rows": 2,
+              "filtered": 100
+            }
+          }
+        }
+      }
+    ]
+  }
+}
+# "Complete match" execution strategy
+analyze format=json select * from t1 where a in (select b from t2);
+ANALYZE
+{
+  "query_block": {
+    "select_id": 1,
+    "r_loops": 1,
+    "r_total_time_ms": "REPLACED",
+    "table": {
+      "table_name": "t1",
+      "access_type": "ALL",
+      "r_loops": 1,
+      "rows": 4,
+      "r_rows": 4,
+      "r_table_time_ms": "REPLACED",
+      "r_other_time_ms": "REPLACED",
+      "filtered": 100,
+      "r_filtered": 25,
+      "attached_condition": "<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "materialization": {
+          "r_strategy": "index_lookup",
+          "r_loops": 3,
+          "query_block": {
+            "select_id": 2,
+            "r_loops": 1,
+            "r_total_time_ms": "REPLACED",
+            "table": {
+              "table_name": "t2",
+              "access_type": "ALL",
+              "r_loops": 1,
+              "rows": 2,
+              "r_rows": 2,
+              "r_table_time_ms": "REPLACED",
+              "r_other_time_ms": "REPLACED",
+              "filtered": 100,
+              "r_filtered": 100
+            }
+          }
+        }
+      }
+    ]
+  }
+}
+# "Partial match" is used due to NOT IN
+# Force rowid-merge partial partial matching
+set @@optimizer_switch="partial_match_rowid_merge=on,partial_match_table_scan=off";
+analyze format=json select * from t1 where a not in (select b from t2);
+ANALYZE
+{
+  "query_block": {
+    "select_id": 1,
+    "r_loops": 1,
+    "r_total_time_ms": "REPLACED",
+    "table": {
+      "table_name": "t1",
+      "access_type": "ALL",
+      "r_loops": 1,
+      "rows": 4,
+      "r_rows": 4,
+      "r_table_time_ms": "REPLACED",
+      "r_other_time_ms": "REPLACED",
+      "filtered": 100,
+      "r_filtered": 50,
+      "attached_condition": "!<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "materialization": {
+          "r_strategy": "index_lookup;array merge for partial match",
+          "r_loops": 4,
+          "r_index_lookups": 3,
+          "r_partial_matches": 1,
+          "r_partial_match_buffer_size": "REPLACED",
+          "r_partial_match_array_sizes": ["2"],
+          "query_block": {
+            "select_id": 2,
+            "r_loops": 1,
+            "r_total_time_ms": "REPLACED",
+            "table": {
+              "table_name": "t2",
+              "access_type": "ALL",
+              "r_loops": 1,
+              "rows": 2,
+              "r_rows": 2,
+              "r_table_time_ms": "REPLACED",
+              "r_other_time_ms": "REPLACED",
+              "filtered": 100,
+              "r_filtered": 100
+            }
+          }
+        }
+      }
+    ]
+  }
+}
+# Force table scan partial matching
+set @@optimizer_switch="partial_match_rowid_merge=off,partial_match_table_scan=on";
+analyze format=json select * from t1 where a not in (select b from t2);
+ANALYZE
+{
+  "query_block": {
+    "select_id": 1,
+    "r_loops": 1,
+    "r_total_time_ms": "REPLACED",
+    "table": {
+      "table_name": "t1",
+      "access_type": "ALL",
+      "r_loops": 1,
+      "rows": 4,
+      "r_rows": 4,
+      "r_table_time_ms": "REPLACED",
+      "r_other_time_ms": "REPLACED",
+      "filtered": 100,
+      "r_filtered": 50,
+      "attached_condition": "!<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "materialization": {
+          "r_strategy": "index_lookup;full scan for partial match",
+          "r_loops": 4,
+          "r_index_lookups": 3,
+          "r_partial_matches": 1,
+          "query_block": {
+            "select_id": 2,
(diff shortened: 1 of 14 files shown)


spetrunia (changes requested):
Currently, there are two kinds of data structures:
* EXPLAIN data structure
* Tracker.

Explain_subq_materialization tries to be both. I know this is caused by
the fact that there's really no query plan.

I would still request to have the Explain class and the tracker class 
explicitly. Even if the Explain class has only one member, the tracker.

The tracker class should follow other tracker classes:
- data members are private.
- execution code calls functions to report data
- printing json output is done by local members.

- constructor initalizes everything to some initial state which 
can already be printed (currently it doesn't). This is to allow
SHOW ANALYZE and also the situation where subquery is never invoked.

sql/item_subselect.cc
```diff
@@ -6966,3 +6992,47 @@ void Item_subselect::init_expr_cache_tracker(THD *thd)
...
+  {
+    writer->add_member("r_exec_strategy").add_str(
+          exec_strategy_str[exec_strategy]);
+    if (partial_match_buffer_size != -1)
+    {
+      writer->add_member("r_partial_match_buffer_size").add_size(
+            partial_match_buffer_size);
+    }
+
+    for(uint i= 0; i < partial_match_merge_keys_counts.elements(); i++)
```
spetrunia, line 7031: Any reason for not printing a JSON array here?

mysql-test/main/subselect_mat_analyze_json.result
```diff
@@ -0,0 +1,910 @@
...
+    "table": {
+      "table_name": "t1",
+      "access_type": "ALL",
+      "rows": 4,
+      "filtered": 100,
+      "attached_condition": "<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "subquery_materialization": {
```
spetrunia, line 21: Can this be just `materialization` ?

spetrunia (commented):
I think it also makes sense to introduce `r_loops` in `subquery_materialization` (to be `materialization`)

sql/item_subselect.cc
Olernov: Agree, fixed in the new revision

mysql-test/main/subselect_mat_analyze_json.result
Olernov: Sure, fixed

sql/item_subselect.cc
```diff
@@ -6966,3 +6992,47 @@ void Item_subselect::init_expr_cache_tracker(THD *thd)
...
+
+
+void Explain_subq_materialization::print_explain_json(Json_writer *writer,
+                                                      bool is_analyze)
+{
+  writer->add_member("subquery_materialization").start_object();
+  if (is_analyze)
+  {
+    writer->add_member("r_exec_strategy").add_str(
+          exec_strategy_str[exec_strategy]);
```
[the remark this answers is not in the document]
Olernov: Agree with `r_strategy`. The enum of strategies indeed looks excessive, `partial_match` and `impossible` don't seem to be used anywhere. Do you think we can shrink `subselect_hash_sj_engine::exec_strategy`?

Olernov:
> I think it also makes sense to introduce `r_loops` in `subquery_materialization` (to be `materialization`)

I don't understand what should be counted/tracked under this counter. Every lookup in the materialized subquery?

sql/item_subselect.cc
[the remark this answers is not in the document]
spetrunia: I would not touch that in this patch.  I would use a `switch(...)` instead  of `exec_strategy_str`.

spetrunia:
> > I think it also makes sense to introduce `r_loops` in `subquery_materialization` (to be `materialization`)
> 
> I don't understand what should be counted/tracked under this counter. Every lookup in the materialized subquery?

Yes.  Sometimes, it is possible to infer this number, but in general case it is not available anywhere else. So it makes a lot of sense to print it.

spetrunia:
Also, note that for non-NULL-aware materialization, it reports `r_loops==r_index_lookup_loops`.

        "materialization": {
          "r_strategy": "index_lookup",
          "r_loops": 10,
          "r_index_lookup_loops": 10,
          "query_block": {

sql/item_subselect.cc
```diff
@@ -4266,17 +4268,21 @@ int subselect_uniquesubquery_engine::exec()
...
+  if (tracker)
+    tracker->increment_loops_count();
   if (in_subs->left_expr_has_null())
   {
     /*
       The case when all values in left_expr are NULL is handled by
       Item_in_optimizer::val_int().
     */
+    if (tracker)
+      tracker->increment_table_scan_loops();
```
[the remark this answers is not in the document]
spetrunia: ... and is this relevant at all ? 
Why does `subselect_uniquesubquery_engine::exec()`  access the "materialization_tracker"?

mysql-test/main/subselect_mat_analyze_json.result
```diff
@@ -0,0 +1,948 @@
...
+      "filtered": 100,
+      "r_filtered": 25,
+      "attached_condition": "<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "materialization": {
+          "r_strategy": "index_lookup",
+          "r_loops": 3,
+          "r_index_lookup_loops": 3,
```
spetrunia, line 61: We should not add "_loops" to all counters. 
Please change to "r_index_lookups".

mysql-test/main/subselect_mat_analyze_json.result
```diff
@@ -0,0 +1,948 @@
...
+      "r_table_time_ms": "REPLACED",
+      "r_other_time_ms": "REPLACED",
+      "filtered": 100,
+      "r_filtered": 50,
+      "attached_condition": "!<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "materialization": {
+          "r_strategy": "partial_match_merge",
```
spetrunia, line 108: This looks confusing: it is not clear that partial matches are only done when we can't do index lookup.
How about changing to 

```index_lookup;array merge for partial match```

``` "r_partial_match_loops": 1,```
Please change to `r_partial_matches`.

``` "r_partial_merge_key_sizes": ["2"],```
please change to `r_partial_match_array_sizes` .

File: mysql-test/main/subselect_mat_analyze_json.result [hunk 7]
```diff
@@ -0,0 +1,948 @@
...
+      "r_table_time_ms": "REPLACED",
+      "r_other_time_ms": "REPLACED",
+      "filtered": 100,
+      "r_filtered": 50,
+      "attached_condition": "!<in_optimizer>(t1.a,t1.a in (subquery#2))"
+    },
+    "subqueries": [
+      {
+        "materialization": {
+          "r_strategy": "partial_match_scan",
```
spetrunia, line 159: Following the above suggestion, let's change `partial_match_scan` into 

```index_lookup;full scan for partial match ```

**Review by spetrunia — CHANGES_REQUESTED** [review]
Also, I do not see any testcase with `r_table_scan_loops`.  Please add one. (or remove r_table_scan_loops if it's dead code, that's what I suspect).

**Comment by spetrunia** [comment]
@Olernov , ok I think there are enough comments to get the next version of the patch.

File: sql/sql_explain.cc [hunk 8]
```diff
@@ -2693,3 +2710,45 @@ void Explain_range_checked_fer::print_json(Json_writer *writer,
     writer->end_object();
   }
 }
+
+
+void Explain_subq_materialization::print_explain_json(Json_writer *writer,
```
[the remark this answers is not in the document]
Olernov: That doesn't align with other `Explain_*` classes architecture. As I can see, existing trackers only track and `Explain_*` classes print. And it seems logical to me, especially taking into account the class names. If `Subq_materialization_tracker` will print data, `Explain_subq_materialization` will become some kind of degenerate proxy class

sql/item_subselect.cc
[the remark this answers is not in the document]
Olernov: > Note that table scan is NOT done if `in_subs->is_top_level_item()` evaluates to TRUE right below.
Agree, better place this increment after the condition.

> ... and is this relevant at all ? Why does `subselect_uniquesubquery_engine::exec()` access the "materialization_tracker"?
Materialization leverages this class as a default engine to implement the `index_lookup` (complete match) strategy, as well as `subselect_partial_match_engine` does for some cases: `subselect_uniquesubquery_engine *lookup_engine`.

Olernov:
> Also, note that for non-NULL-aware materialization, it reports `r_loops==r_index_lookup_loops`.
> 
> ```
>     "materialization": {
>       "r_strategy": "index_lookup",
>       "r_loops": 10,
>       "r_index_lookup_loops": 10,
>       "query_block": {
> ```

Do you mean for the `index_lookup` strategy? Yes, and it seems reasonable. But numbers differ for `partial match` strategies. Or you suggest removing redundant output for `index_lookup` strategy?

spetrunia:
> Do you mean for the `index_lookup` strategy? Yes, and it seems reasonable. But numbers differ for `partial match` strategies. Or you suggest removing redundant output for `index_lookup` strategy?

Let's keep this for now and look at this in the next step.

sql/sql_explain.h
```diff
@@ -115,6 +116,12 @@ class Explain_node : public Sql_alloc
   */
   Expression_cache_tracker* cache_tracker;
 
+  /**
+    This tracker is not NULL if the node explains a materialized subquery.
```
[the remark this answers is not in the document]
Olernov: OK, fixed in a new revision

sql/item_subselect.cc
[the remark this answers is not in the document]
Olernov: After some more examination I decided to remove the counter of table scan loops, it is indeed not applicable in our case

Olernov:
> Also, I do not see any testcase with `r_table_scan_loops`. Please add one. (or remove r_table_scan_loops if it's dead code, that's what I suspect).

I agree, removed the counter

mysql-test/main/subselect_mat_analyze_json.result
Olernov: Fixed

mysql-test/main/subselect_mat_analyze_json.result
Olernov: Done

mysql-test/main/subselect_mat_analyze_json.result
Olernov: Done

Olernov:
> > Do you mean for the `index_lookup` strategy? Yes, and it seems reasonable. But numbers differ for `partial match` strategies. Or you suggest removing redundant output for `index_lookup` strategy?
> 
> Let's keep this for now and look at this in the next step.

I decided to remove `r_index_lookup_loops` from `index_lookup` strategy because it is, indeed, redundant and equals to `r_loops`.

spetrunia:
The result looks good. Please apply this code cleanup patch:
https://gist.github.com/spetrunia/a09ad18f970d4ebe146e8cea5ed28bbd
and update the test results.

Olernov:
@spetrunia, OK. I applied the patch and updated the tests.

PR #3243, opened 2024-05-07
MDEV-34112 Replace one operator name keyword

HugoWenTD:
- [x] *The Jira issue number for this PR is: MDEV-34112*

## Description
Alternative operator name keywords like `and`, `or`, `xor`, etc., are uncommon in MariaDB and can cause obscure build errors when the GCC flag `-fno-operator-names` is applied.

Description of `-fno-operator-names`:
https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html
> Do not treat the operator name keywords `and`, `bitand`, `bitor`, `compl`, `not`, `or` and `xor` as synonyms as keywords.

Part of the build errors:

    /local/p4clients/pkgbuild-LdLa_/workspace/src/RDSMariaDB/sql/sql_select.cc:11171:28: error: expected ‘)’ before ‘and’
    11171 |     DBUG_ASSERT(sel >= 0.0 and sel <= 1.00001);
          |                            ^~~
    /local/p4clients/pkgbuild-LdLa_/workspace/src/RDSMariaDB/include/my_global.h:372:44: note: in definition of macro ‘unlikely’
      372 | #define unlikely(x)     __builtin_expect(((x) != 0),0)
          |                                            ^
    ...

The build failure is caused by using alternative operator name keywords `and` introduced in commit b66cdbd1e.
Replace the `and` keyword with `&&` and target on MariaDB 11.0+ branches which include the commit.

## Release Notes
N/A

## How can this PR be tested?
This is the only occurrence I found using the operator names. With this change, build succeeds with GCC flag `-fno-operator-names`.
Existing MTR tests and builds will verify there's no regression.

## PR quality check

## Copyright

### DIFF
--- a/sql/sql_select.cc
+++ b/sql/sql_select.cc
@@ -11125,7 +11125,7 @@ double table_after_join_selectivity(JOIN *join, uint idx, JOIN_TAB *s,
   else
   {
     sel= records_out / pos->records_read;
-    DBUG_ASSERT(sel >= 0.0 and sel <= 1.00001);
+    DBUG_ASSERT(sel >= 0.0 && sel <= 1.00001);
     if (sel > 1.0)
       sel= 1.0;
   }


LinuxJedi (approved):
Good catch, many thanks!

vuvova:
I'll also push 11a34e746ae into 10.5

PR #3245, opened 2024-05-08
Update test upgrade script for use with latest Fedora

robinnewhouse:
## Description
Fedora 40 introdced wget2 as the default wget program, which caused a
break in the functionality of the test_upgrade.sh script. Removing the
`--spider` option and modifying the determination of the latest
distribution to test on fixes the issue. At the same time, added in
another source for the boost-program-options RPM when the fedora version
is 39.

Additionally added rpm sources for the boost-program-options and openssl 1.1. This is necessary when building older versions of MariaDB (e.g., 10.4) on newer linux distributions (e.g., Fedora 39) that do not always have access to all required dependencies. In the above example older MariaDB versions are not compatible with OpenSSL 3.0+, so require an older version.

## Release Notes
None

## How can this PR be tested?

Start a docker container using the fedora:40 image
```
docker run --rm  -v $(pwd):/mariadb -it fedora:40 bash
```

Execute the upgrade script
```
/mariadb/tests/upgrade_from/test_upgrade.sh 10.11 11.4
```

Try the same for fedora:39

## PR quality check

### DIFF
--- a/tests/upgrade_from/test_upgrade.sh
+++ b/tests/upgrade_from/test_upgrade.sh
@@ -117,11 +117,12 @@ install_mariadb_from_archive() {
     # The missing libraries depned on the latest distribution version, so that is found by scanning the MariaDB archive directory.
 
     log Finding MariaDB RPM repository for version $version
-    if ! $(wget --spider -r --no-parent --level 1 --quiet https://archive.mariadb.org/mariadb-$version/yum/fedora/ -P /tmp/); then
-        log Could not find RPMs
+    latest_distro=$(curl -s https://archive.mariadb.org/mariadb-$version/yum/fedora/ | (grep -Eo '^<a href="[0-9]+/">' || true) | 
+    sort -V | tail -n 1 | sed -n 's/^<a href="\([0-9]*\)\/">/\1/p')
+    if [ -z $latest_distro ]; then
+        log Could not find repository. Exiting.
         exit 1
     fi
-    latest_distro=$(ls -1v /tmp/archive.mariadb.org/mariadb-$version/yum/fedora/ | tail -n 1)
     rpm_repository=https://archive.mariadb.org/mariadb-$version/yum/fedora/$latest_distro/$(uname -m)/
     log RPM repository: $rpm_repository
     # log Fedora distribution: $latest_distro # Currently only supports tests on Fedora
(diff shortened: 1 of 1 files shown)


tests/upgrade_from/test_upgrade.sh
```diff
@@ -154,6 +154,8 @@ EOF
     [[ $latest_distro == 34 ]] && dnf install -y https://repo.almalinux.org/almalinux/9/AppStream/x86_64/os/Packages/boost-program-options-1.75.0-8.el9.x86_64.rpm \
         https://vault.centos.org/centos/8/AppStream/x86_64/os/Packages/liburing-1.0.7-3.el8.x86_64.rpm
     [[ $latest_distro == 35 ]] && dnf install -y https://archives.fedoraproject.org/pub/archive/fedora/linux/updates/36/Everything/x86_64/Packages/b/boost-program-options-1.76.0-12.fc36.x86_64.rpm
+    [[ $latest_distro -le 35 ]] && dnf install -y https://dl.fedoraproject.org/pub/fedora/linux/releases/39/Everything/x86_64/os/Packages/o/openssl1.1-1.1.1q-5.fc39.x86_64.rpm
```
LinuxJedi, line 157: Does this work for things like Fedora 32 as well then?

tests/upgrade_from/test_upgrade.sh
```diff
@@ -154,6 +154,8 @@ EOF
...
+    [[ $latest_distro -ge 39 ]] && dnf install -y https://rpmfind.net/linux/fedora/linux/releases/39/Everything/x86_64/os/Packages/b/boost-program-options-1.81.0-8.fc39.x86_64.rpm
```
LinuxJedi, line 158: This could definitely cause problems with later versions. Why not just do:

```
dnf install -y boost-program-options
```

It will be in the default repo. Unlike other lines in this file which are pulling from another repo on a single version match.

tests/upgrade_from/test_upgrade.sh
```diff
@@ -117,11 +117,11 @@ install_mariadb_from_archive() {
     # The missing libraries depned on the latest distribution version, so that is found by scanning the MariaDB archive directory.
 
     log Finding MariaDB RPM repository for version $version
-    if ! $(wget --spider -r --no-parent --level 1 --quiet https://archive.mariadb.org/mariadb-$version/yum/fedora/ -P /tmp/); then
+    if ! $(wget -r --no-parent --level 1 --quiet https://archive.mariadb.org/mariadb-$version/yum/fedora/ -P /tmp/); then
         log Could not find RPMs
         exit 1
     fi
-    latest_distro=$(ls -1v /tmp/archive.mariadb.org/mariadb-$version/yum/fedora/ | tail -n 1)
+    latest_distro=$(ls -1v /tmp/archive.mariadb.org/mariadb-$version/yum/fedora/ | grep -v 'index.html' | tail -n 1)
```
LinuxJedi, line 124: I feel like this whole section would be better as a single curl | grep | sort | tail or something similar. But this probably works too.

LinuxJedi (changes requested):
I think it looks mostly ok, apart from my last comment.

anson1014 (approved):

LinuxJedi (approved):

PR #3246, opened 2024-05-08
MDEV-33952: Fix flaky galera_create_table_as_select test with debug sync

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-33952*

## Description

The test that triggers multi-master conflict between two CTAS commands uses LOCK/UNLOCK TABLES to block local CTAS from progress. It could result in a race when UNLOCK TABLES command is issued a bit earlier then needed, causing local CTAS to run further and change wsrep transaction state, so that a different code path is taken later and the original error gets overridden, causing the test to fail. The solution is to replace LOCK/UNLOCK TABLES with debug sync points.

## Release Notes

The fix is for 10.5+ versions.

## How can this PR be tested?

MTR test is fixed an re-enabled.

## PR quality check


sysprg:
Thanks, the fix has been merged with the head revision:
https://github.com/MariaDB/server/commit/a6b7203d6564728089df66231739e0cc77b039d7

PR #3250, opened 2024-05-11
MDEV-34057  Inconsistent FTS state in concurrent scenarios

Thirunarayanan:
<!--
- [x] *The Jira issue number for this PR is: MDEV-34057*

## Description
Problem:
=======
- This commit is a merge of mysql commit@129ee47ef994652081a11ee9040c0488e5275b14. InnoDB FTS can be in inconsistent state when sync operation terminates the server before committing the operation. This could lead to incorrect synced doc id and incorrect query results.

Solution:
========
- During sync commit operation, InnoDB should pass the sync transaction to update the max doc id
in the config table.

## How can this PR be tested?
./mtr innodb_fts.fts_sync_commit_resiliency

## PR quality check


dr-m (commented):
This needs to be rebased on 10.5, because 10.4 is EOL. Is the patch for 10.6 identical?

storage/innobase/fts/fts0fts.cc
```diff
@@ -2639,11 +2640,13 @@ fts_cmp_set_sync_doc_id(
 	fts_table.type = FTS_COMMON_TABLE;
 	fts_table.table = table;
 
-	trx = trx_create();
-	if (srv_read_only_mode) {
-		trx_start_internal_read_only(trx);
-	} else {
-		trx_start_internal(trx);
+	if (trx == nullptr) {
+		trx = trx_create();
```
dr-m, line 2644: I think that the `trx_create()` call must be moved immediately before the `retry:` label. If we reach the line `goto retry`, we would leak a transaction object. We should fix also that bug while we are at this.

storage/innobase/fts/fts0fts.cc
```diff
@@ -2715,21 +2718,29 @@ fts_cmp_set_sync_doc_id(
 func_exit:
 
 	if (UNIV_LIKELY(error == DB_SUCCESS)) {
+		if (!created_trx) {
+			return DB_SUCCESS;
+		}
 		fts_sql_commit(trx);
```
dr-m, line 2724: It seems to me that we had better simply add the lines
```c++
	if (!created_trx) {
		return error;
	}
```
right after `func_exit:`. We do not seem to want any `goto retry` in that case.

storage/innobase/fts/fts0fts.cc
```diff
@@ -2607,30 +2607,31 @@ fts_get_next_doc_id(
...
+	trx_t	           *trx=nullptr)
 {
-	trx_t*		trx;
 	pars_info_t*	info;
 	dberr_t		error;
 	fts_table_t	fts_table;
 	que_t*		graph = NULL;
 	fts_cache_t*	cache = table->fts->cache;
 	char		table_name[MAX_FULL_NAME_LEN];
+	bool		created_trx = (trx == nullptr);
```
dr-m, line 2634: We can save one operation:
```c++
	const trx_t* const created_trx = trx;
```
and adjust each check for `!created_trx` to `created_trx`.

storage/innobase/fts/fts0fts.cc
```diff
@@ -2607,30 +2607,31 @@ fts_get_next_doc_id(
...
-	doc_id_t		cmp_doc_id,	/*!< in: Doc ID to compare */
-	ibool			read_only,	/*!< in: TRUE if read the
-						synced_doc_id only */
-	doc_id_t*		doc_id)		/*!< out: larger document id
-						after comparing "cmp_doc_id"
-						to the one stored in CONFIG
-						table */
+	const dict_table_t *table,
+	doc_id_t           cmp_doc_id,
+	ibool	           read_only,
```
dr-m, line 2624: While you are reformatting this, you could also replace `ibool` with `bool`, and format the parameters in the common style, without using TAB.

Do we actually need the parameter `read_only`? Could the initial part of this function (up to the invocation of `fts_eval_sql()`) be refactored to a separate function, which `fts_init_index()` would be invoking directly?

storage/innobase/fts/fts0fts.cc
```diff
@@ -6376,7 +6382,20 @@ fts_init_index(
...
+			trx_start_internal(trx);
+		}
+		if (fts_read_synced_doc_id(table, &start_doc, trx)) {
+			fts_sql_rollback(trx);
+			trx->free();
+			goto func_exit;
+		}
+		fts_sql_commit(trx);
+		if (start_doc) start_doc-= 1;
+		trx->free();
```
dr-m, line 6398: Can we simplify this as follows?
```
trx_t* trx = trx_create();
trx_start_internal_read_only(trx);
dberr_t err = fts_read_synced_doc_id(table, &start_doc, trx);
fts_sql_commit(trx);
trx->free();
if (err != DB_SUCCESS) {
	goto func_exit;
}
if (start_doc) {
	start_doc--;
}
```

storage/innobase/fts/fts0fts.cc
```diff
@@ -2607,89 +2607,89 @@ fts_get_next_doc_id(
...
-	// FIXME: We need to retry deadlock errors
-	if (error != DB_SUCCESS) {
-		goto func_exit;
+	if (trx == nullptr) {
+		trx = trx_create();
+		if (srv_read_only_mode) {
+			trx_start_internal_read_only(trx);
+		} else {
+			trx_start_internal(trx);
+		}
```
dr-m, line 2687: If we are creating a transaction here, it would always appear to be a read-only operation, hence we should not call `trx_start_internal(trx)`.

dr-m (approved):
Thank you, this is OK after addressing my comments.

PR #3253, opened 2024-05-13
MDEV-34129 mariadb-install-db appears to hang on macOS

DaveGosselin-MariaDB:
MDEV-34129 mariadb-install-db appears to hang on macOS

    Immediately close down the signal handler loop when we decide to
    break connections as it's the start of process termination
    anyway, and there's no need to wait once we've invoked break_connections.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -2953,6 +2953,15 @@ static void start_signal_handler(void)
   DBUG_VOID_RETURN;
 }
 
+/** Called only from signal_hand function. */
+static void* exit_signal_handler()
+{
+    my_thread_end();
+    signal_thread_in_use= 0;
+    pthread_exit(0);  // Safety
+    return nullptr;  // Avoid compiler warnings
+}
+
 
 /** This threads handles all signals and alarms. */
 /* ARGSUSED */
@@ -3013,10 +3022,7 @@ pthread_handler_t signal_hand(void *arg __attribute__((unused)))
     if (abort_loop)
     {
       DBUG_PRINT("quit",("signal_handler: calling my_thread_end()"));
-      my_thread_end();
-      signal_thread_in_use= 0;
-      pthread_exit(0);				// Safety
-      return 0;                                 // Avoid compiler warnings
+      return exit_signal_handler();
     }
     switch (sig) {
     case SIGTERM:
@@ -3035,6 +3041,7 @@ pthread_handler_t signal_hand(void *arg __attribute__((unused)))
         PSI_CALL_delete_current_thread();
         my_sigset(sig, SIG_IGN);
         break_connect_loop(); // MIT THREAD has a alarm thread
+        return exit_signal_handler();
       }
       break;
     case SIGHUP:


sql/mysqld.cc
```diff
@@ -2081,39 +2081,29 @@ static void clean_up(bool print_message)
 
 #ifndef EMBEDDED_LIBRARY
 
-/**
-  This is mainly needed when running with purify, but it's still nice to
-  know that all child threads have died when mysqld exits.
-*/
 static void wait_for_signal_thread_to_end()
 {
-  uint i, n_waits= DBUG_EVALUATE("force_sighup_processing_timeout", 5, 100);
```
vuvova, line 2090: you removed "force_sighup_processing_timeout", but it's used in the test

vuvova (commented):
I don't understand the fix. What do you mean by "we could fail to initialize the signal handler but mark it as active anyway"? How could it have happened?

sql/mysqld.cc
DaveGosselin-MariaDB: I removed that part of the test, too, because as far as I can see it wasn't necessary; it was only needed because we had to wait, but we had to wait because of the bug that was in the `setup_hand` function.

DaveGosselin-MariaDB:
> I don't understand the fix. What do you mean by "we could fail to initialize the signal handler but mark it as active anyway"? How could it have happened?

The signal handler will fail to initialize if `my_thread_global_init_done` is false, which is possible and expected in some of our tests (I don't recall which at the moment, but I think in certain plugins).

sql/mysqld.cc
DaveGosselin-MariaDB: I should add that by using `pthread_join`, we'll wait for all `SIGHUP` processing to finish, so there's no need for a special timeout.  Joining pthreads is considered a best practice for the pthread library (unless the thread is created in a detached state, which isn't the case here).

DaveGosselin-MariaDB: `rpl.rpl_shutdown_sighup` continues to succeed with these changes, because the `SIGHUP` processing concludes as expected, without the unnecessary extra waiting semantics.  Those were brittle because `sleep`s (and equivalent) are not reliable for multithreaded coordination.

sql/mysqld.cc
```diff
@@ -2960,9 +2950,17 @@ pthread_handler_t signal_hand(void *arg __attribute__((unused)))
 {
   sigset_t set;
   int sig;
-  my_thread_init();				// Init new thread
+
+  // thread is created, otherwise execution wouldn't reach this point
   signal_thread_in_use= 1;
 
+  // initialize this newly created thread
+  if (my_thread_init()) {
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: This can fail only in one of two ways:
 - calloc fails (extremely unlikely)
 - my_thread_global_init_done is true at start of my_thread_init
I originally wrote this on 11.4 and ported it back to 10.5 before posting.  Now running on latest 11.4, there are new macOS test failures.  I cannot be sure that there was a real world failure, and the slow startup and 'hanging' install are both solved without this change.  From the code perspective, calling `my_end` (which calls `my_thread_global_end` and sets `my_thread_global_init_done` to 0) will cause `my_thread_init` to return with an error, leaving the thread started but not initialized.  In the server, we call `my_end` only once but in tests we call it in multiple places.  During initial development, I had an `abort()` in the body of this conditional to see if we had a case where the init would fail, and I'm certain I saw it, but now cannot replicate.

DaveGosselin-MariaDB:
> I asked specifically because you did not remove that part of the test. It's line 95

Ahh... this was an oversight on my part, I removed that in the latest diff.

sql/mysqld.cc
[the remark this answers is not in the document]
vuvova: * my_thread_global_init_done is true at start of my_thread_init

is false, you mean. And I don't see how it can happen. The server calls `my_init()` at the beginning of `mysqld_main()`

May be it's set back to false because of some early shutdown? This sounds more plausible, but I still couldn't find how it could happen

DaveGosselin-MariaDB:
> Your commit essentially consists of three changes

Yes, this is true.

> * remove the loop in `wait_for_signal_thread_to_end()`
> * handle the case when `my_thread_init()` fails
> * terminate immediately on `SIGTERM`
> 
> The first looks rather risky to me. 

If it seems risky, then can we improve the automated tests for this part of our system to reduce that risk?

>The second — I don't really believe `my_thread_init()` can fail 

If this is true, then should we change `my_thread_init()` to have a `void` return type?

>the third one — looks rather clear and safe, and — incidentally — it's the only change that is needed to fix the bug. With that hunk only the server doesn't hang for me anymore.   Let's only do the third part for now?

Yes, no problem.  I will update the PR to include only that change.  I retested locally both on macOS and FreeBSD and can see that it's sufficient.

vuvova (approved):

PR #3254, opened 2024-05-14
MDEV-34156 InnoDB fails to apply the redo log for compressed tablespace

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34156*

## Description
Problem:
=======
- During recovery, InnoDB fails to apply the redo log for compressed tablespace. The reason is that
InnoDB assumes that pages has been freed while applying the redo log for it. InnoDB does multiple scan due to small buffer pool size. Problematic page has been freed and reinitialize multiple times. InnoDB stores the freed page information before it ran out of memory.
But InnoDB assigns the freed page ranges to tablespace in recv_init_crash_recovery_spaces() even though
InnoDB doesn't have complete freed range information.

Solution:
=========
rec_init_crash_recovery_spaces(): Don't add the freed ranges to tablespace if InnoDB has to read the redo log again.

## PR quality check

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3385,8 +3385,10 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
 			fil_names_dirty(rs.second.space);
 
 			/* Add the freed page ranges in the respective
-			tablespace */
-			if (!rs.second.freed_ranges.empty()
+			tablespace only if InnoDB doesn't need to
+			rescan the redo logs */
+			if (!rescan
+			    && !rs.second.freed_ranges.empty()
 			    && (srv_immediate_scrub_data_uncompressed
 				|| rs.second.space->is_compressed())) {
 


storage/innobase/log/log0recv.cc
```diff
@@ -3385,8 +3385,10 @@ recv_init_crash_recovery_spaces(bool rescan, bool& missing_tablespace)
...
 
 			/* Add the freed page ranges in the respective
-			tablespace */
-			if (!rs.second.freed_ranges.empty()
+			tablespace only if InnoDB doesn't need to
+			rescan the redo logs */
+			if (!rescan
+			    && !rs.second.freed_ranges.empty()
 			    && (srv_immediate_scrub_data_uncompressed
 				|| rs.second.space->is_compressed())) {
```
dr-m, line 3393: Based on this code change, this has something to do with `page_compressed=1` tables, and possibly also with `innodb_immediate_scrub_data_uncompressed=ON`.

dr-m (commented):
The scope and exact impact of this bug are unclear to me.

Is it possible to provide any form of test case for reproducing this?

dr-m (changes requested):
Thank you! This is tricky to review, because I was not able to reproduce this on my own system, no matter what I tried. On the shared environment where this was reproduced in the first place, I debugged the recovery of the data directory whose copy is attached to [MDEV-34156](https://jira.mariadb.org/browse/MDEV-34156), with and without the fix. The problem was that `store_freed_or_init_rec` was not being called on the BLOB page 0x2c:5 in the function with the template parameter `store=false`.

Unfortunately, this patch turns out to be an exact reversion of 941af1fa581a799e59ddc3afcae965852aeceb00 ([MDEV-31803](https://jira.mariadb.org/browse/MDEV-31803)). We need to understand this scenario better and revise the fix accordingly.

A possible answer is that the previous fix was incorrect, and something needs to be adjusted for the `immediate_scrub_data_uncompressed=ON` case elsewhere.

dr-m (approved):
Thank you, great work on the test cases. I confirmed that this will fix the reported issue without reintroducing the bug that was fixed in 941af1fa581a799e59ddc3afcae965852aeceb00. Yes, that change is being reverted, but there is another change to compensate for that.

It would be good to mention both 941af1fa581a799e59ddc3afcae965852aeceb00 and 2f9e264781f702b8da1ed418ac9f4f5e8f8aa843 (MDEV-29911) in the commit message as the "culprits".

PR #3255, opened 2024-05-15
MDEV-34169 Don't allow innodb_open_files to be lesser than number of non-user tablespace

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34169*

## Description
- InnoDB only closes the user tablespace when the number of open files exceeds innodb_open_files limit. In that case, InnoDB should make sure that innodb_open_files value should be greater than number of undo tablespace, system and temporary tablespace.

## How can this PR be tested?
./mtr innodb.open_files_limit 

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/open_files_limit.test
+++ b/mysql-test/suite/innodb/t/open_files_limit.test
@@ -1,0 +1,9 @@
+--source include/have_innodb.inc
+--source include/not_embedded.inc
+call mtr.add_suppression("\\[Warning\\] InnoDB: innodb_open_files=.* is not greater than the number of system tablespace files, temporary tablespace files, innodb_undo_tablespaces=.*");
+let SEARCH_PATTERN= \[Warning\] InnoDB: innodb_open_files=.* is not greater than the number of system tablespace files, temporary tablespace files, innodb_undo_tablespaces=.*;
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+
+CREATE TABLE t1(f1 INT NOT NULL)ENGINE=InnoDB;
+DROP TABLE t1;
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -3860,6 +3860,21 @@ static int innodb_init_params()
 		}
 	}
 
+	ulint min_open_files_limit = srv_undo_tablespaces
+				+ srv_sys_space.m_files.size()
+				+ srv_tmp_space.m_files.size() + 1;
+	if (min_open_files_limit > innobase_open_files) {
+		sql_print_warning(
+			"InnoDB: innodb_open_files=%lu is not greater "
+			"than the number of system tablespace files, "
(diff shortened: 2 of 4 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -3860,6 +3860,17 @@ static int innodb_init_params()
...
 
+	if (srv_undo_tablespaces + 2 >= innobase_open_files) {
+		ib::warn() << "innodb_open_files "
+			   << innobase_open_files << " should be "
+			   << "greater than total number "
+			   << "of undo tablespace, system & temporary "
+			   << "tablespace. So changing the value of "
+			   << "innodb_open_files to "
+			   << srv_undo_tablespaces + 3;
+		innobase_open_files = srv_undo_tablespaces + 3;
```
dr-m, line 3871: Can we please use `sql_print_warning()` here? Does `ib_logfile0` need to be included in the count?

For better common subexpression elimination, the condition could be written as follows:
```c++
if (srv_undo_tablespaces + 3 > innobase_open_files) {
```

dr-m (commented):
Can you provide a test script that shows what would happen when this fix is not present? The current test case does not even check that this warning is actually being issued.

storage/innobase/handler/ha_innodb.cc
Thirunarayanan: `fil_space_t::try_to_close()` tries to close only user tablespace. InnoDB doesn't create fil_space_t for redo log files.

mysql-test/suite/innodb/t/open_files_limit.test
```diff
@@ -0,0 +1,9 @@
+--source include/have_innodb.inc
```
dr-m, line 1: This seems to be missing
```sql
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -3860,6 +3860,16 @@ static int innodb_init_params()
 		}
 	}
 
+	if (srv_undo_tablespaces + 3 > innobase_open_files) {
+		sql_print_warning(
+			"InnoDB: innodb_open_files %ld should be "
+			"greater than total number of undo tablespace, "
+			"system & temporary tablespace. So changing "
+			"the value of innodb_open_files to %ld",
+			innobase_open_files, srv_undo_tablespaces + 3);
```
dr-m, line 3869: I’d suggest to use some more `=` in the message so that it becomes easier to understand for non-native English speakers:
```
InnoDB: innodb_open_files=%ld is smaller than 3+innodb_undo_tablespaces=%ld; adjusting to innodb_open_files=%ld
```
The message does not necessarily need to explain what the 3 is coming from.

By the way, should this take into account system or temporary tablespaces that consist of multiple files? Can you cover that this in the test case?

dr-m (approved):
Thank you, this looks OK.

One more thing that occurred to me is that we could easily extend the test with a multi-file `innodb_temp_data_file_path` to demonstrate if the lower limit of `innodb_open_files` needs to take multi-file system&temporary tablespaces into account.

PR #3256, opened 2024-05-15
MDEV-34166 Server could hang with BP < 80M under stress

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-34166*

## Description
BUF_LRU_MIN_LEN (256) is too high value for low buffer pool(BP) size. For example, for BP size lower than 80M and 16 K page size, the limit is more than 5% of total BP and for lowest BP 5M, it is 80% of the BP. Non-data objects like explicit locks could occupy part of the BP pool reducing the pages available for LRU. If LRU reaches minimum limit and if no free pages are available, server would hang with page cleaner not able to free any more pages.

Fix: To avoid such hang, we adjust the LRU limit lower than the limit for data objects as checked in buf_LRU_check_size_of_non_data_objects() i.e. one page less than 5% of BP.

## Release Notes
This could happen in rare case with BP size < 80M. Too many lock objects created with UPDATE, DELETE, INSERT INTO SELECT from same TABLE with queries over large range in RR or Serializable isolation could leads to the issue.

## How can this PR be tested?

./mtr innodb.lock_memory_debug

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/lock_memory_debug.result
+++ b/mysql-test/suite/innodb/r/lock_memory_debug.result
@@ -1,0 +1,13 @@
+#
+# MDEV-34166 Server could hang with BP < 80M under stress
+#
+call mtr.add_suppression("\\[Warning\\] InnoDB: Over 67 percent of the buffer pool");
+CREATE TABLE t1 (col1 INT) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1),(2),(3),(4),(5);
+SET STATEMENT debug_dbug='+d,innodb_skip_lock_bitmap' FOR
+INSERT INTO t1 SELECT a.* FROM t1 a, t1 b, t1 c, t1 d, t1 e, t1 f, t1 g LIMIT 45000;
+ERROR HY000: The total number of locks exceeds the lock table size
+SELECT COUNT(*) FROM t1;
+COUNT(*)
+5
+DROP TABLE t1;
--- a/mysql-test/suite/innodb/t/lock_memory_debug.test
+++ b/mysql-test/suite/innodb/t/lock_memory_debug.test
@@ -1,0 +1,19 @@
+--source include/have_innodb.inc
+--source include/have_debug.inc
+
+--echo #
+--echo # MDEV-34166 Server could hang with BP < 80M under stress
+--echo #
+
+call mtr.add_suppression("\\[Warning\\] InnoDB: Over 67 percent of the buffer pool");
+
+CREATE TABLE t1 (col1 INT) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1),(2),(3),(4),(5);
+
+--error ER_LOCK_TABLE_FULL
+SET STATEMENT debug_dbug='+d,innodb_skip_lock_bitmap' FOR
+INSERT INTO t1 SELECT a.* FROM t1 a, t1 b, t1 c, t1 d, t1 e, t1 f, t1 g LIMIT 45000;
+
+SELECT COUNT(*) FROM t1;
+
+DROP TABLE t1;
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -1300,9 +1300,21 @@ static void buf_flush_LRU_list_batch(ulint max, bool evict,
   static_assert(FIL_NULL > SRV_TMP_SPACE_ID, "consistency");
   static_assert(FIL_NULL > SRV_SPACE_ID_UPPER_BOUND, "consistency");
 
+  /* BUF_LRU_MIN_LEN (256) is too high value for low buffer pool(BP) size. For
+  example, for BP size lower than 80M and 16 K page size, the limit is more than
+  5% of total BP and for lowest BP 5M, it is 80% of the BP. Non-data objects
+  like explicit locks could occupy part of the BP pool reducing the pages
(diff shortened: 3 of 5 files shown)


mysql-test/suite/innodb/r/lock_memory_debug.result
```diff
@@ -0,0 +1,14 @@
+#
+# MDEV-34166 Server could hang with BP < 80M under stress
+#
+call mtr.add_suppression("\\[Warning\\] InnoDB: Over 67 percent of the buffer pool.*");
```
dr-m, line 4: `.*` is already implied at the start or end of each pattern.

mysql-test/suite/innodb/t/lock_memory_debug.test
```diff
@@ -0,0 +1,23 @@
+--source include/have_innodb.inc
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
```
dr-m, line 3: We do not seem to use `DEBUG_SYNC` here.

storage/innobase/buf/buf0flu.cc
```diff
@@ -1300,9 +1300,21 @@ static void buf_flush_LRU_list_batch(ulint max, bool evict,
...
+  example, for BP size lower than 80M and 16 K page size, the limit is more than
+  5% of total BP and for lowest BP 5M, it is 80% of the BP. Non-data objects
+  like explicit locks could occupy part of the BP pool reducing the pages
+  available for LRU. If LRU reaches minimum limit and if no free pages are
+  available, server would hang with page cleaner not able to free any more
+  pages. To avoid such hang, we adjust the LRU limit lower than the limit for
+  data objects as checked in buf_LRU_check_size_of_non_data_objects() i.e. one
+  page less than 5% of BP. */
+  auto pool_limit = buf_pool.curr_size / 20 - 1;
+  auto buf_lru_min_len = std::min(pool_limit, (ulint)BUF_LRU_MIN_LEN);
```
dr-m, line 1313: I think that we should use `size_t` instead of `auto` here, and write `std::min<size_t>` to avoid the  cast on the constant.

Monty wants to avoid a space before assignment operators:
```c++
size_t pool_limit= buf_pool.curr_size / 20 - 1;
```

mysql-test/suite/innodb/t/lock_memory_debug.test
```diff
@@ -0,0 +1,23 @@
...
+
+CREATE TABLE t1 (col1 INT) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1),(2),(3),(4),(5);
+
+SET DEBUG_DBUG="+d,innodb_skip_lock_bitmap";
+
+--error ER_LOCK_TABLE_FULL
+INSERT INTO t1 SELECT a.* FROM t1 a, t1 b, t1 c, t1 d, t1 e, t1 f, t1 g LIMIT 45000;
+
+SET DEBUG_DBUG="-d,innodb_skip_lock_bitmap";
```
dr-m, line 19: Unfortunately, the use of `-d,` will enable a "DBUG spam" mode because MariaDB does not include a related change that was made in MySQL 5.7. We should use the following pattern:
```sql
SET @save_dbug=@@debug_dbug;
SET debug_dbug='+d,innodb_skip_lock_bitmap';
# …
SET debug_dbug=@save_dbug;
```
Here, an even simpler tweak might work:
```sql
--error ER_LOCK_TABLE_FULL
SET STATEMENT debug_dbug='+d,innodb_skip_lock_bitmap' FOR
INSERT INTO t1 SELECT a.* FROM t1 a, t1 b, t1 c, t1 d, t1 e, t1 f, t1 g LIMIT 45000;
```

dr-m (commented):
The idea looks good. Did you check all other occurrences of `BUF_LRU_MIN_LEN`? Should some of the other references be adjusted?

If the limit is being consulted often, we might consider introducing a global variable `buf_pool.LRU_min_len` (protected by `buf_pool.mutex`) and adjust it in `buf_pool_t::resize()`.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: ok, sure.

mariadb-DebarunBanerjee:
> Did you check all other occurrences of BUF_LRU_MIN_LEN? 

Thanks for pointing it out. I did check the usage. 

1. In 10.5, the only other place it is used is to calculate the minimum BP size. Perhaps the initially it was thought to be sufficient to have just +20%. We don't want to change the min/max spec and it is better to keep this part as is.
  
2. In later versions, we are using the limit in other places but it is not critical and we should be able to carry on with the constant. 

> If the limit is being consulted often, we might consider introducing a global variable buf_pool.LRU_min_len (protected by buf_pool.mutex) and adjust it in buf_pool_t::resize().

I had considered this option and decided otherwise because ...

1. We are not really trying to have a design change where the limit gets modified with BP size constantly. It is rather a special case for low BP size (< 80M) while the constant today works fine in vast majority of the user scenarios. It could start impacting general cases if the constant is replaced by a variable.

2. Changing it to a variable would automatically move all the other usages in higher versions during merge and it could induce bugs implicitly which would be difficult to check/find during merge.

Since the scenario affects only limited  (< 80M BP) cases, I think it is much better to consider them case by case and modify only if there is any visible issue with it.

PR #3258, opened 2024-05-15
MDEV-34175 mtr_t::log_close() warning should change the shutdown condition

Thirunarayanan:

## Description
- InnoDB can print the wrong warning message  like "shutdown is in progress" even though InnoDB is in normal
condition. InnoDB should print the warning message saying "Shutdown is in progress" only when shutdown state is greater than SRV_SHUTDOWN_INITIATED.

## PR quality check


dr-m (commented):
This would seem to be applicable to 10.5 as well.

PR #3260, opened 2024-05-15
MDEV-28800 SIGABRT due to running out of memory for InnoDB locks

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-28800*

## Description
This regression is introduced in 10.6 by following commit. commit 898dcf93a886a68450dce7713208fdad82f9178e
(Cleanup the lock creation)
It removed one important optimization for lock bitmap pre-allocation. We pre-allocate about 8 byte extra space along with every lock object to adjust for similar locks on newly created records on the same page by same transaction. When it is exhausted, a new lock object is created with similar 8 byte pre-allocation. With this optimization removed we are left with only 1 byte pre-allocation. When large number of records are inserted and locked in a single page, we end up creating too many new locks almost in n^2 order.

Fix-1: Bring back LOCK_PAGE_BITMAP_MARGIN for pre-allocation.

Fix-2: Use the extra space (40 bytes) for bitmap in trx->lock.rec_pool.

## Release Notes
Fixes a regression in 10.6 - Server abort due to running out of lock memory while inserting large number of rows using INSERT INTO SELECT statement.

## How can this PR be tested?
./mtr innodb.lock_memory

## PR quality check


storage/innobase/lock/lock0lock.cc
```diff
@@ -1242,6 +1242,13 @@ lock_rec_create_low(
 		type_mode = type_mode & ~(LOCK_GAP | LOCK_REC_NOT_GAP);
 	}
 
+	/* Extra bitmap size in bytes over and above the current number of
+	records when a record lock is created. 8 x LOCK_PAGE_DEFAULT_BITMAP_SIZE
+	extra record locks of same type for newly inserted records can be added
+	without needing to create a new lock object. Useful when the number of
+	records in a page is growing. */
+	static const ulint LOCK_PAGE_DEFAULT_BITMAP_SIZE = 8;
```
dr-m, line 1250: I’d prefer `static constexpr size_t`.

PR #3263, opened 2024-05-16
MDEV-34167 We fail to terminate transaction early with ER_LOCK_TABLE_…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-34167*

## Description
This regression is introduced in 10.6 by following commit. commit b6a2472489accf0ae9ac3655ffe9b2997ab267ba
MDEV-27891: SIGSEGV in InnoDB buffer pool resize

During DML, we check if buffer pool is running out of data pages in buf_pool_t::running_out. Here, if 75% of the buffer pool is occupied by non-data pages, we rollback the current transaction and exit with ER_LOCK_TABLE_FULL.

The integer division (n_chunks_new / 4) becomes zero whenever the total number of chunks are < 4 making the check completely ineffective for such cases. Also the check is inaccurate for larger chunks.

Fix-1: Correct the check in buf_pool_t::running_out.

Fix-2: While waiting for free page, check for
buf_LRU_check_size_of_non_data_objects.

## Release Notes
When system is reaching out of memory in BP due to too many locks, it allows DMLs to exit before reaching the critical stage of server exit. 

## How can this PR be tested?
The test case from MDEV-34166 which is being fixed in 10.5 would also test this patch once merged. So. we would have the automated test running soon. Till that time it is difficult to add an automated test as it requires large BP and records in the order of millions to validate it.

In MDEV, I have updated the way it can be tested manually.

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0lru.cc
+++ b/storage/innobase/buf/buf0lru.cc
@@ -451,7 +451,10 @@ got_block:
     mysql_mutex_unlock(&buf_pool.flush_list_mutex);
     if (my_cond_timedwait(&buf_pool.done_free, &buf_pool.mutex.m_mutex,
                           &abstime))
+    {
       buf_pool.LRU_warn();
+      buf_LRU_check_size_of_non_data_objects();
+    }
   }
 
   goto got_block;
--- a/storage/innobase/include/buf0buf.h
+++ b/storage/innobase/include/buf0buf.h
@@ -1486,7 +1486,7 @@ public:
   {
     return !recv_recovery_is_on() &&
       UT_LIST_GET_LEN(free) + UT_LIST_GET_LEN(LRU) <
-        n_chunks_new / 4 * chunks->size;
+        (n_chunks_new * chunks->size) / 4;
   }
 
   /** @return whether the buffer pool is running low */


storage/innobase/buf/buf0lru.cc
```diff
@@ -450,8 +450,10 @@ buf_block_t *buf_LRU_get_free_block(bool have_mutex)
       buf_pool.page_cleaner_wakeup(true);
     mysql_mutex_unlock(&buf_pool.flush_list_mutex);
     if (my_cond_timedwait(&buf_pool.done_free, &buf_pool.mutex.m_mutex,
-                          &abstime))
+                          &abstime)) {
       buf_pool.LRU_warn();
```
dr-m, line 454: Monty wants do deviate from the K&R style and insert a line break before `{` except in a `switch` statement.

dr-m (approved):
Thank you. Should we also simplify `sel_set_rec_lock()`: declare it `static` instead of `UNIV_INLINE`, and remove the additional condition?
```c++
	if (UT_LIST_GET_LEN(trx->lock.trx_locks) > 10000
	    && buf_pool.running_out()) {
		return DB_LOCK_TABLE_FULL;
	}
```
All other code paths that return `DB_LOCK_TABLE_FULL` ignore the number of locks that are being held by the current transaction.

storage/innobase/buf/buf0lru.cc
mariadb-DebarunBanerjee: right, I should have followed the convention in existing code. 
Corrected.

mariadb-DebarunBanerjee:
> Should we also simplify sel_set_rec_lock():
> UT_LIST_GET_LEN(trx->lock.trx_locks) > 10000

Interesting observation but I think it is not void of impact and I would be sceptic making such changes part of this patch in older versions.  Perhaps we should consider in 11.* versions later.

PR #3264, opened 2024-05-16
MDEV-33935 fix deadlock counter

YVEF:
- [x] *The Jira issue number for this PR is: MDEV-33935*

## Description
Fixed deadlock counter incrementing on every call to cycle finding that was used multiple times during the deadlock detection.

## How can this PR be tested?
```
./mtr mdev-33935
```

## PR quality check


storage/innobase/lock/lock0lock.cc
```diff
@@ -317,6 +316,19 @@ namespace Deadlock
...
+  @param trx Transaction that is waiting for another transaction
+  @return a transaction that is part of a cycle, otherwise nullptr
+  @note the function is not fully idempotent, each call increments a counter */
+  static inline trx_t *detect_deadlock(trx_t *trx)
+  {
+      trx_t *res= nullptr;
+      if ((res = find_cycle(trx)) != nullptr)
+          lock_sys.deadlocks++;
+
+      return res;
```
dr-m, line 330: Instead of adding this function and adjusting both callers of `Deadlock::find_cycle()`, it would seem that we can move the `lock_sys.deadlocks++`  to the beginning of `Deadlock::report()`.

mysql-test/suite/innodb/t/mdev-33935.test
```diff
@@ -0,0 +1,56 @@
+--source include/have_innodb.inc
+--source include/have_debug_sync.inc
```
dr-m, line 2: I would add also `include/have_debug.inc` so that the test can be skipped more quickly on a non-debug build.

Even better would be to write this test without `DEBUG_SYNC` instrumentation, so that it could run on all target architectures. We only have a few debug instrumented builds, not covering all platforms.

There already exists a test `innodb.deadlock_detect`. Could you extend it so that it demonstrates the problem?

dr-m (changes requested):
Thank you for the contribution!

The commit message does not currently explain what is wrong. The [MDEV-33935](https://jira.mariadb.org/browse/MDEV-33935) Description does. I think that it would be better to make source code changes a little more self-contained, not forcing the Jira to be accessed for understanding the problem.

storage/innobase/lock/lock0lock.cc
YVEF: Fixed

mysql-test/suite/innodb/t/mdev-33935.test
YVEF: Based on your suggestion, there is no need to strictly align the test name with mdev-$bug -> mdev-$bug.test, so I decided to adapt the existing test ```deadlock_detect```. For me this is not so graceful, because it checking "parallel" functionality, but in the same logical scope. And if it's okey - it's okey.

dr-m (approved):
Thank you!

PR #3268, opened 2024-05-20
MDEV-34191: Make sure that Debian respects systemd disabled

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-34191*

## Description
Make sure that Debian respects systemd disabled by bumping to deb compat 11 which is available in Debian 10 and Ubuntu 20.04 and it provides better integration with systemd.

Start using dh_installsystemd which is new recommended way in compat 11

## Release Notes
Update Debian deb compat to 12

## PR quality check


debian/compat
```diff
@@ -1 +1 @@
-9
+11
```
tonychen2001, line 1: I'm not familiar with this but it's a bit confusing that the PR description states
>bumping to deb compat 12
but its actually changed to 11

debian/rules
```diff
@@ -191,6 +185,6 @@ override_dh_gencontrol:
 # white list file only starting from Debian Stretch and Ubuntu Xenial.
 # To find more, grep build logs for 'but is not installed to anywhere'.
 %:
-	dh $@ --parallel --with systemd --fail-missing
+	dh $@ --parallel
```
tonychen2001, line 188: Per https://manpages.debian.org/testing/debhelper/dh_missing.1.en.html

> Please note that in compat 11 and earlier without either of these options, dh_missing will silently do nothing. In compat 12, --list-missing is the default. In compat 13 and later, --fail-missing is the default.

From my understanding, this change will make it so that the build process will no longer fail on missing files. Is this intended and safe?

illuusio: I'll add it back as we are using compat 11 not compat 12 which is not building on Debian 10 proberly.

debian/compat
illuusio: I've reworded PR description. It was correct in commit message.

illuusio:
Testing should or should not we have this restart thing but noted that we should not as it was.. so revert to that as this test has passed

tonychen2001:
> Testing should or should not we have this restart thing but noted that we should not as it was.. so revert to that as this test has passed

I assume your referring to the `--restart-after-upgrade` option that was removed?

According to https://man7.org/linux/man-pages/man1/dh_installsystemd.1.html, the `--restart-after-upgrade` option is the default behaviour as of compat 10 for `dh_installsystemd`. So specifying the option is unnecessary since you are also bumping to compat 11.

---

The test failures are unrelated. Changes look good but I will leave it to someone with more experience with Debian packaging for a further review.

illuusio:
Thank you having read documentation more than me! Yes failures are unrelated to this. I'll update the bb branch to get the Deb packages for testing

vizovitin:
Are there any plans to merge this in the foreseeable future?

grooverdan (approved):

vizovitin:
The issue **is not fixed**. See the [MDEV-34191 comment](https://jira.mariadb.org/browse/MDEV-34191?focusedCommentId=289950&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-289950) for details.

PR #3195, opened 2024-04-12
MDEV-33817 WIP: AVX-512 based CRC-32 algorithms

dr-m:
This is based on https://github.com/intel/intel-ipsec-mb/ and not yet tested.

TODO: Test and clean up the code. The current variable names try to correspond to the register names in
intel-ipsec-mb/lib/avx512_t2/crc32_by16_vclmul_avx512.asm (note: XMM registers alias YMM which alias ZMM)

- [x] *The Jira issue number for this PR is: MDEV-______*

## Description
TODO: fill description here

## Release Notes
TODO: What should the release notes say about this change?
Include any changed system variables, status variables or behaviour. Optionally list any https://mariadb.com/kb/ pages that need changing.

## How can this PR be tested?

TODO: modify the automated test suite to verify that the PR causes MariaDB to behave as intended.
Consult the documentation on ["Writing good test cases"](https://mariadb.org/get-involved/getting-started-for-developers/writing-good-test-cases-mariadb-server).

If the changes are not amenable to automated testing, please explain why not and carefully describe how to test manually.

## Basing the PR against the correct MariaDB version
- [ ] *This is a new feature and the PR is based against the latest MariaDB development branch.*
- [ ] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [ ] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [ ] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #3270, opened 2024-05-21
MDEV-33817 fixup: Disable for macOS

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33817*
## Description
According to https://discussions.apple.com/thread/8256853 an attempt to use AVX512 registers on macOS will result in `#UD` (crash at runtime).

The AVX512 intrinsic functions may also be unavailable on Apple Xcode, at least in a version that is based on clang 10.

Furthermore, starting with clang-18 and GCC 14, we must add `evex512` to the target flags. We might also write avx10.1-512, but it could enable some AVX512 flags that we do not have any CPUID check for.
## Release Notes
Nothing; macOS on x86 is not an officially supported platform, and GCC 14 and clang-18 probably are not in any stable operating system distribution yet.

## How can this PR be tested?
See #3195.

## PR quality check

### DIFF
--- a/mysys/crc32/crc32c_x86.cc
+++ b/mysys/crc32/crc32c_x86.cc
@@ -24,7 +24,11 @@
 # endif
 #else
 # include <cpuid.h>
-# if __GNUC__ >= 11 || (defined __clang_major__ && __clang_major__ >= 8)
+# ifdef __APPLE__ /* AVX512 states are not enabled in XCR0 */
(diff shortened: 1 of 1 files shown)


vaintroub (approved):
Looks ok. Maybe worth mention why "evex512" was introduced (e.g _mm512_castsi128_si512 does not work without it on gcc14+)

PR #3272, opened 2024-05-21
MDEV-34200 InnoDB tries to write to read-only system tablespace in buf_dblwr_t::init_or_load_pages()

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34200*

## Description

- InnoDB fails to set the TRX_SYS_DOUBLEWRITE_SPACE_ID_STORED flag in transaction system header page while recreating the undo log tablespaces

buf_dblwr_t::init_or_load_pages(): Tries to reset the space id and try to write into doublewrite buffer even when read_only mode is enabled.

In srv_all_undo_tablespaces_open(), InnoDB should try to open the extra unused undo tablespaces instead of trying to creating it.

## How can this PR be tested?
./mtr innodb.undo_upgrade --mem

## PR quality check

### DIFF
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -370,6 +370,11 @@ inline dberr_t trx_sys_t::reset_page(mtr_t *mtr)
       sys_header->page.frame + TRX_SYS_DOUBLEWRITE
       + FSEG_HEADER_SIZE + TRX_SYS_DOUBLEWRITE_REPEAT,
       sys_header->page.frame + TRX_SYS_DOUBLEWRITE + FSEG_HEADER_SIZE, 12);
+    mtr->write<4>(
+      *sys_header,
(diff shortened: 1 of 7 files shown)


storage/innobase/srv/srv0start.cc
```diff
@@ -822,7 +827,7 @@ static dberr_t srv_all_undo_tablespaces_open(bool create_new_undo,
   {
      char name[OS_FILE_MAX_PATH];
      snprintf(name, sizeof name, "%s/undo%03u", srv_undo_dir, i);
-     uint32_t space_id= srv_undo_tablespace_open(create_new_undo, name, i);
+     uint32_t space_id= srv_undo_tablespace_open(false, name, i);
```
Thirunarayanan, line 830: Only this line should be applicable from 10.5 onwards.

dr-m (approved):

PR #3277, opened 2024-05-22
MDEV-34212 InnoDB transaction recovery is incorrect

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34212*
## Description
Starting with [MDEV-32042](https://jira.mariadb.org/browse/MDEV-32042), we assume that the function `buf_page_get_gen()` is not being invoked while log is being applied during crash recovery. The idea is that any access to buffer pool pages during crash recovery must look up pages via `recv_sys_t::recover()`. This was being violated by the function `trx_undo_mem_create_at_db_start()` during startup.

`trx_undo_mem_create_at_db_start()`: Invoke `recv_sys_t::recover()` instead of `buf_page_get_gen()`, so that all undo log pages will be recovered correctly. Failure to do this could prevent InnoDB from starting up due to `Data structure corruption`, or it could potentially lead to a situation where InnoDB starts up but some transactions were recovered incorrectly.

`recv_sys_t::recover()`: Only acquire a buffer-fix on the pages, not a shared latch. This is adequate protection, because this function is only being invoked during early startup when no "users" are modifying buffer pool pages. The only writes are due to server bootstrap (the data files being created) or crash recovery (changes from `ib_logfile0` being applied).

`buf_page_get_gen()`: Assert that the function is not invoked while crash recovery is in progress, and that the special mode `BUF_GET_RECOVER` is only invoked during crash recovery or server bootstrap.

All this should really have been part of commit 850d61736deba354377634cf344256ee850b93b7 (MDEV-32042).
## Release Notes
InnoDB crash recovery could fail with an error `Data structure corruption`.
## How can this PR be tested?
The failure is very hard to reproduce. I managed it on a MemorySanitizer instrumented build with
```bash
./mtr --parallel=60 innodb.table_flags,64k,strict_full_crc32{,,,,,}{,,,,}{,}
```
If the change to `trx_undo_mem_create_at_db_start()` is omitted, the added assertion in `buf_page_get_gen()` will fail easily.
## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2587,6 +2587,15 @@ buf_page_get_gen(
 {
 	ulint		retries = 0;
 
+	/* BUF_GET_RECOVER is only used by recv_sys_t::recover(),
+	which must be invoked during early server startup when crash
+	recovery may be in progress. The only case when it may be
+	invoked outside recovery is when dict_create() has initialized
+	a new database and is invoking dict_boot(). In this case, the
+	LSN will be small. */
+	ut_ad(mode == BUF_GET_RECOVER
+	      ? recv_recovery_is_on() || log_sys.get_lsn() < 50000
+	      : !recv_recovery_is_on() || recv_sys.after_apply);
 	ut_ad(!mtr || mtr->is_active());
 	ut_ad(mtr || mode == BUF_PEEK_IF_IN_POOL);
 	ut_ad((rw_latch == RW_S_LATCH)
@@ -2608,6 +2617,7 @@ buf_page_get_gen(
 		/* The caller may pass a dummy page size,
 		because it does not really matter. */
 		break;
+	case BUF_GET_RECOVER:
 	case BUF_GET:
 		ut_ad(!mtr->is_freeing_tree());
 		fil_space_t* s = fil_space_get(page_id.space());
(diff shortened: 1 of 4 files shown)


dr-m:
Sorry, the following still fails on my local MemorySanitizer instrumented build, and actually this case does not involve crash recovery, but the first startup right after server bootstrap. I will find out with `rr` what exactly is failing and why, and then revise this fix.
```bash
./mtr --parallel=60 innodb.table_flags,64k,strict_full_crc32{,,,,,,,,,}{,,,,,} --repeat=10
```

Thirunarayanan (approved):

mariadb-DebarunBanerjee (commented):
I don't think I can contribute much to this review. It is more related to the base refactoring done in MDEV-32042.

Looking at MDEV-32042, the change looks a bit risky and not very good for future. If the decision is based on state, it looks much cleaner and less error prone to have a single interface and make the decision there. 

If we don't have other dependencies my suggestion would be to revert MDEV-32042 and re-evaluate the refactoring.

However, I don't think the patch should block on my review if both @dr-m and @Thirunarayanan agree on it.

dr-m:
Let us wait for some test results. I would like to retain the 850d61736deba354377634cf344256ee850b93b7 logic, and actually take it one step forward, similar to what MySQL 8.0 does, and let the WAL recovery run into completion before any pages are accessed in the buffer pool. That is, a follow-up change would remove `recv_sys_t::recover()` altogether.

PR #3280, opened 2024-05-23
Update CMakeLists.txt Specify python version

whsir:
fixes
*** ERROR: ambiguous python shebang in /usr/local/mysql/bin/myrocks_hotbackup: #!/usr/bin/env python. Change it to python3 (or python2) explicitly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -425,7 +425,7 @@ INCLUDE(cpack_tgz)
 INCLUDE(cpack_rpm)
 INCLUDE(cpack_deb)
 
-SET(PYTHON_SHEBANG "/usr/bin/env python" CACHE STRING "python shebang")
+SET(PYTHON_SHEBANG "/usr/bin/env python3" CACHE STRING "python shebang")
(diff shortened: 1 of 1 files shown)


vuvova:
The next planned release date is at the [early August](https://mariadb.com/kb/en/mariadb-release-model/). And 10.4 is reaching End-of-Life [before that](https://mariadb.org/about/#maintenance-policy). Meaning, 10.4 no longer accepts bug fixes as they will never be released.

Also, where and how did you get this error?

whsir:
> 下一个计划发布日期是[8 月初。而 10.4](https://mariadb.com/kb/en/mariadb-release-model/)[在那之前](https://mariadb.org/about/#maintenance-policy)就已到达生命周期结束。这意味着 10.4 不再接受错误修复，因为它们永远不会发布。
> 
> 另外，您在哪里以及如何出现此错误的？

RockyLinux8\AlmaLinux8 cmake throws this error

vuvova:
Thanks.

It was fixed in commit 263f8aff651, present in all versions starting from 10.5.5

PR #3287, opened 2024-05-27
MDEV-34223  Innodb - add status variable for number of bulk inserts

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34223*

## Description
- Added a counter innodb_bulk_operations in INFORMATION_SCHEMA.GLOBAL_STATUS. This counter is incremented whenever a InnoDB undergoes bulk insert operation.

## How can this PR be tested?

## PR quality check


storage/innobase/row/row0ins.cc
```diff
@@ -2749,6 +2749,7 @@ row_ins_clust_index_entry_low(
 #else /* BTR_CUR_HASH_ADAPT */
 			index->table->bulk_trx_id = trx->id;
 #endif /* BTR_CUR_HASH_ADAPT */
+			export_vars.innodb_bulk_operations++;
```
dr-m, line 2752: Could we increment this on transaction commit? Shouldn’t we increment it also in native table-rebuilding `ALTER TABLE`?

In the current form of the patch, this status variable is not protected by anything, so there could be race conditions between concurrent threads updating it. It could be thinkable that an external tool would load data into multiple tables concurrently, and hence race conditions around this counter could lead to incorrect results (some updates of the counter being lost). If we updated this under transaction commit, we could use an existing mutex for protecting it, such as `lock_sys.wait_mutex`.

Another alternative would be to use `Atomic_relaxed` for the counter.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1104,6 +1104,10 @@ static SHOW_VAR innodb_status_variables[]= {
   {"encryption_num_key_requests", &export_vars.innodb_encryption_key_requests,
    SHOW_LONGLONG},
 
+  /* InnoDB bulk operations */
+  {"bulk_operations",
+   &export_vars.innodb_bulk_operations, SHOW_ULONG},
```
dr-m, line 1109: The comment in its current form is redundant, because it is nothing more than duplicating the variable name. If we want to add a comment, then it should be more descriptive.

I would use `SHOW_SIZE_T` and `size_t` for the parameter, to avoid having an artificial difference between LP64 and LLP64 platforms (most notably, Linux and Windows).

storage/innobase/row/row0ins.cc
dr-m: @vuvova just said that incrementing the counter at the start of the transaction is acceptable.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -1104,6 +1104,10 @@ static SHOW_VAR innodb_status_variables[]= {
   {"encryption_num_key_requests", &export_vars.innodb_encryption_key_requests,
    SHOW_LONGLONG},
 
+  /* InnoDB bulk operations */
+  {"num_bulk_insert_operation",
+   &export_vars.innodb_bulk_insert_operation, SHOW_SIZE_T},
```
dr-m, line 1109: The suggestion was `innodb_bulk_operations` in plural form. I don’t think that the components "num" or "insert" add much value.

storage/innobase/include/srv0srv.h
```diff
@@ -760,6 +760,9 @@ struct export_var_t{
 	ulint innodb_encryption_rotation_pages_flushed;
 	ulint innodb_encryption_rotation_estimated_iops;
 	int64_t innodb_encryption_key_requests;
+
+	/* Number of bulk insert operations */
+	Atomic_relaxed<ulint> innodb_bulk_insert_operation;
```
dr-m, line 765: I would suggest the following with no source code comment:
```c++
	Atomic_counter<ulint> innodb_bulk_operations;
```
While we are at this, I think that we should also change `innodb_instant_alter_column` to be of the same type. The two fields could be moved next to each other, because they are logically related.

I think that I previously requested that the counter `innodb_bulk_operations` be also incremented during table-rebuilding native `ALTER TABLE`. The following would seem to be the logical spot for that:
```diff
diff --git a/storage/innobase/handler/handler0alter.cc b/storage/innobase/handler/handler0alter.cc
index 3c5db74e17a..7ed787af22f 100644
--- a/storage/innobase/handler/handler0alter.cc
+++ b/storage/innobase/handler/handler0alter.cc
@@ -7357,6 +7357,7 @@ prepare_inplace_alter_table_dict(
 		ut_d(dict_table_check_for_dup_indexes(user_table,
 						      CHECK_PARTIAL_OK));
 		if (ctx->need_rebuild()) {
+			export_vars.innodb_bulk_insert_operations++;
 			ctx->new_table->acquire();
 		}
 
```

dr-m (changes requested):

mysql-test/suite/innodb/t/insert_into_empty.test
```diff
@@ -26,9 +26,19 @@ DROP TEMPORARY TABLE t;
...
+(SELECT variable_value FROM information_schema.global_status
+WHERE variable_name = 'innodb_bulk_operations');
+
 BEGIN;
 INSERT INTO t1 SELECT * FROM seq_1_to_65536;
 ROLLBACK;
+
+SELECT variable_value-@old_bulk_op bulk_operations
+FROM information_schema.global_status
+WHERE variable_name = 'innodb_bulk_operations';
```
dr-m, line 40: I was a bit uncertain about this, but based on other tests, it appears that `-` is defined on string parameters, by converting the strings to integers first. (The type of `information_schema.global_status.variable_value` is string, not integer.)

storage/innobase/include/srv0srv.h
```diff
@@ -760,6 +760,9 @@ struct export_var_t{
 	ulint innodb_encryption_rotation_pages_flushed;
 	ulint innodb_encryption_rotation_estimated_iops;
 	int64_t innodb_encryption_key_requests;
+
+	/* Number of bulk insert operations */
+	Atomic_relaxed<ulint> innodb_bulk_operations;
```
dr-m, line 765: The comment is misleading, because the counter is also being incremented on some `ALTER TABLE`.

Why do you not use `Atomic_counter` like I suggested? The end result should be the same, though. I also requested to move this counter next to `innodb_instant_alter_column`.

dr-m (approved):

PR #3288, opened 2024-05-27
MDEV-21322: Report slave progress to the master

bnestere:
This PR presents a patch to extend the command `SHOW REPLICA HOSTS` with three columns:

 1) Gtid_State_Sent. This represents that latest GTIDs sent to the
replica in each domain. It will always be populated, regardless of the semi-sync status (i.e. asynchronous connections will still update this column with the latest GTID state sent to the replica).

 2) Gtid_State_Ack. For semi-synchronous connections (only), this
column represents the last GTID in each domain that the replica has acknowledged.

 3) Sync_Status. This value represents the synchronization status of
the replica, and is used to help determine how to interpret the Gtid_State_Ack column. There are four possible values:

   3.1) Initializing. This means the binlog dump thread is still
initializing, and has not yet determined the synchronization status of the replica.

   3.2) Asynchronous: This means the replica is not configured for
semi-sync replication, and thereby, Gtid_State_Ack should always be empty.

   3.3) Semi-sync Stale: This means the replica is configured for
semi-sync replication, however, connected using an old state, and is not readily able to send ACKs for new transactions. Functionally, this means that the primary will try to catch the replica up-to-date by sending transactions which will not be ACKed. Additionally, the value shown by Gtid_State_Ack will be empty until the replica catches up and ACKs its first transaction.

   3.4) Semi-sync Active: This means the replica is configured for
semi-sync replication, and is readily sending ACKs for new transactions it receives. It is possible for Gtid_State_Ack to be empty while Sync_Status is "Semi-sync Active" if no new transactions have been executed on the primary since the replica has connected.

Additionally, this patch creates a new semantic for the configuration rpl_semi_sync_master_timeout=0. That is, now when 0, 1) new transactions will not attempt to wait for an ACK before completing, and 2) the primary will still request ACKs from the replica for new transactions. This means that Gtid_State_Ack will be updated for each ACK from the replica and Sync_Status will read as "Semi-sync Active". Effectively, this creates a mode to mimic the asynchronous connection behavior, while allowing one to monitor the progress at which the primary is sending transactions to the replica via the new columns Gtid_State_Sent and Gtid_State_Ack.

Also note that a new error message was added to account for the case that Gtid_State_(Sent/Ack) represents a binary log file that was purged/cannot be found.

The overall implementation is rather simple. It leverages the existing semi-sync framework, where the replica uses binlog file:pos to ACK transactions, in order to infer GTID state by performing a binlog lookup at the time `SHOW REPLICA HOSTS` is executed. In particular, the Slave_info struct is extended to store 1) the binlog file:pos pair of the transaction which was last sent to the replica, 2) the binlog file:pos pair that was last ACKed by the replica, and 3) and enum to represent the Sync_Status.

This patch was initially started by @JackSlateur in PR#1427, where it was then transferred to @an3l who buffed it out in PR#2374, and final touches were put on by @bnestere.

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_mixed_ddl_dml.result
+++ b/mysql-test/suite/rpl/r/rpl_mixed_ddl_dml.result
@@ -11,8 +11,8 @@ n
 2002
 connection master;
 show slave hosts;
-Server_id	Host	Port	Master_id
-2	127.0.0.1	9999	1
+Server_id	Host	Port	Master_id	Gtid_State_Sent	Gtid_State_Ack	Sync_Status
+2	127.0.0.1	9999	1	0-1-2		Asynchronous
 drop table t1;
 connection slave;
 stop slave;
--- a/sql/repl_failsafe.cc
+++ b/sql/repl_failsafe.cc
@@ -37,18 +37,7 @@
 #include "rpl_filter.h"
 #include "log_event.h"
 #include <mysql.h>
-
-
-struct Slave_info
-{
-  uint32 server_id;
-  uint32 master_id;
-  char host[HOSTNAME_LENGTH*SYSTEM_CHARSET_MBMAXLEN+1];
-  char user[USERNAME_LENGTH+1];
-  char password[MAX_PASSWORD_LENGTH*SYSTEM_CHARSET_MBMAXLEN+1];
-  uint16 port;
-};
-
+#include "semisync_master.h"
 
 Atomic_counter<uint32_t> binlog_dump_thread_count;
 ulong rpl_status=RPL_NULL;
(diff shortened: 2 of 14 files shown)


knielsen (commented):
Hi Brandon,

Review in mail:
https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/thread/62YVCBGSC23PDTPHSEBU4LH74LZVGJD7/

mysql-test/suite/rpl/r/rpl_mixed_ddl_dml.result
```diff
@@ -11,8 +11,8 @@ n
 2002
 connection master;
 show slave hosts;
-Server_id	Host	Port	Master_id
-2	127.0.0.1	9999	1
+Server_id	Host	Port	Master_id	Gtid_State_Sent	Gtid_State_Ack	Sync_Status
+2	127.0.0.1	9999	1	0-1-2		Asynchronous
```
emoonrain, line 15: Have you though/discussed how/if represent "nil" of `Gtid_State_Ack`?
It feels the blank field is not a perfect way.  E.g `NULL` would do imo also to infer `Async...` of the status.

sql/repl_failsafe.cc
```diff
@@ -191,6 +186,50 @@ static my_bool show_slave_hosts_callback(THD *thd, Protocol *protocol)
...
     protocol->store((uint32) si->port);
     protocol->store(si->master_id);
+
+    if (gtid_state_from_binlog_pos(si->gtid_state_sent.log_file,
+                                   (uint32) si->gtid_state_sent.log_pos,
+                                   &gtid_sent, &err_msg))
+    {
+      gtid_sent.length(0);
+      DBUG_ASSERT(err_msg);
+      if (global_system_variables.log_warnings >= 2)
```
emoonrain, line 196: I am sure about the `warning` level. `info` would be carrying lesser :-) incrimination.

emoonrain (commented):
I like the whole work. Couple of aesthetic related notes made.
To the description I would consider here
```
 Additionally, this patch creates a new semantic for the
configuration rpl_semi_sync_master_timeout=0. That is, now when 0...
```
to clarify more with something like
```
In other words such slave's commit state is somewhere in between of `Sent` and  `Ack`.

PR #3290, opened 2024-05-29
MDEV-34256 InnoDB throws out of bound write due to temporary tablespace truncation

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34256*

## Description
 InnoDB fails with out of bound write error after temporary
tablespace truncation. InnoDB fail to clear freed ranges
if shrinking size is the last offset of the freed range.

## How can this PR be tested?
./mtr innodb.temp_truncate_freed

## PR quality check


dr-m (approved):
The commit message had better explicitly mention
```
commit c507678b207a1a8ce6d99ea28fb947bed1d95795 (MDEV-28699)
```
as the source of the regression because c507678b207a1a8ce6d99ea28fb947bed1d95795 added the function `fil_space_t::clear_freed_ranges()`.

PR #3293, opened 2024-05-29
Revert "MDEV-32935: Remove unneeded CMAKE_SYSTEM_PROCESSOR parameter …

LinuxJedi:
…from Debian"

This reverts commit 6914b7804dd29b71f9a85a120f6352d8cd855e38.

With the MDEV-32935 change, building on 32bit Linux with a 64bit CPU causes CMAKE_SYSTEM_PROCESSOR to detect 64bit. This then triggers things such as ColumnStore to build, which definitely should not build on a 32bit platform.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -89,6 +89,7 @@ endif
 	    -DCOMPILATION_COMMENT="mariadb.org binary distribution" \
 	    -DMYSQL_SERVER_SUFFIX="-$(DEB_VERSION_REVISION)" \
 	    -DSYSTEM_TYPE="debian-$(DEB_HOST_GNU_SYSTEM)" \
+	    -DCMAKE_SYSTEM_PROCESSOR=$(DEB_HOST_ARCH) \
 	    -DBUILD_CONFIG=mysql_release \
 	    -DPLUGIN_TOKUDB=NO \
 	    -DPLUGIN_CASSANDRA=NO \


LinuxJedi:
Assigned @illuusio to check this as I don't want this to break MDEV-32936.

grooverdan (changes requested):
reject: We shouldn't hack our way around not wanting to build 32bit as a cross compile (which is what 32 bit builds on 64 bit OS are).

vuvova:
@grooverdan cmake must know whether it's a 32-bit build or a 64-bit build. @LinuxJedi says with this PR that the correct way of determining it is to examine the `CMAKE_SYSTEM_PROCESSOR` variable. If you think it's a hack, you need tell what you think is the correct way of getting this information.

vaintroub:
I think the correct way to determine bitness of the build is to examine CMAKE_SIZEOF_VOID_P variable, like it is done in multiple places https://cmake.org/cmake/help/latest/variable/CMAKE_SIZEOF_VOID_P.html

vuvova:
it's not only that. ColumnStore only supports amd64 (x86_64) and arm64 (aarch64). So "any 64-bit" is not good enough.

vaintroub:
I think for the purpose of excluding columnstore from 32bit builds, it would work pretty well. CMAKE_SIZEOF_VOID_P is more reliably than whatever people pass in CMAKE_SYSTEM_PROCESSOR

vuvova:
Currently ColumnStore has
```cmake
IF(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR
CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64" OR
CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
```

Do you suggest to replace it with  `CMAKE_SIZEOF_VOID_P EQUAL 8` ? It'll allow ppc64. Or to add the new condition with `AND`? Still relying on whatever people pass in `CMAKE_SYSTEM_PROCESSOR`?

vaintroub:
At least testing for CMAKE_SIZEOF_VOID_P EQUAL 8 is better than now. Much better in fact. If  the know they do not build on 32bit.
All these architectures are some predefined constants in gcc (I assume they restrict themselves only to at most gcc and clang). 
So they can write a test to test for those __x86_64_ and whatever other architecture they probably already use in their code.

grooverdan:
`DEB_HOST_ARCH` isn't a correctly formatted processor variable. It contains Debian's weird interpretations (string differences - ppc64el e.g. for a little endian pun on everyone's expense) of what architecture is that don't actually occur on a native build.  To use this hack we'd be maintaining multiple hacks of what architecture is throughout the codebase and `IF(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64"` is a example of this hack. We've got enough different interpretations via Windows and BSDs so lets not add another one.

A proper cmake cross compilation would have the right value and also ensure the other CMake variables include the actual representation of the target hardware.

@illuusio  is looking at something like `setarch i686 eatmydata dpkg-buildpackage  ...` for debian/autobake-deb.sh which fakes the `uname -m`  output. Apparently what Debian does in its build architecture for all the supposed Intel/AMD 32bit users out there.

grooverdan:
or maybe like http://crossqa.debian.net/src/mariadb in autobake-deb.sh when there's not a native arch.
```
dpkg-buildpackage --sanitize-env -ai386 -Pcross,nocheck -us -uc -B -rfakeroot --jobs-try=1
```

vuvova:
@grooverdan I'm not saying `CMAKE_SYSTEM_PROCESSOR` is the correct way of determining the target architecture. I'm just saying, I'm sure there is one.

@vaintroub basically suggests TRY_COMPILE with, like
```c
#if !defined(__x86_64__) && !defined(__aarch64__)
#error
#endif
```
which will definitely work and perhaps we'll have to resort to this. But I'd be very surprised if cmake doesn't provide any builtin way of checking the target architecture.

LinuxJedi:
> reject: We shouldn't hack our way around not wanting to build 32bit as a cross compile (which is what 32 bit builds on 64 bit OS are).

This explicitly wasn't for a 64bit OS, but the method CMake uses can still detect a 64bit CPU on a 32bit OS when you are not cross-compiling.

grooverdan:
close since #3303 has fixed it?

LinuxJedi-Test (approved):
Test Review

PR #3294, opened 2024-05-29
MDEV-33616 Fix memleak in pfs_noop

DaveGosselin-MariaDB:
Invoke cleanup routine at the end of pfs_noop.

### DIFF
--- a/storage/perfschema/unittest/pfs_noop-t.cc
+++ b/storage/perfschema/unittest/pfs_noop-t.cc
@@ -240,6 +240,7 @@ int main(int, char **)
 
   MY_INIT("pfs_noop-t");
   test_noop();
+  my_end(MY_CHECK_ERROR);
   return (exit_status());
 }
 


vuvova (commented):
Let's split this in two. A fix for memleak in pfs_noop.c is obviously ok. As for the safemalloc changes, perhaps they should be inside my_addr_resolve? And definitely they should be in a separate commit.

DaveGosselin-MariaDB:
> Let's split this in two. A fix for memleak in pfs_noop.c is obviously ok. As for the safemalloc changes, perhaps they should be inside my_addr_resolve? And definitely they should be in a separate commit.

No problem, will do.  We don't need any changes to `my_addr_resolve` as the implementation compiled on macOS (the  `MY_ADDR_RESOLVE_FORK` `#elif` branch) appears to be working.

vuvova (approved):

PR #3295, opened 2024-05-29
MDEV-33616 Show correct function names in safemalloc backtrace

DaveGosselin-MariaDB:
Fix safemalloc's backtrace functionality to work on macOS.

### DIFF
--- a/mysys/safemalloc.c
+++ b/mysys/safemalloc.c
@@ -224,6 +224,11 @@ static void print_stack(void **frame)
 {
   const char *err;
   int i;
+#ifdef __APPLE__
+  char** strptr= 0;
+#else
+  my_addr_loc loc;
+#endif
 
   if ((err= my_addr_resolve_init()))
   {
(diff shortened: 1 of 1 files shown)


vuvova:
sorry, I've missed it when it was created. Why safemalloc backtrace functionality doesn't work on Mac OS X?

DaveGosselin-MariaDB:
> sorry, I've missed it when it was created. Why safemalloc backtrace functionality doesn't work on Mac OS X?

The `addr2line` doesn't exist on macOS, so we cannot rely on it for address resolution during `my_addr_resolve`.  `addr2line` is available via other third-party package systems (like `ports`) but we shouldn't assume it's on the system.  Before this change, a test with a memory leak would produce a failure like this on macOS:

Error: Freeing unallocated data or underrun buffer 0x1278ebe38 at 0x101da0870, Printing to addr2line failed

With this change applied, the same error now resolves function names:

Error: Freeing unallocated data or underrun buffer 0x1318da038 at 0   mariadbd                            0x0000000105ba484c sf_free + 52, 1   mariadbd                            0x0000000105b8c420 my_free + 284, 2   mariadbd                            0x00000001050a3a30 _ZN25Sys_var_gtid_binlog_state13global_updateEP3THDP7set_var + 376, 3   mariadbd                            0x0000000104ccf294 _ZN7sys_var6updateEP3THDP7set_var + 156, 4   mariadbd                            0x0000000104cd1dd8 _ZN7set_var6updateEP3THD + 64, 5   mariadbd                            0x0000000104cd1620 _Z17sql_set_variablesP3THDP4ListI12set_var_baseEb + 408, 6   mariadbd                            0x0000000104e65394 _Z21mysql_execute_commandP3THD + 30360, 7   mariadbd                            0x0000000104e55ec8 _Z11mysql_parseP3THDPcjP12Parser_statebb + 1092

This isn't the most attractive output, but it is more helpful than a list of addresses.

vuvova:
No, that's not how it should work. True, addr2line might be not installed on Mac OS X, but it might be not installed on Linux either. Which is often the case in production.

We can have a generic fallback inside `my_addr_resolve.c` to use another resolver if addr2line couldn't be forked.

vuvova (changes requested):
No, that's not how it should work. True, addr2line might be not installed on Mac OS X, but it might be not installed on Linux either. Which is often the case in production.

We can have a generic fallback inside `my_addr_resolve.c` to use another resolver if addr2line couldn't be forked.

DaveGosselin-MariaDB:
I closed this PR because (1) we can use `objdump` on macOS to decode the addresses and (2) I'm not sure of an acceptable fallback that will work across operating systems.

PR #3298, opened 2024-05-30
MDEV-34265 Possible hang during IO burst with innodb_flush_sync enabled

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-34265*

## Description
When checkpoint age goes beyond the sync flush threshold and buf_flush_sync_lsn is set, page cleaner enters into "furious flush" stage to aggressively flush dirty pages from flush list and pull checkpoint LSN above safe margin. In this stage, page cleaner skips doing LRU flush and eviction.

In 10.6, all other threads entirely rely on page cleaner to generate free pages. If free pages get over while page cleaner is busy in "furious flush" stage, a session thread could wait for free page in the middle of a min-transaction(mtr) while holding latches on other pages.

It, in turn, can prevent page cleaner to flush such pages preventing checkpoint LSN to move forward creating a deadlock situation. Even otherwise, it could create a stall and hang like situation for large BP with plenty of dirty pages to flush before the stage could finish.

Fix: During furious flush, check and evict LRU pages after each flush iteration.

## Release Notes
Fixes a hang and possible stall under stress. When buffer pool runs out of free pages and checkpoint age is high in comparison to redo log capacity.

## How can this PR be tested?
Best reproduced with innodb-page-size=4k. Usually repeats within 5 to 10 runs.
./mtr --mem --force-restart --repeat=50 innodb.insert_into_empty

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/insert_into_empty.test
+++ b/mysql-test/suite/innodb/t/insert_into_empty.test
@@ -163,9 +163,9 @@ DROP TABLE t1,t2;
 
 CREATE TABLE t1 (a INT KEY) ENGINE=InnoDB;
 
---error 0,1193
+--error 0,ER_UNKNOWN_SYSTEM_VARIABLE
 SET @save_limit = @@GLOBAL.innodb_limit_optimistic_insert_debug;
---error 0,1193
+--error 0,ER_UNKNOWN_SYSTEM_VARIABLE
 SET GLOBAL innodb_limit_optimistic_insert_debug = 2;
 
 BEGIN;
@@ -289,4 +289,30 @@ CREATE TABLE t1(a INT KEY)ENGINE=InnoDB
 INSERT INTO t1 VALUES(1);
 UPDATE t1 SET a = 2 WHERE a = 1;
 DROP TABLE t1;
+
+--echo #
+--echo #  MDEV-34265 Possible hang during IO burst with innodb_flush_sync enabled
+--echo #
+CREATE TABLE t1(f1 MEDIUMTEXT)ENGINE=InnoDB;
+
+--error 0,ER_UNKNOWN_SYSTEM_VARIABLE
+SET @save_dbug=@@GLOBAL.debug_dbug;
+--error 0,ER_UNKNOWN_SYSTEM_VARIABLE
+SET @@GLOBAL.debug_dbug='+d,ib_page_cleaner_sleep';
+
+if ($have_debug) {
+  SET STATEMENT debug_dbug='+d,ib_free_page_sleep' FOR
+  INSERT INTO t1 VALUES(REPEAT(1, 8459264));
+}
+if (!$have_debug) {
+  --echo SET STATEMENT debug_dbug='+d,ib_free_page_sleep' FOR
+  INSERT INTO t1 VALUES(REPEAT(1, 8459264));
+}
+
+--error 0,ER_UNKNOWN_SYSTEM_VARIABLE
+SET @@GLOBAL.debug_dbug=@save_dbug;
+
+SELECT length(f1) FROM t1;
+DROP TABLE t1;
+
 --echo # End of 10.6 tests
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -2025,70 +2025,94 @@ ATTRIBUTE_COLD void buf_flush_ahead(lsn_t lsn, bool furious)
 /** Conduct checkpoint-related flushing for innodb_flush_sync=ON,
 and try to initiate checkpoints until the target is met.
 @param lsn   minimum value of buf_pool.get_oldest_modification(LSN_MAX) */
-ATTRIBUTE_COLD static void buf_flush_sync_for_checkpoint(lsn_t lsn)
+ATTRIBUTE_COLD ATTRIBUTE_NOINLINE
+static void buf_flush_sync_for_checkpoint(lsn_t lsn)
 {
   ut_ad(!srv_read_only_mode);
   mysql_mutex_assert_not_owner(&buf_pool.flush_list_mutex);
 
-  for (;;)
+  /* During furious flush, we need to keep generating free pages. Otherwise
+  concurrent mtrs could be blocked holding latches for the pages to be flushed
+  causing deadlock in rare occasion.
+
+  Ideally we should be acquiring buffer pool mutex for the check but it is more
+  expensive and we are not using the mutex while calling need_LRU_eviction() as
+  of today. It is a quick and dirty read of the LRU and free list length.
+  Atomic read of try_LRU_scan should eventually let us do the eviction.
+  Correcting the inaccuracy would need more consideration to avoid any possible
+  performance regression. */
+  if (buf_pool.need_LRU_eviction())
   {
-    if (ulint n_flushed= buf_flush_list(srv_max_io_capacity, lsn))
-    {
-      MONITOR_INC_VALUE_CUMULATIVE(MONITOR_FLUSH_SYNC_TOTAL_PAGE,
-                                   MONITOR_FLUSH_SYNC_COUNT,
-                                   MONITOR_FLUSH_SYNC_PAGES, n_flushed);
-    }
-
-    switch (srv_file_flush_method) {
-    case SRV_NOSYNC:
-    case SRV_O_DIRECT_NO_FSYNC:
-      break;
-    default:
-      fil_flush_file_spaces();
-    }
-
-    mysql_mutex_lock(&log_sys.mutex);
-    const lsn_t newest_lsn= log_sys.get_lsn();
-    mysql_mutex_lock(&log_sys.flush_order_mutex);
     mysql_mutex_lock(&buf_pool.flush_list_mutex);
-    lsn_t measure= buf_pool.get_oldest_modification(0);
-    mysql_mutex_unlock(&log_sys.flush_order_mutex);
-    const lsn_t checkpoint_lsn= measure ? measure : newest_lsn;
-
-    if (!recv_recovery_is_on() &&
-        checkpoint_lsn > log_sys.last_checkpoint_lsn + SIZE_OF_FILE_CHECKPOINT)
-    {
-      mysql_mutex_unlock(&buf_pool.flush_list_mutex);
-      log_checkpoint_low(checkpoint_lsn, newest_lsn);
-      mysql_mutex_lock(&buf_pool.flush_list_mutex);
-      measure= buf_pool.get_oldest_modification(LSN_MAX);
-    }
-    else
-    {
-      mysql_mutex_unlock(&log_sys.mutex);
-      if (!measure)
-        measure= LSN_MAX;
-    }
-
-    mysql_mutex_assert_not_owner(&log_sys.mutex);
-
-    /* After attempting log checkpoint, check if we have reached our target. */
-    const lsn_t target= buf_flush_sync_lsn;
-
-    if (measure >= target)
-      buf_flush_sync_lsn= 0;
-    else if (measure >= buf_flush_async_lsn)
-      buf_flush_async_lsn= 0;
-
-    /* wake up buf_flush_wait() */
-    pthread_cond_broadcast(&buf_pool.done_flush_list);
+    buf_pool.page_cleaner_set_idle(false);
+    buf_pool.n_flush_inc();
     mysql_mutex_unlock(&buf_pool.flush_list_mutex);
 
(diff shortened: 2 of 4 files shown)


storage/innobase/buf/buf0flu.cc
```diff
@@ -2347,9 +2364,12 @@ static void buf_flush_page_cleaner()
       }
       while (false);
 
+      mysql_mutex_lock(&buf_pool.flush_list_mutex);
       if (!buf_pool.need_LRU_eviction())
```
dr-m, line 2368: This check actually depends on `buf_pool.mutex`, not `buf_pool.flush_list_mutex`. Maybe we should remove the ThreadSanitizer suppression and add `mysql_mutex_assert_owner(&mutex)` to `buf_pool_t::need_LRU_eviction()`? It would seem to be the proper thing to do with respect to the bug scenario that we are attempting to fix. Note that the `buf_pool.mutex` would have to be acquired before `buf_pool.flush_list_mutex`. But, we should also keep in mind that the result of this check would be stale as soon as we no longer hold `buf_pool.mutex`.

storage/innobase/buf/buf0lru.cc
```diff
@@ -406,17 +406,32 @@ buf_block_t *buf_LRU_get_free_block(bool have_mutex)
...
+    DBUG_EXECUTE_IF("ib_free_page_sleep",
+    {
+      static bool do_sleep = true;
+      if (do_sleep && signalled)
+      {
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+        do_sleep = false;
+      }
+    });
+
```
dr-m, line 434: This is debug injection code for one particular test, so this does not matter much. Nevertheless, I can’t resist some nitpicking:
* We could use `buf_pool.mutex` to protect the shared variable `do_sleep`. However,  in the `!have_mutex` case we would already have released that mutex.
* It would be more efficient to let global or `static` variables to be zero-initialized, so that no special initialization code has to be emitted. I would suggest `static bool skip_sleep;`.
* It would be more efficient to check the local variable `signalled` before loading the global variable from memory.
```c++
static bool skip_sleep;
if (signalled && !skip_sleep)
{
  skip_sleep= true;
  if (!have_mutex)
    mysql_mutex_unlock(&buf_pool.mutex);
  std::this_thread::sleep_for(std::chrono::milliseconds(10));
  if (!have_mutex)
    mysql_mutex_lock(&buf_pool.mutex);
}
```

mysql-test/suite/innodb/t/insert_into_empty.test
```diff
@@ -279,4 +279,21 @@ CREATE TABLE t1(a INT KEY)ENGINE=InnoDB
...
+--echo #
+CREATE TABLE t1(f1 MEDIUMTEXT)ENGINE=InnoDB;
+
+SET @save_dbug=@@GLOBAL.debug_dbug;
+SET @@GLOBAL.debug_dbug='+d,ib_page_cleaner_sleep';
+
+SET STATEMENT debug_dbug='+d,ib_free_page_sleep' FOR
+INSERT INTO t1 VALUES(REPEAT(1, 8459264));
+
+SET @@GLOBAL.debug_dbug=@save_dbug;
```
dr-m, line 294: These need to be guarded with
```sql
--error 0,1193
```
like elsewhere in the test, or better yet, all occurrences should be changed to
```sql
--error 0,ER_UNKNOWN_SYSTEM_VARIABLE
```
We want this test to be executed also on non-debug builds, to have wide platform coverage.  If needed, you may write like this:
```sql
if ($have_debug) {
SET STATEMENT debug_dbug='+d,ib_free_page_sleep' FOR
INSERT INTO t1 VALUES(REPEAT(1, 8459264));
}
if (!$have_debug) {
--echo SET STATEMENT debug_dbug='+d,ib_free_page_sleep' FOR
INSERT INTO t1 VALUES(REPEAT(1, 8459264));
}
```
The reason why I suggest this trick is that in this way, we can avoid having a `,debug.rdiff` file that could become a maintenance burden.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2286,11 +2278,36 @@ static void buf_flush_page_cleaner()
...
+      concurrent mtrs could be blocked holding latches for the pages to be
+      flushed causing deadlock in rare occasion. */
+      mysql_mutex_lock(&buf_pool.flush_list_mutex);
+      buf_pool.page_cleaner_set_idle(false);
+      bool need_eviction= buf_pool.need_LRU_eviction();
+      mysql_mutex_unlock(&buf_pool.flush_list_mutex);
+
+      if (need_eviction)
+      {
+        mysql_mutex_lock(&buf_pool.mutex);
```
dr-m, line 2300: `need_eviction` is actually protected by `buf_pool.mutex`, so we’d better acquire that mutex before invoking the check. This code branch is not time critical; we can afford to widen the critical section of `buf_pool.mutex` here.

dr-m (commented):
Thank you. I have some quick initial comments. I did not review the changes to `buf_flush_sync_for_checkpoint()` yet.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2018,65 +2018,57 @@ ATTRIBUTE_COLD static void buf_flush_sync_for_checkpoint(lsn_t lsn)
...
 
-    lsn= std::max(lsn, target);
+  if (measure >= target)
+    buf_flush_sync_lsn= 0;
+  else if (measure >= buf_flush_async_lsn)
+    buf_flush_async_lsn= 0;
 
-    if (measure >= lsn)
-      return;
-  }
```
dr-m, line 2079: The actual change to this function is the removal of these lines as well as the removal of the `for (;;) {` near the start of the function. The remaining difference is only due to deindenting the former `for` body by 2 spaces.

An alternative fix that would not only result in a smaller difference region but also avoid expanding `buf_flush_page_cleaner()` for this rare-case scenario would be to retain the `for (;;)` loop and its above termination condition, but add calls to `buf_pool.need_LRU_eviction()` and `buf_flush_LRU()` to this function. The debug instrumentation `ib_page_cleaner_sleep` might have to be moved to this function (and renamed accordingly) as well.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: yes, thanks for pointing it out. Surprisingly we never acquire buf_pool.mutex before calling need_LRU_eviction(). May be performance reasons as the atomic variable try_LRU_scan would always be correct eventually. I added buf_pool.flush_list_mutex to make it similar to all other calls in the function but I think is not necessary. I have removed this change.

storage/innobase/buf/buf0lru.cc
mariadb-DebarunBanerjee: I see your point but we are supposed to use "ib_free_page_sleep" from one connection. Even if we need to set it across multiple connections, we would probably need to use thread local to keep the decision independent. Doing all that seems a bit of over engineering and I would rather like to skip it for future if any such need arises.

mysql-test/suite/innodb/t/insert_into_empty.test
mariadb-DebarunBanerjee: Thanks for pointing it out. I have now made the changes. Please have a look,

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: Indeed you are right. My inference was based on the existing calls to need_LRU_eviction and it seems we never acquire buf_pool.mutex before calling it. I would refrain from modifying existing call flow to acquire stricter mutex as it could introduce performance regression. This is not an issue functionally as the atomic variable try_LRU_scan should save us.

Looks like buf_pool.flush_list_mutex is not necessary and I have removed it and added some code comments. I will also update the MDEV to capture this finding and possible future improvement.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: I think it is best to keep the decision for high level calls like LRU flush to page cleaner main thread. I have now introduced a callback mechanism keeping the loop as it is. Please have a look.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2024,14 +2025,20 @@ ATTRIBUTE_COLD void buf_flush_ahead(lsn_t lsn, bool furious)
...
+static void buf_flush_sync_for_checkpoint(lsn_t lsn, std::function<void()> cbk)
 {
   ut_ad(!srv_read_only_mode);
   mysql_mutex_assert_not_owner(&buf_pool.flush_list_mutex);
 
   for (;;)
   {
+    /* Callback before every iteration. Currently used to check and free LRU
+    pages when buffer pool is running out of free pages. */
+    cbk();
```
dr-m, line 2040: There is only one caller of `buf_flush_sync_for_checkpoint()`, and it is always passing the same function `lru_check` as the callback. That function is not being invoked anywhere else.

To avoid any runtime overhead, I would suggest to define `lru_check` as a `static` function right before this function, so that the compiler will be able to inline the invocation. Its name could be `buf_flush_sync_LRU` or something similar.

I don’t see any way to reuse that function in `buf_flush_page_cleaner()`. The conditions for invoking `buf_flush_LRU()` are slightly different in the case of non-furious flushing. The only thing we might do is to move the mutex release inside `buf_flush_LRU()` and maybe rename that function to `buf_flush_and_release_LRU()` or similar to indicate that the `buf_pool.mutex` will be released.

mariadb-DebarunBanerjee: Isn't the dynamic overhead should be negligible here ? The reason for having the callback is to abstract the LRU flush check etc. from the rest of the flow. If you do think dynamic overhead is more, please re-consider my previous implementation.

dr-m: I am all for abstractions that make code more readable, preferrably zero overhead abstractions. In this case, there seems to be nontrivial overhead (beyond what I would have expected):
```asm
   0x0000000001178b77 <+23>:	cmpq   $0x0,0x10(%r12)
   0x0000000001178b7d <+29>:	jne    0x1178b84 <buf_flush_sync_for_checkpoint(lsn_t, std::function<void()>)+36>
   0x0000000001178b7f <+31>:	call   0x13f6a60 <_ZSt25__throw_bad_function_callv@plt>
   0x0000000001178b84 <+36>:	mov    %r12,%rdi
   0x0000000001178b87 <+39>:	call   *0x18(%r12)
```
In addition to this we would of course have some function preamble and cleanup overhead. The parameter `cbk` also needs to be preserved in the loop.

Because of this, I checked the previous approach 0c209095660542f20a84a9ccceaeed880f58e86c again. I like it better. Basically, it removes the `for` loop from `buf_flush_sync_for_checkpoint()` and adds some code to its only caller.

I would suggest the moving back to more like 0c209095660542f20a84a9ccceaeed880f58e86c:
1. Add the needed code to the start of `buf_flush_sync_for_checkpoint()` and do not add anything else than some debug injection to `buf_flush_page_cleaner()`.
1. Consider declaring the function `buf_flush_sync_for_checkpoint()` not only `ATTRIBUTE_COLD` but also `ATTRIBUTE_NOINLINE`. I do not know why and I did not test other compilers, but `g++-14` would inline the cold function invocation, which was not my intention.

mariadb-DebarunBanerjee: ok, modified the way you suggested.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2025,70 +2025,85 @@ ATTRIBUTE_COLD void buf_flush_ahead(lsn_t lsn, bool furious)
...
   mysql_mutex_assert_not_owner(&buf_pool.flush_list_mutex);
 
-  for (;;)
+  /* During furious flush, we need to keep generating free pages. Otherwise
+  concurrent mtrs could be blocked holding latches for the pages to be flushed
+  causing deadlock in rare occasion.
+
+  Ideally we should be acquiring buffer pool mutex for this check but it is more
+  expensive and we are not using the mutex while calling need_LRU_eviction() as
+  of today. It is a quick and dirty read of the LRU and free list length. Atomic
```
dr-m, line 2040: The line length should be limited to 79 columns, not 80.

dr-m (approved):

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: done, thanks

PR #3299, opened 2024-05-30
MDEV-33616 Fix main.connect2 test on macOS to be stable

DaveGosselin-MariaDB:
This test simulates a test failure during client connect, yet due to differences in setsockopt on macOS versus Linux, the test client will fail to connect and then map that failure to success.  In the case of a failed connection, do not map to success and try again, up to the configured maximum retry attempts.

### DIFF
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -5899,7 +5899,10 @@ int connect_n_handle_errors(struct st_command *command,
       Otherwise, give it some extra time to rule out race-conditions.
       If extra-time doesn't help, we have an unexpected error and
       must abort -- just proceeding to handle_error() when second
-      and third chances are used up will handle that for us.
+      and third chances are used up will handle that for us.  Sometimes
+      we fail because mysql_real_connect returns NULL yet the errno
+      associated with con is zero (such as in a failure to set a
+      socket option within the PVIO plugin): in that case, try again.
 
       There are various user-limits of which only max_user_connections
       and max_connections_per_hour apply at connect time. For the
@@ -5910,7 +5913,8 @@ int connect_n_handle_errors(struct st_command *command,
     */
 
     if (((mysql_errno(con) == ER_TOO_MANY_USER_CONNECTIONS) ||
-         (mysql_errno(con) == ER_USER_LIMIT_REACHED)) &&
+         (mysql_errno(con) == ER_USER_LIMIT_REACHED) ||
+         (mysql_errno(con) == 0)) &&
         (failed_attempts++ < opt_max_connect_retries))
     {
       int i;


vaintroub:
It is not difference in thread scheduling macOS vs Linux. It fails the same on Windows at least, too. There is not less difference in thread scheduling.
The actual error is likely, that client does not fail to connect, even when server closes the socket, and this is not a good thing.
What does client do, then, if it does not fail to connect?

DaveGosselin-MariaDB:
> It is not difference in thread scheduling macOS vs Linux. It fails the same on Windows at least, too. There is quite a difference in thread scheduling between Linux and Windows, too.
> 
> The actual error is, likely, that client does not fail to connect, even when server closes the socket, and this is not a good thing. What does client do, then, if it does _not_ fail to connect?What are the symptoms, that make the test fail? Does client hang? Does it throw a different error?

On macOS, the client doesn't fail to connect but succeeds and the test gets no error (well, error 0).  Are you sure that it's not due to a difference in threading?  As I mentioned, slightly delaying the server allows the failure to manifest at the client, which strongly indicates an issue at the server.  In any case, the failure is completely artificial and in production isn't constrained to occur at this location but at any time; hence why I fixed it in this manner.

vaintroub:
If client does not fail to connect, there is an definitely error on client side. The login did not finish, the server did not send OK packet, which is how every login would end, so why would client indicate a success?

DaveGosselin-MariaDB:
> If client does not fail to connect, there is an definitely error on client side. The login did not finish, the server did not send OK packet, which is how every login would end, so why would client indicate a success? That failure is artificial, but the same or similar error in connect.test is not artificial, and happens not just in a debug setting.

I will investigate and report back.

DaveGosselin-MariaDB:
> If client does not fail to connect, there is an definitely error on client side. The login did not finish, the server did not send OK packet, which is how every login would end, so why would client indicate a success? That failure is artificial, but the same or similar error in connect.test is not artificial, and happens not just in a debug setting.

Hey @vaintroub I debugged from the client side.  I found that: when this test is running on macOS, and during `pvio_socket_change_timeout`, the calls to `setsockopt` sometimes fail with EINVAL and return -1.  This in turn causes `pvio_socket_connect` (called by `ma_pvio_connect` in mariadb_lib.c) to fail and we return NULL to the caller which causes `mysql_errno` (mariadb_lib.c version) to map the errno to 0 (success).  When this test is running on Linux, the calls to `setsockopt` always return 0, success (although sometimes errno is set to 11 but the rc will be 0).

Seems like there could be a couple of issues: (1) `setsockopt` failing unexpectedly on macOS and (2) `connect_n_handle_errors` in mysqltest.cc should expect `mysql_errno(con) == 0` in the case that `mysql_real_connect` fails.

vaintroub:
> > to (1) - the client library actually should not be relying on socket timeouts SO_SNDTIMEO/SO_RECVTIMEO, but it looks like it does.
> 
> Really it's the PVIO plugin. The client relies on the default mariadb connection methods dispatch which in turn calls `mthd_my_real_connect` and that invokes `ma_pvio_connect` directly.
> 

yes, I'm saying that socket "pvio" is implemented incorrectly, be it plugin or not does not matter much.

PR #3300, opened 2024-06-03
MDEV-34283 A misplaced btr_cur_need_opposite_intention() check may fail to prevent hangs

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34283*
## Description
`btr_cur_t::search_leaf()`: Invoke `btr_cur_need_opposite_intention()` after positioning `page_cur.rec` so that the record will be in the intended page. This is something that was broken in f2096478d5750b983f9a9cc4691d20e152dafd4a or de4030e4d49805a7ded5c0bfee01cc3fd7623522 or related changes.
## Release Notes
In extremely rare cases, InnoDB could hang due to a mispredicted page split or merge.
## How can this PR be tested?
`./mtr innodb.temp_table_savepoint` and many other tests cover this.
## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -756,6 +756,7 @@ static bool btr_cur_need_opposite_intention(const buf_page_t &bpage,
                                             ulint compress_limit,
                                             const rec_t *rec)
 {
+  ut_ad(bpage.frame == page_align(rec));
   if (UNIV_LIKELY_NULL(bpage.zip.data) &&
       !page_zip_available(&bpage.zip, is_clust, node_ptr_max_size, 1))
     return true;
@@ -1437,11 +1438,6 @@ release_tree:
           !btr_block_get(*index(), btr_page_get_next(block->page.frame),
                          RW_X_LATCH, false, mtr, &err))
         goto func_exit;
-      if (btr_cur_need_opposite_intention(block->page, index()->is_clust(),
-                                          lock_intention,
-                                          node_ptr_max_size, compress_limit,
-                                          page_cur.rec))
-        goto need_opposite_intention;
     }
 
   reached_latched_leaf:
@@ -1462,6 +1458,13 @@ release_tree:
     ut_ad(up_match != ULINT_UNDEFINED || mode != PAGE_CUR_GE);
     ut_ad(up_match != ULINT_UNDEFINED || mode != PAGE_CUR_LE);
     ut_ad(low_match != ULINT_UNDEFINED || mode != PAGE_CUR_LE);
+
+    if (latch_mode == BTR_MODIFY_TREE &&
+        btr_cur_need_opposite_intention(block->page, index()->is_clust(),
+                                        lock_intention,
+                                        node_ptr_max_size, compress_limit,
+                                        page_cur.rec))
+        goto need_opposite_intention;
 
 #ifdef BTR_CUR_HASH_ADAPT
     /* We do a dirty read of btr_search_enabled here.  We will


mariadb-DebarunBanerjee (commented):
The patch looks straight forward. Can you please see if a testcase/scenario can be created that can validate the patch ? Other than validation, it would also help us map any customer issue that is possibly caused by it.

dr-m:
> The patch looks straight forward. Can you please see if a testcase/scenario can be created that can validate the patch ? Other than validation, it would also help us map any customer issue that is possibly caused by it.

What kind of testing are you looking for? Several `mtr` tests will fail when the debug assertion is added but the fix is omitted. The description mentions one of them: `./mtr innodb.temp_table_savepoint`.

storage/innobase/btr/btr0cur.cc
```diff
@@ -756,6 +756,7 @@ static bool btr_cur_need_opposite_intention(const buf_page_t &bpage,
                                             ulint compress_limit,
                                             const rec_t *rec)
 {
+  ut_ad(bpage.frame == page_align(rec));
   if (UNIV_LIKELY_NULL(bpage.zip.data) &&
```
mariadb-DebarunBanerjee, line 760: > What kind of testing are you looking for? Several mtr tests will fail when the debug assertion is added but the fix is omitted. The description mentions one of them: ./mtr innodb.temp_table_savepoint.
1. The bug is tagged as hang. Can we have some scenario which can produce a hang ?
2. I guess you mean the assert above, Let me check it out.

mariadb-DebarunBanerjee: Ok, The added assert fails. The block was misplaced and the removal looks correct. Looks like page_cur.rec is not in block->page at this point. Let me check bit more on this.

storage/innobase/btr/btr0cur.cc
```diff
@@ -1463,6 +1459,13 @@ dberr_t btr_cur_t::search_leaf(const dtuple_t *tuple, page_cur_mode_t mode,
     ut_ad(up_match != ULINT_UNDEFINED || mode != PAGE_CUR_LE);
     ut_ad(low_match != ULINT_UNDEFINED || mode != PAGE_CUR_LE);
 
+    if (latch_mode == BTR_MODIFY_TREE &&
+        btr_cur_need_opposite_intention(block->page, index()->is_clust(),
+                                        lock_intention,
+                                        node_ptr_max_size, compress_limit,
+                                        page_cur.rec))
+        goto need_opposite_intention;
```
mariadb-DebarunBanerjee, line 1467: An interesting point to check should be that the code is working as expected in the newly moved position. It is likely there is not test to validate this functionality too. Otherwise we should already have failures.

storage/innobase/btr/btr0cur.cc
dr-m: I don’t think that the hang would be practical to reproduce. The closest that we have gotten to that is the assertion failure that was reported in [MDEV-31815](https://jira.mariadb.org/browse/MDEV-31815). Because I do not know if it is exactly due to this bug, I filed this bug separately. The purpose of that assertion is to catch potential hangs when the [WL#6326](https://dev.mysql.com/worklog/task/?id=6326) rules are being violated. That is, the index pages would be accessed in the wrong order while not holding an exclusive latch on the index tree.

Yes, the added assertion would fail quite easily if we omitted the fix from the caller.

storage/innobase/btr/btr0cur.cc
dr-m: The `goto need_opposite_intention` is something that will escalate to exclusively acquiring `index->lock`, via `btr_cur_t::pessimistic_search_leaf()`. If we invoked that code path unnecessarily, there would be a performance regression due to increased contention on `index->lock`.

The cursor should be positioned correctly even if this fix is not present, but a later operation in the same mini-transaction could hang because it could violate the [WL#6326](https://dev.mysql.com/worklog/task/?id=6326) rules. The rules are supposed to be enforced here. Further code is only trying to catch misuse in some debug assertions (checking that we are holding an exclusive `index->lock` if we have to acquire new page latches up or back in the tree).

storage/innobase/btr/btr0cur.cc
mariadb-DebarunBanerjee: The removal is correct.

storage/innobase/btr/btr0cur.cc
mariadb-DebarunBanerjee: yes, I understand it could be difficult to create a scenario. Perhaps I will give it a try later. The check looks correct to me in theory.
> because it could violate the [WL#6326](https://dev.mysql.com/worklog/task/?id=6326) rules.

Right. The WL is the best that we probably have for the rules. I remember there was some addition/correction needed there last time I was reviewing one bug with Yasufumi.

mariadb-DebarunBanerjee (approved):
Looks correct! Thanks for identifying and resolving it.

PR #3303, opened 2024-06-04
MDEV-34261: Detect if build is running under 32-bit container

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-34261*

## Description
When building on 64-bit kernel machine in 32-bit docker container CMake falsely (but it works as expected) detects that  container runtime in also 64-bits. Use linux32 command to change runtime environment to 32-bit and then CMake will  correctly disable for example ColumnStore and not try to build it

This commit only works with debian/autobake-debs.sh

## Release Notes
Should add nothing as this brings back i386 building in 11.5 and above

## How can this PR be tested?
This is bit difficult to test as problem start to rise after 11.5. They have `__int128` which is not supported and probably won't be supported in i386 environment. But basics is that have i386 container and launch it with `docker run -it quay.io/mariadb-foundation/bb-worker:debian12-386 bash`. Then try to compile with `debian/autobake-debs.sh`

## PR quality check

### DIFF
--- a/debian/autobake-deb.sh
+++ b/debian/autobake-deb.sh
@@ -62,6 +62,7 @@ replace_uring_with_aio()
 }
 
 architecture=$(dpkg-architecture -q DEB_BUILD_ARCH)
+uname_machine=$(uname -m)
 
 # Parse release name and number from Linux standard base release
 # Example:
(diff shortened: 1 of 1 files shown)


grooverdan (approved):
lgtm

LinuxJedi (approved):

PR #3305, opened 2024-06-04
MDEV-34297 get_rnd_value() of ib_counter_t is unnecessarily complex

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34297*
## Description
The shared counter template `ib_counter_t` uses the function `my_timer_cycles()` as a source of pseudo-random numbers to pick a shard. On some platforms, `my_timer_cycles()` could return the constant value 0.

`get_rnd_value()`: Remove.

`my_pseudo_random()`: Implement as an alias of `my_timer_cycles()` or a wrapper for `pthread_self()`.
## Release Notes
This is a minor performance improvement, not worth mentioning.
## How can this PR be tested?
For example, disassemble the function `sync_array_print_innodb` and observe that a call to `pthread_self` is present prior to this fix and absent after this fix (unless you happen to compile for a rare operating system and ISA).
## PR quality check

### DIFF
--- a/include/my_rdtsc.h
+++ b/include/my_rdtsc.h
@@ -177,9 +177,22 @@ static inline ulonglong my_timer_cycles(void)
   /* gethrtime may appear as either cycle or nanosecond counter */
   return (ulonglong) gethrtime();
 #else
+# define MY_TIMER_CYCLES_IS_ZERO
   return 0;
 #endif
 }
+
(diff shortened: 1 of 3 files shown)


vaintroub (approved):
Ok for me.

As noted, pthread_self() workaround won't work if pthread_t is not numeric, and a function that returns a numeric  thread id would be needed instead. Just a remark, to be fixed only when the need arises.

PR #3306, opened 2024-06-04
MDEV-34296 extern thread_local is a CPU waste

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34296*
## Description
In commit 99bd22605938c42d876194f2ec75b32e658f00f5 (MDEV-31558) we wrongly thought that there would be minimal overhead for accessing a thread-local variable `mariadb_stats`.

It turns out that in C++11, each access to an `extern thread_local` variable requires conditionally invoking an initialization function. In fact, we do have a custom initializer of this very variable, and that call is actually unavoidable.

In C++20, one could declare `constinit thread_local` variables, but the address of a `thread_local` variable (`&mariadb_dummy_stats`) is not a compile-time constant. We did not want to declare `mariadb_dummy_stats` without `thread_local`, because then the dummy accesses could lead to cache line contention between threads.

`mariadb_stats`: Declare as `__thread` or `__declspec(thread)` so that there will be no dynamic initialization, but zero-initialization.

`mariadb_dummy_stats`: Remove. It is lesser evil to let the environment perform zero-initialization and check if `!mariadb_stats`.
## Release Notes
This is a relatively minor performance improvement, maybe not worth mentioning specially.
## How can this PR be tested?
```sh
./mtr main.analyze_engine_stats2 main.analyze_engine_stats main.rowid_filter_innodb
```
## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2165,6 +2165,12 @@ void buf_page_free(fil_space_t *space, uint32_t page, mtr_t *mtr)
   mtr->memo_push(block, MTR_MEMO_PAGE_X_MODIFY);
 }
 
+static void buf_inc_get(ha_handler_stats *stats)
+{
+  mariadb_increment_pages_accessed(stats);
+  ++buf_pool.stat.n_page_gets;
+}
+
 /** Get read access to a compressed page (usually of type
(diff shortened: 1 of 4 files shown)


spetrunia (approved):
I have to trust your statement that compilers guarantee that `__thread` or `__declspec(thread)` are zero-initialized.

Please add a comment about this above the declaration of `mariadb_stats`. 

Other than that, looks good to me.

PR #3307, opened 2024-06-04
MDEV-21166 Add Mroonga initialized check to Mroonga UDFs

kou:
- [x] *The Jira issue number for this PR is: MDEV-21166*

## Description

If Mroonga UDFs such as `last_insert_id()` is used without `INSTALL PLUGIN mroonga`, MariaDB is crashed. See MDEV-21166 for unexpected behavior details.

## Release Notes

I think that we don't need to add this to the release notes because this is unexpected usage. Mroonga UDFs are available only with `INSTALL PLUGIN mroonga`. This changes just add validations for unexpected usage.

## How can this PR be tested?

1. Execute `CREATE FUNCTION last_insert_grn_id RETURNS INTEGER SONAME 'ha_mroonga.so'` without `INSTALL PLUGIN mroonga`
2. Execute `SELECT last_insert_grn_id();`
3. It returns an error (`ERROR 1123 (HY000): Can't initialize function 'last_insert_grn_id'; last_insert_grn_id(): Mroonga isn't initialized`) without crash

We can't add `.test` for this because we enable Mroonga for all tests under `storage/mroonga/mysql-test/mroonga/storage/`: https://github.com/MariaDB/server/blob/3d4bdf76d35611f1aed066824fb56988c5ba16e9/storage/mroonga/mysql-test/mroonga/storage/suite.opt#L1

## PR quality check


LinuxJedi (commented):
Many thanks for this contribution, in-general I think it looks good, but it should be against 10.5 instead of 11.6. This is because bug fixes are merged up through versions.

As for testing, whilst it is true there is a suite wide option for "storage" and "wrapper", there is no reason why another directory / suite could not be created with this as the first test in it.

storage/mroonga/ha_mroonga.cpp
```diff
@@ -296,6 +297,7 @@ static PSI_mutex_info mrn_mutexes[] =
 #endif
 
 /* global variables */
+bool mrn_initialized = false;
```
HugoWenTD, line 300: Is this safe for multi-threading or CPU optimization? Is there a need to add mrn_mutexes similar to other global variables?

kou: Yes. It's safe for multi-threading. This is changed only by `st_mysql_plugin::init` and `st_mysql_plugin::deinit` and they aren't called with multiple threads at once.

I'm not sure for CPU optimization. What situation do you think about?

kou:
> it should be against 10.5 instead of 11.6.

I changed the base branch to 10.5 from 11.6 on GitHub. It causes many commits to this PR... I close this and reopen a new PR...

kou:
https://github.com/MariaDB/server/pull/3329

PR #3312, opened 2024-06-06
MDEV-34321: UBSAN runtime error: call to function crc32c_3way through…

bnestere:
MDEV-34321: UBSAN runtime error: call to function crc32c_3way through pointer to incorrect function type

my_crc32_t takes function signature:

```
unsigned int (*)(unsigned int, const void *, unsigned long)
```

yet the crc32c_3way derivation has a definition with a second parameter of a different type:

```
uint32_t crc32c_3way(uint32_t crc, const char *buf, size_t len)
```

which leads to UBSAN errors. This patch creates a type-safe wrapper to call crc32c_3way.

Assigning to @dr-m because you authored the original MDEV-24745. 

I also considered changing the definition of crc32c_3way() in crc32c_amd64.c, yet that feels more fragile (it seems to be pulled in from a third party).

### DIFF
--- a/mysys/crc32/crc32c_amd64.cc
+++ b/mysys/crc32/crc32c_amd64.cc
@@ -184,9 +184,9 @@ static inline uint64_t CombineCRC(
 // Compute CRC-32C using the Intel hardware instruction.
 extern "C"
 USE_PCLMUL
-uint32_t crc32c_3way(uint32_t crc, const char *buf, size_t len)
+uint32_t crc32c_3way(uint32_t crc, const void *buf, size_t len)
 {
-  const unsigned char* next = (const unsigned char*)buf;
+  const unsigned char* next = static_cast<const unsigned char*>(buf);
   uint64_t count;
   uint64_t crc0, crc1, crc2;
   crc0 = crc ^ 0xffffffffu;
--- a/mysys/crc32/crc32c.cc
+++ b/mysys/crc32/crc32c.cc
@@ -526,7 +526,7 @@ extern "C" const char *my_crc32c_implementation()
 }  // namespace crc32c
 }  // namespace mysys_namespace
 
-extern "C" unsigned my_crc32c(unsigned int crc, const char *buf, size_t size)
+extern "C" uint32 my_crc32c(uint32 crc, const void *buf, size_t size)
 {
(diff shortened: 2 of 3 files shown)


dr-m:
af5e18a121465632a6d442e2811b783eaad8344c would seem to introduce runtime overhead in the form of an extra function call. Even if were a simple tail call that is translated to a jump, it is not free.

Based on my testing with Clang 18, the minimum that needs to be done to silence the error is to change the `const char*` parameter of `crc32c_3way` to be of type `const void*`. It will be cast to `const unsigned char*` anyway.

I did this (as well as some unification of the 32-bit CRC parameters and return values) in 
73225b624b863f7e463c32b542280fe7e64c26fc. Sure, neither LLVM nor any of our thinkable target platforms (mainly LP32, LP64, LLP64) cannot distinguish `uint32`, `uint32_t` and `unsigned`, but we could do a slightly better job there.

mysys/crc32/crc32c_amd64.cc
```diff
@@ -184,9 +184,9 @@ static inline uint64_t CombineCRC(
 // Compute CRC-32C using the Intel hardware instruction.
 extern "C"
 USE_PCLMUL
-uint32_t crc32c_3way(uint32_t crc, const char *buf, size_t len)
+uint32_t crc32c_3way(uint32_t crc, const void *buf, size_t len)
 {
-  const unsigned char* next = (const unsigned char*)buf;
+  const unsigned char* next = static_cast<const unsigned char*>(buf);
```
vaintroub, line 189: I do  not think changing this line is necessary. It is , for the most part, not our code anyway, so l'd leave it as-is (it is also shorter, which is not a bad thing.

vaintroub (approved):
Ok, but suggest to remove the unnecessary   C++ cast change.

dr-m:
> Ok, but suggest to remove the unnecessary C++ cast change.

Thank you, but I’d retain it, for reasons mentioned here: https://stackoverflow.com/questions/32168/c-cast-syntax-styles

Curiously, https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rcpl-C does not explicitly say anything about C-style casts.

PR #3315, opened 2024-06-07
MDEV-31658 : Deadlock found when trying to get lock; try restarting t…

janlindstrom:
…ransaction, Error_code: 1213; handler error HA_ERR_LOCK_DEADLOCK; the event's master log FIRST, end_log_pos xxx, Internal MariaDB error code: 1213

- [x] *The Jira issue number for this PR is: MDEV-31658*

## Description
Problem was that there was two non-conflicting local idle transactions in node_1 that both inserted a key to primary key. Then two transactions from other nodes inserted also a key to primary key so that insert from node_2 conflicted one of the local transactions in node_1 so that there would be duplicate key if both are committed. For this insert from other node tries to acquire S-lock for this record and because this insert is high priority brute force (BF) transaction it will kill idle local transaction.

Concurrently, second insert from node_3 conflicts the second idle insert transaction in node_1. Again, it tries to acquire S-lock for this record and kills idle local transaction.

At this point we have two non-conflicting high priority transactions holding S-lock on different records in node_1. For example like this: rec s-lock-node2-rec s-lock-node3-rec rec.

Because these high priority BF-transactions do not wait each other insert from node3 that has later seqno compared to insert from node2 can continue. It will try to acquire insert intention for record it tries to insert (to avoid duplicate key to be inserted by local transaction). Hower, it will note that there is conflicting S-lock in same gap between records. This will lead deadlock error as we have defined that BF-transactions may not wait for record lock but we can't kill conflicting BF-transaction because it has lower seqno and it should commit first.

BF-transactions are executed concurrently because their values to primary key are different i.e. they do not conflict.

Galera certification will make sure that inserts from other nodes i.e these high priority BF-transactions can't insert duplicate keys. Local transactions naturally can but they will be killed when BF-transaction
acquires required record locks.

Therefore, we can allow situation where there is conflicting S-lock and insert intention lock regardless of their seqno order and let both continue with no wait. This will lead to situation where we need to allow BF-transaction to wait when lock_rec_has_to_wait_in_queue is called because this function is also called from
lock_rec_queue_validate and because lock is waiting there would be assertion in ut_a(lock->is_gap()
|| lock_rec_has_to_wait_in_queue(cell, lock));

lock_wait_wsrep_kill
  Add debug sync points for BF-transactions killing
  local transaction.

wsrep_assert_no_bf_bf_wait
  Print also requested lock information

lock_rec_has_to_wait
  Reformat function as we make bigger changes. Add
  clear conditions when wsrep transaction does not
  need to wait.

lock_rec_has_to_wait_in_queue
  Remove wsrep exception, in this function all
  conflicting locks need to wait in queue.
  Conflicts between BF and local transactions
  are handled in lock_wait.

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/sql/service_wsrep.cc
+++ b/sql/service_wsrep.cc
@@ -249,18 +249,23 @@ extern "C" my_bool wsrep_thd_skip_locking(const THD *thd)
 
 extern "C" my_bool wsrep_thd_order_before(const THD *left, const THD *right)
 {
-  if (wsrep_thd_is_BF(left, false) &&
-      wsrep_thd_is_BF(right, false) &&
-      wsrep_thd_trx_seqno(left) < wsrep_thd_trx_seqno(right)) {
-    WSREP_DEBUG("BF conflict, order: %lld %lld\n",
-                (long long)wsrep_thd_trx_seqno(left),
-                (long long)wsrep_thd_trx_seqno(right));
-    return TRUE;
-  }
-  WSREP_DEBUG("waiting for BF, trx order: %lld %lld\n",
-              (long long)wsrep_thd_trx_seqno(left),
-              (long long)wsrep_thd_trx_seqno(right));
-  return FALSE;
+  my_bool before= (wsrep_thd_is_BF(left, false) &&
+                   wsrep_thd_is_BF(right, false) &&
+                   wsrep_thd_trx_seqno(left) < wsrep_thd_trx_seqno(right));
+
+  WSREP_DEBUG("wsrep_thd_order_before: %s thread=%llu seqno=%llu query=%s "
+              "%s %s thread=%llu, seqno=%llu query=%s",
+              (wsrep_thd_is_BF(left, false) ? "BF" : "def"),
+              thd_get_thread_id(left),
+              wsrep_thd_trx_seqno(left),
+              wsrep_thd_query(left),
+              (before ? " TRUE " : " FALSE "),
+              (wsrep_thd_is_BF(right, false) ? "BF" : "def"),
+              thd_get_thread_id(right),
+              wsrep_thd_trx_seqno(right),
+              wsrep_thd_query(right));
+
+  return before;
 }
 
 /** Check if wsrep transaction is aborting state.
--- a/storage/innobase/lock/lock0lock.cc
+++ b/storage/innobase/lock/lock0lock.cc
@@ -503,8 +503,10 @@ this BF-BF wait correct and if not report BF wait and assert.
 
 @param[in]	lock_rec	other waiting record lock
 @param[in]	trx		trx requesting conflicting record lock
+@param[in]	type_mode	lock type mode of requesting trx
 */
-static void wsrep_assert_no_bf_bf_wait(const lock_t *lock, const trx_t *trx)
+static void wsrep_assert_no_bf_bf_wait(const lock_t *lock, const trx_t *trx,
+                                       const unsigned type_mode = LOCK_NONE)
 {
 	ut_ad(!lock->is_table());
 	lock_sys.assert_locked(*lock);
@@ -546,6 +548,15 @@ static void wsrep_assert_no_bf_bf_wait(const lock_t *lock, const trx_t *trx)
 		return;
 	}
 
+	if (type_mode != LOCK_NONE)
+		ib::error() << " Requested lock "
+			    << ((type_mode & LOCK_TABLE) ? "on table " : " on record ")
+			    << ((type_mode & LOCK_WAIT) ? " WAIT " : " ")
+			    << ((type_mode & LOCK_GAP) ? " GAP " : " ")
+			    << ((type_mode & LOCK_REC_NOT_GAP) ? " RECORD " : " ")
+			    << ((type_mode & LOCK_INSERT_INTENTION) ? " INSERT INTENTION " : " ")
+			    << ((type_mode & LOCK_X) ? " LOCK_X " : " LOCK_S ");
+
 	mtr_t mtr;
 
 	ib::error() << "Conflicting lock on table: "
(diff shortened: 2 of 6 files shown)


storage/innobase/lock/lock0lock.cc
```diff
@@ -546,6 +548,17 @@ static void wsrep_assert_no_bf_bf_wait(const lock_t *lock, const trx_t *trx)
...
+	if (type_mode)
+	{
+		ib::error() << " Requested lock "
+			    << ((type_mode & LOCK_TABLE) ? "on table " : " on record ")
+			    << ((type_mode & LOCK_WAIT) ? " WAIT " : " ")
+			    << ((type_mode & LOCK_GAP) ? " GAP " : " ")
+			    << ((type_mode & LOCK_REC_NOT_GAP) ? " RECORD " : " ")
+			    << ((type_mode & LOCK_INSERT_INTENTION) ? " INSERT INTENTION " : " ")
+			    << ((type_mode & LOCK_X) ? " LOCK_X " : " LOCK_S ");
+	}
```
dr-m, line 560: This introduces an excessive number of calls to `operator<<()` functions. Why not just simply write the following:
```c++
	sql_print_error("InnoDB: Requested lock %03x", type_mode);
```
If someone is interested in particular values, 4-digit hexadecimal values should be much easier to filter (for example with regular expressions) than any such partially "decoded" output.

What is the check for `type_mode == 0` good for? It would seem to be an almost valid value that would correspond to an intention-shared record lock. (To my knowledge, intention locks are only possible on tables, not index records.)

storage/innobase/lock/lock0lock.cc
```diff
@@ -584,141 +597,177 @@ ATTRIBUTE_NOINLINE static bool wsrep_is_BF_lock_timeout(const trx_t &trx)
...
+bool lock_rec_has_to_wait(const trx_t *trx, unsigned type_mode,
+                          const lock_t* lock2, bool lock_is_on_supremum)
+{
+  ut_ad(trx);
+  ut_ad(!lock2->is_table());
+  ut_d(lock_sys.hash_get(type_mode).assert_locked(
+         lock2->un_member.rec_lock.page_id));
+
+  if (trx == lock2->trx
+      || lock_mode_compatible(
```
dr-m, line 625: If you are reformatting a function to follow the common formatting rules, then you should consistently follow those rules:
* Never break a line before any binary operator such as `||`.
* Never break a line after `(`.

storage/innobase/lock/lock0lock.cc
```diff
@@ -937,7 +986,7 @@ void wsrep_report_error(const lock_t* victim_lock, const trx_t *bf_trx)
   lock_rec_print(stderr, bf_trx->lock.wait_lock, mtr);
   WSREP_ERROR("victim holding lock: ");
   lock_rec_print(stderr, victim_lock, mtr);
-  wsrep_assert_no_bf_bf_wait(victim_lock, bf_trx);
+  wsrep_assert_no_bf_bf_wait(victim_lock, bf_trx, 0);
```
dr-m, line 989: The value `0` would seem to correspond to a intention-shared record lock. This is debug code, and such locks should be impossible. In any case, it would be clearer to use the value `LOCK_NONE` here; in fact, you could declare that as a default value of the added function parameter, so that you would not have to change this invocation.

storage/innobase/lock/lock0lock.cc
```diff
@@ -584,141 +597,177 @@ ATTRIBUTE_NOINLINE static bool wsrep_is_BF_lock_timeout(const trx_t &trx)
...
-	(brute force). If conflicting transaction is also wsrep high
-	priority transaction we should avoid lock conflict because
-	ordering of these transactions is already decided and
-	conflicting transaction will be later replayed. */
-	if (trx->is_wsrep_UK_scan()
-	    && wsrep_thd_is_BF(lock2->trx->mysql_thd, false)) {
-		return false;
-	}
+  if (trx->is_wsrep())
+  {
```
dr-m, line 714: It would be better to avoid reformatting this entire function when the essential change is here. I would suggest to add a separate `static ATTRIBUTE_NOINLINE ATTRIBUTE_COLD` function for the Galera specific code, and invoke it at the end of the function:
```c++
#ifdef WITH_WSREP
	if (trx->is_wsrep()) {
		return lock_rec_has_to_wait_wsrep(…);
	}
#endif
	return true;
```

dr-m:
It would be good to properly fill in the pull request template, also paying attention to the formatting. The first line of the commit message is way too long. The output of `git log --oneline` should be short, definitely not 227 characters per line.

storage/innobase/lock/lock0lock.cc
janlindstrom: I do not agree on hexadecimal values, I prefer clear strings. type_mode = 0 is case when actual requested lock type is not known but your suggestion LOCK_NONE is good.

storage/innobase/lock/lock0lock.cc
janlindstrom: No it is not that. Here we do not know what lock type was requested, only conflicting lock is known.

storage/innobase/lock/lock0lock.cc
janlindstrom: Okay

sql/service_wsrep.cc
```diff
@@ -249,18 +249,26 @@ extern "C" my_bool wsrep_thd_skip_locking(const THD *thd)
 
 extern "C" my_bool wsrep_thd_order_before(const THD *left, const THD *right)
 {
+  my_bool before=FALSE;
+
   if (wsrep_thd_is_BF(left, false) &&
       wsrep_thd_is_BF(right, false) &&
-      wsrep_thd_trx_seqno(left) < wsrep_thd_trx_seqno(right)) {
```
dr-m, line 254: This could be written more concisely as
```c++
  bool before= (wsrep_thd_is_BF(left, false) &&
      wsrep_thd_is_BF(right, false) &&
      wsrep_thd_trx_seqno(left) < wsrep_thd_trx_seqno(right);
```

storage/innobase/lock/lock0lock.cc
```diff
@@ -582,6 +593,80 @@ ATTRIBUTE_NOINLINE static bool wsrep_is_BF_lock_timeout(const trx_t &trx)
...
+ATTRIBUTE_NOINLINE ATTRIBUTE_COLD
+bool lock_rec_has_to_wait_wsrep(const trx_t* trx,
+                                const unsigned type_mode,
+                                const lock_t* lock2)
+{
+  const trx_t* trx2= lock2->trx;
+
+  if (trx->is_wsrep_UK_scan() &&
+      wsrep_thd_is_BF(trx2->mysql_thd, false))
+  {
```
dr-m, line 619: Please omit `{` when the body consists of a single statement.

storage/innobase/lock/lock0lock.cc
```diff
@@ -691,31 +776,10 @@ lock_rec_has_to_wait(
...
-	if (trx->is_wsrep() &&
-	   (type_mode & LOCK_MODE_MASK) == LOCK_X &&
-	   (lock2->type_mode & LOCK_MODE_MASK) == LOCK_X &&
-	   wsrep_thd_order_before(trx->mysql_thd,
-				  lock2->trx->mysql_thd)) {
-		return false;
+	if (trx->is_wsrep()) {
+	  /* We have some additional rules for wsrep transactions. */
+	  return (lock_rec_has_to_wait_wsrep(trx, type_mode, lock2));
 	}
```
dr-m, line 782: This function is formatted using the archaic InnoDB formatting rules. Hence we should use TAB instead of spaces for indentation. The comment does not add much value; this should be obvious from the name of the tail-called function. The parentheses after `return` are redundant and should better be omitted.

storage/innobase/lock/lock0lock.cc
```diff
@@ -582,6 +593,80 @@ ATTRIBUTE_NOINLINE static bool wsrep_is_BF_lock_timeout(const trx_t &trx)
...
+                            set on the same record as in the new
+	                    lock we are setting.
+@return TRUE if new lock has to wait for lock2 to be removed */
+
+ATTRIBUTE_NOINLINE ATTRIBUTE_COLD
+bool lock_rec_has_to_wait_wsrep(const trx_t* trx,
+                                const unsigned type_mode,
+                                const lock_t* lock2)
+{
+  const trx_t* trx2= lock2->trx;
```
dr-m, line 615: In the common formatting rules, there should be a space before but not after `*`.

dr-m (approved):
I didn’t fully review the logic of the Galera specific changes. In InnoDB, this is improving the encapsulation of some Galera related logic. One of my comments about `wsrep_assert_no_bf_bf_wait()` was disregarded. Because that function is only present in debug builds, I can live with that.

sql/service_wsrep.cc
janlindstrom: Fixed.

storage/innobase/lock/lock0lock.cc
janlindstrom: I keep this because it is easier to read as it is (long comment).

storage/innobase/lock/lock0lock.cc
janlindstrom: Fixed.

storage/innobase/lock/lock0lock.cc
janlindstrom: Fixed.

sysprg:
Thanks, fix merged with head revision: https://github.com/MariaDB/server/commit/ee974ca5e0f00e928873179c5dfc2f040ab5aefd

PR #3317, opened 2024-06-07
MDEV-34178 Performance regression with IO-bound insert benchmark

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34178*
## Description
Performance testing conducted by @mdcallag identified multiple scalability bottlenecks around the spin loop logic for `buf_page_t::lock`. It also suggests that we might better enable spin loops also for `dict_index_t::lock`.
## Release Notes
Performance of I/O bound workloads was improved.
## How can this PR be tested?
I think that this is best tested by @mdcallag. Enabling the spin loops for `index_lock` (79039bdc6f028c86355f0045c70dc3b7c97e05aa) may be turn out to be unnecessary in the advent of the other changes.

This is low-level code, and it is in part covered by by the `innodb_sync-t` unit test. I ran it (as well as the entire regression test suite) for both `PLUGIN_PERFSCHEMA=NO`  and `PLUGIN_PERFSCHEMA=STATIC`.
## PR quality check

### DIFF
--- a/storage/innobase/include/srw_lock.h
+++ b/storage/innobase/include/srw_lock.h
@@ -223,12 +223,7 @@ public:
 
   bool u_lock_try()
   {
-    if (!writer.wr_lock_try())
-      return false;
-    IF_DBUG_ASSERT(uint32_t lk=,)
-    readers.fetch_add(1, std::memory_order_acquire);
-    DBUG_ASSERT(lk < WRITER - 1);
-    return true;
+    return writer.wr_lock_try();
   }
 
   bool wr_lock_try()
@@ -248,9 +243,6 @@ public:
   void u_lock()
   {
     writer.wr_lock();
-    IF_DBUG_ASSERT(uint32_t lk=,)
-    readers.fetch_add(1, std::memory_order_acquire);
-    DBUG_ASSERT(lk < WRITER - 1);
   }
   void wr_lock()
   {
@@ -272,15 +264,15 @@ public:
   void u_wr_upgrade()
   {
     DBUG_ASSERT(writer.is_locked());
-    uint32_t lk= readers.fetch_add(WRITER - 1, std::memory_order_acquire);
-    if (lk != 1)
-      wr_wait(lk - 1);
+    uint32_t lk= readers.fetch_add(WRITER, std::memory_order_acquire);
+    if (lk)
+      wr_wait(lk);
   }
   void wr_u_downgrade()
   {
     DBUG_ASSERT(writer.is_locked());
     DBUG_ASSERT(is_write_locked());
-    readers.store(1, std::memory_order_release);
+    readers.store(0, std::memory_order_release);
     /* Note: Any pending rd_lock() will not be woken up until u_unlock() */
   }
 
@@ -293,10 +285,6 @@ public:
   }
   void u_unlock()
   {
-    IF_DBUG_ASSERT(uint32_t lk=,)
-    readers.fetch_sub(1, std::memory_order_release);
-    DBUG_ASSERT(lk);
-    DBUG_ASSERT(lk < WRITER);
     writer.wr_unlock();
   }
   void wr_unlock()
--- a/storage/innobase/include/srw_lock.h
+++ b/storage/innobase/include/srw_lock.h
@@ -402,7 +402,7 @@ typedef srw_spin_lock_low srw_spin_lock;
 class ssux_lock
 {
   PSI_rwlock *pfs_psi;
-  ssux_lock_impl<false> lock;
+  ssux_lock_impl<true> lock;
 
   ATTRIBUTE_NOINLINE void psi_rd_lock(const char *file, unsigned line);
   ATTRIBUTE_NOINLINE void psi_wr_lock(const char *file, unsigned line);
--- a/storage/innobase/include/srw_lock.h
+++ b/storage/innobase/include/srw_lock.h
@@ -72,10 +72,13 @@ inline void pthread_mutex_wrapper<true>::wr_lock()
 # endif
 #endif
 
+template<bool spinloop> class ssux_lock_impl;
+
 /** Futex-based mutex */
 template<bool spinloop>
 class srw_mutex_impl final
 {
+  friend ssux_lock_impl<spinloop>;
   /** The lock word, containing HOLDER + 1 if the lock is being held,
   plus the number of waiters */
   std::atomic<uint32_t> lock;
@@ -97,6 +100,8 @@ private:
   inline void wait(uint32_t lk);
   /** Wake up one wait() thread */
   void wake();
+  /** Wake up all wait() threads */
+  inline void wake_all();
 public:
   /** @return whether the mutex is being held or waited for */
   bool is_locked_or_waiting() const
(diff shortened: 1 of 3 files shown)


storage/innobase/sync/srw_lock.cc
```diff
@@ -392,14 +391,22 @@ template void ssux_lock_impl<false>::wr_wait(uint32_t);
...
-      break;
+    const unsigned delay= srw_pause_delay();
+
+    for (auto spin= srv_n_spin_wait_rounds; spin; spin--)
+    {
+      srw_pause(delay);
+      if (rd_lock_try())
+        return;
+    }
   }
```
dr-m, line 404: We might end the `if (spinloop)` body in `this->rd_wait<false>(); return;` to avoid some code duplication.

storage/innobase/sync/srw_lock.cc
```diff
@@ -392,14 +391,22 @@ template void ssux_lock_impl<false>::wr_wait(uint32_t);
...
+
+  writer.wr_lock();
+  ut_d(uint32_t lk=) readers.fetch_add(1, std::memory_order_relaxed);
+  writer.wr_unlock();
+  ut_ad(!(lk & WRITER));
```
dr-m, line 409: We could try adding a ‘pre-filtering’ step before these lines, to avoid unnecessarily locking `writer`:
```c++
#ifndef SUX_LOCK_GENERIC
  while (writer.lock.load(std::memory_order_acquire) & HOLDER)
  {
    writer.wait(lk);
    if (rd_lock_try())
      return;
  }
  if (rd_lock_try())
    return;
#endif
```

dr-m (commented):
I got some further thoughts on the `rd_wait()`.

storage/innobase/sync/srw_lock.cc
dr-m: This is not so simple because the template parameter is for the type and no for the member function. Besides, with a change like 79039bdc6f028c86355f0045c70dc3b7c97e05aa no `spinloop=false` variant would need to be instantiated.

storage/innobase/sync/srw_lock.cc
dr-m: It is not this simple. Before invoking `wait()` we would have to inform the holder that we are waiting, so that it would know to wake us once it is releasing the exclusive lock. 9430e397e0e83697d7b987ff8dfc8348b8dcf707 implements this, and it actually also works with `SUX_LOCK_GENERIC`.

dr-m: In 9430e397e0e83697d7b987ff8dfc8348b8dcf707 the `writer.wr_lock()` seems to be virtually unreachable during tests that I conducted. As far as I can tell, it is reachable with a tiny probability. We can remove this unnecessary contention point by implementing an infinite loop, even if the `HOLDER` flag on `writer.lock` is intermittently clear.

storage/innobase/sync/srw_lock.cc
```diff
@@ -392,14 +402,55 @@ template void ssux_lock_impl<false>::wr_wait(uint32_t);
...
+    bool acquired;
+
+    while (writer.HOLDER & writer_lock)
+    {
+      writer.wait(writer_lock);
+      acquired= rd_lock_try();
+      if (acquired)
+        goto acquired;
+      writer_lock = writer.lock.load(std::memory_order_acquire);
+    }
```
dr-m, line 432: If `rd_lock_try()` returned `false` because the `readers` lock word value was exactly `WRITER`, we must invoke `writer.wake()` to pass the `writer.wait()` signal that we had consumed, to a concurrent `wr_lock()` that just succeeded.

storage/innobase/include/srw_lock.h
```diff
@@ -223,12 +228,7 @@ class ssux_lock_impl
...
-    if (!writer.wr_lock_try())
-      return false;
-    IF_DBUG_ASSERT(uint32_t lk=,)
-    readers.fetch_add(1, std::memory_order_acquire);
-    DBUG_ASSERT(lk < WRITER - 1);
-    return true;
+    return writer.wr_lock_try();
   }
 
   bool wr_lock_try()
```
dr-m, line 234: `ssux_lock_impl::wr_lock_try()` had better be refactored so that we can avoid unnecessary calls to `writer.wr_unlock()` as part of blocking operations such as `dict_sys_t::lock()`.

dr-m (commented):
I will revise this further.

storage/innobase/include/srw_lock.h
dr-m: To check this, I renamed various definitions of `wr_lock_try()` and checked every caller in the chain. We can improve `psi_wr_lock()` so that if only the `lock.writer.wr_lock_try()` succeeds but some shared locks in `lock.readers` exist, we will retain the `lock.writer` latch and invoke `lock.u_wr_upgrade()` after registering the wait in `PERFORMANCE_SCHEMA`.

`dict_sys.lock(SRW_LOCK_CALL)` calls are rather rare, mostly occurring during database startup and DDL. The only calls during a DML-only workload should be related to loading tables into or evicting them from the `dict_sys` cache. Because `dict_sys.latch` can also be `SRWLOCK` or `rw_lock_t` and because the exclusive lock requests are rare, it it probably does not make sense to optimize that case.

dr-m: I don’t think that it makes sense to improve `dict_sys_t::lock()`. Only `psi_wr_lock()` can be improved: 509d5e761a4db8b598616b4165f992f08c319c69. All other callers of any `wr_lock_try()` would back off more and not proceed to directly wait for the exclusive lock.

storage/innobase/sync/srw_lock.cc
dr-m: 0ba2fbe07a8e5d5fb59015cd137e27ce39533814 addresses this.

storage/innobase/sync/srw_lock.cc
dr-m: 0ba2fbe07a8e5d5fb59015cd137e27ce39533814 addresses this.

storage/innobase/sync/srw_lock.cc
```diff
@@ -392,14 +402,45 @@ template void ssux_lock_impl<false>::wr_wait(uint32_t);
...
+  uint32_t wl= 1 + writer.lock.fetch_add(1, std::memory_order_acquire);
+
   for (;;)
   {
-    writer.wr_lock();
-    bool acquired= rd_lock_try();
-    writer.wr_unlock();
-    if (acquired)
+    if (UNIV_LIKELY(writer.HOLDER & wl))
+      writer.wait(wl);
```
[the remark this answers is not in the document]
dr-m: You are right. Before we get into this member function `rd_wait()`, we will already have failed to acquire a shared latch. It is possible that another thread meanwhile invoked `wr_unlock()` and yet another one invoked `u_lock()` (I renamed the MySQL 5.7 "SX" to "U"). In the `spinloop=false` state we would unnecessarily wait for `u_unlock()` in the other thread. This wait could also be ‘rescued’ by another `rd_wait()` thread that would invoke `writer.wake_all()`.

storage/innobase/sync/srw_lock.cc
```diff
@@ -392,14 +402,45 @@ template void ssux_lock_impl<false>::wr_wait(uint32_t);
...
+    uint32_t lk= rd_lock_try_low();
+    if (!lk)
       break;
+    if (UNIV_UNLIKELY(lk == WRITER)) /* A wr_lock() just succeeded. */
+      writer.wake_all(); /* Immediately wake up (also) wr_lock(). */
```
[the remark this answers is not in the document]
dr-m: There could be several of these `for (;;)` loops executing in multiple threads, each of them ‘consuming’ wake-ups in `wait()`. The `wr_unlock()` would only wake up one of these threads.

If we invoked `wake()` instead of `wake_all()`, then this wake-up signal could go into a waste, waking up another `rd_wait()` instead of waking the `wr_lock()` which is the only thread that can make progress. Only after the blocked `wr_lock()` thread does `wr_unlock()` can all the `rd_lock()` make progress.

The `wake_all()` call at the end I added ‘just in case’; initially it was `wake()`. I don’t remember if I observed a hang when I had a `wake()` at that spot.

PR #3320, opened 2024-06-07
MDEV-32176 Contention in ha_innobase::info_low()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-32176*
## Description
During a Sysbench `oltp_point_select` workload with *1 table* and 400 concurrent connections, a bottleneck on `dict_table_t::lock_mutex` was observed in `ha_innobase::info_low()`.

`dict_table_t::lock_latch`: Replaces `lock_mutex`.

In `ha_innobase::info_low()` and several other places, we will acquire a shared `dict_table_t::lock_latch` or we may elide the latch if hardware memory transactions are available.

`innobase_build_v_templ()`: Remove the parameter `bool locked`, and require the caller to hold exclusive `dict_table_t::lock_latch` (instead of holding an exclusive `dict_sys.latch`). This could help some workloads that make use of virtual columns.
## Release Notes
This is a performance improvement that may not be worth mentioning, because it affects a special case (massively parallel read-only workload on a small number of tables).
## How can this PR be tested?
This is best tested with a read-only workload using 1 table (or a very small number of tables).

The changed code is at rather low level and relatively well covered by the regression test suite.

Note that this branch also includes #3317 for performance testing purposes.
## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -3267,9 +3267,9 @@ static bool innobase_query_caching_table_check_low(
 	}
 #endif
 
-	table->lock_mutex_lock();
+	table->lock_shared_lock();
 	auto len= UT_LIST_GET_LEN(table->locks);
-	table->lock_mutex_unlock();
+	table->lock_shared_unlock();
 	return len == 0;
 }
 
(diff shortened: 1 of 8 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -5581,43 +5581,29 @@ is done when the table first opened.
...
-	bool			locked)
+	const dict_add_v_col_t*	add_v)
 {
 	ulint	ncol = unsigned(ib_table->n_cols) - DATA_N_SYS_COLS;
 	ulint	n_v_col = ib_table->n_v_cols;
 	bool	marker[REC_MAX_N_FIELDS];
 
 	DBUG_ENTER("innobase_build_v_templ");
 	ut_ad(ncol < REC_MAX_N_FIELDS);
+	ut_ad(ib_table->lock_mutex_is_owner());
```
dr-m, line 5598: We could actually (mis)use the `ib_table->autoinc_mutex` for this. As far as I can tell (@FooBarrior please correct me if I am wrong), we only need mutual exclusion of all callers of this function.

vaintroub (approved):
Looks good to me.

PR #3321, opened 2024-06-07
WIP - MDEV-33408 Initial support for vector DELETE and UPDATE 

HugoWenTD:
- [x] *The Jira issue number for this PR is: MDEV-33408*

## Description
When the source row is deleted, mark the corresponding node in HNSW index by setting `tref` to null. An index is added for the `tref` in secondary table for faster searching of the to-be-marked nodes.

The nodes marked as deleted will still be used for search, but will not be included in the final query results.

As skipping deleted nodes and not adding deleted nodes for new-inserted nodes' neighbor list could impact the performance, we now only skip these nodes in search results.

TODO:
- for some reason the bitmap is not set for hlindex during the delete so I had to temporarily comment out [this line](https://github.com/MariaDB/server/blob/bb-11.6-MDEV-32887-vector/sql/sql_base.cc#L9874)

## Release Notes
N/A - Vector support development

## How can this PR be tested?

MTR test has been updated to verify the DELETE and UPDATE.

Benchmarking test had been done with `DATASET=random-s-100-euclidean` and `ef_construction_multiplier = 1` ( have to set this to 1 because default 4 always returns recall of 1 and hard to compare ).

#### Before
```
Insert time:  195.7592477798462
Recall:       0.872
QPS:          701.594
```

#### After
first run
```
Insert time:  218.7464780807495
Recall:       0.885
QPS:          722.408
```
second run
```
Insert time:  172.49292421340942
Recall:       0.892
QPS:          778.593
```

## PR quality check

## Copyright


vuvova:
As far as I can see, you went with the approach to always store the complete vector in the secondary table. okay.
Did you benchmark the overhead as you wrote in MDEV-33408?

It, of course, doesn't apply at all to bb-11.6-MDEV-32887-vector branch. But also, it assumes that all refs are unique, which is not the case if one can delete and update rows. I'll try to find a solution for this.

HugoWenTD:
Hi @vuvova , 

> As far as I can see, you went with the approach to always store the complete vector in the secondary table. okay. Did you benchmark the overhead as you wrote in MDEV-33408?

Yes, I've done some benchmarking a couple of weeks ago. As I noted in previous email, the search speed was significantly impacted, **approximately 5 times slower due to the frequent calling of `ha_index_read_map` for the second table**. However, if the new branch provides improved cache management, the results are likely to be better.

I was just storing the entire vector in the secondary table in the initial change. The primary performance bottleneck was the frequent calls to the `ha_index_read_map` function, as shown in the flame graphs generated by the benchmarking tool. I don't think scalar quantization would significantly improving the performance in this case.

#### Insert
perf data inserting 2024-06-06-14-08-05
#### Search
image


> It, of course, doesn't apply at all to bb-11.6-MDEV-32887-vector branch. But also, it assumes that all refs are unique, which is not the case if one can delete and update rows. I'll try to find a solution for this.

You are right Sergei. I could not think of a good method to handle duplicate refs while keep the primary index clean, but I'd like to hear your suggestions on it.

----

What's your opinion about next steps?

- Do you think it worthwhile to continue testing the benchmark with this approach and add scalar quantization on bb-11.4-vec-preview branch?
- Would you like me to apply a similar approach to bb-11.6-MDEV-32887-vector and do benchmark? Do you have other proposed solutions?

vuvova:
I'm trying to change the storage format (= table structure) to support deletions. It'll need a day or two. After that it could be possible to apply your commit to bb-11.6-MDEV-32887-vector.

How do you plan to do scalar quantization?

HugoWenTD:
Thank you @vuvova . Please let me know when the branch is ready.

#### For the scalar quantization, here are some thoughts. I look forward to hearing your feedback.

Vector quantization that could improve performance (mostly for speeding up index creation) and reduce the index size including:

1. reducing the size of each dimension (e.g., from 4-byte to 2-byte, 1-byte, or even 1-bit) or
2. decreasing the dimensionality of the vector (e.g., from 100 to 10 dimensions).

While users can partially achieve dimensionality reduction through preprocessing before inserting the data, reducing the size of each dimension requires additional support from the MariaDB server.

1. One option is to introduce support for vector data types composed of 4-byte, 2-byte float, or 1-bit dimensions, and support indexing on these data types. With these dedicated data types, one can specify how to parse the binary data. The index logic is mostly common but just the data size difference.
    1. Users will have a choice to directly store the quantized data to save space.
    2. Seems MariaDB does not support functional indexes but users can create separate invisible column and create index against it. I didn’t not have experience with it may need more research on it.
2. Another option is to do quantization for the vector data only when building and searching the index, there need to be a way to allow user selecting the quantization type while creating the index.
    1. In this way there is less storage overhead but less flexible for users comparing to option 1.
    2. I’m not sure what’s the best way to allow user select the quantization function for an index, probably could add a new [index option](https://mariadb.com/kb/en/create-index/)?

Pg-vector supports functional indexes and they are similar to option 1 of supporting different vector data types. Their tests show that converting from 4-byte float to 2-byte float significantly reduces index building time and index size without substantially impacting recall. Conversion to 1-bit further accelerates the process but may result in poor recall rates if the dimensionality is not sufficiently high.

With the quantization, performance improvements in MariaDB might differ from those observed in Pg-vector due to potential differences in bottlenecks.

What’s your opinion about the options? Maybe  we should start with saving 2-type float in index, and see how the performance differs?

vuvova:
I've pushed the new storage format. It gives ~10-20% slowdown for inserts and up to 5x slowdown for selects.
But it doesn't really matter, because with a cache (that I have in my local branch) this slowdown should disappear.

Now it should be easy to implement deletion, duplicate refs won't be an issue.

----

As far as quantization is concerned, using `_Float16` seems to be a rather easy change. Going down to one byte or bit is more work and more recall loss.

I would definitely suggest to keep user data as is and only change what we store in the index.

Reducing dimensionality — I see it as a very promising technique, but, again, only in the index. It's more tricky, so let's do it later.

And, yes, I'm afraid we'll need index options for this eventually.

HugoWenTD:
@vuvova Thank you for the quick turnaround and branch updates. It appears that your commit https://github.com/MariaDB/server/commit/77a016686ec2a2617dd6489a756b1f9f11a78d9f introduced `gref`, the ref to the corresponding row in the second graph table. It also removed the need for searching in the second table, as each node knows precisely where to load its neighbors. That's an excellent idea!

I've also learned that it assumes the gref of a node will not change even if the row is updated. I presume this holds true for most storage engines?

With this new structure, as you mentioned in the inline comment, the rows of the second graph table must not be deleted. If we later want to support the cleanup or partial rebuild of deleted vectors, these rows should still be retained, but I don't think that's a big issue. And if the user plans to delete a large number of records, they should probably consider rebuilding the index instead.

---

There might be additional index options required, such as ef_construct and max_edge_per_node. Currently, they are session variables, but it might make more sense to make them index options.

---

I can start implementing the delete functionality against your branch next week and update this PR.

vuvova:
`gref` is not changed — it is true at least for MyISAM and Aria (because they don't update all indexes on UPDATE, only changed ones) and for InnoDB (where `gref` is a hidden PK, so a logical row id anyway).

You can mark a deleted row by setting `ref` to NULL — note that unlike all other columnts, `ref` is nullable. For this very reason.

HugoWenTD:
Hi @vuvova, I've drafted this commit https://github.com/MariaDB/server/pull/3321/commits/b75c1aebe34c06c0dd06713159c9ecd8e5700330 for DELETE and UPDATE.

There are two items on which I would like to hear your guidance. Thank you in advance!

- on handler::ha_delete_all_rows() - what is the best way to clean up the secondary table when user delete all rows from table, any code examples?
- for some reason the bitmap is not set for hlindex during the delete. so I had to temporarily comment out [this line](https://github.com/MariaDB/server/blob/bb-11.6-MDEV-32887-vector/sql/sql_base.cc#L9874). Do you know where in the code it should be set?

HugoWenTD:
I'm now using `this->hlindex->file->ha_delete_all_rows();` to delete all rows in second table and it seems to be working well [compare/b75c1aeb..24f39d50](https://github.com/MariaDB/server/compare/b75c1aebe34c06c0dd06713159c9ecd8e5700330..24f39d50af88b433889a0e4c4f46972d0512c789).
Please let me know if there's better approach.

sql/vector_mhnsw.cc
```diff
@@ -735,7 +766,7 @@ int mhnsw_first(TABLE *table, KEY *keyinfo, Item *dist, ulonglong limit)
   *(ulonglong*)context= limit;
   context+= context_size;
 
-  while (limit--)
+  while (candidates.elements && limit--)
```
vuvova, line 769: not needed, I've done this few days ago with
{code:c}
  if (limit > candidates.num)
    limit= candidates.num;
{code}

sql/handler.cc
```diff
@@ -8126,6 +8131,9 @@ int handler::ha_update_row(const uchar *old_data, const uchar *new_data)
   DBUG_ASSERT(new_data == table->record[0]);
   DBUG_ASSERT(old_data == table->record[1]);
 
+  if (table->open_hlindexes_for_write())
+    return 1;
+
```
vuvova, line 8136: strange, `ha_write_row()` doesn't do it, it opens the table once, not for every row. perhaps update/delete should do the same?

vuvova:
merged, thanks

PR #3322, opened 2024-06-10
MDEV-33161 Function pointer signature mismatch in LF_HASH

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33161*
## Description
In `cmake -DWITH_UBSAN=ON` builds with `clang` but not with GCC, `-fsanitize=undefined` will flag several runtime errors on function pointer mismatch related to the lock-free hash table `LF_HASH`.

Let us use matching function signatures and remove function pointer casts in order to avoid potential bugs due to undefined behaviour.
## Release Notes
This is probably not worth mentioning, because no actual bugs related to this are known.
## How can this PR be tested?
I ran the following after both `cmake -DCMAKE_BUILD_TYPE=Debug` and with `cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo` using `clang` 18.1.6 and `-DWITH_UBSAN=ON`:
```sh
UBSAN_OPTIONS=halt_on_error=0 ./mtr --parallel=auto --max-test-fail=0 --force --big-test
grep -li lf_ mysql-test/var/*/log/mysqld*err*
```
The `mtr` driver would appear to hang due to the sheer amount of test failures. For me, the output in the window title listed 5235 tests, 4924 failures and 985 tests left at the time of the apparent hang.

Only one false alarm (due to a similar bug elsewhere) was flagged on my system:
```
CURRENT_TEST: innodb.trx_sys_t_find_lf_hash_error
…
mysys/hash.c:197:20: runtime error: call to function get_sys_var_length(sys_var const*, unsigned long*, char) through pointer to incorrect function type 'unsigned char *(*)(const unsigned char *, unsigned long *, char)'
sql/set_var.cc:53: note: get_sys_var_length(sys_var const*, unsigned long*, char) defined here
```
## PR quality check

### DIFF
--- a/mysys/lf_alloc-pin.c
+++ b/mysys/lf_alloc-pin.c
@@ -242,8 +242,10 @@ void lf_pinbox_put_pins(LF_PINS *pins)
   return;
 }
 
-static int ptr_cmp(void **a, void **b)
+static int ptr_cmp(const void *pa, const void *pb)
 {
+  const void *const*a= pa;
+  const void *const*b= pb;
   return *a < *b ? -1 : *a == *b ? 0 : 1;
 }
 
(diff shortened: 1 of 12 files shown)


vaintroub (approved):
Ok, but 
1) Can you file a bug about enabling this clang flag you found about, the flag that forbids incompatible function pointer casting. 
2) at some point we'll need to have some type-aware callbacks, specifically for this trx hash in Innodb. Or maybe C++ style iterator, rather than callbacks. This patch erased types from function signatures in many cases, making it less idiomatic C++.

dr-m:
As requested, I filed [MDEV-34348](https://jira.mariadb.org/browse/MDEV-34348) MariaDB is violating `clang-16 -Wcast-function-type-strict`

PR #3323, opened 2024-06-10
MDEV-32892: IO Thread Reports False Error When Stopped During Connect…

bnestere:
MDEV-32892: IO Thread Reports False Error When Stopped During Connecting to Primary

The IO thread can report error code 2013 into the error log when it is stopped during the initial connection process to the primary, as well as when trying to read an event. However, because the IO thread is being stopped, its connection to the primary is force-killed by the signaling thread (see THD::awake_no_mutex()), and thereby these connection errors should be ignored.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_stop_slave_error.test
+++ b/mysql-test/suite/rpl/t/rpl_stop_slave_error.test
@@ -6,9 +6,24 @@ source include/master-slave.inc;
 
 connection master;
 sync_slave_with_master;
+
+--let $iter=100
+--echo # MDEV-32892: Repeatedly starting/stopping io_thread..
+--disable_query_log
+while ($iter)
+{
+  stop slave io_thread;
+  start slave io_thread;
+  --dec $iter
+}
+--enable_query_log
 source include/stop_slave.inc;
+
 let SEARCH_FILE=$MYSQLTEST_VARDIR/tmp/slave_log.err;
 let SEARCH_PATTERN=Error reading packet from server: Lost connection;
+source include/search_pattern_in_file.inc;
+
+let SEARCH_PATTERN=error code: 2013;
 source include/search_pattern_in_file.inc;
 
 source include/start_slave.inc;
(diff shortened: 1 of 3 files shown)


mysql-test/suite/rpl/t/rpl_stop_slave_error.test
```diff
@@ -6,11 +6,26 @@ source include/master-slave.inc;
...
+--let $iter=1000
+--echo # MDEV-32892: Repeatedly starting/stopping io_thread..
+--disable_query_log
+while ($iter)
+{
+  stop slave io_thread;
+  start slave io_thread;
+  --dec $iter
+}
+--enable_query_log
```
knielsen, line 19: Can we use something less than 1000 iterations here? I can imagine it was useful during development testing, but I think it will take a long time to run this part of the test. So I'd reduce this a lot before pushing, say so that it doesn't take more than a couple seconds to run - any later regression in the code should still be caught in the buildbot sooner or later.

knielsen (approved):
Looks good, with just one suggestion for the testcase.

mysql-test/suite/rpl/t/rpl_stop_slave_error.test
bnestere: On my laptop, 1000 iterations takes less than a second (`0.8s`), but sure, I can reduce it just in case, say for  overwhelmed BB runs.

knielsen:

>> +--let $iter=1000
> +--echo # MDEV-32892: Repeatedly starting/stopping io_thread..
> +--disable_query_log
> +while ($iter)
> +{
> +  stop slave io_thread;
> +  start slave io_thread;
> +  --dec $iter
> +}
> +--enable_query_log
>
> On my laptop, 1000 iterations takes less than a second (`0.8s`), but
> sure, I can reduce it just in case, say for overwhelmed BB runs.

Oh, I see. I think I was confusing it with --source
include/stop_slave_io.inc, which IIRC would take a lot longer. Then it's
less critical, but still ok to reduce if you want.

 - Kristian.

PR #3327, opened 2024-06-11
MDEV-33894: Resurrect innodb_log_write_ahead_size

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33894*
## Description
In commit 685d958e38b825ad9829be311f26729cccf37c46 (MDEV-14425) we assumed that the physical block size (typically 512 or 4096 bytes) would be an optimal ib_logfile0 write size. In fact, some file systems may have a much larger allocation block size.
## Release Notes
We will resurrect the parameter `innodb_log_write_ahead_size`, so that read-on-write to the ib_logfile0 can be avoided. This parameter must be a power of 2 multiple of the physical block size, and the `innodb_log_file_size` must be an integer multiple of it.
## How can this PR be tested?
```sh
```
FIXME: The test innodb.log_file_size_online occasionally fails (server fails to restart due to corrupted redo log).

FIXME: The test currently fails when the log is memory-mapped.
## PR quality check


storage/innobase/handler/ha_innodb.cc
```diff
@@ -3809,6 +3816,20 @@ static int innodb_init_params()
 		DBUG_RETURN(HA_ERR_INITIALIZATION);
 	}
 
+	MYSQL_SYSVAR_NAME(log_write_ahead_size).max_val = log_sys.buf_size;
```
mariadb-DebarunBanerjee, line 3819: Not an issue for now as innodb_log_buffer_size cannot change dynamically. If we ever try to make it dynamic, the innodb_write_ahead_size and the max_limit needs to be adjusted accordingly.

mariadb-DebarunBanerjee (commented):
Good to see the feature back. Perhaps it was too early to remove it with 4k/8k sectors storages still in use. The “read-on-write”avoidance is crucial and I agree with the decision.

I have reviewed the spec and high level changes and here are some initial comments. 
This MDEV is more like a feature as opposed to a bug. I think both QA and Doc should be signalled.

I have now started reviewing the core implementation in log0log.cc and finish soon. 

One thing I observed is that with 4k sector size the very DB creation hits an assert in debug mode. Any mtr test would fail on my HDD for the same reason. It is not repeatable on SSD or RAM disk with 512 bytes physical sector size.

```
sudo fdisk -l /dev/sda
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 4096 bytes
I/O size (minimum/optimal): 4096 bytes / 4096 bytes
```

```
debanerj@deb-7010:/home/hdd/deb/maria-src5/bld_install$ mariadb-install-db --srcdir=/home/hdd/deb/maria-src5/ --datadir=./maria_data
Installing MariaDB/MySQL system tables in './maria_data' ...
mariadbd: /home/hdd/deb/maria-src5/storage/innobase/log/log0log.cc:852: lsn_t log_t::write_buf() [with bool release_latch = true; lsn_t = long unsigned int]: Assertion `write_size_1 >= block_size_1' failed.
240620 18:12:48 [ERROR] mysqld got signal 6 ;

log/log0log.cc:853(unsigned long log_t::write_buf<true>())[0x57899ef7e046]
log/log0log.cc:1040(log_write_up_to(unsigned long, bool, completion_callback const*))[0x57899ef7b1a3]
buf/buf0flu.cc:1954(log_checkpoint_low(unsigned long, unsigned long))[0x57899f22b9e8]
buf/buf0flu.cc:1998(log_checkpoint())[0x57899f22bc34]
buf/buf0flu.cc:2098(buf_flush_wait_flushed(unsigned long))[0x57899f22c396]
buf/buf0flu.cc:2004(log_make_checkpoint())[0x57899f22bc5a]
srv/srv0start.cc:246(create_log_file(bool, unsigned long))[0x57899f12bd28]
srv/srv0start.cc:1356(srv_start(bool))[0x57899f12fccd]
handler/ha_innodb.cc:4251(innodb_init(void*))[0x57899ee78325]
sql/handler.cc:655(ha_initialize_handlerton(st_plugin_int*))[0x57899ea4a51f]
sql/sql_plugin.cc:1454(plugin_do_initialize(st_plugin_int*, unsigned int&))[0x57899e6b002f]
sql/sql_plugin.cc:1507(plugin_initialize(st_mem_root*, st_plugin_int*, int*, char**, bool))[0x57899e6b0376]
sql/sql_plugin.cc:1765(plugin_init(int*, char**, int))[0x57899e6b10e8]
sql/mysqld.cc:5256(init_server_components())[0x57899e4f1e31]
sql/mysqld.cc:5882(mysqld_main(int, char**))[0x57899e4f317f]
sql/main.cc:34(main)[0x57899e4e799d]
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -19989,20 +20021,19 @@ i_s_innodb_sys_virtual,
...
+  MYSQL_SYSVAR_NAME(max_undo_log_size).max_val=
+    1ULL << (32U + srv_page_size_shift);
+  MYSQL_SYSVAR_NAME(max_undo_log_size).min_val=
+    MYSQL_SYSVAR_NAME(max_undo_log_size).def_val=
+    ulonglong{SRV_UNDO_TABLESPACE_SIZE_IN_PAGES} << srv_page_size_shift;
+  MYSQL_SYSVAR_NAME(max_undo_log_size).max_val=
+    1ULL << (32U + srv_page_size_shift);
+
+  ut_ad(MYSQL_SYSVAR_NAME(log_write_ahead_size).max_val == log_sys.buf_size);
+  log_sys.set_write_ahead_size(innodb_log_write_ahead_size);
```
dr-m, line 20036: On Linux and Microsoft Windows, we also have to adjust the `min_val` to `log_sys.get_block_size()`, which can be 4096 bytes. I failed to test this on my 4096-byte-sector-size HDD.

storage/innobase/log/log0log.cc
```diff
@@ -824,64 +842,119 @@ template<bool release_latch> inline lsn_t log_t::write_buf() noexcept
...
   {
+    ut_ad(write_lock.is_owner());
     ut_ad(!recv_no_log_write);
     write_lock.set_pending(lsn);
     ut_ad(write_lsn >= get_flushed_lsn());
     const size_t block_size_1{get_block_size() - 1};
-    lsn_t offset{calc_lsn_offset(write_lsn) & ~lsn_t{block_size_1}};
+    size_t write_size_1{innodb_log_write_ahead_size - 1};
+    ut_ad(ut_is_2pow(write_size_1 + 1));
+    ut_ad(write_size_1 >= block_size_1);
```
dr-m, line 852: We may have `write_size_1<block_size_1` during `srv_start()` when running with 4096-byte sector size. We’d better adjust the `innodb_log_write_ahead_size` during `create_log_file()` to keep this logic simple.

storage/innobase/handler/ha_innodb.cc
dr-m: For some future-proofing it would make sense to add a test for `SET GLOBAL innodb_log_buffer_size` with a comment that it must be consistent with `innodb_log_write_ahead_size`.

storage/innobase/handler/ha_innodb.cc
dr-m: Adjusted in 910d3266d9cd0813f3323bd7b4befb4efe19e078.

storage/innobase/log/log0log.cc
dr-m: Adjusted in 910d3266d9cd0813f3323bd7b4befb4efe19e078 and 61506d9eef08e95458fcba01f5f87c6d41342378, before the calls to `log_sys.set_block_size()`.

dr-m:
The experiment ef96260d56853357a97e93cde0bee393a619b4fc seems to break things massively.

jkrauss82:
While configuring a new MariaDB instance running on ZFS I have stumbled over the parameter `innodb-log-write-ahead-size`.

According to this [guide](https://shatteredsilicon.net/mysql-mariadb-innodb-on-zfs/), we should set the parameter to match the recordsize of the zfs dev (128k in our case):

> [innodb_log_write_ahead_size](https://mariadb.com/kb/en/innodb-system-variables/#innodb_log_write_ahead_size)=16384: In order to prevent torn pages and avoid read-on-write overhead, we should set this to the underlying [file system block size](https://shatteredsilicon.net/disk-and-file-system-optimisation/). If we are keeping the InnoDB logs in the data directory (as is the default), we should set this to what we set the ZFS recordsize to.

I recognize the current max value of the parameter in 10.11.9-MariaDB-deb12-log is `4096`.

As I think their recommendation makes sense it would be great if larger sizes (128k+) could be supported in future releases.

PR #3331, opened 2024-06-12
MDEV-33746 Supply missing override markings

DaveGosselin-MariaDB:
Find and fix missing virtual override markings.  Updates CMAKE_CXX_FLAGS to include -Wsuggest-override and -Winconsistent-missing-override.  In a few places, redundant virtual markings were removed.

### DIFF
--- a/sql/item_cmpfunc.h
+++ b/sql/item_cmpfunc.h
@@ -236,12 +236,12 @@ public:
   Item_bool_func(THD *thd, Item *a, Item *b, Item *c): Item_int_func(thd, a, b, c) {}
   Item_bool_func(THD *thd, List<Item> &list): Item_int_func(thd, list) { }
   Item_bool_func(THD *thd, Item_bool_func *item) :Item_int_func(thd, item) {}
-  const Type_handler *type_handler() const { return &type_handler_bool; }
-  const Type_handler *fixed_type_handler() const { return &type_handler_bool; }
-  CHARSET_INFO *compare_collation() const { return NULL; }
-  bool fix_length_and_dec() { decimals=0; max_length=1; return FALSE; }
-  uint decimal_precision() const { return 1; }
-  bool need_parentheses_in_default() { return true; }
+  const Type_handler *type_handler() const override { return &type_handler_bool; }
+  const Type_handler *fixed_type_handler() const override { return &type_handler_bool; }
+  CHARSET_INFO *compare_collation() const override { return NULL; }
+  bool fix_length_and_dec() override { decimals=0; max_length=1; return FALSE; }
+  uint decimal_precision() const override { return 1; }
+  bool need_parentheses_in_default() override { return true; }
 };
 
 
@@ -253,11 +253,11 @@ public:
 class Item_func_truth : public Item_bool_func
 {
 public:
-  virtual bool val_bool();
-  virtual longlong val_int();
-  virtual bool fix_length_and_dec();
-  virtual void print(String *str, enum_query_type query_type);
-  enum precedence precedence() const { return CMP_PRECEDENCE; }
+  bool val_bool() override;
+  longlong val_int() override;
+  bool fix_length_and_dec() override;
+  void print(String *str, enum_query_type query_type) override;
+  enum precedence precedence() const override { return CMP_PRECEDENCE; }
 
 protected:
   Item_func_truth(THD *thd, Item *a, bool a_value, bool a_affirmative):
@@ -287,8 +287,8 @@ class Item_func_istrue : public Item_func_truth
 public:
   Item_func_istrue(THD *thd, Item *a): Item_func_truth(thd, a, true, true) {}
   ~Item_func_istrue() = default;
-  virtual const char* func_name() const { return "istrue"; }
-  Item *get_copy(THD *thd)
+  const char* func_name() const override { return "istrue"; }
+  Item *get_copy(THD *thd) override
   { return get_item_copy<Item_func_istrue>(thd, this); }
 };
 
@@ -303,11 +303,11 @@ public:
   Item_func_isnottrue(THD *thd, Item *a):
     Item_func_truth(thd, a, true, false) {}
   ~Item_func_isnottrue() = default;
-  virtual const char* func_name() const { return "isnottrue"; }
-  bool find_not_null_fields(table_map allowed) { return false; }
-  Item *get_copy(THD *thd)
+  const char* func_name() const override { return "isnottrue"; }
+  bool find_not_null_fields(table_map allowed) override { return false; }
+  Item *get_copy(THD *thd) override
   { return get_item_copy<Item_func_isnottrue>(thd, this); }
-  bool eval_not_null_tables(void *) { not_null_tables_cache= 0; return false; }
+  bool eval_not_null_tables(void *) override { not_null_tables_cache= 0; return false; }
 };
 
 
@@ -320,8 +320,8 @@ class Item_func_isfalse : public Item_func_truth
 public:
   Item_func_isfalse(THD *thd, Item *a): Item_func_truth(thd, a, false, true) {}
   ~Item_func_isfalse() = default;
-  virtual const char* func_name() const { return "isfalse"; }
-  Item *get_copy(THD *thd)
+  const char* func_name() const override { return "isfalse"; }
+  Item *get_copy(THD *thd) override
   { return get_item_copy<Item_func_isfalse>(thd, this); }
 };
 
@@ -336,11 +336,11 @@ public:
   Item_func_isnotfalse(THD *thd, Item *a):
     Item_func_truth(thd, a, false, false) {}
   ~Item_func_isnotfalse() = default;
-  virtual const char* func_name() const { return "isnotfalse"; }
-  bool find_not_null_fields(table_map allowed) { return false; }
-  Item *get_copy(THD *thd)
+  const char* func_name() const override { return "isnotfalse"; }
+  bool find_not_null_fields(table_map allowed) override { return false; }
+  Item *get_copy(THD *thd) override
   { return get_item_copy<Item_func_isnotfalse>(thd, this); }
-  bool eval_not_null_tables(void *) { not_null_tables_cache= 0; return false; }
+  bool eval_not_null_tables(void *) override { not_null_tables_cache= 0; return false; }
 };
 
 
@@ -427,10 +427,10 @@ public:
   Item_bool_func2(THD *thd, Item *a, Item *b):
     Item_bool_func(thd, a, b) { }
 
-  bool is_null() { return MY_TEST(args[0]->is_null() || args[1]->is_null()); }
+  bool is_null() override { return MY_TEST(args[0]->is_null() || args[1]->is_null()); }
   COND *remove_eq_conds(THD *thd, Item::cond_result *cond_value,
-                        bool top_level);
-  bool count_sargable_conds(void *arg);
+                        bool top_level) override;
+  bool count_sargable_conds(void *arg) override;
   /*
     Specifies which result type the function uses to compare its arguments.
     This method is used in equal field propagation.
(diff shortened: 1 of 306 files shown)


sql/item_cmpfunc.h
```diff
@@ -303,11 +303,11 @@ class Item_func_isnottrue : public Item_func_truth
   Item_func_isnottrue(THD *thd, Item *a):
     Item_func_truth(thd, a, true, false) {}
   ~Item_func_isnottrue() = default;
-  virtual const char* func_name() const { return "isnottrue"; }
-  bool find_not_null_fields(table_map allowed) { return false; }
-  Item *get_copy(THD *thd)
+  virtual const char* func_name() const override { return "isnottrue"; }
+  bool find_not_null_fields(table_map allowed) override { return false; }
```
dr-m, line 307: While we are changing the lines, we could also fix any `-Wunused-parameter` warnings:
```c++
  bool find_not_null_fields(table_map) override { return false; }
```

sql/item_row.h
```diff
@@ -57,104 +57,104 @@ class Item_row: public Item_fixed_hybrid,
...
-  double val_real()
+  double val_real() override
   {
     illegal_method_call((const char*)"val");
     return 0;
   };
-  longlong val_int()
+  longlong val_int() override
   {
     illegal_method_call((const char*)"val_int");
```
dr-m, line 81: Side note (to be handled in a separate fix): C-style casts should be avoided, and string literals are already compatible with `const char*` by default.

Some savings could be had if `Item_row::illegal_method_call(const char*)` returned a `size_t` of 0. In that case, tail call optimization could be enabled:
```c++
return illegal_method_call("val_int");
```
or
```c++
return reinterpret_cast<String*>(illegal_method_call("val_str"));
```

sql/item_strfunc.h
```diff
@@ -864,14 +864,14 @@ class Item_func_encrypt :public Item_str_binary_checksum_func
...
-  bool fix_length_and_dec() { maybe_null=1; max_length = 13; return FALSE; }
-  const char *func_name() const { return "encrypt"; }
-  bool check_vcol_func_processor(void *arg)
+  String *val_str(String *) override;
+  bool fix_length_and_dec() override { maybe_null=1; max_length = 13; return FALSE; }
+  const char *func_name() const override { return "encrypt"; }
+  bool check_vcol_func_processor(void *arg) override
   {
     return FALSE;
   }
```
dr-m, line 873: `arg` is unused.

sql/item_subselect.h
```diff
@@ -360,12 +360,12 @@ class Item_maxmin_subselect :public Item_singlerow_subselect
 public:
   Item_maxmin_subselect(THD *thd, Item_subselect *parent,
 			st_select_lex *select_lex, bool max);
-  virtual void print(String *str, enum_query_type query_type);
-  void cleanup();
+  virtual void print(String *str, enum_query_type query_type) override;
```
dr-m, line 363: I’d be consistent and remove `virtual` from anything that is declared `override`.

sql/log_event.h
```diff
@@ -3002,28 +3004,30 @@ class Rand_log_event: public Log_event
...
-  Log_event_type get_type_code() { return RAND_EVENT;}
-  int get_data_size() { return 16; /* sizeof(ulonglong) * 2*/ }
+  Log_event_type get_type_code() override { return RAND_EVENT;}
+  int get_data_size() override { return 16; /* sizeof(ulonglong) * 2*/ }
 #ifdef MYSQL_SERVER
-  bool write();
+  bool write() override;
+#ifdef HAVE_REPLICATION
+  bool is_part_of_group() override { return 1; }
 #endif
```
dr-m, line 3022: It would be better to indent nested preprocessor conditions with one space per nesting level after the initial `#`:
```c++
#ifdef MYSQL_SERVER
  bool write() override;
# ifdef HAVE_REPLICATION
  bool is_part_of_group() override { return true; }
# endif
```

sql/log_event.h
```diff
@@ -5659,18 +5676,20 @@ class Ignorable_log_event : public Log_event {
   virtual ~Ignorable_log_event();
 
 #ifndef MYSQL_CLIENT
-  void pack_info(Protocol*);
+#ifdef HAVE_REPLICATION
+  void pack_info(Protocol*) override;
+#endif
 #endif
 
 #ifdef MYSQL_CLIENT
```
dr-m, line 5684: This could be replaced with `#else`.

sql/sp_head.h
```diff
@@ -1576,17 +1576,17 @@ class sp_instr_jump : public sp_instr_opt_meta
...
+  virtual uint opt_shortcut_jump(sp_head *sp, sp_instr *start) override;
 
-  virtual void opt_move(uint dst, List<sp_instr> *ibp);
+  virtual void opt_move(uint dst, List<sp_instr> *ibp) override;
 
-  virtual void backpatch(uint dest, sp_pcontext *dst_ctx)
+  virtual void backpatch(uint dest, sp_pcontext *dst_ctx) override
   {
     /* Calling backpatch twice is a logic flaw in jump resolution. */
     DBUG_ASSERT(m_dest == 0);
```
dr-m, line 1592: `dst_ctx` is an unused parameter and should be omitted. Also the `virtual` is redundant because it is already implied by `override`.

dr-m (approved):
Thank you. As far as I understand, this may omit some dead replication code when the code is built without `HAVE_REPLICATION`, I guess, as part of the embedded server library.

I have some additional suggestions:
* Ensure that `git grep 'virtual.*override'` returns no results.
* Add the warning flags to `cmake/maintainer.cmake` in order to prevent regressions.
* Try to ensure that the names of any unused parameters will be omitted from the changed definitions, to suppress `-Wunused-parameter`.

DaveGosselin-MariaDB:
> Thank you. As far as I understand, this may omit some dead replication code when the code is built without `HAVE_REPLICATION`, I guess, as part of the embedded server library.

Especially in log_event.h, there are some methods that are marked `override` that, without the additional `#ifdef`s I added, would be marked during compilation as an error because they don't override any `virtual` methods.  For example, the `virtual is_part_of_group()` method is present only when `#if defined(MYSQL_SERVER) && defined(HAVE_REPLICATION)` at line 1478 is true.

> I have some additional suggestions:
> 
> * Ensure that `git grep 'virtual.*override'` returns no results.

Completely agree, will incorporate in this patch.

> * Add the warning flags to `cmake/maintainer.cmake` in order to prevent regressions.

Great suggestion, will do.

> * Try to ensure that the names of any unused parameters will be omitted from the changed definitions, to suppress `-Wunused-parameter`.

This is a great idea, but I'd prefer to implement it in a separate delivery, mostly because it's independent of `override` markings.

sql/item_cmpfunc.h
DaveGosselin-MariaDB: Agreed that these should be fixed but in a different patch.

sql/item_subselect.h
DaveGosselin-MariaDB: Agreed, will do as part of this patch.

storage/perfschema/ha_perfschema.h
```diff
@@ -86,25 +86,25 @@ class ha_perfschema final : public handler
     Operations supported by indexes.
     None, there are no indexes.
   */
-  ulong index_flags(uint , uint , bool ) const
+  ulong index_flags(uint , uint , bool ) const override
   { return 0; }
 
-  uint max_supported_record_length(void) const
+  uint max_supported_record_length(void) const override
```
dr-m, line 92: In C++, the empty parameter list should be written as `()` and not `(void)`. This would be easy to check with the following:
```sh
git show|grep '^+.*(void)'
```

storage/perfschema/pfs_engine_table.cc
```diff
@@ -343,13 +343,13 @@ static PFS_engine_table_share *all_shares[]=
 class PFS_silent_check_intact : public Table_check_intact
 {
 protected:
-  virtual void report_error(uint code, const char *fmt, ...) {}
+  void report_error(uint code, const char *fmt, ...) override {}
```
dr-m, line 346: `code` and `fmt` are unused.

storage/perfschema/pfs_engine_table.h
```diff
@@ -340,8 +340,8 @@ class PFS_readonly_acl : public ACL_internal_table_access
 
   ~PFS_readonly_acl() = default;
 
-  virtual ACL_internal_access_result check(privilege_t want_access,
-                                           privilege_t *save_priv) const;
+    ACL_internal_access_result check(privilege_t want_access,
+                                     privilege_t *save_priv) const override;
```
dr-m, line 344: It looks like this should be indented by 2 columns less.

storage/test_sql_discovery/test_sql_discovery.cc
```diff
@@ -64,35 +64,35 @@ class ha_tsd: public handler
...
            HA_CAN_INDEX_BLOBS | HA_AUTO_PART_KEY | HA_CAN_RTREEKEYS |
            HA_CAN_FULLTEXT;
   }
 
-  ulong index_flags(uint inx, uint part, bool all_parts) const { return 0; }
+  ulong index_flags(uint inx, uint part, bool all_parts) const override { return 0; }
 
   THR_LOCK_DATA **store_lock(THD *thd, THR_LOCK_DATA **to,
-                             enum thr_lock_type lock_type)
+                             enum thr_lock_type lock_type) override
```
dr-m, line 77: `thd` is unused.

storage/test_sql_discovery/test_sql_discovery.cc
```diff
@@ -64,35 +64,35 @@ class ha_tsd: public handler
...
-  ulonglong table_flags() const
+  ulonglong table_flags() const override
   { // NO_TRANSACTIONS and everything that affects CREATE TABLE
     return HA_NO_TRANSACTIONS | HA_CAN_GEOMETRY | HA_NULL_IN_KEY |
            HA_CAN_INDEX_BLOBS | HA_AUTO_PART_KEY | HA_CAN_RTREEKEYS |
            HA_CAN_FULLTEXT;
   }
 
-  ulong index_flags(uint inx, uint part, bool all_parts) const { return 0; }
+  ulong index_flags(uint inx, uint part, bool all_parts) const override { return 0; }
```
dr-m, line 74: All parameters are unused.

storage/perfschema/table_setup_instruments.h
```diff
@@ -97,21 +97,21 @@ class table_setup_instruments : public PFS_engine_table
...
+                      unsigned char *buf,
+                      Field **fields,
+                      bool read_all) override;
+    
+  int update_row_values(TABLE *table,
+                        const unsigned char *old_buf,
+                        const unsigned char *new_buf,
+                        Field **fields) override;
+    
   table_setup_instruments();
```
dr-m, line 115: Please do not add lines with trailing space.

storage/perfschema/table_setup_objects.h
```diff
@@ -70,24 +70,24 @@ class table_setup_objects : public PFS_engine_table
...
+  int read_row_values(TABLE *table,
+                      unsigned char *buf,
+                      Field **fields,
+                      bool read_all) override;
+
+  int update_row_values(TABLE *table,
+                        const unsigned char *old_buf,
+                        const unsigned char *new_buf,
+                        Field **fields) override;
+
```
dr-m, line 87: Trailing space should be easy to spot with
```sh
git show|grep '^+.*[	 ]$'
```

storage/sequence/sequence.cc
```diff
@@ -64,45 +64,45 @@ class ha_seq final : public handler
...
+  int rnd_next(unsigned char *buf) override;
+  void position(const uchar *record) override;
+  int rnd_pos(uchar *buf, uchar *pos) override;
+  int info(uint flag) override;
 
   /* indexes */
-  ulong index_flags(uint inx, uint part, bool all_parts) const
+  ulong index_flags(uint inx, uint part, bool all_parts) const override
   { return HA_READ_NEXT | HA_READ_PREV | HA_READ_ORDER |
            HA_READ_RANGE | HA_KEYREAD_ONLY; }
```
dr-m, line 93: All parameters are unused.

dr-m (approved):
Great work. I found some formatting nitpicks, but did not flag all of them.

I would suggest running a few more formatting checks:
```sh
git show|grep '^+.\{80\}'
git show|grep '^+.*unused'
```
You can ignore the long-line output for files that are already clearly violating it, such as `storage/sphinx`.

This does not need any additional review from me.

PR #3332, opened 2024-06-13
MDEV-34381 InnoDB fails to truncate during recovery

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34381*

## Description
- There is a possibility that InnoDB fails to truncate the undo tablespace when there are no pages to recover while applying the redo log. trx_lists_init_at_db_start() could've applied the redo logs for undo tablespace page0.

## How can this PR be tested?
 Need to write solid mtr test case.

## PR quality check


dr-m (approved):
Thank you, good catch. Can you please edit the MDEV title (and the commit message) to clarify that this is about the recovery of `innodb_undo_log_truncate=ON`?

PR #3334, opened 2024-06-14
MDEV-19052 Range-type window frame supports only numeric datatype

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-19052*

## Description

When there is no bounds on the upper or lower part of the window, it doesn't matter if the type is numeric.

## Release Notes

Window functions over unbounded ranges can now be non-numeric types.

## How can this PR be tested?

mtr tests included.

## PR quality check


LinuxJedi (changes requested):
Appears to break the result for `encryption.tempfiles_encrypted`. I'm assuming this imports `win.test`.

sql/sql_window.cc
```diff
@@ -275,9 +275,12 @@ setup_windows(THD *thd, Ref_ptr_array ref_pointer_array, TABLE_LIST *tables,
       /*
         "The declared type of SK shall be numeric, datetime, or interval"
         we don't support datetime or interval, yet.
+	But it really only matters if the frame is bounded.
       */
       Item_result rtype= win_spec->order_list->first->item[0]->result_type();
-      if (rtype != REAL_RESULT && rtype != INT_RESULT && 
+      if (!win_spec->window_frame->top_bound->is_unbounded() &&
+          !win_spec->window_frame->bottom_bound->is_unbounded() &&
```
Olernov, line 282: Do you want the restriction to be applied when only one frame is unbounded or both? If the latter, I guess the new condition must look like 
```
!(win_spec->window_frame->top_bound->is_unbounded() &&
          win_spec->window_frame->bottom_bound->is_unbounded())
```

Olernov, line 282: I think the new condition fits better to the `if` above:
```
if (win_spec->window_frame && 
        win_spec->window_frame->units == Window_frame::UNITS_RANGE)
```
(given that we only touch the case when both window frames are unbounded).
This will however break `win.test` , but probably it can be amended?

grooverdan: both, so done.

grooverdan: Done. By moving this there is now multiple ORDER BY items allowed. Added a test case to ensure that its ordering the window correctly.

spetrunia (commented):
The patch looks good.  We can support this special case.  It would be nice to support all kinds of range bounds but one can take that in steps. 

@grooverdan Do you really intend to get this into 10.5?  The patch looks safe but it doesn't look like a fix for a critical bug...

grooverdan:
> The patch looks good. We can support this special case.

As @Olernov found out, MySQL 8.0 and Postgres both support this syntax.

> It would be nice to support all kinds of range bounds but one can take that in steps.

Hoping so too.

> @grooverdan Do you really intend to get this into 10.5?

Yes.

> The patch looks safe but it doesn't look like a fix for a critical bug...

Yes, safety was my main criteria and thanks for confirming. It may not be critical in the sense of a crashing or wrong results but when its safe, I'd rather not have another MySQL or Postgres incompatibility to document, or for a migrating user to hit a case that worked in their former database but not MariaDB.

spetrunia (approved):

PR #3335, opened 2024-06-14
MDEV-33837: wsrep_sst_common.sh does not have shebang

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-33837*

## Description
As script scripts/wsrep_sst_common.sh is mainly meant to be sourced in bash/zsh script but there is also possiblity to run to independent also. It does not hurt to have shebang on script file.

Fixes lintian warning:
 * W: mariadb-server: executable-not-elf-or-script [usr/bin/wsrep_sst_common]

## Release Notes
Should not appear in Release notes

## How can this PR be tested?
Test is `usr/bin/wsrep_sst_common` work the same with or without `/bin/sh` shebang

## PR quality check

### DIFF
--- a/scripts/wsrep_sst_common.sh
+++ b/scripts/wsrep_sst_common.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 # Copyright (C) 2017-2022 MariaDB
 # Copyright (C) 2012-2015 Codership Oy
 #


grooverdan:
its sourced, its never directed executed.

tonychen2001 (approved):

illuusio:
> its sourced, its never directed executed.

Then there should probably be removed all the independed parts or at least move to something like `/usr/share/mariadb` probably?

grooverdan:
Need a more compatible plan. The files sourcing this file move also needs to be considered. The bash rather than sh might have a impact on BSD systems that I'm not prepared for.

PR #3337, opened 2024-06-14
MDEV-34389  Don't write FILE_CHECKPOINT during recovery when log file is insufficient

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34389*

## Description
- InnoDB tries to write FILE_CHECKPOINT marker when log file size is insufficient. This could corrupt the
log file and make it unrecoverable.

recv_recovery_From_checkpoint_start(): Before writing FILE_CHECKPOINT marker, check whether
we have sufficient log file capacity

## How can this PR be tested?
Pasted the test case in MDEV page.

## PR quality check

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3738,7 +3738,9 @@ completed:
 	if (!srv_read_only_mode
             && srv_operation <= SRV_OPERATION_EXPORT_RESTORED
 	    && (~log_t::FORMAT_ENCRYPTED & log_sys.log.format)
-	    == log_t::FORMAT_10_5) {
+	    == log_t::FORMAT_10_5
+	    && recv_sys.recovered_lsn - log_sys.last_checkpoint_lsn
+	    < log_sys.log_capacity) {
 		/* Write a FILE_CHECKPOINT marker as the first thing,
 		before generating any other redo log. This ensures
 		that subsequent crash recovery will be possible even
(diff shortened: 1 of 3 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -3738,7 +3738,9 @@ recv_recovery_from_checkpoint_start(lsn_t flush_lsn)
 	if (!srv_read_only_mode
             && srv_operation <= SRV_OPERATION_EXPORT_RESTORED
 	    && (~log_t::FORMAT_ENCRYPTED & log_sys.log.format)
-	    == log_t::FORMAT_10_5) {
+	    == log_t::FORMAT_10_5
+	    && recv_sys.recovered_lsn - log_sys.last_checkpoint_lsn
+	       < log_sys.log_capacity) {
```
dr-m, line 3743: These lines should be indented to the same level.

dr-m (commented):
I think that the bug title needs to emphasize "during early recovery" because we would be writing a checkpoint marker later on in `log_checkpoint_low()`. The point of this fix is that if the server is killed during recovery, we are narrowing down the window of opportunity that the log will not be recoverable. Also this should be explained in the commit message.

Can you write a test case for this? In [MDEV-34156](https://jira.mariadb.org/browse/MDEV-34156) I was able to reproduce this, but maybe not reliably.

dr-m (approved):

PR #3340, opened 2024-06-16
MDEV-34311: Alter USER should reset all account limit counters

cvicentiu:


sql/sql_acl.cc
```diff
@@ -14442,7 +14450,8 @@ bool acl_authenticate(THD *thd, uint com_change_user_pkt_len)
     case CR_AUTH_USER_CREDENTIALS:
       errors.m_authentication= 1;
       if (thd->password && !mpvio.make_it_fail)
-        handle_password_errors(acl_user->user.str, acl_user->host.hostname, PASSWD_ERROR_INCREMENT);
+        handle_password_errors(acl_user->user.str, acl_user->host.hostname,
```
grooverdan, line 14453: Though was a problem with the previous implementation, we've got a `acl_user` (`const` needs dropping),  and we're taking its user/host to look the user up again in `handle_password_errors`? Seems wasteful. Can `handle_password_errors` and `update_password_errors` be dropped and a `clear_password_errors` / `increment_password_errors` exist in the `AC_USER` class.

I'm assuming the `acl_cache->lock` was related to the `find_user_exact` function, but if its for the increment/clear itself, perhaps `password_errors` should be a `Atomic_counter`.

grooverdan (changes requested):

tonychen2001:
Just want to say that this makes definitely makes sense from a user point of view.

I also agree with comments from grooverdan:
>we've got a acl_user (const needs dropping), and we're taking its user/host to look the user up again in handle_password_errors

But as mentioned this is simply adhering to the current implementation.

cvicentiu:
@grooverdan @tonychen2001 Thank you for the reviews.

The MPVIO_EXT::ACL_USER is never the same object as the one stored in `acl_users`. That is why one can use the `acl_user` without having taken the acl_cache lock. It is created via ACL_USER::copy. This is (as far as I can understand from the code) a way to prevent security attacks by having the server do the exact same work for an existing user as well as for a non-existing user. See the "make_it_fail" flag.

That is why we need to search for the actual user in acl_users if we are to modify the password attempts counter for it.



When one must modify some value in an ACL_USER that must e persistent for

cvicentiu:
I got a review from @montywi for this code, with only an error message suggestion change that I incorporated. The suggestion was to say ALTER USER / FLUSH PRIVILEGES, in this order.

sql/sql_acl.cc
cvicentiu: acl_cache->lock guards all acl_users and some other related data structures. One must hold the lock anyway for the small period of time need to find and increment the counter. In the future we could improve this, but not in the scope of this bug fix.

Thank you for looking into this.

PR #3341, opened 2024-06-18
MDEV-34392 Inplace algorithm violates the foreign key constraint

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34392*

## Description
MDEV-34392 Inplace algorithm violates the foreign key constraint
    
    Changing the child foreign key column from NULL TO NOT NULL
    should be avoided if
    
    1) Foreign key constraint is ON UPDATE SET NULL
    2) Foreign key constraint is ON DELETE SET NULL
    3) Foreign key constraint is UPDATE CASCADE and parent referenced
    column allows NULL values
    
    Changing the parent foreign key column from NOT NULL to NULL
    should be avoided if the foreign key constraint is UPDATE CASCADE
    and child foreign columns doesn't allow NULL values
    
    get_foreign_key_info(): Adds the additional information that
    parent referenced columns exists or accepts NULL/NON_NULL values
    
    fk_check_column_changes(): Enforce the above rules for COPY
    algorithm
    
    innobase_check_foreign_drop_col(): Checks whether the dropped
    column exists in existing foreign key relation
    
    innobase_check_foreign_low() : Enforce the above rules for
    INPLACE algorithm

## How can this PR be tested?
./mtr innodb.fk_col_non_null

## PR quality check


mysql-test/suite/innodb/r/fk_col_alter.result
```diff
@@ -132,3 +132,26 @@ ADD CONSTRAINT t2_fk FOREIGN KEY(f3) REFERENCES t1(f1);
...
+SET FOREIGN_KEY_CHECKS=1;
+ALTER TABLE t2 MODIFY f2 INT UNSIGNED NOT NULL, ALGORITHM=COPY;
+ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 'fk_id'
+ALTER TABLE t2 MODIFY f2 INT UNSIGNED NOT NULL, ALGORITHM=INPLACE;
+ERROR HY000: Column 'f2' cannot be NOT NULL: needed in a foreign key constraint 'test/fk_id' SET NULL
+SET FOREIGN_KEY_CHECKS=0;
+ALTER TABLE t2 MODIFY f2 INT UNSIGNED NOT NULL, ALGORITHM=COPY;
+ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 'fk_id'
+ALTER TABLE t2 MODIFY f2 INT UNSIGNED NOT NULL, ALGORITHM=INPLACE;
+ERROR HY000: Column 'f2' cannot be NOT NULL: needed in a foreign key constraint 'test/fk_id' SET NULL
```
dr-m, line 152: There is no `ON DELETE SET NULL` or `ON UPDATE SET NULL` clause in `CONSTRAINT fk_id FOREIGN KEY (f2) REFERENCES t1(f1)`.

dr-m (changes requested):
I do not see why we should forbid the addition or removal of a `NOT NULL` constraint when there is no `FOREIGN KEY` constraint that could cause the column to be updated to `NULL`. Also `ON UPDATE CASCADE` can cause that, unless the parent column is declared `NOT NULL`.

Related to that, it would be good to add another test that attempts to remove a `NOT NULL` constraint from a parent table column that is being referenced by a column in a child table, with `NOT NULL REFERENCES parent(col) ON UPDATE CASCADE`.

storage/innobase/include/dict0mem.h
```diff
@@ -1490,6 +1490,26 @@ struct dict_foreign_t{
...
+
+	bool col_exists_in_reference(const char *name) const
+	{
+          for (unsigned f = 0; f < n_fields; f++)
+          {
+            if (!strcmp(referenced_col_names[f], name))
+              return true;
+	  }
+	  return false;
+	}
```
dr-m, line 1512: This is using some common code, which I think would better be in a non-inlined function, to avoid object code duplication. Something like this should be optimal also from the ABI point of view:
```c++
private:
  bool col_exists(const char *name, const char **names) const;
public:
  bool col_exists_in_foreign(const char *name) const
  { return col_exists(name, foreign_col_names); }
  bool col_exists_in_referenced(const char *name) const
  { return col_exists(name, referenced_col_names); }
```
Alternatively, make the non-inline function `static` and add `n_fields` as its first parameter.

sql/table.h
```diff
@@ -1870,6 +1870,9 @@ enum enum_schema_table_state
 enum enum_fk_option { FK_OPTION_UNDEF, FK_OPTION_RESTRICT, FK_OPTION_NO_ACTION,
   FK_OPTION_CASCADE, FK_OPTION_SET_NULL, FK_OPTION_SET_DEFAULT };
 
+enum enum_fk_relation { FIELDS_NOT_EXIST, FIELDS_NULL,
+  FIELDS_NOT_NULL };
```
dr-m, line 1874: These values had better be documented with `/** */` comments. What exactly do `FIELDS_NULL` and `FIELDS_NOT_NULL` mean if there are multiple columns and only some of them allow `NULL` values, some not?

mysql-test/suite/innodb/t/fk_col_non_null.combinations
```diff
@@ -0,0 +1,5 @@
+[COPY]
+--alter_algorithm=copy
+
+[INPLACE]
+--alter_algorithm=inplace
```
dr-m, line 5: I think that there were plans to remove this parameter. It could be better to examine `MTR_COMBINATION_COPY` and `MTR_COMBINATION_INPLACE` in the `.test` file and not specify any parameter here.

A better name for the test would be `foreign_not_null` or `foreign_null`, to better match the already is a SQL phrases `NOT NULL` or `ON…SET NULL`. The `null` already implies that this is about columns.

sql/table.h
```diff
@@ -1870,6 +1870,14 @@ enum enum_schema_table_state
 enum enum_fk_option { FK_OPTION_UNDEF, FK_OPTION_RESTRICT, FK_OPTION_NO_ACTION,
   FK_OPTION_CASCADE, FK_OPTION_SET_NULL, FK_OPTION_SET_DEFAULT };
 
+enum enum_fk_nullablity
+{
+  /** Foreign table or Referenced table doesn't exist */
+  NOT_EXIST,
+  ALLOWS_NULL,
+  ALLOWS_NOT_NULL
```
dr-m, line 1878: The literals `ALLOWS_NULL` and `ALLOWS_NOT_NULL` need to be documented. In particular, I want to know what would be the values when there is a `CONSTRAINT FOREIGN KEY (a_null,b_not_null) REFERENCES parent(c_not_null,d_null)`.

The logic of `get_foreign_key_info()` looks like `ALLOWS_NULL` will take priority, that is, we would only set `ALLOWS_NOT_NULL` if a suitable index is found and `ALLOWS_NULL` was not set yet. It would be more descriptive to use the name `INDEX_EXISTS` instead of `ALLOWS_NOT_NULL`.

storage/innobase/handler/handler0alter.cc
```diff
@@ -4211,103 +4211,179 @@ innobase_dropping_foreign(
...
-	dict_foreign_t**	drop_fk,
-	ulint			n_drop_fk,
-	const char*		col_name,
-	bool			drop)
+bool innobase_check_foreigns_nullability(
+  const dict_table_t *user_table,
+  dict_foreign_t     **drop_fk,
+  ulint              n_drop_fk,
+  const char*        col_name,
+  uint32_t           new_field_flags)
```
dr-m, line 4258: The changed formatting is neither according to the old InnoDB style nor according to the common style.

storage/innobase/handler/handler0alter.cc
```diff
@@ -4211,103 +4211,179 @@ innobase_dropping_foreign(
...
+  ut_ad(mutex_own(&dict_sys.mutex));
 
-		for (unsigned f = 0; f < foreign->n_fields; f++) {
-			char display_name[FN_REFLEN];
+  /* Check if any FOREIGN KEY constraints are defined on this column. */
+  for (dict_foreign_set::const_iterator it=
+         user_table->foreign_set.begin();
+       it != user_table->foreign_set.end(); ++it)
+  {
+    dict_foreign_t *foreign= *it;
```
dr-m, line 4347: We are not using `it` anywhere, so it would be better to write
```c++
  for (dict_foreign_t *foreign : user_table->foreign_set)
```

storage/innobase/handler/handler0alter.cc
```diff
@@ -4211,103 +4211,179 @@ innobase_dropping_foreign(
...
-			my_error(ER_FK_COLUMN_CANNOT_DROP_CHILD,
-				 MYF(0), col_name, foreign->id,
-				 display_name);
+  /* Check if any FOREIGN KEY constraints in other tables are
+  referring to the column that is being dropped. */
+  for (dict_foreign_set::const_iterator it=
+         user_table->referenced_set.begin();
+       it != user_table->referenced_set.end(); ++it)
+  {
+    dict_foreign_t *foreign= *it;
```
dr-m, line 4366:
```c++
  for (dict_foreign_t *foreign : user_table->referenced_set)
```

storage/innobase/handler/handler0alter.cc
```diff
@@ -4342,10 +4418,21 @@ innobase_check_foreigns(
...
+			if (innobase_check_foreign_drop_col(
+				user_table, drop_fk, n_drop_fk,
+				(*fp)->field_name.str)) {
+				return true;
+			}
+			continue;
+		}
+
+
+		if ((it->flags & NOT_NULL_FLAG)
```
dr-m, line 4431: You could remove the `continue` and add `else` before the `if` to make this slightly more readable.

mysql-test/suite/innodb/r/foreign_null,INPLACE.rdiff
```diff
@@ -0,0 +1,32 @@
+--- foreign_null.result	2024-06-26 02:26:06.830377918 +0530
++++ foreign_null.reject	2024-06-26 02:26:12.838781149 +0530
```
dr-m, line 2: Please do not add any timestamps.
```
--- foreign_null.result
+++ foreign_null.result,INPLACE
```

sql/sql_table.cc
```diff
@@ -9628,17 +9632,48 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
...
         /*
           Column in a FK has changed significantly and it
           may break referential intergrity.
         */
+data_change:
         *bad_column_name= column->str;
         return FK_COLUMN_DATA_CHANGE;
       }
+      else if ((new_field->flags & NOT_NULL_FLAG) !=
+               (old_field->flags & NOT_NULL_FLAG))
```
dr-m, line 9646: We can simplify the arithmetic:
```c++
      else if ((new_field->flags ^ old_field->flags) & NOT_NULL_FLAG)
```

sql/sql_table.cc
```diff
@@ -9628,17 +9632,48 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
...
+            NOT NULL to NULL when foreign key relation is
+            UPDATE CASCADE and foreign key column doesn't allow NULL */
+            if (fk->update_method == FK_OPTION_CASCADE
+                && fk->foreign_fields_nullable == ALLOWS_NOT_NULL)
+              goto data_change;
+          }
+        }
+        else
+        {
+          if (new_field->flags & NOT_NULL_FLAG)
```
dr-m, line 9662: This can be written as `else if`.

sql/sql_table.cc
```diff
@@ -9628,17 +9632,48 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
...
+          {
+            /* Don't allow the foreign column to change
+            from NULL to NOT NULL when foreign key type is
+            1) UPDATE SET NULL
+            2) DELETE SET NULL
+            3) UPDATE CASCADE and referenced column allows NULL */
+            if (fk->update_method == FK_OPTION_SET_NULL
+                || fk->delete_method == FK_OPTION_SET_NULL
+                || (fk->update_method == FK_OPTION_CASCADE
+                    && fk->referenced_fields_nullable == ALLOWS_NULL))
```
dr-m, line 9672: Lines should be split after binary operators, not before them.

sql/sql_table.cc
```diff
@@ -9628,17 +9632,48 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
...
+      else if ((new_field->flags & NOT_NULL_FLAG) !=
+               (old_field->flags & NOT_NULL_FLAG))
+      {
+        if (referenced)
+        {
+          if (!(new_field->flags & NOT_NULL_FLAG))
+          {
+            /* Don't allow referenced column to change from
+            NOT NULL to NULL when foreign key relation is
+            UPDATE CASCADE and foreign key column doesn't allow NULL */
```
dr-m, line 9654: The phrase "foreign key column" can refer to "referenced column" or  "referencing column". Let’s make it easier for the reader and end this with "ON UPDATE CASCADE and the referencing column is declared NOT NULL" .

mysql-test/suite/innodb/r/foreign_null,INPLACE.rdiff
```diff
@@ -0,0 +1,32 @@
...
++ERROR HY000: Column 'f2' cannot be NOT NULL: needed in a foreign key constraint 'test/t2_ibfk_2' SET NULL
++ALTER TABLE t2 MODIFY COLUMN f1 INT NOT NULL, ALGORITHM=INPLACE;
++ERROR HY000: Column 'f1' cannot be NOT NULL: needed in a foreign key constraint 'test/t2_ibfk_1' SET NULL
+ DROP TABLE t2, t1;
+ CREATE TABLE t1(f1 INT, f2 INT NOT NULL, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+ CREATE TABLE t2(f1 INT NOT NULL, FOREIGN KEY(f1) REFERENCES t1(f2) ON UPDATE CASCADE)ENGINE=InnoDB;
+-ALTER TABLE t1 MODIFY COLUMN f2 INT, ALGORITHM=COPY;
+-ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 't2_ibfk_1' of table 'test.t2'
++ALTER TABLE t1 MODIFY COLUMN f2 INT, ALGORITHM=INPLACE;
++ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 'test/t2_ibfk_1' of table 'test/t1'
```
dr-m, line 29: The `test/` prefix needs to be omitted in the `INPLACE` case, to have more consistent reporting.

mysql-test/suite/innodb/t/foreign_null.test
```diff
@@ -0,0 +1,55 @@
...
+
+if ($algorithm == "INPLACE") {
+ let $error_code = ER_FK_COLUMN_NOT_NULL;
+}
+
+# MODIFY CHILD COLUMN..NOT NULL ON UPDATE CASCADE.. PARENT COLUMN NULL
+CREATE TABLE t1(f1 INT, f2 INT, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT, FOREIGN KEY(f1) REFERENCES t1(f2) ON UPDATE CASCADE)ENGINE=InnoDB;
+--error $error_code
+eval ALTER TABLE t2 MODIFY COLUMN f1 INT NOT NULL, algorithm=$algorithm;
```
dr-m, line 19: Can we use `replace_result` to remove the `algorithm=` difference from the `.rdiff` file? It would make the output a little easier to read.

sql/sql_table.cc
```diff
@@ -9588,9 +9588,13 @@ enum fk_column_change_type
...
 fk_check_column_changes(THD *thd, Alter_info *alter_info,
-                        List<LEX_CSTRING> &fk_columns,
-                        const char **bad_column_name)
+                        FOREIGN_KEY_INFO *fk,
+                        const char **bad_column_name,
+                        bool referenced=false)
 {
+  List<LEX_CSTRING> fk_columns= referenced
+                                ? fk->referenced_fields
+                                : fk->foreign_fields;
```
dr-m, line 9597: I believe that the ternary operator should be indented by exactly 2 additional spaces (in total, 4 columns from the start).

storage/innobase/handler/handler0alter.cc
```diff
@@ -4211,103 +4211,179 @@ innobase_dropping_foreign(
...
+  goto func_exit;
+ref_field_check:
+  for (dict_foreign_t *foreign : user_table->referenced_set)
+  {
+    if ((foreign->type & DICT_FOREIGN_ON_UPDATE_CASCADE)
+        && foreign->foreign_index)
+    {
+      for (unsigned f = 0;f < foreign->n_fields; f++)
+      {
+        dict_field_t* field= dict_index_get_nth_field(
```
dr-m, line 4306: Only in the archaic InnoDB style are we allowed to break a line right after `(`.

storage/innobase/handler/handler0alter.cc
```diff
@@ -4211,103 +4211,179 @@ innobase_dropping_foreign(
...
-
-	for (dict_foreign_set::const_iterator it = user_table->foreign_set.begin();
-	     it != user_table->foreign_set.end();
-	     ++it) {
+  ut_ad(mutex_own(&dict_sys.mutex));
 
-		foreign = *it;
+  /* Changing from NOT NULL to NULL. So check referenced set */
+  if (!(new_field_flags & NOT_NULL_FLAG))
+    goto ref_field_check;
```
dr-m, line 4264: We could remove the `goto` by inverting the condition and adding `{`. With 2 spaces per indentation level (instead of 8-space TAB) there should be no need for such tricks.

storage/innobase/handler/handler0alter.cc
```diff
@@ -4211,103 +4211,179 @@ innobase_dropping_foreign(
...
+        foreign->referenced_index)
+    {
+      for (unsigned f = 0;f < foreign->n_fields; f++)
+      {
+        dict_field_t* field= dict_index_get_nth_field(
+                               foreign->referenced_index, f);
+        if (!(field->col->prtype & DATA_NOT_NULL))
+          goto col_exist_check;
+      }
+    }
```
dr-m, line 4281: Between the two `}` it would be helpful to add a comment
```c++
/* All referenced columns are NOT NULL */
```
But, wouldn’t be more readable to write `foreign->referenced_index->get_n_nullable(foreign->n_fields)`? Yes, that would skip the short-circuit `goto`, but this is not time critical code.

storage/innobase/handler/handler0alter.cc
```diff
@@ -4211,103 +4211,179 @@ innobase_dropping_foreign(
...
-	}
+    if (foreign->type & DICT_FOREIGN_ON_DELETE_SET_NULL
+        || foreign->type & DICT_FOREIGN_ON_UPDATE_SET_NULL)
+    {
+col_exist_check:
+      if (col_exists(foreign->n_fields, col_name,
+                     foreign->foreign_col_names))
+      {
+        my_error(ER_FK_COLUMN_NOT_NULL, MYF(0), col_name,
+                 foreign->id);
```
dr-m, line 4291: This is slightly out of sync with `get_foreign_key_info()`, because here we are comparing column names even if no underlying index exists. Can we refer to `foreign->foreign_index` here to make it consistent?

dr-m (changes requested):

sql/table.h
dr-m: I thought more about this, and I think that we need to match each column. This might require ‘pushing down’ the check to InnoDB.
Consider the valid case `CONSTRAINT FOREIGN KEY (a_null,b_not_null) REFERENCES parent(c_null,d_not_null) ON UPDATE CASCADE`. As far as I understand, with the current logic this would be disallowed due to `ALLOWS_NULL` being reported for `parent.c_null`. But, there is no actual problem because `a_null` does allow `NULL` values.

I think that the test case must be expanded with 2-column `FOREIGN KEY` constraints like this, covering all 4 combinations of `NULL` and `NOT NULL` for the 2+2 columns.

dr-m: I think that @FooBarrior would be a good candidate for reviewing the changes outside InnoDB, related to this.

mysql-test/suite/innodb/t/foreign_null.test
```diff
@@ -0,0 +1,110 @@
...
+let $algorithm=`select regexp_replace('$MTR_COMBINATIONS', 'innodb,\|,innodb', '')`;
+let $error_code = 0;
+
+if ($algorithm == "COPY") {
+ let $error_code = ER_FK_COLUMN_CANNOT_CHANGE;
+}
+
+if ($algorithm == "INPLACE") {
+ let $error_code = ER_FK_COLUMN_NOT_NULL;
+}
```
mariadb-DebarunBanerjee, line 11: Ideally it is better experience for user if he gets same error for COPY and INPLACE algorithm. Do you think we can have ER_FK_COLUMN_CANNOT_CHANGE for both ? If needed we can also raise ER_FK_COLUMN_NOT_NULL for inplace which should be stacked and shown with "show errors".

mysql-test/suite/innodb/t/foreign_null.test
```diff
@@ -0,0 +1,110 @@
...
+
+# MODIFY CHILD COLUMN..NOT NULL ON UPDATE CASCADE.. PARENT COLUMN NULL
+CREATE TABLE t1(f1 INT, f2 INT, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT, FOREIGN KEY(f1) REFERENCES t1(f2) ON UPDATE CASCADE)ENGINE=InnoDB;
+replace_result ,ALGORITHM=COPY '' ,ALGORITHM=INPLACE '';
+--error $error_code
+eval ALTER TABLE t2 MODIFY COLUMN f1 INT NOT NULL,ALGORITHM=$algorithm;
```
mariadb-DebarunBanerjee, line 18: 1. The comment is actually useful in result file. I would also suggest changing NULL to NULLABLE or NO CONSTRAINT for parent table in the comment.
`--echo Test-1: MODIFY CHILD COLUMN..NOT NULL ON UPDATE CASCADE.. PARENT COLUMN NULLABLE`
2. Can we add a case for ON DELETE CASCADE to make sure we are allowing it to pass ?

mysql-test/suite/innodb/t/foreign_null.test
```diff
@@ -0,0 +1,110 @@
...
+DROP TABLE t2, t1;
+
+# MODIFY CHILD COLUMN..NOT NULL ON DELETE|ON UPDATE SET NULL
+CREATE TABLE t1(f1 INT, f2 INT, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT, f2 INT, FOREIGN KEY(f1) REFERENCES t1(f1) ON UPDATE SET NULL, FOREIGN KEY (f2) REFERENCES t1(f2) ON DELETE SET NULL)ENGINE=InnoDB;
+replace_result ,ALGORITHM=COPY '' ,ALGORITHM=INPLACE '';
+--error $error_code
+eval ALTER TABLE t2 MODIFY COLUMN f2 INT NOT NULL,ALGORITHM=$algorithm;
```
mariadb-DebarunBanerjee, line 26: I would suggest to separate out ON DELETE and ON UPDATE in 2 test cases.

mysql-test/suite/innodb/t/foreign_null.test
```diff
@@ -0,0 +1,110 @@
...
+
+replace_result ,ALGORITHM=COPY '' ,ALGORITHM=INPLACE '';
+--error $error_code
+eval ALTER TABLE t2 MODIFY COLUMN f1 INT NOT NULL,ALGORITHM=$algorithm;
+DROP TABLE t2, t1;
+
```
mariadb-DebarunBanerjee, line 32: Let's also add the positive case which should work with both algorithm that the filer actually complained about.
`--echo Test-3: MODIFY CHILD COLUMN..NOT NULL /* ON UPDATE RESTRICT */.. PARENT COLUMN NULLABLE`

mysql-test/suite/innodb/t/foreign_null.test
```diff
@@ -0,0 +1,110 @@
...
+if ($algorithm == "COPY") {
+ let $error_code = ER_FK_COLUMN_CANNOT_CHANGE;
+}
+
+if ($algorithm == "INPLACE") {
+ let $error_code = ER_FK_COLUMN_NOT_NULL;
+}
+
+CREATE TABLE t1(f1 INT, f2 INT, INDEX(f1, f2))ENGINE=InnoDB;
```
mariadb-DebarunBanerjee, line 74: We are back to testing the child table CREATE/ALTER. Can you please add a comment for each test similar to the tests in the beginning.

mariadb-DebarunBanerjee (commented):
Thiru,
Please find my comments on the FK test. I will check the code soon.

FooBarrior (approved):

sql/sql_table.cc
```diff
@@ -9651,12 +9697,15 @@ fk_check_column_changes(THD *thd, Alter_info *alter_info,
...
+      result= FK_COLUMN_DROPPED;
+      goto func_exit;
     }
+    n_col++;
   }
-
-  return FK_COLUMN_NO_CHANGE;
+func_exit:
+  if (result != FK_COLUMN_NO_CHANGE)
+    *bad_column_name= column->str;
```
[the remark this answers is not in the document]
FooBarrior: When result is not `FK_COLUMN_NO_CHANGE`, it represents different types of changes. `bad_column_name` has to be set only when the change is found.

So overall the way it's written now is fine for `sql`.

Lets avoid auxiliary things like `if (false)`:
```
  return FK_COLUMN_NO_CHANGE;
err_exit:
  *bad_column_name= column->str; 
  return result;
```

Also I'd suggest to use other flow control techniques, than goto: if-else, function extraction, or break/continue on the poor end. The choice is up to the author though, goto is not forbidden.

mariadb-DebarunBanerjee (commented):
@Thirunarayanan, Some quick points on the Innodb changes. I will need some more time to complete it.

Since the change is reaching close to 1 KLOC, we should have a good amount of testing for this change. I would also suggest to check that added lines are all covered by the existing and added test.

mysql-test/suite/innodb/t/foreign_null.test
```diff
@@ -0,0 +1,167 @@
+--source include/have_innodb.inc
+let $algorithm=`select regexp_replace('$MTR_COMBINATIONS', 'innodb,\|,innodb', '')`;
+let $error_code = 0;
+let $error_code = ER_FK_COLUMN_NOT_NULL;
```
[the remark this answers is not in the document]
Thirunarayanan: Removed all error code from test case and added error code directly

mysql-test/suite/innodb/r/foreign_null,INPLACE.rdiff
```diff
@@ -0,0 +1,20 @@
+--- foreign_null.result
++++ foreign_null,INPLACE.result
+@@ -45,7 +45,7 @@
+ CREATE TABLE t1(f1 INT, f2 INT NOT NULL, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+ CREATE TABLE t2(f1 INT NOT NULL, FOREIGN KEY(f1) REFERENCES t1(f2) ON UPDATE CASCADE)ENGINE=InnoDB;
+ ALTER TABLE t1 MODIFY COLUMN f2 INT;
+-ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 't2_ibfk_1' of table 'test.t2'
++ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 't2_ibfk_1' of table '`test`.`t2`'
```
dr-m, line 8: I don’t think that this file is allowed to exist. The error message output needs to be identical in both cases. See 024a18dbcb7f9c44094db03daa013fd8ba3c58b7 how the table name can be reported properly.

sql/sql_table.cc
```diff
@@ -9611,17 +9612,25 @@ enum fk_column_change_type
                                  change in foreign key column.
   @retval FK_COLUMN_RENAMED      Foreign key column is renamed.
   @retval FK_COLUMN_DROPPED      Foreign key column is dropped.
+  @retval FK_COLUMN_NOT_NULL     Foreign key column cannot be null.
```
dr-m, line 9615: The comment is a little confusing. Should it be: "ON...SET NULL or ON UPDATE CASCADE conflicts with NOT NULL"?

I see that there are some tests cases that cover the following type of scenario:
```sql
create table t1 (f1 int null unique) engine=innodb;
create table t2 (f1 int primary key,
    constraint c1 foreign key (f1) references t1(f1)
    on update cascade) engine=innodb;
insert into t1 set f1=0;
insert into t2 set f1=0;
update t1 set f1=null;
```
Here, if we did not refuse to create such a child table, the `update` should lead to a transaction rollback.

sql/sql_table.cc
```diff
@@ -9611,17 +9612,25 @@ enum fk_column_change_type
...
 
 static enum fk_column_change_type
-fk_check_column_changes(THD *thd, Alter_info *alter_info,
-                        List<LEX_CSTRING> &fk_columns,
-                        const char **bad_column_name)
+fk_check_column_changes(THD *thd, const TABLE *table,
+                        Alter_info *alter_info,
+                        const FOREIGN_KEY_INFO *fk,
+                        const char **bad_column_name,
+                        bool referenced=false)
```
dr-m, line 9623: No `@param referenced` was added to the preceding comment.

sql/sql_table.cc
```diff
@@ -9611,17 +9612,25 @@ enum fk_column_change_type
...
 {
+  List<LEX_CSTRING> fk_columns= referenced
+      ? fk->referenced_fields
+      : fk->foreign_fields;
```
dr-m, line 9627: Could this be as follows:
```c++
  const List<LEX_CSTRING> &fk_columns= referenced
    ? fk->referenced_fields
    : fk->foreign_fields;
```
That is, do not copy the list, but have a `const` reference to it.

mysql-test/suite/innodb/r/foreign_null.result
```diff
@@ -0,0 +1,156 @@
...
+# modify child column NOT NULL ON DELETE NO ACTION..PARENT COLUMN NULL
+CREATE TABLE t1(f1 INT, f2 INT, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT, f2 INT, FOREIGN KEY (f2) REFERENCES t1(f2) ON DELETE NO ACTION)ENGINE=InnoDB;
+ALTER TABLE t2 MODIFY COLUMN f2 INT NOT NULL;
+DROP TABLE t2, t1;
+# modify parent column NULL ON UPDATE CASCADE child column NOT NULL
+CREATE TABLE t1(f1 INT, f2 INT NOT NULL, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT NOT NULL, FOREIGN KEY(f1) REFERENCES t1(f2) ON UPDATE CASCADE)ENGINE=InnoDB;
+ALTER TABLE t1 MODIFY COLUMN f2 INT;
+ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 't2_ibfk_1' of table 'test.t2'
```
dr-m, line 49: I don’t see an immediate problem if the child column allows `NULL` values and the parent column is `NOT NULL`. Yes, the foreign key constraint would prevent a `NULL` from being inserted into the child table.

In fact I think that that this needs to be allowed. Someone could want to change the column from `NOT NULL` to `NULL` in both the child and the parent table. The logical way to do that is to first alter the child table and then the parent table.

mysql-test/suite/innodb/r/foreign_null.result
```diff
@@ -0,0 +1,156 @@
...
+DROP TABLE t2, t1;
+# modify parent column NULL ON DELETE CASCADE child column NOT NULL
+CREATE TABLE t1(f1 INT, f2 INT NOT NULL, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT NOT NULL, FOREIGN KEY(f1) REFERENCES t1(f2) ON DELETE CASCADE)ENGINE=InnoDB;
+ALTER TABLE t1 MODIFY COLUMN f2 INT;
```
dr-m, line 54: It is logical for this one to be accepted. I would suggest to add a DML test as well, something like:
```sql
INSERT INTO t1 VALUES(1,1);
INSERT INTO t2 VALUES(1);
--error
UPDATE t1 SET f2=NULL;
SELECT * FROM t1;
SELECT * FROM t2;
```

sql/sql_table.cc
```diff
@@ -9811,6 +9862,7 @@ static bool fk_prepare_copy_alter_table(THD *thd, TABLE *table,
                f_key->foreign_id->str, buff.c_ptr());
       DBUG_RETURN(true);
     }
+    case FK_COLUMN_NOT_NULL:
     default:
       DBUG_ASSERT(0);
```
dr-m, line 9867: It would be good to add a source code comment that explains why this value is not possible here.

sql/table.h
```diff
@@ -1882,6 +1883,53 @@ typedef struct st_foreign_key_info
...
+    @param field_no field number
+    @param foreign  set foreign field
+  */
+  void set_nullable(bool referenced, unsigned field)
+  {
+    unsigned n_field= get_n_fields();
+    DBUG_ASSERT(field < n_field);
+    size_t bit= size_t{field} + referenced * n_field;
+    fields_nullable[bit]= (unsigned char) 1;
+  }
```
dr-m, line 1916: `assign_nullable()` allocates an array of `char` of two bits per column, but `set_nullable()` is treating it as 1 byte per column. I think that this could lead to a buffer overflow when there are many columns.

I’d suggest to allocate an array of `size_t` and accessing it something like this:
```c++
    fields_nullable[bit / (CHAR_BIT * sizeof(size_t))]|=
      size_t{1} << (bit % (CHAR_BIT * sizeof(size_t));
```

storage/innobase/handler/handler0alter.cc
```diff
@@ -4220,103 +4220,145 @@ innobase_dropping_foreign(
...
+  {
+    for (dict_foreign_t *foreign : user_table->referenced_set)
+    {
+      if ((foreign->type & foreign->UPDATE_CASCADE)
+          && foreign->check_fk_null_referenced(col_name))
+      {
+        char display_name[FN_REFLEN];
+	const int dblen=
+          int(table_name_t(const_cast<char*>(
+                             foreign->foreign_table_name)).dblen());
```
dr-m, line 4291: In the common formatting style, lines must not be split immediately before a binary operator, such as `&&`, or immediately after `(`.

storage/innobase/handler/handler0alter.cc
```diff
@@ -4220,103 +4220,145 @@ innobase_dropping_foreign(
...
+	sprintf(display_name, "%.*s.%s",
+                dblen, foreign->foreign_table_name,
+		foreign->foreign_table_name + dblen + 1);
```
dr-m, line 4294: We depend on C99, so we had better use `snprintf`.

storage/innobase/include/dict0mem.h
```diff
@@ -1451,45 +1451,149 @@ typedef std::set<dict_v_col_t*, std::less<dict_v_col_t*>,
...
+  }
+
+public:
+  /** Check whether the name exists in the foreign key column names
+  @retval offset in case of success
+  @retval UINT_MAX in case of failure */
+  unsigned col_fk_exists(const char *name) const
+  {
+    return col_exists(name, foreign_col_names);
+  }
```
dr-m, line 1523: Let’s add `noexcept` to any member function that can’t throw exceptions (such as `std::bad_alloc` from C++ heap memory allocations). No member functions here should do that.

mysql-test/suite/innodb/r/foreign_null.result
dr-m: Sorry, I had overlooked the `ON UPDATE CASCADE`. This one is definitely behaving correctly.

If there is not yet this kind of a case that uses something else (say, `ON DELETE CASCADE` or no `ON` clause at all), please add those variants.

mysql-test/suite/innodb/r/foreign_null.result
Thirunarayanan: Reason to accept this one because of  `ON DELETE CASCADE` . Don't know how this test case applicable to this scenario.

mysql-test/suite/innodb/r/foreign_null.result
Thirunarayanan: This variant is already added as a part of the test case.

dr-m (commented):
Thank you, this is close to being approved.

I checked the code coverage. The changed code is mostly covered by
```sh
./mtr innodb.innodb innodb.innodb-index innodb.foreign_null
```
Can you change the table names in the new tests to be `t$1` and `t$2` instead of `t1` and `t2`? That should exercise the pretty-printing logic.

storage/innobase/handler/handler0alter.cc
```diff
@@ -4220,103 +4220,144 @@ innobase_dropping_foreign(
...
+        char display_name[FN_REFLEN];
+	const int dblen=  int(table_name_t(const_cast<char*>(foreign->
+                                           foreign_table_name)).dblen());
+	snprintf(display_name, FN_REFLEN - 1, "%.*s.%s",
+                 dblen, foreign->foreign_table_name,
+		 foreign->foreign_table_name + dblen + 1);
+        const char* fid = strchr(foreign->id, '/');
+        fid= fid ? fid + 1 : foreign->id;
+        my_error(ER_FK_COLUMN_CANNOT_CHANGE_CHILD, MYF(0), col_name,
+                 fid, display_name);
```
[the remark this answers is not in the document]
Thirunarayanan:
```CREATE TABLE t$1(f1 INT, f2 INT NOT NULL, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
CREATE TABLE t$2(f1 INT NOT NULL, FOREIGN KEY(f1) REFERENCES t$1(f2) ON UPDATE CASCADE)ENGINE=InnoDB;
SET FOREIGN_KEY_CHECKS=0;
ALTER TABLE t$2 DROP INDEX f1,ALGORITHM=COPY;
SET FOREIGN_KEY_CHECKS=1;
ALTER TABLE t$1 MODIFY COLUMN f2 INT,ALGORITHM=COPY;
innodb.f1                                [ fail ]
        Test ended at 2024-09-27 21:45:05

CURRENT_TEST: innodb.f1
mysqltest: At line 8: query 'ALTER TABLE t$1 MODIFY COLUMN f2 INT,ALGORITHM=COPY' failed: 1025: Error on rename of './test/#sql-alter-7cd8f-4' to './test/t@00241' (errno: 150 "Foreign key constraint is incorrectly formed")
```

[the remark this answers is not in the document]
Thirunarayanan: Since the test case has eval statement, it tries to print $ value and fails. So I pasted the case here for your reference.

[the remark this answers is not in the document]
Thirunarayanan:
```
CREATE TABLE t$1(f1 INT, f2 INT NOT NULL, PRIMARY KEY(f1), KEY(f2))ENGINE=InnoDB;
CREATE TABLE t$2(f1 INT NOT NULL, FOREIGN KEY(f1) REFERENCES t$1(f2) ON UPDATE CASCADE)ENGINE=InnoDB;
ALTER TABLE t$1 MODIFY COLUMN f2 INT,ALGORITHM=INPLACE;
innodb.f1                                [ fail ]
        Test ended at 2024-09-27 21:54:51

CURRENT_TEST: innodb.f1
mysqltest: At line 4: query 'ALTER TABLE t$1 MODIFY COLUMN f2 INT,ALGORITHM=INPLACE' failed: 1833: Cannot change column 'f2': used in a foreign key constraint 't$2_ibfk_1' of table 'test.t@00242'
```

mysql-test/suite/innodb/r/foreign_null,INPLACE.rdiff
```diff
@@ -0,0 +1,27 @@
+--- foreign_null.result
++++ foreign_null,INPLACE.result
+@@ -48,7 +48,7 @@
+ FOREIGN KEY(f1) REFERENCES `t#1`(f2)
+ ON UPDATE CASCADE)ENGINE=InnoDB;
+ ALTER TABLE `t#1` MODIFY COLUMN f2 INT;
+-ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 't#2_ibfk_1' of table 'test.t#2'
++ERROR HY000: Cannot change column 'f2': used in a foreign key constraint 't#2_ibfk_1' of table 'test.t@00232'
+ DROP TABLE `t#2`, `t#1`;
```
dr-m, line 9: This difference looks unacceptable. One of the messages is incorrectly using a `my_charset_filename` encoded name of the table.

sql/table.h
```diff
@@ -1882,6 +1883,69 @@ typedef struct st_foreign_key_info
...
+
+  /**
+    Set nullable bit for the field in the given field
+    @param referenced set null bit for referenced column
+    @param field field number
+    @param n_fields number of fields
+  */
+  void set_nullable(bool referenced, unsigned field,
+                    unsigned n_fields) noexcept
```
dr-m, line 1924: It would seem to be safer to remove the parameter `n_fields` and remember the `num_fields` parameter that was passed to `assign_nullable()`. This object has a short lifetime and is not part of time-critical code.

I see that every call `assign_nullable()` is followed by a conditional call to `set_nullable()`. It looks like we could use a single member function that is a combination of the two. The member current function `set_nullable()` is only missing the parameter `thd` that is needed when we need to set a value to `true` and the object has not been allocated previously.

PR #3344, opened 2024-06-20
MDEV-34430: remove Debian character-set-collations

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34430*

## Description

Since MDEV-25829 Change default Unicode collation to uca1400_ai_ci there is no need to set character-set-collations explicitly as its the default.

Further mode the change in defaults affects all character sets that support the uca1400_ai_ci collation.

## Release Notes

Remove default Debian character-set-collations so that all character sets that support uca1400_ai_ci use it as default.

## How can this PR be tested?

## PR quality check


abarkov:
The patch is OK. Thanks.

abarkov (approved):
Ok to push

PR #3348, opened 2024-06-20
MDEV-34435 Increase code coverage for debug_dbug test case during startup

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34435*

## Description
- Few of test case should make sure that InnoDB does hit the debug sync point during startup of the server. InnoDB can remove the double quotes of debug point in restart parameters.

## How can this PR be tested?
./mtr innodb.innodb_buffer_pool_fail innodb.alter_copy innodb.recovery_memory

## PR quality check

### DIFF
--- a/mysql-test/suite/binlog/t/binlog_recover_checksum_error.test
+++ b/mysql-test/suite/binlog/t/binlog_recover_checksum_error.test
@@ -16,7 +16,7 @@ call mtr.add_suppression("Error in Log_event::read_log_event");
 
 # Proof of no crash follows.
 # There's no need for actual bin-loggable queries to the server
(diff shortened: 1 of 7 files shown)


dr-m (commented):
I think that the commit message needs to refer to 061adae9a26e9086f0072e49e06823299f8ce34f, where this problem was introduced. Neither `write_line` no `$restart_parameters` allow quotes around the parameters.

I found a few more errors that I think you missed:
```sh
git grep 'restart_parameters.*"'
```

My quick read of the output of the following seems to be OK.
```sh
git grep 'write_line.*"'
```

dr-m (approved):
I approve this on the condition that you check also the couple of tests outside InnoDB that use `$restart_parameters` and `"` around some arguments.

PR #3349, opened 2024-06-20
MDEV-33921: Replication fails when XA transactions are used where the slave has replicate_do_db...

bnestere:
MDEV-33921: Replication fails when XA transactions are used where the slave has replicate_do_db set and the client has touched a different database when running DML such as inserts.

There are two problems addressed by this PR:

First, replication fails when XA transactions are used where the
slave has replicate_do_db set and the client has touched a different
database when running DML such as inserts. This is because XA
commands are not treated as keywords, and are thereby not exempt
from the replication filter. The effect of this is that during an XA
transaction, if its logged “use db” from the master is filtered out
by the replication filter, then XA END will be ignored, yet its
corresponding XA PREPARE will be executed in an invalid state,
thereby breaking replication.

Second, if the slave replicates an XA transaction which results in
an empty transaction, the XA START through XA PREPARE first phase of
the transaction won’t be binlogged, yet the XA COMMIT will be
binlogged. This will break replication in chain configurations. 

This patch fixes the first problem by treating XA commands in
Query_log_events as keywords, allowing them to bypass the
replication filter. 

This patch fixes the second problem by always having the slave
binlog an empty XA START through XA PREPARE when its execution
results in an empty transaction.

The PR is organized as follows:
1. The first commit is the regression, which initially shows the first problem.
2. The second commit fixes the first problem, and its fix then highlights the second problem in the test
3. The third commit fixes the second problem

### DIFF
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -1444,6 +1444,22 @@ int ha_prepare(THD *thd)
       error=1;
     }
   }
+  else if (thd->rgi_slave)
+  {
+    /*
+      Slave threads will always process XA COMMITs in the binlog handler (see
+      MDEV-25616 and MDEV-30423), so if this is a slave thread preparing a
+      transaction which proved empty during replication (e.g. because of
+      replication filters) then mark it as XA_ROLLBACK_ONLY so the follow up
+      XA COMMIT will know to roll it back, rather than try to commit and binlog
+      a standalone XA COMMIT (without its preceding XA START - XA PREPARE).
+
+      If the xid_cache is cleared before the completion event comes, before
+      issuing ER_XAER_NOTA, first check if the event targets an ignored
+      database, and ignore the error if so.
+    */
+    thd->transaction->xid_state.set_rollback_only();
+  }
 
   DBUG_RETURN(error);
 }
--- a/sql/log_event.h
+++ b/sql/log_event.h
@@ -845,6 +845,7 @@ typedef struct st_print_event_info
   uint lc_time_names_number;
   uint charset_database_number;
   uint verbose;
+  uchar gtid_ev_flags2;
   uint32 flags2;
   uint32 server_id;
   uint32 domain_id;
@@ -916,6 +917,8 @@ typedef struct st_print_event_info
       copy_event_cache_to_file_and_reinit(&body_cache, file);
     fflush(file);
   }
+
+  my_bool is_xa_trans();
 } PRINT_EVENT_INFO;
 #endif  // MYSQL_CLIENT
 
@@ -2173,7 +2176,7 @@ public:        /* !!! Public in this patch to allow old usage */
     If true, the event always be applied by slave SQL thread or be printed by
     mysqlbinlog
    */
-  bool is_trans_keyword()
+  bool is_trans_keyword(bool is_xa)
   {
     /*
       Before the patch for bug#50407, The 'SAVEPOINT and ROLLBACK TO'
@@ -2186,10 +2189,11 @@ public:        /* !!! Public in this patch to allow old usage */
       but we don't handle these cases and after the patch, both quiries are
       binlogged in upper case with no comments.
      */
-    return !strncmp(query, "BEGIN", q_len) ||
-      !strncmp(query, "COMMIT", q_len) ||
-      !strncasecmp(query, "SAVEPOINT", 9) ||
-      !strncasecmp(query, "ROLLBACK", 8);
+    return is_xa ? !strncasecmp(query, C_STRING_WITH_LEN("XA "))
+                 : (!strncmp(query, "BEGIN", q_len) ||
+                    !strncmp(query, "COMMIT", q_len) ||
+                    !strncasecmp(query, "SAVEPOINT", 9) ||
+                    !strncasecmp(query, "ROLLBACK", 8));
   }
   virtual bool is_begin()    { return !strcmp(query, "BEGIN"); }
   virtual bool is_commit()   { return !strcmp(query, "COMMIT"); }
--- a/sql/xa.cc
+++ b/sql/xa.cc
@@ -180,6 +180,13 @@ void XID_STATE::set_error(uint error)
     xid_cache_element->rm_error= error;
 }
 
+void XID_STATE::set_rollback_only()
+{
+  xid_cache_element->xa_state= XA_ROLLBACK_ONLY;
+  if (current_thd)
+    MYSQL_SET_TRANSACTION_XA_STATE(current_thd->m_transaction_psi,
+                                   XA_ROLLBACK_ONLY);
+}
 
 void XID_STATE::er_xaer_rmfail() const
 {
@@ -547,8 +554,21 @@ bool trans_xa_prepare(THD *thd)
     }
     else
     {
-      thd->transaction->xid_state.xid_cache_element->xa_state= XA_PREPARED;
-      MYSQL_SET_TRANSACTION_XA_STATE(thd->m_transaction_psi, XA_PREPARED);
+      if (thd->transaction->xid_state.xid_cache_element->xa_state !=
+          XA_ROLLBACK_ONLY)
+      {
+        thd->transaction->xid_state.xid_cache_element->xa_state= XA_PREPARED;
+        MYSQL_SET_TRANSACTION_XA_STATE(thd->m_transaction_psi, XA_PREPARED);
+      }
+      else
+      {
+        /*
+          In the non-err case, XA_ROLLBACK_ONLY should only be set by a slave
+          thread which prepared an empty transaction, to prevent binlogging a
+          standalone XA COMMIT.
+        */
+        DBUG_ASSERT(thd->rgi_slave && !(thd->transaction->all.ha_list));
+      }
       res= thd->variables.pseudo_slave_mode || thd->slave_thread ?
         slave_applier_reset_xa_trans(thd) : 0;
     }
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -1071,7 +1071,7 @@ Exit_status process_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
     case QUERY_COMPRESSED_EVENT:
     {
       Query_log_event *qe= (Query_log_event*)ev;
-      if (!qe->is_trans_keyword())
+      if (!qe->is_trans_keyword(print_event_info->is_xa_trans()))
       {
         if (shall_skip_database(qe->db))
           goto end;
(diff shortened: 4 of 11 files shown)


sql/handler.cc
```diff
@@ -1436,14 +1436,25 @@ int ha_prepare(THD *thd)
...
+unlog_xa_prepare:
     DEBUG_SYNC(thd, "at_unlog_xa_prepare");
-
     if (tc_log->unlog_xa_prepare(thd, all))
     {
       ha_rollback_trans(thd, all);
       error=1;
     }
   }
+  else if (thd->rgi_slave)
```
andrelkin, line 1447: Wouldn't it be better to OR this condition with l.1415 one?
That optimizes away the next `goto` and l.1439.

andrelkin (approved):
Brandon, thanks for a good piece of analysis!

I made a suggestion to the sources code part.
Also please consider to detail in in the commit message to say that the empty trx
can also occur due to replicate filters.

`having the slave binlong an empty ... results in

sql/log_event.h
```diff
@@ -2189,7 +2189,8 @@ class Query_log_event: public Log_event
     return !strncmp(query, "BEGIN", q_len) ||
       !strncmp(query, "COMMIT", q_len) ||
       !strncasecmp(query, "SAVEPOINT", 9) ||
-      !strncasecmp(query, "ROLLBACK", 8);
+      !strncasecmp(query, "ROLLBACK", 8) ||
+      !strncasecmp(query, C_STRING_WITH_LEN("XA "));
   }
```
knielsen, line 2194: Ouch, yet another strcmp() added here :-(
Really, XA should be handled with a separate event like the XID_LOG_EVENT, not with yet another magic QUERY_LOG_EVENT that the code then needs to strcmp() against all over the place.

knielsen (changes requested):
> This patch fixes the second problem by always having the slave
> binlog an empty XA START through XA PREPARE when its execution
> results in an empty transaction.

That's not right. Filtered transactions are not binlogged, that's the whole point of filteri ng. Try what happens with a non-XA transaction, it will not result in an empty transaction in the binlog, neither with master-side nor with slave-side filtering.

Filtering a transaction should mean the transaction is not binlogged (master side) / not executed nor binlogged (slave side).

andrelkin:
> Filtering a transaction should mean the transaction is not binlogged

Strictly speaking a metadata of a transaction such as its identifier does not belong to the transaction.
At any rate logging of such data naturally shouldn't cause any issue to replication.

Unlike normal transaction, the XA replication at times needs such measure to cope with few "fringe" use case.
In a typical one

```
 # server start: --binlog-ignore-db=db_ign
 create table db_ign.t (a int) engine=innodb;
 use db_ign;
 xa start '1';
 insert into t set a=1;
 xa end '1';
 xa prepare '1';
 shutdown;

 # server restart
 XA COMMIT '1';
```

`XA COMMIT '1'` has to be put into binlog as at that time its handler may not know what happened
to XA-PREPARE with regard to binlogging of the transaction.
And because of that XA-PREPARE "also" had to be logged as an empty group of events.

sql/log_event.h
andrelkin: > XA should be handled with a separate event like the XID_LOG_EVENT

Perhaps.
More useful I'd argue would be to first [check](https://github.com/MariaDB/server/blob/10.5-MDEV-33921/sql/log_event_server.cc#L1718)
**unlikely**  presence of replicate-filters and not to invoke
` is_trans_keyword()` altogether.

knielsen:
Yes, XA COMMIT '1' knows what happened to the XA-PREPARE, it's in the binlog. If the XA PREPARE '1' is not found in the binlog, then it was filtered. Check the branch knielsen_mdev32020, it implements keeping track of what is in the binlog.

It's not acceptable to binlog and replicate extra GTIDs just to work-around wrong design or code. The user is filtering events for a reason, this way they could end up with unlimited amounts of garbage events on the slave that slow down replication and might even cause problems if it introduces unexpected domain_id. Quite apart from being a regression from 10.4 and being inconsistent with non-XA transactions.

If replication filters don't work with XA the correct thing is to give an error if they are used. But it seems just checking if the prepare was binlogged and only binlogging the commit if so, should solve the problem in this case.

andrelkin:
> Yes, XA COMMIT '1' knows what happened to the XA-PREPARE, it's in the binlog. If the XA PREPARE '1' is not found in the binlog, then it was filtered. 

Or binlog was purged. 

> Check the branch knielsen_mdev32020, it implements keeping track of what is in the binlog.

Right. Or like in my branch for MDEV-32830 a list of xids is  passed over at binlog rotation.
I did not mean to say the empty trx logging is the only way to deal with **that** use case.
For its virtue it just exists at this time when some fixes are indeed requested by a user.

> 
> It's not acceptable to binlog and replicate extra GTIDs just to work-around wrong design or code. The user is filtering events for a reason, this way they could end up with unlimited amounts of garbage events on the slave

MDEV-32830 fixes are going to end, or at least limit to minimum, the empty groups creation. 

> 
> If replication filters don't work with XA the correct thing is to give an error if they are used.

Well, even if there's a risk of a great number of empty groups, I would still try our best to help out the current user
with a simple solution.
And I would be with you if I had any hint or prove that even after that he would still face any slowdown etc that you're talking about.

> But it seems just checking if the prepare was binlogged and only binlogging the commit if so, should solve the problem in this case.

Indeed. MDEV-32830 which is a part of improved slave xa performance is not far away.

knielsen:
I find this endless arguing about continuing to implement things that are obviously wrong very frustrating :-(

I explained already *two years ago* the problem with replicating the XA PREPAREs on the slave, and how to implement it correctly. Nothing has been done to address this since then.

I explained in MDEV-32020 how the problem is a design that fundamentally goes counter to the core way MariaDB replication works, I even implemented a prototype implementation of a way that handles XA replication correctly. This is just being ignored, and work continues down the wrong path, making the code ever harder to eventually fix.

The problem in this bug, with the introduction of potentially arbitrary large overhead of empty transactions, is just a consequence of one part of MDEV-32020, the splitting of one transaction into two separate GTIDs, each interdependent on the other. Working around this particular issue doesn't solve the underlying design issue, and we're going to see this endless stream of similarly caused bugs continue forever.

sql/log_event.h
bnestere: In addition to @andrelkin 's suggestion, we can further optimize this with a mode to either strncmp for XA commands _or_ regular transaction commands (not both). The mode would be determined by `rgi->gtid_flags2 & (FL_PREPARED_XA | FL_COMPLETED_XA)`, so when true, we'd only do the `XA` check, and when false, we only check against regular transaction commands. I quickly prototyped a patch and it passed tests (though there is Galera and mysqlbinlog that also use this function which don't have immediate access to XA state, and may need additional extensions for them to make use of the optimization).

bnestere:
From my understanding, the current XA 2-phase replication behavior is actively used for "instant"/automatic fail-over. I imagine changing anything, at least in GA, would take a fair amount of (more) discussion/time. Whereas we have a user right now that is suffering from this bug. 

Can we proceed with the idea of this fix? I agree binlogging empty transactions is sub-optimal, but there doesn't seem to be a non-potentially replication-breaking alternative in the current state of XA. 

I can extend the other `if` statement with another comment to mention it is for 2 phase XA replication for easy tracking too.

knielsen:
Then just change the XA replication code to ignore the ER_XAER_NOTA for XA COMMIT if the prepare was filtered out. That way at least we don't change things for the worse by binlogging more empty transactions, and only touch the XA replication code for XA commit.

It's not like fixing this will make XA work correctly in GA versions...

bnestere:
Well, the server which silently corrupts the binlog never sees the `ER_XAER_NOTA`. I.e., server_2 filters and only binlogs the `XA COMMIT` (without errors), and server_3 errors from seeing the standalone `XA COMMIT` (without the presence of filters).

Another option I considered was on server_2, to set the xid cache element state to `XA_ROLLBACK_ONLY` if preparing an empty transaction so the `XA COMMIT` would know to rollback. But that wouldn't work if the server is shut down before processing the commit (as Andrei points out above). Though I suppose when we restart, we would run into `ER_XAER_NOTA` when processing the `XA COMMIT`, at which point we could check its `db` against the filter (because it is a `Query_log_event`), and if it is ignored, then we could ignore the error. Though it won't work for all cases, e.g. if the filter changes, or if the `db` of the `XA COMMIT` doesn't match the `db` of the filtered out DMLs.

andrelkin:
> Another option I considered was on server_2, to set the xid cache element state to `XA_ROLLBACK_ONLY` if preparing an empty transaction ... at which point we could check its `db` against the filter (because it is a `Query_log_event`), and if it is ignored, then we could ignore the error. 

I think that's robust enough for the moment.

knielsen:
If you're going to push this, then let's at least try to make sure it only affects XA code.

The two added conditionals for thd->rgi_slave are already in xa-only code.

But the added strcmp() in is_trans_keyword() is adding overhead to the execution of every query event on the slave when called from Query_log_event::do_apply_event().

One option to avoid this in the common case of no filter set, is to push the check for is_trans_keyword() into the rpl_filter::db_ok() function, adding the query string as an optional second argument.

Another option is to do a check for empty filter (with eg. a new function rpl_filter db_empty()) like this, though it duplicates the empty check:

  if (likely(rpl_filter->db_empty()) || is_trans_keyword() || rpl_filter->db_ok(thd->db.str)) { ... }

Also, there are two other uses of is_trans_keyword(), one in mysqlbinlog and some Galera stuff. Did you consider if the check for XA ... * is required there? If not (in particular Galera doesn't even allow XA AFAIK), then better add the XA check only in the place(s) where it is needed.

knielsen (approved):
Ok now, I guess.

 \- Kristian.

sql/xa.cc
```diff
@@ -547,8 +554,23 @@ bool trans_xa_prepare(THD *thd)
...
     {
-      thd->transaction->xid_state.xid_cache_element->xa_state= XA_PREPARED;
-      MYSQL_SET_TRANSACTION_XA_STATE(thd->m_transaction_psi, XA_PREPARED);
+      if (thd->transaction->xid_state.xid_cache_element->xa_state !=
+          XA_ROLLBACK_ONLY)
+      {
+        thd->transaction->xid_state.xid_cache_element->xa_state= XA_PREPARED;
+        MYSQL_SET_TRANSACTION_XA_STATE(thd->m_transaction_psi, XA_PREPARED);
+      }
+#ifndef DBUG_OFF
```
andrelkin, line 563: The #if/#end clauses must be unnecessary as they wrap solely DBUG_ASSERT.

andrelkin (approved):
Looks good, just missing one small note.

PR #3355, opened 2024-06-24
MDEV-34446 SIGSEGV on SET GLOBAL innodb_log_file_size with memory-mapped log file

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34446*
## Description
`log_t::resize_write()`: Advance `log_sys.resize_lsn` and reset the `resize_log` offset to `START_OFFSET` whenever the memory-mapped buffer would wrap around.

Previously, in case the initial target offset would be beyond the requested `innodb_log_file_size`, we only adjusted the offset but not the LSN. An incorrect LSN would cause `log_sys.buf_free` to be out of bounds when the log resizing completes.

The `log_sys.lsn_lock` will cover the entire duration of replicating memory-mapped log for resizing. We just need any mutex that is compatible with the caller holding `log_sys.latch`. While the choice of `mtr_t::finisher` (for normal log writes) depends on `mtr_t::spin_wait_delay`, replicating the log during resizing is a rare operation where we can afford possible additional context switching overhead.
## Release Notes
This is a race condition that is rather unlikely to trigger. Persistent memory is very rarely used. So, this does not need to be mentioned in release notes.
## How can this PR be tested?
This is best tested in the stress test framework that reproduced the issue in the first place.
## PR quality check

### DIFF
--- a/storage/innobase/mtr/mtr0mtr.cc
+++ b/storage/innobase/mtr/mtr0mtr.cc
@@ -1217,19 +1217,43 @@ inline void log_t::resize_write(lsn_t lsn, const byte *end, size_t len,
     if (!resize_flush_buf)
     {
       ut_ad(is_pmem());
+      lsn_lock.wr_lock();
       const size_t resize_capacity{resize_target - START_OFFSET};
-      const lsn_t resizing{resize_in_progress()};
-      if (UNIV_UNLIKELY(lsn < resizing))
       {
-        size_t l= resizing - lsn;
-        if (l >= len)
-          return;
-        end+= l - len;
-        len-= l;
-        lsn+= l;
+        const lsn_t resizing{resize_in_progress()};
+        /* For memory-mapped log, log_t::resize_start() would never
+        set log_sys.resize_lsn to less than log_sys.lsn. It cannot
+        execute concurrently with this thread, because we are holding
+        log_sys.latch and it would hold an exclusive log_sys.latch. */
+        if (UNIV_UNLIKELY(lsn < resizing))
+        {
+          /* This function may execute in multiple concurrent threads
+          that hold a shared log_sys.latch. Before we got lsn_lock,
+          another thread could have executed resize_lsn.store(lsn) below
+          with a larger lsn than ours.
+
+          append_prepare() guarantees that the concurrent writes
+          cannot overlap, that is, our entire log must be discarded.
+          Besides, incomplete mini-transactions cannot be parsed anyway. */
+          ut_ad(resizing >= lsn + len);
+          goto pmem_done;
+        }
+
+        s= START_OFFSET;
+
+        if (UNIV_UNLIKELY(lsn - resizing + len >= resize_capacity))
+        {
+          resize_lsn.store(lsn, std::memory_order_relaxed);
+          lsn= 0;
+        }
+        else
+        {
+          lsn-= resizing;
+          s+= lsn;
+        }
       }
-      lsn-= resizing;
-      s= START_OFFSET + lsn % resize_capacity;
+
+      ut_ad(s + len <= resize_target);
 
       if (UNIV_UNLIKELY(end < &buf[START_OFFSET]))
       {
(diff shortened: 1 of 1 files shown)


storage/innobase/mtr/mtr0mtr.cc
```diff
@@ -1223,7 +1223,12 @@ inline void log_t::resize_write(lsn_t lsn, const byte *end, size_t len,
         lsn+= l;
       }
       lsn-= resizing;
-      s= START_OFFSET + lsn % resize_capacity;
+      if (UNIV_UNLIKELY(lsn >= resize_capacity))
+      {
+        lsn-= resize_capacity;
+        resize_lsn.fetch_add(resize_capacity);
+      }
+      s= START_OFFSET + lsn;
```
mariadb-DebarunBanerjee, line 1231: Is it guaranteed that the gap between resize_lsn to lsn is less than 2 x resize_capacity ? If not then please consider having a loop here which should execute once or none in most cases (Also possible to % and / to get to the numbers).

Also, please consider putting an assert like ut_a(lsn < resize_capacity); after we are done advancing the resize_lsn. I think release assert is better here as anything wrong could lead to corruption.

mariadb-DebarunBanerjee (commented):
Looks like the issue is introduced when the feature was developed,
MDEV-27812 Allow SET GLOBAL innodb_log_file_size

Also, went through the change in MDEV-27812 to get a better understanding. The patch looks correct to me. Perhaps it is possible to design an mtr test to block checkpoint and executing repeated dmls (calling mtr_t::commit -> ...log_t::resize_write).

I have one doubt added as comment.

storage/innobase/mtr/mtr0mtr.cc
dr-m: I had a similar doubt myself. In fact, the atomic operations on `resize_lsn` may not be safe, because multiple calls can execute concurrently, under shared `log_sys.latch`. I will think of something else, such as acquiring `log_sys.lsn_lock`. Then, `std::memory_order_relaxed` loads and stores should be safe.

dr-m: In 7064ffadc8400836a9a4b8589bf835d394cc03a3 I revised this so that in case a wrap-around is noticed, the rest of `log_t::resize_write()` will be protected by `lock_lsn()`.

storage/innobase/mtr/mtr0mtr.cc
```diff
@@ -1217,19 +1217,35 @@ inline void log_t::resize_write(lsn_t lsn, const byte *end, size_t len,
...
-        lsn+= l;
+        const lsn_t resizing{resize_in_progress()};
+        if (UNIV_UNLIKELY(lsn < resizing))
+        {
+          size_t l= resizing - lsn;
+          if (l >= len)
+            goto pmem_done;
+          end+= l - len;
+          len-= l;
+          lsn+= l;
```
mariadb-DebarunBanerjee, line 1231: `end+= l - len;` looks incorrect to me. When resize_lsn is ahead, we need to skip copying the initial part of the buffer i.e. `l` (resizing - lsn). Why should we subtract `len` here ?

Can we have a test covering this part ? If resize LSN is set when LSN is in the beginning of a write block (`write_size`), the subsequent mtr should follow this path. Once the resize is complete a crash and restart should reflect the issue if I am correct.

I am assuming it may not be very difficult to have a test now we are setting the resize_lsn ahead of log_sys LSN in the other patch. Perhaps this code was unreachable be before the other change.

storage/innobase/mtr/mtr0mtr.cc
```diff
@@ -1217,19 +1217,35 @@ inline void log_t::resize_write(lsn_t lsn, const byte *end, size_t len,
     if (!resize_flush_buf)
     {
       ut_ad(is_pmem());
```
mariadb-DebarunBanerjee, line 1219: I have a general question/suggestion about testing the is_pmem() T/F path on linux. Since the paths are significantly different, can we have a set of testcases (redo and recovery related) to be forced to run 2 times following each path ? Currently on linux, I see that is_pmem() path is followed when I use RAM disk (./mtr --mem). Probably CI also runs same way.

mariadb-DebarunBanerjee (commented):
The simplified code seems good to me. I have one concern on the existing code related to the current issue. Can you please check.

storage/innobase/mtr/mtr0mtr.cc
dr-m: We could add a debug build specific parameter for enabling or disabling the special handling of  `/dev/shm` . That would allow both code paths to be covered easily. Currently, I think that on any qualifying 64-bit Linux platform, https://buildbot.mariadb.org and https://buildbot.mariadb.net would run with memory-mapped files on `/dev/shm`. That is, this code path (and not the `pread` and `pwrite` based one) is what is being tested most.

storage/innobase/mtr/mtr0mtr.cc
dr-m: "The other patch" would be #3458, which only affects the non-PMEM code path. For *memory-mapped log* (which is the domain of this fix: `is_pmem()` holds if and only if `!flush_buf` and `!resize_flush_buf`), the only way how `log_sys.resize_lsn` can be less than `log_sys.lsn` is when it is 0 (online log resizing is not in progress) or 1 (`log_t::resize_start()` is creating and allocating an `ib_logfile101` that aims to be the resized log file).

It would seem that this is in fact dead code and needs to be replaced with a debug assertion.

The critical path of `log_t::resize_start()` is enclosed between `log_resize_acquire()` and `log_resize_release()`, which includes an exclusive `log_sys.latch`. The callers of this function `log_t::resize_write()` must hold a shared or exclusive `log_sys.latch` until they have completed their write. Therefore, `resize_write()` and `resize_start()` cannot run concurrently and there is no chance that a resizing that is about to start would cause `lsn < resizing` to hold or otherwise be able to conflict with an older `log_t::resize_write()`.

mariadb-DebarunBanerjee: Ok, let's remove the dead code then.

storage/innobase/mtr/mtr0mtr.cc
dr-m: To test the non-memory-mapped log writes on RAM disk, the following is possible already, and used by me:
```sh
mkdir /run/user/$UID/var
rm -fr mysql-test/var
ln -s /run/user/$UID/var mysql-test/var
```
In my current Debian Sid system, I could also symlink `mysql-test/var` to a subdirectory of `/tmp`, which would be on RAM disk nowadays.

There is no PMEM support on at least one 64-bit Linux architecture on https://buildbot.mariadb.org: IBM System Z (s390x). We also have a 32-bit Linux target (x86) where we do not support PMEM. We also have FreeBSD, IBM AIX, Apple macOS without PMEM support. I think that the regular file I/O path is adequately covered on our CI systems.

If a production user stores the write-ahead log in `/dev/shm`, they should not care which impelmentation is being used. I wanted to implement `/dev/shm` as a special case so that we will get adequate test coverage of the PMEM code path, and so that there is a convenient way to test the relative performance of memory-mapped and `pwrite` based log writes.

storage/innobase/mtr/mtr0mtr.cc
dr-m: Sorry, I realized that it is not dead code at all. Regular log writers would only hold a shared `log_sys.latch`, and therefore a concurrent `log_t::resize_write()` in another thread could have updated the `log_sys.resize_lsn`. If a concurrent `resize_write()` with a larger LSN was able to grab the `log_sys.lsn_lock` first, then any older writes would have to be discarded.

dr-m: The code after `goto pmem_done` is actually dead, because the `resize_lsn.store(lsn)` in another thread must have advanced the log beyond our `lsn+len`. The concurrent mini-transaction writes to `log_sys.buf` cannot overlap, so they must not overlap here for `log_sys.resize_buf` either.

mariadb-DebarunBanerjee (approved):
Thanks for the changes. Looks good to me.

storage/innobase/mtr/mtr0mtr.cc
mariadb-DebarunBanerjee: Right, Thanks for explaining.

storage/innobase/mtr/mtr0mtr.cc
mariadb-DebarunBanerjee: Ok. For the record let me cut-paste the messages from slack discussion that would help us identifying if mapped redo memory is used.

```
2024-09-04 14:22:12 0 [Note] InnoDB: File system buffers for log disabled (block size=512 bytes)
2024-09-04 14:22:12 0 [Note] InnoDB: End of log at LSN=58172

2024-09-04 14:22:56 0 [Note] InnoDB: log sequence number 56287 (memory-mapped); transaction id 16
```

PR #3364, opened 2024-06-27
MDEV-34471 add test case to show faking subject/issuer certs cannot connect

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34471*

## Description

mysql-test: generate-ssl-certs.sh and openssl.cnf
where updated so they where self sufficient.

The script was run and the generated files where
replaces with *_faked.*:
* cacert.pem
* client-cert.pem
* client-key.pem

## Release Notes

(test)

## How can this PR be tested?

mtr test included

- [] *This is a new feature or a refactoring, and the PR is based against the latest MariaDB development branch.*
- [] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*
- [*] *This is a test case that could be backported without merge conflicts.

## PR quality check
- [*] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.


vuvova:
I don't quite understand, what does it test?

grooverdan:
A client cert by same issuer subject in a certificate by a different CA than the ssl-ca isn't accepted

vuvova:
you mean, you test that an invalid certificate isn't accepted? I'm sure there are existing tests for that

LinuxJedi:
Question: if we don't already have tests for this (I can't see obvious ones that would cover this specific case), should it go in `ssl_ca.test`?

LinuxJedi (approved):
Whilst my question still stands, I don't see a need to hold it up for that.

grooverdan:
> you mean, you test that an invalid certificate isn't accepted? I'm sure there are existing tests for that

per Andrew's comment there isn't. In ssl_ca.test there are tests without a CA, but none with a CA that look similar but isn't the same as the issued certificate.

As this test is about attempting to fake the authentication of ssl_user3 and ssl_user4 I've kept these beside the original connection tests.

fake keys regenerated as the standard is now 4k RSA.

removed asan exclusion added in 2019. There's no ASAN errors.

vuvova:
This is already tested, in the very same file, few lines below:

https://github.com/MariaDB/server/blob/bfcd2674a3b20b105393970d3937cdc18f43ce23/mysql-test/main/openssl_1.test#L67-L76

grooverdan:
Its not the same. Testing with an old expired 512bit CA, that would pass within the regex of a fail for either of these reasons, with a different subject line isn't the same as testing an identical certificate information with a different private/public key.

At some level its just an invalid CA, but if you leave it so open with a large number of differences the point at which is can be rejected are many. By constraining it as much as possible it provide a stronger test case of the validation of the certificate chain.

Andrew saw this.

> A client cert by *same issuer subject* in a certificate by a different CA than the ssl-ca isn't accepted

This is checking the the server does a full validation of the certificate change rather than relying on subject information.

vuvova:
I believed the server will reject an invalid certificate before it'll start looking at the subject, so it doesn't matter how similar the subject is. Are you saying this is not the case and the subject value is important even if the cert is invalid?

PR #3368, opened 2024-06-27
MDEV-34477 galera.galera_gcache_recover_manytrx sporadic failures

sjaakola:
- [x] *The Jira issue number for this PR is: MDEV-34477*

## Description
Fixing sporadic test failures with galera.galeraa_gcache_recover_manytrx

The problem was in error message suppression, which did not match the actual warning messages, due to bad quotations. Changed warnings message suppressions to more simple format.

## How can this PR be tested?
galera.galera_gcache_recover_manytrx should not fail for warning of type:
[ERROR] Error in Log_event::read_log_event(): 'Event truncated', data_len: 10485802, event_type: 23

- [*] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check


mysql-test/suite/galera/r/galera_gcache_recover_manytrx.result
```diff
@@ -131,9 +130,9 @@ DROP PROCEDURE update_simple;
 DROP PROCEDURE insert_1k;
 DROP PROCEDURE insert_1m;
 connection node_1;
-call mtr.add_suppression("Error in Log_event::read_log_event\(\):.*");
+call mtr.add_suppression("Error in Log_event::read_log_event.*");
```
dr-m, line 133: The trailing `.*` is actually redundant. I think that passing a `\` to the regular expression layer would require escaping it as `\\`. The `\(\)` would actually be equivalent to `()`, which in PCRE refers to a group, so originally this would have matched `…read_log_event:` while the actual output is `…read_log_event():` So, either the literal parentheses have to be omitted, or represented correctly as `\\(\\)`.

dr-m (approved):

sysprg:
Thanks, the fix merged with the head revision: https://github.com/MariaDB/server/commit/8d61a94b8b1cfc18e1a07f8a29dcaef1dfe9d73b

PR #3369, opened 2024-06-27
MDEV-34466 XA prepare don't release unmodified records for some cases

vlad-lesin:
There is no need to exclude exclusive non-gap locks from the procedure
    of locks releasing on XA PREPARE execution in
    lock_release_on_prepare_try() after commit
    17e59ed3aad481918f1a01d8afbc071c316d5930 (MDEV-33454), because
    lock_rec_unlock_unmodified() should check if the record was modified
    with the XA, and release the lock if it was not.
    
    lock_release_on_prepare_try(): don't skip X-locks, let
    lock_rec_unlock_unmodified() to process them.
    
    lock_sec_rec_some_has_impl(): add template parameter for not acquiring
    trx_t::mutex for the case if a caller already holds the mutex, don't
    crash if lock's bitmap is clean.
    
    row_vers_impl_x_locked(), row_vers_impl_x_locked_low(): add new argument
    to skip trx_t::mutex acquiring.
    
    rw_trx_hash_t::validate_element(): don't acquire trx_t::mutex if the
    current thread already holds it.


dr-m (commented):
The changes look OK, I only have some minor nitpicks. I would suggest to write
`commit 17e59ed3aad481918f1a01d8afbc071c316d5930 (MDEV-33454)`
in the commit message so that it is a little more convenient to find the change that it is referring to. In the GitHub web user interface the 17e59ed3aad481918f1a01d8afbc071c316d5930 should be displayed as a hyperlink.

Once you have addressed these, @mariadb-DebarunBanerjee should review this.

storage/innobase/include/trx0sys.h
```diff
@@ -500,12 +500,13 @@ class rw_trx_hash_t
     ut_ad(!trx->read_only || !trx->rsegs.m_redo.rseg);
     ut_ad(!trx->is_autocommit_non_locking());
     /* trx->state can be anything except TRX_STATE_NOT_STARTED */
-    ut_d(trx->mutex_lock());
+    bool acquire_trx_mutex = !trx->mutex_is_owner();
+    ut_d(if (acquire_trx_mutex) trx->mutex_lock());
```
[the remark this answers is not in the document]
vlad-lesin: `rw_trx_hash_t::validate_element()` is under `#ifdef UNIV_DEBUG`.

mariadb-DebarunBanerjee (commented):
Hi Vladislav,
Thanks for the patch. I have reviewed the code and included my comments. The design level comments are from my general DB/MySQL understanding and I may not be completely correct with MariaDB. I still think it is worth to point them out for your consideration.

As this code is enabling the major part of the base MDEV work, we should consider testing the feature well for impacts.

It could have significant performance impact in Primary while checking for 2ndary index implicit lock especially with concurrent load modifying index records.

Since this early release of locks hardly helps the primary (rather impacts the performance/concurrency), we should evaluate how much it indeed helps the replica with scenarios where it is effective. 

For row level replication, the replica should be updating the rows that are modified directly and the locks acquired during range scanning for update in Primary should not be there while replaying. So, I think there are only very specific cases, like unique index insert S locks where the optimization for releasing S lock early is really useful.  

For statement level replication, I think releasing S locks early has a functional issue related to transaction ordering. Statement level replication is essentially dependent on Serializability (Note that RR is serializable when only DMLs are considered) and releasing S lock at XA PREPARE impacts the ordering.I would try to repeat it with mtr and update.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4501,7 +4503,9 @@ static void lock_rec_unlock_unmodified(hash_cell_t &cell, lock_t *lock,
   {
     if (UNIV_UNLIKELY(!page_is_leaf(block->page.frame)))
     {
-      ut_ad("corrupted lock system" == 0);
+      /* There can be locks with clean bitmap on non-leaf pages in the case of
+      page split. See lock_move_rec_list_end() call stack for details. */
+      ut_ad(lock_rec_find_set_bit(lock) == ULINT_UNDEFINED);
```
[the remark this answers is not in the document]
vlad-lesin: Yes, this test is innodb.lock_update_split_rc. 

The stack trace is the following:
``` text
#0  0x00005ebaaabe8da7 in lock_rec_reset_nth_bit (lock=0x7244198d7d00, i=4) at ./storage/innobase/include/lock0priv.h:489
#1  0x00005ebaaabd1041 in lock_move_rec_list_end (new_block=0x7244190e6700, block=0x7244190e6490, rec=0x7244195a8063 "infimum")
    at ./storage/innobase/lock/lock0lock.cc:2879                                
#2  0x00005ebaaac82a02 in page_copy_rec_list_end (new_block=0x7244190e6700, block=0x7244190e6490, rec=0x7244195a8063 "infimum", index=0x7243f000de68,
    mtr=0x7244111a8db0, err=0x7244111a893c) at ./storage/innobase/page/page0page.cc:663
#3  0x00005ebaaae17284 in btr_root_raise_and_insert (flags=0, cursor=0x7244111a8a80, offsets=0x7244111a8a00, heap=0x7244111a89f8, tuple=0x7243fc036df8,
    n_ext=0, mtr=0x7244111a8db0, err=0x7244111a893c) at ./storage/innobase/btr/btr0btr.cc:1987
#4  0x00005ebaaae44049 in btr_cur_pessimistic_insert (flags=0, cursor=0x7244111a8a80, offsets=0x7244111a8a00, heap=0x7244111a89f8, entry=0x7243fc036df8,
    rec=0x7244111a8a08, big_rec=0x7244111a89f0, n_ext=0, thr=0x7243fc036298, mtr=0x7244111a8db0)
    at ./storage/innobase/btr/btr0cur.cc:2754                                   
#5  0x00005ebaaacfbfff in row_ins_clust_index_entry_low (flags=0, mode=BTR_MODIFY_TREE, index=0x7243f000de68, n_uniq=1, entry=0x7243fc036df8, n_ext=0,
    thr=0x7243fc036298) at ./storage/innobase/row/row0ins.cc:2855               
#6  0x00005ebaaacfd6c1 in row_ins_clust_index_entry (index=0x7243f000de68, entry=0x7243fc036df8, thr=0x7243fc036298, n_ext=0)
    at ./storage/innobase/row/row0ins.cc:3256
```

It looks like it was done intentionally. Here is the comment of lock_update_root_raise() function:
```text
/** Update locks when the root page is copied to another in                     
btr_root_raise_and_insert(). Note that we leave lock structs on the             
root page, even though they do not make sense on other than leaf                
pages: the reason is that in a pessimistic update the infimum record            
of the root page will act as a dummy carrier of the locks of the record         
to be updated. */                                                               
void lock_update_root_raise(const buf_block_t &block, const page_id_t root)
```

There is also lock_update_copy_and_discard() function, which discards lock objects from the certain page, and it's used for merging pages, but it's not used in the above call stack.

It does not cause lock objects leak because the lock objects will be released on transaction commit or rollback.

storage/innobase/include/btr0btr.h
```diff
@@ -83,18 +83,19 @@ btr_root_adjust_on_import(
...
+@param		page_get_mode	page get mode
 @return block */
-buf_block_t *btr_block_get(const dict_index_t &index,
-                           uint32_t page, rw_lock_type_t mode, bool merge,
-                           mtr_t *mtr, dberr_t *err= nullptr,
-                           bool *first= nullptr);
+buf_block_t *btr_block_get(const dict_index_t &index, uint32_t page,
+                           rw_lock_type_t latch_mode, bool merge, mtr_t *mtr,
+                           dberr_t *err= nullptr, bool *first= nullptr,
+                           ulint page_get_mode= BUF_GET);
```
dr-m, line 98: The parameter `mode` was renamed to `latch_mode`, but not in the function comment in `btr0btr.cc`. In this file, the formatting of the `@param` comments got uglier beacuse one of the parameter names is longer. Could we retain the old naming?

Instead of changing the function signature, could we add a debug-only field to `mtr_t` to denote that `BUF_GET` should be treated as `BUF_GET_POSSIBLY_FREED`? That would avoid a potential performance regression while keeping the debug check happy. After all, `BUF_GET` and `BUF_GET_POSSIBLY_FREED` are equivalent in non-debug builds.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4583,12 +4588,49 @@ static bool lock_release_on_prepare_try(trx_t *trx)
...
+        {
+          ulint set_nth_bit_calls= trx->lock.set_nth_bit_calls;
+          latch->release();
+          lock_sys.rd_unlock();
+          trx->mutex_unlock();
+          mtr.start();
+          buf_block_t *block= btr_block_get(
+              *lock->index, lock->un_member.rec_lock.page_id.page_no(),
+              RW_S_LATCH, true, &mtr, nullptr, nullptr,
+              BUF_GET_POSSIBLY_FREED);
```
dr-m, line 4610: Could we set a new debug-only field in `mtr` to signal that `btr_block_get()` should not complain if the page has been freed meanwhile?

storage/innobase/include/trx0trx.h
```diff
@@ -407,6 +407,9 @@ struct trx_lock_t
 
   /** number of record locks; protected by lock_sys.assert_locked(page_id) */
   ulint n_rec_locks;
+  /** number of lock_rec_set_nth_bit() calls since the start of transaction;
+  protected by lock_sys.is_holder() */
+  ulint set_nth_bit_calls;
```
dr-m, line 412: This seems to be protected by `trx_t::mutex` as well. Is that the case? `lock_sys.is_holder()` can also refer to a non-exclusive `lock_sys.latch`.

dr-m (commented):
My comments are about the MDEV-34690 patch.

I see that there is some rather similar code between `lock_release_on_prepare_try()` and `lock_release_on_prepare()` for releasing the record locks, but there seem to be enough differences so that the similar code can’t be refactored to a separate function.

mysql-test/suite/sys_vars/r/sysvars_innodb.result
```diff
@@ -535,6 +535,18 @@ NUMERIC_BLOCK_SIZE	NULL
 ENUM_VALUE_LIST	OFF,ON
 READ_ONLY	YES
 COMMAND_LINE_ARGUMENT	NONE
+VARIABLE_NAME	INNODB_ENABLE_XAP_UNLOCK_UNMODIFIED_FOR_PRIMARY_DEBUG
```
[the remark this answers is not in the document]
vlad-lesin: Why do we need to run the test for all targets? It just checks the presence of system variables, `have_debug.inc` gives more coverage for that. We don't have release-only variables, so we don't need a separate run for release build.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4692,61 +4756,85 @@ void lock_release_on_prepare(trx_t *trx)
...
+          buf_block_t *block= btr_block_get(
+              *lock->index, lock->un_member.rec_lock.page_id.page_no(),
+              RW_S_LATCH, true, &mtr, nullptr, nullptr
+#if defined(UNIV_DEBUG) || !defined(DBUG_OFF)
+              , BUF_GET_POSSIBLY_FREED
+#endif /* defined(UNIV_DEBUG) || !defined(DBUG_OFF) */
+          );
+          lock_sys.wr_lock(SRW_LOCK_CALL);
+          if (block)
+            lock_rec_unlock_unmodified(block, *cell, lock, offsets, heap);
```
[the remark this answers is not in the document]
vlad-lesin: Yes, you are absolutely right, this is race condition. I will fix it. But we can't place it in `lock_rec_unlock_unmodified()`. Because in `lock_release_on_prepare()`  it's invoked under `lock_sys.wr_lock()`, but in `lock_release_on_prepare_try()`  it's a combination of `lock_sys.rd_lock()`  + `latch->try_acquire()`. But we can re-get the cell in  `lock_release_on_prepare()`  and `lock_release_on_prepare_try()`  under the correspondent latches after a block is gotten.

storage/innobase/lock/lock0lock.cc
```diff
@@ -60,11 +61,19 @@ my_bool innodb_deadlock_detect;
 /** The value of innodb_deadlock_report */
 ulong innodb_deadlock_report;
 
+#if defined(UNIV_DEBUG) || \
+    defined(INNODB_ENABLE_XAP_UNLOCK_UNMODIFIED_FOR_PRIMARY)
+/** Enable unmodified records unlocking on XA PREPARE for master. */
+extern my_bool innodb_enable_xap_unlock_unmodified_for_primary_debug;
+#endif /* defined(UNIV_DEBUG) ||
+          defined(INNODB_ENABLE_XAP_UNLOCK_UNMODIFIED_FOR_PRIMARY) */
```
dr-m, line 69: We could omit a subsequent occurrence of "`#if` spam" by defining an `#else` branch:
```c++
#if defined UNIV_DEBUG || defined INNODB_VERY_LONG_NAME
extern my_bool innodb_very_long_name;
#else
static constexpr bool innodb_very_long_name=false;
#endif
```
In this way, the "user" of this parameter may simply refer to it by name, without any run-time penalty.

I don’t think that it is necessary to repeat the `#if` condition in comments after `#else` or `#endif` if the enclosed code sections are short, like in this case.

storage/innobase/lock/lock0lock.cc
```diff
@@ -2310,8 +2320,8 @@ static void lock_rec_dequeue_from_page(lock_t *in_lock, bool owns_wait_mutex)
 
 	const page_id_t page_id{in_lock->un_member.rec_lock.page_id};
 	auto& lock_hash = lock_sys.hash_get(in_lock->type_mode);
-	ut_ad(lock_sys.is_writer() || in_lock->trx->mutex_is_owner());
-
+	ut_ad(lock_sys.is_writer() || in_lock->trx->mutex_is_owner()
+		|| lock_sys.is_cell_locked(*in_lock));
```
dr-m, line 2324: In the old InnoDB formatting style, which this function is following, lines should be indeed split before binary operators such as `||`. But, the continuation lines should be indented so that they align with the preceding `(`. In this case, instead of TAB+TAB the indentation should be TAB+6 spaces.

There are a few other places like this in this file.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4692,61 +4768,90 @@ void lock_release_on_prepare(trx_t *trx)
...
+          mtr.start();
+          buf_block_t *block= btr_block_get(
+              *lock->index, lock->un_member.rec_lock.page_id.page_no(),
+              RW_S_LATCH, true, &mtr, nullptr, nullptr
+#if defined(UNIV_DEBUG) || !defined(DBUG_OFF)
+              , BUF_GET_POSSIBLY_FREED
+#endif /* defined(UNIV_DEBUG) || !defined(DBUG_OFF) */
+          );
+          lock_sys.wr_lock(SRW_LOCK_CALL);
+          if (block) {
```
dr-m, line 4812: In the common MariaDB formatting style, Monty wants us to deviate from K&R and split the line before `{` except when it is after a `switch` expression.

dr-m (approved):
Thank you, this looks good to me. I only have a few formatting nitpicks.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4692,61 +4820,92 @@ void lock_release_on_prepare(trx_t *trx)
...
-                     level > TRX_ISO_READ_COMMITTED */
-                  lock->mode() == LOCK_X ||
-                  !lock_rec_get_nth_bit(lock, PAGE_HEAP_NO_SUPREMUM));
-            lock_rec_unlock_unmodified(*cell, lock, offsets, heap, mtr);
+            /* concurent lock_sys_t::resize() call could make a previously
+            computed cell address invalid */
+            cell= lock_sys.rec_hash.cell_get(lock->un_member.rec_lock.
+                                             page_id.fold());
+            (void) lock_rec_unlock_unmodified(block, cell, lock, offsets, heap,
+                                              GLOBAL);
```
dr-m, line 4871: @mariadb-DebarunBanerjee recently taught us that the correct C++11 way to declare that a result is being discarded is to assign it to `std::ignore`:
```c++
            std::ignore=
              lock_rec_unlock_unmodified(block, cell, lock, offsets_, GLOBAL);
```
See my other comment regarding removing `offsets` and `heap` from the caller.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4591,22 +4660,70 @@ static bool lock_release_on_prepare_try(trx_t *trx)
...
       bool rec_granted_exclusive_not_gap=
         lock->is_rec_granted_exclusive_not_gap();
-      if (!supremum_bit && rec_granted_exclusive_not_gap)
-        continue;
       if (UNIV_UNLIKELY(lock->type_mode & (LOCK_PREDICATE | LOCK_PRDT_PAGE)))
         continue; /* SPATIAL INDEX locking is broken. */
-      auto cell=
-        lock_sys.rec_hash.cell_get(lock->un_member.rec_lock.page_id.fold());
+      auto cell= lock_sys.rec_hash.cell_get(lock->un_member.rec_lock.
+                                            page_id.fold());
```
dr-m, line 4666: I’d suggest that we assign the `fold` into a local variable, because we may be reusing it:
```c++
      const auto fold= lock->un_member.rec_lock.page_id.fold();
      auto cell= lock_sys.rec_hash.cell_get(fold);
```

storage/innobase/lock/lock0lock.cc
```diff
@@ -4591,22 +4660,70 @@ static bool lock_release_on_prepare_try(trx_t *trx)
...
+              , BUF_GET_POSSIBLY_FREED
+#endif /* defined(UNIV_DEBUG) || !defined(DBUG_OFF) */
+          );
+          lock_sys.rd_lock(SRW_LOCK_CALL);
+          if (block)
+          {
+            /* concurent lock_sys_t::resize() call could make a previously
+            computed cell address invalid */
+            cell= lock_sys.rec_hash.cell_get(lock->un_member.rec_lock.
+                                             page_id.fold());
```
dr-m, line 4700: I’d reuse a previously computed `fold` value.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4498,64 +4523,108 @@ static void lock_rec_unlock(hash_cell_t &cell, lock_t *lock, ulint heap_no)
...
+@param offsets    storage for rec_get_offsets()
+@param heap       storage for rec_get_offsets()
+@param latch_type how the caller of the function latched lock_sys,
+                  GLOBAL or CELL
+@return true if the cell was latched successfully or if latch_type is GLOBAL,
+        false otherwise */
+static bool lock_rec_unlock_unmodified(buf_block_t *block, hash_cell_t *cell,
+                                       lock_t *lock, rec_offs *&offsets,
+                                       mem_heap_t *&heap,
+                                       lock_sys_latch_type latch_type)
```
dr-m, line 4538: `latch_type` could also be a `template<bool global>`; the function looks rather small, so expanding it twice should not incur too much code bloat. It’s up to you; this should not be frequently executed or time-critical code, so replacing the run-time conditions on `latch_type` with machine code duplication might not matter much.

dr-m (commented):
Thank you for asking for a re-review. The first commit was not changed in the force-push. I compared the second commit as of 737dd09046f0f8c4c97c64d99e76ecc57badcecd and 37fcbc1024ebcff77f36745d0a2f35fde6a68133.

This looks OK, but I have a few minor suggestions that could be good to address.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4680,73 +4800,119 @@ and release possible other transactions waiting because of these locks. */
...
+#endif
 
   mtr_t mtr;
-  rec_offs offsets_[REC_OFFS_NORMAL_SIZE];
+  /* Reserve enough offsets for the key and PRIMARY KEY. */
+  rec_offs offsets_[REC_OFFS_HEADER_SIZE + 2 * MAX_REF_PARTS + 1];
   rec_offs *offsets= offsets_;
   mem_heap_t *heap= nullptr;
 
   rec_offs_init(offsets_);
```
dr-m, line 4825: Both `offsets` and `heap` should be removed. A local variable `heap` should be declared in `lock_rec_unlock_unmodified()`, which this function would invoke on `offsets_`.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4498,64 +4523,108 @@ static void lock_rec_unlock(hash_cell_t &cell, lock_t *lock, ulint heap_no)
...
-              lock->trx->id)
-            continue;
-        unlock_rec:
-          lock_rec_unlock(cell, lock, i);
+          lock_sys_t::hash_table::latch(cell)->release();
+          lock_sys.rd_unlock();
+        }
+
+        offsets= rec_get_offsets(rec, index, offsets, index->n_core_fields,
+                                 ULINT_UNDEFINED, &heap);
```
dr-m, line 4583: Add before and after this:
```c++
mem_heap_t *heap= nullptr;
ut_ad(!heap);
```

dr-m (commented):
Sorry, it looks like I forgot to publish some earlier comments.

dr-m (approved):

PR #3371, opened 2024-06-28
MDEV-34489 innodb.innodb_row_lock_time_ms fails

vlad-lesin:
The test fails because 'variable_value' has not the same type as the value which it's compared to. The fix is to use CAST() not only for 'variable_value', but also for 'count_reset' to avoid future failures.

Reviewed by Marko Mäkelä.

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/innodb_row_lock_time_ms.test
+++ b/mysql-test/suite/innodb/t/innodb_row_lock_time_ms.test
@@ -19,18 +19,18 @@ SELECT * FROM t FOR UPDATE;
 --connection default
 COMMIT;
 
-SELECT variable_value > 100 FROM information_schema.global_status
+SELECT CAST(variable_value AS INT) > 100 FROM information_schema.global_status
   WHERE LOWER(variable_name) = 'innodb_row_lock_time';
-SELECT variable_value > 100 FROM information_schema.global_status
+SELECT CAST(variable_value AS INT) > 100 FROM information_schema.global_status
   WHERE LOWER(variable_name) = 'innodb_row_lock_time_max';
-SELECT variable_value > 100 FROM information_schema.global_status
+SELECT CAST(variable_value AS INT) > 100 FROM information_schema.global_status
   WHERE LOWER(variable_name) = 'innodb_row_lock_time_avg';
 
-SELECT count_reset > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
+SELECT CAST(count_reset AS INT) > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
   WHERE NAME="lock_row_lock_time";
-SELECT count_reset > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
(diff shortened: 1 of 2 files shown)


mysql-test/suite/innodb/t/innodb_row_lock_time_ms.test
```diff
@@ -19,18 +19,18 @@ SELECT * FROM t FOR UPDATE;
...
 
-SELECT count_reset > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
+SELECT CAST(count_reset AS INT) > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
   WHERE NAME="lock_row_lock_time";
-SELECT count_reset > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
+SELECT CAST(count_reset AS INT) > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
   WHERE NAME="lock_row_lock_time_max";
-SELECT count_reset > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
+SELECT CAST(count_reset AS INT) > 100 FROM INFORMATION_SCHEMA.INNODB_METRICS
   WHERE NAME="lock_row_lock_time_avg";
```
dr-m, line 34: As far as I can tell, `count_reset` is declared as `SLonglong()` in `storage/innobase/handler/i_s.cc`. Yet, we have observed some failures where the output differs. I don’t think that the `CAST` will help here.

mysql-test/suite/innodb/t/innodb_row_lock_time_ms.test
```diff
@@ -19,18 +19,18 @@ SELECT * FROM t FOR UPDATE;
 --connection default
 COMMIT;
 
-SELECT variable_value > 100 FROM information_schema.global_status
+SELECT CAST(variable_value AS INT) > 100 FROM information_schema.global_status
   WHERE LOWER(variable_name) = 'innodb_row_lock_time';
```
dr-m, line 23: To obtain more information from the failures, I think that we should follow the example of `storage/rocksdb/mysql-test/rocksdb/t/2pc_group_commit.test`, with something like this:
```sql
SELECT IF(CAST(variable_value AS INT) > 100, 'OK', variable_value) AS innodb_row_lock_time
FROM information_schema.global_status WHERE variable_name='Innodb_row_lock_time';
```

dr-m (commented):
I think that this needs more effort, maybe running the test a few thousand times under `rr` until we get a failure. There have been examples where we have a result difference for `count_reset`, and it does not look like it would be explained by an invalid comparison, like it was the case in ead61d9bd96b773a3bbacf16c20df7cc34d85c96.

In fact, if we think of it, `variable_value > 100` would hold for most string values that are possible in `variable_value`. If the comparison was like `variable_value > 90` then it would be obviously incorrect for a case like `'100' > 90` which I think would be evaluated as `'100' > '90'`, which would not hold.

vlad-lesin:
This pull request is closed, because another fix was applied: https://github.com/MariaDB/server/pull/3809.

PR #3372, opened 2024-06-28
MDEV-23983 Assertion failure on multiple equalities present in HAVING…

Olernov:
… during execution phase

During optimization, there is a step where the HAVING clause is checked for being an impossible condition (i.e, always FALSE). This check is performed using the `remove_eq_conds()` function. However, a side effect of this function is that it may convert some predicates into multiple equalities, which are not allowed during the execution stage, causing an assertion failure.

To address this issue, a clone of the HAVING clause is created before calling `remove_eq_conds()`. This preserves the original HAVING clause from modification, ensuring it remains unchanged while only checking for an impossible condition

- [x] *The Jira issue number for this PR is: MDEV-_23983_*

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/main/having.result
+++ b/mysql-test/main/having.result
@@ -953,5 +953,20 @@ WHERE (0, a) IN ((0,-1),(+1,0))
 ORDER BY 1+AVG(a) OVER (ORDER BY a)) ORDER BY a;
 DROP TABLE t;
 #
+# MDEV-23983: Assertion failure on multiple equalities present
+#             in HAVING during execution phase
+#
+CREATE TABLE t1 (id INT PRIMARY KEY, a INT, descr TEXT, tp INT);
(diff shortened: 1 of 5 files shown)


spetrunia (changes requested):
The testcase passes when I try it on the tree before the patch.  Is it possible (with reasonable effort) to construct a testcase that would fail?

Olernov:
Execution plan in the test case was unstable, I added `ANALYZE TABLE` to stabilize it.

spetrunia:
This patch seems go in the wrong direction. Details:
https://jira.mariadb.org/browse/MDEV-23983?focusedCommentId=288610&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-288610

PR #3376, opened 2024-06-30
MDEV-34141: Implements the function ST_Simplify

StefanoPetrilli:
- [x] *The Jira issue number for this PR is: [MDEV-34141](https://jira.mariadb.org/browse/MDEV-34141)

## Description

MariaDB GIS functionalities are limited when compared to other DBMs. To facilitate the migration from MySQL to MariaDB the GIS functionality should be the same. This PR implements the GIS function `ST_Simplify(Geometry g, double max_distance)`.  It applies the [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm] ) algorithm on `g` and returns the resulting geometry.
The goal of the Douglas-Peucker algorithm is to find a geometry that is similar to `g` but uses only a subset of points. The algorithm may produce self-intersections and therefore result in invalid geometries.

If the `max_distance` is not positive or is NULL, an ER_WRONG_ARGUMENT will occur. The same function is also accessible using `simplify(Geometry g1, double max_distance)`.

## Release Notes
Adds the function:
`ST_Simplify(Geometry g, double max_distance)`, `Simplify(Geometry g, double max_distance)` :  returns the simplified version of g using the Ramer–Douglas–Peucker algorithms.

## How can this PR be tested?

## PR quality check


grooverdan:
Note, new error messages go at the end as they are numbered in order of the file. This seems to be the cause of most MTR test failures.

holyfoot (approved):
Looks ok to push.

PR #3378, opened 2024-07-01
MDEV-34502 InnoDB debug mode build - asserts with Valgrind

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34502*

## Description

\Valgrind looks at the assertions as examining uninitialized values.

As the assertions are tested in other Debug builds we know it isn't all invalid.

Account for Valgrind by removing the assertion under the WITH_VALGRIND=1 compilation.

## Release Notes

N/A

## How can this PR be tested?

Valgrind builder with newer 3.23.0 version run leaktest like:

 mysql-test/mysql-test-run.pl --valgrind="--leak-check=summary --gen-suppressions=yes --num-callers=10"  innodb.alter_not_null

## PR quality check


grooverdan:
> should clear out the last byte

Is negated *~*0xFFFFU,  so it clears, all except the last uninitialized byte. Technically it does read the uninit value, it just doesn't make such sense to report this as an error as the assertion purpose is to detect this. Technically valgrind can't tell the purpose of an `assert`.

PR #3379, opened 2024-07-01
MDEV-34503 Archive Storage Engine fails to compile with gcc-14

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34503*

## Description

gcc has started to recognise the difference between NULL and 0.

As such the NULL_Z defination in the azlib.h is an inconsistent uplicate.

As the function return an integer, make the return value 0 rather than Z_NULL.

## Release Notes
Archive storage engine can now compile with gcc-14.

Or nothing as its compile only.

## How can this PR be tested?

Compile with gcc-14.

## PR quality check

### DIFF
--- a/storage/archive/azio.c
+++ b/storage/archive/azio.c
@@ -99,7 +99,7 @@ int az_open (azio_stream *s, const char *path, int Flags, File fd)
     if (err != Z_OK)
     {
       destroy(s);
-      return Z_NULL;
+      return 0;
     }
   } else {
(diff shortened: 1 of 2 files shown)


cvicentiu:
@grooverdan I can't reproduce the problem. What's the zlib version you have? It doesn't seem to be gcc 14.1 specific.

grooverdan:
I tried and couldn't reproduce it either. Thanks for looking.

PR #3384, opened 2024-07-03
MDEV-34137: Implement the GIS function ST_Validate

StefanoPetrilli:
- [x] *The Jira issue number for this PR is: [MDEV-34137](https://jira.mariadb.org/browse/MDEV-34137)

## Description

MariaDB GIS functionalities are limited when compared to other DBMs. To facilitate the migration from MySQL to MariaDB the GIS functionality should be the same. This PR implements the GIS function `ST_Validate(Geometry g)`.  The function verifies that the input geometry:
- Is compliant with the [Well-Known Binary(WKB)](https://mariadb.com/kb/en/well-known-binary-wkb-format/) format and [Spatial Reference System Identifier [(SRID)](https://mariadb.com/kb/en/st_srid/) syntax.
- Is geometrically valid

If the input is valid return it, else it returns `NULL`. The use case of this function is to filter out invalid geometry data.

The following exceptions apply:

- If an input Polygon or MultiPolygon has clockwise rings, those rings are reversed before checking for validity. The value with the reversed rings is returned if the geometry is valid.
- The only valid empty geometry is represented as an empty geometry collection value. ST_Validate() returns it directly without further checks in this case.
- If the geometry has a geographic SRS with a longitude or latitude that is out of range, an error occurs. If a longitude value is not in the range (−180, 180], an ER_GEOMETRY_PARAM_LONGITUDE_OUT_OF_RANGE error occurs. If a latitude value is not in the range [−90, 90], an ER_GEOMETRY_PARAM_LATITUDE_OUT_OF_RANGE error occurs.

## Release Notes
Adds the function:
`ST_Validate(Geometry g)`, `Validate(Geometry g)` : the function returns the argument if the argumentis valid, otherwise returns `NULL`.

## How can this PR be tested?

## PR quality check


holyfoot (approved):
ok to push.

PR #3385, opened 2024-07-04
MDEV-34528: bundle fmt version 11.0.0

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34528*

## Description

Numerous fixes included.

## Release Notes

Bundled libfmt version is now 11.0.0. This applies only if not linked to a distro package of libfmt.

## How can this PR be tested?

mtr included. fmt changes by upstream devs left to their judgement :-)

- [X] *This is a bump of a bundled package*

## PR quality check


cmake/libfmt.cmake
```diff
@@ -15,10 +15,10 @@ MACRO(BUNDLE_LIBFMT)
   ExternalProject_Add(
     libfmt
     PREFIX   "${dir}"
-    URL      "https://github.com/fmtlib/fmt/archive/refs/tags/8.0.1.zip"
-    URL_MD5  e77873199e897ca9f780479ad68e25b1
+    URL      "https://github.com/fmtlib/fmt/archive/refs/tags/11.0.0.zip"
+    URL_MD5  f690d14b38d0fa473ea414ecf4e9c1a2
+    CMAKE_ARGS -DFMT_UNICODE=ON
```
vaintroub, line 20: I'm not exactly sure what you're trying to do here. Run CMake, but neither build or install? Can you explain the effect?

If it is just header-only library, I would think you just need to unpack it.

If you want to do the way  libfmt authors supposed it to be used, simplest way is probably FetchContent, as here https://stackoverflow.com/a/66531890,  which will download, build, install, import config, so below  the only thing needed to use library (no include directories, or path to binaries needed)
```cmake
target_link_libraries(sql fmt::fmt)
```

sql/CMakeLists.txt
```diff
@@ -211,6 +211,9 @@ ENDIF()
 IF(WIN32)
   SET(SQL_SOURCE ${SQL_SOURCE} handle_connections_win.cc winmain.cc)
 ENDIF()
+IF(MSVC)
+  ADD_COMPILE_FLAGS(item_strfunc.cc COMPILE_FLAGS "/utf-8")
```
vaintroub, line 215: Not sure why it is needed, can you explain?  I think /utf-8 is fine option to use generally, and so no use *not* to use them, our sources are already prepared for that, since 1db4fc543bba0cadbdcecdd82a6584875559a667

If it is needed, it is best to add them generally, not just to a single file. Reason is that compiler would need less invocations - 
MSVC compiler cl.exe, as it is called from VS Generator, will compile multiple source files of a library with a single compiler invocation, but only for the files for which all compile options are the same.,

With that, you can add options e.g like this (also using "modern CMake" supported way)
```diff
diff --git a/cmake/os/Windows.cmake b/cmake/os/Windows.cmake
index f951eed7651..ec1029e20ac 100644
--- a/cmake/os/Windows.cmake
+++ b/cmake/os/Windows.cmake
@@ -273,6 +273,7 @@ IF(MSVC)
     STRING(APPEND CMAKE_CXX_FLAGS_RELEASE " /d2OptimizeHugeFunctions")
     STRING(APPEND CMAKE_CXX_FLAGS_RELWITHDEBINFO " /d2OptimizeHugeFunctions")
   ENDIF()
+  ADD_COMPILE_OPTIONS($<$<COMPILE_LANGUAGE:C,CXX>:/utf-8>)
 ENDIF()
```

vaintroub (changes requested):
Asked some questions inline, could you please answer?

cmake/libfmt.cmake
grooverdan: this was enabling the utf flag in the headers:

https://github.com/fmtlib/fmt/blob/master/CMakeLists.txt#L360-L363

Which  becomes redundant once its used in a generator

previous attempts: https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F3385%2Fmerge

sql/CMakeLists.txt
grooverdan: sure, can do that.

cmake/libfmt.cmake
vaintroub: Oh, I see . This /utf-8 would be enabled automatically if we used target_link_libraries(sql fmt) , but we dont 

Looked up the where /utf-8 error was coming from in previous builds

Author assumes that 
```c++
unsigned char buffer[]="\u00A0"
```
has `buffer` encoded in UTF8. Seem too strong for general case, why would gcc have -finput-charset and -fexec-charset

Anyway, just enabling CONFIGURE would not do anything good for us, and it is good you removed it again

vaintroub (approved):
Looks good to me. Thanks!

sql/CMakeLists.txt
vuvova: @vaintroub should then the file require an appropriately modern CMake? If you're using modern CMake syntax here. Only on Windows, I mean.

vaintroub: We're guaranteed to have CMake version around, with MSVC. The VS2022 comes with at least [3.24](https://learn.microsoft.com/en-us/cpp/overview/what-s-new-for-visual-cpp-in-visual-studio?view=msvc-170) (currently 3.28), VS2019 comes with [3.19](https://learn.microsoft.com/en-us/cpp/overview/what-s-new-for-cpp-2019?view=msvc-170)

older CMakes do not even know VS2022 or VS2019 generators.
So, on Windows it is guaranteed, but if needed I can add an assertion. 

However we need to update general CMake to something greater than 2.8.12 in general case, not on Windows, not on MSVC too, at some point, now that centos-7 is dead . We already are using something deprecated, and CMake is already complaining.

> CMake Deprecation Warning at libmariadb/CMakeLists.txt:5 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

vuvova: generally, yes. but here I specifically mean to add something like
```cmake
CMAKE_MINIMUM_REQUIRED(VERSION 3.19)
```
to `Windows.cmake`.

On the other hand, if all cmakes that support VS2019+ generators are fine — then may be we don't need that line indeed.

vaintroub: That line requires 3.0 iirc, external pcre requires 3.5 (this is in their CMakeLists.txt), the use of TARGET_OBJECTs of non-object library in sql/CMakeLists.txt requires iirc 3.14, so overall we might require the highest of these in the top-level CMakeLists.txt, for MSVC, but I think it is fine also to do nothing, until real problems occur.

PR #3386, opened 2024-07-04
MDEV-34519 innodb_log_checkpoint_now crashes when innodb_read_only is enabled

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34519*

## Description
During read only mode, InnoDB doesn't allow checkpoint to happen. So InnoDB should throw the warning when InnoDB tries to force the checkpoint when innodb_read_only = 1 or
innodb_force_recovery = 6.
## How can this PR be tested?
./mtr innodb.log_file_overwrite 

## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -18378,28 +18378,42 @@ wait_background_drop_list_empty(THD*, st_mysql_sys_var*, void*, const void*)
 Force innodb to checkpoint. */
 static
 void
-checkpoint_now_set(THD*, st_mysql_sys_var*, void*, const void* save)
+checkpoint_now_set(THD* thd, st_mysql_sys_var*, void*, const void* save)
 {
-	if (*(my_bool*) save) {
-		mysql_mutex_unlock(&LOCK_global_system_variables);
-
-		lsn_t lsn;
-
-		while (log_sys.last_checkpoint_lsn.load(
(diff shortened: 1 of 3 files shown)


grooverdan (approved):
lgtm

storage/innobase/handler/ha_innodb.cc
```diff
@@ -18378,9 +18378,21 @@ wait_background_drop_list_empty(THD*, st_mysql_sys_var*, void*, const void*)
...
+		if (srv_read_only_mode) {
+			push_warning_printf(
+			  thd, Sql_condition::WARN_LEVEL_WARN,
+			  HA_ERR_UNSUPPORTED,
+			  "InnoDB doesn't force checkpoint when %s",
+			  (srv_force_recovery == SRV_FORCE_NO_LOG_REDO)
+			  ? "innodb-force-recovery=6."
+			  : "innodb-read-only=1.");
+			return;
+		}
```
mariadb-DebarunBanerjee, line 18394: This is a suggestion for your consideration. We might want to avoid the if condition nesting as follows. 
```
if (!*(my_bool*) save)
  return;

if (srv_read_only_mode) {
  ...
  return;
}
...

mariadb-DebarunBanerjee (approved):
Thanks for fixing it @Thirunarayanan. Looks good to me. 
I have one suggestion.

PR #3387, opened 2024-07-05
MDEV-34099: AddressSanitizer running out of memory regardless of stack_thread size

grooverdan:

## Description

Address Sanitizer's know how to detect stack overrun, so there's no point in us doing it.

As evidenced by perfschema tests where signficant test failures because this function failed under ASAN (MDEV-33210).

Also, so since clang-16, we cannot assume much about how local variables are allocated on the stack (MDEV-31605).

Disabling check idea thanks to Sanja.

## Release Notes

Disable stack overrun detection when an alternate Address Sanitizer computation is used.

## How can this PR be tested?

With clang-16+ and -DWITH_ASAN=1 run the mtr --suite=perfschema, will now pass.

## PR quality check

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -7614,6 +7614,7 @@ __attribute__((optimize("-O0")))
 #endif
 check_stack_overrun(THD *thd, long margin, uchar *buf __attribute__((unused)))
 {
+#ifndef __SANITIZE_ADDRESS__
   long stack_used;
   DBUG_ASSERT(thd == current_thd);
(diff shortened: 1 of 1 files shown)


LinuxJedi (approved):
Interestingly, I think the way ASAN's stack sanitizer works would break this kind of checking that we do: https://github.com/google/sanitizers/wiki/AddressSanitizerUseAfterReturn

LinuxJedi:
Auto-merge enabled based on Sanja's review too.

PR #3392, opened 2024-07-10
MDEV-34542  Assertion `lock_trx_has_sys_table_locks(trx) == __null' failed in void row_mysql_unfreeze_data_dictionary(trx_t*)

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34542*

## Description
- During XA PREPARE, InnoDB releases the non-exclusive locks. But it fails to remove the non-exclusive table lock from the transaction table locks. In the mean time, main thread evicts the table from the LRU cache. While rollbacking the XA transaction, InnoDB iterates through the table locks to check whether it holds lock on any system tables and wrongly assumes the evicted table as system table since the table id is 0

Fix:
===
During XA PREPARE, remove the table locks of the transaction while releasing the non-exclusive locks.

## How can this PR be tested?
./mtr innodb.lock_release

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/lock_release.result
+++ b/mysql-test/suite/innodb/r/lock_release.result
@@ -1,0 +1,23 @@
+#
+# MDEV-34542  Assertion `lock_trx_has_sys_table_locks(trx) == __null'
+#   failed in void row_mysql_unfreeze_data_dictionary(trx_t*)
+#
+#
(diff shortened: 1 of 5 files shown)


mariadb-DebarunBanerjee (approved):
Looks good.

It is worth noting that we don't call lock_trx_table_locks_remove in lock_release in commit/rollback path . This is fine as we clear the vector immediately afterwards in trx_t::release_locks.

PR #3395, opened 2024-07-11
11.6 mdev 34278

StefanoPetrilli:

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check


holyfoot (changes requested):
MariaDB (and MySQL too) has the internal List<> implemetation.
Along with the MEM_ROOT mechanics it works much faster and safer than the std list.
So please modify your code to use that List type instead for the geometry storage.
I can recommend the send_show_create_event() function in sql/event.cc as an example for the List<> usage.
So that your Item_func_collect class would host the MEM_ROOT that it uses as an allocator for your lists and will be freed in the ::clear() method.

holyfoot (changes requested):
Issues i'd like to be fixed:
1. You made the Item_func_collect::mem_root to point to the thd->mem_root.
This is not nice as the function can be counted multiple times in one query whlie the thd->mem_root is only freed at the query's end.
So it's better to have just the devoted MEM_ROOT member that is freed in the ::cleanup()
2. Since we have the MEM_ROOT it's much faster to allocate the String objects there to. Just do like 'String *buffer= new (&mem_root) String(wkb-length())'
3. Since we collect the data in RAM we have to limit the possible size so we don't end up eating all the memory thus stopping the server. For the value of the limit i recommed the existing group_concat_max_len server variable.
You can get the value as thd->variables.group_concat_max_len. So sum the lengths of the strings copied and return NULL + warning if it gets bigger than the group_concat_max_len  value.

Issues 1. and 2. are optional. Can be pushed as it is now. We have to fix the 3. thoug.

StefanoPetrilli:
@holyfoot I pushed a new version that solves point number 3

DaveGosselin-MariaDB:
> @holyfoot I pushed a new version that solves point number 3

@StefanoPetrilli the `main.spatial_utility_function_collect` test fails because the result file wasn't updated, can you please attach those changes to the PR?  Thanks.

StefanoPetrilli:
> > @holyfoot I pushed a new version that solves point number 3
> 
> @StefanoPetrilli the `main.spatial_utility_function_collect` test fails because the result file wasn't updated, can you please attach those changes to the PR? Thanks.

@DaveGosselin-MariaDB I forgot to add that file to the changes. Now it should be all right.

PR #3396, opened 2024-07-12
MDEV-34543  Shutdown hangs while freeing the asynchronous i/o slots

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34543*

## Description
Problem:
========
 During shutdown, InnoDB tries to free the asynchronous I/O slots and hangs. The reason is that InnoDB disables asynchronous I/O before waiting for pending asynchronous I/O to finish.

buf_load(): InnoDB aborts the buffer pool load due to user requested shutdown and doesn't wait for the asynchronous read to get completed. This could lead to debug assertion in buf_flush_buffer_pool() during shutdown

Fix:
===
os_aio_free(): Should wait all read_slots and write_slots to finish before disabling the aio.

buf_load(): Should wait for pending read request to complete even though it was aborted.

## How can this PR be tested?
I could repeat the debug assertion in buf_flush_buffer_pool()

large buffer pool load queue. break though load process, break point in aio call back before its freed from the slots. continue load thread. Instigate shutdown without resuming AIO thread

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/lock_release.result
+++ b/mysql-test/suite/innodb/r/lock_release.result
@@ -1,0 +1,23 @@
+#
+# MDEV-34542  Assertion `lock_trx_has_sys_table_locks(trx) == __null'
+#   failed in void row_mysql_unfreeze_data_dictionary(trx_t*)
+#
+#
+CREATE TABLE t1 (c1 CHAR(1) ,c2 INT) ENGINE=INNODB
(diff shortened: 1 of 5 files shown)


grooverdan (approved):

mariadb-DebarunBanerjee (approved):
Good catch Thiru! The patch looks good and safe to me.

Thirunarayanan:
Pushed as a part of pr#3392. So closing this pr

PR #3400, opened 2024-07-15
MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-condition

DaveGosselin-MariaDB:
Emit warnings if finished processing input files before reaching the boundary conditions indicated by either `--stop-position` or `--stop-datetime`.

### DIFF
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -142,6 +142,7 @@ static char *charset= 0;
 static uint verbose= 0;
 
 static ulonglong start_position, stop_position;
+static const longlong stop_position_default= (longlong)(~(my_off_t)0);
 #define start_position_mot ((my_off_t)start_position)
 #define stop_position_mot  ((my_off_t)stop_position)
 
@@ -1684,8 +1685,8 @@ static struct my_option my_options[] =
    "Stop reading the binlog at position N. Applies to the last binlog "
    "passed on the command line.",
    &stop_position, &stop_position, 0, GET_ULL,
-   REQUIRED_ARG, (longlong)(~(my_off_t)0), BIN_LOG_HEADER_SIZE,
-   (ulonglong)(~(my_off_t)0), 0, 0, 0},
+   REQUIRED_ARG, stop_position_default, BIN_LOG_HEADER_SIZE,
+   (ulonglong)stop_position_default, 0, 0, 0},
   {"table", 'T', "List entries for just this table (affects only row events).",
    &table, &table, 0, GET_STR_ALLOC, REQUIRED_ARG,
    0, 0, 0, 0, 0, 0},
@@ -2937,7 +2938,20 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
               llstr(old_off,llbuff));
         goto err;
       }
-      // file->error == 0 means EOF, that's OK, we break in this case
+      // else file->error == 0 means EOF, that's OK, we break in this case
+
+      /*
+        Emit a warning in the event that we finished processing input
+        before reaching the boundary indicated by --stop-position.
+      */
+      if (((longlong)stop_position != stop_position_default) &&
+          stop_position > my_b_tell(file))
+      {
+          retval = OK_STOP;
+          warning("Did not reach stop position %llu before "
+                  "end of input", stop_position);
+      }
+
       goto end;
     }
     if ((retval= process_event(print_event_info, ev, old_off, logname)) !=
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -2847,6 +2847,7 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
   IO_CACHE cache,*file= &cache;
   uchar tmp_buff[BIN_LOG_HEADER_SIZE];
   Exit_status retval= OK_CONTINUE;
+  my_time_t last_ev_when= MY_TIME_T_MAX;
 
   if (logname && strcmp(logname, "-") != 0)
   {
@@ -2952,8 +2953,21 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
                   "end of input", stop_position);
       }
 
+      /*
+        Emit a warning in the event that we finished processing input
+        before reaching the boundary indicated by --stop-datetime.
+      */
+      if (stop_datetime != MY_TIME_T_MAX &&
+          stop_datetime > last_ev_when)
+      {
+          retval = OK_STOP;
+          warning("Did not reach stop datetime '%s' "
+                  "before end of input", stop_datetime_str);
+      }
+
       goto end;
     }
+    last_ev_when= ev->when;
     if ((retval= process_event(print_event_info, ev, old_off, logname)) !=
         OK_CONTINUE)
       goto end;
--- a/mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_datetime.test
+++ b/mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_datetime.test
@@ -1,0 +1,42 @@
+--echo
+--echo # MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-datetime
+--echo
+
+--source include/have_binlog_format_statement.inc
+
+--let ignored_output_file= $MYSQLTEST_VARDIR/tmp/warn_pos_test_file.out
+
+set timestamp=1000000000;
+CREATE TABLE t1(word VARCHAR(20));
+set timestamp=1000000010;
+INSERT INTO t1 VALUES ("abirvalg");
+set timestamp=1000000020;
+INSERT INTO t1 SELECT * FROM t1;
+--let MYSQLD_DATADIR= `select @@datadir;`
+flush logs;
+
+--echo Case: Default, must not see warning.
+--echo # MYSQL_BINLOG --short-form MYSQLD_DATADIR/master-bin.000001 --result-file=ignored_output_file
+--exec $MYSQL_BINLOG --short-form $MYSQLD_DATADIR/master-bin.000001 --result-file=$ignored_output_file 2>&1
+
+--echo Case: Stop datetime before EOF, must not see warning.
+--echo # MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:50' MYSQLD_DATADIR/master-bin.000001 --result-file=ignored_output_file
+--exec $MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:50' $MYSQLD_DATADIR/master-bin.000001 --result-file=$ignored_output_file 2>&1
+
+--echo Case: Stop datetime between records, must not see warning.
+--echo # MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:55' MYSQLD_DATADIR/master-bin.000001 --result-file=ignored_output_file
+--exec $MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:55' $MYSQLD_DATADIR/master-bin.000001 --result-file=$ignored_output_file 2>&1
+
+--echo Case: Stop datetime at EOF, must not see warning.
+--echo # MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:55' MYSQLD_DATADIR/master-bin.000001 --result-file=ignored_output_file
+--exec $MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:55' $MYSQLD_DATADIR/master-bin.000001 --result-file=$ignored_output_file 2>&1
+
+--echo Case: Stop datetime after EOF, must see warning.
+--echo # MYSQL_BINLOG --short-form --stop-datetime='2035-01-19 03:14:05' MYSQLD_DATADIR/master-bin.000001 --result-file=ignored_output_file
+--exec $MYSQL_BINLOG --short-form --stop-datetime='2035-01-19 03:14:05' $MYSQLD_DATADIR/master-bin.000001 --result-file=$ignored_output_file 2>&1
+
+DROP TABLE t1;
+
+--remove_file $ignored_output_file
+
+--echo # End of binlog_mysqlbinlog_warn_stop_datetime.test
--- a/mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
+++ b/mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
@@ -1,0 +1,26 @@
+--echo
+--echo # MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-condition
+--echo
+
+--let assert_file= $MYSQLTEST_VARDIR/tmp/warn_pos_test_file.out
+--let data_file= $MYSQLTEST_VARDIR/std_data/master-bin.000001
+
+--echo Case: Default stop position, WARNING must not appear
+--echo # MYSQL_BINLOG --short-form --start-position=4 mysql-test/std_data/master-bin.000001 --result-file=warn_pos_test_file.out 2>&1
+--exec $MYSQL_BINLOG --short-form --start-position=4 $data_file --result-file=$assert_file 2>&1
+
+--echo Case: Stop position before EOF, WARNING must not appear
+--echo # MYSQL_BINLOG --short-form --start-position=4 --stop-position=97 mysql-test/std_data/master-bin.000001 --result-file=warn_pos_test_file.out 2>&1
+--exec $MYSQL_BINLOG --short-form --start-position=4 --stop-position=97 $data_file --result-file=$assert_file 2>&1
+
+--echo Case: Stop position at EOF, WARNING must not appear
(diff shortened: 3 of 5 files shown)


client/mysqlbinlog.cc
```diff
@@ -2922,6 +2923,17 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
 
     Log_event* ev = Log_event::read_log_event(file, glob_description_event,
                                               opt_verify_binlog_checksum);
+
+    // If we're not in raw mode, then emit a warning in the event that we
+    // finished processing input before reaching the boundary indicated by
+    // --stop-position.
+    if (!opt_raw_mode &&  // --stop-position is ignored in raw mode
```
bnestere, line 2930: We shouldn't need to check for each event seen, but we should only check once after we have processed everything.  `process_event()` will return `OK_STOP` if it has encountered the stop position, and this `Exit_status` value should propagate up to main. So after we've dumped events from all files, we should be able to check if `stop_position` (and for consistency, probably `stop_datetime` too, as a last-minute thought :sweat_smile: And can you add a test for that too?) is set and `retval != OK_STOP`, then we warn.

client/mysqlbinlog.cc
```diff
@@ -2922,6 +2923,17 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
 
     Log_event* ev = Log_event::read_log_event(file, glob_description_event,
                                               opt_verify_binlog_checksum);
+
+    // If we're not in raw mode, then emit a warning in the event that we
```
bnestere, line 2927: Just a small formatting remark, but the general style (at least in replication code) for formatting multi-line comments is to wrap it like:
```
/*
  line 1
  line 2
*/
```

Though note, there are a few "legacy" files in the replication code base which unfortunately deviate in their own way, and we try to just keep the formatting consistent per-file. `mysqlbinlog.cc` uses the above though.

mysql-test/suite/rpl/t/rpl_warn_statement.test
```diff
@@ -0,0 +1,2 @@
+--source include/have_binlog_format_statement.inc
+--source rpl_warn.inc
```
bnestere, line 2: These tests would be better suited in the `binlog` suite (`./mysql-test/suite/binlog/t`). The `rpl` suite is used for tests which use a `master` and `slave`.  Then since it is a mysqlbinlog specific test, there is another convention of using the test prefix `binlog_mysqlbinlog_*`.

mysql-test/suite/rpl/t/rpl_warn.inc
```diff
@@ -0,0 +1,30 @@
...
+--let MYSQLD_DATADIR= `select @@datadir;`
+
+--echo Case: Stop position before EOF
+--exec $MYSQL_BINLOG --short-form --start-position=$position --stop-position=$before_eof $MYSQLD_DATADIR/master-bin.000001 2>&1
+
+--echo Case: Stop position at EOF
+--exec $MYSQL_BINLOG --short-form --start-position=$position --stop-position=$at_eof $MYSQLD_DATADIR/master-bin.000001 2>&1
+
+--echo Case: Stop position after EOF
+--exec $MYSQL_BINLOG --short-form --start-position=$position --stop-position=$after_eof $MYSQLD_DATADIR/master-bin.000001 2>&1
```
bnestere, line 28: Generally it is better to avoid writing event output information to stdout (i.e. the `.result` file) if we can. Since this test isn't really about the event data, you can use the `--result-file` option to direct stdout to the file (and keep `2>&1` so the error output goes there too). Then you can assert that the warning is (or is not) in the file as intended, e.g. using `--source include/assert_grep.inc` (my preferred way, as it is more intentional) or `--source include/search_pattern_in_file.inc` (which just outputs the number of times the pattern exists).  A few bonuses of this too, are that you don't need 1) three different tests, as then they won't be `binlog_format` dependent anymore (you'll need to remove the `SHOW MASTER STATUS` too) , nor 2) the `set timestamp` (though kudos for figuring out those patterns :) ) 

Note too, an example you can follow for method 1 is in `rpl_xa_empty_transaction.test` for test case `33921.3`.

And while there, can you also add one more pattern from that file, to `--echo` the `MYSQL_BINLOG` command before the `--exec`?

mysql-test/suite/rpl/t/rpl_warn.inc
```diff
@@ -0,0 +1,30 @@
...
+
+DROP TABLE t1;
```
bnestere, line 30: As the last line of the test, can you add a final echo for
```
--echo # End of <test_name>.test
```

bnestere (changes requested):
Hi @DaveGosselin-MariaDB!

Thanks for the PR! Please see my notes.

client/mysqlbinlog.cc
DaveGosselin-MariaDB: I see: we will know (1) if `stop_position` was set by the user and (2) whether we hit EOF first or `stop_position` first as in the latter case we will see `OK_STOP`; then we can deduce if we need to warn or not at the top-level.  The only trouble with this approach is that we cannot distinguish between when we've read exactly to EOF or when we've read past it, both cases will appear the same.  This is because `OK_STOP` is returned when the position is greater-than or equal to `stop_position`.  We could disambiguate with for example `OK_STOP_BEFORE_EOF` and `OK_STOP_AT_EOF`.
Yes, I'll add checks and tests for `stop_datetime` as well 👍

client/mysqlbinlog.cc
DaveGosselin-MariaDB: Thanks, will fix!

mysql-test/suite/rpl/t/rpl_warn_statement.test
DaveGosselin-MariaDB: Great idea, I will move them.

mysql-test/suite/rpl/t/rpl_warn.inc
DaveGosselin-MariaDB: Very nice advice, I will implement it.

client/mysqlbinlog.cc
bnestere: Great observation! I don't necessarily think we need the new status options though. I think we can just override `retval` to `OK_STOP` in such case. E.g., in `dump_local_log_entries`, at the line

```
      // file->error == 0 means EOF, that's OK, we break in this case
```
we check if we have hit our stop condition, and if so, set `retval= OK_STOP`. It seems more consistent this way too because stopping should be the same whether we are at or before the EOF (note `OK_EOF` isn't actually used for `dump_local_log_entries`; the alternative is `OK_CONTINUE`)

DaveGosselin-MariaDB: If we override `retval` to `OK_STOP` and in the case that `end_io_cache` returns a failure, we will inadvertently swallow the warning because `retval` will become `ERROR_STOP` (ed of `dump_local_log_entries`).  That may be fine for the user, but I'm not sure what the user expectations are; in the case of an error, should the user see any/all warnings that may have occurred?

bnestere: Yeah it'd be better not to swallow the warning, good point. And with the prior amendment where we are already adding a check after processing each binlog at
```
      // file->error == 0 means EOF, that's OK, we break in this case
```
we may as well write the warning there and avoid a redundant later check in `main`. We could still set `retval= OK_STOP` too, it seems more consistent.

DaveGosselin-MariaDB: I was just looking at doing that, we're on the same brainwave 😄

mysql-test/suite/rpl/t/rpl_warn.inc
DaveGosselin-MariaDB: In testing, the combined stdout and stderr don't end up in `--result-file`, but the warning does appear in the test's `.result` file.

DaveGosselin-MariaDB: I updated the PR to include the changes for `--stop-position` with the thought that I would submit a separate PR for `--stop-datetime`.  From a software archaeology perspective, it would be cleaner because the changes and tests will standalone from those for `--stop-position`.  Will that work?

client/mysqlbinlog.cc
```diff
@@ -2937,7 +2938,21 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
...
       }
-      // file->error == 0 means EOF, that's OK, we break in this case
+      // else file->error == 0 means EOF, that's OK, we break in this case
+
+      /*
+        If we're not in raw mode, then emit a warning in the event that we
+        finished processing input before reaching the boundary indicated by
+        --stop-position.
+      */
+      if (!opt_raw_mode &&  // --stop-position is ignored in raw mode
```
bnestere, line 2948: I don't think we need to check for raw mode, as I don't think we can actually be in this function when that is set (as its path needs a remote server, which goes to `dump_remote_log_entries()`).

client/mysqlbinlog.cc
```diff
@@ -142,6 +142,7 @@ static char *charset= 0;
 static uint verbose= 0;
 
 static ulonglong start_position, stop_position;
+static const longlong stop_position_default= (longlong)(~(my_off_t)0);
```
bnestere, line 145: Why `longlong`? Seems like it should be `ulonglong` like the original type.

client/mysqlbinlog.cc
```diff
@@ -1684,7 +1685,7 @@ static struct my_option my_options[] =
    "Stop reading the binlog at position N. Applies to the last binlog "
    "passed on the command line.",
    &stop_position, &stop_position, 0, GET_ULL,
-   REQUIRED_ARG, (longlong)(~(my_off_t)0), BIN_LOG_HEADER_SIZE,
+   REQUIRED_ARG, stop_position_default, BIN_LOG_HEADER_SIZE,
```
bnestere, line 1688: Let's make the `max_value` arg also use `stop_position_default` so it is clear from the go that the default is the max

client/mysqlbinlog.cc
```diff
@@ -2937,7 +2938,21 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
...
+      // else file->error == 0 means EOF, that's OK, we break in this case
+
+      /*
+        If we're not in raw mode, then emit a warning in the event that we
+        finished processing input before reaching the boundary indicated by
+        --stop-position.
+      */
+      if (!opt_raw_mode &&  // --stop-position is ignored in raw mode
+          ((longlong)stop_position != stop_position_default) &&
+          stop_position > my_b_tell(file)) {
```
bnestere, line 2950: Just formatting, but the  `{` should be on a new line.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
```diff
@@ -0,0 +1,20 @@
+--echo
+--echo # MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-condition
+--echo
+
+--let assert_file= $MYSQLTEST_VARDIR/tmp/warn_pos_test_file.out
+--let data_file= $MYSQLTEST_VARDIR/std_data/master-bin.000001
+
+--echo Case: Stop position before EOF, WARNING must not appear
+--echo MYSQL_BINLOG --short-form --start-position=4 --stop-position=97 mysql-test/std_data/master-bin.000001 --result-file=warn_pos_test_file.out 2>&1
```
bnestere, line 9: Nice and simple test, I like it :) Do we still need the `2>&1` for the warning to show up in the test result output? Let's also clean up the result file at the end of the test (with MTR command `--remove_file`).

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
```diff
@@ -0,0 +1,20 @@
+--echo
+--echo # MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-condition
+--echo
+
+--let assert_file= $MYSQLTEST_VARDIR/tmp/warn_pos_test_file.out
+--let data_file= $MYSQLTEST_VARDIR/std_data/master-bin.000001
+
+--echo Case: Stop position before EOF, WARNING must not appear
```
bnestere, line 8: For the `--echo`'s can you prefix the output with some indicator to show the origin is from an echo? `#` is pretty common (I think because there was an original idea of being able to use the `.result` file as input to `mariadbd`, but I've never done it myself or seen anyone actually do that), but I don't think it matters much as long as it is clear. It makes it easier to distinguish when reading the `.result` file.

bnestere (changes requested):
Thanks @DaveGosselin-MariaDB for a quick update, the patch is coming along nicely :smile:. I left a few more notes.

mysql-test/suite/rpl/t/rpl_warn.inc
bnestere: I'm all for a cleaner software history. Though I think the `--stop-position` and `--stop-datetime` changes should be pushed at the same time for a consistent code base. What about using the same PR but have a separate commit?

client/mysqlbinlog.cc
DaveGosselin-MariaDB: I thought the same and tried it, but the compiler complains about type narrowing in the argument parsing.  We tell the arg parser that it is a ULL type which should be sufficient.  We could explore that further; my factoring preserves the original behavior.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
```diff
@@ -0,0 +1,20 @@
+--echo
+--echo # MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-condition
+--echo
+
+--let assert_file= $MYSQLTEST_VARDIR/tmp/warn_pos_test_file.out
+--let data_file= $MYSQLTEST_VARDIR/std_data/master-bin.000001
```
bnestere, line 6: Though another thought comes to me now. In `10.11`, `--stop-position` also supports a GTID value, and this test binlog file doesn't have any transactions. We should have a `10.11` specific patch to also warn when a GTID specified is not hit. And for that we will need transactions. It might be easier if we prepare for that here, and revert  back to something like your last test file which used its own binlog.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
DaveGosselin-MariaDB: Thanks! 😊 You're right, we no longer need `2>&1`.  Didn't know about `--remove_file`, will clean it up.

DaveGosselin-MariaDB: Oops, I spoke too soon; we need `2>&1` 😅

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
DaveGosselin-MariaDB: Great idea, I didn't know about that additional support in 10.11. 👍

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
DaveGosselin-MariaDB: Definitely, I'll prepend it with `#` since that's what's common in the codebase already.

mysql-test/suite/rpl/t/rpl_warn.inc
DaveGosselin-MariaDB: Sounds good!  I'm most interested in keeping them in separate commits; using the same PR works great for that 👍

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
DaveGosselin-MariaDB: I created https://jira.mariadb.org/browse/MDEV-34614 to capture this work as it's in a different version of the codebase.

DaveGosselin-MariaDB:
I need to investigate why some of the builders failed on my new test for the second commit while others did not.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_datetime.test
```diff
@@ -0,0 +1,37 @@
+--echo
+--echo # MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-datetime
+--echo
+
+--source include/have_binlog_format_statement.inc
+
+--let assert_file= $MYSQLTEST_VARDIR/tmp/warn_pos_test_file.out
```
bnestere, line 7: I think the name `assert_file` makes it seem like the file itself has content that will be validated, esp hinting towards `assert_grep.inc`. But it is actually ignored. I think the name should set the expectation that the file is ignored.

bnestere (approved):
Looks good, thank you @DaveGosselin-MariaDB ! I just had one last minor thought for cleanup.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_datetime.test
DaveGosselin-MariaDB: Thanks, I made that change and added `--echo` commands that reflect what's `--exec`'d (to have parity with the `...warn_stop_position.test` file).

PR #3401, opened 2024-07-15
MDEV-34680 Asynchronous and Buffered Logging in MySQL Audit Plugin

HashimSyedUBC:
## Description

1. **Buffered Logging**:

Implement an in-memory buffer to reduce disk I/O by batching log writes.

2. **Asynchronous Logging**:

Delegate logging to a dedicated logger thread for independent log handling,
minimizing wait times for main audit threads

3. **Dynamic Buffer Management**:

Add `log_buffer_size` and `log_buffer_time` system variables for user-customizable
flushing based on data volume or time, adapting to varying workloads.

4. **Confirm Performance Improvement**:

An average of 15% speed improvements confirmed in benchmark tests, demonstrating the
effectiveness of these logging enhancements in high-load environments.

## Release Notes

N/A - Asynchronous and Buffered Logging.

## How can this PR be tested?

The mtr tests server_audit.test, server_audit_buffer_size.test, and server_audit_buffer_time.test have been updated to account for asynchronous and buffered logging.

## PR quality check

### DIFF
--- a/plugin/server_audit/server_audit.cc
+++ b/plugin/server_audit/server_audit.cc
@@ -283,8 +283,10 @@ static my_off_t loc_tell(File fd)
 #define localtime_r(a, b) localtime_s(b, a)
 #endif /*WIN32*/
 
-
-extern MYSQL_PLUGIN_IMPORT char server_version[];
+extern "C"
+{
+  extern MYSQL_PLUGIN_IMPORT char server_version[];
+}
 static const char *serv_ver= NULL;
 const char *(*thd_priv_host_ptr)(MYSQL_THD thd, size_t *length);
 static int started_mysql= 0;
@@ -550,7 +552,10 @@ static long log_write_failures= 0;
 static char current_log_buf[FN_REFLEN]= "";
 static char last_error_buf[512]= "";
 
-extern void *mysql_v4_descriptor;
+extern "C"
+{
+  extern void *mysql_v4_descriptor;
+}
 
 static struct st_mysql_show_var audit_status[]=
 {
@@ -558,7 +563,7 @@ static struct st_mysql_show_var audit_status[]=
   {"server_audit_current_log", current_log_buf, SHOW_CHAR},
   {"server_audit_writes_failed", (char *)&log_write_failures, SHOW_LONG},
   {"server_audit_last_error", last_error_buf, SHOW_CHAR},
-  {0,0,0}
+  {0,0, SHOW_UNDEF}
 };
 
 #ifdef HAVE_PSI_INTERFACE
@@ -724,9 +729,9 @@ static int coll_insert(struct user_coll *c, char *n, size_t len)
   {
     c->n_alloced+= 128;
     if (c->users == NULL)
-      c->users= malloc(c->n_alloced * sizeof(c->users[0]));
+      c->users= (user_name*)malloc(c->n_alloced * sizeof(user_name));
     else
-      c->users= realloc(c->users, c->n_alloced * sizeof(c->users[0]));
+      c->users= (user_name*)realloc(c->users, c->n_alloced * sizeof(user_name));
 
     if (c->users == NULL)
       return 1;
@@ -1848,7 +1853,7 @@ do_log_query:
   if (query_len > (message_size - csize)/2)
   {
     size_t big_buffer_alloced= (query_len * 2 + csize + 4095) & ~4095L;
-    if(!(big_buffer= malloc(big_buffer_alloced)))
+    if(!(big_buffer = (char*)malloc(big_buffer_alloced)))
       return 0;
 
     memcpy(big_buffer, message, csize);
@@ -2135,7 +2140,7 @@ struct connection_info cn_error_buffer;
 
 
 #define FILTER(MASK) (events == 0 || (events & MASK))
-void auditing(MYSQL_THD thd, unsigned int event_class, const void *ev)
+extern "C" void auditing(MYSQL_THD thd, unsigned int event_class, const void *ev)
 {
   struct connection_info *cn= 0;
   int after_action= 0;
@@ -2296,8 +2301,10 @@ struct mysql_event_general_v8
 };
 
 
-static void auditing_v8(MYSQL_THD thd, struct mysql_event_general_v8 *ev_v8)
+static void auditing_v8(MYSQL_THD thd, unsigned int event_class, const void *ev)
 {
+  struct mysql_event_general_v8 *ev_v8 = (struct mysql_event_general_v8 *) ev;
+
 #ifdef __linux__
 #ifdef DBUG_OFF
   #ifdef __x86_64__
@@ -2366,9 +2373,9 @@ static void auditing_v8(MYSQL_THD thd, struct mysql_event_general_v8 *ev_v8)
 }
 
 
-static void auditing_v13(MYSQL_THD thd, unsigned int *ev_v0)
+static void auditing_v13(MYSQL_THD thd, unsigned int event_class, const void *ev_v0)
 {
-  struct mysql_event_general event= *(const struct mysql_event_general *) (ev_v0+1);
+  struct mysql_event_general event= *(const struct mysql_event_general *) (((const unsigned int *) ev_v0)+1);
 
   if (event.general_query_length > 0)
   {
@@ -2376,11 +2383,11 @@ static void auditing_v13(MYSQL_THD thd, unsigned int *ev_v0)
     event.general_command= "Query";
     event.general_command_length= 5;
   }
-  auditing(thd, ev_v0[0], &event);
+  auditing(thd, ((const unsigned int *) ev_v0)[0], &event);
 }
 
 
-int get_db_mysql57(MYSQL_THD thd, char **name, size_t *len)
+extern "C" int get_db_mysql57(MYSQL_THD thd, char **name, size_t *len)
 {
 #ifdef __linux__
   int db_off;
@@ -2511,7 +2518,7 @@ static int server_audit_init(void *p __attribute__((unused)))
 {
   if (!serv_ver)
   {
-    serv_ver= find_sym("server_version");
+    serv_ver= (const char*)find_sym("server_version");
   }
 
   if (!mysql_57_started)
@@ -2525,12 +2532,12 @@ static int server_audit_init(void *p __attribute__((unused)))
     if (!my_hash_init_ptr)
       return 1;
 
-    thd_priv_host_ptr= dlsym(RTLD_DEFAULT, "thd_priv_host");
+    thd_priv_host_ptr= (const char* (*)(THD*, size_t*)) dlsym(RTLD_DEFAULT, "thd_priv_host");
   }
 
-  if(!(int_mysql_data_home= find_sym("mysql_data_home")))
+  if(!(int_mysql_data_home= (char**)find_sym("mysql_data_home")))
   {
-    if(!(int_mysql_data_home= find_sym("?mysql_data_home@@3PADA")))
+    if(!(int_mysql_data_home= (char**)find_sym("?mysql_data_home@@3PADA")))
       int_mysql_data_home= &default_home;
   }
 
@@ -3090,12 +3097,12 @@ void __attribute__ ((constructor)) audit_plugin_so_init(void)
       if (sc <= 10)
       {
         mysql_descriptor.interface_version= 0x0200;
-        mysql_descriptor.event_notify= (void *) auditing_v8;
+        mysql_descriptor.event_notify= auditing_v8;
       }
       else if (sc < 14)
       {
         mysql_descriptor.interface_version= 0x0200;
-        mysql_descriptor.event_notify= (void *) auditing_v13;
+        mysql_descriptor.event_notify= auditing_v13;
       }
     }
     else if (serv_ver[0] == '5' && serv_ver[2] == '6')
--- a/plugin/server_audit/server_audit.cc
+++ b/plugin/server_audit/server_audit.cc
@@ -21,7 +21,18 @@
 #define _my_thread_var loc_thread_var
 
 #include <my_config.h>
+#include <condition_variable>
+#include <mutex>
+#include <queue>
+#include <future>
+#include <chrono>
+#include <vector>
+#include <string>
 #include <assert.h>
+
+typedef char my_bool;
+typedef unsigned long long int ulonglong; /* ulong or unsigned long long */
+#include <thr_timer.h>
 
 #ifndef _WIN32
 #define DO_SYSLOG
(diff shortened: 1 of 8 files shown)


LinuxJedi:
I know this is a WIP, but just in case you missed it, there are compiler errors with this PR:

```
/home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c: In function ‘void flush_buffer(const std::vector<std::__cxx11::basic_string<char> >&)’:
/home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:1096:20: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::__cxx11::basic_string<char> >::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
 1096 |   for (int i= 0; i < all_messages.size(); i++)
      |                  ~~^~~~~~~~~~~~~~~~~~~~~
/home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c: In function ‘void signal_log(const string&)’:
/home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:1189:22: error: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Werror=sign-compare]
 1189 |   if (log_queue_size >= max_size_log_queue)
      |       ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c: In function ‘void audit_plugin_so_init()’:
/home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:3267:84: error: cast between incompatible function types from ‘void (*)(THD*, mysql_event_general_v8*)’ to ‘void (*)(THD*, unsigned int, const void*)’ [-Werror=cast-function-type]
 3267 |         mysql_descriptor.event_notify= (void (*)(THD*, unsigned int, const void*)) auditing_v8;
      |                                                                                    ^~~~~~~~~~~
/home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:3272:84: error: cast between incompatible function types from ‘void (*)(THD*, unsigned int*)’ to ‘void (*)(THD*, unsigned int, const void*)’ [-Werror=cast-function-type]
 3272 |         mysql_descriptor.event_notify= (void (*)(THD*, unsigned int, const void*)) auditing_v13;
      |                                                                                    ^~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [plugin/server_audit/CMakeFiles/server_audit.dir/build.make:63: plugin/server_audit/CMakeFiles/server_audit.dir/server_audit.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:9048: plugin/server_audit/CMakeFiles/server_audit.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

HashimSyedUBC:
> I know this is a WIP, but just in case you missed it, there are compiler errors with this PR:
> 
> ```
> /home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c: In function ‘void flush_buffer(const std::vector<std::__cxx11::basic_string<char> >&)’:
> /home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:1096:20: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::__cxx11::basic_string<char> >::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
>  1096 |   for (int i= 0; i < all_messages.size(); i++)
>       |                  ~~^~~~~~~~~~~~~~~~~~~~~
> /home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c: In function ‘void signal_log(const string&)’:
> /home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:1189:22: error: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Werror=sign-compare]
>  1189 |   if (log_queue_size >= max_size_log_queue)
>       |       ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
> /home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c: In function ‘void audit_plugin_so_init()’:
> /home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:3267:84: error: cast between incompatible function types from ‘void (*)(THD*, mysql_event_general_v8*)’ to ‘void (*)(THD*, unsigned int, const void*)’ [-Werror=cast-function-type]
>  3267 |         mysql_descriptor.event_notify= (void (*)(THD*, unsigned int, const void*)) auditing_v8;
>       |                                                                                    ^~~~~~~~~~~
> /home/buildbot/amd64-ubuntu-2004-debug/build/plugin/server_audit/server_audit.c:3272:84: error: cast between incompatible function types from ‘void (*)(THD*, unsigned int*)’ to ‘void (*)(THD*, unsigned int, const void*)’ [-Werror=cast-function-type]
>  3272 |         mysql_descriptor.event_notify= (void (*)(THD*, unsigned int, const void*)) auditing_v13;
>       |                                                                                    ^~~~~~~~~~~~
> cc1plus: all warnings being treated as errors
> make[2]: *** [plugin/server_audit/CMakeFiles/server_audit.dir/build.make:63: plugin/server_audit/CMakeFiles/server_audit.dir/server_audit.c.o] Error 1
> make[1]: *** [CMakeFiles/Makefile2:9048: plugin/server_audit/CMakeFiles/server_audit.dir/all] Error 2
> make[1]: *** Waiting for unfinished jobs....
> ```

This problem is now resolved.

HashimSyedUBC:
Currently Investigating: The server_audit_buffer_size.test is failing inconsistently on the CI but passing on my local machine. The hypothesis is that there might be a race condition causing a very small amount of logs to be displaced. Note that the logs are not missing but rather appear in a later position in the file.

Would love any feedback on the implementation.

LinuxJedi:
> Currently Investigating: The server_audit_buffer_size.test is failing inconsistently in the CI environment but passing on my local machine. The hypothesis is that there might be a race condition causing a very small number of logs to be displaced. Note that the logs are not missing but rather appear in a later position in the file.

That is likely to be expected with something that is asynchronous.

> Update: This race condition also shows up in the server_audit.test on CI when testing with the default original synchronous direct logging implementation. Race Condition is likely not due to this implementation.

Are you sure? I have not seen evidence of this, and it isn't on own know failures list.

LinuxJedi:
I'll get someone to look over the implementation of this. But some versioning changes will likely need to be made. I'm specifically thinking about `VERSION_logger` in `include/service_versions.h` and the version `plugin_audit_v4.h`.

vaintroub:
I think "dedicated thread" that is doing nothing, just waiting most of the time is a wrong idea. MySQL has multiplied threads for anything possible, "a thread that is waiting for another thread and signals third thread", I would not be like them and multiply threads without necessity. In fact, MariaDB already has timers in mysys, and I think it makes sense to explore them. Whenever you write to buffer at position 0, and not immediately flush, you can set up timer that does that for you, in 1 sec.

HashimSyedUBC:
> I think "dedicated thread" that is doing nothing, just waiting most of the time is a wrong idea. MySQL has multiplied threads for anything possible, "a thread that is waiting for another thread and signals third thread", I would not be like them and multiply threads without necessity. In fact, MariaDB already has timers in mysys, and I think it makes sense to explore them instead. It is just 3 functions you'd need - `thr_timer_init`, `thr_timer_settime`, `thr_timer_end`. Whenever you write to buffer at position 0, and not immediately flush, you can do `thr_timer_setttime` that flushes buffer for you, in 1 sec from now, or whatever the timeout is. The job will still be performed by another thread (yes, there is a dedicated timer thread, that handles all timers), but this one already exists,it has started, and you do not need to repeat its job.

Thank you so much for the feedback. I will be looking into this ASAP.

HashimSyedUBC:
@LinuxJedi @vaintroub I have now updated the branch with your suggestions. The details about the implementation and its benchmarking scores can be found on this [JIRA](https://jira.mariadb.org/browse/MDEV-34680?filter=-2). Please let me know if you any other suggestions. Thank you !

plugin/server_audit/server_audit.cc
```diff
@@ -1121,6 +1273,7 @@ static int start_logging()
     strncpy(current_log_buf, "[SYSLOG]", sizeof(current_log_buf)-1);
     compile_time_assert(sizeof current_log_buf > sizeof "[SYSLOG]");
   }
+  thr_timer_set_period(&audit_timer, log_buffer_time*1000);
```
vaintroub, line 1276: You do not really need periodic timer, as far as I can see  Periodic will tick, even if there is no work to do, this is a bit suboptimal. On the other hand, I think you know exactly when set that timer - if you the push first message to  queue, without flushing.

plugin/server_audit/server_audit.cc
```diff
@@ -1055,6 +1085,128 @@ static int get_user_host(const char *uh_line, unsigned int uh_len,
...
+static void resize_flush(size_t new_size) 
+{
+  std::unique_lock<std::mutex> lock(queue_mutex);
+  if (log_queue.size() != 0) 
+  {
+      flush_buffer(log_queue);
+  }
+  log_queue.clear();
+  max_size_log_queue= new_size;
+  thr_timer_settime(&audit_timer, log_buffer_time*1000);
```
vaintroub, line 1146: The queue is empty now, I think you do not need this settime here.

plugin/server_audit/server_audit.cc
```diff
@@ -1055,6 +1085,128 @@ static int get_user_host(const char *uh_line, unsigned int uh_len,
...
+
+void push_log(const std::string& message) 
+{
+  std::unique_lock<std::mutex> lock(queue_mutex);  
+  log_queue.emplace_back(message);
+  if (log_queue.size() >= max_size_log_queue)
+  {
+      flush_buffer(log_queue);
+      log_queue.clear();
+      thr_timer_settime(&audit_timer, log_buffer_time*1000);
```
vaintroub, line 1171: you only need `thr_timer_settime` if you added *very first* message, and you did *not* flush.

That is, only  start timer with ` thr_timer_settime`  if  log_queue.size()` == 1 after `emplace_back`, and `max_size_log_queue` > 1

Otherwise, either you flushed yourself, or, if  `log_queue.size()` > 1 there is already a pending timer

plugin/server_audit/server_audit.cc
```diff
@@ -2864,6 +2976,44 @@ static int check_excl_users(MYSQL_THD thd  __attribute__((unused)),
...
+{
+  log_buffer_time= *(unsigned long long *) save;
+  error_header();
+  fprintf(stderr, "Log buffer time was changed to '%lld'.\n", 
+  log_buffer_time);
+  if (!logging || output_type != OUTPUT_FILE) 
+    return;
+
+  mysql_prlock_wrlock(&lock_operations);
+  thr_timer_set_period(&audit_timer, log_buffer_time*1000);
```
vaintroub, line 3010: You do not need that either :)

vaintroub (changes requested):
This is much better, thanks! you got rid of extra thread that I was complaining about. 

But, as I commented,  there is
1. No reason for periodic timer, you know when to start it.
2. If only need to call thr_timer_settime in one place, when you push first message to queue, without flushing it yourself.

I think I got the logic right, or is there something I'd miss?

HashimSyedUBC:
> This is much better, thanks! you got rid of extra thread that I was complaining about.
> 
> But, as I commented, there is
> 
> 1. No reason for periodic timer, you know when to start it.
> 2. You _only_ need to call thr_timer_settime in one case, when you push _first_ message to queue, without flushing it yourself.
> 
> I think I got the logic right, or is there something I'd miss?

Hey @vaintroub,

Thank you for your suggestions. I have updated the implementation accordingly. Now, the audit plugin only calls thr_timer_init(), thr_timer_end, and thr_timer_settime (once). I've tested this on a local server, and everything is working well. Please let me know if you have any further comments.

plugin/server_audit/server_audit.cc
```diff
@@ -1055,6 +1085,130 @@ static int get_user_host(const char *uh_line, unsigned int uh_len,
...
+  std::unique_lock<std::mutex> lock(queue_mutex);  
+  log_queue.emplace_back(message);
+  if (log_queue.size() >= max_size_log_queue)
+  {
+      flush_buffer(log_queue);
+      log_queue.clear();
+      return;
+  }
+  
+  if (log_queue.size() == 1 && audit_timer.expired)
```
vaintroub, line 1173: I was wondering if you need to test this audit_timer.expired flag, then looked at the sources to find out that thr_timer_setting on non-expired timer is forbidden with a DBUG_ASSERT. I do not know the reasons behind that, but you could also do thr_timer_end (which does not destroy the timer itself, only removes from timer queue, if it is pending) prior to thr_timer_settime . This would be a tiny bit more thread-safe (all checks for "expired" flag under the same mutex). But I would not argue, even if you leave it as it is now.  Leaving it up to you.

vaintroub (approved):
Looks good to me. Hope others will comment on other aspects of the patch, I just looked into how timer is used.

holyfoot (changes requested):
This buffering improves the performance in some workloads indeed.
Though i'd really prefer it to be done in different style. The similar logging happens in some other places so it makes sense to move the buffering inside the logger_file so this can be employed by everybody. Also i don't like how the vector classes are extensively used. That leads to lots of heap allocations/deallocations which can consume resourses expecially in the context of the working server. It seems we can get by with one pre-allocated buffer per logger_file, or employ the IO_CACHE implemented in the server.

plugin/server_audit/server_audit.cc
```diff
@@ -340,6 +353,13 @@ struct connection_info
   int proxy_host_length;
 };
 
+std::vector<std::string> log_queue(1);
```
janlindstrom, line 356: This is not a good idea.  Instead of adding every sting to a vector, I would rather see one big buffer where strings are concatenated e.g.

  // We are using a pointer to string to guarantee that we can decrease the size
  // of the buffer
  using data_t = std::unique_ptr<std::string>;
  // memory buffer storing the log messages
  data_t data;

and then when you reserve it

  data_t new_buffer(new std::string());
  new_buffer->reserve(buffer_size);
  data.swap(new_buffer);

janlindstrom (changes requested):

plugin/server_audit/server_audit.cc
```diff
@@ -1050,6 +1085,130 @@ static int get_user_host(const char *uh_line, unsigned int uh_len,
 #define S_ISDIR(x) ((x) & _S_IFDIR)
 #endif /*_WIN32 && !S_ISDIR*/
 
+/*
+  Flushes the buffer of log strings to the file.
+
+  @param all_messages Contains the log strings
+*/
+void flush_buffer(const std::vector<std::string>& all_messages)
```
janlindstrom, line 1093: I would push this filename and buffer handling to file_logger.c so that other plugins and/or server code could use buffered logging.

vuvova:
MDEV-34680 is already done, pushed and released in 12.1.1.
Main functionality is in `file_logger.cc` as @janlindstrom suggested above. And it uses `IO_CACHE` instead of implementing yet another caching logic, that made the implementation much simpler and smaller.

PR #3402, opened 2024-07-16
MDEV-32014 Rename binlog cache temporary file to binlog file for large transactions

SongLibing:
# Description
When a transaction commits, it copies the binlog events from
binlog cache to binlog file. Very large transactions
(eg. gigabytes) can stall other transactions for a long time
because the data is copied while holding LOCK_log, which blocks
other commits from binlogging.

The solution in this patch is to rename the binlog cache file to
a binlog file instead of copy, if the commiting transaction has
large binlog cache. Rename is a very fast operation, it doesn't
block other transactions a long time.

# Design
* binlog_large_commit_threshold
  type: ulonglong
  scope: global
  dynamic: yes
  default: 128MB

  Only the binlog cache temporary files large than 128MB are
  renamed to binlog file.

* #binlog_cache_files directory
  To support rename, all binlog cache temporary files are managed
  as normal files now. `#binlog_cache_files` directory is in the same
  directory with binlog files. It is created at server startup if it doesn't
  exist. Otherwise, all files in the directory is deleted at startup.

  The temporary files are named with ML_ prefix and the memorary address
  of the binlog_cache_data object which guarantees it is unique.

* Reserve space
  To supprot rename feature, It must reserve enough space at the
  begin of the binlog cache file. The space is required for
  Format description, Gtid list, checkpoint and Gtid events when
  renaming it to a binlog file.

  Since binlog_cache_data's cache_log is directly accessed by binlog log,
  online alter and wsrep. It is not easy to update all the code. Thus
  binlog cache will not reserve space if it is not session binlog cache or
  wsrep session is enabled.

  - m_file_reserved_bytes
    Stores the bytes reserved at the begin of the cache file.
    It is initialized in write_prepare() and cleared by reset().

    The reserved file header is hide to callers. Thus there is no
    change for callers. E.g.
    - get_byte_position() still get the length of binlog data
      written to the cache, but not the file length.
    - truncate(0) will truncate the file to m_file_reserved_bytes but not 0.

  - write_prepare()
    write_prepare() is called everytime when anything is being written
    into the cache. It will call init_file_reserved_bytes() to  create
    the cache file (if it doesn't exist) and reserve suitable space if
    the data written exceeds buffer's size.

* Binlog_commit_by_rotate
  It is used to encapsulate the code for remaing a binlog cache
  tempoary file to binlog file.
  - should_commit_by_rotate()
    it is called by write_transaction_to_binlog_events() to check if
    a binlog cache should be rename to a binlog file.
  - commit()
    That is the entry to rename a binlog cache and commit the
    transaction. Both rename and commit are protected by LOCK_log,
    Thus not other transactions can write anything into the renamed
    binlog before it.

    Rename happens in a rotation. After the new binlog file is generated,
    replace_binlog_file() is called to:
    - copy data from the new binlog file to its binlog cache file.
    - write gtid event.
    - rename the binlog cache file to binlog file.

    After that the rotation will continue to succeed. Then the transaction
    is committed in a seperated group itself. Its cache file will be
    detached and cache log will be reset before calling
    trx_group_commit_with_engines(). Thus only Xid event be written.

## Release Notes

## How can this PR be tested?

- [x] *This is a new feature or a refactoring, and the PR is based against the latest MariaDB development branch.*

## PR quality check

### DIFF
--- a/mysql-test/main/mysqld--help.result
+++ b/mysql-test/main/mysqld--help.result
@@ -109,6 +109,16 @@ The following specify which files/extra groups are read (specified before remain
  --binlog-ignore-db=name 
  Tells the master that updates to the given database
  should not be logged to the binary log
+ --binlog-large-commit-threshold=# 
+ Increases transaction concurrency for large transactions
+ (i.e. those with sizes larger than this value) by using
+ the large transaction's cache file as a new binary log,
+ and rotating the active binary log to the large
+ transaction's cache file at commit time. This avoids the
+ default commit logic that copies the transaction cache
+ data to the end of the active binary log file while
+ holding a lock that prevents other transactions from
+ binlogging
  --binlog-legacy-event-pos 
  Fill in the end_log_pos field of _all_ events in the
  binlog, even when doing so costs performance. Can be used
@@ -614,7 +624,9 @@ The following specify which files/extra groups are read (specified before remain
  --max-binlog-cache-size=# 
  Sets the total size of the transactional cache
  --max-binlog-size=# Binary log will be rotated automatically when the size
- exceeds this value
+ exceeds this value, unless
+ `binlog_large_commit_threshold` causes rotation
+ prematurely
  --max-binlog-stmt-cache-size=# 
  Sets the total size of the statement cache
  --max-binlog-total-size=# 
@@ -1569,9 +1581,12 @@ The following specify which files/extra groups are read (specified before remain
  temporary table exceeds this size, MariaDB will
  automatically convert it to an on-disk MyISAM or Aria
  table
- -t, --tmpdir=name   Path for temporary files. Several paths may be specified,
- separated by a colon (:), in this case they are used in a
- round-robin fashion
+ -t, --tmpdir=name   Path for temporary files. Files that are created in
+ background for binlogging by user threads are placed in a
+ separate location (see `binlog_large_commit_threshold`
+ option). Several paths may be specified, separated by a
+ colon (:), in this case they are used in a round-robin
+ fashion
  --transaction-alloc-block-size=# 
  Allocation block size for transactions to be stored in
  binary log
@@ -1630,6 +1645,7 @@ binlog-format MIXED
 binlog-gtid-index TRUE
 binlog-gtid-index-page-size 4096
 binlog-gtid-index-span-min 65536
+binlog-large-commit-threshold 134217728
 binlog-legacy-event-pos FALSE
 binlog-optimize-thread-scheduling TRUE
 binlog-row-event-max-size 8192
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -163,6 +163,128 @@ static SHOW_VAR binlog_status_vars_detail[]=
   {NullS, NullS, SHOW_LONG}
 };
 
+/**
+  This class implementes the feature to rename a binlog cache temporary file to
+  a binlog file. It is used to avoid holding LOCK_log long time when writting a
+  huge binlog cache to binlog file.
+
+  With this feature, temporary files of binlog caches will be created in
+  BINLOG_CACHE_DIR which is created in the same directory to binlog files
+  at server startup.
+*/
+class Binlog_commit_by_rotate
+{
+public:
+  Binlog_commit_by_rotate() {}
+
+  /**
+    Check whether rename to binlog should be executed on the cache_data.
+
+    @param group_commit_entry object of current transaction
+
+    @retval true    it should do rename
+    @retval false   it should do normal commit.
+  */
+  bool should_commit_by_rotate(
+      const MYSQL_BIN_LOG::group_commit_entry *entry) const;
+
+  /**
+    This function is the entry function to rename a binlog cache to a binlog
+    file. It first, rotate the binlog, then rename the temporary file of the
+    binlog cache to new binlog file, after that it commits the transaction.
+
+    @param entry, group_commit_entry object of current transaction.
+
+    @retval true    Succeeds to rename binlog cache and commit the transaction
+    @retval false   Fails if error happens or the cache cannot be renamed
+  */
+  bool commit(MYSQL_BIN_LOG::group_commit_entry *entry);
+
+  /**
+    During binlog rotate, after creating the new binlog file and writing the
+    events that describe its state (e.g. Format description event)}, copy
+    them into the the binlog cache, delete the binlog file and then rename
+    the binlog cache to the new binlog file.
+
+    @retval true    Succeeds to replace the binlog file.
+    @retval false   Failed to replace the binlog file. It only return
+                    true if some error happened after the new binlog file
+                    is deleted. In this situation rotate process will fail.
+  */
+  bool replace_binlog_file();
+
+  /**
+    The space left is more than a gtid event required, thus the extra
+    space is padded into the gtid event as 0. This function is used
+    to calculate the real gtid size with pad.
+  */
+  size_t get_gtid_event_pad_data_size();
+
+  /**
+    The space required for session binlog caches to reserve. It is calculated
+    from the length of current binlog file when it is generated and aligned
+    to IO_SIZE;
+
+    @param header_len  header length of current binlog file.
+  */
+  void set_reserved_bytes(uint32 header_len)
+  {
+    // Add reserved space for gtid event
+    header_len+= LOG_EVENT_HEADER_LEN + Gtid_log_event::max_data_length +
+                 BINLOG_CHECKSUM_LEN;
+
+    // reserved size is aligned to IO_SIZE.
+    header_len= (header_len + (IO_SIZE - 1)) & ~(IO_SIZE - 1);
+    if (header_len != m_reserved_bytes)
+      m_reserved_bytes= header_len;
+  }
+
+  /**
+    Return reserved space required for binlog cache. It is NOT defined as
+    an atomic variable, while it is get and set in parallel. Synchronizing
+    between set and get is not really necessary, m_reserved_bytes doesnot
+    be updated often. It may read an old value, but it just effects
+    current transaction. Next transaction will get the fresh value.
+    And reserving space is a transaction level action, so there alway
+    are some transactions reserving space with the old value.
+  */
+  uint32 get_reserved_size()
+  {
+    return m_reserved_bytes;
+  }
+private:
+  /* Singleton object, disable the constructors to prevent copying by mistake */
+  Binlog_commit_by_rotate &operator=(const Binlog_commit_by_rotate &);
+  Binlog_commit_by_rotate(const Binlog_commit_by_rotate &);
+
+  /**
+    The commit entry of current transaction which is committed by renaming
+    it binlog cache to binlog file.
+  */
+  MYSQL_BIN_LOG::group_commit_entry *m_entry{nullptr};
+
+  /**
+    The cache_data which will be renamed to binlog, it is used with
+    LOCK_log acquired.
+  */
+  binlog_cache_data *m_cache_data{nullptr};
+
+  /**
+    It will be set to true if rename operation succeeds,
+    it is used with LOCK_log acquired.
+  */
+  bool m_replaced{false};
+
+  uint32 m_reserved_bytes {IO_SIZE};
+};
+static Binlog_commit_by_rotate binlog_commit_by_rotate;
+ulonglong opt_binlog_commit_by_rotate_threshold= 128 * 1024 * 1024;
+
+uint32 binlog_cache_reserved_size()
+{
+  return binlog_commit_by_rotate.get_reserved_size();
+}
+
 /*
   Variables for the binlog background thread.
   Protected by the MYSQL_BIN_LOG::LOCK_binlog_background_thread mutex.
@@ -3761,7 +3883,8 @@ bool MYSQL_BIN_LOG::open(const char *log_name,
                          enum cache_type io_cache_type_arg,
                          ulong max_size_arg,
                          bool null_created_arg,
-                         bool need_mutex)
+                         bool need_mutex,
+                         bool commit_by_rotate)
 {
   xid_count_per_binlog *new_xid_list_entry= NULL, *b;
   DBUG_ENTER("MYSQL_BIN_LOG::open");
@@ -4027,14 +4150,23 @@ bool MYSQL_BIN_LOG::open(const char *log_name,
         goto err;
       bytes_written+= description_event_for_queue->data_written;
     }
+
+    /*
+      Offset must be saved before replace_binlog_file(), it will update the
+      file position
+    */
+    my_off_t offset= my_b_tell(&log_file);
+
+    if (commit_by_rotate && binlog_commit_by_rotate.replace_binlog_file())
+      goto err;
+
     if (flush_io_cache(&log_file) ||
         mysql_file_sync(log_file.file, MYF(MY_WME)))
       goto err;
 
-    my_off_t offset= my_b_tell(&log_file);
-
     if (!is_relay_log)
     {
+      binlog_commit_by_rotate.set_reserved_bytes((uint32)offset);
       /* update binlog_end_pos so that it can be read by after sync hook */
       reset_binlog_end_pos(log_file_name, offset);
 
@@ -4126,8 +4258,7 @@ bool MYSQL_BIN_LOG::open(const char *log_name,
   /* Notify the io thread that binlog is rotated to a new file */
   if (is_relay_log)
     signal_relay_log_update();
-  else
-    update_binlog_end_pos();
+
   DBUG_RETURN(0);
 
 err:
@@ -5708,7 +5839,7 @@ int MYSQL_BIN_LOG::new_file()
 {
   int res;
   mysql_mutex_lock(&LOCK_log);
-  res= new_file_impl();
+  res= new_file_impl(false);
   mysql_mutex_unlock(&LOCK_log);
   return res;
 }
@@ -5717,9 +5848,9 @@ int MYSQL_BIN_LOG::new_file()
   @retval
     nonzero - error
  */
-int MYSQL_BIN_LOG::new_file_without_locking()
+int MYSQL_BIN_LOG::new_file_without_locking(bool commit_by_rotate)
 {
-  return new_file_impl();
+  return new_file_impl(commit_by_rotate);
 }
 
 
@@ -5734,7 +5865,7 @@ int MYSQL_BIN_LOG::new_file_without_locking()
     binlog_space_total will be updated if binlog_space_limit is set
 */
 
-int MYSQL_BIN_LOG::new_file_impl()
+int MYSQL_BIN_LOG::new_file_impl(bool commit_by_rotate)
 {
   int error= 0, close_on_error= FALSE;
   char new_name[FN_REFLEN], *new_name_ptr, *old_name, *file_to_open;
@@ -5856,7 +5987,8 @@ int MYSQL_BIN_LOG::new_file_impl()
   {
     /* reopen the binary log file. */
     file_to_open= new_name_ptr;
-    error= open(old_name, new_name_ptr, 0, io_cache_type, max_size, 1, FALSE);
+    error= open(old_name, new_name_ptr, 0, io_cache_type, max_size, 1, FALSE,
+                commit_by_rotate);
   }
 
   /* handle reopening errors */
@@ -5959,7 +6091,7 @@ bool MYSQL_BIN_LOG::append_no_lock(Log_event* ev,
   if (flush_and_sync(0))
     goto err;
   if (my_b_append_tell(&log_file) > max_size)
-    error= new_file_without_locking();
+    error= new_file_without_locking(false);
 err:
   update_binlog_end_pos();
   DBUG_RETURN(error);
@@ -6018,7 +6150,7 @@ bool MYSQL_BIN_LOG::write_event_buffer(uchar* buf, uint len)
   if (flush_and_sync(0))
     goto err;
   if (my_b_append_tell(&log_file) > max_size)
-    error= new_file_without_locking();
+    error= new_file_without_locking(false);
 err:
   my_safe_afree(ebuf, len);
   if (likely(!error))
@@ -6207,11 +6339,11 @@ static binlog_cache_mngr *binlog_setup_cache_mngr(THD *thd)
                                   sizeof(binlog_cache_mngr),
                                                    MYF(MY_ZEROFILL));
   if (!cache_mngr ||
-      open_cached_file(&cache_mngr->stmt_cache.cache_log, mysql_tmpdir,
-                       LOG_PREFIX, (size_t)binlog_stmt_cache_size,
+      open_cached_file(&cache_mngr->stmt_cache.cache_log, binlog_cache_dir,
+                       LOG_PREFIX, (size_t) binlog_stmt_cache_size,
                        MYF(MY_WME | MY_TRACK_WITH_LIMIT)) ||
-      open_cached_file(&cache_mngr->trx_cache.cache_log, mysql_tmpdir,
-                       LOG_PREFIX, (size_t)binlog_cache_size,
+      open_cached_file(&cache_mngr->trx_cache.cache_log, binlog_cache_dir,
+                       LOG_PREFIX, (size_t) binlog_cache_size,
                        MYF(MY_WME | MY_TRACK_WITH_LIMIT)))
   {
     my_free(cache_mngr);
@@ -6866,6 +6998,7 @@ Event_log::prepare_pending_rows_event(THD *thd, TABLE* table,
 bool
 MYSQL_BIN_LOG::write_gtid_event(THD *thd, bool standalone,
                                 bool is_transactional, uint64 commit_id,
+                                bool commit_by_rotate,
                                 bool has_xid, bool is_ro_1pc)
 {
   rpl_gtid gtid;
@@ -6933,6 +7066,9 @@ MYSQL_BIN_LOG::write_gtid_event(THD *thd, bool standalone,
     thd->variables.server_id= global_system_variables.server_id;
   }
 #endif
+
+  if (unlikely(commit_by_rotate))
+    gtid_event.pad_to_size= binlog_commit_by_rotate.get_gtid_event_pad_data_size();
 
   if (write_event(&gtid_event))
     DBUG_RETURN(true);
@@ -7262,7 +7398,7 @@ bool MYSQL_BIN_LOG::write(Log_event *event_info, my_bool *with_annotate)
                                              commit_name.length);
           commit_id= entry->val_int(&null_value);
         });
-      res= write_gtid_event(thd, true, using_trans, commit_id);
+      res= write_gtid_event(thd, true, using_trans, commit_id, false);
       if (mdl_request.ticket)
         thd->mdl_context.release_lock(mdl_request.ticket);
       thd->backup_commit_lock= 0;
@@ -7623,7 +7759,8 @@ MYSQL_BIN_LOG::do_checkpoint_request(ulong binlog_id)
   @retval
     nonzero - error in rotating routine.
 */
-int MYSQL_BIN_LOG::rotate(bool force_rotate, bool* check_purge)
+int MYSQL_BIN_LOG::rotate(bool force_rotate, bool *check_purge,
+                          bool commit_by_rotate)
 {
   int error= 0;
   ulonglong binlog_pos;
@@ -7664,7 +7801,7 @@ int MYSQL_BIN_LOG::rotate(bool force_rotate, bool* check_purge)
     */
     mark_xids_active(binlog_id, 1);
 
-    if (unlikely((error= new_file_without_locking())))
+    if (unlikely((error= new_file_without_locking(commit_by_rotate))))
     {
       /** 
          Be conservative... There are possible lost events (eg, 
@@ -7965,11 +8102,13 @@ int Event_log::write_cache_raw(THD *thd, IO_CACHE *cache)
 
 int Event_log::write_cache(THD *thd, binlog_cache_data *cache_data)
 {
-  int res;
   IO_CACHE *cache= &cache_data->cache_log;
   DBUG_ENTER("Event_log::write_cache");
 
   mysql_mutex_assert_owner(&LOCK_log);
+
+  if (cache_data->init_for_read())
+    DBUG_RETURN(ER_ERROR_ON_WRITE);
 
   /*
     If possible, just copy the cache over byte-by-byte with pre-computed
@@ -7979,13 +8118,12 @@ int Event_log::write_cache(THD *thd, binlog_cache_data *cache_data)
       likely(!crypto.scheme) &&
       likely(!opt_binlog_legacy_event_pos))
   {
-    int res= my_b_copy_all_to_cache(cache, &log_file);
-    status_var_add(thd->status_var.binlog_bytes_written, my_b_tell(cache));
+    int res=
+        my_b_copy_to_cache(cache, &log_file, cache_data->length_for_read());
+    status_var_add(thd->status_var.binlog_bytes_written,
+                   cache_data->length_for_read());
     DBUG_RETURN(res ? ER_ERROR_ON_WRITE : 0);
   }
-
-  if ((res= reinit_io_cache(cache, READ_CACHE, 0, 0, 0)))
-    DBUG_RETURN(ER_ERROR_ON_WRITE);
 
   /* Amount of remaining bytes in the IO_CACHE read buffer. */
   size_t log_file_pos;
@@ -8700,10 +8838,29 @@ end:
   DBUG_RETURN(result);
 }
 
-bool
+inline bool
 MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
 {
+  if (unlikely(binlog_commit_by_rotate.should_commit_by_rotate(entry)))
+  {
+    if (binlog_commit_by_rotate.commit(entry))
+      return true;
+  }
+  else if (write_transaction_with_group_commit(entry))
+    return true;
+
+  if (likely(!entry->error))
+    return false;
+
+  write_transaction_handle_error(entry);
+  return true;
+}
+
+bool
+MYSQL_BIN_LOG::write_transaction_with_group_commit(group_commit_entry *entry)
+{
   int is_leader= queue_for_group_commit(entry);
+
 #ifdef WITH_WSREP
   /* commit order was released in queue_for_group_commit() call,
      here we check if wsrep_commit_ordered() failed or if we are leader */
@@ -8808,7 +8965,13 @@ MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
 
   if (likely(!entry->error))
     return entry->thd->wait_for_prior_commit();
+  else
+    write_transaction_handle_error(entry);
+  return true;
+}
 
+void MYSQL_BIN_LOG::write_transaction_handle_error(group_commit_entry *entry)
+{
   switch (entry->error)
   {
   case ER_ERROR_ON_WRITE:
@@ -8837,8 +9000,6 @@ MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
   if (entry->cache_mngr->using_xa && entry->cache_mngr->xa_xid &&
       entry->cache_mngr->need_unlog)
     mark_xid_done(entry->cache_mngr->binlog_id, true);
-
-  return 1;
 }
 
 /*
@@ -8854,65 +9015,74 @@ MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
 void
 MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
 {
-  uint xid_count= 0;
-  my_off_t UNINIT_VAR(commit_offset);
   group_commit_entry *current, *last_in_queue;
   group_commit_entry *queue= NULL;
-  bool check_purge= false;
-  ulong UNINIT_VAR(binlog_id);
-  uint64 commit_id;
   DBUG_ENTER("MYSQL_BIN_LOG::trx_group_commit_leader");
 
-  {
 #ifdef ENABLED_DEBUG_SYNC
-    DBUG_EXECUTE_IF("inject_binlog_commit_before_get_LOCK_log",
-      DBUG_ASSERT(!debug_sync_set_action(leader->thd, STRING_WITH_LEN
-        ("commit_before_get_LOCK_log SIGNAL waiting WAIT_FOR cont TIMEOUT 1")));
-    );
+  DBUG_EXECUTE_IF("inject_binlog_commit_before_get_LOCK_log",
+    DBUG_ASSERT(!debug_sync_set_action(leader->thd, STRING_WITH_LEN
+      ("commit_before_get_LOCK_log SIGNAL waiting WAIT_FOR cont TIMEOUT 1")));
+  );
 #endif
-    /*
-      Lock the LOCK_log(), and once we get it, collect any additional writes
-      that queued up while we were waiting.
-    */
-    DEBUG_SYNC(leader->thd, "commit_before_get_LOCK_log");
-    mysql_mutex_lock(&LOCK_log);
-    DEBUG_SYNC(leader->thd, "commit_after_get_LOCK_log");
+  /*
+    Lock the LOCK_log(), and once we get it, collect any additional writes
+    that queued up while we were waiting.
+  */
+  DEBUG_SYNC(leader->thd, "commit_before_get_LOCK_log");
+  mysql_mutex_lock(&LOCK_log);
+  DEBUG_SYNC(leader->thd, "commit_after_get_LOCK_log");
 
-    mysql_mutex_lock(&LOCK_prepare_ordered);
-    if (opt_binlog_commit_wait_count)
-      wait_for_sufficient_commits();
-    /*
-      Note that wait_for_sufficient_commits() may have released and
-      re-acquired the LOCK_log and LOCK_prepare_ordered if it needed to wait.
-    */
-    current= group_commit_queue;
-    group_commit_queue= NULL;
-    mysql_mutex_unlock(&LOCK_prepare_ordered);
-    binlog_id= current_binlog_id;
+  mysql_mutex_lock(&LOCK_prepare_ordered);
+  if (opt_binlog_commit_wait_count)
+    wait_for_sufficient_commits();
+  /*
+    Note that wait_for_sufficient_commits() may have released and
+    re-acquired the LOCK_log and LOCK_prepare_ordered if it needed to wait.
+  */
+  current= group_commit_queue;
+  group_commit_queue= NULL;
+  mysql_mutex_unlock(&LOCK_prepare_ordered);
 
-    /* As the queue is in reverse order of entering, reverse it. */
-    last_in_queue= current;
-    while (current)
-    {
-      group_commit_entry *next= current->next;
-      /*
-        Now that group commit is started, we can clear the flag; there is no
-        longer any use in waiters on this commit trying to trigger it early.
-      */
-      current->thd->waiting_on_group_commit= false;
-      current->next= queue;
-      queue= current;
-      current= next;
-    }
-    DBUG_ASSERT(leader == queue /* the leader should be first in queue */);
-
-    /* Now we have in queue the list of transactions to be committed in order. */
-  }
-    
-  DBUG_ASSERT(is_open());
-  if (likely(is_open()))                       // Should always be true
+  /* As the queue is in reverse order of entering, reverse it. */
+  last_in_queue= current;
+  while (current)
   {
-    commit_id= (last_in_queue == leader ? 0 : (uint64)leader->thd->query_id);
+    group_commit_entry *next= current->next;
+    /*
+      Now that group commit is started, we can clear the flag; there is no
+      longer any use in waiters on this commit trying to trigger it early.
+    */
+    current->thd->waiting_on_group_commit= false;
+    current->next= queue;
+    queue= current;
+    current= next;
+  }
+  DBUG_ASSERT(leader == queue /* the leader should be first in queue */);
+  /* Now we have in queue the list of transactions to be committed in order. */
+
+  trx_group_commit_with_engines(leader, last_in_queue, false);
+  DBUG_VOID_RETURN;
+}
+
+void MYSQL_BIN_LOG::trx_group_commit_with_engines(group_commit_entry *leader,
+                                                  group_commit_entry *tail,
+                                                  bool commit_by_rotate)
+{
+  uint xid_count= 0;
+  bool check_purge= false;
+  ulong UNINIT_VAR(binlog_id);
+  my_off_t UNINIT_VAR(commit_offset);
+  group_commit_entry *current;
+
+  DBUG_ENTER("MYSQL_BIN_LOG::trx_group_commit_with_engines");
+  mysql_mutex_assert_owner(&LOCK_log);
+  binlog_id= current_binlog_id;
+
+  if (likely(is_open())) // Binlog could be closed if rotation fails
+  {
+    uint64 commit_id= (leader == tail ? 0 : (uint64) leader->thd->query_id);
+
     DBUG_EXECUTE_IF("binlog_force_commit_id",
       {
         const LEX_CSTRING commit_name= { STRING_WITH_LEN("commit_id") };
@@ -8933,7 +9103,7 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
       current->error and let the thread do the error reporting itself once
       we wake it up.
     */
-    for (current= queue; current != NULL; current= current->next)
+    for (current= leader; current != NULL; current= current->next)
     {
       set_current_thd(current->thd);
       binlog_cache_mngr *cache_mngr= current->cache_mngr;
@@ -8944,10 +9114,11 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
       */
       DBUG_ASSERT(!cache_mngr->stmt_cache.empty() ||
                   !cache_mngr->trx_cache.empty()  ||
-                  current->thd->transaction->xid_state.is_explicit_XA());
+                  current->thd->transaction->xid_state.is_explicit_XA() ||
+                  commit_by_rotate);
 
-      if (unlikely((current->error= write_transaction_or_stmt(current,
-                                                              commit_id))))
+      if (unlikely((current->error= write_transaction_or_stmt(
+                        current, commit_id, commit_by_rotate))))
         current->commit_errno= errno;
 
       strmake_buf(cache_mngr->last_commit_pos_file, log_file_name);
@@ -8980,7 +9151,7 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
     bool synced= 0;
     if (unlikely(flush_and_sync(&synced)))
     {
-      for (current= queue; current != NULL; current= current->next)
+      for (current= leader; current != NULL; current= current->next)
       {
         if (!current->error)
         {
@@ -9000,7 +9171,7 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
       mysql_mutex_assert_not_owner(&LOCK_after_binlog_sync);
       mysql_mutex_assert_not_owner(&LOCK_commit_ordered);
 
-      for (current= queue; current != NULL; current= current->next)
+      for (current= leader; current != NULL; current= current->next)
       {
 #ifdef HAVE_REPLICATION
         /*
@@ -9098,7 +9269,7 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
 
     bool first __attribute__((unused))= true;
     bool last __attribute__((unused));
-    for (current= queue; current != NULL; current= current->next)
+    for (current= leader; current != NULL; current= current->next)
     {
       last= current->next == NULL;
 #ifdef HAVE_REPLICATION
@@ -9151,8 +9322,8 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
       in this function, so parent does not need to and we need not set these
       values).
     */
-    last_in_queue->check_purge= check_purge;
-    last_in_queue->binlog_id= binlog_id;
+    tail->check_purge= check_purge;
+    tail->binlog_id= binlog_id;
 
     /* Note that we return with LOCK_commit_ordered locked! */
     DBUG_VOID_RETURN;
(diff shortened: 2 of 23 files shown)


bnestere (changes requested):
Hi @SongLibing !

Thanks for the submission! I was hoping I could do some profiling to compare this PR against the other open PR for non-blocking binlog (MDEV-33487/PR-3087), but unfortunately I haven't been able to get your PR to complete a benchmark. I only gave a high-level review for now (mostly to address the found problems), and once the patch is working consistently, I'll do another review with more in-depth remarks. Note that Kristian Nielsen reviewed PR-3087, and some of his comments on that PR are relevant to your patch as well. I'll add some of the more relevant points as comments here too, where appropriate (though it'd be good to read through his review too, as his general remarks are insightful).

One concrete problem I came across in trying to benchmark is that I saw binary logs could be completely empty (0 bytes), while there were later full binlogs with Gtid_list_log_events that had skipped events from the empty binlog.

As well as a couple assertions, which were pretty frequent:
```
mariadbd: /home/brandon/workspace/server/sql/log.cc:8986: void MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *, bool): Assertion `!cache_mngr->stmt_cache.empty() || !cache_mngr->trx_cache.empty() || current->thd->transaction->xid_state.is_explicit_XA()' failed.
```

```
mariadbd: /home/brandon/workspace/server/sql/sql_error.cc:335: void Diagnostics_area::set_ok_status(ulonglong, ulonglong, const char *): Assertion `!is_set() || (m_status == DA_OK_BULK && is_bulk_op())' failed.
```

For my benchmark, I was just using sysbench to update an initial set of rows in tables organized as `create table t<thread_number> (a int primary key, b longtext) engine=innodb` (with large values for `b`). My env is Ubuntu 24.04, with a DEBUG build compiled using clang-18.

sql/log.cc
```diff
@@ -12863,3 +12936,215 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+{
+  size_t begin_pos = my_b_tell(&mysql_bin_log.log_file);
+  size_t empty_size= m_cache_data->file_reserved_bytes() - begin_pos
+                     - LOG_EVENT_HEADER_LEN;
+
+  if (binlog_checksum_options != BINLOG_CHECKSUM_ALG_OFF)
+    empty_size-= BINLOG_CHECKSUM_LEN;
+
+  Empty_log_event empty_ev(m_entry->thd, (int) empty_size);
+  if (mysql_bin_log.write_event(&empty_ev) ||
```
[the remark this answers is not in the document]
SongLibing: Look the code in Gtid_log_event::write(), gtid log event also has variable length. with/o XID, single/multiple engine, it has different length.  I also thought about using a Query_log_event instead of a empty event, but that is also hack and a little bit complex than an empty event.

sql/log.cc
```diff
@@ -8858,7 +8932,6 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
     /* Now we have in queue the list of transactions to be committed in order. */
   }
     
-  DBUG_ASSERT(is_open());
```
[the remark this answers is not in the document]
SongLibing: If error happens in normal rotate, binlog will be closed, the following transactions will go commit with binlog closed. so I think the assertion is not correct and should be removed. I just added a test to simulate the situation when doing free flush.

sql/log.cc
```diff
@@ -8652,7 +8706,16 @@ MYSQL_BIN_LOG::queue_for_group_commit(group_commit_entry *orig_entry)
 bool
 MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
 {
-  int is_leader= queue_for_group_commit(entry);
+  int is_leader;
+
+  if (binlog_cache_free_flush.check(entry) &&
```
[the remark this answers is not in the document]
SongLibing: done.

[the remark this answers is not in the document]
SongLibing: I didn't finalize the patch yet. So there may be more name, comment, coding style related problems.

sql/log.cc
```diff
@@ -12863,3 +12936,215 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+    not supported.
+  */
+  if (!mysql_bin_log.is_open() || opt_binlog_legacy_event_pos ||
+      (expected_alg != cache_data->checksum_opt))
+  {
+    mysql_mutex_unlock(&mysql_bin_log.LOCK_log);
+    return false;
+  }
+
+  m_is_free_flushing= true;
```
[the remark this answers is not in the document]
SongLibing: I originally wanted to avoid to add an argument to other functions. But you are right, LOCK_log is freed earlier that make the logic a little complex. So I followed your suggestion to add a argument to related functions.

SongLibing:
@bnestere Thanks for reviewing and testing it. I fixed the problem you mentioned and I ran the new patch with sysbench. It works well. You can go ahead now.  I will add more coverage test cases later.

sql/log.cc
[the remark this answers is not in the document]
bnestere: You should be able to reserve the maximum size of a GTID event (which you already pre-calculate into `Gtid_log_event::max_data_length`) and pad the unused data with `0`s. The padded `0`s should just be treated as reserve bytes (see the Gtid_log_event class comment for more details), provided the `event_length` field is consistent with the maximum size. 

And another thought comes to mind, I think the GTID event should be written before the log rotation/renaming happens. That way, the rotate event never points to a corrupt log.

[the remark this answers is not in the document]
SongLibing: You are right, Gtid_log_event has fixed length. thanks for the information. 
As you said Gtid_list_log_event's length may change, and 
checkpoint event's length may change as well in the situation that the suffix number is larger than 999999. 
So we cannot suppose it has fixed length most time. but that is not very often to change.  We probable can detect the change when rotating binlog and updated the reserved length for binlog caches.

In that way, we can remove the empty event. But there is still some extra work to do. Binlog cache is aligned with IO_SIZE(4K), thus it will waste the first reserved size bytes of memory, or we need to add our own buffer instead of IO_CACHE's buffer.  I will try to remove the empty event.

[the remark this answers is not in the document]
SongLibing: Hi @bnestere, I removed the empty event, in the latest patch. The the space left is padded into Gtid event.
It still reserved 4K(IO_SIZE) space for easily handle the alignment. 

I tried to write gtid event earlier but that is a little complex, because the gtid needs to be written into the temp file.
so I think it is better not to spread free_flush related context to more places and  keep the original binlog_cache flush process. However I have no strong opinion on it.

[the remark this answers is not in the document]
bnestere: > It still reserved 4K(IO_SIZE) space for easily handle the alignment.

Hmm, sorry, why does aligning the cache data to the start of a page help in this case? I don't see a performance benefit (though that doesn't mean there isn't one :innocent:), and it adds complexity in the code, and awkwardness into the event.

> I tried to write gtid event earlier but that is a little complex, because the gtid needs to be written into the temp file. so I think it is better not to spread free_flush related context to more places and keep the original binlog_cache flush process. However I have no strong opinion on it.

Can you write it in `replace_binlog_file()` before doing the actual file replacement? That'd keep the spread minimal while also keeping the state consistent.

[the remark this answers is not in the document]
SongLibing: > Hmm, sorry, why does aligning the cache data to the start of a page help in this case? I don't see a performance benefit (though that doesn't mean there isn't one 😇), and it adds complexity in the code, and awkwardness into the event.
look the code below
```
  if (type == WRITE_CACHE)
    info->write_end=
      info->buffer+info->buffer_length- (seek_offset & (IO_SIZE-1));
```
the buffer of IO_CACHE is 4K aligned. Since we reserved e.g. 400 bytes, the seek_offset will be 400, thus only 4k-400B can be used, though 4K is allocated. It has no performance improvement, it just eliminates memory waste.

sql/log.cc
```diff
@@ -12914,3 +13024,221 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+  }
+
+  sql_print_information("Renamed binlog cache to binlog %s",
+                        mysql_bin_log.get_log_fname());
+
+  mysql_file_close(mysql_bin_log.log_file.file, MYF(MY_WME));
+  mysql_bin_log.log_file.file= new_log_fd;
+  new_log_fd= -1;
+
+  if (mysql_bin_log.write_gtid_event(
```
bnestere, line 13207: I still think this should happen before the actual replacement. I imagine you can create an `IO_CACHE` of the temp file, and pass it in as a new optional parameter to `write_gtid_event`, which in-turn would use 
```
bool Event_log::write_event(Log_event *ev, binlog_cache_data *data,
                            IO_CACHE *file)
```

if the parameter is not null. Then I imagine the copying of the header events could use the IO cache too, and then everything would be flushed together.

sql/log.cc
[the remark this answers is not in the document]
bnestere: Ok, I suppose it would be a trade-off either way. Though keeping with the aligned approach, I'd suggest adding some additional logic when setting `binlog_cache_data::m_file_reserved_bytes` to calculate the size of the header bytes to see if it exceeds `IO_SIZE` (if the `Gtid_list_log_event` is very big) and set it to the next multiple of `IO_SIZE` if so.

sql/log.cc
```diff
@@ -12914,3 +13024,221 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+  size_t binlog_size= my_b_tell(&mysql_bin_log.log_file);
+  size_t required_size= binlog_size;
+  // space for Gtid_log_event
+  required_size+= LOG_EVENT_HEADER_LEN + Gtid_log_event::max_data_length +
+                  BINLOG_CHECKSUM_LEN;
+
+  if (DBUG_IF("simulate_required_size_too_big") ||
+      required_size > m_cache_data->file_reserved_bytes())
+  {
+    sql_print_information("Could not rename binlog cache to binlog, "
```
bnestere, line 13149: This case should be unlikely, so 1) I'd wrap the `if` condition check with `unlikely()` and 2) can we just revert to the old behavior in this case? We shouldn't have yet written the `rotate` event, so we should be in a consistent state.

SongLibing: If error happens here, binlog system will be closed. That is not what I expected. In fact, replace_binlog_file() is called after written rotate event. the old file has been closed and  the new file has been created. In the situation, it will go finish the rotate. But the transaction will go to the original path to copy its cache to the new binlog file. The only different is that there is an extra rotate.

sql/log.cc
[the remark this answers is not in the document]
SongLibing: ok, In the original patch, it records the header size of the new binlog file when rotating. and use it(aligned to IO_SIZE) as the reserved space. I will add it back. I  had  removed the logic just because I thought  mariadb's gtid_list_log_event would not be  big.

sql/log.cc
SongLibing: I didn't use unlikely in this function, that is because this function is called not often.

sql/log.cc
SongLibing: moved it before actual replacement.

mysql-test/main/mysqld--help.result
```diff
@@ -95,6 +95,10 @@ The following specify which files/extra groups are read (specified before remain
  statement-based binary logging (smaller binary logs),
  MIXED for statement-based binary logging when it's safe
  with fall back to row-based otherwise
+ --binlog-free-flush-threshold=#
```
bnestere, line 98: I think `binlog-free-flush-threshold` is a little hard to decipher off the bat. Perhaps if we leverage the existing binlog group commit convention, this can be `binlog_group_commit_threshold`, which defines the size for transactions that, when smaller are considered for group commit , and when larger, commit standalone using their own transaction cache as a new binary log file. Though before making any changes, @andrelkin and @knielsen should provide input too.

mysql-test/suite/binlog/t/binlog_free_flush_atomic.test
```diff
@@ -0,0 +1,110 @@
+################################################################################
+# MDEV-32014 Rename binlog cache to binlog file
+#
+# It verifies that the rename logic is handled correct if error happens.
+################################################################################
+--source include/have_binlog_format_row.inc
+--source include/have_innodb.inc
+--source include/have_debug.inc
```
bnestere, line 8: Should also have
```
--source include/have_debug_sync.inc
```

mysql-test/suite/rpl/t/rpl_binlog_free_flush.test
```diff
@@ -0,0 +1,217 @@
...
+# - an ignorable event is generated just after the Gtid_log_event of the
+#   transaction to take the reserved spaces which is unused.
+#
+# It also verifies that rename is not supported in below cases
+# though the cache is larger than the threshold
+# - both statement and transaction cache should be flushed.
+# - the cache's checksum option is not same to binlog_checksum
+# - binlog_legacy_event_pos is enabled.
+################################################################################
+--source include/master-slave.inc
```
bnestere, line 17:
```
--source include/master-slave.inc
```
^ This include should always be the last of the header `--source`'s

mysql-test/suite/rpl/t/rpl_binlog_free_flush.test
```diff
@@ -0,0 +1,217 @@
...
+--echo #
+--echo # Not rename to binlog file If binlog_legacy_event_pos is on
+--echo #
+SET GLOBAL binlog_legacy_event_pos = ON;
+UPDATE t1 SET c1 = repeat('9', 5242880);
+SET GLOBAL binlog_legacy_event_pos = OFF;
+--let $binlog_file= query_get_value(SHOW MASTER STATUS, File, 1)
+--let $assert_cond= "$binlog_file" = "master-bin.000004"
+--let $assert_text= Binlog is not rotated
+--source include/assert.inc
```
bnestere, line 209: I think this test should also validate
 1. The binlogging of multiple concurrent transactions of:
   1a. A smaller transaction which is able to write to binlog (show the purpose of the JIRA :) )
   1b. Multiple large transactions
 2. the slave's binlogging behavior for all test cases

sql/log.cc
```diff
@@ -163,6 +163,111 @@ static SHOW_VAR binlog_status_vars_detail[]=
...
+
+    @retval true    free flush succeeds.
+    @retval false   free flush fails, it should go to normal commit process.
+  */
+  bool commit(MYSQL_BIN_LOG::group_commit_entry *entry);
+
+  /**
+    After rotate has created the new binlog file, it copies the content
+    of the new binlog file to the binlog cache, delete the new binlog file
+    and then rename the binlog cache to the new binlog file.
```
bnestere, line 205: Perhaps here, we can be specific about "the content" as to the actual binlog events that are copied.

sql/log.cc
```diff
@@ -163,6 +163,111 @@ static SHOW_VAR binlog_status_vars_detail[]=
...
+
+    @retval true   Succeeds to replace the binlog file.
+    @retval false    Failed to replace the binlog file. It only return
+                    true if some error happened after the new binlog file
+                    is deleted. In this situation rotate process will fail.
+  */
+  bool replace_binlog_file();
+  /**
```
bnestere, line 213: Space between the last function dec and this comment

sql/log.cc
```diff
@@ -163,6 +163,111 @@ static SHOW_VAR binlog_status_vars_detail[]=
...
+                 BINLOG_CHECKSUM_LEN;
+    header_len= header_len - (header_len % IO_SIZE) + IO_SIZE;
+    if (header_len != m_reserved_bytes)
+      m_reserved_bytes= header_len;
+  }
+
+  /**
+    Return reserved space required for binlog cache. It is NOT defined as
+    an atomic variable, while it is get and set in parallel. Synchronizing
+    between set and get is not really necessary.
```
bnestere, line 240: Can you extend a bit on why synchronizing isn't necessary

sql/log.cc
```diff
@@ -163,6 +163,111 @@ static SHOW_VAR binlog_status_vars_detail[]=
...
+  /** The cache_data which will be renamed to binlog. */
+  binlog_cache_data *m_cache_data{nullptr};
+
+  /** It will be set to true if rename operation succeeds */
+  bool m_replaced{false};
+
+  uint32 m_reserved_bytes {IO_SIZE};
+};
+static Binlog_free_flush binlog_free_flush;
+ulonglong opt_binlog_free_flush_threshold= 10 * 1024 * 1024;
```
bnestere, line 264: I think this mismatches with the default `128 * 1024 * 1024` mentioned in `sys_vars.cc`

bnestere (changes requested):
Thanks for such quick updates @SongLibing! Here's another pass of your commit with minor suggestions.

sql/log_event.h
```diff
@@ -3404,6 +3412,11 @@ class Gtid_log_event: public Log_event
   static const uchar FL_EXTRA_THREAD_ID= 16; // thread_id like in BEGIN Query
 
 #ifdef MYSQL_SERVER
+  static const uint max_data_length= GTID_HEADER_LEN + 2 + sizeof(XID)
```
andrelkin, line 3415: This new member does not count the size of the user XA's xid which the event embeds with
`FL_*_XA` flags.

andrelkin (commented):
While I am collecting more, here comes the first short question.

sql/log.cc
SongLibing: I didn't write the specific events down, because the function doesn't care about the exact events. It just do bytes copy. Even new events added in future, this function doesn't need to be changed.

sql/log.cc
SongLibing: extend the comment like:
```
 /**
    Return reserved space required for binlog cache. It is NOT defined as
    an atomic variable, while it is get and set in parallel. Synchronizing
    between set and get is not really necessary, m_reserved_bytes doesnot
    be updated often. It may read an old value, but it just effects
    current transaction. Next transaction will get the fresh value.
    And reserving space is a transaction level action, so there alway
    are some transactions reserving space with the old value.
 */
```

sql/log.cc
```diff
@@ -8944,10 +9083,11 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
       */
       DBUG_ASSERT(!cache_mngr->stmt_cache.empty() ||
                   !cache_mngr->trx_cache.empty()  ||
-                  current->thd->transaction->xid_state.is_explicit_XA());
+                  current->thd->transaction->xid_state.is_explicit_XA() ||
+                  is_free_flush);
 
-      if (unlikely((current->error= write_transaction_or_stmt(current,
-                                                              commit_id))))
+      if (unlikely((current->error= write_transaction_or_stmt(
```
andrelkin, line 9089: ... inside `write_transaction_or_stmt`?

sql/log.cc
```diff
@@ -8703,7 +8821,16 @@ MYSQL_BIN_LOG::queue_for_group_commit(group_commit_entry *orig_entry)
 bool
 MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
 {
-  int is_leader= queue_for_group_commit(entry);
+  int is_leader;
+
+  if (binlog_free_flush.should_free_flush(entry) &&
```
andrelkin, line 8826: There's an invocation of  `trx_group_commit_leader()` inside which I fear not least 'cos
that new decision to compute the leader skips `queue_for_group_commit()` (and something else I would add later).
Would it be actually possible to execute all the renaming logics in the flush loop of `trx_group_commit_leader()` ? I [am pointing](https://github.com/MariaDB/server/pull/3402/commits/75e36f431bd08fe4e2b192b496456751466a018b#r1734958967) to the exact place ...

sql/log_event.h
SongLibing: sizeof(XID) is the one for external XID.

sql/log.cc
bnestere: Right.. Though even without being specific, I think the comment should provide an expectation for the content of the rotated binlog file (i.e. at what point in rotation). Perhaps something like
> During binlog rotate, after creating the new binlog file and writing the events that describe its state (e.g. Format description event)}, copy ...

knielsen (changes requested):
Full review sent to mailing list: https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/thread/FQ2QVONAGBHYBSX6EISZQNLICNF7OWRK/

Also pasted below:

```
Hi Libing Song,

Here is my full review of the patch for MDEV-32014 Rename binlog cache
temporary file to binlog file for large transaction.

First let me say that thanks, reviewing this patch was a pleasant surprise.
The binlog code that you are integrating this new feature in is complex to
work with, but overall I think you have managed to do it a pretty clean way.
That is great work, and I think a lot of effort must have gone into getting
it this way. I have a number of comments and requests for changes below, so
I wanted to give this overall good impression up-front.

Also thanks to Brandon for doing first review. My apologies if I am
repeating something already discussed, but Github pull request reviews of
non-trivial patches are almost impossible to read, with reviews being split
in fragments, patches disappearing after force-pushes, etc.

The root cause for the problem being addressed by the patch is a fundamental
limitation in the current binlog format. It requires all transactions to be
binlogged as a single event group, using consecutive bytes in a single
binlog file. Thus, while writing the large transaction to the binlog during
commit, all other transaction commits will have to wait before they can
commit.

Eventually, we should move towards an enhanced binlog format that doesn't
have this limitation, so that transactions can be partially binlogged during
execution, avoiding this bottlenect. There are a number of other problems
caused by this limitation that are not addressed by this patch. Changing the
binlog format is a complex task not easily done, so this patch should be
seen as a temporary solution to one specific bottleneck, until if/when this
underlying limitation can hopefully be removed.

First some overall comments:


As Brandon also remarked, the name "binlog_free_flush" does not describe the
feature very clearly. I would suggest to use the term "rotate" in the name,
as the concept of binlog rotation is well known in MariaDB replication, and
the patch effectively performs large commits by doing it as a binlog
rotation instead of a simple write to the existing binlog file.

So I suggest to name the option --binlog-commit-by-rotate-threshold. And I
suggest in the code (names and comments) to use the phrase
"rotate_by_commit" instead of "free_flush".


The patch enables the feature by default, with a threshold of 128MB.
We need to consider if that is appropriate, or if the feature should be
disabled by default.

Reasons for disabling by default:

 - The feature causes some binlog files to be cut short of the configured
   --max-binlog-size, which will be a surprise to users.

 - The feature adds an overhead of 4096 pre-allocated bytes at the start of
   every trx cache (though the overhead is minimal).

 - A default of 128MB is somewhat arbitrary.

 - Releasing this feature disabled by default is somewhat less risky wrt.
   regressions for users that upgrade.

Reasons to enable by default:

 -  Most users will otherwise not know the feature exists, and will thus not
    benefit from this work.

 - It will get much less testing "in the wild".

So the decision will be a compromise between these two. If I had to make the
choice, I would default to this feature being disabled. But I'm open to hear
opinions the other way.


The patch calls the trx_group_commit_leader() function. That is ackward, as
commit-by-rotate does not actually participate in group commit. But the
problem is with the existing function trx_group_commit_leader(), which does
too much in a single function.

So please do a small refactoring of trx_group_commit_leader() and keep only
the first part of that function that fetches the group commit queue. Move
the second part to a new function trx_group_commit_with_engines() that is
called by trx_group_commit_leader() passing the "queue" and "last_in_queue"
values. And assert in trx_group_commit_with_engines() that LOCK_log is held
on entry and comment that this function will release the lock.

Then you can from your Binlog_free_flush::commit() function call into
trx_group_commit_with_engines() directly, avoiding the need for adding a
true/false flag, and also avoiding that the trx_group_commit_leader()
function is sometimes called with LOCK_log held and sometimes not.


Same situation with MYSQL_BIN_LOG::write_transaction_to_binlog_events(),
this function is also doing too much and makes integrating your code
ackward.

Please refactor the MYSQL_BIN_LOG::write_transaction_to_binlog_events()
function something like this:

bool
MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
{
  if (binlog_commit_by_rotate.should_commit_by_rotate(entry) &&
      binlog_commit_by_rotate.commit(entry))
  {
    /* Commit done by renaming the trx/stmt cache. */
  }
  else
  {
    if (write_transaction_with_group_commit(entry))
      return true;  /* Error */
  }

  if (unlikely(entry->error))
  {
    write_transaction_handle_error(entry);
    return 1;
  }
  return 0;
}

Introducing two new functions:

  bool write_transaction_with_group_commit(group_commit_entry *entry)

This function contains the code from write_transaction_to_binlog_events() up
to and including this:

  if (likely(!entry->error))
    return entry->thd->wait_for_prior_commit();

You don't need the wait_for_prior_commit() in the commit-by-rotate case, as
you already did this in Binlog_free_flush::commit()).

You also don't need to handle the (!opt_optimize_thread_scheduling) case for
commit-by-rotate, this is old code that is being deprecated. Just add a
condition in should_free_flush() to skip if opt_optimize_thread_scheduling
is set.

Other new function:

  void write_transaction_handle_error(group_commit_entry *entry)

This function contains the remainder of the
write_transaction_to_binlog_events() function starting from
"switch (entry->error) ...".

This way, the commit-by-rotate case is cleanly separated from the normal
group commit case and the code of write_transaction_to_binlog_events()
becomes much cleaner.


Now follows more detailed comments on specific parts of the patch:



> diff --git a/sql/log.cc b/sql/log.cc
> index 34f9ad745fc..3dc57b21c05 100644
> --- a/sql/log.cc
> +++ b/sql/log.cc
> @@ -163,6 +163,111 @@ static SHOW_VAR binlog_status_vars_detail[]=


> +  void set_reserved_bytes(uint32 header_len)
> +  {
> +    // Gtid event length
> +    header_len+= LOG_EVENT_HEADER_LEN + Gtid_log_event::max_data_length +
> +                 BINLOG_CHECKSUM_LEN;
> +    header_len= header_len - (header_len % IO_SIZE) + IO_SIZE;

This has an off-by-one error, it will round up IO_SIZE to 2*IO_SIZE.
Use eg. header_len = (header_len + (IO_SIZE - 1)) & ~(IO_SIZE - 1)
You can use static_assert((IO_SIZE & (IO_SIZE - 1)) == 0, "power of two")
if you want to catch if anyone would ever change IO_SIZE to be not a power
of two.


+private:
+  Binlog_free_flush &operator=(const Binlog_free_flush &);
+  Binlog_free_flush(const Binlog_free_flush &);

Add before this a comment "singleton object, prevent copying by mistake".


> +  char m_cache_dir[FN_REFLEN];

This member appears to be unused?

> +  /** The cache_data which will be renamed to binlog. */
> +  binlog_cache_data *m_cache_data{nullptr};

Also appears unused.

Maybe add a comment that m_entry and m_replaced are protected by LOCK_log?


> +static Binlog_free_flush binlog_free_flush;
> +ulonglong opt_binlog_free_flush_threshold= 10 * 1024 * 1024;

This initializes the value to 10 MB, but the default value in sys_vars.cc is
128MB, so that seems inconsistent.


> @@ -4126,8 +4238,7 @@ bool MYSQL_BIN_LOG::open(const char *log_name,
>    /* Notify the io thread that binlog is rotated to a new file */
>    if (is_relay_log)
>      signal_relay_log_update();
> -  else
> -    update_binlog_end_pos();
> +

Why do you remove update_binlog_end_pos() here? Doesn't the binlog dump
threads need this update?


> @@ -8703,7 +8821,16 @@ MYSQL_BIN_LOG::queue_for_group_commit(group_commit_entry *orig_entry)
>  bool
>  MYSQL_BIN_LOG::write_transaction_to_binlog_events(group_commit_entry *entry)
>  {
> -  int is_leader= queue_for_group_commit(entry);
> +  int is_leader;
> +
> +  if (binlog_free_flush.should_free_flush(entry) &&
> +      binlog_free_flush.commit(entry))
> +  {
> +    is_leader= 1;
> +    goto commit;
> +  }
> +
> +  is_leader= queue_for_group_commit(entry);

As described above, re-factor the write_transaction_to_binlog_events()
function to avoid this assignment of is_leader and the goto.


> @@ -8863,6 +8992,16 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader)
>    uint64 commit_id;
>    DBUG_ENTER("MYSQL_BIN_LOG::trx_group_commit_leader");
>  
> +  /*
> +    When move a binlog cache to a binlog file, the transaction itself is
> +    a group.
> +  */
> +  if (unlikely(is_free_flush))
> +  {
> +    last_in_queue= leader;
> +    queue= leader;
> +  }
> +  else

As described above, re-factor trx_group_commit_leader() to avoid this check
in trx_group_commit_leader(), instead calling directly into the relevant
code now in trx_group_commit_with_engines().


> +inline bool Binlog_free_flush::should_free_flush(
> +    const MYSQL_BIN_LOG::group_commit_entry *entry) const
> +{

> +  /* Do not free flush if total_bytes smaller than limit size. */
> +  if (likely(cache_data->get_byte_position() <=
> +             opt_binlog_free_flush_threshold))
> +    return false;

Put this check as the first check in the function. So the common case of a
small transaction does not need to run any of the other checks.

And as suggested above, put a check:

  if (unlikely(!opt_optimize_thread_scheduling))
    return false;

(This is not critical, but just to make things a bit easier; there is no
reason for you/us to worry about making commit-by-rotate work with disabled
optimize_thread_scheduling, this option isn't used any more. Otherwise you'd
need to test that your patch works correctly with optimize_thread_scheduling
disabled, which would just be a waste of time.)


> +bool Binlog_free_flush::commit(MYSQL_BIN_LOG::group_commit_entry *entry)
> +{

> +  // It will be released by trx_group_commit_leader
> +  mysql_mutex_lock(&mysql_bin_log.LOCK_log);

After suggested changes, comment should read "It will be released by
trx_group_commit_with_engines()."


> +  mysql_bin_log.trx_group_commit_leader(entry, true /* is_free_flush */);

As discussed above, change this to call a new function
mysql_bin_log.trx_group_commit_with_engines(), which contains the second
part of trx_group_commit_leader() that you need to run. Then there is no
longer a need to pass a `true` flag here to skip the first part.

Also, before the call, this is the place to set entry->next to nullptr, to
convert the stand-alone entry into a (singleton) list (instead of
initializing it in MYSQL_BIN_LOG::write_transaction_to_binlog().)


> @@ -8301,6 +8418,7 @@ MYSQL_BIN_LOG::write_transaction_to_binlog(THD *thd,
>      DBUG_RETURN(0);
>    }
>  
> +  entry.next= nullptr;
>    entry.thd= thd;
>    entry.cache_mngr= cache_mngr;
>    entry.error= 0;

Do you need to initialize entry->next here, and if so why?
At this point, the entry is not a list, it is a stand-alone struct that is
not yet linked into any list, the next pointer should be set appropriately
when/if the struct is later put into a list.

More important than the minor cost of initializing, by redundantly
initializing here we are hiding bugs (eg. Valgrind/msan/asan) if some code
would incorrectly read the next pointer.



> +bool Binlog_free_flush::replace_binlog_file()
> +{
> +  size_t binlog_size= my_b_tell(&mysql_bin_log.log_file);
> +  size_t required_size= binlog_size;
> +  // space for Gtid_log_event
> +  required_size+= LOG_EVENT_HEADER_LEN + Gtid_log_event::max_data_length +
> +                  BINLOG_CHECKSUM_LEN;

Just a remark, here you actually know the real required size of the GTID
event, so you don't need to conservatively estimate
Gtid_log_event::max_data_length. But I think doing it this way is fine, it
is simpler and in most cases we are rounding up to IO_SIZE anyway, so I'm
fine with keeping it this way.


> +    sql_print_information("Could not rename binlog cache to binlog, "
> +                          "require %llu bytes but only %llu bytes reserved.",
> +                          required_size, m_cache_data->file_reserved_bytes());

This log message will perhaps appear a bit out-of-context in the error log.
Suggested clarification:

  "Could not rename binlog cache to binlog (as requested by
  --binlog-commit-by-rotate-threshold). Required ..."


> +  // Set the cache file as binlog file.
> +  mysql_file_close(mysql_bin_log.log_file.file, MYF(MY_WME));

Do not to pass MY_WME here (instead pass MYF(0)). Since you are ignoring the
return of the function, so we don't want mysql_file_close to call my_error()
leaving an error code set even though the operation succeeds.


> +size_t Binlog_free_flush::get_gtid_event_pad_size()
> +{
> +  size_t begin_pos= my_b_tell(&mysql_bin_log.log_file);
> +  size_t pad_to_size=
> +      m_cache_data->file_reserved_bytes() - begin_pos - LOG_EVENT_HEADER_LEN;
> +
> +  if (binlog_checksum_options != BINLOG_CHECKSUM_ALG_OFF)
> +    pad_to_size-= BINLOG_CHECKSUM_LEN;

Adjusting with BINLOG_CHECKSUM_LEN is to account for the 4 bytes of checksum
at the end of the GTID_EVENT, right?

But why do you subtract LOG_EVENT_HEADER_LEN when calculating pad_to_size?

(I mean, it is probably correct, otherwise nothing would work I think. I
just did not understand it, maybe you can explain in a short comment. Or
maybe Gtid_log_event::write() could be modified to do the adjusting for
LOG_EVENT_HEADER_LEN and BINLOG_CHECKSUM_LEN).


> +    // offset must be saved before replace_binlog_file(), it will update the pos
> +    my_off_t offset= my_b_tell(&log_file);

Suggest to clarify comment to "..., it will update the file position".

> @@ -756,18 +759,20 @@ class MYSQL_BIN_LOG: public TC_LOG, private Event_log
>      new_file() is locking. new_file_without_locking() does not acquire
>      LOCK_log.
>    */
> -  int new_file_impl();
> +  int new_file_impl(bool is_free_flush= false);
>    void do_checkpoint_request(ulong binlog_id);
> -  int write_transaction_or_stmt(group_commit_entry *entry, uint64 commit_id);
> +  int write_transaction_or_stmt(group_commit_entry *entry, uint64 commit_id,
> +                                bool is_free_flush= false);
>    int queue_for_group_commit(group_commit_entry *entry);
>    bool write_transaction_to_binlog_events(group_commit_entry *entry);
> -  void trx_group_commit_leader(group_commit_entry *leader);
> +  void trx_group_commit_leader(group_commit_entry *leader,
> +                               bool is_free_flush= false);
>    bool is_xidlist_idle_nolock();
>    void update_gtid_index(uint32 offset, rpl_gtid gtid);
>  
>  public:
>    void purge(bool all);
> -  int new_file_without_locking();
> +  int new_file_without_locking(bool is_free_flush= false);
>    /*
>      A list of struct xid_count_per_binlog is used to keep track of how many
>      XIDs are in prepared, but not committed, state in each binlog. And how
> @@ -997,7 +1002,8 @@ class MYSQL_BIN_LOG: public TC_LOG, private Event_log
>  	    enum cache_type io_cache_type_arg,
>  	    ulong max_size,
>              bool null_created,
> -            bool need_mutex);
> +            bool need_mutex,
> +            bool is_free_flush = false);
>    bool open_index_file(const char *index_file_name_arg,
>                         const char *log_name, bool need_mutex);
>    /* Use this to start writing a new log file */
> @@ -1037,7 +1043,7 @@ class MYSQL_BIN_LOG: public TC_LOG, private Event_log
>    bool is_active(const char* log_file_name);
>    bool can_purge_log(const char *log_file_name, bool interactive);
>    int update_log_index(LOG_INFO* linfo, bool need_update_threads);
> -  int rotate(bool force_rotate, bool* check_purge);
> +  int rotate(bool force_rotate, bool* check_purge, bool is_free_flush= false);
>    void checkpoint_and_purge(ulong binlog_id);
>    int rotate_and_purge(bool force_rotate, DYNAMIC_ARRAY* drop_gtid_domain= NULL);
>    /**

Style preference: please add the these new boolean parameters without
default value, and instead change existing callers to pass false explicitly.
(And the parameter will be "commit_by_rotate" after name change.)


> +
> +extern ulonglong opt_binlog_free_flush_threshold;
> +static Sys_var_ulonglong Sys_binlog_free_flush_threshold(
> +    "binlog_free_flush_threshold",
> +    "Try to rename the binlog cache temporary file of the commiting "
> +    "transaction to a binlog file when its binlog cache size "
> +    "is bigger than the value of this variable",
> +    GLOBAL_VAR(opt_binlog_free_flush_threshold),
> +    CMD_LINE(REQUIRED_ARG), VALID_RANGE(10 * 1024 * 1024, ULLONG_MAX),
> +    DEFAULT(128 * 1024 * 1024), BLOCK_SIZE(1));

As described above:

 - Suggested name: "binlog_commit_by_rotate_threshold".

 - Suggested text: "For transactions that exceed this size in the binlog,
   commit by a more efficient method that forces a binlog rotation and
   re-uses the pre-existing temporary file as the next binlog file."

 - Let me hear opinions on whether this should default to 128 MB or should
   default to eg. ULONGLONG_MAX to disable the feature by default.


> diff --git a/mysql-test/main/tmp_space_usage.test b/mysql-test/main/tmp_space_usage.test
> index af7b295f343..1685dbbc450 100644
> --- a/mysql-test/main/tmp_space_usage.test
> +++ b/mysql-test/main/tmp_space_usage.test
> @@ -215,7 +215,8 @@ select count(distinct concat(seq,repeat('x',1000))) from seq_1_to_1000;
>  
>  set @save_max_tmp_total_space_usage=@@global.max_tmp_total_space_usage;
>  set @@global.max_tmp_total_space_usage=64*1024*1024;
> -set @@max_tmp_session_space_usage=1179648;
> +# Binlog cache reserve 4096 bytes at the begin of the temporary file.
> +set @@max_tmp_session_space_usage=1179648+65536;
>  select @@max_tmp_session_space_usage;
>  set @save_aria_repair_threads=@@aria_repair_threads;
>  set @@aria_repair_threads=2;
> @@ -224,6 +225,7 @@ set @@max_heap_table_size=16777216;
>  
>  CREATE TABLE t1 (a CHAR(255),b INT,INDEX (b));
>  INSERT INTO t1 SELECT SEQ,SEQ FROM seq_1_to_100000;
> +set @@max_tmp_session_space_usage=1179648;
>  --error 200
>  SELECT * FROM t1 UNION SELECT * FROM t1;
>  DROP TABLE t1;
> @@ -266,11 +268,15 @@ SELECT MIN(VARIABLE_VALUE) OVER (), NTILE(1) OVER (), MAX(VARIABLE_NAME) OVER ()
>  
>  connect(c1, localhost, root,,);
>  set @@binlog_format=row;
> -CREATE OR REPLACE TABLE t1 (a DATETIME) ENGINE=MyISAM;
> +CREATE OR REPLACE TABLE t1 (a DATETIME) ENGINE=InnoDB;
> +# Use the transaction to keep binlog cache temporary file large enough
> +BEGIN;
>  INSERT INTO t1 SELECT NOW() FROM seq_1_to_6000;
> +
>  SET max_tmp_session_space_usage = 64*1024;
>  --error 200
>  SELECT * FROM information_schema.ALL_PLUGINS LIMIT 2;
> +ROLLBACK;
>  drop table t1;

Can you explain why you need these changes to the test case? I was not able
to understand that from the patch.

With these changes, I approve of this patch / pull-request to merge into the
next MariaDB release.

 - Kristian.
```

sql/log.cc
```diff
@@ -9090,8 +9081,26 @@ MYSQL_BIN_LOG::trx_group_commit_leader(group_commit_entry *leader,
       */
       DBUG_ASSERT(!cache_mngr->stmt_cache.empty() ||
                   !cache_mngr->trx_cache.empty()  ||
-                  current->thd->transaction->xid_state.is_explicit_XA() ||
-                  is_free_flush);
+                  current->thd->transaction->xid_state.is_explicit_XA());
```
andrelkin, line 9084: @knielsen, as your review mail mentions
```
The patch calls the trx_group_commit_leader()
```
I think you did not see this block in the latest commit that addresses (no more call of the func) this concern to my understanding without any major refactoring.

sql/log.cc
SongLibing: fine

sql/log.cc
SongLibing: moved the code to where you suggested.

sql/sys_vars.cc
```diff
@@ -7391,3 +7391,13 @@ static Sys_var_enum Sys_block_encryption_mode(
   "AES_ENCRYPT() and AES_DECRYPT() functions",
   SESSION_VAR(block_encryption_mode), CMD_LINE(REQUIRED_ARG),
   block_encryption_mode_values, DEFAULT(0));
+
+extern ulonglong opt_binlog_commit_by_rotate_threshold;
+static Sys_var_ulonglong Sys_binlog_free_flush_threshold(
+    "binlog_large_commit_threshold",
+    "When the size of the commiting transaction's binlog events is larger than "
+    "the threshold, try to commit the transaction in the way moving the "
+    "binlog cache file to a binlog file instead of copying the binlog events",
```
bnestere, line 7400: I think this description can be improved given the name of the variable is now more vague. I suggest something along the lines of
```
Increases transaction concurrency for large transactions (i.e. 
those who's size is larger than this value) by using the large
transaction's cache file as a new binary log, and rotating the
active binary log to the large transaction's cache file at commit
time. This avoids the default commit logic that copies the
transaction cache data to the end of the active binary log file
while holding a lock that prevents other transactions from
binlogging.
```

bnestere (approved):
There are a couple test failures, and I left one last comment about the sys_var description, but otherwise looks good!  Thank you for a great contribution, @SongLibing !

Test failures for convenience:
```
rpl.rpl_binlog_commit_by_rotate 'row'    w35 [ fail ]
        Test ended at 2024-09-05 14:47:35

CURRENT_TEST: rpl.rpl_binlog_commit_by_rotate
--- /home/buildbot/amd64-debian-11-msan-clang-16/build/mysql-test/suite/rpl/r/rpl_binlog_commit_by_rotate.result  2024-09-05 14:34:42.000000000 +0000
+++ /home/buildbot/amd64-debian-11-msan-clang-16/build/mysql-test/suite/rpl/r/rpl_binlog_commit_by_rotate.reject  2024-09-05 14:47:33.669269404 +0000
@@ -83,6 +83,7 @@
 slave-bin.000002 # Table_map # # table_id: # (test.t1)
 slave-bin.000002 # Write_rows_v1 # # table_id: # flags: STMT_END_F
 slave-bin.000002 # Xid # # COMMIT /* XID */
+slave-bin.000002 # Binlog_checkpoint # # slave-bin.000002
 slave-bin.000002 # Rotate  # # slave-bin.000003;pos=POS
 include/show_binlog_events.inc
 Log_name Pos Event_type  Server_id End_log_pos Info
@@ -97,6 +98,7 @@
 slave-bin.000003 # Table_map # # table_id: # (test.t1)
 slave-bin.000003 # Write_rows_v1 # # table_id: # flags: STMT_END_F
 slave-bin.000003 # Xid # # COMMIT /* XID */
+slave-bin.000003 # Binlog_checkpoint # # slave-bin.000003
 include/stop_slave.inc
 SET GLOBAL binlog_large_commit_threshold = @saved_binlog_large_commit_threshold;
 SET GLOBAL slave_parallel_workers = @saved_slave_parallel_workers;

Result length mismatch
```

And 
```
sys_vars.sysvars_server_embedded '64bit,unix' w15 [ fail ]
        Test ended at 2024-09-05 14:42:40

CURRENT_TEST: sys_vars.sysvars_server_embedded
Warning: /home/buildbot/amd64-debian-10-debug-embedded/build/libmysqld/examples/mariadb-test-embedded: unknown option '--loose-disable-ssl-verify-server-cert'
Warning: /home/buildbot/amd64-debian-10-debug-embedded/build/libmysqld/examples/mariadb-test-embedded: unknown variable 'loose-ssl-ca=/home/buildbot/amd64-debian-10-debug-embedded/build/mysql-test/std_data/cacert.pem'
Warning: /home/buildbot/amd64-debian-10-debug-embedded/build/libmysqld/examples/mariadb-test-embedded: unknown variable 'loose-ssl-cert=/home/buildbot/amd64-debian-10-debug-embedded/build/mysql-test/std_data/client-cert.pem'
Warning: /home/buildbot/amd64-debian-10-debug-embedded/build/libmysqld/examples/mariadb-test-embedded: unknown variable 'loose-ssl-key=/home/buildbot/amd64-debian-10-debug-embedded/build/mysql-test/std_data/client-key.pem'
Warning: /home/buildbot/amd64-debian-10-debug-embedded/build/libmysqld/examples/mariadb-test-embedded: unknown option '--loose-skip-ssl'
--- /home/buildbot/amd64-debian-10-debug-embedded/build/mysql-test/suite/sys_vars/r/sysvars_server_embedded.result  2024-09-05 14:34:43.000000000 +0000
+++ /home/buildbot/amd64-debian-10-debug-embedded/build/mysql-test/suite/sys_vars/r/sysvars_server_embedded.reject  2024-09-05 14:42:39.893126446 +0000
@@ -465,7 +465,7 @@
 VARIABLE_NAME  BINLOG_LARGE_COMMIT_THRESHOLD
 VARIABLE_SCOPE GLOBAL
 VARIABLE_TYPE  BIGINT UNSIGNED
-VARIABLE_COMMENT  When the size of the commiting transaction's binlog events is larger than the threshold, try to commit the transaction in the way moving the binlog cache file to a binlog file instead of copying the binlog events
+VARIABLE_COMMENT When the size of the commiting transaction's binlog events is larger than the threshold, try to commit the transaction in the way moving the binlog cache file to a binlog file instead of copying the binlog events
 NUMERIC_MIN_VALUE  10485760
 NUMERIC_MAX_VALUE  18446744073709551615
 NUMERIC_BLOCK_SIZE 1

Result length mismatch
```

sql/sys_vars.cc
```diff
@@ -7391,3 +7391,13 @@ static Sys_var_enum Sys_block_encryption_mode(
   "AES_ENCRYPT() and AES_DECRYPT() functions",
   SESSION_VAR(block_encryption_mode), CMD_LINE(REQUIRED_ARG),
   block_encryption_mode_values, DEFAULT(0));
+
+extern ulonglong opt_binlog_commit_by_rotate_threshold;
+static Sys_var_ulonglong Sys_binlog_free_flush_threshold(
+    "binlog_large_commit_threshold",
+    "When the size of the commiting transaction's binlog events is larger than "
```
andrelkin, line 7398: `max_binlog_size` docs, l.1797 needs adjustment. 
Let me propose here
```       "max_binlog_size",
       "Binary log will be rotated automatically when the size exceeds this "
-       "value",
+      "value, unless `binlog_large_commit_threshold` causes rotation prematurely",

File: sql/sys_vars.cc (hunk 19 above)
Comment (SongLibing) [reply sql/sys_vars.cc]: fixed

File: sql/sys_vars.cc (hunk 20 above)
Comment (SongLibing) [reply sql/sys_vars.cc]: fixed

File: sql/log.h [hunk 21]
```diff
@@ -600,9 +600,12 @@ class binlog_cache_mngr;
 class binlog_cache_data;
 struct rpl_gtid;
 struct wait_for_commit;
+class Binlog_commit_by_rotate;
 
 class MYSQL_BIN_LOG: public TC_LOG, private Event_log
 {
+  friend Binlog_commit_by_rotate;
```
andrelkin, line 607: As I see there is only one instance of the class declared `static` in `log.cc` and it is always accessed
within `MYSQL_BIN_LOG::method`:s. Except one time in ` binlog_cache_reserved_size()`.
Of course locally `Binlog_commit_by_rotate` is a pure sub-class in the sense that its instance (one) can not (does not have to) exist without  `mysql_bin_log` instance.
I wonder why the class was not made as internal to  `MYSQL_BIN_LOG`. The latter function then would access `mysql_bin_log` instance instead of the current way   of  `return binlog_commit_by_rotate.get_reserved_size()`.

sql/log.cc
```diff
@@ -163,6 +163,128 @@ static SHOW_VAR binlog_status_vars_detail[]=
...
+  bool m_replaced{false};
+
+  uint32 m_reserved_bytes {IO_SIZE};
+};
+static Binlog_commit_by_rotate binlog_commit_by_rotate;
+ulonglong opt_binlog_commit_by_rotate_threshold= 128 * 1024 * 1024;
+
+uint32 binlog_cache_reserved_size()
+{
+  return binlog_commit_by_rotate.get_reserved_size();
```
andrelkin, line 285: The note is merely to refer to this point.

sql/log.cc
```diff
@@ -12934,3 +13110,227 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+  /* Seek binlog file to the end */
+  reinit_io_cache(&mysql_bin_log.log_file, WRITE_CACHE,
+                  cache_data->temp_file_length(), false, true);
+  status_var_add(m_entry->thd->status_var.binlog_bytes_written,
+                 cache_data->get_byte_position());
+  m_cache_data->detach_temp_file();
+
+  entry->next= nullptr;
+  mysql_bin_log.trx_group_commit_with_engines(entry, entry, true);
+
```
andrelkin, line 13230: Please put `mysql_mutex_assert_not_owner(&LOCK_log);` at this point.

SongLibing: done

sql/log.h
SongLibing: I didn't make it as a member of MYSQL_BIN_LOG, because MYSQL_BIN_LOG is shared between binlog and relay log. But binlog_commit_by_rotate is not really necessary for relay log.

sql/sys_vars.cc
```diff
@@ -7391,3 +7392,18 @@ static Sys_var_enum Sys_block_encryption_mode(
   "AES_ENCRYPT() and AES_DECRYPT() functions",
   SESSION_VAR(block_encryption_mode), CMD_LINE(REQUIRED_ARG),
   block_encryption_mode_values, DEFAULT(0));
+
+extern ulonglong opt_binlog_commit_by_rotate_threshold;
+static Sys_var_ulonglong Sys_binlog_large_commit_threshold(
+  "binlog_large_commit_threshold",
```
andrelkin, line 7398: I believe we also have to make it clear in [--tmpdir docs lines](https://github.com/MariaDB/server/blob/39d134b19907af1368e037e97939aa2f3e6169d7/sql/sys_vars.cc#L3259) that the user thread's binlog IO_Cache temporary files are relocated into a folder in the same directory with the binlog files.

andrelkin:
```
       "tmpdir", "Path for temporary files.
+      Files that are created in background for binlogging by user threads are placed in a separate location (see `binlog_large_commit_threshold` option).
```

sql/log.cc
```diff
@@ -12934,3 +13110,227 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+  if (!m_replaced)
+  {
+    mysql_mutex_unlock(&mysql_bin_log.LOCK_log);
+    if (check_purge)
+      mysql_bin_log.checkpoint_and_purge(prev_binlog_id);
+    // It cannot do rename, so go to group commit
+    return mysql_bin_log.write_transaction_with_group_commit(entry);
+  }
+
+  /* Seek binlog file to the end */
```
andrelkin, line 13221: I would place this 3 instruction block inside [Binlog_commit_by_rotate::replace_binlog_file](https://github.com/MariaDB/server/blob/39d134b19907af1368e037e97939aa2f3e6169d7/sql/log.cc#L13318) right after `m_replaced= true`.

Esp that it looks a bit worrying  that `detach_temp_file()` can change the size of the new active binlog file after `rotate()`  l.13204 is done. To me it'd be natural to invoke `my_chsize` closer to `m_replaced` sets to `true`. As to the worry, what if the server crashes right after `rotate()`? I imagine (from [here])(https://github.com/MariaDB/server/blob/39d134b19907af1368e037e97939aa2f3e6169d7/sql/log_cache.cc#L50) the renamed file may have some garbage at its tail part.

sql/log.cc
```diff
@@ -7979,14 +8118,12 @@ int Event_log::write_cache(THD *thd, binlog_cache_data *cache_data)
       likely(!crypto.scheme) &&
       likely(!opt_binlog_legacy_event_pos))
   {
-    int res= my_b_copy_all_to_cache(cache, &log_file);
+    int res=
+        my_b_copy_to_cache(cache, &log_file, cache_data->length_for_read());
     status_var_add(thd->status_var.binlog_bytes_written, my_b_tell(cache));
```
andrelkin, line 8123: Apparently `my_b_tell(cache)` the increment to `binlog_bytes_written`  needs adjustment now to subtract `cache_data->length_for_read()` that is not copied to the binlog file.

sql/log.cc
```diff
@@ -12934,3 +13110,227 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+  binlog_cache_data *cache_data= cache_mngr->get_binlog_cache_data(true);
+  if (unlikely(!entry->using_trx_cache || cache_data->empty()))
+    cache_data= cache_mngr->get_binlog_cache_data(false);
+
+  /* Call them before enter log_lock to avoid holding the lock long */
+  if (cache_data->sync_temp_file() || thd->wait_for_prior_commit())
+    return true;
+
+  // It will be released by trx_group_commit_leader
+  mysql_mutex_lock(&mysql_bin_log.LOCK_log);
```
knielsen, line 13170: This should now read: "It will be released by trx_group_commit_with_engines()".

sql/log.cc
```diff
@@ -12934,3 +13110,227 @@ void wsrep_register_binlog_handler(THD *thd, bool trx)
...
+        goto err;
+      }
+      copy_len+= read_len;
+    }
+
+    mysql_file_close(read_fd, MYF(0));
+  }
+
+  // Set the cache file as binlog file.
+  mysql_file_close(mysql_bin_log.log_file.file, MYF(MY_WME));
```
knielsen, line 13293: Don't pass MYF(MY_WME) here, pass MYF(0).

knielsen (approved):
Thanks Libing Song for the updates to the patch and the explanations. And
thanks for the patience with the reviews.

This is a complex piece of work and very well done, showing deep
understanding of the server code and great attention to detail. If you have
more replication patches in the future, I will try to prioritise reviewing
them so the process can hopefully be quicker.

The patch looks good to me to merge. I have made only two very minor
suggestions that you can consider at your choice.

I agree with making the patch enabled by default, with default threshold of
128MB.

 \- Kristian.

sql/log.cc
SongLibing: fixed

sql/log.cc
SongLibing: fixed.

sql/log.cc
SongLibing: fixed

sql/log.cc
SongLibing: Good point, thanks! Originally, detach_temp_file() happened before replace_bnlog_file(), so I put my_chsize() in it.
Now, I would put my_chsize() into Binlog_commit_by_rotate::commit() right before acquired LOCK_log. my_chsize() may take long time if it shrinks large space.

sql/sys_vars.cc
SongLibing: added

sql/log.cc
SongLibing: The 3 line placed here, because the first line cannot be done in replace_binlog_file(),
```  
reinit_io_cache(&mysql_bin_log.log_file, WRITE_CACHE,
                  m_cache_data->temp_file_length(), false, true);" 
```
So I will put my_chsize just into Binlog_commit_by_rotate::commit() just before acquire LOCK_log.

SongLibing: I also added a test case to verify the behavior.

andrelkin:
Thanks LiBing for a very good piece of work! Approved from my side.

andrelkin (approved):

bnestere:
Pushed into 11.7 as 72cc58bb7127b5f17c083ce4b4cc704288347be0

PR #3404, opened 2024-07-16
MDEV-34158: Implement the GIS function ST_Geohash

StefanoPetrilli:
- [x] *The Jira issue number for this PR is: [MDEV-34158](https://jira.mariadb.org/browse/MDEV-34158)

## Description

MariaDB GIS functionalities are limited when compared to other DBMs. To facilitate the migration from MySQL to MariaDB the GIS functionality should be the same. This PR implements the GIS function `ST_GeoHash(longitude, latitude, max_length)` and  `ST_GeoHash(point, max_length) `. The functions return the geohash corresponding to the input values.

The longitude parameter is a numeric value in the interval [180, -180], the latitude is a numeric value in the interval [90, -90].
In the case of a point, the x coordinate is treated as the latitude and the y coordinate is treated as the latitude. Even in the case of a point, the same constraints apply.
The max_length parameter is the upper limit on the resulting string size and cannot exceed 100.

## Release Notes
Adds the function:
`ST_GeoHash(longitude, latitude, max_length)`, `GeoHash(longitude, latitude, max_length)`, `ST_GeoHash(point, max_length)` and `GeoHash(point, max_length)` :  The functions return the geohash corresponding to the input values.

## How can this PR be tested?

## PR quality check


holyfoot (approved):
This implementation looks good enough to push.

PR #3405, opened 2024-07-17
 MDEV-34159: Implement the GIS function ST_LatFromGeoHash

StefanoPetrilli:
- [x] *The Jira issue number for this PR is: [MDEV-34159](https://jira.mariadb.org/browse/MDEV-34159)

## Description

MariaDB GIS functionalities are limited when compared to other DBMs. To facilitate the migration from MySQL to MariaDB the GIS functionality should be the same. This PR implements the GIS function `ST_LatFromGeoHash(geohash_str)`. The GIS function ST_LatFromGeoHash takes in input a geohash and returns its latitude. The latitude is returned as a numeric value in the interval [90, -90].
If the argument is NULL, the return value is NULL. If the argument is invalid, an ER_INCORRECT_TYPE is thrown.

## Release Notes
Adds the function:
`ST_LatFromGeoHash(geohash_str)` and `LatFromGeoHash(geohash_str)`:  takes in input a geohash and returns its latitude.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/main/spatial_utility_function_geohash.result
+++ b/mysql-test/main/spatial_utility_function_geohash.result
@@ -1,3 +1,254 @@
+CREATE TABLE geohashes  (gid INT NOT NULL PRIMARY KEY, hash_value VARCHAR(255));
+INSERT INTO geohashes VALUES
+(1, "000000000000000000000"),
+(2, "zzzzzzzzzzzzzzzzzzzzz"),
+(3, NULL),
+(4, "s00t"),
+(5, "7zzzm"),
+(6, "s00d"),
+(7, "0"),
+(8, "z"),
+(9, "3ejh6z75ddt2d839zh2u"),
+(10, "twtsuqg3q7vh3nrbt0nn"),
+(11, "yw8s10dxddhe4s06nsph"),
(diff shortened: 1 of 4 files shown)


holyfoot (changes requested):
+    char input_character = my_tolower(&my_charset_latin1, (*geohash)[i]);
+
+    if (is_invalid_character(input_character))
+      return true;
+
+    int converted_character= from_geohash_alphabet_to_int(input_character);

Here for each character we do 3 function calls.
Wouldn't it be easier to have one int geohash_alphabet[255] where for each character we'd specify it's geohash value and something negative to mark the invalid characters?
Then we could replace these calls with one 
  converted_character= geohash_alphabet[(*geohash)[i]))

holyfoot (changes requested):
+double Item_func_latlongfromgeohash::val_real()
+{
+  String buf;

I'd recommend to make this 'buf' a member of the class instead of having in local variable.
That reduces the number of malloc calls and heap fragmentation for large tables.

holyfoot:
As we now have the geohash_alphabet, there's no reason to call the 'to_lower'. Just specify the same alphabet codes for uppercase letters.
Otherwise looks ok to push.

PR #3406, opened 2024-07-17
Fix Duplicate Log Entries for Partitioned Tables

HashimSyedUBC:
- [x] *The Jira issue number for this PR is: MDEV-7451*

## Description

Prevent duplicate log entries when creating partitioned tables by
adding a boolean flag, is_partition, to the notify function. Avoid
calling auditing for each partition to eliminate redundant TABLE
event logs.

Amazon Web Services, Inc.

## Release Notes

N/A

## How can this PR be tested?

Added tests in `server_audit.test` that ensure that creating tables with partitions results in no more than 1 log message. All mtr tests pass.

## PR quality check


LinuxJedi (commented):
Looks good for me. As this is a bug fix, it should go into the oldest maintained release, which is 10.5.

Can you please rebase to that? (we might be able to do it if you cannot)

HashimSyedUBC:
> Looks good for me. As this is a bug fix, it should go into the oldest maintained release, which is 10.5.
> 
> Can you please rebase to that? (we might be able to do it if you cannot)

I have just rebased it to 10.5. Thank you for the clarification.

sql/handler.cc
```diff
@@ -5264,7 +5264,8 @@ handler::ha_create(const char *name, TABLE *form, HA_CREATE_INFO *info_arg)
     info_arg->options|= HA_LEX_CREATE_GLOBAL_TMP_TABLE;
   int error= create(name, form, info_arg);
   if (!error &&
-      !(info_arg->options & (HA_LEX_CREATE_TMP_TABLE | HA_CREATE_TMP_ALTER)))
+      !(info_arg->options & (HA_LEX_CREATE_TMP_TABLE | HA_CREATE_TMP_ALTER)) &&
+      !info_arg->is_partition)
```
LinuxJedi, line 5268: Another question, could we not use the following?:
```c
if (share->db_type() == partition_hton)
```
If this works, it would save adding a new variable.

HashimSyedUBC: Hey Andrew thank your for your comment. I implemented your suggestion locally and tested it. Unfortunately, using (share->db_type() != partition_hton) as the condition removes all CREATE logs for the table creation. The desired behavior is to retain one CREATE log and remove the duplicates.

LinuxJedi (approved):
Many thanks for answering my questions.

holyfoot (changes requested):
I don't like to have the specific is_partition member just for that.
We can detect the 'in_partition' state there using the fact that table_share->db_type() and handler::ht are different in this case.
Something like:
if (... table_share->db_type() == partition_hton && ht != partition_hton)

FooBarrior:
@holyfoot we have handler::is_root_handler for that

FooBarrior:
Also let me express my overall opinion on the approach.

All the time, we have some problem with double-something with partitions. Most of things are worked around in ha_partition::*_row, but more and more are found.

It's because the logic lies in ha_*_row. For example, ha_partition::update_row invokes partition's ha_update_row, duplicating that logic's invocation 

Overall, the correct approach would be calling a virtual, protected update_row, which would solve the issue.

holyfoot:
Dear Hashim!
We'd like to have this issue fixed in the next release, so if you won't update this patch with my last suggestion, after July 15-th i'm going to do that myself otherwise it'll miss the release.

vuvova:
We couldn't have waited any longer and implemented a fix ourselves

PR #3407, opened 2024-07-19
MDEV-33971 fix --view-protocol under main.name_const_replacment test

DaveGosselin-MariaDB:
MDEV-33971 fix --view-protocol test failure

    Allow the NAME_CONST unwrap optimization when the client is not
    in the PREPARE step of prepared statement nor in the view
    analysis mode.

### DIFF
--- a/sql/item.cc
+++ b/sql/item.cc
@@ -2214,8 +2214,9 @@ bool Item_name_const::fix_fields(THD *thd, Item **ref)
   */
   if ((thd->where == THD_WHERE::WHERE_CLAUSE ||
        thd->where == THD_WHERE::ON_CLAUSE) &&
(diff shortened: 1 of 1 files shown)


spetrunia:
> Allow the NAME_CONST unwrap optimization when the client is not connected in PS or view modes.

This may be read as if the client who is using Prepared Statement would not get the benefit of the optimization.  They would, but the rewrite will be performed when the client issues a command 
to execute the prepared statement (and not when "preparing" it).

So would changing 
> when the client is not connected in PS or view modes.
to
```
... when the client is not in PREPARE step of prepared statement
 or in the view analysis mode.
```
be a better wording?

spetrunia (approved):
Other than that, Ok to push.

PR #3408, opened 2024-07-19
MDEV-34625 Fix undefined behavior of using uninitialized member variables

HugoWenTD:
- [x] _The Jira issue number for this PR is: MDEV-34625_

## Description

Commit a8a75ba2d causes the MariaDB server to crash, usually with signal 11, at random code locations due to invalid pointer values during any table operation. This issue occurs when the server is built with -O3 and other customized compiler flags.

For example, the command `use db1;` causes server to crash in the `check_table_access` function at line sql_parse.cc:7080 because `tables->correspondent_table` is an invalid pointer value of 0x1.

The crashes are due to undefined behavior from using uninitialized variables. The problematic commit a8a75ba2d introduces code that allocates memory and sets it to 0 using thd-\>calloc before initializing it with a placement new operation. This process depends on setting memory to 0 to initialize member variables not explicitly set in the constructor. However, the compiler can optimize out the memset/bfill, leading to uninitialized values and unpredictable issues.

Once a constructor function initializes an object, any uninitialized variables within that object are subject to undefined behavior. The state of memory before the constructor runs, whether it involves memset or was used for other purposes, is irrelevant after the placement new operation.

Now we ensure the constructor initializes variables correctly by running the reset() function in the constructor to perform the memset/bfill(0) operation. After applying the fix, the crash is gone.

### Detailed explanation of the issue

[Following code](https://github.com/MariaDB/server/blob/a8a75ba2d0dd639b2b14337a6c4f495fcaaff7e4/sql/sql_parse.cc#L8352-L8357 "Follow link") mainly does three things:

1. It allocates memory from mem_root using the `calloc` function implemented in `THD` class.
2. In the same `calloc` , it sets all allocated memory to 0.
3. In the `placement new` it initializes the `TABLE_LIST` at the memory allocated in step 1.

```
  void *memregion= thd->calloc(sizeof(TABLE_LIST));
  TABLE_LIST *ptr= new (memregion) TABLE_LIST(thd, db, fqtn, alias_str,
                                              has_alias_ptr, table, lock_type,
                                              mdl_type, table_options,
                                              info_schema, this,
                                              index_hints_arg, option);
```

There's a critical issue in above code because it depends on step 2 to initialize all the member variables of TABLE_LIST that are not explicitly set in the constructor function. There's no guarantee that the memset/bfill in step 2 will always happen as compiler could optimize it out. Using the uninitialized value can lead to unpredictable issues which are hard to debug.

This behavior can be demonstrated with this test https://gcc.godbolt.org/z/5n87z1raG I wrote to examine the assembly code. The code in MariaDB can be abstracted to the following, though it has many layers wrapped around it and more complex logic, causing slight differences in optimization in the MariaDB build. ( and in default MariaDB build, it seems it's not optimized out so there's no issue, but there are chances to be optimized out with some customized flags which are happening in our build )

_To summarize, on x86, the memset in the following code is optimized out with both -O2 and -O3 in GCC 13, and is only preserved in the much older GCC 4.9._

```
struct S {
  int i;     // uninitialized in consturctor
  S() {};
};

int bar() {
  void *buf = malloc(sizeof(S));
  memset(buf, 0, sizeof(S));       // optimized out
  S* s = new(buf) S;
  return s->i;
}
```

GCC13 -O3

```
bar():
        sub     rsp, 8
        mov     edi, 4
        call    malloc
        mov     eax, DWORD PTR [rax]
        add     rsp, 8
        ret
```

GCC 4.9 -O3

```
bar():
        sub     rsp, 8
        mov     edi, 4
        call    malloc
        mov     DWORD PTR [rax], 0
        xor     eax, eax
        add     rsp, 8
        ret
```

## Release Notes

What should the release notes say about this change?

N/A

## How can this PR be tested?

This change does not change any server behavior so all MTR tests should pass.

- [ ] _This is a new feature and the PR is based against the latest MariaDB development branch._
- [x] _This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced._

The fix is targeting on 10.6 which is the lowest major version includes the problematic commit a8a75ba2d.

## PR quality check

## Copyright

All new code of the whole pull request, including one or several files that are either new files or modified ones, are contributed under the BSD-new license. I am contributing on behalf of my employer Amazon Web Services.

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -8375,7 +8375,7 @@ TABLE_LIST *st_select_lex::add_table_to_list(THD *thd,
   }
 
   bool has_alias_ptr= alias != nullptr;
-  void *memregion= thd->calloc(sizeof(TABLE_LIST));
+  void *memregion= thd->alloc(sizeof(TABLE_LIST));
   TABLE_LIST *ptr= new (memregion) TABLE_LIST(thd, db, fqtn, alias_str,
                                               has_alias_ptr, table, lock_type,
                                               mdl_type, table_options,
--- a/sql/table.cc
+++ b/sql/table.cc
@@ -5865,6 +5865,7 @@ TABLE_LIST::TABLE_LIST(THD *thd,
                        List<Index_hint> *index_hints_ptr,
                        LEX_STRING *option_ptr)
 {
+  reset();
   db= db_str;
   is_fqtn= fqtn;
   alias= alias_str;


sql/sql_parse.cc
```diff
@@ -8375,7 +8375,7 @@ TABLE_LIST *st_select_lex::add_table_to_list(THD *thd,
   }
 
   bool has_alias_ptr= alias != nullptr;
-  void *memregion= thd->calloc(sizeof(TABLE_LIST));
```
DaveGosselin-MariaDB, line 8378: This `calloc` was present at line 8328 prior to the change, yet without crashes.  Is there something about the original code that prevents the compiler from optimizing away the memset/bfill?

DaveGosselin-MariaDB: Can you please send along the other compiler flags that you mentioned in your commit message?  Thanks.

sql/table.cc
```diff
@@ -5865,6 +5865,7 @@ TABLE_LIST::TABLE_LIST(THD *thd,
                        List<Index_hint> *index_hints_ptr,
                        LEX_STRING *option_ptr)
 {
+  reset();
```
DaveGosselin-MariaDB, line 5868: This is an inline method that calls bzero, how can we be sure that the compiler won't optimize it away?

sql/sql_parse.cc
HugoWenTD: >  Is there something about the original code that prevents the compiler from optimizing away the memset/bfill?

#### It used to work before https://github.com/MariaDB/server/commit/a8a75ba2d0dd639b2b14337a6c4f495fcaaff7e4 because it was using the C standard way to allocate memory and set variable values. 
The behavior was similar to following code https://gcc.godbolt.org/z/rWvcadcvb:
```
struct S {
  int i;
  S() {};
};

S* bar() {
  void *buf = malloc(sizeof(S));
  memset(buf, 0, sizeof(S));
  S* s = (S*) buf;     // later simply reads the allocated memory but no `new` function call
  return s;
}
```
which will be complied to `calloc` call (malloc + memset):
```
bar():
        mov     esi, 1
        mov     edi, 4
        jmp     calloc
```

#### With https://github.com/MariaDB/server/commit/a8a75ba2d0dd639b2b14337a6c4f495fcaaff7e4, the code uses the C++ `placement new` function. And the C++ `new` appears to rely on the constructor to initialize the memory contents of the new object. The compiler may optimize away any changes to the memory made before the constructor function is called.

Similar to following code https://gcc.godbolt.org/z/9rKhPc8s1:
```
struct S {
  int i;
  S() {};
};

S* bar() {
  void *buf = malloc(sizeof(S));
  memset(buf, 0, sizeof(S)); 
  S* s = new (buf) S;        // C++ way to initialize the object
  return s;
}
```
Which will be translate to a `malloc` call:
```
bar():
        mov     edi, 4
        jmp     malloc
```

About the optimization differences in GCC versions, GCC may argue that, rather than calling it a regression comparing GCC 4.9 to GCC 13, it's more like an optimizer enhancement leading to behavior change. As using C++ `new` to initialize the object means constructor function should handle the whole initialization so memory contents before `new` does not matter and is not guaranteed to be maintained, and using uninitialized variables of the object is still undefined behavior.

HugoWenTD: > Can you please send along the other compiler flags that you mentioned in your commit message? Thanks.

There are lots of customization in CMake files in our build to adapt our internal build tools.

In general we are using GCC 13.2.0 and Link Time Optimization with -flto=auto. 

Some build flags from the build logs, when building sql_parse.cc.o:
```
-DHAVE_CONFIG_H -DHAVE_EVENT_SCHEDULER -DHAVE_OPENSSL -DHAVE_POOL_OF_THREADS -DMYSQL_SERVER -D_FILE_OFFSET_BITS=64 -IXXX -pthread -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fdiagnostics-color=auto -fno-operator-names -D_GNU_SOURCE -flto=auto -pie -fPIC -fstack-protector --param=ssp-buffer-size=4 -O3 -march=core2 -mtune=generic -DNDEBUG -g -fgcse-after-reload -fipa-cp-clone -floop-interchange -floop-unroll-and-jam -fpeel-loops -fpredictive-commoning -fsplit-loops -fsplit-paths -ftree-loop-distribution -ftree-partial-pre -funswitch-loops -fvect-cost-model=dynamic -fversion-loops-for-strides -flto=auto -fno-omit-frame-pointer -D_FORTIFY_SOURCE=2 -DDBUG_OFF -Wall -Wenum-compare -Wenum-conversion -Wextra -Wmissing-braces -Wno-format-truncation -Wno-init-self -Wno-nonnull-compare -Wno-unused-parameter -Woverloaded-virtual -Wnon-virtual-dtor -Wvla -Wwrite-strings -std=gnu++11 -MD -MT
```
when linking mariadbd:
```
-pthread -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fdiagnostics-color=auto -fno-operator-names -D_GNU_SOURCE -flto=auto -pie -fPIC -fstack-protector --param=ssp-buffer-size=4 -O3 -march=core2 -mtune=generic -DNDEBUG -g -O2 -fgcse-after-reload -fipa-cp-clone -floop-interchange -floop-unroll-and-jam -fpeel-loops -fpredictive-commoning -fsplit-loops -fsplit-paths -ftree-loop-distribution -ftree-partial-pre -funswitch-loops -fvect-cost-model=dynamic -fversion-loops-for-strides -flto=auto -fno-omit-frame-pointer -D_FORTIFY_SOURCE=2 -DDBUG_OFF -Wall -Wenum-compare -Wenum-conversion -Wextra -Wmissing-braces -Wno-format-truncation -Wno-init-self -Wno-nonnull-compare -Wno-unused-parameter -Woverloaded-virtual -Wnon-virtual-dtor -Wvla -Wwrite-strings -pthread -rdynamic -Wl,--as-needed -flto=auto -Wl,-z,relro,-z,now  -static-libstdc++ -fuse-ld=bfd  -Wl,--export-dynamic sql/CMakeFiles/mariadbd.dir/main.cc.o -o sql/mariadbd -LXXX -Wl,-rpath,XXX  sql/libsql.a  sql/libsql_builtins.a  LIB  vio/libvio.a  LIB  -lcrypt  sql/libwsrep.a  sql/libonline_alter_log.a  storage/maria/libaria.a  mysys_ssl/libmysys_ssl.a  LIB  LIB  sql/libpartition.a  storage/perfschema/libperfschema.a  sql/libsql_sequence.a  storage/csv/libcsv.a  storage/heap/libheap.a  storage/innobase/libinnobase.a  tpool/libtpool.a  LIB  LIB  storage/myisam/libmyisam.a  mysys/libmysys.a  dbug/libdbug.a  strings/libstrings.a  mysys/libmysys.a  dbug/libdbug.a  strings/libstrings.a  LIB  -lm  -ldl  storage/myisammrg/libmyisammrg.a  storage/sequence/libsequence.a  plugin/auth_socket/libauth_socket.a  plugin/provider_bzip2/libprovider_bzip2.a  -Wl,-Bstatic  -lbz2  plugin/provider_lz4/libprovider_lz4.a  -llz4  plugin/provider_lzma/libprovider_lzma.a  -llzma  plugin/provider_lzo/libprovider_lzo.a  -llzo2  plugin/rds_performance_insights/librds_performance_insights.a  plugin/type_geom/libtype_geom.a  plugin/type_inet/libtype_inet.a  plugin/type_uuid/libtype_uuid.a  plugin/user_variables/libuser_variables.a  plugin/userstat/libuserstat.a  sql/libthread_pool_info.a  -Wl,-Bdynamic
```

sql/table.cc
HugoWenTD: Thank you @vuvova for sharing your insights, I had the same understanding as you.

It can also be verified with following examples:

Simply moving the memset (or bzero) into the constructor function, from:
https://gcc.godbolt.org/z/6vY5Ma517
```
struct S {
  int i;
  S() {};
};

S* bar() {
  void *buf = malloc(sizeof(S));
  memset(buf, 0, sizeof(S)); 
  S* s = new (buf) S;
  return s;
}
```
to:
https://gcc.godbolt.org/z/aPss1sj59
```
struct S {
  int i;
  S() {  memset(this, 0, sizeof(S)); };
};

S* bar() {
  void *buf = malloc(sizeof(S));
  S* s = new(buf) S;
  return s;
}
```

The assembly code changes from a simple malloc:
```
# Assembly code:
bar():
        mov     edi, 4
        jmp     malloc
```
to malloc + set 0:
```
bar():
        sub     rsp, 8
        mov     edi, 4
        call    malloc
        mov     DWORD PTR [rax], 0
        add     rsp, 8
        ret
```

DaveGosselin-MariaDB (approved):

dr-m:
For the record, I believe that the relevant GCC flag is `-flifetime-dse` (dead store elimination). See e070cfe398072b6020193f7eac5c3a1f7281979b.

dr-m:
Sorry, a better example is 32de60bb2e9f06b24c76fc41bef81e35c14ca528 ([MDEV-18749](https://jira.mariadb.org/browse/MDEV-18749)).

HugoWenTD:
Yes. Looks exactly very similar issue. Thank you @dr-m!

PR #3411, opened 2024-07-22
 MDEV-34177: Implement the GIS function ST_PointFromGeoHash

StefanoPetrilli:
- [x] *The Jira issue number for this PR is: [MDEV-34177](https://jira.mariadb.org/browse/MDEV-34177)

## Description

MariaDB GIS functionalities are limited when compared to other DBMs. To facilitate the migration from MySQL to MariaDB, the GIS functionality should be the same. This PR implements the GIS function `ST_PointFromGeoHash(geohash_str)`. The GIS function ST_PointFromGeoHash takes in input a geohash and returns a `ST_Point` where the x and y coordinate are respectively the longitude in the range [−180, 180] and the latitude in the range [−90, 90]. If the argument is NULL, the return value is NULL. If the argument is invalid, an ER_GIS_INVALID_DATA is thrown.

## Release Notes
Adds the function:
`ST_PointFromGeoHash(geohash_str)` and `ST_PointFromGeoHash(geohash_str)`:  takes in input a geohash and returns a `ST_Point` where the x and y coordinate are respectively the longitude in the range [−180, 180] and the latitude in the range [−90, 90].

## How can this PR be tested?

## PR quality check


holyfoot (changes requested):
+String *Item_func_pointfromgeohash::val_str(String *str)
+{
+  DBUG_ASSERT(fixed());
+  String buf;
Please move the buf to be the class member instead of the local variable.
otherwise looks ok to push.

PR #3420, opened 2024-07-24
MDEV-34181  Instant table aborts after discard tablespace

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34181*

## Description
- commit 85db5347311340e39753b0200fb9d459a5024535 (MDEV-33400) avoid the clearing of the instant alter metadata during discard tablespace. But it fails to remove the assignment of n_core_null_bytes with total nullable fields exist in the index for discarded tablespace.

## How can this PR be tested?
./mtr innodb.import_bugs

## PR quality check

### DIFF
--- a/storage/innobase/handler/handler0alter.cc
+++ b/storage/innobase/handler/handler0alter.cc
@@ -374,10 +374,12 @@ found_j:
 			}
 		}
 
-		/* In case of discarded tablespace, InnoDB can't
-		read the root page. So assign the null bytes based
-		on nullabled fields */
-		if (!oindex.table->space) {
+		/* Discard tablespace doesn't remove the instantness
+		from the table definition. if n_core_null_bytes wasn't
+		initialized then assign it based on nullable fields */
+		if (!oindex.table->space
+		    && oindex.n_core_null_bytes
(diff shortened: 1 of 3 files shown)


storage/innobase/handler/handler0alter.cc
```diff
@@ -374,14 +374,6 @@ inline void dict_table_t::prepare_instant(const dict_table_t& old,
...
 		}
 
-		/* In case of discarded tablespace, InnoDB can't
-		read the root page. So assign the null bytes based
-		on nullabled fields */
-		if (!oindex.table->space) {
-			oindex.n_core_null_bytes = static_cast<uint8_t>(
-				UT_BITS_IN_BYTES(unsigned(oindex.n_nullable)));
-		}
-
```
dr-m, line 384: It would be good if the commit message explained the sequence of events that goes wrong here. (Okay, one could get that relatively easily from `./mtr --rr`.)

It would be even better to explain why this code was needed in the first place. I see that it was added in 99de8cc02878ec8c20bffeaf17443e704d6edbf7 but the commit message does not mention that.

dr-m (approved):
OK to push after fixing `./mtr --embedded innodb.import_bugs`.

PR #3421, opened 2024-07-25
MDEV-33755 Some test cause MSAN alarms due to uninitialized Item_func…

Olernov:
- [x] *The Jira issue number for this PR is: MDEV-33755*

## Description
Commit 09d991d01c47e22030879e5bf0c7a4893a598199 resolved false MSAN alarms by initializing `Item_func::not_null_tables_cache` with 0 during construction. However, this solution inadvertently disabled Valgrind diagnostics, which we want to retain to detect potential issues (e.g., failing to set not_null_tables_cache after calling Item_func::quick_fix_field()).

This commit introduces the macro `MEM_UNDEFINED_VALGRIND_ONLY`, enabling Valgrind diagnostic without triggering MSAN.
Now, Item_func::not_null_tables_cache is marked as undefined during construction using this macro

## Release Notes

## How can this PR be tested?

No need for special test cases, it's enough that all tests with Valgrind diagnostic enabled pass

## PR quality check

### DIFF
--- a/include/my_valgrind.h
+++ b/include/my_valgrind.h
@@ -29,6 +29,7 @@
 # define HAVE_valgrind
 # define HAVE_MEM_CHECK
 # define MEM_UNDEFINED(a,len) __msan_allocated_memory(a,len)
+# define MEM_UNDEFINED_VALGRIND_ONLY(a, len) ((void) 0)
 # define MEM_MAKE_ADDRESSABLE(a,len) MEM_UNDEFINED(a,len)
 # define MEM_MAKE_DEFINED(a,len) __msan_unpoison(a,len)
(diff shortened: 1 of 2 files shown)


dr-m (changes requested):
The MemorySanitizer diagnostics that is mentioned in the description is not bogus. According to https://stackoverflow.com/questions/27421419/is-passing-an-uninitialised-variable-in-a-function-parameter-list-well-defined it is undefined behaviour to pass an uninitialized function parameter by value. I could not find the quoted text in the latest ISO/IEC 14882 working draft at https://eel.is/c++draft/dcl.init, though.

I don't think that it is a good idea to attempt to work around any undefined behaviour.

PR #3422, opened 2024-07-26
MDEV-14231 MATCH() AGAINST( IN BOOLEAN MODE), results mismatch

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-14231*

## Description
- Added plugin_debug.test, multiple_index.test to innodb_fts suite from mysql-5.7.
- multiple_index result changes due to stat_n_rows value. Rollback properly sets stat_n_rows in MariaDB.
- plugin_debug result changes is due to position wasn't stored in fts default parser. Simple parser doesn't support phrase search query.

## How can this PR be tested?
./mtr innodb_fts.plugin_debug innodb_fts.multiple_index

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb_fts/r/multiple_index.result
+++ b/mysql-test/suite/innodb_fts/r/multiple_index.result
@@ -1,0 +1,210 @@
+CREATE TABLE t1 (
+id INT UNSIGNED AUTO_INCREMENT NOT NULL PRIMARY KEY,
+a VARCHAR(200),
+b TEXT
+) ENGINE = InnoDB STATS_PERSISTENT=0;
+INSERT INTO t1 (a,b) VALUES
+('MySQL Tutorial','DBMS stands for DataBase ...')  ,
+('How To Use MySQL Well','After you went through a ...'),
+('Optimizing MySQL','In this tutorial we will show ...');
+ALTER TABLE t1 ADD FULLTEXT INDEX idx_1 (a);
+ALTER TABLE t1 ADD FULLTEXT INDEX idx_2 (b);
+SHOW CREATE TABLE t1;
+Table	Create Table
+t1	CREATE TABLE `t1` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `a` varchar(200) DEFAULT NULL,
+  `b` text DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  FULLTEXT KEY `idx_1` (`a`),
+  FULLTEXT KEY `idx_2` (`b`)
+) ENGINE=InnoDB AUTO_INCREMENT=4 DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci STATS_PERSISTENT=0
+START TRANSACTION;
+INSERT INTO t1 (a,b) VALUES
+('1001 MySQL Tricks','1. Never run mysqld as root. 2. ...'),
+('MySQL vs. YourSQL','In the following database comparison ...'),
+('MySQL Security','When configured properly, MySQL ...');
+ROLLBACK;
+SELECT * FROM t1 WHERE MATCH (a)
+AGAINST ('Tutorial' IN NATURAL LANGUAGE MODE);
+id	a	b
+1	MySQL Tutorial	DBMS stands for DataBase ...
+select * from t1 where MATCH(a) AGAINST("+mysql +Tutorial" IN BOOLEAN MODE);
+id	a	b
+1	MySQL Tutorial	DBMS stands for DataBase ...
+select * from t1 where MATCH(b) AGAINST("+Tutorial" IN BOOLEAN MODE);
+id	a	b
+3	Optimizing MySQL	In this tutorial we will show ...
+select * from t1 where MATCH(b) AGAINST("+stands +(DataBase)" IN BOOLEAN MODE);
+id	a	b
+1	MySQL Tutorial	DBMS stands for DataBase ...
+select * from t1 where MATCH(b) AGAINST("+DataBase -(comparison)" IN BOOLEAN MODE);
+id	a	b
+1	MySQL Tutorial	DBMS stands for DataBase ...
+select *, MATCH(a) AGAINST("Optimizing MySQL" IN BOOLEAN MODE) as x from t1;
+id	a	b	x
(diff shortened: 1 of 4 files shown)


mysql-test/suite/innodb_fts/t/multiple_index.test
```diff
@@ -0,0 +1,173 @@
...
+ALTER TABLE t1 ADD FULLTEXT INDEX idx_2 (b);
+EVAL SHOW CREATE TABLE t1;
+
+# check mutiple index with transaction
+START TRANSACTION;
+# Insert rows
+INSERT INTO t1 (a,b) VALUES
+        ('1001 MySQL Tricks','1. Never run mysqld as root. 2. ...'),
+        ('MySQL vs. YourSQL','In the following database comparison ...'),
+	('MySQL Security','When configured properly, MySQL ...');
```
dr-m, line 34: There is no TAB in the MySQL 5.7 test:
```diff
@@ -31,7 +31,7 @@
 INSERT INTO t1 (a,b) VALUES
         ('1001 MySQL Tricks','1. Never run mysqld as root. 2. ...'),
         ('MySQL vs. YourSQL','In the following database comparison ...'),
-	('MySQL Security','When configured properly, MySQL ...');
+        ('MySQL Security','When configured properly, MySQL ...');
 ROLLBACK;
 
 # Select word "tutorial" in the table
```
I don’t think we should add one either.

mysql-test/suite/innodb_fts/r/multiple_index.result
```diff
@@ -0,0 +1,211 @@
...
+id	a	b
+1	MySQL Tutorial	DBMS stands for DataBase ...
+select * from t1 where MATCH(b) AGAINST("+DataBase -(comparison)" IN BOOLEAN MODE);
+id	a	b
+1	MySQL Tutorial	DBMS stands for DataBase ...
+select *, MATCH(a) AGAINST("Optimizing MySQL" IN BOOLEAN MODE) as x from t1;
+id	a	b	x
+1	MySQL Tutorial	DBMS stands for DataBase ...	0.000000001885928302414186
+2	How To Use MySQL Well	After you went through a ...	0.000000001885928302414186
+3	Optimizing MySQL	In this tutorial we will show ...	0.22764469683170319
```
dr-m, line 49: The commit message should specifically say that the `MATCH(a)` values are smaller than in MySQL because `ROLLBACK` updates the `stat_n_rows`. There are also some other, more benign, result differences.

mysql-test/suite/innodb_fts/t/plugin_debug.test
```diff
@@ -0,0 +1,293 @@
...
+        id INT UNSIGNED AUTO_INCREMENT NOT NULL PRIMARY KEY,
+        title VARCHAR(200),
+        body TEXT,
+        FULLTEXT (title, body) WITH PARSER simple_parser
+        ) ENGINE=InnoDB DEFAULT CHARACTER SET gb2312 COLLATE gb2312_chinese_ci;
+
+INSERT INTO articles (title, body) VALUES
+  ('数据库 是 数据 的 结构化 集合','它 可以 是 任何 东西'),
+  ('从 简单 的 购物清单 到 画展','或 企业网络 中 的 海量信息'),
+  ('要 想 将 数据 添加 到 数据库','或 访问、处理 计算机 数据库 中 保存 >的 数据'),
```
dr-m, line 246: There is no `>` in the MySQL 5.7 test file. There are also some white space differences, which I think must be reverted. The following difference is acceptable:
```diff
@@ -2,9 +2,7 @@
 --source include/have_debug.inc
 --source include/have_simple_parser.inc
 
-# Install fts parser plugin
---replace_regex /\.dll/.so/
-eval INSTALL PLUGIN simple_parser SONAME '$SIMPLE_PARSER';
+INSTALL PLUGIN simple_parser SONAME 'mypluglib';
 
 -- echo # Test Part 2: Create Index Test(CREATE TABLE WITH FULLTEXT INDEX)
 CREATE TABLE articles (
```

mysql-test/suite/innodb_fts/r/plugin_debug.result
```diff
@@ -0,0 +1,306 @@
...
+1	数据库 是 数据 的 结构化 集合	它 可以 是 任何 东西
+4	需要 使用 数据库 管理系统	计算机 是 处理 大量 数据 的 理想 工具
+SELECT * FROM articles WHERE
+MATCH(title, body) AGAINST('+数据库 -计算机');
+id	title	body
+1	数据库 是 数据 的 结构化 集合	它 可以 是 任何 东西
+SELECT * FROM articles WHERE
+MATCH(title, body) AGAINST('"计算机 数据库"' IN BOOLEAN MODE);
+id	title	body
+DROP TABLE articles;
```
dr-m, line 291: In MySQL 5.7, this would return the row where we inadvertently injected the `>` character. But, as far as I can tell, the character was not added in the middle of this phrase.

The commit message does mention some missing results, but I think it would be clearer if it used some specific syntax to refer to that. I see that many but not all `IN BOOLEAN MODE` queries return an empty result or fewer results than MySQL. Is there a more precise way to explain this difference?

mysql-test/suite/innodb_fts/r/plugin_debug.result
```diff
@@ -0,0 +1,306 @@
...
+1	MySQL Tutorial	DBMS stands for MySQL DataBase ...
+SELECT * FROM articles WHERE
+MATCH(title, body) AGAINST('+("dbms stands" search) -good' IN BOOLEAN MODE);
+id	title	body
+5	1001 MySQL Tricks	How to use full-text search engine
+1	MySQL Tutorial	DBMS stands for MySQL DataBase ...
+SELECT * FROM articles WHERE
+MATCH(title, body) AGAINST('+("dbms stands" "full text") -good' IN BOOLEAN MODE);
+id	title	body
+1	MySQL Tutorial	DBMS stands for MySQL DataBase ...
```
dr-m, line 251: In MySQL 5.7, this and many other searches for `full text` or `full text search` would also return another row:
```
5	1001 MySQL Tricks	How to use full-text search engine
```
This difference is not explained in the commit message. Is it a difference in tokenization? Is there a way to make `-` considered a word separator? Could additional searches for `full-text` be added to demonstrate that the records would be found?

dr-m (approved):

PR #3360, opened 2024-06-26
[MDEV-21978 part 2] Enable and Fix GCC `-Wformat` Checks on `my_vsnprintf` and Users

ParadoxV5:
* *The Jira issue number for this PR is: [MDEV-21978](https://jira.mariadb.org/browse/MDEV-21978)*
  * [Zulip thread](https://mariadb.zulipchat.com/#narrow/stream/118759-general/topic/MDEV-21978.20make.20my_vsnprintf.20to.20use.20gcc-compatible.20format)
  * Possibly supercedes [MDEV-19784](https://jira.mariadb.org/browse/MDEV-19784)
* Resolves https://github.com/MariaDB/server/pull/1531#issuecomment-626620929
* Prerequisite (MDEV-21978 & GSoC part 1): #3309
  * This PR bases on that prerequisite PR and currently shows up inheriting those commits.
    I will keep this branch updated with that prerequisite branch.
    When that prerequisite merges, I will rebase this PR to the latest branch directly.
* [Google Summer of Code (GSoC) 2024](https://summerofcode.withgoogle.com/programs/2024/projects/BNEGRwkS) milestone 2
  * … And it even assigns me a responsibility I’ve rarely taken before – housekeeping. Everyone dreads chores, amirite?
  * My experience was also regular C, not C++, and I thought I wouldn’t need to deal with C++ for this project. :face_in_clouds:

<details><summary>

### Why wasn’t this project merged during GSoC?
(Yes, I’m linking this PR as my GSoC final submission.)

</summary>

[_Sporadic (inconsistent)_ CI failures *on the base branch*](https://buildbot.mariadb.org/#/grid?branch=11.6) have been *consistently* plaguing the MariaDB/server repo _before and during_ this GSoC event. How do we like them :apple:s?
These issues confuse both new (such as me and other GSoC participants) and old contributors, leading to wild *geese* chases and even alert fatigue.[^m]

[^m]: O. Kekäläinen et al., “Can we get MariaDB GitHub CI to consistently be green?” [Mailing list]. Available: https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/thread/NRMKHZ6JRDEWQ73HKV4XYVKHSEY7X3WF/

MariaDB organizers are aiming to improve coördination and address those bugs promptly.
Unfortunately, this shift in focus puts new features such as most of the GSoC projects in the back seat.
For more information, folks are tracking those failures at
[...]

PR #3429, opened 2024-07-28
Use `strmake` over `my_snprintf` to copy strings

ParadoxV5:
* [ ] *The Jira issue number for this PR is: ~~MDEV-21978~~*
  * This commit was rebased out of https://github.com/MariaDB/server/pull/3360#discussion_r1660019479 to make that PR more focused.
    * This commit will likely conflict with that PR; in which case, these `strmake`s take precedence.

There is no need to parse a format string if we just need to clone chars over (and truncate if the buffer overflows).
In fact, when #3360 enabled `ATTRIBUTE_FORMAT` on `my_snprintf`, `-Werror=format-security` complained several of these that “format not a string literal and no format arguments”.

There are alternatives to `strmake`, but I’ve determined on [Zulip chat for #3360](https://mariadb.zulipchat.com/#narrow/stream/118759-general/topic/MDEV-21978.20make.20my_vsnprintf.20to.20use.20gcc-compatible.20format/near/448003346) that `strmake` is the best fit.
(Although, in one case, `strcpy` suffice because it’s the first write to the buffer and will countably not overflow.)

Unless I messed up – not at all? I just swapped function A with function B.

## Release Notes
Um… General code cleanup?

## How can this PR be tested?
This switch does not require new tests – existing tests will do.

* [x] *This is a new feature or a refactoring, and the PR is based against the latest MariaDB development branch.*
  * Uh, I suppose this is a refactoring.
* [ ] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check


storage/connect/tabrest.cpp
```diff
@@ -66,7 +66,7 @@ int Xcurl(PGLOBAL g, PCSZ Http, PCSZ Uri, PCSZ filename)
 			my_snprintf(buf, sizeof(buf)-1, "%s/%s", Http, Uri);
 
 	} else
-		my_snprintf(buf, sizeof(buf)-1, "%s", Http);
+		strmake(buf, Http, sizeof(buf)-2);
```
ParadoxV5, line 69: questionable double-`-1`

PR #3430, opened 2024-07-29
MDEV-34565: SIGILL due to OS not supporting AVX512

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34565*
## Description
It is not sufficient to check that the CPU supports the necessary instructions. Also the operating system (or virtual machine hypervisor) must enable all the AVX registers to be saved and restored on a context switch.

Because clang 8 does not support the compiler intrinsic `_xgetbv()` we will require clang 9 or later for enabling the use of `vpclmulqdq` and the related AVX512 features.
## Release Notes
MariaDB Server would catch SIGILL on a system where the CPU but not the operating system supports AVX-512.
## How can this PR be tested?
`InnoDB: Using AVX512 instructions` should continue to be reported on the `amd64-debian-11-msan` builder.

Unfortunately, I do not have the ability to change the boot parameters of an environment to test that `_xgetbv()` would prevent the use of AVX512 instructions when the operating system does not enable saving and restoring the registers.
## PR quality check

### DIFF
--- a/mysys/crc32/crc32c_x86.cc
+++ b/mysys/crc32/crc32c_x86.cc
@@ -28,7 +28,8 @@
 # elif __GNUC__ >= 14 || (defined __clang_major__ && __clang_major__ >= 18)
 #  define TARGET "pclmul,evex512,avx512f,avx512dq,avx512bw,avx512vl,vpclmulqdq"
 #  define USE_VPCLMULQDQ __attribute__((target(TARGET)))
-# elif __GNUC__ >= 11 || (defined __clang_major__ && __clang_major__ >= 8)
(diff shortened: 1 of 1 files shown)


dr-m:
https://ci.mariadb.org/48820/logs/amd64-debian-11-msan/mysqld.1.err.1 contains the following:
```
2024-07-29  8:21:45 0 [Note] InnoDB: Using AVX512 instructions
```
The timestamp seems to be in UTC.

vaintroub (commented):
OK to push. would be better if there was a test env, where we can switch AVX on or off, so we don't do this blindly, but we got what we got.

vaintroub (approved):

PR #3432, opened 2024-07-29
MDEV-34664: Add an option to fix InnoDB's doubling of secondary index…

spetrunia:
… cardinalities

(This is Monty's patch with some trivial fix-ups)

Added option fix_innodb_cardinality to optimizer_adjust_secondary_key_costs

Using fix_innodb_cardinality disables the 'divide by 2' of rec_per_key_int in InnoDB that in effect doubles the Cardinality for secondary keys. This has the biggest effect for indexes where a few rows has the same key value. Using this may also cause table scans for very small tables (which in some cases may be better than an index scan).

The user visible effect is that 'SHOW INDEX FROM table_name' will for InnoDB show the true Cardinality (and not 2x the real value). It will also allow the optimizer to chose a better index in some cases as the division by 2 could have a bad effect for tables with 2-5 identical values per key.

A few notes about using fix_innodb_cardinality:
- It has direct affect for SHOW INDEX FROM table_name. SHOW INDEX will also update the statistics in table share.
- The effect of fix_innodb_cardinality for query plans or EXPLAIN is only visible after first open of the table. This is why one must do a flush tables or use SHOW INDEX for the option to take effect.
- Using fix_innodb_cardinality can thus affect all user in their query plans if they are using the same tables.

Because of this, it strongy recommended that one uses optimizer_adjust_secondary_key_costs=fix_innodb_cardinality mainly in configuartion files to not cause issues for other users.

- [x] *The Jira issue number for this PR is: MDEV-34664*

### DIFF
--- a/mysql-test/main/mysqld--help.result
+++ b/mysql-test/main/mysqld--help.result
@@ -719,9 +719,10 @@ The following specify which files/extra groups are read (specified before remain
  costs. disable_max_seek = Disable 'max_seek optimization'
  for secondary keys and slight adjustment of filter cost.
  disable_forced_index_in_group_by = Disable automatic
- forced index in GROUP BY. This variable will be deleted
- in MariaDB 11.0 as it is not needed with the new 11.0
- optimizer.
+ forced index in GROUP BY. fix_innodb_cardinality =
+ Disable doubling of the Cardinality for InnoDB secondary
+ keys. This variable will be deleted in MariaDB 11.0 as it
+ is not needed with the new 11.0 optimizer.
  Use 'ALL' to set all combinations.
  --optimizer-max-sel-arg-weight=# 
  The maximum weight of the SEL_ARG graph. Set to 0 for no
--- a/mysql-test/main/secondary_key_costs.test
+++ b/mysql-test/main/secondary_key_costs.test
@@ -1,6 +1,8 @@
 --source include/have_sequence.inc
 --source include/not_embedded.inc
 --source include/have_innodb.inc
+# Testcase for MDEV-33306 takes ~6 minutes with valgrind:
+--source include/not_valgrind.inc
 
 #
 # Show the costs for rowid filter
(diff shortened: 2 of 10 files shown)


mysql-test/main/secondary_key_costs.test
```diff
@@ -1,6 +1,7 @@
 --source include/have_sequence.inc
 --source include/not_embedded.inc
 --source include/have_innodb.inc
+--source include/not_valgrind.inc
```
dr-m, line 4: Hereby we are losing Valgrind memcheck coverage of some earlier tests in this file. It would be useful to have a comment here that explains why we are doing it. Is it only because Valgrind is extremely slow compared to `-fsanitize=address` and `-fsanitize=memory` instrumented builds?

mysql-test/main/mysqld--help.result
```diff
@@ -719,9 +719,10 @@ The following specify which files/extra groups are read (specified before remain
  costs. disable_max_seek = Disable 'max_seek optimization'
  for secondary keys and slight adjustment of filter cost.
  disable_forced_index_in_group_by = Disable automatic
- forced index in GROUP BY. This variable will be deleted
- in MariaDB 11.0 as it is not needed with the new 11.0
- optimizer.
+ forced index in GROUP BY. fix_innodb_cardinality =
+ Disable doubling of the Cardinality for InnoDB secondary
+ keys. This variable will be deleted in MariaDB 11.0 as it
+ is not needed with the new 11.0 optimizer.
```
dr-m, line 725: The 11.0 branch is no longer maintained. Should the documentation only mention long-term-support branches, such as 11.4?

sql/sql_class.cc
```diff
@@ -5374,6 +5374,15 @@ extern "C" int thd_current_status(MYSQL_THD thd)
 }
 
 
+extern "C" int thd_double_innodb_cardinality(MYSQL_THD thd)
+{
+  if (thd->variables.optimizer_adjust_secondary_key_costs &
+      OPTIMIZER_FIX_INNODB_CARDINALITY)
+    return 0;                                   // Do not double
+  return 1;                                     // original behaviour
```
dr-m, line 5382: Some compilers might rewrite this already, but I’d just write
```c++
  return !(thd->variables.optimizer_adjust_secondary_key_costs &
           OPTIMIZER_FIX_INNODB_CARDINALITY);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15013,7 +15013,8 @@ ha_innobase::info_low(
 				index selectivity is 2 times better than
 				our estimate: */
 
-				rec_per_key_int = rec_per_key_int / 2;
+                                if (thd_double_innodb_cardinality(m_user_thd))
+					rec_per_key_int = rec_per_key_int / 2;
```
dr-m, line 15017: The first line is being indented with spaces, while the second line is indented by TABs. I would prefer arithmetics to conditional evaluation, to be more friendly towards modern superscalar CPU cores:
```c++
				rec_per_key_int /= 1
					+ thd_double_innodb_cardinality(m_user_thd);
```
That corresponds to the archaic original InnoDB formatting that this function is in.

mysql-test/main/secondary_key_costs.test
spetrunia: Yes, valgrind is very slow on this test.  
Actually not due to this testcase, due to MDEV-33306.

spetrunia: Added a note about this

mysql-test/main/mysqld--help.result
spetrunia: I think the statement "is it is not needed with the new 11.0 optimizer" should stay there  even if 11.0 is EOL-ed.

Then, "This variable will be deleted in MariaDB 11.0 " can stay as well in order to not be confusing.  Changing 11.0 to 11.4 doesn't make the message much better.

PR #3433, opened 2024-07-29
MDEV-34422 InnoDB writes corrupted log on macOS and AIX due to uninitialized log_sys.lsn_lock

DaveGosselin-MariaDB:
- [x] *The Jira issue number for this PR is: MDEV-34422*
## Description
On systems where `log_sys.lsn_lock` is not based on a `futex` like system call but the `pthread_mutex_wrapper`, we must invoke `lsn_lock.init()` to avoid a corrupted `ib_logfile0` due to race conditions on writes to `log_sys.buf`.
## Release Notes
On Apple macOS, IBM AIX, Solaris derivatives and some other platforms, InnoDB would corrupt its `ib_logfile0`, which would not only break crash recovery and `mariadb-backup --backup` but often prevent a normal restart as well.
## How can this PR be tested?
Without this fix, the test suite on our CI would fail very early on Apple macOS as well as IBM AIX, as well as on Microsoft Windows if we unconditionally enabled `SUX_LOCK_GENERIC` (disabled the futex-like system calls).
## PR quality check

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -132,6 +132,7 @@ bool log_t::create()
 #endif
 
   latch.SRW_LOCK_INIT(log_latch_key);
+  lsn_lock.init();
 
   last_checkpoint_lsn= FIRST_LSN;
   log_capacity= 0;
@@ -1330,6 +1331,7 @@ void log_t::close()
 #endif
 
   latch.destroy();
+  lsn_lock.destroy();
 
   recv_sys.close();
 


storage/innobase/include/srw_lock.h
```diff
@@ -39,13 +39,25 @@ template<bool spinloop>
...
 public:
+  pthread_mutex_wrapper()
+  {
+    init();
+  }
+
   void init()
   {
+    if (initialized)
+      return;
```
dr-m, line 53: I would like to try another way of fixing this, that is, adding a missing `init()` call somewhere. It should be a no-op for other mutex implementations. It is better to do things without additional state variables and run-time conditions if possible.

For my experiment ab931d656dde299301e291731e2df89a6c448f9c in https://buildbot.mariadb.org/#/builders/234/builds/31470/steps/8/logs/stdio I got a stack trace that identifies a culprit object:
```
server.dll!pthread_mutex_wrapper<0>::wr_lock()[srw_lock.h:67]
server.dll!log_t::append_prepare<0,0>()[mtr0mtr.cc:1042]
```

dr-m: f7b11442e2f992226ea95026d80f72c9e92821ef is an one-liner fix on top of my instrumentation experiment:
```diff
diff --git a/storage/innobase/log/log0log.cc b/storage/innobase/log/log0log.cc
index a485a875ee1..5396806be82 100644
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -132,6 +132,7 @@ bool log_t::create()
 #endif
 
   latch.SRW_LOCK_INIT(log_latch_key);
+  lsn_lock.init();
 
   last_checkpoint_lsn= FIRST_LSN;
   log_capacity= 0;
```
It was successfully tested on 4 targets where the tests would previously fail:
* Microsoft Windows ([non-debug](https://buildbot.mariadb.org/#/builders/239/builds/22932) and [debug](https://buildbot.mariadb.org/#/builders/234/builds/31489)); the instrumentation caused these to fail as mentioned above.
* [Apple macOS on ARMv8](https://buildbot.mariadb.org/#/builders/590/builds/7443) (in progress)
* [IBM AIX on POWER](https://buildbot.mariadb.org/#/builders/622/builds/3810) (in progress).

I will force-push this pull request with that fix. @DaveGosselin-MariaDB Thank you for your debugging. If we hadn’t been lucky with my instrumentation on Windows, I would have been at a complete loss.

storage/innobase/log/log0log.cc
```diff
@@ -132,6 +132,7 @@ bool log_t::create()
 #endif
 
   latch.SRW_LOCK_INIT(log_latch_key);
+  lsn_lock.init();
```
mariadb-DebarunBanerjee, line 135: This is now corrects the removal of `init_lsn_lock()` in MDEV-33515. 

1. Looking further at the changes done by this MDEV I see it also removed `destroy_lsn_lock()` in `log_t::close` which should be done to destroy the pthread mutex object at the end. So, we should add a call to `lsn_lock.destory` in `log_t::close`.
2. I would suggest to add a debug bool variable `initialized` for both mutex and latch implementation and assert while acquiring. This would detect the problem immediately in debug run. 
3. To allow us to detect such issues in linux, I think we should have a debug way to run different mutex implementations and have some mtr for them.

mariadb-DebarunBanerjee (commented):
Marko,
@dr-m @DaveGosselin-MariaDB 
It is great that finally we could catch and fix the issue. Such issues in low level constructs like mutex and latches are hard to catch and it is usually the last thing one doubts while analyzing and there by could cost lot of developer time. Please see if my suggestions make sense.

storage/innobase/log/log0log.cc
dr-m: 1. You are right, we’d better have also this to avoid potential resource leaks:
```diff
diff --git a/storage/innobase/log/log0log.cc b/storage/innobase/log/log0log.cc
index 6fdb8499d23..93f8db6bfc0 100644
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1331,6 +1331,7 @@ void log_t::close()
 #endif
 
   latch.destroy();
+  lsn_lock.destroy();
 
   recv_sys.close();
 
```
2. This is a good idea, but I think that it should be done separately in the 10.6 branch. This is not entirely trivial. We have a problem that the `sux_lock::readers_lock` in debug builds is being reinitialized. We would need something like this. This is not complete; at least the test `innodb.undo_truncate` would fail because `trx_sys_create_rsegs()` is reinitializing a latch in `trx_rseg_t::init()` after `srv_undo_tablespaces_reinit()` had already done it properly (without a prior `destroy()`):
```diff
diff --git a/storage/innobase/include/srw_lock.h b/storage/innobase/include/srw_lock.h
index f952722426d..fd7e08dc3be 100644
--- a/storage/innobase/include/srw_lock.h
+++ b/storage/innobase/include/srw_lock.h
@@ -30,7 +30,7 @@ this program; if not, write to the Free Software Foundation, Inc.,
 # define SUX_LOCK_GENERIC /* fall back to generic synchronization primitives */
 #endif
 
-#if !defined SUX_LOCK_GENERIC && 0 /* defined SAFE_MUTEX */
+#if !defined SUX_LOCK_GENERIC // && 0 /* defined SAFE_MUTEX */
 # define SUX_LOCK_GENERIC /* Use dummy implementation for debugging purposes */
 #endif
 
@@ -39,15 +39,26 @@ template<bool spinloop>
 class pthread_mutex_wrapper final
 {
   pthread_mutex_t lock;
+#ifdef UNIV_DEBUG
+  bool initialized{false};
 public:
-  void init()
+  ~pthread_mutex_wrapper() { ut_ad(!initialized); }
+#endif
+public:
+  void init(bool maybe_reinit= false)
   {
+    ut_ad(maybe_reinit || !initialized);
+    ut_d(initialized= true);
     if (spinloop)
       pthread_mutex_init(&lock, MY_MUTEX_INIT_FAST);
     else
       pthread_mutex_init(&lock, nullptr);
   }
-  void destroy() { pthread_mutex_destroy(&lock); }
+  void destroy()
+  {
+    ut_ad(initialized); ut_d(initialized=false);
+    pthread_mutex_destroy(&lock);
+  }
 # ifdef PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP
   void wr_lock() { pthread_mutex_lock(&lock); }
 # else
diff --git a/storage/innobase/include/sux_lock.h b/storage/innobase/include/sux_lock.h
index a8f51fdcdba..fe2562ff0ba 100644
--- a/storage/innobase/include/sux_lock.h
+++ b/storage/innobase/include/sux_lock.h
@@ -294,7 +294,7 @@ template<> inline void sux_lock<ssux_lock_impl<true>>::init()
   lock.init();
   ut_ad(!writer.load(std::memory_order_relaxed));
   ut_ad(!recursive);
-  ut_d(readers_lock.init());
+  ut_d(readers_lock.init(true));
 #ifdef UNIV_DEBUG
   if (auto r= readers.load(std::memory_order_relaxed))
     ut_ad(r->empty());
diff --git a/storage/innobase/sync/srw_lock.cc b/storage/innobase/sync/srw_lock.cc
index 6c37d120f73..5db6e627590 100644
--- a/storage/innobase/sync/srw_lock.cc
+++ b/storage/innobase/sync/srw_lock.cc
@@ -661,6 +661,7 @@ void srw_lock_debug::destroy()
     ut_ad(r->empty());
     delete r;
   }
+  readers_lock.destroy();
   srw_lock::destroy();
 }
 
```
3. It is possible to add `-DSUX_LOCK_DEBUG` to `CMAKE_CXX_FLAGS`, or if you do not want to recompile everything, you could patch the start of `srw_lock.h` as shown above.

mariadb-DebarunBanerjee: I didn't mean manual run. It would be useful if we could automate it but it is not necessarily be done part of this patch.

mariadb-DebarunBanerjee (approved):

dr-m:
379c7135be5997a163720efa1a0b3ad1d42e26af is a work-in-progress implementation of the debug instrumentation. It passed the test suite for me locally. I still need to figure out a solution for `buf_page_t::init()` and the `readers_lock` that we use in debug builds. I think that the additional overhead for `dict_index_t::lock` and `trx_rseg_t::latch` should be acceptable. I will file an MDEV and a pull request on this soon.

PR #3434, opened 2024-07-30
MDEV-34669: ER_NEED_REPREPARE on SELECT DEFAULT(name) FROM table1_containing_sequence

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34669*

## Description

A prepared SELECT statement because of CF_REEXECUTION_FRAGILE needs to check the table is the same defination as previously otherwise a re-prepare of the statement can occur.

When running many 'SELECT DEFAULT(name) FROM table1_containing_sequence' in parallel the TABLE_LIST::is_the_same_definition may be called when m_table_ref_type is TABLE_REF_NULL because it hasn't been checked yet.

In this case populate the TABLE_LIST with the values determined by the TABLE_SHARE and allow the execution to continue.

## Release Notes

Prepared statements that check the table defination are less likely to falsely return ER_NEED_REPREPARE

## How can this PR be tested?

Per MDEV.

The race conditions made a mtr test difficult to write.

## PR quality check


LinuxJedi:
Looks like this is causing `ps_ddl` to fail.

PR #3437, opened 2024-07-30
MDEV-32363 Shut down Galera networking and logging on fatal signal

temeo:
- [x] *The Jira issue number for this PR is: MDEV-32363*

## Description

When handling fatal signal, shut down Galera networking before printing out stack trace and writing core file. This is to achieve fail-silent semantics on crashes which may keep the process running for a long time, but not fully responding e.g. due to core dumping or symbol resolving.

Also suppress all Galera/wsrep logging to avoid logging from background threads to garble crash information from signal handler.

Notice that for fully fail-silent crash, Galera 26.4.19 is needed.

## How can this PR be tested?

Deterministic test case does not exist. In order to test, start a 3 node cluster and kill one of the nodes by `kill -SIGABRT`. Observe that the killed node logs 
```
WSREP: Suppressing further logging
WSREP: Shutting down network communications
```
before printing out stacktrace. Other nodes should report lost connection almost immediately.

## PR quality check

### DIFF
--- a/sql/wsrep_server_state.cc
+++ b/sql/wsrep_server_state.cc
@@ -18,6 +18,8 @@
 #include "wsrep_server_state.h"
 #include "wsrep_binlog.h" /* init/deinit group commit */
 
+#include "my_stacktrace.h" /* my_safe_printf_stderr() */
+
 mysql_mutex_t LOCK_wsrep_server_state;
 mysql_cond_t  COND_wsrep_server_state;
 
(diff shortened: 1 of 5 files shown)


sql/wsrep_server_state.cc
```diff
@@ -82,3 +84,24 @@ void Wsrep_server_state::destroy()
...
+
+void Wsrep_server_state::handle_fatal_signal()
+{
+  if (m_instance)
+  {
+    /* Galera background threads are still running and the logging may be
+       relatively verbose in case of networking error. Silence all wsrep
+       logging before shutting down networking to avoid garbling signal
+       handler output. */
+    my_safe_printf_stderr("WSREP: Suppressing further logging\n");
```
dr-m, line 96: Are you really sure that the code of `my_safe_printf_stderr` is safe to invoke in a signal handler function? `man 7 signal-safety` on my GNU/Linux system does not list even `printf` or any of the stream oriented functions. A straight `write` to `STDERR_FILENO` might be safe.

Where is an invocation of this function added? I missed it in this pull request.

janlindstrom: sql/sql_signal_handler.cc first changed file in this pull request. Same function is used in mysys/stacktrace.c.

sysprg:
Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/54a10a429334a9579558a5d284c510d6f8b5bc97

PR #3438, opened 2024-07-31
MDEV-34670 IMPORT TABLESPACE unnecessarily traverses tablespace list

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV- 34670 *

## Description
Problem:
========
- After the commit ada1074bb10359342ee00e220fe9c172574265fb (MDEV-14398) fil_crypt_set_encrypt_tables() iterates through all tablespaces to fill the default_encrypt tables list. This was a trigger to encrypt or decrypt when key rotation age is set to 0. But import tablespace does call fil_crypt_set_encrypt_tables() unnecessarily. The motivation for the call is to signal the encryption threads.

Fix:
====
ha_innobase::discard_or_import_tablespace: Remove the fil_crypt_set_encrypt_tables() and add the import tablespace to the default encrypt list if necessary

## How can this PR be tested?
```

CREATE TABLE t1 (c1 CHAR(1) ,c2 INT) ENGINE=INNODB
                PARTITION BY LINEAR HASH ((c2)) PARTITIONS 8192;
CREATE TABLE t2(f1 INT, f2 INT)ENGINE=InnoDB;
CREATE TABLE t2_imp like t2;
ALTER TABLE t2_imp DISCARD TABLESPACE;
FLUSH TABLE t2 FOR EXPORT;
let $datadir=`select @@datadir`;
--copy_file $datadir/test/t2.ibd $datadir/test/t2_imp.ibd
--copy_file $datadir/test/t2.cfg $datadir/test/t2_imp.cfg
UNLOCK TABLES;
ALTER TABLE t2_imp IMPORT TABLESPACE;
DROP TABLE t2, t2_imp, t1;
```
./mtr encryption.innodb_encryption_discard_import

## PR quality check

### DIFF
--- a/mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
+++ b/mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
@@ -2,7 +2,8 @@
 -- source include/have_example_key_management_plugin.inc
 -- source include/not_valgrind.inc
 -- source include/not_embedded.inc
-
+SET @start_encr_threads = @@global.innodb_encryption_threads;
+SET @start_encrypt_tables = @@global.innodb_encrypt_tables;
 let MYSQLD_DATADIR = `SELECT @@datadir`;
 
 --let SEARCH_RANGE = 10000000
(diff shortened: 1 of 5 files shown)


mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
```diff
@@ -124,3 +125,56 @@ FLUSH TABLES t1, t2, t3 FOR EXPORT;
...
+SET GLOBAL innodb_encryption_threads= 0;
+
+CREATE TABLE t1(f1 int,f2 text)ENGINE=InnoDB;
+INSERT INTO t1 VALUES(1, "InnoDB");
+CREATE TABLE t2 LIKE t1;
+ALTER TABLE t2 DISCARD TABLESPACE;
+FLUSH TABLES t1 FOR EXPORT;
+--copy_file $MYSQLD_DATADIR/test/t1.ibd $MYSQLD_DATADIR/test/t2.ibd
+UNLOCK TABLES;
+ALTER TABLE t2 IMPORT TABLESPACE;
```
dr-m, line 143: We should also rename the `t1.cfg` to `t2.cfg` in order to avoid a warning on `IMPORT`.

mysql-test/suite/encryption/t/innodb_encryption_discard_import.test
```diff
@@ -124,3 +125,56 @@ FLUSH TABLES t1, t2, t3 FOR EXPORT;
...
+AND NAME NOT LIKE 'innodb_undo%' AND NAME NOT LIKE 'mysql/innodb_%_stats' AND NAME NOT LIKE 'mysql/transaction_registry';
+
+--sorted_result
+SELECT NAME FROM INFORMATION_SCHEMA.INNODB_TABLESPACES_ENCRYPTION WHERE MIN_KEY_VERSION <> 0
+AND NAME NOT LIKE 'innodb_undo%' AND NAME NOT LIKE 'mysql/innodb_%_stats' AND NAME NOT LIKE 'mysql/transaction_registry';
+
+SET GLOBAL innodb_encryption_threads=@start_encr_threads;
+SET GLOBAL innodb_encrypt_tables=@start_encrypt_tables;
+SELECT * FROM t2;
+DROP TABLE t1, t2;
```
dr-m, line 180: What is the purpose of the `SELECT`? Because the server was not restarted at any point, the data should remain cached in the buffer pool all the time. If we want to demonstrate that the contents of the data file is readable, then we perhaps should access the `t2.ibd` file directly by some means, such as invoking `innochecksum`, between `FLUSH TABLE t2 FOR EXPORT` and `UNLOCK TABLES`.

storage/innobase/fil/fil0crypt.cc
```diff
@@ -2364,6 +2364,24 @@ fil_crypt_set_rotation_iops(
...
 }
 
+/** Add the import tablespace to default_encrypt list
+if necessary and signal fil_crypt_threads
+@param space imported tablespace */
+void fil_crypt_add_imported_space(fil_space_t *space)
+{
+  mutex_enter(&fil_system.mutex);
+
+  if (fil_crypt_must_default_encrypt())
```
dr-m, line 2374: It would be good to assert in `fil_crypt_must_default_encrypt()` that `fil_system.mutex` is being held, with a comment saying that it prevents a race condition with `fil_crypt_set_rotate_key_age()`.

dr-m (approved):
Thank you, this looks OK with a few minor nitpicks.

PR #3439, opened 2024-07-31
MDEV-34678 pthread_mutex_init() without pthread_mutex_destroy()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34678*
## Description
When `SUX_LOCK_GENERIC` is defined, the `srw_mutex`, `srw_lock`, `sux_lock` are implemented based on `pthread_mutex_t` and `pthread_cond_t`.  This is the only option for systems that lack a `futex`-like system call.

In the `SUX_LOCK_GENERIC` mode, if `pthread_mutex_init()` is allocating some resources that need to be freed by `pthread_mutex_destroy()`, a memory leak could occur when we are repeatedly invoking `pthread_mutex_init()` without a `pthread_mutex_destroy()` in between.

`pthread_mutex_wrapper::initialized`: A debug field to track whether `pthread_mutex_init()` has been invoked.  This also helps find bugs like the one that was fixed by commit 1c8af2ae53a3a48d64b4167d21b20ff5e1caef36 (MDEV-34422); one simply needs to add `-DSUX_LOCK_GENERIC` to the CMAKE_CXX_FLAGS to catch that particular bug on the initial server bootstrap.

`buf_block_init()`, `buf_page_init_for_read()`: Invoke `block_lock::init()` because `buf_page_t::init()` will no longer do that.

`buf_page_t::init()`: Instead of invoking `lock.init()`, assert that it has already been invoked (the lock is vacant).

`add_fts_index()`, `build_fts_hidden_table()`: Explicitly invoke `index_lock::init()` in order to avoid a `pthread_mutex_destroy()` invocation on an uninitialized object.

`srw_lock_debug::destroy()`: Invoke `readers_lock.destroy()`.

`trx_sys_t::create()`: Invoke `trx_rseg_t::init()` on all rollback segments in order to guarantee a deterministic state for shutdown, even if InnoDB fails to start up.

`trx_rseg_array_init()`, `trx_temp_rseg_create()`, `trx_rseg_create()`: Invoke `trx_rseg_t::destroy()` before `trx_rseg_t::init()` in order to balance `pthread_mutex_init()` and `pthread_mutex_destroy()` calls.
## Release Notes
On operating systems that lack a `futex`-like system call (such as macOS, AIX, Solaris), memory could be leaked as a result of repeated calls to `pthread_mutex_init()` without a call to `pthread_mutex_destroy()` in between.
## How can this PR be tested?
If the system does support a `futex` like system call (Linux, Windows, FreeBSD, OpenBSD, NetBSD, Dragonfly BSD), add `-DSUX_LOCK_GENERIC` to the `CMAKE_CXX_FLAGS` so that this fallback code will be used.

The default `./mtr` regression test suites should pass. I tested it on Debian GNU/Linux with both `-DPLUGIN_PERFSCHEMA=NO` and  `-DPLUGIN_PERFSCHEMA=STATIC`.

I tested this also on 10.11 cc8eefb0dca1372378905fbae11044f20364c42d (right before the #3433 fix) and got an assertion failure on server bootstrap, as expected, for both "fake PMEM" on `/dev/shm` and on the `pwrite` based log writes:
```
mariadbd: /mariadb/10.11/storage/innobase/include/srw_lock.h:64: void pthread_mutex_wrapper<spinloop>::wr_lock() [with bool spinloop = false]: Assertion `initialized' failed.
```
## PR quality check

### DIFF
--- a/storage/innobase/include/buf0buf.h
+++ b/storage/innobase/include/buf0buf.h
@@ -648,10 +648,10 @@ public:
   void init(uint32_t state, page_id_t id)
   {
     ut_ad(state < REMOVE_HASH || state >= UNFIXED);
+    ut_ad(!lock.is_locked_or_waiting());
     id_= id;
     zip.fix= state;
     oldest_modification_= 0;
-    lock.init();
     ut_d(in_zip_hash= false);
     ut_d(in_free_list= false);
     ut_d(in_LRU_list= false);
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -842,6 +842,7 @@ buf_block_init(buf_block_t* block, byte* frame)
 	MEM_MAKE_DEFINED(&block->modify_clock, sizeof block->modify_clock);
 	ut_ad(!block->modify_clock);
 	MEM_MAKE_DEFINED(&block->page.lock, sizeof block->page.lock);
+	block->page.lock.init();
 	block->page.init(buf_page_t::NOT_USED, page_id_t(~0ULL));
 #ifdef BTR_CUR_HASH_ADAPT
 	MEM_MAKE_DEFINED(&block->index, sizeof block->index);
--- a/storage/innobase/buf/buf0rea.cc
+++ b/storage/innobase/buf/buf0rea.cc
@@ -212,6 +212,7 @@ static buf_page_t* buf_page_init_for_read(ulint mode, const page_id_t page_id,
     page_zip_set_size(&bpage->zip, zip_size);
     bpage->zip.data = (page_zip_t*) data;
 
+    bpage->lock.init();
     bpage->init(buf_page_t::READ_FIX, page_id);
     bpage->lock.x_lock(true);
 
--- a/storage/innobase/include/srw_lock.h
+++ b/storage/innobase/include/srw_lock.h
@@ -40,32 +40,45 @@ template<bool spinloop>
 class pthread_mutex_wrapper final
 {
   pthread_mutex_t lock;
+#ifdef UNIV_DEBUG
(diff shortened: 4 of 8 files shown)


storage/innobase/include/buf0buf.h
```diff
@@ -648,10 +648,10 @@ class buf_page_t
   void init(uint32_t state, page_id_t id)
   {
     ut_ad(state < REMOVE_HASH || state >= UNFIXED);
+    ut_ad(!lock.is_locked_or_waiting());
     id_= id;
     zip.fix= state;
     oldest_modification_= 0;
-    lock.init();
```
mariadb-DebarunBanerjee, line 654: I see this lock.init() call is moved to the caller. Can we assert here that the lock is initialized ?

mariadb-DebarunBanerjee (approved):
Thank you @dr-m for pursuing this work. I see that this allowed us to find the init() and destroy() calls.

storage/innobase/include/buf0buf.h
dr-m: The added assertion checks a part of the effect of `lock.init()`. It will not cover the debug-only field `readers_lock` or `readers` for keeping track of threads that are holding S latches, nor the `pthread_mutex_t` or `pthread_cond_t` that would be embedded in `ssux_lock_impl` when `SUX_LOCK_GENERIC` is defined (`futex`-like system calls are not available).

However, other checks about the initialization or destruction of the debug field `readers_lock` would easily fire when `SUX_LOCK_GENERIC` is defined and the initialization was forgotten. The reason why the `lock.init()` call was moved out of here was to ensure that all `init()` and `destroy()` calls are paired.

PR #3442, opened 2024-08-01
MDEV-29010	Table cannot be loaded after instant ALTER

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-29010*
## Description
Reason:
======
- InnoDB fails to load the instant alter table metadata from clustered index while loading the table definition. The reason is that InnoDB metadata blob has the column length exceeds maximum fixed length column size.

Fix:
===
- InnoDB should treat the long fixed length column as variable length fields that needs external storage while initializing the field map for instant alter operation

## How can this PR be tested?
./mtr innodb.instant_alter_bugs

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/instant_alter_bugs.test
+++ b/mysql-test/suite/innodb/t/instant_alter_bugs.test
@@ -553,3 +553,14 @@ SELECT COUNT(*) FROM t1;
 DROP TABLE t1;
 
 --echo # End of 10.4 tests
+
+--echo #
+--echo #  MDEV-29010  Table cannot be loaded after instant ALTER
+--echo #
+CREATE TABLE t1 (a CHAR(255), b INT,
+		 c INT as (b) VIRTUAL)ENGINE=InnoDB CHARACTER SET utf32;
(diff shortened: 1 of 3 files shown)


mysql-test/suite/innodb/t/instant_alter_bugs.test
```diff
@@ -553,3 +553,12 @@ SELECT COUNT(*) FROM t1;
...
 --echo # End of 10.4 tests
+
+--echo #
+--echo #  MDEV-29010  Table cannot be loaded after instant ALTER
+--echo #
+CREATE TABLE t1 (a CHAR(255), b INT,
+		 c INT as (b) VIRTUAL)ENGINE=InnoDB CHARACTER SET utf32;
+ALTER TABLE t1 DROP COLUMN a;
+ALTER TABLE t1 DROP COLUMN c;
+DROP TABLE t1;
```
dr-m, line 564: The following needs to be added to the end:
```sql
--echo # End of 10.5 tests
```

dr-m (approved):

PR #3443, opened 2024-08-01
MDEV-34665 Simplify IN predicate processing for NULL-aware materializ…

Olernov:
…ation involving only one column

- [x] *The Jira issue number for this PR is: MDEV-34665

It was found that unnecessary work of building Ordered_key structures is being done when processing NULL-aware materialization for IN predicates having only one column. In fact, the logic for that simplified case can be expressed as follows.
Say we have predicate `left_expr IN (SELECT <subq1>)`, where `left_expr` is scalar(not a tuple). Then
   ```
 if (left_expr is NULL) {
      if (subq1 produced any rows) {
        // note that we don't care if subq1 has produced
        // NULLs or not.
        NULL IN (<some values>) -> UNKNOWN, i.e. NULL.
      } else {
        NULL IN ({empty-set}) -> FALSE.
      }
    } else {
      // left_expr is a non-NULL value
      if (subq1 output has a match for left_expr) {
        left_expr IN (..., left_expr ...) -> TRUE
      } else {
        // no "known" matches.
        if (subq1 output has a NULL) {
          left_expr IN ( ... NULL ...) ->
           (NULL could have been a match or not)
           -> NULL.
        } else {
          // subq1 didn't produce any "UNKNOWNs" so
          // we're positive there weren't any matches
          -> FALSE.
        }
      }
    }
```

This commit introduces subselect_single_column_partial_engine class implementing the logic described.
## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/main/subselect_mat.test
+++ b/mysql-test/main/subselect_mat.test
@@ -296,3 +296,76 @@ UNION ALL
 ;
 
 drop table t1, t2;
+
+--echo #
+--echo # MDEV-34665: Simplify IN predicate processing for NULL-aware
+--echo #             materialization involving only one column
+
+set @save_optimizer_switch=@@optimizer_switch;
+set @@optimizer_switch = "materialization=on,in_to_exists=off,semijoin=off";
+
+create table t1 (a int);
+create table t2 (b int);
+insert into t1 values (null), (1), (2), (3);
+
+--echo # Query against empty t2
+--sorted_result
+select a, a in (select b from t2) from t1;
+
+--echo # Insert some not-NULL values
+insert into t2 values (3), (4);
+--sorted_result
+select a, a in (select b from t2) from t1;
+--echo # Ensure the correct strategy is tested
+--source include/analyze-format.inc
+analyze format=json select a, a in (select b from t2) from t1;
+
+--echo # Insert NULL value (so there are both NULLs and and not-NULL values)
+insert into t2 values (null);
+--sorted_result
+select a, a in (select b from t2) from t1;
+--source include/analyze-format.inc
+analyze format=json select a, a in (select b from t2) from t1;
+
+delete from t2;
+--echo # Insert NULL values only
+insert into t2 values (null), (null);
+--sorted_result
+select a, a in (select b from t2) from t1;
+--source include/analyze-format.inc
+analyze format=json select a, a in (select b from t2) from t1;
+
+--echo # Test UPDATE
+insert into t2 values (3), (4);
+update t1 set a=a+1 where a not in (select b from t2);
+--echo # Nothing updated due to NULLs on both sides of IN
+select * from t1;
+--echo # Remove NULLs from the right side
+delete from t2 where b is null;
+update t1 set a=a+1 where a not in (select b from t2);
+--echo # Now some rows are updated:
+select * from t1;
+--source include/analyze-format.inc
+analyze format=json update t1 set a=a+1 where a not in (select b from t2);
+
+--echo # Test DELETE
+--echo # Restore initial data-set:
+delete from t1;
+insert into t1 values (null), (1), (2), (3);
+--echo # Add some NULL values to the right side of IN
+insert into t2 values (null), (null);
+delete from t1 where a not in (select b from t2);
+--echo # Nothing deleted due to NULLs on both sides of IN
+select * from t1;
+--echo # Remove NULLs from the right side
+delete from t2 where b is null;
+delete from t1 where a not in (select b from t2);
+--echo # Now some rows are deleted:
+select * from t1;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: sql/item_subselect.h [hunk 1]
```diff
@@ -1175,6 +1176,8 @@ class subselect_hash_sj_engine : public subselect_engine
     PARTIAL_MATCH,  /* Use some partial matching strategy. */
     PARTIAL_MATCH_MERGE, /* Use partial matching through index merging. */
     PARTIAL_MATCH_SCAN,  /* Use partial matching through table scan. */
+    PARTIAL_MATCH_SINGLE_COLUMN, /* Use simplifeid partial matching when
```
spetrunia, line 1179: Typo, `simplifeid`

sql/item_subselect.h
```diff
@@ -1559,6 +1563,25 @@ class subselect_table_scan_engine: public subselect_partial_match_engine
   enum_engine_type engine_type() override { return TABLE_SCAN_ENGINE; }
 };
 
+
+class subselect_single_column_partial_engine:
```
spetrunia, line 1567: This needs a comment.  
Something like
```
/*
  An engine to handle NULL-aware Materialization for subqueries
  that compare one column:

    col1 IN (SELECT t2.col2 FROM t2 ...)

  When only one column is used, we need to handle NULL values of 
  col1 and col2 but don't need to perform "partial" matches when only
  a subset of compared columns is NULL.
  This allows to save on some data structures.
*/
```

sql/item_subselect.h
```diff
@@ -1641,6 +1664,8 @@ class Subq_materialization_tracker
         return "index_lookup;array merge for partial match";
       case Strategy::PARTIAL_MATCH_SCAN:
         return "index_lookup;full scan for partial match";
+      case Strategy::PARTIAL_MATCH_SINGLE_COLUMN:
+        return "index_lookup;constant for partial match";
```
spetrunia, line 1668: This sounds very cryptic for user.  Internally, we handle this through `subselect_partial_match_engine` but 
imagine a user running
``` 
 col1 in (select col2 from t2)
```
and getting "partial match"?   Please consider changing to 
`check nulls; index_lookup`  or `null-aware index_lookup`  or something like that.   (It should be still different from just `index_lookup` to allow us developers to distinguish from COMPLETE_MATCH).

sql/item_subselect.cc
```diff
@@ -6978,6 +6999,44 @@ void subselect_table_scan_engine::cleanup()
...
+    - there is only one column in the materialized table;
+    - its current value of left_expr is NULL (otherwise we would have hit
+         the earlier "index lookup" branch at subselect_partial_match::exec());
+    - the materialized table does not have NULL values (for a similar reason);
+    - the materialized table is not empty.
+    The case when materialization produced no rows (empty table) is handled at
+    subselect_hash_sj_engine::exec(), the result of IN predicate is always
+    FALSE in that case.
+    After all those preconditions met, the result of the partial match is TRUE.
+  */
```
spetrunia, line 7032: Very nice comment, thanks!

mysql-test/main/subselect_mat.test
```diff
@@ -296,3 +296,50 @@ UNION ALL
...
+insert into t1 values (null), (1), (2), (3);
+
+--echo # Query against empty t2
+--sorted_result
+select * from t1 where a not in (select b from t2);
+--sorted_result
+select a from t1 join t2 on a not in (select b from t2);
+--sorted_result
+select a from t1 left join t2 on a not in (select b from t2);
+
```
spetrunia, line 318: As discussed on Slack: the 3 variants are excessive. 
It's better to put the subquery into the select list:
```
select a,  a in (select ....) as SUBQ ,,,
```
This puts the subquery into a context where the subquery's arguments and return value can be easily seen.   (and the subquery is not a "top-level item" so optimizer doesn't take shortcuts).   Using a join outside the subquery is redundant.

sql/item_subselect.h
spetrunia: ... The same with `subselect_single_column_partial_engine`.  This looks puzzling. Maybe, "subselect_single_column_match_engine" ?

Olernov: Good point, I like the names you suggested. I have also renamed the strategy to `Strategy::SINGLE_COLUMN_MATCH` in a next revision.

spetrunia (changes requested):
Ok this patch is making progress.

But I still see duplicate queries in subselect_mat.test:
```
select * from t1 where a not in (select b from t2);
--sorted_result
select a, a in (select b from t2) from t1;
```
Please fix this.

spetrunia (approved):

PR #3446, opened 2024-08-02
MDEV-34689 Redo log corruption at high load

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-34689*

## Description
Issue: During mtr_t:commit, if there is not enough space available in redo log buffer, we flush the buffer. During flush, the log system latch is released allowing other concurrent mtr to commit. After flush we reacquire the latch but use the old LSN obtained before check. It could lead to redo log corruption. As the LSN moves backwards with the possibility of data loss and unrecoverable server if the server aborts for any reason or if server is shutdown with innodb_fast_shutdown=2. With normal shutdown, recovery fails to map the checkpoint LSN to correct offset.

In debug mode it hits log0log.cc:863: lsn_t log_t::write_buf() Assertion `new_buf_free == ((lsn - first_lsn) & write_size_1)' failed.

In release mode, after normal shutdown, restart fails. [ERROR] InnoDB: Missing FILE_CHECKPOINT(8416546) at 8416546 [ERROR] InnoDB: Log scan aborted at LSN 8416546

Backup fails reading the corrupt redo log.
[00] 2024-07-31 20:59:10 Retrying read of log at LSN=7334851 [00] FATAL ERROR: 2024-07-31 20:59:11 Was only able to copy log from 7334851 to 7334851, not 8416446; try increasing innodb_log_file_size

Unless a backup is tried or the server is shutdown or killed immediately, the corrupt redo part is eventually truncated and there may not be any visible issues seen in release mode.

This issue was introduced by the following commit.

commit a635c40648519fd6c3729c9657872a16a0a20821
    MDEV-27774 Reduce scalability bottlenecks in mtr_t::commit()

Fix: If we need to release latch and flush redo before writing mtr logs, make sure to get the latest system LSN after reacquiring the redo system latch.

## Release Notes
We could mention the error scenarios that could occur because of the issue.
1. Server startup failure:
[ERROR] InnoDB: Missing FILE_CHECKPOINT

2. Backup failure: 
[00] 2024-07-31 20:59:10 Retrying read of log at LSN=7334851
[00] FATAL ERROR: 2024-07-31 20:59:11 Was only able to copy log from 7334851 to 7334851, not 8416446; try increasing innodb_log_file_size

## How can this PR be tested?
The fix can be tested by manually running the testcase provided in MDEV-34689. With high concurrency and low redo buffer size the issue might be hit.

It is hard to add a automated test for the issue as the redo log generation depends on the background operations and it could be hard to maintain too a the amount of log generated by mtrs could change in future.

## PR quality check

### DIFF
--- a/storage/innobase/mtr/mtr0mtr.cc
+++ b/storage/innobase/mtr/mtr0mtr.cc
@@ -1042,13 +1042,19 @@ std::pair<lsn_t,byte*> log_t::append_prepare(size_t size, bool ex) noexcept
   size_t b{spin ? lock_lsn() : buf_free.load(std::memory_order_relaxed)};
   write_to_buf++;
 
-  const lsn_t l{lsn.load(std::memory_order_relaxed)}, end_lsn{l + size};
+  lsn_t l{lsn.load(std::memory_order_relaxed)}, end_lsn{l + size};
 
   if (UNIV_UNLIKELY(pmem
                     ? (end_lsn -
                        get_flushed_lsn(std::memory_order_relaxed)) > capacity()
                     : b + size >= buf_size))
+  {
     b= append_prepare_wait<spin>(b, ex, l);
+    /* While flushing log, we had released the lsn lock and LSN could have
+    progressed in the meantime. */
+    l= lsn.load(std::memory_order_relaxed);
+    end_lsn= l + size;
+  }
 
   size_t new_buf_free= b + size;
   if (pmem && new_buf_free >= file_size)


storage/innobase/mtr/mtr0mtr.cc
```diff
@@ -1042,13 +1042,19 @@ std::pair<lsn_t,byte*> log_t::append_prepare(size_t size, bool ex) noexcept
...
   if (UNIV_UNLIKELY(pmem
                     ? (end_lsn -
                        get_flushed_lsn(std::memory_order_relaxed)) > capacity()
                     : b + size >= buf_size))
+  {
     b= append_prepare_wait<spin>(b, ex, l);
+    /* While flushing log, we had released the log system latch and LSN could
+    have progressed in the meantime. */
+    l = lsn.load(std::memory_order_relaxed);
+    end_lsn = l + size;
```
dr-m, line 1056: I think that the comment is slightly misleading, speaking about "the log system latch" while there are multiple latches in `log_sys`. An intuitive interpretation of the term would be `log_sys.latch`, but that is wrong. (Holding an exclusive `log_sys.latch` would freeze `log_sys.lsn` but most writers are holding a shared `log_sys.latch`.) What is relevant is `lsn_lock` or `lock_lsn()`. They protect the allocation of LSN. So, the comment should be reworded.

Furthermore, the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/11.6/CODING_STANDARDS.md) linked from the pull request description says that there should be no space before an assignment operator. I would suggest the following:
```c++
    /* The LSN could change while we were not holding lsn_lock or lock_lsn() */
    l= lsn.load(std::memory_order_relaxed);
    end_lsn= l + size;
```

dr-m (approved):
Thank you; good catch! This could explain a few sporadic recovery failures that we have observed both on https://buildbot.mariadb.org and on our local systems (typically, when avoiding the use of `/dev/shm` on Linux.)

This code path is easy to cover on RAM disk on Linux by running the following in the `mysql-test` subdirectory of the build directory:
```sh
mkdir /run/user/$UID/var
ln -sf /run/user/$UID/var var
```
Some recent GNU/Linux distributions also create `/tmp` on RAM disk.

storage/innobase/mtr/mtr0mtr.cc
mariadb-DebarunBanerjee: right, it is the lsn_lock. Thanks for pointing it out. And I keep forgetting about the assignment rule :)

PR #3450, opened 2024-08-04
MDEV-33756: Deprecate binlog_optimize_thread_scheduling

bnestere:
The option binlog_optimize_thread_scheduling was initially added to provide a safe alternative for the newly added binlog group commit logic, such that when 0, it would disable a leader thread from performing the binlog write for all transactions that are a part of the group commit. Any problems related to the binlog group commit optimization should be sorted out by now, so we can deprecate-to-eventually-remove the option altogether.

This commit performs the deprecation, and the removal is tracked by MDEV-33745. Note, as the option is only able to be provided via configuration at startup time, users will not see a deprecation message unless looking through the CLI help message.


knielsen (approved):
Sure, that's fine.
I doubt anyone is actually using this option anyway.

vuvova (approved):

bnestere:
Pushed as 9811d23b6d05c89e03ebbdf4202bbbe123e02ab3

PR #3451, opened 2024-08-06
MDEV-34707: BUF_GET_RECOVER assertion failure on upgrade

dr-m:
- [x] *The Jira issue number for this PR is: [MDEV-34707](https://jira.mariadb.org/browse/MDEV-34707)*
## Description
`buf_page_get_gen()`: Relax the assertion once more. The LSN may grow by invoking `ibuf_upgrade()`, that is, when upgrading files where `innodb_change_buffering≠none` was used. The LSN may also have been recovered from a log that needs to be upgraded to the current format.
## Release Notes
N/A. This was a false alarm in a debug assertion whose intention is to catch the misuse of an internal API.
## How can this PR be tested?
1. Download the attachment from [MDEV-34707](https://jira.mariadb.org/browse/MDEV-34707)
2. `7z x dt.7z`
2. (Optional:) Start up and shut down MariaDB Server 10.5 or 10.6 with `--datadir` pointing to the `dt` subdirectory that was extracted in the previous step.
3. Start up MariaDB Server 11.x with `--datadir` pointing to the `dt `subdirectory.
The earliest applicable branch is 11.1, but I think that this will be pushed after the pending quarterly releases, at which point the 11.2 branch would be the last maintained one.
## PR quality check

### DIFF
--- a/storage/innobase/srv/srv0start.cc
+++ b/storage/innobase/srv/srv0start.cc
@@ -1163,8 +1163,11 @@ static dberr_t srv_log_rebuild_if_needed()
   return srv_log_rebuild();
 }
 
+ut_d(bool ibuf_upgrade_was_needed;)
+
 ATTRIBUTE_COLD static dberr_t ibuf_log_rebuild_if_needed()
 {
+  ut_d(ibuf_upgrade_was_needed= true;)
   mysql_mutex_lock(&recv_sys.mutex);
   recv_sys.apply(true);
   mysql_mutex_unlock(&recv_sys.mutex);
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2606,8 +2606,11 @@ buf_page_get_gen(
 	would typically be around 60000 with the default
 	innodb_undo_tablespaces=3, and less than 110000 with the maximum
 	innodb_undo_tablespaces=127. */
+	ut_d(extern bool ibuf_upgrade_was_needed;)
 	ut_ad(mode == BUF_GET_RECOVER
 	      ? recv_recovery_is_on() || log_sys.get_lsn() < 120000
+	      || log_sys.get_lsn() == recv_sys.lsn + SIZE_OF_FILE_CHECKPOINT
+	      || ibuf_upgrade_was_needed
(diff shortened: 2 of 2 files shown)


storage/innobase/srv/srv0start.cc
```diff
@@ -1182,6 +1185,10 @@ static tpool::task rollback_all_recovered_task(trx_rollback_all_recovered,
 					       nullptr,
 					       &rollback_all_recovered_group);
 
+#ifdef UNIV_DEBUG
+bool must_upgrade_ibuf;
+#endif
+
```
mariadb-DebarunBanerjee, line 1191: I could not understand this change. There is no reference to this global. There is one local variable (non-debug) exists by the same name and used. Something to remove ?

mariadb-DebarunBanerjee (approved):
Thanks for fixing the debug specific issue. Please go ahead.

We might want to eliminate the assert for checking LSN altogether. I checked the reasoning but It doesn't seem to be good validating against some constant value. Also, checking for LSN in buf_page_get_gen looks a bit misplaced.

storage/innobase/srv/srv0start.cc
dr-m: Right, this should not have been added. I added `ibuf_upgrade_was_needed` instead.

dr-m:
> Thanks for fixing the debug specific issue. Please go ahead.
> 
> We might want to eliminate the assert for checking LSN altogether. I checked the reasoning but It doesn't seem to be good validating against some constant value. Also, checking for LSN in buf_page_get_gen looks a bit misplaced.

The purpose of that assertion is to ensure that `mode == BUF_GET_RECOVER` and `recv_sys.recover()` are not being misused. Very few page read operations may be invoked while the log-based recovery is in progress. Similarly, we want to ensure that any page reads during recovery will use `mode == BUF_GET_RECOVER`. The reason for this special logic is that 850d61736deba354377634cf344256ee850b93b7 simplified some code that had been added in 6697135c6d03935118c3dfa1c97faea7fa76afa6 to fix a recovery bug. Sometimes when `recv_sys.recover()` is invoked, recovery may not actually be running. The LSN assertions are related to that.

PR #3454, opened 2024-08-12
MDEV-20153: Slave error message incorrectly mentions server_uuid

bnestere:
Correct an error message thrown by the slave to remove the descriptor "slave_uuid", as it was never ported to MariaDB


knielsen (approved):
Sure.

I don't have a strong opinion on whether to do it in 10.5, or only in a later version to be careful with changes in GA versions. It's good to be more careful about GA changes, and the bug is of course very minor. On the other hand, the risk of breaking things is very low as there's an error code to match against, which is the one that can be assumed stable.

If you want my opinion, I'd suggest some 11.x, just because it's so minor, but up to you.

andrelkin (approved):

bnestere:
Pushed into 11.7 as a812dba6dc881a8a7fbd9bcc1451d2e54c92575c

PR #3458, opened 2024-08-13
MDEV-34750 SET GLOBAL innodb_log_file_size is not crash safe

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34750*
## Description
The recent commit 4ca355d863e3b6a14439eebbb2958afccb3548e3 (MDEV-33894) caused a serious regression for online InnoDB `ib_logfile0` resizing, breaking crash-safety unless the memory-mapped log file interface is being used. However, the log resizing was broken also before this.
## Release Notes
If InnoDB was killed soon after executing `SET GLOBAL innodb_log_file_size`, it could fail to recover.
## How can this PR be tested?
To prevent such regressions in the future, we will extend the test `innodb.log_file_size_online` with a kill and restart of the server and with some writes running concurrently with the log size change.

I tested this on a 4-thread system with the following invocation, not using memory-mapped log (`/dev/shm` or `WITH_INNODB_PMEM`): 
```bash
./mtr --parallel=auto --repeat=10 innodb.log_file_size_online{,}
```
Memory-mapped log files were not affected by this, but by #3355.
## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/log_file_size_online.test
+++ b/mysql-test/suite/innodb/t/log_file_size_online.test
@@ -25,17 +25,28 @@ SHOW VARIABLES LIKE 'innodb_log_file_size';
 let SEARCH_PATTERN = InnoDB: Resized log to 4\\.000MiB;
 --source include/search_pattern_in_file.inc
 
-UPDATE t SET b='' WHERE a<10;
+send SET GLOBAL innodb_log_file_size=5242880;
 
-SET GLOBAL innodb_log_file_size=5242880;
+--connect con1,localhost,root
+send UPDATE t SET b='' WHERE a<10;
+
+--connection default
+reap;
 SHOW VARIABLES LIKE 'innodb_log_file_size';
 SELECT global_value FROM information_schema.system_variables
 WHERE variable_name = 'innodb_log_file_size';
 
+--connection con1
+reap;
+--disconnect con1
+--connection default
+
+--let $shutdown_timeout=0
 --let $restart_parameters=
 --source include/restart_mysqld.inc
 
 SELECT * FROM t WHERE a<10;
+SELECT COUNT(*),LENGTH(b) FROM t GROUP BY b;
 
 SHOW VARIABLES LIKE 'innodb_log_file_size';
 let SEARCH_PATTERN = InnoDB: Resized log to 5\\.000MiB;
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -18551,6 +18551,7 @@ static void innodb_log_file_size_update(THD *thd, st_mysql_sys_var*,
       ib_senderrf(thd, IB_LOG_LEVEL_ERROR, ER_CANT_CREATE_HANDLER_FILE);
       break;
     case log_t::RESIZE_STARTED:
+      const lsn_t start{log_sys.resize_in_progress()};
       for (timespec abstime;;)
       {
         if (thd_kill_level(thd))
@@ -18561,13 +18562,30 @@ static void innodb_log_file_size_update(THD *thd, st_mysql_sys_var*,
 
         set_timespec(abstime, 5);
         mysql_mutex_lock(&buf_pool.flush_list_mutex);
-        const bool in_progress(buf_pool.get_oldest_modification(LSN_MAX) <
-                               log_sys.resize_in_progress());
-        if (in_progress)
+        lsn_t resizing= log_sys.resize_in_progress();
+        if (resizing > buf_pool.get_oldest_modification(0))
+        {
+          buf_pool.page_cleaner_wakeup(true);
           my_cond_timedwait(&buf_pool.done_flush_list,
                             &buf_pool.flush_list_mutex.m_mutex, &abstime);
+          resizing= log_sys.resize_in_progress();
+        }
         mysql_mutex_unlock(&buf_pool.flush_list_mutex);
-        if (!log_sys.resize_in_progress())
+        if (start > log_sys.get_lsn())
+        {
+          ut_ad(!log_sys.is_pmem());
+          /* The server is almost idle. Write dummy FILE_CHECKPOINT records
+          to ensure that the log resizing will complete. */
+          log_sys.latch.wr_lock(SRW_LOCK_CALL);
+          while (start > log_sys.get_lsn())
+          {
+            mtr_t mtr;
+            mtr.start();
+            mtr.commit_files(log_sys.last_checkpoint_lsn);
+          }
+          log_sys.latch.wr_unlock();
+        }
+        if (!resizing || resizing > start /* only wait for our resize */)
           break;
       }
     }
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -516,17 +516,14 @@ log_t::resize_start_status log_t::resize_start(os_offset_t size) noexcept
         resize_target= size;
         resize_buf= static_cast<byte*>(ptr);
         resize_flush_buf= static_cast<byte*>(ptr2);
+        start_lsn= get_lsn();
+
         if (is_pmem())
-        {
           resize_log.close();
-          start_lsn= get_lsn();
-        }
         else
-        {
-          memcpy_aligned<16>(resize_buf, buf, (buf_free + 15) & ~15);
           start_lsn= first_lsn +
-            (~lsn_t{write_size - 1} & (write_lsn - first_lsn));
-        }
+            (~lsn_t{write_size - 1} &
+             (lsn_t{write_size - 1} + start_lsn - first_lsn));
       }
       resize_lsn.store(start_lsn, std::memory_order_relaxed);
       status= success ? RESIZE_STARTED : RESIZE_FAILED;
@@ -780,19 +777,26 @@ inline void log_t::persist(lsn_t lsn) noexcept
 #endif
 
 ATTRIBUTE_COLD ATTRIBUTE_NOINLINE
-/** Write resize_buf to resize_log.
-@param length  the used length of resize_buf */
-void log_t::resize_write_buf(size_t length) noexcept
+void log_t::resize_write_buf(const byte *b, size_t length) noexcept
 {
   const size_t block_size_1= write_size - 1;
+  ut_ad(b == resize_buf || b == resize_flush_buf);
   ut_ad(!(resize_target & block_size_1));
   ut_ad(!(length & block_size_1));
   ut_ad(length > block_size_1);
   ut_ad(length <= resize_target);
-  const lsn_t resizing{resize_in_progress()};
-  ut_ad(resizing <= write_lsn);
-  lsn_t offset= START_OFFSET +
-    ((write_lsn - resizing) & ~lsn_t{block_size_1}) %
+
+  int64_t d= int64_t(write_lsn - resize_in_progress());
+  if (UNIV_UNLIKELY(d <= 0))
+  {
+    d&= ~int64_t(block_size_1);
+    if (int64_t(d + length) <= 0)
+      return;
+    length+= d;
+    b-= d;
+    d= 0;
+  }
+  lsn_t offset= START_OFFSET + (lsn_t(d) & ~lsn_t{block_size_1}) %
     (resize_target - START_OFFSET);
 
   if (UNIV_UNLIKELY(offset + length > resize_target))
@@ -804,7 +808,7 @@ void log_t::resize_write_buf(size_t length) noexcept
   }
 
   ut_a(os_file_write_func(IORequestWrite, "ib_logfile101", resize_log.m_file,
-                          buf, offset, length) == DB_SUCCESS);
+                          b, offset, length) == DB_SUCCESS);
 }
 
 /** Write buf to ib_logfile0.
@@ -838,6 +842,7 @@ template<bool release_latch> inline lsn_t log_t::write_buf() noexcept
     ut_ad(write_size_1 >= 511);
 
     const byte *const write_buf{buf};
+    const byte *const re_write_buf{resize_buf};
     offset&= ~lsn_t{write_size_1};
 
     if (length <= write_size_1)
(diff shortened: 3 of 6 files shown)


mariadb-DebarunBanerjee (approved):

dr-m:
Sorry, this is not this simple. I think that I need a way of reproducing the failure myself. I got an `rr replay` trace where the server was killed just after completing log resizing. Just before that, it would have written 512 bytes of `0xa5` (the `TRASH_ALLOC` value) to the log file offset 0x3000.

storage/innobase/log/log0log.cc
```diff
@@ -516,17 +516,14 @@ log_t::resize_start_status log_t::resize_start(os_offset_t size) noexcept
...
-          start_lsn= get_lsn();
-        }
         else
-        {
-          memcpy_aligned<16>(resize_buf, buf, (buf_free + 15) & ~15);
           start_lsn= first_lsn +
-            (~lsn_t{write_size - 1} & (write_lsn - first_lsn));
-        }
+            (~lsn_t{write_size - 1} &
+             (lsn_t{write_size - 1} + start_lsn - first_lsn));
```
mariadb-DebarunBanerjee, line 526: Here we are changing the start_lsn from write_lsn (aligned down) to log_sys.lsn (aligned up). If I am right, this change is independent to the other issue/bug that is being resolved by this patch. 

1. I could not understand what issue is being resolved by this part. Is there any issue if the current write_lsn is used ? Is it solving anything or some attempted improvement ?
2.  This is setting the start lsn after resize to a possible future LSN. Is this causing the resize to wait in the test ?

> This may block until some InnoDB persistent data is changed.

AFAICS, this change would cause behaviour change in a negative way.

storage/innobase/log/log0log.cc
```diff
@@ -780,19 +777,26 @@ inline void log_t::persist(lsn_t lsn) noexcept
...
+  if (UNIV_UNLIKELY(d <= 0))
+  {
+    d&= ~int64_t(block_size_1);
+    if (int64_t(d + length) <= 0)
+      return;
+    length+= d;
+    b-= d;
+    d= 0;
+  }
+  lsn_t offset= START_OFFSET + (lsn_t(d) & ~lsn_t{block_size_1}) %
```
mariadb-DebarunBanerjee, line 799: Just a note that this part of the change looks to be required only because we changed the resize lsn ahead of write_lsn in log_t::resize_start.

mariadb-DebarunBanerjee (changes requested):
Hi @dr-m,

Thanks for identifying and fixing this critical issue for resize. While I agree with one part of the change I have some concern for changing the resize LSN ahead of current LSN. Please see my comments.

mysql-test/suite/innodb/t/log_file_size_online.test
```diff
@@ -4,16 +4,25 @@
 
 let SEARCH_FILE = $MYSQLTEST_VARDIR/log/mysqld.1.err;
 
-SET GLOBAL innodb_log_file_size=4194304;
```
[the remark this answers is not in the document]
dr-m: Yes, this change to the test is described in the commit message. I can reword that in the present tense if that is less confusing.

storage/innobase/log/log0log.cc
```diff
@@ -804,7 +808,7 @@ void log_t::resize_write_buf(size_t length) noexcept
   }
 
   ut_a(os_file_write_func(IORequestWrite, "ib_logfile101", resize_log.m_file,
-                          buf, offset, length) == DB_SUCCESS);
+                          b, offset, length) == DB_SUCCESS);
```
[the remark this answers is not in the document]
dr-m: The revision to the test `innodb.log_file_size_online` in this pull request does catch it. The logic got more wrong in 4ca355d863e3b6a14439eebbb2958afccb3548e3, where the parameter `resize_flush_buf` was changed to `buf` (which in the more complex version #3327 was a function parameter). Before 4ca355d863e3b6a14439eebbb2958afccb3548e3, each invocation of `log_t::write_buf()` would invoke `std::swap(resize_buf, resize_flush_buf)` and there was no problem with this logic. Now we only swap the buffers conditionally.

However, before 4ca355d863e3b6a14439eebbb2958afccb3548e3 there was a problem that we’d write garbage to the beginning of `log_sys.resize_log`. Also the impact of that is revealed by the revised test `innodb.log_file_size_online`, from time to time, if you revert the changes to `log_t::resize_start()`.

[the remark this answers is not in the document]
dr-m: In 4ca355d863e3b6a14439eebbb2958afccb3548e3 `log_t::write_buf()` was optimized so that in case the last written log block is incomplete `(length <= write_size_1)`, we will keep filling the current `log_sys.buf` or `log_sys.resize_buf`. This will avoid a `memcpy` of up to several kilobytes per call when the same block is being written over and over until it is completely filled.

I would recommend you to run the modified test `innodb.log_file_size_online` some 100 times, with some of these changes reverted. If you revert any of the changes to `innodb_log_file_size_update()`, `log_t::resize_start()`, `log_t::resize_write_buf()`, `log_t::write_buf()`, the test will start to fail.

storage/innobase/log/log0log.cc
dr-m: Before I came up with this simple solution, I had added a data member to `log_sys` that would be a copy of the current local variable `write_buf` in `log_t::write_buf()`. It is this buffer that should form the ‘seed’ of `resize_buf`. However, that would not be enough; the test `innodb.log_file_size_online` would still fail, rather rarely. The problem is that  the old `ib_logfile0` may have wrapped around an odd number of times, and therefore the sequence bit at the end of a mini-transaction would be 0 and not the initial value 1.

In the non-memory-mapped I/O, there are some constraints around `log_sys.buf_free`. To keep things simple, we align the `start_lsn` so that we can write the same blocks of data (except for the sequence bit) to both `log_sys.log` and `log_sys.resize_log` . The formula here used to be able to choose a `start_lsn` that is smaller than the current `log_sys.lsn`. If the log resizing completes with such a LSN, then recovery could start parsing the resized log from such an LSN. That would be all fine and good if the ‘seed’ buffer contained sequence bits that are 1. However, if the sequence bits are 0, recovery would determine that it reached the end of the circular log, and stop right at the checkpoint. This was what motivated me to file #3476; the failure was not at all obvious.

If we wanted to adjust the sequence bits so that recovery can work, we’d have to add even more bookkeeping, just to satisfy a hard-to-test condition of a rarely used feature. We’d basically have to remember each "log block" or mini-transaction boundary, because the log can only be scanned forward, not backwards. The logic regarding adjusting the sequence bit (the assignments `resize_buf[s]= 1` in `log_t::resize_write()`) would only be enabled after this code has been executed. I would not want to execute that code unconditionally, just in case someone is going to initiate a log file size change on an idle or read-only server.

Any added bookkeeping in this area would not come for free. In bf0b82d24ba5cfa6d3cf42fba501ba4a57f266ff I spent a lot of effort to optimize the cache line usage, so that two 64-byte x86-64 cache lines pointing to `log_sys` are being used in a sensible way during `mtr_t::commit()`. If we added more data fields to a busy code path, a serious performance regression could occur.

storage/innobase/log/log0log.cc
dr-m: That is correct.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -18561,11 +18561,15 @@ static void innodb_log_file_size_update(THD *thd, st_mysql_sys_var*,
...
         mysql_mutex_lock(&buf_pool.flush_list_mutex);
-        const bool in_progress(buf_pool.get_oldest_modification(LSN_MAX) <
-                               log_sys.resize_in_progress());
-        if (in_progress)
+        const lsn_t oldest{buf_pool.get_oldest_modification(LSN_MAX)};
+        const lsn_t resizing{log_sys.resize_in_progress()};
+        static_assert(LSN_MAX + 1 == 0, "oldest + 1 must wrap around to 0");
+        if (resizing > oldest || resizing > oldest + 1)
+        {
+          buf_pool.page_cleaner_wakeup(true);
```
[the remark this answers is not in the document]
dr-m: You are right, this should be equivalent to the single line:
```c++
        if (log_sys.resize_in_progress() > buf_pool.get_oldest_modification(0))
```
I tested that with two batches of the following invocation:
```bash
./mtr --parallel=84 --repeat=10 innodb.log_file_size_online{,,,,,,,,,,,,,}{,,}
```
On the first batch, there was an unrelated failure (we still have other recovery bugs to analyze and fix):
```
2024-08-28  8:56:55 0 [Note] InnoDB: Resized log to 5.000MiB; start LSN=14486316
… (kill and restart)
2024-08-28  8:57:02 0 [Note] InnoDB: Starting crash recovery from checkpoint LSN=14487797
2024-08-28  8:57:02 0 [Note] InnoDB: End of log at LSN=14769434
2024-08-28  8:57:02 0 [Note] InnoDB: To recover: 200 pages
2024-08-28  8:57:02 0 [ERROR] InnoDB: OPT_PAGE_CHECKSUM mismatch on [page id: space=7, page number=229]
2024-08-28  8:57:02 0 [ERROR] InnoDB: Set innodb_force_recovery=1 to ignore corruption.
2024-08-28  8:57:02 0 [Note] InnoDB: Set innodb_force_recovery=1 to ignore corrupted pages.
2024-08-28  8:57:02 0 [ERROR] InnoDB: Unable to apply log to corrupted page [page id: space=7, page number=229]
2024-08-28  8:57:02 0 [ERROR] InnoDB: Failed to read page 229 from file './test/t.ibd': Page read from tablespace is corrupted.
2024-08-28  8:57:02 0 [Note] InnoDB: Set innodb_force_recovery=1 to ignore corrupted pages.
2024-08-28  8:57:02 0 [ERROR] InnoDB: Plugin initialization aborted at srv0start.cc[1512] with error Data structure corruption
```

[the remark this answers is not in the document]
dr-m: Side note: @mleich1 reproduced a hang during log resizing for 9f1995d0fd0dd627155059b973b0f8e96d800f3b (where this change is not present). So, the hang (or 1 thread busy looping) was possible also unrelated to the change to `log_t::resize_start()`.

storage/innobase/log/log0log.cc
dr-m: By the way, I had tried to justify this change in the commit message  e63dae8a15ad52f127818bc0c3a4e4b5f6a51358. I will try to make it even clearer.

mysql-test/suite/innodb/t/log_file_size_online.test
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: No, I didn't mean to change the commit message.  My doubt was if the new tests are to be added separately or in this patch. I could not see any kill/restart in the test modifications of the current patch.

storage/innobase/log/log0log.cc
dr-m: Yes, in [MDEV-14425](https://jira.mariadb.org/browse/MDEV-14425) (685d958e38b825ad9829be311f26729cccf37c46) the 31-bit `LOG_BLOCK_HDR_NO` was shrunk to 1 bit. That bit will flip if and only if the the sequence byte (right before the CRC-32C of the mini-transaction bytes) is at a lower file offset (near `log_sys.START_OFFSET`) than the start of the mini-transaction. That does make the value of the bit dependent on the `innodb_log_file_size`.

We can (and actually during `SET GLOBAL innodb_log_file_size` as well as in `mariadb-backup --backup` already do) replicate log records into an *append-only* fashion, always setting the sequence bit as 1. I specifically chose the initial value as 1 so that we can more conveniently detect the end of the log when the `ib_logfile0` has initially been filled with NUL bytes.

I would be very reluctant to change a file format.

To better understand the problem scenario, you can revert this code change and run
```bash
while ./mtr --rr --parallel=auto innodb.log_file_size_online{,,,,,,,,,,,,,,}; do :; done
```
and finally check the failure with `rr replay`. That is how I came up with this solution.

mariadb-DebarunBanerjee: @dr-m explained to me that the problem is smaller in magnitude than I perceived. That is we already generate a different buffers at mtr commit for resize log and the sequence BIT is set there correctly. This double copy may not look very optimal but should be fine only during resize time.

So, the only issue left was the log that were already written by mtrs but not yet flushed, i.e. write_lsn -> log_sys.lsn. The current patch is trying to get rid of this problem by trying to set the start LSN for resized log to log_sys.lsn instead. However, due to alignment requirement we need to go either ahead or behind log_sys LSN. Going behind would need addressing the incorrect sequence BIT and going ahead (the current patch) could lead to indefinite wait when server is idle.

So, it seems ...

Either 

- we accept the apparent hang of the SET GLOBAL in idle server and document it. User needs to have some concurrent operation to let the SET GLOBAL complete.

Or
- make more code changes to address it.

Possible solutions seems to be to parse the redo logs from write_lsn -> log_sys LSN and populate the resize_buf with correct sequence BIT or start from log_sys.lsn rounded up and device dummy mtrs to progress the LSN.

dr-m: To avoid a hang of `SET GLOBAL innodb_log_file_size`, `innodb_log_file_size_update()` can write enough dummy log records so that the LSN will advance enough. There is an `#if 0` section code `log_pad()` for that purpose. Recovery can deal with such dummy records.

dr-m: 984606d74705e1a4c67945531edeff81692f403d addresses this by making `innodb_log_file_size_update()` write redundant `FILE_CHECKPOINT` records if needed. The changes to the test `innodb.log_file_size_online` are smaller, so that it is likely that the first `SET GLOBAL innodb_log_file_size` would hang if this tweak were not present. The tweak (as well as the hang) was validated by running 3×2×10 rounds of that test. At least 1 of the 6 concurrent test runs would typically hang when `mysql-test/var` is linked to `/run/user/$UID/var`.

mysql-test/suite/innodb/t/log_file_size_online.test
```diff
@@ -4,16 +4,25 @@
...
-SHOW VARIABLES LIKE 'innodb_log_file_size';
-SELECT global_value FROM information_schema.system_variables
-WHERE variable_name = 'innodb_log_file_size';
+--connect con1,localhost,root
+# This may block until some InnoDB persistent data is changed.
+send SET GLOBAL innodb_log_file_size=4194304;
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Thanks for resolving the issue.

mariadb-DebarunBanerjee (approved):
Hi @dr-m,
Thanks for explaining all the design details and addressing the concern 100%. The patch looks good to me.
-Deb

PR #3459, opened 2024-08-13
MDEV-33075 [backport/2f5174e556] [10.11] Resolve server shutdown issues on macOS, Solaris, and FreeBSD

sitano:
Backport of 2f5174e556cd247133aa14d7b37372ae49fe83c5: MDEV-33075 Resolve server shutdown issues on macOS, Solaris, and FreeBSD.

This commit addresses multiple server shutdown problems observed on macOS,
Solaris, and FreeBSD:

1. Corrected a non-portable assumption where socket shutdown was expected
to wake up poll() with listening sockets in the main thread.

Use more robust self-pipe to wake up poll() by writing to the pipe's write
end.

2. Fixed a random crash on macOS from pthread_kill(signal_handler)
when the signal_handler was detached and the thread had already exited.

Use more robust `kill(getpid(), SIGTERM)` to wake up the signal handler
thread.

3. Made sure, that signal handler thread always exits once `abort_loop` is
set, and also calls `my_thread_end()` and clears `signal_thread_in_use`
when exiting.

This fixes warning "1 thread did not exit"  by `my_global_thread_end()`
seen on FreeBSD/macOS when the process is terminated via signal.

Additionally, the shutdown code underwent light refactoring
for better readability and maintainability:

- Modified `break_connect_loop()` to no longer wait for the main thread,
  aligning behavior with Windows (since 10.4).
- Removed dead code related to the unused `USE_ONE_SIGNAL_HAND`
  preprocessor constant.
- Eliminated support for `#ifndef HAVE_POLL` in `handle_connection_sockets`
  This code is also dead, since 10.4

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -1395,6 +1395,9 @@ Dynamic_array<MYSQL_SOCKET> listen_sockets(PSI_INSTRUMENT_MEM, 0);
 bool unix_sock_is_online= false;
 static int systemd_sock_activation; /* systemd socket activation */
 
+/** wakeup listening(main) thread by writing to this descriptor */
+static int termination_event_fd= -1;
+
 
 C_MODE_START
 #ifdef WITH_PERFSCHEMA_STORAGE_ENGINE
@@ -1651,44 +1654,20 @@ static my_bool warn_threads_active_after_phase_2(THD *thd, void *)
 
 static void break_connect_loop()
 {
-#ifdef EXTRA_DEBUG
-  int count=0;
-#endif
-
   abort_loop= 1;
 
 #if defined(_WIN32)
   mysqld_win_initiate_shutdown();
 #else
-  /* Avoid waiting for ourselves when thread-handling=no-threads. */
-  if (pthread_equal(pthread_self(), select_thread))
-    return;
-  DBUG_PRINT("quit", ("waiting for select thread: %lu",
-                      (ulong)select_thread));
-
   mysql_mutex_lock(&LOCK_start_thread);
-  while (select_thread_in_use)
+  if (termination_event_fd >= 0)
   {
-    struct timespec abstime;
-    int UNINIT_VAR(error);
-    DBUG_PRINT("info",("Waiting for select thread"));
-
-#ifndef DONT_USE_THR_ALARM
-    if (pthread_kill(select_thread, thr_client_alarm))
-      break;					// allready dead
-#endif
-    set_timespec(abstime, 2);
-    for (uint tmp=0 ; tmp < 10 && select_thread_in_use; tmp++)
+    uint64_t u= 1;
+    if(write(termination_event_fd, &u, sizeof(uint64_t)) < 0)
     {
-      error= mysql_cond_timedwait(&COND_start_thread, &LOCK_start_thread,
-                                  &abstime);
-      if (error != EINTR)
-        break;
+      sql_print_error("Couldn't send event to terminate listen loop");
+      abort();
     }
-#ifdef EXTRA_DEBUG
-    if (error != 0 && error != ETIMEDOUT && !count++)
-      sql_print_error("Got error %d from mysql_cond_timedwait", error);
-#endif
   }
   mysql_mutex_unlock(&LOCK_start_thread);
 #endif /* _WIN32 */
@@ -6324,6 +6303,25 @@ void handle_connections_sockets()
     fds.push(local_fds);
     set_non_blocking_if_supported(listen_sockets.at(i));
   }
+  int termination_fds[2];
+  if (pipe(termination_fds))
+  {
+    sql_print_error("pipe() failed %d", errno);
+    DBUG_VOID_RETURN;
+  }
+#ifdef FD_CLOEXEC
+  for (int fd : termination_fds)
+    (void)fcntl(fd, F_SETFD, FD_CLOEXEC);
+#endif
+
+  mysql_mutex_lock(&LOCK_start_thread);
+  termination_event_fd= termination_fds[1];
+  mysql_mutex_unlock(&LOCK_start_thread);
+
+  struct pollfd event_fd;
+  event_fd.fd= termination_fds[0];
+  event_fd.events= POLLIN;
+  fds.push(event_fd);
 
   sd_notify(0, "READY=1\n"
             "STATUS=Taking your SQL requests now...\n");
@@ -6388,6 +6386,12 @@ void handle_connections_sockets()
       }
     }
   }
+  mysql_mutex_lock(&LOCK_start_thread);
+  for(int fd : termination_fds)
+    close(fd);
+  termination_event_fd= -1;
+  mysql_mutex_unlock(&LOCK_start_thread);
+
   sd_notify(0, "STOPPING=1\n"
             "STATUS=Shutdown in progress\n");
   DBUG_VOID_RETURN;
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -2092,6 +2092,7 @@ static void clean_up(bool print_message)
 */
 static void wait_for_signal_thread_to_end()
 {
+#ifndef _WIN32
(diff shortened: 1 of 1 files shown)


sitano:
I did not have time to understand if this part needed and missing in 11.x

```
#ifdef USE_ONE_SIGNAL_HAND
    case THR_SERVER_ALARM:
      process_alarm(sig);			// Trigger alarms.
      break;
#endif
```

The critical part for me of all those 4 parts is the thread exit on abort in signal_hand - without it the whole 10.x branch just hangs on exit and db init on MacOS.

sql/mysqld.cc
```diff
@@ -2092,33 +2092,17 @@ static void clean_up(bool print_message)
...
-  for (i= 0 ; i < n_waits && signal_thread_in_use; i++)
-  {
-    err= pthread_kill(signal_thread, MYSQL_KILL_SIGNAL);
-    if (err)
-      break;
-    my_sleep(100000); // Give it time to die, .1s per iteration
-  }
-
-  if (err && err != ESRCH)
+  for (uint i= 0 ; i < 100 && signal_thread_in_use; i++)
```
vaintroub, line 2079: https://buildbot.mariadb.org/#/builders/572/builds/11911 is failing, this is testing this code.

The current code in 11.6 is slightly different https://github.com/MariaDB/server/blob/68e369e3a97940303173b124fdfb7c626effbe6b/sql/mysqld.cc#L2095

It retains kill() instead of pthread_kill(), like here, but also has some warnings, if signal thread did not shutdown in time, and pthread_join(), that was removed here

sitano: yeah, thx for point this out, seems like I was a bit confused by the amounts of changes in https://github.com/MariaDB/server/blob/2f5174e556cd247133aa14d7b37372ae49fe83c5/sql/mysqld.cc#L2074-L2088 that did not left anything from this func. I will fix it.

however, speaking of failures its interesting that its actually failing for Debian and Ubuntu in a segfault assert in https://github.com/MariaDB/server/blob/10.11/mysql-test/suite/rpl/t/rpl_shutdown_sighup.test (I am not sure which commit it uses in https://buildbot.mariadb.org/#/builders/369/builds/23487/steps/7/logs/stdio : line 3404).

Anyway it seems like it segfaults in signal_hand() that is a different function, and i did expect it not to call a reload:
```
[Current thread is 1 (Thread 0x7f5f147a5700 (LWP 104156))]
#0  0x00007f5f227fcf68 in pthread_kill () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00005625f2a46126 in my_write_core (sig=11) at /home/buildbot/amd64-ubuntu-2004-debug/build/mysys/stacktrace.c:424
#2  0x00005625f21b8c3e in handle_fatal_signal (sig=11) at /home/buildbot/amd64-ubuntu-2004-debug/build/sql/signal_handler.cc:358
#3  <signal handler called>
#4  0x00005625f1d50b5d in hash_filo::clear (this=0x0, locked=false) at /home/buildbot/amd64-ubuntu-2004-debug/build/sql/hash_filo.h:87
#5  0x00005625f21d6cb2 in hostname_cache_refresh () at /home/buildbot/amd64-ubuntu-2004-debug/build/sql/hostname.cc:134
#6  0x00005625f20157a2 in reload_acl_and_cache (thd=0x0, options=2147499823, tables=0x0, write_to_binlog=0x7f5f147a4c10) at /home/buildbot/amd64-ubuntu-2004-debug/build/sql/sql_reload.cc:379
[-------------------]
#7  0x00005625f1c84d21 in signal_hand () at /home/buildbot/amd64-ubuntu-2004-debug/build/sql/mysqld.cc:3277
#8  0x00005625f24aa749 in pfs_spawn_thread (arg=0x5625f6888268) at /home/buildbot/amd64-ubuntu-2004-debug/build/storage/perfschema/pfs.cc:2201
#9  0x00007f5f227f4609 in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#10 0x00007f5f223c5353 in clone () from /lib/x86_64-linux-gnu/libc.so.6
```

I will look into that.

vaintroub: The amount of changes to this function is has to do with a different bug,  MDEV-30260: Slave crashed:reload_acl_and_cache during shutdown , which they fixed after my changes that you have ported.
This is 952ab9a59659b28f10ce2770e17249e0ff5b0681 

It is a rare condition, where SIGHUP signal handler (which does "reloads" in signal handler thread) coincidences with shutdown.

Will it still be working OK, if you just copy that function verbatim from 11.6 (I think 11.6 should work ok on macOS and FreeBSD etc, I think)?

Also, it makes sense to base the patch tor the earliest affected and still supported branch, which is 10.5. It will be merged to all subsequent versions anyway.

sitano: Got you @vaintroub. Surprisingly I have eliminated pthread_join in a previous version. Reverted needed pieces back. Now it must be ok. I have ported work to 10.5 at https://github.com/MariaDB/server/pull/3469, the code there is a bit different for termiation_fd commit so had to keep it separate.

Also, any ideas why this may fail? Does it fail sometimes or probably I missed something again? Didn't look at logs yet:

```
Completed: All 6112 tests were successful.

Errors/warnings were found in logfiles during server shutdown after running the
following sequence(s) of tests:
    binlog_encryption.rpl_parallel_missed_error_handling

298 tests were skipped, 213 by the test itself.

mysql-test-run: *** ERROR: There where errors/warnings in server logs after running test cases.
```

sitano:
```
***Warnings generated in error logs during shutdown after running tests: binlog_encryption.rpl_parallel_missed_error_handling
Warning: Memory not freed: -2072
```

2072 is an interesting number. Close to 2KB.

sitano: 2nd attempt has different failure. https://buildbot.mariadb.org/#/builders/640/builds/4586/steps/7/logs/stdio

```
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f13c8737f44 in pthread_kill () from /lib/x86_64-linux-gnu/libpthread.so.0
[Current thread is 1 (Thread 0x7f13c7eb3980 (LWP 230258))]
#0  0x00007f13c8737f44 in pthread_kill () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x000055f48e7c92ca in handle_fatal_signal ()
#2  0x000055f48daf333b in SignalHandler(int) ()
#3  <signal handler called>
#4  0x00007f13c82219af in poll () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x000055f48dac3ff3 in poll ()
#6  0x000055f48db1bce5 in handle_connections_sockets() ()
#7  0x000055f48db0bac2 in mysqld_main(int, char**) ()
#8  0x00007f13c8155d0a in __libc_start_main () from /lib/x86_64-linux-gnu/libc.so.6
#9  0x000055f48da7945a in _start ()
```

during atomic.alter_table.

@vaintroub how do you think, its my changes or the msan ci tests are flaky?

vaintroub: Does not look like your changes. That test ran into timeout, and was terminated with SIGABRT, which mtr does to produce coredump, to analyze a potential hang with gdb . but there is nothing interesting in the crash , there is not even shutdown or anything. Maybe it just took too long.

vaintroub: yes, this test [regularly runs into timeout,](https://buildbot.mariadb.net/ci/reports/cross_reference#branch=&revision=&platform=&fail_name=atomic.alter_table&fail_variant=&fail_info_full=&typ=&info=&dt=&limit=100&fail_info_short=) and very often on "msan" builders

sitano: should I re-execute CI until it succeeds?

vaintroub: I do not think it is necessary

sitano:
10.5 seems to be ready green too

cvicentiu:
@vaintroub can you please confirm if this should be merged or not any more?

grooverdan:
Fairly sure the 10.5 backport and subsequent merge up handled this. Thanks @sitano for pushing for the backport.

PR #3461, opened 2024-08-13
MDEV-34753 memory pressure - erroneous termination condition

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34753*

## Description

The 'if (!m_abort) break' condition was inverted by accident.

Thanks Kristian Nielsen for noticing.

## Release Notes

Linux memory pressure events are now correctly handled.

## How can this PR be tested?

innodb.mem_pressure tests shows the handling:

Version: '10.11.10-MariaDB-debug-log'  socket: '/home/dan/repos/build-mariadb-server-10.11-debug/mysql-test/var/tmp/mysqld.1.sock'  port: 19000  Source distribution
2024-08-14  8:23:29 4 [Note] InnoDB: Memory pressure event freed 146 pages
2024-08-14  8:23:29 0 [Note] InnoDB: Memory pressure event freed 0 pages

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/mem_pressure.test
+++ b/mysql-test/suite/innodb/t/mem_pressure.test
@@ -1,5 +1,5 @@
 --source include/have_debug.inc
---source include/linux.inc
+--source include/have_cgroupv2.inc
 --source include/not_embedded.inc
 --source include/have_innodb.inc
 --source include/have_sequence.inc
@@ -31,11 +31,15 @@ WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 set debug_dbug="d,trigger_garbage_collection";
 SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
 
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+# either a fail or the pressure event
+let SEARCH_PATTERN= [Mm]emory pressure.*;
+--source include/search_pattern_in_file.inc
+
 SELECT CAST(VARIABLE_VALUE AS INTEGER) < @dirty_prev AS LESS_DIRTY_IS_GOOD
 FROM INFORMATION_SCHEMA.GLOBAL_STATUS
 WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 
-let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
 let SEARCH_PATTERN= InnoDB: Memory pressure event freed.*;
 --source include/search_pattern_in_file.inc
 
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -830,6 +830,7 @@ public:
       my_close(m_fds[m_num_fds].fd, MYF(MY_WME));
       m_fds[m_num_fds].fd= -1;
     }
+    m_event_fd= -1;
   }
 
   static void pressure_routine(mem_pressure *m);
@@ -838,10 +839,8 @@ public:
   void trigger_collection()
   {
     uint64_t u= 1;
-    if (m_event_fd >=0 && write(m_event_fd, &u, sizeof(uint64_t)) != sizeof(uint64_t))
+    if (m_event_fd < 0 || write(m_event_fd, &u, sizeof(uint64_t)) != sizeof(uint64_t))
       sql_print_information("InnoDB: (Debug) Failed to trigger memory pressure");
-    else /* assumed failed to meet intialization criteria, so trigger directy */
-      buf_pool.garbage_collect();
   }
 #endif
 
@@ -900,7 +899,7 @@ void mem_pressure::pressure_routine(mem_pressure *m)
       else
         break;
     }
-    if (!m->m_abort)
+    if (m->m_abort)
       break;
 
     for (pollfd &p : st_::span<pollfd>(m->m_fds, m->m_num_fds))
--- a/mysql-test/include/have_cgroupv2.inc
+++ b/mysql-test/include/have_cgroupv2.inc
@@ -1,0 +1,28 @@
+--source include/linux.inc
+--error 0,1
+perl;
+use strict;
+use warnings;
+use File::Spec;
+
+# Read the cgroup file
+my $cgroup_file = '/proc/self/cgroup';
+open my $fh, '<', $cgroup_file or exit 1;
+my $line = <$fh>;
+close $fh;
+
+# Process the cgroup file content
+$line =~ s/^0:://;
+chomp($line);
+
+# Construct the path
+my $path = File::Spec->catdir('/sys/fs/cgroup', $line, 'memory.pressure');
+
+# Check if the file is writable exists
+exit 0 if -w $path;
+exit 1;
+EOF
+if ($errno)
+{
+  --skip Requires cgroupv2
+}
--- a/mysql-test/suite/innodb/r/mem_pressure.result
+++ b/mysql-test/suite/innodb/r/mem_pressure.result
@@ -15,6 +15,7 @@ FROM INFORMATION_SCHEMA.GLOBAL_STATUS
 WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 set debug_dbug="d,trigger_garbage_collection";
 SET GLOBAL innodb_buffer_pool_size=@@innodb_buffer_pool_size;
+FOUND 1 /[Mm]emory pressure.*/ in mysqld.1.err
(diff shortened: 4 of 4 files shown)


knielsen (changes requested):
Yes, the patch is easy, but what about testing?

Has this code for handling memory pressure been tested at all? And if so, how?

I'm thinking something like running a VM with limited amount of memory, running some random load (sysbench, whatever) with a large buffer pool, maybe running some memory-hungry job on the side and see that the mariadbd gets OOM-killed without the feature. And then test the same with the feature enabled.

And a mysql-test-run testcase that uses some debug facility to artificially trigger the code path at least, to test against something like this typo.

[I'm asking, because it seems until now the entire logic of the memory pressure handling has been dead code? But with this fix it will become active. So we should be careful not to activate new untested code in 10.11 GA release...]

grooverdan:
Updated existing innodb.mem_pressure to trigger the same event loop that would have come from the OS.

storage/innobase/buf/buf0buf.cc
```diff
@@ -838,10 +839,11 @@ class mem_pressure
...
   {
     uint64_t u= 1;
-    if (m_event_fd >=0 && write(m_event_fd, &u, sizeof(uint64_t)) != sizeof(uint64_t))
+    if (m_event_fd < 0 || write(m_event_fd, &u, sizeof(uint64_t)) != sizeof(uint64_t))
+    {
       sql_print_information("InnoDB: (Debug) Failed to trigger memory pressure");
-    else /* assumed failed to meet intialization criteria, so trigger directy */
       buf_pool.garbage_collect();
+    }
   }
```
knielsen, line 847: Aha. So this is the reason that the existing test case did not fail from the bug. There were actually two bugs, the inverted m_abort test, and this incorrect test which (IIUC) ran the garbage_collect() when it _succeeded_ in triggering the memory pressure thread, rather than when it failed to trigger (as I assume was intended?).

So the two bugs kind of cancelled each other, and the garbage_collect() was triggered anyway.

But why have this condition at all? What is the purpose of triggering buf_pool.garbage_collect() directly here if you cannot signal the thread? That seems to only hide bugs (as indeed it did in this case).

So better just remove this call of garbage_collect()?

knielsen: So now with this change, you have a race condition in the test case. Because the test case checks for message from the garbage_collect() in the error log, but this message only comes asynchroneously, and thread scheduling can delay the mem_pressure thread and make the test not see the message and fail.

You can easily see this by putting a my_sleep(1000000) in the thread routine just before running the garbage_collect(), then the test fails.

This is going to trigger random failures in buildbot, so needs to be fixed.

mysql-test/suite/innodb/t/mem_pressure.test
```diff
@@ -41,4 +41,35 @@ let SEARCH_PATTERN= InnoDB: Memory pressure event freed.*;
...
+SELECT CAST(VARIABLE_VALUE AS INTEGER) < @dirty_prev AS LESS_DIRTY_IS_GOOD
+FROM INFORMATION_SCHEMA.GLOBAL_STATUS
+WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
+
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+let SEARCH_PATTERN= InnoDB: Memory pressure event freed.*;
+--source include/search_pattern_in_file.inc
+
+set debug_dbug=@save_dbug;
+
```
knielsen, line 74: Well, if you really want to run this twice, do it in a while ($i < 2) { ... inc $i; } loop, don't copy-paste duplicate the code.

Or maybe it's fine to run it just once as originally, if you fix the debug code that triggers the condition from debug injection. That seems to have been the reason that this test could succeed, even though the code in the thread could never run.

But you still need to fix the race condition in this test, see other comment.

knielsen (changes requested):
Thanks, Daniel.

The most important question from before still remains: was this ever tested (eg. manually) on actual OOM conditions? This is complex logic interacting with linux kernel interfaces and logic, I think it needs to be tested at least once that this code can actually do something sensible in the actual condition it is intended to handle, even if that condition cannot be set up in mysql-test-run.

Also see additional comments.

 \- Kristian.

grooverdan:
> Thanks, Daniel.
> 
> The most important question from before still remains: was this ever tested (eg. manually) on actual OOM conditions? This is complex logic interacting with linux kernel interfaces and logic, I think it needs to be tested at least once that this code can actually do something sensible in the actual condition it is intended to handle, even if that condition cannot be set up in mysql-test-run.
> 
> Also see additional comments.
> 
> - Kristian.

Tests where done in https://jira.mariadb.org/browse/MDEV-25341.  It wasn't until later I discovered eventfds and implemented #2864 that the bug disabled the functionality.

storage/innobase/buf/buf0buf.cc
grooverdan: yep - `garbage_collect` removed and replaced with a mtr test case for cgroupsv2 only.

Race condition fixed.

knielsen (changes requested):
> Race condition fixed.

Where? I don't see anything in the patch that would fix the race.
In fact the race is still there, and the test still fails, did you even try
it?
I just tried to apply your patch in a 11.4 checkout and add this one-liner,
and then the testcase fails:
```
index fee6b959493..39da10aec1b 100644
@@ -902,6 +902,7 @@ void mem_pressure::pressure_routine(mem_pressure *m)
     if (m->m_abort)
       break;
 
+my_sleep(1000000);
     for (pollfd &p : st_::span<pollfd>(m->m_fds, m->m_num_fds))
     {
       if (p.revents & POLLPRI)
```

```
innodb.mem_pressure                      [ fail ]
        Test ended at 2024-08-20 08:52:39

CURRENT_TEST: innodb.mem_pressure
--- /kvm/src/my/sn-mdev30255/mysql-test/suite/innodb/r/mem_pressure.result	2024-08-20 08:47:23.750571295 +0200
+++ /kvm/src/my/sn-mdev30255/mysql-test/suite/innodb/r/mem_pressure.reject	2024-08-20 08:52:38.224915283 +0200
@@ -20,7 +20,7 @@
 FROM INFORMATION_SCHEMA.GLOBAL_STATUS
 WHERE VARIABLE_NAME='Innodb_buffer_pool_pages_dirty';
 LESS_DIRTY_IS_GOOD
-1
-FOUND 1 /InnoDB: Memory pressure event freed.*/ in mysqld.1.err
+0
+NOT FOUND /InnoDB: Memory pressure event freed.*/ in mysqld.1.err
 set debug_dbug=@save_dbug;
 # End of 10.11 tests

Result length mismatch
```
> yep - garbage_collect removed and replaced with a mtr test case for cgroupsv2 only.

Aha - so the reason to have the direct trigger in case of the failure was to
have the same behaviour on systems that does not have the cgroups and memory
pressure reporting facility available. I missed that in the prior review -
that is why I asked "What is the purpose of triggering
buf_pool.garbage_collect() directly here if you cannot signal the thread?".

So now I understand the purpose the prior code makes more sense. Anyway, the
new way is fine too if it makes the test work, whichever you prefer.

> Tests where done in https://jira.mariadb.org/browse/MDEV-25341. It wasn't until later I discovered eventfds and implemented #2864 that the bug disabled the functionality.

Ok, that makes sense, thanks for the explanation. So the code was tested
prior to a48c1b89c849a390be9e5b210759369b35777b39 which changed to eventfd
and added the reversed if (!m->m_abort) condition. I was otherwise wondering
how it was at all possible to test what looked like dead code in the version
I was looking at. It was just unfortunate (and somewhat of a reviewer miss)
that two distinct bugs ended up cancelling one another and make the test
case not catch the problem.

You still didn't explain how the code was tested, but I will take your word
for it that the code was exercised at least once on a system where real
memory pressure caused kernel to report and InnoDB to release memory.

The sporadic test failure still needs to be fixed, otherwise looks ok to me.

A word of advice, maybe slow down a bit? This is the third round of review.
It's good to take the time to understand exactly what the problem is before
typing out code. And if reviewer comments are not clear, better ask; the
reviewer does not know the code as well as the author, and will often have
misunderstood. One thing that helps me when I feel even simple issues are
taking too long time with endless detail, is to remember that MariaDB is one
of the highest-profile open-source projects out there, and running many
critical systems. So extra care is totally worth it before the push that
sends our code more or less directly to production systems.

knielsen:
Suggested patch to fix the race in the test case (goes on top of the current commit in this pull request): https://github.com/MariaDB/server/commit/28e2e02c8b8fa79982785ce9000db49d53ce11e0
Ok to push/merge with this fix (or some variant or other way to fix the test case if you prefer).

knielsen (approved):
Approved after fixing race in test case as suggested

grooverdan:
merge other variant before I saw this one. Thanks @knielsen

PR #3464, opened 2024-08-16
MDEV-34756  Validation of new foreign key skipped if innodb_alter_copy_bulk= ON

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34756*

## Description
- During copy algorithm, InnoDB should disable bulk insert operation if the table has foreign key relation and foreign key check is enabled.
## How can this PR be tested?
./mtr innodb.alter_copy_bulk

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/alter_copy_bulk.test
+++ b/mysql-test/suite/innodb/t/alter_copy_bulk.test
@@ -42,3 +42,28 @@ INSERT INTO t1 SELECT seq, seq * 2 FROM seq_3_to_65536;
 # Alter should use temporary file for sorting
 ALTER TABLE t1 ADD INDEX(f2);
 DROP TABLE t1;
+
+--echo #
+--echo # MDEV-34756 Validation of new foreign key skipped
+--echo #            if innodb_alter_copy_bulk=ON
+--echo #
+CREATE TABLE t1(f1 INT NOT NULL PRIMARY KEY,
+		f2 INT NOT NULL)ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT NOT NULL PRIMARY KEY,
+		f2 INT NOT NULL)ENGINE=InnoDB;
+--enable_info
+ALTER TABLE t2 ADD CONSTRAINT FOREIGN KEY(f1) REFERENCES t1(f1);
+--disable_info
+INSERT INTO t1 VALUES (1, 1);
+INSERT INTO t2 VALUES (1, 2);
+--replace_regex /#sql-alter-[0-9a-f-]*/#sql-alter/
+--error ER_NO_REFERENCED_ROW_2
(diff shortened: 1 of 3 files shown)


mysql-test/suite/innodb/r/alter_copy_bulk,OFF.rdiff
```diff
@@ -1,5 +1,5 @@
---- bulk_copy_alter.result
-+++ bulk_copy_alter,non_bulk_alter_copy.result
+--- alter_copy_bulk.result
++++ alter_copy_bulk,OFF.result
```
dr-m, line 2: This change seems to be unnecessary. The `---` and `+++` lines only need to exist at the start of the file; the exact contents should not matter.

mysql-test/suite/innodb/t/alter_copy_bulk.test
```diff
@@ -42,3 +42,22 @@ INSERT INTO t1 SELECT seq, seq * 2 FROM seq_3_to_65536;
...
+		f2 INT NOT NULL)ENGINE=InnoDB;
+INSERT INTO t2 VALUES (1, 1);
+--replace_regex /#sql-alter-[0-9a-f-]*/#sql-alter/
+--error ER_NO_REFERENCED_ROW_2
+ALTER TABLE t2 ADD CONSTRAINT FOREIGN KEY(f2) REFERENCES t1(f1);
+SET session FOREIGN_KEY_CHECKS=0;
+INSERT INTO t1 VALUES(1, 1);
+ALTER TABLE t2 ADD CONSTRAINT FOREIGN KEY(f2) REFERENCES t1(f1);
+SET session FOREIGN_KEY_CHECKS=1;
+ALTER TABLE t1 FORCE;
```
dr-m, line 62: I would suggest `--enable_info` and `--disable_info` around this, to demonstrate that the `ALGORITHM=COPY` code path is being used here. It should report "1 row(s) affected", while `ALGORITHM=INPLACE` should always report "0 row(s) affected".

Please add an `ALTER TABLE` right before the `INSERT`, to show that no error will be reported for the violation. I would expect that one to use the `ALGORITHM=INPLACE` code path, which should be indicated by the `--enable_info` output.

I’d also suggest writing
```sql
SET STATEMENT foreign_key_checks=0 FOR
ALTER TABLE t2 ADD CONSTRAINT FOREIGN KEY(f2) REFERENCES t1(f1);
```
so that there is no need to "permanently" change the session variable.

dr-m (commented):
Sorry for the delay. The code change looks good to me. I have some suggestions related to the regression test.

mysql-test/suite/innodb/t/alter_copy_bulk.test
Thirunarayanan: OK. Fixed the issue.

dr-m (approved):

PR #3466, opened 2024-08-16
MDEV-34529  Shrink the system tablespace when system tablespace contains MDEV-30671 leaked undo pages

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34529*

## Description
- InnoDB fails to shrink the system tablespace when it contains
the leaked undo log pages caused by MDEV-30671.

- InnoDB does free the unused segment in system tablespace
before shrinking the tablespace.

inode_info: Structure to store the inode page and offsets.

fil_space_t::garbage_collect():
Frees the system tablespace unused segment

fsp_free_unused_seg(): Frees the unused segment

fsp_get_sys_used_segment(): Iterates through all default
file segment and index segment present in system tablespace.

## How can this PR be tested?

Added the test case innodb.undo_leak which basically leaks the undo segment by purge thread.

## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARD- (https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.

### DIFF
--- a/mysql-test/suite/innodb/r/undo_leak_fail.result
+++ b/mysql-test/suite/innodb/r/undo_leak_fail.result
@@ -1,0 +1,94 @@
+call mtr.add_suppression("InnoDB: Cannot free the unused segments in system tablespace because a previous shutdown was not with innodb_fast_shutdown=0");
+call mtr.add_suppression("InnoDB: :autoshrink failed to read the used segment");
+call mtr.add_suppression("InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FULL list");
+call mtr.add_suppression("InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FREE list");
+call mtr.add_suppression("InnoDB: :autoshrink failed to free the segment");
+call mtr.add_suppression("Found .* prepared XA transactions");
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	10485760
+SET GLOBAL INNODB_FILE_PER_TABLE=0;
+Warnings:
+Warning	1287	'@@innodb_file_per_table' is deprecated and will be removed in a future release
+CREATE TABLE t1(f1 INT NOT NULL, f2 INT NOT NULL)STATS_PERSISTENT=0, ENGINE=InnoDB;
+XA START 'x';
+insert into t1 values (1, 1);
+XA END 'x';
+XA PREPARE 'x';
+set GLOBAL innodb_fast_shutdown=0;
+# restart
+# Fail to free the segment due to XA PREPARE transaction
+FOUND 2 /InnoDB: Cannot free the unused segments in system tablespace because a previous shutdown was not with innodb_fast_shutdown=0/ in mysqld.1.err
+XA COMMIT 'x';
+DROP TABLE t1;
+# restart: --debug_dbug=d,undo_segment_leak
+SET GLOBAL INNODB_FILE_PER_TABLE=0;
+Warnings:
+Warning	1287	'@@innodb_file_per_table' is deprecated and will be removed in a future release
+CREATE TABLE t1(f1 INT NOT NULL, f2 INT NOT NULL)STATS_PERSISTENT=0, ENGINE=InnoDB;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+UPDATE t1 SET f1 = f1 + 1 WHERE f1 > 1000;
+UPDATE t1 SET f2 = f2 + 1 WHERE f1 > 1000;
+UPDATE t1 SET f1 = f2 + 1 WHERE f1 > 1000;
+UPDATE t1 SET f2 = f1 + 1 WHERE f1 > 1000;
+DELETE FROM t1;
+DROP TABLE t1;
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	77594624
+# restart: --debug_dbug=d,unused_undo_free_fail_1
+# Fail to free the segment due to previous shutdown
+FOUND 4 /InnoDB: Cannot free the unused segments in system tablespace because a previous shutdown was not with innodb_fast_shutdown=0/ in mysqld.1.err
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	15728640
+SET GLOBAL innodb_fast_shutdown= 0;
+# Fail to free the segment while finding the used segments
+# restart: --debug_dbug=d,unused_undo_free_fail_2
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+IF(file_size>10485760,'ok',file_size)
+ok
+FOUND 1 /InnoDB: :autoshrink failed to read the used segment/ in mysqld.1.err
+FOUND 1 /InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FULL list/ in mysqld.1.err
+SET GLOBAL innodb_fast_shutdown= 0;
+# Fail to free the segment while finding the used segments
+# restart: --debug_dbug=d,unused_undo_free_fail_3
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+IF(file_size>10485760,'ok',file_size)
+ok
+FOUND 1 /InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FREE list/ in mysqld.1.err
+SET GLOBAL innodb_fast_shutdown= 0;
+# Fail to free the segment while freeing the unused segments
+# restart: --debug_dbug=d,unused_undo_free_fail_4
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+IF(file_size>10485760,'ok',file_size)
+ok
+FOUND 1 /InnoDB: :autoshrink failed to free the segment .* in page .*/ in mysqld.1.err
+SET GLOBAL innodb_fast_shutdown= 0;
+# Fail to free the segment while freeing the used segments
+# restart: --debug_dbug=d,unused_undo_free_fail_5
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+IF(file_size>10485760,'ok',file_size)
+ok
+FOUND 1 /InnoDB: :autoshrink failed to free the segment .* in page .*/ in mysqld.1.err
+SET GLOBAL innodb_fast_shutdown= 0;
+# restart
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	10485760
+# Fail to reset the TRX_SYS_FSEG_HEADER during undo tablespace
+# reinitialization. garbage_collect() shouldn't free the
+# TRX_SYS_FSEG_HEADER index node
+set global innodb_fast_shutdown=0;
+# restart: --innodb_undo_tablespaces=2 --debug_dbug=d,sys_fseg_header_fail
+FOUND 1 /InnoDB: :autoshrink freed the segment .* in page .*/ in mysqld.1.err
+set global innodb_fast_shutdown=0;
+# restart
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	10485760
--- a/mysql-test/suite/innodb/t/undo_leak_fail.test
+++ b/mysql-test/suite/innodb/t/undo_leak_fail.test
@@ -1,0 +1,119 @@
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+--source include/have_debug.inc
+--source include/not_embedded.inc
+
+call mtr.add_suppression("InnoDB: Cannot free the unused segments in system tablespace because a previous shutdown was not with innodb_fast_shutdown=0");
+call mtr.add_suppression("InnoDB: :autoshrink failed to read the used segment");
+call mtr.add_suppression("InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FULL list");
+call mtr.add_suppression("InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FREE list");
+call mtr.add_suppression("InnoDB: :autoshrink failed to free the segment");
+call mtr.add_suppression("Found .* prepared XA transactions");
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+
+SET GLOBAL INNODB_FILE_PER_TABLE=0;
+CREATE TABLE t1(f1 INT NOT NULL, f2 INT NOT NULL)STATS_PERSISTENT=0, ENGINE=InnoDB;
+XA START 'x';
+insert into t1 values (1, 1);
+XA END 'x';
+XA PREPARE 'x';
+set GLOBAL innodb_fast_shutdown=0;
+--source include/restart_mysqld.inc
+--echo # Fail to free the segment due to XA PREPARE transaction
+let SEARCH_PATTERN= InnoDB: Cannot free the unused segments in system tablespace because a previous shutdown was not with innodb_fast_shutdown=0;
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+XA COMMIT 'x';
+DROP TABLE t1;
+
+let $restart_parameters=--debug_dbug=d,undo_segment_leak;
+--source include/restart_mysqld.inc
+SET GLOBAL INNODB_FILE_PER_TABLE=0;
+CREATE TABLE t1(f1 INT NOT NULL, f2 INT NOT NULL)STATS_PERSISTENT=0, ENGINE=InnoDB;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+INSERT INTO t1 SELECT seq, seq FROM seq_1_to_4096;
+UPDATE t1 SET f1 = f1 + 1 WHERE f1 > 1000;
+UPDATE t1 SET f2 = f2 + 1 WHERE f1 > 1000;
+UPDATE t1 SET f1 = f2 + 1 WHERE f1 > 1000;
+UPDATE t1 SET f2 = f1 + 1 WHERE f1 > 1000;
+DELETE FROM t1;
+DROP TABLE t1;
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+let $shutdown_timeout=0;
+let $restart_parameters=--debug_dbug=d,unused_undo_free_fail_1;
+--source include/restart_mysqld.inc
+
+let $shutdown_timeout=;
+--echo # Fail to free the segment due to previous shutdown
+--let SEARCH_PATTERN= InnoDB: Cannot free the unused segments in system tablespace because a previous shutdown was not with innodb_fast_shutdown=0
+
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+SET GLOBAL innodb_fast_shutdown= 0;
+
+--echo # Fail to free the segment while finding the used segments
+let $restart_parameters=--debug_dbug=d,unused_undo_free_fail_2;
+--source include/restart_mysqld.inc
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+let SEARCH_PATTERN= InnoDB: :autoshrink failed to read the used segment;
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+
+let SEARCH_PATTERN= InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FULL list;
+--source include/search_pattern_in_file.inc
+SET GLOBAL innodb_fast_shutdown= 0;
+
+--echo # Fail to free the segment while finding the used segments
+let $restart_parameters=--debug_dbug=d,unused_undo_free_fail_3;
+--source include/restart_mysqld.inc
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+
+let SEARCH_PATTERN= InnoDB: :autoshrink failed due to .* in FSP_SEG_INODES_FREE list;
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+SET GLOBAL innodb_fast_shutdown= 0;
+
+--echo # Fail to free the segment while freeing the unused segments
+let $restart_parameters=--debug_dbug=d,unused_undo_free_fail_4;
+--source include/restart_mysqld.inc
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+
+let SEARCH_PATTERN= InnoDB: :autoshrink failed to free the segment .* in page .*;
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+
+SET GLOBAL innodb_fast_shutdown= 0;
+
+--echo # Fail to free the segment while freeing the used segments
+let $restart_parameters=--debug_dbug=d,unused_undo_free_fail_5;
+--source include/restart_mysqld.inc
+SELECT IF(file_size>10485760,'ok',file_size) FROM information_schema.innodb_sys_tablespaces WHERE space=0;
+
+let SEARCH_PATTERN= InnoDB: :autoshrink failed to free the segment .* in page .*;
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+
+SET GLOBAL innodb_fast_shutdown= 0;
+let $restart_parameters=;
+--source include/restart_mysqld.inc
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+
+--echo # Fail to reset the TRX_SYS_FSEG_HEADER during undo tablespace
+--echo # reinitialization. garbage_collect() shouldn't free the
+--echo # TRX_SYS_FSEG_HEADER index node
+set global innodb_fast_shutdown=0;
+let $restart_parameters=--innodb_undo_tablespaces=2 --debug_dbug=d,sys_fseg_header_fail;
+--source include/restart_mysqld.inc
+let SEARCH_PATTERN= InnoDB: :autoshrink freed the segment .* in page .*;
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+--source include/search_pattern_in_file.inc
+let $restart_parameters=;
+set global innodb_fast_shutdown=0;
+--source include/restart_mysqld.inc
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
--- a/storage/innobase/dict/dict0load.cc
+++ b/storage/innobase/dict/dict0load.cc
@@ -172,36 +172,24 @@ name_of_col_is(
 }
 #endif /* UNIV_DEBUG */
 
-/********************************************************************//**
-This function gets the next system table record as it scans the table.
-@return the next record if found, NULL if end of scan */
-static
 const rec_t*
-dict_getnext_system_low(
-/*====================*/
-	btr_pcur_t*	pcur,		/*!< in/out: persistent cursor to the
-					record*/
-	mtr_t*		mtr)		/*!< in: the mini-transaction */
+dict_getnext_system_low(btr_pcur_t *pcur, mtr_t *mtr)
 {
-	rec_t*	rec = NULL;
-
-	while (!rec) {
-		btr_pcur_move_to_next_user_rec(pcur, mtr);
-
-		rec = btr_pcur_get_rec(pcur);
-
-		if (!btr_pcur_is_on_user_rec(pcur)) {
-			/* end of index */
-			btr_pcur_close(pcur);
-
-			return(NULL);
-		}
-	}
-
-	/* Get a record, let's save the position */
-	btr_pcur_store_position(pcur, mtr);
-
-	return(rec);
+  rec_t *rec = nullptr;
+  while (!rec)
+  {
+    btr_pcur_move_to_next_user_rec(pcur, mtr);
+    rec = btr_pcur_get_rec(pcur);
+    if (!btr_pcur_is_on_user_rec(pcur))
+    {
+      /* end of index */
+      btr_pcur_close(pcur);
+      return nullptr;
+    }
+  }
+  /* Get a record, let's save the position */
+  btr_pcur_store_position(pcur, mtr);
+  return rec;
 }
 
 /********************************************************************//**
--- a/storage/innobase/fsp/fsp0fsp.cc
+++ b/storage/innobase/fsp/fsp0fsp.cc
@@ -25,11 +25,8 @@ Created 11/29/1995 Heikki Tuuri
 ***********************************************************************/
 
 #include "fsp0fsp.h"
-#include "buf0buf.h"
-#include "fil0fil.h"
 #include "fil0crypt.h"
 #include "mtr0log.h"
-#include "ut0byte.h"
 #include "page0page.h"
 #include "srv0srv.h"
 #include "srv0start.h"
@@ -37,12 +34,16 @@ Created 11/29/1995 Heikki Tuuri
 #include "btr0sea.h"
 #include "dict0boot.h"
 #include "log0log.h"
+#include "dict0load.h"
 #include "dict0mem.h"
-#include "fsp0types.h"
+#include "btr0pcur.h"
+#include "trx0sys.h"
 #include "log.h"
 #ifndef DBUG_OFF
 # include "trx0purge.h"
 #endif
+#include <unordered_set>
+#include "trx0undo.h"
 
 /** Returns the first extent descriptor for a segment.
 We think of the extent lists of the segment catenated in the order
@@ -1477,19 +1478,19 @@ MY_ATTRIBUTE((nonnull))
 @param[in,out]	inode		segment inode
 @param[in,out]	iblock		segment inode page
 @param[in,out]	mtr		mini-transaction */
-static void fsp_free_seg_inode(fil_space_t *space, fseg_inode_t *inode,
-                               buf_block_t *iblock, mtr_t *mtr)
+static dberr_t fsp_free_seg_inode(fil_space_t *space, fseg_inode_t *inode,
+                                  buf_block_t *iblock, mtr_t *mtr)
 {
   ut_d(space->modify_check(*mtr));
 
   dberr_t err;
   buf_block_t *header= fsp_get_header(space, mtr, &err);
   if (!header)
-    return;
+    return err;
   if (UNIV_UNLIKELY(memcmp(FSEG_MAGIC_N_BYTES, FSEG_MAGIC_N + inode, 4)))
   {
     space->set_corrupted();
-    return;
+    return DB_CORRUPTION;
   }
 
   const ulint physical_size= space->physical_size();
@@ -1499,24 +1500,27 @@ static void fsp_free_seg_inode(fil_space_t *space, fseg_inode_t *inode,
                                                       physical_size))
   {
     /* Move the page to another list */
-    if (flst_remove(header, FSP_HEADER_OFFSET + FSP_SEG_INODES_FULL,
-                    iblock, FSEG_INODE_PAGE_NODE, limit, mtr) != DB_SUCCESS)
-      return;
-    if (flst_add_last(header, FSP_HEADER_OFFSET + FSP_SEG_INODES_FREE,
-                      iblock, FSEG_INODE_PAGE_NODE, limit, mtr) != DB_SUCCESS)
-      return;
+    err= flst_remove(header, FSP_HEADER_OFFSET + FSP_SEG_INODES_FULL,
+                     iblock, FSEG_INODE_PAGE_NODE, limit, mtr);
+    if (err == DB_SUCCESS)
+      err= flst_add_last(header, FSP_HEADER_OFFSET + FSP_SEG_INODES_FREE,
+                         iblock, FSEG_INODE_PAGE_NODE, limit, mtr);
+    if (err)
+      return err;
   }
 
   mtr->memset(iblock, page_offset(inode) + FSEG_ID, FSEG_INODE_SIZE, 0);
 
   if (ULINT_UNDEFINED != fsp_seg_inode_page_find_used(iblock->page.frame,
                                                       physical_size))
-    return;
+    return DB_SUCCESS;
 
   /* There are no other used headers left on the page: free it */
-  if (flst_remove(header, FSP_HEADER_OFFSET + FSP_SEG_INODES_FREE,
-                  iblock, FSEG_INODE_PAGE_NODE, limit, mtr) == DB_SUCCESS)
-    fsp_free_page(space, iblock->page.id().page_no(), mtr);
+  err= flst_remove(header, FSP_HEADER_OFFSET + FSP_SEG_INODES_FREE,
+                   iblock, FSEG_INODE_PAGE_NODE, limit, mtr);
+  if (err != DB_SUCCESS)
+    return err;
+  return fsp_free_page(space, iblock->page.id().page_no(), mtr);
 }
 
 MY_ATTRIBUTE((nonnull(1,4,5), warn_unused_result))
@@ -2510,20 +2514,21 @@ try_to_extend:
 
 MY_ATTRIBUTE((nonnull, warn_unused_result))
 /** Frees a single page of a segment.
-@param[in]	seg_inode	segment inode
 @param[in,out]	space		tablespace
-@param[in]	offset		page number
+@param[in]	seg_inode	segment inode
+@param[in,out]	iblock		block where segment inode are kept
 @param[in,out]	mtr		mini-transaction
+@param[in]	offset		page number
 @param[in]	ahi		Drop adaptive hash index
 @return error code */
 static
 dberr_t
 fseg_free_page_low(
+	fil_space_t*		space,
 	fseg_inode_t*		seg_inode,
 	buf_block_t*		iblock,
-	fil_space_t*		space,
-	uint32_t		offset,
-	mtr_t*			mtr
+	mtr_t*			mtr,
+	uint32_t		offset
 #ifdef BTR_CUR_HASH_ADAPT
 	,bool			ahi=false
 #endif /* BTR_CUR_HASH_ADAPT */
@@ -2656,7 +2661,7 @@ dberr_t fseg_free_page(fseg_header_t *seg_header, fil_space_t *space,
   {
     if (!space->full_crc32())
       fil_block_check_type(*iblock, FIL_PAGE_INODE, mtr);
-    return fseg_free_page_low(seg_inode, iblock, space, offset, mtr);
+    return fseg_free_page_low(space, seg_inode, iblock, mtr, offset);
   }
 
   return err;
@@ -2812,6 +2817,71 @@ remove:
 	return DB_SUCCESS;
 }
 
+/** Free the extent and fragment page associated with
+the segment.
+@param space       tablespace where segment resides
+@param inode       index node information
+@param iblock      page where segment header are placed
+@param mtr         mini-transaction
+@param hdr_page_no segment header page number field
+@param ahi         adaptive hash index
+@return DB_SUCCESS_LOCKED_REC when freeing wasn't completed
+@return DB_SUCCESS or other error code  when freeing was completed */
+static
+dberr_t fseg_free_step_low(fil_space_t *space, fseg_inode_t *inode,
+                           buf_block_t *iblock, mtr_t *mtr,
+                           const byte *hdr_page_no
+#ifdef BTR_CUR_HASH_ADAPT
+                           , bool ahi=false
+#endif /* BTR_CUR_HASH_ADAPT */
+                           )
+{
+  dberr_t err= DB_SUCCESS;
+  if (xdes_t *descr= fseg_get_first_extent(inode, space, mtr, &err))
+  {
+    err= fseg_free_extent(inode, iblock, space,
+                          xdes_get_offset(descr), mtr
+#ifdef BTR_CUR_HASH_ADAPT
+                          , ahi
+#endif /* BTR_CUR_HASH_ADAPT */
+                        );
+    return err == DB_SUCCESS ? DB_SUCCESS_LOCKED_REC : err;
+  }
+
+  if (err != DB_SUCCESS)
+    return err;
+
+  /* Free a fragment page. If there are no fragment pages
+  exist in the array then free the file segment inode */
+  ulint n = fseg_find_last_used_frag_page_slot(inode);
+  if (UNIV_UNLIKELY(n == ULINT_UNDEFINED))
+    return hdr_page_no
+      ? DB_SUCCESS
+      : fsp_free_seg_inode(space, inode, iblock, mtr);
+
+  if (hdr_page_no && !memcmp_aligned<2>(hdr_page_no + FIL_PAGE_OFFSET,
+			                inode + FSEG_FRAG_ARR +
+                                        n * FSEG_FRAG_SLOT_SIZE, 4))
+    /* header_page_no is only passed by fseg_free_step_not_header().
+    In that case, the header page must be preserved, to be freed
+    when we're finally called by fseg_free_step(). */
+    return DB_SUCCESS;
+
+  uint32_t page_no= fseg_get_nth_frag_page_no(inode, n);
+  err= fseg_free_page_low(space, inode, iblock, mtr, page_no
+#ifdef BTR_CUR_HASH_ADAPT
+                          , ahi
+#endif /* BTR_CUR_HASH_ADAPT */
+                          );
+  if (err != DB_SUCCESS)
+    return err;
+  buf_page_free(space, page_no, mtr);
+  if (!hdr_page_no &&
+      fseg_find_last_used_frag_page_slot(inode) == ULINT_UNDEFINED)
+    return fsp_free_seg_inode(space, inode, iblock, mtr);
+  return DB_SUCCESS_LOCKED_REC;
+}
+
 /** Frees part of a segment. This function can be used to free
 a segment by repeatedly calling this function in different
 mini-transactions. Doing the freeing in a single mini-transaction
@@ -2831,7 +2901,6 @@ fseg_free_step(
 #endif /* BTR_CUR_HASH_ADAPT */
 	)
 {
-	ulint		n;
 	fseg_inode_t*	inode;
 
 	const uint32_t space_id = page_get_space_id(page_align(header));
@@ -2864,54 +2933,11 @@ fseg_free_step(
 		fil_block_check_type(*iblock, FIL_PAGE_INODE, mtr);
 	}
 
-	dberr_t err;
-	descr = fseg_get_first_extent(inode, space, mtr, &err);
-
-	if (descr) {
-		/* Free the extent held by the segment */
-		return fseg_free_extent(inode, iblock, space,
-					xdes_get_offset(descr), mtr
+	return fseg_free_step_low(space, inode, iblock, mtr, nullptr
 #ifdef BTR_CUR_HASH_ADAPT
-					, ahi
+				  , ahi
 #endif /* BTR_CUR_HASH_ADAPT */
-					) != DB_SUCCESS;
-	}
-
-	if (err != DB_SUCCESS || space->is_stopping()) {
-		return true;
-	}
-
-	/* Free a frag page */
-	n = fseg_find_last_used_frag_page_slot(inode);
-
-	if (n == ULINT_UNDEFINED) {
-		/* Freeing completed: free the segment inode */
-		fsp_free_seg_inode(space, inode, iblock, mtr);
-		return true;
-	}
-
-	uint32_t page_no = fseg_get_nth_frag_page_no(inode, n);
-
-	if (fseg_free_page_low(inode, iblock, space, page_no, mtr
-#ifdef BTR_CUR_HASH_ADAPT
-			       , ahi
-#endif /* BTR_CUR_HASH_ADAPT */
-			       ) != DB_SUCCESS) {
-		return true;
-	}
-
-	buf_page_free(space, page_no, mtr);
-
-	n = fseg_find_last_used_frag_page_slot(inode);
-
-	if (n == ULINT_UNDEFINED) {
-		/* Freeing completed: free the segment inode */
-		fsp_free_seg_inode(space, inode, iblock, mtr);
-
-		return true;
-	}
-
-	return false;
+				  ) != DB_SUCCESS_LOCKED_REC;
 }
 
 bool
@@ -2923,16 +2949,16 @@ fseg_free_step_not_header(
 #endif /* BTR_CUR_HASH_ADAPT */
 	)
 {
-	fseg_inode_t*	inode;
-
-	const uint32_t space_id = page_get_space_id(page_align(header));
+	const page_t *page = page_align(header);
+	const uint32_t space_id = page_get_space_id(page);
 	ut_ad(mtr->is_named_space(space_id));
 
 	fil_space_t*		space = mtr->x_lock_space(space_id);
 	buf_block_t*		iblock;
 
-	inode = fseg_inode_try_get(header, space_id, space->zip_size(),
-				   mtr, &iblock);
+	fseg_inode_t *inode = fseg_inode_try_get(header, space_id,
+						 space->zip_size(),
+						 mtr, &iblock);
 	if (space->is_stopping()) {
 		return true;
 	}
@@ -2940,7 +2966,7 @@ fseg_free_step_not_header(
 	if (!inode) {
 		ib::warn() << "Double free of "
 			   << page_id_t(space_id,
-					page_get_page_no(page_align(header)));
+					page_get_page_no(page));
 		return true;
 	}
 
@@ -2948,43 +2974,11 @@ fseg_free_step_not_header(
 		fil_block_check_type(*iblock, FIL_PAGE_INODE, mtr);
 	}
 
-	dberr_t err;
-	if (xdes_t* descr = fseg_get_first_extent(inode, space, mtr, &err)) {
-		/* Free the extent held by the segment */
-		return fseg_free_extent(inode, iblock, space,
-					xdes_get_offset(descr),
-					mtr
+	return fseg_free_step_low(space, inode, iblock, mtr, page
 #ifdef BTR_CUR_HASH_ADAPT
-					, ahi
+				  , ahi
 #endif /* BTR_CUR_HASH_ADAPT */
-					) != DB_SUCCESS;
-	} else if (err != DB_SUCCESS) {
-		return true;
-	}
-
-	/* Free a frag page */
-
-	ulint n = fseg_find_last_used_frag_page_slot(inode);
-
-	if (UNIV_UNLIKELY(n == ULINT_UNDEFINED)) {
-		return true;
-	}
-
-	uint32_t page_no = fseg_get_nth_frag_page_no(inode, n);
-
-	if (page_no == page_get_page_no(page_align(header))) {
-		return true;
-	}
-
-	if (fseg_free_page_low(inode, iblock, space, page_no, mtr
-#ifdef BTR_CUR_HASH_ADAPT
-			       , ahi
-#endif /* BTR_CUR_HASH_ADAPT */
-			       ) != DB_SUCCESS) {
-		return true;
-	}
-	buf_page_free(space, page_no, mtr);
-	return false;
+				  ) != DB_SUCCESS_LOCKED_REC;
 }
 
 /** Returns the first extent descriptor for a segment.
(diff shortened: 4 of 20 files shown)


storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3574,14 +3599,138 @@ dberr_t fsp_sys_tablespace_validate()
...
+
 void fsp_system_tablespace_truncate()
 {
+  dberr_t err= DB_SUCCESS;
+  uint32_t max_used_extent= 0;
+  if (srv_sys_shrink_extensive)
+  {
+    std::vector<uint32_t> sys_root_pages;
+    err= dict_get_sys_tablespace_indexes(sys_root_pages);
+    if (err) return;
```
dr-m, line 3722: There is no diagnostic in case an error was reported, and no suggestion how to resolve the corruption (such as displaying the name of the problematic index and table that had been created with `innodb_file_per_table=0`). We wouldn’t even attempting a normal truncation in this case.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3574,14 +3599,138 @@ dberr_t fsp_sys_tablespace_validate()
...
+
+    max_used_extent= std::max(max_used_extent, buf_dblwr.block2_page());
+    fsp_get_max_sys_used_index_extent(sys_root_pages, max_used_extent);
```
dr-m, line 3725: I thought that extents and page numbers are not the same. It would be good to add some comments here.

What about the indexes of the hard-coded system tables `SYS_TABLES`, `SYS_COLUMNS`, `SYS_INDEXES` and `SYS_FIELDS`? Should they be included in `sys_root_pages`?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -19307,6 +19307,12 @@ static MYSQL_SYSVAR_BOOL(optimize_fulltext_only, innodb_optimize_fulltext_only,
   "Only optimize the Fulltext index of the table",
   NULL, NULL, FALSE);
 
+static MYSQL_SYSVAR_BOOL(sys_tablespace_shrink_extensive,
+  srv_sys_shrink_extensive,
+  PLUGIN_VAR_OPCMDARG,
+  "Shrink the system tablespace by doing extensive check",
+  NULL, NULL, TRUE);
```
dr-m, line 19314: Instead of repeating the word `extensive`, can we try to explain the effect of setting this option? Such as: "Mark any pages in the system tablespace free that do not belong to any table that resides in the system tablespace. Has no effect if innodb_undo_tablespaces<2."

This option should really be disabled by default, in order to avoid changing any default behaviour after a GA release.

storage/innobase/include/dict0load.h
```diff
@@ -213,4 +213,6 @@ dict_process_sys_foreign_col_rec(
 					in referenced table */
 	ulint*		pos);		/*!< out: column position */
 
+dberr_t dict_get_sys_tablespace_indexes(
+  std::vector<uint32_t> &sys_root_pages);
```
dr-m, line 217: Can we define this function `static` in the compilation unit that needs it? Maybe there should be a separate compilation unit for system tablespace shrinking? All those functions could be declared `ATTRIBUTE_COLD`.

dr-m (commented):
As far as I understand it, the current version of this change is tweaking the shrinking of the system tablespace so that any pages that do not belong to index trees that are declared in `SYS_INDEXES` will be ignored when determining the maximum allocated page number within the system tablespace.

This seems to be both wrong and insufficient.

1. Some pages of `SYS_TABLES`, `SYS_INDEXES`, `SYS_COLUMNS` and `SYS_FIELDS` could actually be stored in an area that the current patch would deem safe for truncation.
2. There could also be other objects that are allocated in the system tablespace, such as undo log pages that are reachable via the `TRX_SYS` page. (I think that the `TRX_SYS` page itself as well as the empty change buffer tree would not cause a problem.)
3. We fail to mark any leaked pages that are below the tweaked upper limit as free.

I think that for development and debugging, it would be good to implement something that pretty-prints the allocated page bitmap, possibly using different characters for representing allocated pages in different roles. For example, `'0' + (index_id & 63)`` for index pages, and some characters in the 0x21 to 0x2f range for anything else. And 0x20 (space) for pages that are marked as free. This should nicely illustrate the page leaks due to MDEV-31234 and allow us to understand if we have missed anything.

I think that it would be better to implement an isolated step that would mark as freed any leaked pages in the system tablespace. It could be independent of the shrinking logic.

storage/innobase/dict/dict0load.cc
```diff
@@ -3224,3 +3224,55 @@ dict_load_foreigns(
...
+  std::vector<uint32_t> *root_pages)
+{
+  /* Add system table index root pages */
+  dict_sys.sys_tables->add_index_root_pages(root_pages);
+  dict_sys.sys_indexes->add_index_root_pages(root_pages);
+  dict_sys.sys_columns->add_index_root_pages(root_pages);
+  dict_sys.sys_fields->add_index_root_pages(root_pages);
+  dict_sys.sys_foreign->add_index_root_pages(root_pages);
+  dict_sys.sys_foreign_cols->add_index_root_pages(root_pages);
+  dict_sys.sys_virtual->add_index_root_pages(root_pages);
```
dr-m, line 3241: These should be added by the `dict_getnext_system_low()` loop below. Is that not the case?

storage/innobase/dict/dict0load.cc
```diff
@@ -3224,3 +3224,55 @@ dict_load_foreigns(
...
+    field= rec_get_nth_field_old(
+      rec, DICT_FLD__SYS_INDEXES__PAGE_NO, &len);
+    if (len != 4)
+    {
+      err= DB_CORRUPTION;
+      break;
+    }
+    uint32_t root_page_no= mach_read_from_4(field);
+    if (root_page_no != FIL_NULL)
+      root_pages->push_back(root_page_no);
```
dr-m, line 3272: What if there are duplicate `SYS_INDEXES.PAGE` entries? Should `root_pages` rather be a `std::unordered_set`?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -1255,9 +1260,17 @@ static dberr_t fsp_free_page(fil_space_t *space, uint32_t offset, mtr_t *mtr)
...
+		} else if (ignore_freed_page) {
+			return err;
 		}
+	}
+		/* fall through */
+	case XDES_FREE:
+		if (ignore_freed_page) return err;
 		/* fall through */
 	default:
 		space->set_corrupted();
```
dr-m, line 1276: Could we tweak the rarely invoked function `fil_space_t::set_corrupted()` instead of adding a parameter and complex logic to a rather commonly invoked operation?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -1647,18 +1660,20 @@ fseg_find_last_used_frag_page_slot(
 /** Calculate reserved fragment page slots.
 @param inode  file segment index
 @return number of fragment pages */
-static ulint fseg_get_n_frag_pages(const fseg_inode_t *inode)
+static ulint fseg_get_n_frag_pages(
+  const fseg_inode_t *inode, range_set *page_ranges= nullptr)
```
dr-m, line 1664: Instead of adding a parameter to an existing rather commonly used function, I think that it would be better to add a separate function for the rare case that `page_ranges` is specified. That parameter also needs to be documented in a function comment.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3574,6 +3589,294 @@ dberr_t fsp_sys_tablespace_validate()
...
+  const uint32_t len= flst_get_len(base->page.frame + boffset);
+  fil_addr_t addr= flst_get_first(base->page.frame + boffset);
+  for (uint32_t i= len; i--; )
+  {
+    ut_ad(addr.boffset >= FIL_PAGE_DATA);
+    ut_ad(addr.boffset < base->physical_size() - FIL_PAGE_DATA_END);
+    const buf_block_t *b=
+      buf_page_get_gen(page_id_t(base->page.id().space(), addr.page),
+                       base->zip_size(), RW_SX_LATCH, nullptr, BUF_GET, mtr);
+    ut_ad(b);
```
dr-m, line 3610: An error from the function must be handled properly.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3574,6 +3589,294 @@ dberr_t fsp_sys_tablespace_validate()
...
+  dberr_t err= DB_SUCCESS;
+  mtr_t mtr;
+  mtr.start();
+  for (uint32_t i= 0; i < space->free_limit; i++)
+  {
+    if (n_words == 16)
+    {
+      fprintf(stderr, "\n\t");
+      n_words= 0;
+    }
```
dr-m, line 3664: It could be good to display the page number in hexadecimal at the start of each line.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3574,6 +3589,294 @@ dberr_t fsp_sys_tablespace_validate()
...
+    if (!block || block->page.state() <= buf_page_t::FREED)
+    {
+      fprintf(stderr, "F");
+      err= DB_SUCCESS;
+      continue;
+    }
+    ulint type= fil_page_get_type(block->page.frame);
+    switch (type) {
+    case FIL_PAGE_INDEX:
+      fprintf(stderr, "I");
```
dr-m, line 3684: It would be good to add a comment that we can’t trust the `FIL_PAGE_TYPE` field unless `space->full_crc32()` holds. For other system tablespaces, the contents can be any garbage. The only guarantee is that if the page belongs to a B-tree, then the `FIL_PAGE_TYPE` will contain `FIL_PAGE_INDEX`. The opposite is _not_ guaranteed.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3574,6 +3589,294 @@ dberr_t fsp_sys_tablespace_validate()
...
+      break;
+    default:
+      ut_ad(0);
+    }
+  }
+  mtr.commit();
+  fprintf(stderr, "\nNote the letter description of the pages\n\t");
+  fprintf(stderr, "I - Index node page\n\t"
+                  "U - Undo log page\n\t"
+                  "Z - Compressed page\n\t"
```
dr-m, line 3730: There is nothing formatted in the output; therefore we could as well use `fputs` or `putc`.

I was hoping that we would display something like `'A' + ((index_id - 10) % ('~' - 'A' + 1))` for pages that look like index pages. That would help identify them. This would leave the ASCII range `' ' to '?'` (0x20 to 0x3f) available for displaying other leaked pages.

Compressed pages can’t exist in the system tablespace, neither `page_compressed` nor `ROW_FORMAT=COMPRESSED`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3574,6 +3589,294 @@ dberr_t fsp_sys_tablespace_validate()
...
+static
+dberr_t fsp_system_tablespace_used_pages(range_set *page_ranges)
+{
+  page_ranges->add_value_range(FSP_XDES_OFFSET, FSP_DICT_HDR_PAGE_NO + 1);
+  /* TRX_SYS page and Doublewrite buffer segment */
+  mtr_t mtr;
+  mtr.start();
+  buf_block_t *block= buf_page_get(page_id_t(0, TRX_SYS_PAGE_NO),
+                                   0, RW_S_LATCH, &mtr);
+  ut_ad(block);
```
dr-m, line 3757: We could very easily pass `&err` to the function and add error handling here.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -19792,6 +19786,20 @@ static MYSQL_SYSVAR_BOOL(encrypt_temporary_tables, innodb_encrypt_temporary_tabl
   "Enrypt the temporary table data.",
   NULL, NULL, false);
 
+static MYSQL_SYSVAR_BOOL(system_garbage_collect,
+  srv_sys_garbage_collect, 0,
+  "During slow shutdown, InnoDB does free the unused pages "
+  "in the tablespace", NULL, NULL, FALSE);
```
dr-m, line 19792: We are supposed to do this only on startup, right?

storage/innobase/dict/dict0load.cc
```diff
@@ -3224,3 +3224,55 @@ dict_load_foreigns(
...
+{
+  mtr_t mtr;
+  btr_pcur_t pcur;
+  ulint len;
+  dberr_t err= DB_SUCCESS;
+  mtr.start();
+  dict_sys.lock(SRW_LOCK_CALL);
+  for (const rec_t *rec= dict_startscan_system(
+         &pcur, &mtr, dict_sys.sys_indexes);
+       rec; rec= dict_getnext_system_low(&pcur, &mtr))
```
dr-m, line 3242: A shared latch (`freeze` instead of `lock`) should suffice.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,6 +3579,238 @@ dberr_t fsp_sys_tablespace_validate()
...
+  void insert(uint32_t page, uint16_t offset)
+  {
+    m_used_inodes[page].insert(offset);
+  }
+
+  void insert_seg(byte *inode)
+  {
+    m_used_inodes[mach_read_from_4(inode + 4)].insert(
+      mach_read_from_2(inode + 8));
+  }
```
dr-m, line 3598: Is there a reason why we use `insert` instead of `emplace`? Would `unordered_map` work instead of `map`?

In the common formatting style, lines should never be broken right after an opening parenthesis `(` or `[`. You can break the line after `.` for example.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,6 +3579,238 @@ dberr_t fsp_sys_tablespace_validate()
...
+  { return m_used_inodes.begin(); }
+  typename std::map<uint32_t, std::set<uint16_t>>::iterator end()
+  { return m_used_inodes.end(); }
+};
+
+/* Get the used inode from the system tablespace
+@param used_inode list of inode information used in
+                  system tablespace
+@return error code */
+dberr_t fsp_get_sys_used_segment(inodeInfo *usedInodes)
```
dr-m, line 3619: The `@param` does not match the parameter name.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,6 +3579,238 @@ dberr_t fsp_sys_tablespace_validate()
...
+ }
+
+ block= buf_page_get_gen(page_id_t{0, DICT_HDR_PAGE_NO}, 0,
+                         RW_S_LATCH, nullptr, BUF_GET, &mtr, &err);
+ if (!block)
+   goto func_exit;
+ usedInodes->insert_seg(
+   block->page.frame + DICT_HDR + DICT_HDR_FSEG_HEADER);
+ block= buf_page_get_gen(page_id_t{0, FSP_IBUF_HEADER_PAGE_NO}, 0,
+                         RW_S_LATCH, nullptr, BUF_GET, &mtr, &err);
```
dr-m, line 3662: Are we following the latching order here? It would be useful to add a comment that gives an example of a function that uses the same latching order.

If the server is in a kind of single-threaded mode (only this thread and some I/O handler threads) and if we are only reading here, then it could be better to use `buf_pool.page_fix()`; no need to hold multiple page latches to prevent concurrent modification.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,6 +3579,238 @@ dberr_t fsp_sys_tablespace_validate()
...
+/** Remove the unused segment in system tablespace
+@return error code */
+dberr_t fsp_garbage_collect()
+{
+  if (!trx_sys.is_undo_empty())
+  {
+    sql_print_warning("InnoDB: Can't free the unused segment "
+                      "in system tablespace due to undo log exists.");
+    return DB_ERROR;
+  }
```
dr-m, line 3773: Could we rename this function to `fil_space_t::garbage_collect()`? Invoked as `fil_system.sys_space->garbage_collect()`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,6 +3579,238 @@ dberr_t fsp_sys_tablespace_validate()
...
+   block->page.frame + DICT_HDR + DICT_HDR_FSEG_HEADER);
+ block= buf_page_get_gen(page_id_t{0, FSP_IBUF_HEADER_PAGE_NO}, 0,
+                         RW_S_LATCH, nullptr, BUF_GET, &mtr, &err);
+ if (!block)
+   goto func_exit;
+ usedInodes->insert_seg(block->page.frame + PAGE_DATA);
+
+ std::unordered_set<uint32_t> root_pages;
+ /* Get all index root page exists on system tablespace */
+ err= dict_get_sys_tablespace_indexes(&root_pages);
```
dr-m, line 3669: Do we really need the additional collection `root_pages`, or could we extend `usedInodes` directly?

dr-m (commented):
Some quick comments. I don’t think that an additional configuration parameter for enabling this is absolutely necessary.

storage/innobase/fsp/fsp0fsp.cc
dr-m: Besides, could the function that registers the `BTR_SEG_LEAF` and `BTR_SEG_TOP` for known indexes use the `mtr` from this function? There would seem to be no benefit to create multiple `mtr` objects.

storage/innobase/fsp/fsp0fsp.cc
dr-m: Shouldn’t we report an error if something had already been registered in `m_used_inodes`? That is, if multiple entries point to the same file segment, we should treat the system tablespace as corrupted and refuse to free any file segments or to shrink the system tablespace.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,16 +3579,326 @@ dberr_t fsp_sys_tablespace_validate()
...
+    field= rec_get_nth_field_old(
+      rec, DICT_FLD__SYS_INDEXES__PAGE_NO, &len);
+    if (len != 4)
+    {
+      err= DB_CORRUPTION;
+      break;
+    }
+    uint32_t root_page_no= mach_read_from_4(field);
+    if (root_page_no != FIL_NULL)
+      root_pages.insert(root_page_no);
```
dr-m, line 3638: We are ignoring duplicate `root_pages` here. Can we avoid introducing a local variable `root_pages` and update `inodes` directly? It would seem to require a further function to be invoked here:
```c++
  uint32_t root= mach_reaad_from_4(field);
  if (root != FIL_NULL && (err= fsp_table_inodes_root(mtr, root) != DB_SUCCESS)
    break;
```

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,16 +3579,326 @@ dberr_t fsp_sys_tablespace_validate()
...
+  }
+  mtr->rollback_to_savepoint(savepoint);
+  dict_sys.unfreeze();
+
+  if (err) return err;
+  /* Add system table index root pages */
+  dict_sys.sys_tables->add_index_root_pages(&root_pages);
+  dict_sys.sys_indexes->add_index_root_pages(&root_pages);
```
dr-m, line 3646: Each of these should invoke the new function `fsp_table_indexes_root()`.

storage/innobase/include/dict0mem.h
```diff
@@ -2446,6 +2447,16 @@ struct dict_table_t {
...
+
+  /** Iterate through indexes and store their root page
+  in the given list.
+  @param root_pages root page list */
+  void add_index_root_pages(std::unordered_set<uint32_t> *root_pages)
+  {
+    for (auto i= UT_LIST_GET_FIRST(indexes);
+         i != nullptr; i= UT_LIST_GET_NEXT(indexes, i))
+      root_pages->insert(i->page);
+  }
```
dr-m, line 2459: There is only one user of this function. I think that this can be defined as a `static` non-member function in the compilation unit that needs it. I would not declare it as `inline` because there will be many calls.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,16 +3579,326 @@ dberr_t fsp_sys_tablespace_validate()
...
+ }
+ return err;
+}
+
+/* Get the used inode from the system tablespace
+@param inodes  inode information used found in system tablespace
+@return error code */
+dberr_t fsp_get_sys_used_segment(inode_info *inodes)
+{
+  mtr_t mtr;
```
dr-m, line 3683: This needs to be declared `static` and take also `mtr_t *mtr` as a parameter.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,16 +3579,326 @@ dberr_t fsp_sys_tablespace_validate()
...
+  TRX_SYS_PAGE */
+  block= buf_page_get_gen(
+    page_id_t{0, TRX_SYS_PAGE_NO}, 0, RW_NO_LATCH, nullptr,
+    BUF_GET, &mtr, &err);
+  if (!block)
+  {
+err_exit:
+    mtr.commit();
+    return err;
+  }
```
dr-m, line 3697: If we took `mtr` as a parameter, each `goto err_exit;` would reduce to `return err;`.

I think that it could be even better to use `buf_pool.page_fix()` to read the pages. I would like to get rid of the `RW_NO_LATCH` parameter of `buf_page_get_gen()` and use that function exclusively for latched block access. In that case we would have to remember to invoke `block->unfix()` after we are done with a block.

dr-m (commented):
The logic looks good to me. My comments are mainly about coding style or API efficiency.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,339 @@ dberr_t fsp_sys_tablespace_validate()
...
+public:
+  __attribute__((warn_unused_result))
+  /** Register an inode */
+  bool insert_seg(const byte *inode)
+  {
+    uint32_t page= mach_read_from_4(inode + 4);
+    uint32_t offset= mach_read_from_2(inode + 8);
+    return page >= fil_system.sys_space->free_limit ||
+      offset < FIL_PAGE_DATA || offset >= srv_page_size - FIL_PAGE_DATA_END ||
+      emplace(uint64_t{page} << 16 | offset).second;
```
dr-m, line 3596: The return value is not documented. According to this [description of std::unordered_set::emplace()](https://cplusplus.com/reference/unordered_set/unordered_set/emplace/), the `.second` component is `true` when the element to insert did not previously exist. So, the "successful" return value ought to be `true`. But the callers currently seem to expect `false`!

The conditions before `||` seem to be backwards. We would seem to want the following:
```c++
    return page < fil_system.sys_space->free_limit &&
      offset >= FIL_PAGE_DATA && offset < srv_page_size - FIL_PAGE_DATA_END &&
      emplace(uint64_t{page} << 16 | offset).second;
```

dr-m (changes requested):

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,339 @@ dberr_t fsp_sys_tablespace_validate()
...
+  if (!block)
+    return err;
+
+  if (!inodes->insert_seg(block->page.frame + PAGE_HEADER + PAGE_BTR_SEG_TOP))
+    return DB_CORRUPTION;
+
+  if (!inodes->insert_seg(block->page.frame + PAGE_HEADER + PAGE_BTR_SEG_LEAF))
+    return DB_CORRUPTION;
+  block->page.unfix();
+  return err;
```
dr-m, line 3625: The two `return` must be replaced with `err= DB_CORRUPTION` assignments. We must not leave the page fixed forever.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,339 @@ dberr_t fsp_sys_tablespace_validate()
...
+@param mtr    mini-transaction
+@return error code */
+static dberr_t fsp_table_inodes(inode_info *inodes, mtr_t *mtr)
+{
+  btr_pcur_t pcur;
+  ulint len;
+  const auto savepoint= mtr->get_savepoint();
+  dberr_t err= DB_SUCCESS;
+  dict_sys.freeze(SRW_LOCK_CALL);
+  std::unordered_set<uint32_t> root_pages;
```
dr-m, line 3656: The variable `root_pages` is unused.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,339 @@ dberr_t fsp_sys_tablespace_validate()
...
+    return err;
+
+  if (!inodes->insert_seg(block->page.frame + TRX_SYS + TRX_SYS_FSEG_HEADER))
+    return DB_CORRUPTION;
+
+  if (!inodes->insert_seg(block->page.frame + TRX_SYS_DOUBLEWRITE
+                          + TRX_SYS_DOUBLEWRITE_FSEG))
+    return DB_CORRUPTION;
+
+  block->page.unfix();
```
dr-m, line 3719: The `unfix()` must not be skipped.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,339 @@ dberr_t fsp_sys_tablespace_validate()
...
+  block= buf_pool.page_fix(page_id_t{0, DICT_HDR_PAGE_NO}, &err,
+                           buf_pool_t::FIX_WAIT_READ);
+  if (!block)
+    return err;
+
+  if (!inodes->insert_seg(block->page.frame + DICT_HDR + DICT_HDR_FSEG_HEADER))
+    return DB_CORRUPTION;
+
+  block->page.unfix();
```
dr-m, line 3728: Do not skip the `unfix()`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,339 @@ dberr_t fsp_sys_tablespace_validate()
...
+  block= buf_pool.page_fix(page_id_t{0, FSP_IBUF_HEADER_PAGE_NO},
+                           &err, buf_pool_t::FIX_WAIT_READ);
+  if (!block)
+    return err;
+  if (!inodes->insert_seg(block->page.frame + PAGE_DATA))
+    return DB_CORRUPTION;
+
+  block->page.unfix();
```
dr-m, line 3736: Do not skip `unfix()`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,339 @@ dberr_t fsp_sys_tablespace_validate()
...
+    trx_rseg_t *rseg= &trx_sys.rseg_array[rseg_id];
+    if (rseg->space->id == 0)
+    {
+      block= buf_pool.page_fix(rseg->page_id(), &err,
+                               buf_pool_t::FIX_WAIT_READ);
+      if (!block) return err;
+      if (!inodes->insert_seg(block->page.frame + TRX_RSEG +
+                              TRX_RSEG_FSEG_HEADER))
+        return DB_CORRUPTION;
+      block->page.unfix();
```
dr-m, line 3749: Do not skip `unfix()`.

storage/innobase/srv/srv0start.cc
```diff
@@ -347,6 +355,12 @@ inline dberr_t trx_sys_t::reset_page(mtr_t *mtr)
   mtr->write<2>(*sys_header, FIL_PAGE_TYPE + sys_header->page.frame,
                 FIL_PAGE_TYPE_TRX_SYS);
 
+  mtr->write<4>(*sys_header,
+		TRX_SYS + TRX_SYS_FSEG_HEADER + 4 +
+                sys_header->page.frame, trx_fseg_addr.page);
+  mtr->write<2>(*sys_header,
+		TRX_SYS + TRX_SYS_FSEG_HEADER + 8 +
+                sys_header->page.frame, trx_fseg_addr.boffset);
   mtr->write<4>(*sys_header,
```
dr-m, line 364: Please do not use TAB for indentation.

We should use one of the `mtr_t::memcpy()` functions to write these 6 consequent bytes, instead of writing two `WRITE` records.

storage/innobase/srv/srv0start.cc
```diff
@@ -333,6 +333,14 @@ inline dberr_t trx_sys_t::reset_page(mtr_t *mtr)
 
   if (!sys_header) return err;
 
+  ut_a(mach_read_from_4(sys_header->page.frame + TRX_SYS + TRX_SYS_FSEG_HEADER) == 0);
+
+  /* Store the TRX_SYS_FSEG_HEADER page, offset */
+  fil_addr_t trx_fseg_addr{mach_read_from_4(sys_header->page.frame +
+                                            TRX_SYS + TRX_SYS_FSEG_HEADER + 4),
+                           mach_read_from_2(sys_header->page.frame +
+                                            TRX_SYS + TRX_SYS_FSEG_HEADER + 8)};
```
dr-m, line 342: Could we just allocate `char fseg_addr[6]` and `memcpy(fseg_addr, sys_header->page.frame + …, 6);` to it? We do not seem to need the decoded `trx_fseg_addr` anywhere; we’re only writing it back to the page.

storage/innobase/srv/srv0start.cc
```diff
@@ -1781,11 +1795,14 @@ dberr_t srv_start(bool create_new_db)
...
-			fsp_system_tablespace_truncate();
-			DBUG_EXECUTE_IF("crash_after_sys_truncate",
+		if (!high_level_read_only) {
+
+			if (srv_sys_space.can_auto_shrink()) {
+				fsp_system_tablespace_truncate(false);
+				DBUG_EXECUTE_IF(
+					"crash_after_sys_truncate",
 					return srv_init_abort(DB_ERROR););
+			}
```
dr-m, line 1805: This looks like a nonfunctional change, replacing short-circuit evaluation with two nested `if`. Am I right that this hunk could be reverted without having any observable effect?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,347 @@ dberr_t fsp_sys_tablespace_validate()
...
-  if (err != DB_SUCCESS)
+  dberr_t err= space->garbage_collect(&mtr, shutdown);
+  if (err)
   {
 func_exit:
-    sql_print_warning("InnoDB: Cannot shrink the system tablespace "
-                      "due to %s", ut_strerr(err));
+    sql_print_warning(
+      "InnoDB: Cannot shrink the system tablespace due to %s",
+      ut_strerr(err));
```
dr-m, line 3916: I think that this white-space-only change had better be reverted.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,347 @@ dberr_t fsp_sys_tablespace_validate()
...
+    if (rseg->space->id == id)
+    {
+      buf_block_t *block= buf_page_get_gen(rseg->page_id(), 0,
+                                           RW_X_LATCH, nullptr,
+                                           BUF_GET, mtr, &err);
+      if (!block)
+        return err;
+      mtr->memset(block,
+                  TRX_RSEG + TRX_RSEG_UNDO_SLOTS,
+                  TRX_RSEG_N_SLOTS * TRX_RSEG_SLOT_SIZE, 0xff);
```
dr-m, line 3896: Can you add a comment that says which existing function this is mimicing?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,347 @@ dberr_t fsp_sys_tablespace_validate()
...
+  for (ulint rseg_id= 0; rseg_id < TRX_SYS_N_RSEGS; rseg_id++)
+  {
+    trx_rseg_t *rseg= &trx_sys.rseg_array[rseg_id];
+    if (rseg->space->id == id)
+    {
+      buf_block_t *block= buf_page_get_gen(rseg->page_id(), 0,
+                                           RW_X_LATCH, nullptr,
+                                           BUF_GET, mtr, &err);
+      if (!block)
+        return err;
```
dr-m, line 3893: Could we latch all these pages before making any changes? In that way, we would not do anything irreversible before trying to make sure that the operation can succeed?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,24 +3580,347 @@ dberr_t fsp_sys_tablespace_validate()
...
+    sql_print_warning(
+      "InnoDB: Cannot shrink the system tablespace due to %s",
+      ut_strerr(err));
     mtr.commit();
     return;
   }
+
+  err= fsp_traverse_extents(space, &last_used_extent, &mtr);
+  if (err != DB_SUCCESS)
+    goto func_exit;
```
dr-m, line 3923: `space->garbage_collect()` already changed some things. Can we invoke `mtr.commit()` and `mtr.start()` before this step? This is about actually shrinking the system tablespace file size, right? If it is about something else, then could we do it as part of `garbage_collect()`? The goal should be to first ensure that the operation is not going to fail, before starting to modify things. There is no rollback, after all.

storage/innobase/include/dict0load.h
```diff
@@ -213,4 +213,11 @@ dict_process_sys_foreign_col_rec(
 					in referenced table */
 	ulint*		pos);		/*!< out: column position */
 
+/** This function gets the next system table record as it scans
+the table.
+@param pcur persistent cursor
+@param mtr  mini-transaction
+@return the next record if found, NULL if end of scan */
```
dr-m, line 220: `@retval nullptr at the end of the table`

dr-m (commented):
Thank you, this looks mostly OK. I have a few nitpicks.

storage/innobase/fsp/fsp0fsp.cc
Thirunarayanan: This change is wrong. we shouldn't reset the rseg undo slots. Because undo object has already constructed
before this function. If we do that, we need to remove the respective undo objects. I think that is un-necessary.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3196,6 +3200,68 @@ class fsp_xdes_old_page
...
+      buf_page_get_gen(page_id_t{0, page_no}, 0, RW_SX_LATCH,
+                       nullptr, BUF_GET, mtr, &err);
+    if (!block)
+      return err;
+    buf_block_t *old= buf_LRU_get_free_block(have_no_mutex_soft);
+    if (!old) return DB_OUT_OF_MEMORY;
+
+    memcpy_aligned<UNIV_PAGE_SIZE_MIN>(
+      old->page.frame, block->page.frame, srv_page_size);
+    m_old_inode_entry[page_no]= old;
```
dr-m, line 3225: At the start of this function, the `operator[]` call already inserted a `nullptr` to the `std::map`. We could have saved a reference to that, e.g.,
```c++
buf_block_t *&old= m_old_inode_entry[page_no];
if (old)
  return DB_SUCCESS;
//…
old= buf_LRU_get_free_block(have_no_mutex_soft);
return old ? DB_OUT_OF_MEMORY : DB_SUCCESS;
```
However, I do not quite understand the need for storing a copy of the page. Logic like this would make sense when we are shrinking the system tablespace and we expect to free most pages, rewriting only a few surviving pages of the system tablespace. In that case we wanted to minimize the amount of redo log written, so that everything can be done in a single atomic mini-transaction, with at most 2 MiB of log records.

Here we are handling a special case and we are allowed to use as many mini-transactions as it takes.

Can we structure the mini-transactions in such a way that we free only one leaked page per mini-transaction, after performing all necessary consistency checks? In that way, there should be no need to allocate any shadow pages. Yes, we may be modifying the same pages over and over again, but the buffer pool will cache the write-backs. Yes, we may write a lot of log records, but this is a very rarely executed clean-up of corrupted data, and it will not be executed by default.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,23 +3643,408 @@ dberr_t fsp_sys_tablespace_validate()
...
+  xdes_t *descr= fseg_get_first_extent(inode, space, &mtr, &err);
+  while (descr && err == DB_SUCCESS)
+  {
+    old_inodes_entry.insert_descr(descr, &mtr);
+    err= fseg_free_extent(inode, iblock, space,
+                          xdes_get_offset(descr), &mtr);
+    DBUG_EXECUTE_IF("unused_undo_free_fail_4",
+                    err= DB_CORRUPTION;);
+    if (err)
+      goto err_exit;
```
dr-m, line 3887: Can we split `fseg_free_extent()` to two separate: checking the consistency, and modifying the data? We should first only check the consistency.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,23 +3643,408 @@ dberr_t fsp_sys_tablespace_validate()
...
+    uint32_t page_no = fseg_get_nth_frag_page_no(inode, i);
+    if (page_no != FIL_NULL)
+    {
+      err= fseg_free_page_low(inode, iblock, space, page_no, &mtr);
+      DBUG_EXECUTE_IF("unused_undo_free_fail_5",
+                      err= DB_CORRUPTION;);
+      if (err)
+        goto err_exit;
+    }
+  }
```
dr-m, line 3905: Can we first execute a loop that checks that all pages that we are going to free have been marked as allocated? And only after all consistency checks have passed, execute the modification?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,23 +3643,408 @@ dberr_t fsp_sys_tablespace_validate()
...
+    old_inodes_entry.insert(prev_addr.page, &mtr);
+
+  const fil_addr_t next_addr=
+    flst_get_next_addr(iblock->page.frame + FSEG_INODE_PAGE_NODE);
+
+  if (next_addr.page != FIL_NULL)
+    old_inodes_entry.insert(next_addr.page, &mtr);
+
+  fsp_free_seg_inode(space, inode, iblock, &mtr);
+  mtr.commit();
```
dr-m, line 3923: I do not quite understand the purpose of the `old_insert_entry.insert()` calls here. Are they really needed? If they are, why are we ignoring the return value?

storage/innobase/fsp/fsp0fsp.cc
dr-m: There seems to be a simpler approach:
* We would execute this extra freeing of leaked pages *only on startup* (not on shutdown) if the `:autoshrink` attribute is set on the system tablespace.
* We will structure the mini-transactions normally, with no special low-level error handling.
* If we notice any error, then we will:
  1. display an error message saying that `:autoshrink` is being ignored due to an error
  2. invoke `log_buffer_flush_to_disk()`
  3. execute InnoDB shutdown as if `innodb_fast_shutdown=2` had been set
  4. pretend that `:autoshrink` was not set
  5. restart InnoDB

In this way, all changes right before the current mini-transaction will be recovered, and the server should start up without corrupting a corrupted system tablespace too much further.

storage/innobase/dict/dict0load.cc
```diff
@@ -172,36 +172,24 @@ name_of_col_is(
...
-
-			return(NULL);
-		}
-	}
-
-	/* Get a record, let's save the position */
-	btr_pcur_store_position(pcur, mtr);
-
-	return(rec);
+  rec_t *rec = NULL;
```
dr-m, line 178: If we are reformatting this function, we might also replace `NULL` with the C++11 `nullptr`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2864,43 +2922,12 @@ fseg_free_step(
...
 #ifdef BTR_CUR_HASH_ADAPT
 			       , ahi
 #endif /* BTR_CUR_HASH_ADAPT */
-			       ) != DB_SUCCESS) {
-		return true;
-	}
-
-	buf_page_free(space, page_no, mtr);
+                                ))
+         return true;
```
dr-m, line 2930: I would retain the original formatting here, only removing the ` != DB_SUCCESS` since the function returns `bool` instead of `dberr_t`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2812,6 +2816,60 @@ fseg_free_extent(
...
+  if (xdes_t* descr= fseg_get_first_extent(inode, space, mtr, &err))
+    /* Free the extent held by the segment */
+    return fseg_free_extent(inode, iblock, space,
+                            xdes_get_offset(descr), mtr
+#ifdef BTR_CUR_HASH_ADAPT
+                            , ahi
+#endif /* BTR_CUR_HASH_ADAPT */
+                           ) != DB_SUCCESS;
+  if (err != DB_SUCCESS)
+    return true;
```
dr-m, line 2848: The first `return` could be a tail call and the second one would avoid loading a constant if this function returned `dberr_t` instead of `bool`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2812,6 +2816,60 @@ fseg_free_extent(
...
+
+  /* Free a frag page */
+  ulint n = fseg_find_last_used_frag_page_slot(inode);
+  if (UNIV_UNLIKELY(n == ULINT_UNDEFINED))
+  {
+    if (all_pages_freed)
+      *all_pages_freed= true;
+    return true;
+  }
```
dr-m, line 2857: Instead of having an output parameter `all_pages_freed` we could return a special `dberr_t` value that is not used by the other code, say, `DB_SUCCESS_LOCKED_REC`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+  bool insert_inode(uint32_t page, uint16_t offset)
+  {
+    return page < fil_system.sys_space->free_limit &&
+      offset >= FIL_PAGE_DATA && offset < srv_page_size - FIL_PAGE_DATA_END &&
+      emplace(uint64_t{page} << 32 | offset).second;
+  }
+
+  __attribute__((warn_unused_result))
+  /** Register an inode */
+  bool insert_seg(const byte *inode)
```
dr-m, line 3589: I think that both these need a `@return` comment.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+      break;
+  }
+  mtr->rollback_to_savepoint(savepoint);
+  dict_sys.unfreeze();
+
+  if (err == DB_SUCCESS)
+  {
+    err= add_index_root_pages(inodes, dict_sys.sys_tables);
+    if (err) return err;
+    err= add_index_root_pages(inodes, dict_sys.sys_indexes);
```
dr-m, line 3690: Errors are unlikely. We might also write these as
```c++
if (err == DB_SUCCESS)
  err= add_index_root_pages(inodes, dict_sys.sys_tables);
if (err == DB_SUCCESS)
  err= add_index_root_pages(inodes, dict_sys.sys_columns);
if (err == DB_SUCCESS)
  err= add_index_root_pages(inodes, dict_sys.sys_indexes);
if (err == DB_SUCCESS)
  err= add_index_root_pages(inodes, dict_sys.sys_fields);
return err;
```

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+  block= buf_pool.page_fix(page_id_t{0, TRX_SYS_PAGE_NO}, &err,
+                           buf_pool_t::FIX_WAIT_READ);
+  if (!block)
+    return err;
+
+  if (!inodes->insert_seg(block->page.frame + TRX_SYS + TRX_SYS_FSEG_HEADER))
+    err= DB_CORRUPTION;
+
+  if (!inodes->insert_seg(block->page.frame + TRX_SYS_DOUBLEWRITE
+                          + TRX_SYS_DOUBLEWRITE_FSEG))
```
dr-m, line 3718: Split the line after `+`, not before it.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+  for (ulint rseg_id= 0; rseg_id < TRX_SYS_N_RSEGS && err == DB_SUCCESS;
+       rseg_id++)
+  {
+    trx_rseg_t *rseg= &trx_sys.rseg_array[rseg_id];
+    if (rseg->space->id == 0)
+    {
+      block= buf_pool.page_fix(rseg->page_id(), &err,
+                               buf_pool_t::FIX_WAIT_READ);
+      if (!block)
+        continue;
```
dr-m, line 3758: Shouldn’t this be `break` or `return err` to avoid ignoring any error?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+      /* Even after slow shutdown, there is a possiblity that
+      cached undo log can exist. So store the segment as used one */
+      for (trx_undo_t *undo= UT_LIST_GET_FIRST(rseg->undo_cached);
+           undo && err == DB_SUCCESS;
+           undo= UT_LIST_GET_NEXT(undo_list, undo))
+      {
+        block= buf_pool.page_fix(page_id_t{0, undo->hdr_page_no}, &err,
+                                 buf_pool_t::FIX_WAIT_READ);
+        if (!block)
+	  continue;
```
dr-m, line 3773: Shouldn’t this be `return err;`?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+  buf_block_t *iblock= buf_page_get_gen(page_id_t{0, page_no}, 0,
+                                        RW_X_LATCH, nullptr, BUF_GET,
+                                        &mtr);
+
+  DBUG_EXECUTE_IF("unused_undo_free_fail_4", iblock= nullptr;);
+  if (!iblock)
+  {
+    mtr.commit();
+    return;
+  }
```
dr-m, line 3809: I’d suggest `goto func_exit` instead, to avoid code duplication.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+    mtr.memo_push(iblock, MTR_MEMO_PAGE_X_FIX);
+  }
+
+  DBUG_EXECUTE_IF("unused_undo_free_fail_6", mtr.commit(); return;);
+  if (all_pages_freed)
+    /* Freeing completed: free the segment inode */
+    fsp_free_seg_inode(space, inode, iblock, &mtr);
+  mtr.commit();
+  return;
+}
```
dr-m, line 3834: `return` is redundant at the end of a function body that returns `void`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+/** Free the unused segment
+@param boffset    offset of the list containing segment header
+@param space      system tablespace
+@param usedInodes Used system tablespace index node */
+static void fsp_free_unused_seg(inode_info *unused_inodes)
+{
+  for (auto it= unused_inodes->start();
+       it != unused_inodes->last(); it++)
+    fseg_inode_free((uint32_t)((*it) >> 32), (uint16_t)(*it));
+}
```
dr-m, line 3845: It looks like this had better be a member function of `inode_info`. The `@param` comments are wrong. Would the following work in a member function?
```c++
  for (auto i : *this)
    fseg_inode_free(uint32_t(i >> 32), uint16_t(i));
```

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+
+static dberr_t fsp_get_sys_unused_segment(inode_info *inodes,
+                                          uint16_t boffset,
+                                          inode_info *unused_inodes)
```
dr-m, line 3849: `inodes` could be a pointer to `const`, right? Did you consider making this a member function with a bit shorter name?
```c++
dberr_t inode_info::get_unused(uint16_t boffset, inode_info *unused) const
```

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+  {
+    ut_ad(addr.boffset >= FIL_PAGE_DATA);
+    ut_ad(addr.boffset < block->physical_size() - FIL_PAGE_DATA_END);
+    block= buf_pool.page_fix(page_id_t{0, addr.page}, &err,
+                             buf_pool_t::FIX_WAIT_READ);
+    if (!block)
+    {
+      addr.page= FIL_NULL;
+      continue;
+    }
```
dr-m, line 3869: It would be nice to add a comment that explains why we would tolerate this error.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+
+    fil_addr_t next_addr= flst_get_next_addr(block->page.frame + addr.boffset);
+    for (uint32_t i= 0;
+         i < FSP_SEG_INODES_PER_PAGE(srv_page_size) && err == DB_SUCCESS;
+         i++)
+    {
+      fseg_inode_t *inode=
```
dr-m, line 3876: Missing `const`

dr-m (commented):
This is really close to completion. I did not debug the test case yet, but will do so on my final review.

storage/innobase/fsp/fsp0fsp.cc
Thirunarayanan: It was added to avoid the assert. This has been fixed now.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+                                        buf_pool_t::FIX_WAIT_READ);
+  if (!block)
+    return err;
+  buf_block_t *header= block;
+  const uint32_t len= flst_get_len(block->page.frame + boffset);
+  fil_addr_t addr= flst_get_first(block->page.frame + boffset);
+  for (uint32_t i= len; i-- && err == DB_SUCCESS; )
+  {
+    ut_ad(addr.boffset >= FIL_PAGE_DATA);
+    ut_ad(addr.boffset < block->physical_size() - FIL_PAGE_DATA_END);
```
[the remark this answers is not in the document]
Thirunarayanan: Replaced the assert with error. But we're storing the next address before unfixing the block. Evicting and reloaded shouldn't be a issue.

mysql-test/suite/innodb/t/undo_leak_fail.test
```diff
@@ -0,0 +1,96 @@
...
+DELETE FROM t1;
+DROP TABLE t1;
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+let $shutdown_timeout=0;
+let $restart_parameters=--debug_dbug=d,unused_undo_free_fail_1;
+--source include/restart_mysqld.inc
+
+let $shutdown_timeout=;
+--echo # Fail to free the segment due to previous shutdown
+--let SEARCH_PATTERN= \[Warning\] InnoDB: Cannot free the unused segments in system tablespace because a previous shutdown was not with innodb_fast_shutdown=0
```
dr-m, line 33: The `\` in this and other `SEARCH_PATTERN` assignments had better be doubled. Yes, currently `\` before a character that does not require escaping in a string literal may be interpreted as a literal `\`, but I think that it is bad practice to rely on that. While this is Perl and `mtr` instead of `grep` and `/bin/sh´, I think that it is useful to refer to https://superuser.com/questions/1063992/backslash-in-single-quotes-giving-unpredictable-output-in-shell-script-linux

mysql-test/suite/innodb/r/undo_leak_fail.result
```diff
@@ -0,0 +1,75 @@
...
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	15728640
+SET GLOBAL innodb_fast_shutdown= 0;
+# Fail to free the segment while finding the used segments
+# restart: --debug_dbug=d,unused_undo_free_fail_2
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	13631488
+FOUND 2 /\[Warning\] InnoDB: Cannot free the unused segment in system tablespace due to Data structure corruption/ in mysqld.1.err
```
dr-m, line 36: I reproduced a test failure with
```sh
./mtr --rr innodb.undo_leak_fail
```
where the `information_schema.innodb_sys_tablespaces.file_size` was not changed between these two steps. I suggest you to do the same, to make the test as stable as possible. Below is the entire result difference:
```diff
CURRENT_TEST: innodb.undo_leak_fail
--- /mariadb/11/mysql-test/suite/innodb/r/undo_leak_fail.result	2024-10-07 08:49:07.780465875 +0300
+++ /mariadb/11/mysql-test/suite/innodb/r/undo_leak_fail.reject	2024-10-07 09:27:52.853385081 +0300
@@ -32,28 +32,28 @@
 # restart: --debug_dbug=d,unused_undo_free_fail_2
 SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
 NAME	FILE_SIZE
-innodb_system	13631488
+innodb_system	15728640
 FOUND 2 /\[Warning\] InnoDB: Cannot free the unused segment in system tablespace due to Data structure corruption/ in mysqld.1.err
 SET GLOBAL innodb_fast_shutdown= 0;
 # Fail to free the segment while finding the used segments
 # restart: --debug_dbug=d,unused_undo_free_fail_3
 SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
 NAME	FILE_SIZE
-innodb_system	13631488
+innodb_system	15728640
 FOUND 1 /\[Warning\] InnoDB: Cannot free the unused segment in system tablespace due to .*/ in mysqld.1.err
 SET GLOBAL innodb_fast_shutdown= 0;
 # Fail to free the segment while freeing the unused segments
 # restart: --debug_dbug=d,unused_undo_free_fail_4
 SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
 NAME	FILE_SIZE
-innodb_system	13631488
+innodb_system	15728640
 FOUND 5 /\[Warning\] InnoDB: Cannot free the unused segment in system tablespace due to Data structure corruption/ in mysqld.1.err
 SET GLOBAL innodb_fast_shutdown= 0;
 # Fail to free the segment while freeing the used segments
 # restart: --debug_dbug=d,unused_undo_free_fail_5
 SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
 NAME	FILE_SIZE
-innodb_system	11534336
+innodb_system	14680064
 FOUND 1 /\[Note\] InnoDB: Freeing the unused segment in page .*/ in mysqld.1.err
 SET GLOBAL innodb_fast_shutdown= 0;
 # restart
```
Does the exact intermediate output serve any useful purpose, or would it suffice to execute something like the following?
```sql
SELECT IF(file_size>10485760,'ok',file_size)
FROM information_schema.innodb_sys_tablespaces WHERE space=0;
```
It is worth noting that in my test run, the final file size was 10 MiB, as expected.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -37,12 +37,16 @@ Created 11/29/1995 Heikki Tuuri
 #include "btr0sea.h"
 #include "dict0boot.h"
 #include "log0log.h"
+#include "dict0load.h"
 #include "dict0mem.h"
-#include "fsp0types.h"
+#include "btr0pcur.h"
+#include "trx0sys.h"
```
dr-m, line 43: `trx0sys.h` includes many headers, such as `buf0buf.h`, `fil0fil.h` and `ut0byte.h`. The redundant headers should be removed.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3582,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+  if (err == DB_SUCCESS)
+    err= fsp_table_inodes(inodes, mtr);
+  return err;
+}
+
+/** Free the extents, fragment page from the given inode
+@param page_no index node page number
+@param offset  index node offset within page
+@return error code */
+static void fseg_inode_free(uint32_t page_no, uint16_t offset)
```
dr-m, line 3871: No error is being returned. It seems to me that any error needs to be propagated to the caller of `fil_space_t::garbage_collect()`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3582,401 @@ dberr_t fsp_sys_tablespace_validate()
...
+    err= used_inodes.get_unused(FSP_HEADER_OFFSET + FSP_SEG_INODES_FREE,
+                                &unused_inodes);
+  DBUG_EXECUTE_IF("unused_undo_free_fail_3", err= DB_CORRUPTION;);
+  mtr.commit();
+  if (err)
+    return err;
+
+  /* Ignore the error just like trx_purge_free_segment() */
+  unused_inodes.free_segs();
+  return DB_SUCCESS;
```
dr-m, line 3954: It is up to the caller to ignore the error (after reporting it). This had better be as follows:
```c++
  return unused_inodes.free_segs();
```

dr-m (commented):
This is getting better, but some error handling could improved further.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2812,6 +2816,76 @@ fseg_free_extent(
...
+
+  if (free_inode)
+  {
+    n= fseg_find_last_used_frag_page_slot(inode);
+    if (UNIV_UNLIKELY(n == ULINT_UNDEFINED))
+    {
+      fsp_free_seg_inode(space, inode, iblock, mtr);
+      return DB_SUCCESS;
+    }
+  }
```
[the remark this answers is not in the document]
dr-m: Sorry, I realized that `fsp_free_seg_inode()` requires that the allocation bitmap be all clear upfront. We have to retain the current order because of that.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2812,6 +2816,76 @@ fseg_free_extent(
...
+  if (UNIV_UNLIKELY(n == ULINT_UNDEFINED))
+  {
+    if (free_inode)
+      fsp_free_seg_inode(space, inode, iblock, mtr);
+    return DB_SUCCESS;
+  }
+
+  uint32_t page_no = fseg_get_nth_frag_page_no(inode, n);
+  if (header && page_no == page_get_page_no(page_align(header)))
+     return DB_SUCCESS;
```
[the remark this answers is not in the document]
dr-m: Should there be a `.` after the first line of the comment? There’s also a grammar error "should not freed".

This function is called by 3 functions. I see that there is a third possible caller `fseg_inode_free()`, but I don’t think we need to mention that in the comment. I would place the comment between the `if` and the `return` (which by the way is indented 1 column too deep), like this. Furthermore, I would name the parameter `const byte *header_page_no` so that we will avoid calling `page_align(header)` twice (once here and another time in our caller):

(We can compare the 4 bytes directly.)
```c++
  if (header_page_no && !memcmp_aligned<4>(header_page_no,
                                   inode + FSEG_FRAG_ARR +
                                   n * FSEG_FRAG_SLOT_SIZE, 4))
    /* header_page_no is only passed by fseg_free_step_not_header().
    In that case, the header page must be preserved, to be freed when we're
    finally called by fseg_free_step(). */
    return DB_SUCCESS;
```
Furthermore, the parameter `bool free_inode` is redundant and should be replaced by `!header_page_no`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2812,6 +2816,74 @@ fseg_free_extent(
...
+  if (err != DB_SUCCESS)
+    return err;
+  buf_page_free(space, page_no, mtr);
+  if (free_inode)
+  {
+    n= fseg_find_last_used_frag_page_slot(inode);
+    if (UNIV_UNLIKELY(n == ULINT_UNDEFINED))
+      return fsp_free_seg_inode(space, inode, iblock, mtr);
+  }
+  return DB_SUCCESS_LOCKED_REC;
```
dr-m, line 2884: I’d simplify this a bit:
```c++
  if (!header_page_no &&
      fseg_find_last_used_frag_page_slot(inode) == ULINT_UNDEFINED)
    return fsp_free_seg_inode(space, inode, iblock, mtr);
  return DB_SUCCESS_LOCKED_REC;
```
Can we define the parameters of `fseg_free_page_low()` in more similar order as `fseg_free_seg_inode()`? That is: `space, inode, iblock, mtr, page_no, ahi`? Likewise, the first parameters of this function `fseg_free_step_low()` should be `space, inode, iblock, header_page_no, ahi`. In that way, there would be less shuffling of parameter registers needed.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -2948,42 +2979,14 @@ fseg_free_step_not_header(
...
-	if (UNIV_UNLIKELY(n == ULINT_UNDEFINED)) {
-		return true;
-	}
-
-	uint32_t page_no = fseg_get_nth_frag_page_no(inode, n);
-
-	if (page_no == page_get_page_no(page_align(header))) {
+                               ) != DB_SUCCESS_LOCKED_REC) {
 		return true;
 	}
```
dr-m, line 2988: We could replace `if (fseg_free_step_low(…) != DB_SUCCESS_LOCKED_REC)` with
```c++
	return fseg_free_step_low(…) != DB_SUCCESS_LOCKED_REC;
```
But, is that what we really want here? Wouldn’t we want `== DB_SUCCESS` instead? That is, our caller should keep freeing as long there was no corruption and we have not freed everything yet? Our callers are `btr_free_but_not_root()`, which is followed by `btr_free_root()`, and `trx_purge_free_segment()`. I’d suggest to carefully review the logic in both. A source code comment could be useful here.

On a related note, I see that `fseg_inode_free()` is checking for `DB_SUCCESS_LOCKED_REC`. It seems to be the right course of action there.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3580,413 @@ dberr_t fsp_sys_tablespace_validate()
...
+            !unused->insert_inode(addr.page, offset))
+        {
+          err= DB_CORRUPTION;
+          break;
+        }
+      }
+      addr= next_addr;
+      block->page.unfix();
+      if (err)
+        break;
```
dr-m, line 3676: OK, the first `break`is indeed breaking out of both loops.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3580,413 @@ dberr_t fsp_sys_tablespace_validate()
...
+/** Free the unused segment
+@return error code */
+dberr_t inode_info::free_segs()
+{
+  for (auto i : *this)
+  {
+    dberr_t err= fseg_inode_free(uint32_t(i >> 32), uint16_t(i));
+    if (err)
+      return err;
+  }
```
dr-m, line 3922: We could avoid the braces if we wrote just
```c++
  for (auto i : *this)
    if (dberr_t err= fseg_inode_free(uint32_t(i >> 32), uint16_t(i)))
      return err;
```

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3580,413 @@ dberr_t fsp_sys_tablespace_validate()
...
+  fseg_inode_t *inode= nullptr;
+  DBUG_EXECUTE_IF("unused_undo_free_fail_4", iblock= nullptr;);
+  if (!iblock)
+    goto func_exit;
+
+  sql_print_information("InnoDB: Freeing the unused segment in page "
+                        UINT32PF, page_no);
+  inode= iblock->page.frame + offset;
+  err= fseg_free_step_low(iblock, inode, space, &mtr, nullptr, true);
+  while (err == DB_SUCCESS_LOCKED_REC)
```
dr-m, line 3894: It looks like this call could be made inside the `while` block, to avoid duplicating the call at the end of the `while` body:
```c++
  while ((err= fseg_free_step_low(…) == DB_SUCCESS_LOCKED_REC)
```
I would like to see a source code comment after the `while` loop body, before `func_exit:` that would explain why there is no call to `fseg_free_step()`, like we have in `trx_purge_free_segment()`. Is the header page being freed as well? If not, what will remain in the header page?

dr-m (commented):
The error handling is almost there. I’d like to see some clarifying source code comments still.

storage/innobase/fsp/fsp0fsp.cc
Thirunarayanan: If the `fseg_free_step_low()` returns any error code other than `DB_SUCCESS_LOCKED_REC`  then InnoDB
faced corruption or freed the segment successfully. So I think this condition is correct.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,413 @@ dberr_t fsp_sys_tablespace_validate()
...
+  err= fsp_traverse_extents(space, &last_used_extent, &mtr);
   if (err != DB_SUCCESS)
   {
-func_exit:
+err_exit:
+    mtr.commit();
     sql_print_warning("InnoDB: Cannot shrink the system tablespace "
                       "due to %s", ut_strerr(err));
-    mtr.commit();
     return;
```
dr-m, line 3981: This code does not seem to be covered by the regression tests.

dr-m (commented):
I reviewed the test execution in a bit more detail. I think that we need to cover the existence of `XA PREPARE` transactions and improve some error handling and code coverage.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3577,22 +3571,429 @@ dberr_t fsp_sys_tablespace_validate()
...
+        this segment as unused one */
+        if (seg_id == 0 || seg_id == 2)
+          continue;
+	uint16_t offset= uint16_t(inode - block->page.frame);
+        if (!find(addr.page, offset) &&
+            !unused->insert_inode(addr.page, offset))
+        {
+          err= DB_CORRUPTION;
+          break;
+        }
```
dr-m, line 3662: Please distinguish these as `DB_RECORD_NOT_FOUND` and `DB_DUPLICATE_KEY` so that we can output more accurate messages.

dr-m (commented):
Sorry, this still needs some minor refinement.

storage/innobase/trx/trx0rseg.cc
```diff
@@ -780,3 +780,29 @@ void trx_rseg_update_binlog_offset(buf_block_t *rseg_header,
...
+      undo= UT_LIST_GET_NEXT(undo_list, undo);
+    }
+
+    undo= UT_LIST_GET_FIRST(rseg.undo_cached);
+    while (undo)
+    {
+      if (undo->state == TRX_UNDO_PREPARED)
+        return true;
+      undo= UT_LIST_GET_NEXT(undo_list, undo);
+    }
```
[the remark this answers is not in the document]
Thirunarayanan: Yes, I was wrong. `undo_cached` must carry the state `TRX_UNDO_CACHED`  and state has been changed during `trx_purge_add_undo_to_history()` . So we should check only for `rseg.undo_list`

storage/innobase/fsp/fsp0fsp.cc
dr-m: Sorry, I misread the condition as `||` instead of `&&`. We can only return one error here, `DB_DUPLICATE_KEY`.

Thirunarayanan: There is no need for `DB_RECORD_NOT_FOUND`

dr-m (approved):

PR #3467, opened 2024-08-16
MDEV-34759: buf_page_get_low() is unnecessarily acquiring exclusive latch

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34759*
## Description
Basically any request on a secondary index leaf page would unnecessarily first acquire an exclusive page latch, in case any buffered changes would have to be merged into the page. This would unnecessarily slow down read operations.

`buf_page_ibuf_merge_try()`: A separate function for invoking `ibuf_merge_or_delete_for_page()` when needed. Use the already requested page latch for determining if the call is necessary. If it is and if we are currently holding `rw_latch==RW_S_LATCH`, upgrading to an exclusive latch may involve waiting that another thread acquires and releases a U or X latch on the page. If we have to wait, we must recheck if the call to `ibuf_merge_or_delete_for_page()` is still needed. If the page turns out to be corrupted, we will release and fail the operation. Finally, the exclusive page latch will be downgraded to the originally requested latch.

`ssux_lock_impl::rd_u_upgrade_try()`, `sux_lock::s_x_upgrade_try()`: Attempt to upgrade a shared lock to an update lock.

`ssux_lock_impl::u_rd_downgrade()`, `sux_lock::u_s_downgrade()`: Downgrade an update lock to shared.
## Release Notes
The performance of secondary index read operations was improved.
## How can this PR be tested?
This is a low-level change that is well covered by `./mtr`, except for the change buffer.

The change buffer needs to be subjected to a stress test, with `innodb_change_buffering=all`.

Performance can be tested with Sysbench `select_random_ranges` using `--tables=1 --table_size=10000`. On my system with 2×10×2 threads, I tested `--threads=20` against the baseline:
| revision | throughput/qps |
| ---------| --------------:|
| 267c0fce5696a6be7f2e04bc01035db855cc970f | 43141.66 |
| 3bf05a4c201c5393e7707ae4c1662e1f2cf0f65e | 70529.46 |

This is also applicable to the 10.5 branch, but this fix depends on 2bd661ca1038e78296289d9285d54b57954d48e6 (#3317), which depends on prior refactoring that was done in the 10.6 branch.
## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -2404,6 +2404,78 @@ err_exit:
 	return(FALSE);
 }
 
+ATTRIBUTE_COLD
+/** Try to merge buffered changes to a buffer pool page.
+@param block     buffer-fixed and latched block
+@param rw_latch  RW_X_LATCH, RW_SX_LATCH, RW_S_LATCH held on block
+@param err       error code
+@return whether the page is invalid (corrupted) */
+static bool buf_page_ibuf_merge_try(buf_block_t *block, ulint rw_latch,
+                                    dberr_t *err)
+{
+  ut_ad(block->page.lock.have_any());
+  ut_ad(block->page.buf_fix_count());
+
+  if (fil_page_get_type(block->page.frame) != FIL_PAGE_INDEX ||
+      !page_is_leaf(block->page.frame))
+    return false;
+
+  if (rw_latch != RW_X_LATCH)
+  {
+    if (rw_latch == RW_S_LATCH)
+    {
+      if (!block->page.lock.s_x_upgrade())
+      {
+        uint32_t state;
+        state= block->page.state();
+        if (state < buf_page_t::UNFIXED)
+        {
+        fail:
+          block->page.lock.x_unlock();
+          return true;
+        }
+        ut_ad(state & ~buf_page_t::LRU_MASK);
+        ut_ad(state < buf_page_t::READ_FIX);
+        if (state < buf_page_t::IBUF_EXIST || state >= buf_page_t::REINIT)
+          /* ibuf_merge_or_delete_for_page() was already invoked in
+          another thread. */
+          goto downgrade_to_s;
+      }
+    }
+    else
+    {
+      ut_ad(rw_latch == RW_SX_LATCH);
+      block->page.lock.u_x_upgrade();
+    }
+  }
+
+  ut_ad(block->page.lock.have_x());
+  block->page.clear_ibuf_exist();
+  if (dberr_t e= ibuf_merge_or_delete_for_page(block, block->page.id(),
+                                               block->zip_size()))
+  {
+    if (err)
+      *err= e;
+    goto fail;
+  }
+
+  switch (rw_latch) {
+  default:
+    ut_ad(rw_latch == RW_X_LATCH);
+    break;
+  case RW_SX_LATCH:
+    block->page.lock.x_u_downgrade();
+    break;
+  case RW_S_LATCH:
+  downgrade_to_s:
+    block->page.lock.x_u_downgrade();
+    block->page.lock.u_s_downgrade();
+    break;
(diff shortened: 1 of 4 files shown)


storage/innobase/buf/buf0buf.cc
```diff
@@ -2404,6 +2404,81 @@ buf_zip_decompress(
...
+
+  if (rw_latch != RW_X_LATCH)
+  {
+    if (rw_latch == RW_S_LATCH)
+    {
+      if (!block->page.lock.s_x_upgrade_try())
+      {
+        block->page.lock.s_unlock();
+        block->page.lock.x_lock();
+        state= block->page.state();
```
mariadb-DebarunBanerjee, line 2433: We can declare the variable state here. It is not used outside the scope.

storage/innobase/buf/buf0buf.cc
```diff
@@ -2404,6 +2404,81 @@ buf_zip_decompress(
...
+        {
+        fail:
+          block->page.lock.x_unlock();
+          return true;
+        }
+        ut_ad(state & ~buf_page_t::LRU_MASK);
+        ut_ad(state < buf_page_t::READ_FIX);
+        if (state < buf_page_t::IBUF_EXIST || state >= buf_page_t::REINIT)
+          /* ibuf_merge_or_delete_for_page() was already invoked in
+          another thread. */
```
mariadb-DebarunBanerjee, line 2444: Essentially we are re-checking the condition again after acquiring the latch. Looks all good. May be we can bring down the same check from the caller to this function and have some lamda which we can use in the beginning of the function and here.

storage/innobase/include/srw_lock.h
```diff
@@ -282,6 +282,8 @@ class ssux_lock_impl
 #endif
   }
 
+  bool rd_u_upgrade_try() { return writer.wr_lock_try(); }
```
mariadb-DebarunBanerjee, line 285: May be better to reduce the S lock count (s_unlock) here if the attempt is successful. That makes the upgrade to U complete. Currently it is being done in the caller.

mariadb-DebarunBanerjee (approved):
@dr-m Thanks for the patch.

The fix looks correct and good to me. It is good observation that we must only try for upgrading S -> X in current design and `not wait` which could lead to deadlock. I have some minor points for your consideration.

One way people implement the upgrade with wait Q(queue) based model is to jump the Q ahead of other waiting locks and just wait for granted locks eliminating any deadlock possibility. This is not possible here with two level locking where the waiting X lock could have already passed the fist step and waiting on the readers. It is not possible to jump ahead of it. We don't need to worry though as releasing and reacquiring should be fine here as implemented in the patch.

storage/innobase/buf/buf0buf.cc
dr-m: Earlier, you [suggested something similar](https://github.com/MariaDB/server/pull/3022#discussion_r1481081933) and I found out that a lambda function would not be inlined. It might work here, but the emitted code would have to be reviewed on a number of compilers and targets. I would rather keep it ‘ugly’ for now in order to avoid surprises.

storage/innobase/include/srw_lock.h
dr-m: I think that it could be confusing and error prone to release a lock in a `trylock` operation. However, I can write a convenience member function `sux_lock::s_x_upgrade()` that is usable for `block_lock` but not necessarily elsewhere. The reasoning would be that `buf_page_t` is typically also protected by `fix()` and a `dict_index_t::lock` or some higher-level lock while we end up waiting.

storage/innobase/buf/buf0buf.cc
dr-m: Right, we can, but the initialization must be separate from the definition, because otherwise a warning would be emitted for the `goto fail` that leads into this scope.

PR #3470, opened 2024-08-20
MDEV-34772  Overuse of big stackvariables results in stackoverflows

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34772*

## Description
- This patch removes the PRAGMA_DISABLE_CHECK_STACK_FRAME usage
    inside innodb and also made the large variable into thread_local
    storage.

## How can this PR be tested?
I ran with gcc -Wstack-usage=8192

cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS=-Wstack-usage=8192

## PR quality check

### DIFF
--- a/extra/mariabackup/fil_cur.cc
+++ b/extra/mariabackup/fil_cur.cc
@@ -244,15 +244,12 @@ xb_fil_cur_open(
 	return(XB_FIL_CUR_SUCCESS);
 }
 
-/* Stack usage 131224 with clang */
-PRAGMA_DISABLE_CHECK_STACK_FRAME
-
 static bool page_is_corrupted(const byte *page, ulint page_no,
 			      const xb_fil_cur_t *cursor,
 			      const fil_space_t *space)
 {
-	byte tmp_frame[UNIV_PAGE_SIZE_MAX];
-	byte tmp_page[UNIV_PAGE_SIZE_MAX];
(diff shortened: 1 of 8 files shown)


dr-m (changes requested):
I think that we need to investigate further how the `thread_local` variables are being allocated. It would seem logical to allocate a contiguous area of memory for each thread, covering both `thread_local` variables and the stack. If that holds in some environment, this change could actually make things worse, because I could assume that the `thread_local` variables will be assigned a static offset relative to the thread local storage (TLS), while the stack allocation is dynamic (stack frames are being allocated at runtime by modifying the stack pointer).

In fact, https://elixir.bootlin.com/glibc/glibc-2.40.9000/source/nptl/allocatestack.c#L351 is mentioned in https://issues.chromium.org/issues/335342168 with the following comment:
>`pthread_create` checks for a couple of things when determining if the provided stack size is large enough. One of them is the size of the static TLS area which is determined by the dynamic loader

For example, `innobase_raw_format()` and `ha_innobase::ft_init_ext()`  both declare a `char buf_tmp[8192]`. If it is not `thread_local`, this buffer will only be reserved when one of these functions is being called. These functions are not calling each other. If the variables are `thread_local`, then it is thinkable that 2*8192 bytes of address space per thread will be wasted, no matter if these functions are ever invoked.

dr-m:
This was fixed differently in #4051.

PR #3471, opened 2024-08-20
MDEV-34565 MariaDB crashes with SIGILL because the OS does not support AVX512

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34565*
## Description
In commit 232d7a5e2dc50181294b927e2bc4b02e282725a8 (#3430) we almost got the detection logic right. However, the `xgetbv` instruction would crash if Linux was started up with the option `noxsave`.
## Release Notes
MariaDB Server should no longer crash on startup when the CPU but not the operating system supports AVX512.

`have_vpclmulqdq()`: Check for the XSAVE flag at the correct position and also for the AVX flag.
## How can this PR be tested?
This was tested by invoking `unittest/mysys/crc32-t` on Ubuntu 22.04, on its Linux 5.15 kernel running with and without the `noxsave` option.
## PR quality check

### DIFF
--- a/mysys/crc32/crc32c_x86.cc
+++ b/mysys/crc32/crc32c_x86.cc
@@ -39,7 +39,7 @@ extern "C" unsigned crc32c_sse42(unsigned crc, const void* buf, size_t size);
 
 constexpr uint32_t cpuid_ecx_SSE42= 1U << 20;
 constexpr uint32_t cpuid_ecx_SSE42_AND_PCLMUL= cpuid_ecx_SSE42 | 1U << 1;
-constexpr uint32_t cpuid_ecx_XSAVE= 1U << 26;
+constexpr uint32_t cpuid_ecx_AVX_AND_XSAVE= 1U << 28 | 1U << 27;
 
 static uint32_t cpuid_ecx()
 {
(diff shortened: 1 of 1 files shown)


vaintroub (approved):
Seems ok

Although for aesthetic reasons I'd prefer individual flags rather than combined ones, and checking with `(ecx & bit)` rather than (arguably harder to read) `(~ecx & combined_bits)`

dr-m:
> Seems ok
> 
> Although for aesthetic reasons I'd prefer individual flags rather than combined ones, and checking with `(ecx & bit)` rather than (arguably harder to read) `(~ecx & combined_bits)`

For aesthetic reasons, I prefer fewer machine instructions. I was too lazy to check which compiler versions (if any) would be smart enough to undo such pessimization.

vaintroub:
I do not think single AND instruction, even if not optimized by any compiler, would take as much time as it takes a reader to understand Yoda notation :)

dr-m:
The straightforward translation of logical `&&` or `||` would be into conditional branches that implement short-circuit evaluation. Conditional branches are poison for modern superscalar processors. Depending on the target ISA and the smartness of the compiler, the conditional branches could be avoided. On recent x86, fancy instructions such as `cmov` or `setCC` might help if they are available in `-march=i686` or `-march=amd64`.

The forest of `cpuid` leaves is rather confusing to navigate anyway.

PR #3472, opened 2024-08-20
MDEV-34787: invalid instruction signal 4

DaveGosselin-MariaDB:
Changes for powerpc in git commit dec3f8ca69e5 assume power8 VSX
    extensions for faster crc32 support, but earlier processors
    (e.g. power5) don't have them.  On unexpected powerpc chips,
    we will not use an optimized crc32 implementation.  CMake does
    not provide a facility to detect processor features or in the
    case of powerpc, flavors of the chip.  Create a place where we
    can detect such flavors and caveat as necessary.

### DIFF
--- a/mysys/CMakeLists.txt
+++ b/mysys/CMakeLists.txt
@@ -134,7 +134,10 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   ENDIF()
 ENDIF()
 
-IF(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|powerpc64" OR CMAKE_SYSTEM_NAME MATCHES AIX)
+cmake_host_system_information(RESULT procname QUERY PROCESSOR_NAME)
+string(FIND ${procname} "PPC970" proc_is_ppc970)
+IF((proc_is_ppc970 EQUAL -1) AND
+   (CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|powerpc64" OR CMAKE_SYSTEM_NAME MATCHES AIX))
   SET(MYSYS_SOURCES ${MYSYS_SOURCES} crc32/crc32_ppc64.c crc32/crc32c_ppc.c)
   SET_SOURCE_FILES_PROPERTIES(crc32/crc32_ppc64.c crc32/crc32c_ppc.c PROPERTIES
         COMPILE_FLAGS "${COMPILE_FLAGS} -maltivec -mvsx -mpower8-vector -mcrypto -mpower8-vector")
--- a/mysys/crc32ieee.cc
+++ b/mysys/crc32ieee.cc
@@ -50,9 +50,6 @@ static my_crc32_t init_crc32()
 
 static const my_crc32_t my_checksum_func= init_crc32();
 
-#ifdef __powerpc64__
-# error "my_checksum() is defined in mysys/crc32/crc32_ppc64.c"
-#endif
 extern "C"
 unsigned int my_checksum(unsigned int crc, const void *data, size_t len)
 {


mysys/CMakeLists.txt
```diff
@@ -134,7 +134,10 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   ENDIF()
 ENDIF()
 
-IF(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|powerpc64" OR CMAKE_SYSTEM_NAME MATCHES AIX)
+cmake_host_system_information(RESULT procname QUERY PROCESSOR_NAME)
+string(FIND ${procname} "PPC970" proc_is_ppc970)
+IF((proc_is_ppc970 EQUAL -1) AND
```
DaveGosselin-MariaDB, line 139: On my system, `string()` populates `proc_is_ppc970` with `0` because `procname` is `PPC970` at this point, implying the offset to the substring is at the start of the string (it is).

mysys/crc32ieee.cc
```diff
@@ -50,9 +50,6 @@ static my_crc32_t init_crc32()
 
 static const my_crc32_t my_checksum_func= init_crc32();
 
-#ifdef __powerpc64__
```
DaveGosselin-MariaDB, line 53: The `__powerpc64__` define is true regardless of processor flavor, so this has to go.

mysys/CMakeLists.txt
DaveGosselin-MariaDB: Please double check this, I don't want to break the ppc-optimized case but I don't have a newer chip to test on.

dr-m: As far as I understand, the SIMD implementation of CRC-32 is based on https://github.com/antonblanchard/crc32-vpmsum. The POWER v2.07 ISA was introduced in 2013. The IBM PowerPC 970 should implement the v2.03 ISA.

I suppose that @grooverdan had assumed that nobody would be running MariaDB on anything older than POWER8, which should already be 10 years old.

What operating system do you run? https://wiki.debian.org/ppc64el says that the minimum requirement is POWER8 (or the POWER v2.07 ISA). This suggests that a compile-time check (like the one you attempted) should be sufficient for older hardware and there might be no need to add a runtime check.

mysys/crc32ieee.cc
dr-m: Right, this is a redundant compile-time guard against multiple definition of `my_checksum()` in case an error in the CMake logic would cause both implementations of  `my_checksum()`.  A duplicated definition should result in a link-time error anyway.

mysys/CMakeLists.txt
grooverdan: > As far as I understand, the SIMD implementation of CRC-32 is based on https://github.com/antonblanchard/crc32-vpmsum. The POWER v2.07 ISA was introduced in 2013. The IBM PowerPC 970 should implement the v2.03 ISA.
> 
> I suppose that @grooverdan had assumed that nobody would be running MariaDB on anything older than POWER8, which should already be 10 years old.

I certainly assumed that `arch_ppc_probe` would return false if the hardware wasn't capable and therefore wouldn't SIGILL, even if the optimized code appeared in the compiled code.

Some lesser constraints existed for OSes that only supported POWER8+.

To complicate it more the Debian builds are chroots rather than a cross compile https://buildd.debian.org/status/package.php?p=mariadb where compile time checks may detect the OS of the host. Luckily powerpc is 32bit and somehow works.

> What operating system do you run? https://wiki.debian.org/ppc64el says that the minimum requirement is POWER8 (or the POWER v2.07 ISA). This suggests that a compile-time check (like the one you attempted) should be sufficient for older hardware and there might be no need to add a runtime check.

dr-m: fabaac86a18cfa026ca53b4a62e9edbf0bfd32d4 introduced a runtime check for something else that is based on trapping `SIGILL`. That could be an option here. We could have a compile-time check to enable the run-time detection only on operating systems that support older POWER ISA.

PR #3476, opened 2024-08-23
MDEV-34802 Recovery fails to note some log corruption

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34802*
## Description
`recv_recovery_from_checkpoint_start()`: Abort startup due to log corruption if we were unable to parse the entire log between the latest log checkpoint and the corresponding `FILE_CHECKPOINT` record.

Also, reduce some code bloat related to log output and  `log_sys.mutex`.
## Release Notes
InnoDB crash recovery would wrongly succeed when the log file is corrupted. This would lead to corruption later, with messages noting "log sequence number … is in the future".
## How can this PR be tested?
This is somewhat covered by the regression test suite. A version of this could be useful to test on 10.11 before the fix of #3458 to ensure that there will be no assertion failures or errors, but instead recovery failures. (Also, there should not be excessively many new recovery failures for a generic workload.)
## PR quality check

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -3474,10 +3474,9 @@ recv_recovery_from_checkpoint_start(lsn_t flush_lsn)
 	ut_d(mysql_mutex_unlock(&buf_pool.flush_list_mutex));
 
 	if (srv_force_recovery >= SRV_FORCE_NO_LOG_REDO) {
-
-		ib::info() << "innodb_force_recovery=6 skips redo log apply";
-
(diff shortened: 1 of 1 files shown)


mariadb-DebarunBanerjee (approved):
Looks good,

dr-m:
Thank you. I will wait for some test results.

Note to myself: I want to see if `recv_max_page_lsn` and `recv_lsn_checks_on` can be removed from 10.6 onwards, thanks to f2c17cc9d9bcd634887846d3064bcb71243f9cc0.

PR #3477, opened 2024-08-23
[WiP] Validate password does not match any username

SouradeepSaha:
## Description
Validate password does not match any username
    
Currenly, simple_password_check plugin only checks that a password does not match the username who's password is being set. However, the password can be set to another user in the table. Thus, implement logic to prevent password from matching any username in the table.
    
All new code of the whole pull request, including one or several files that are either new files or modified ones, are contributed under the
BSD-new license. I am contributing on behalf of my employer Amazon Web Services, Inc.

## Release Notes
A system variable, simple_password_check_usernames would be added with the allowed values ON/OFF, which would allow customers to turn this feature on/off. The following page would need to be updated: https://mariadb.com/kb/en/simple-password-check-plugin/

## How can this PR be tested?

MTR tests will be added to verify the functionality of the new changes.

## PR quality check

### DIFF
--- a/plugin/simple_password_check/simple_password_check.c
+++ b/plugin/simple_password_check/simple_password_check.c
@@ -22,6 +22,57 @@
 
 static unsigned min_length, min_digits, min_letters, min_others;
 
+#define SQL_BUFF_LEN 2048
+#define USERNAME_VIEW_NAME "username_view"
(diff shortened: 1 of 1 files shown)


vuvova:
I don't think we'll want to have it. It allows a user to probe what other user names exist in the database, this is a pretty serious information leak.

PR #3478, opened 2024-08-23
Update mysqldump

mrgharabaghi:
## Description
Since `MySQL` updated `utf8` to `utf8mb4`, it's better to reflect that for `MariaDB`.


mrgharabaghi:
[MySQL repository](https://github.com/mysql/mysql-server/blob/596f0d238489a9cf9f43ce1ff905984f58d227b6/client/mysqldump.cc#L3009C56-L3009C64)

PR #3481, opened 2024-08-27
Skip FK checks in Galera during applying in IST

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-34822*

## Description

Appliers need to verify foreign key constraints during normal operation, in multi-active topologies, and for this reason appliers are configured to enable FK checking.

However, during node joining, in IST and latter catch up period, the node is still idle (from local connections), and only source for incoming transactions is the cluster sending certified write sets for applying. IST happens with parallel applying, and there is a possibility that foreign key check cause lock conflicts between appliers accessing FK child and parent tables. Also, the excessive FK checking will slow down IST process somewhat.

For this reasons, we could relax FK checks for appliers during IST and catch up periods. The relaxed FK check mode should, however, be configurable e.g. by wsrep_mode flag: SKIP_APPLIER_FK_CHECKS_IN_IST. When this operation mode is set, and the node is processing IST or catch up, appliers should skip FK checking.

## How can this PR be tested?

This PR consists of two separate but related changes split into two independent commits:

- the first one makes `wsrep_ready` flag reads lock-free by converting it into `std::atomic` variable
- the second commit is the actual feature implementation with a test.

- [x] *This is a new feature or a refactoring, and the PR is based against the latest MariaDB development branch.*

## PR quality check


sysprg:
Thanks, the main commit has been added to temporary test branch. Also, a separate commit with atomic variable is already included in 10.5, because it is not directly related to the new feature and is useful for all versions, improving their performance:
https://github.com/MariaDB/server/commit/9f61aa4f8aeb3cf828b765855af80eac2543e7c2

denis-protivensky:
Thanks! Indeed the commit with atomic is good to include to all versions.

PR #3484, opened 2024-08-28
MDEV-34825 FreeBSD fails to build under clang natively

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34825*

## Description

clang doesn't have /usr/local/lib in the path. As such there are various depedency linkages that will fail.

For example pcre and libfmt.`

## Release Notes

Add /usr/local/lib to linking path for FreeBSD.

## How can this PR be tested?

Check BB FreeBSD worker to ensure that pcre is linked rather than bundled.

## PR quality check

@Sp1l is this too much of a hack?


grooverdan:
test: https://buildbot.mariadb.org/#/builders/619/builds/6305

- pcre2 system libraries detected
- no pcre2 download and build occured
- my_apc-t unit test passed

Removed on unneeded commit, but all [should pass again](https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F3484%2Fmerge).

cmake/pcre.cmake
```diff
@@ -1,12 +1,12 @@
-INCLUDE (CheckCSourceRuns)
 INCLUDE (ExternalProject)
 
 SET(WITH_PCRE "auto" CACHE STRING
    "Which pcre to use (possible values are 'bundled', 'system', or 'auto')")
 
 MACRO(BUNDLE_PCRE2)
   SET(dir "${CMAKE_BINARY_DIR}/extra/pcre2")
-  SET(PCRE_INCLUDES ${dir}/src/pcre2-build ${dir}/src/pcre2/src)
+  SET(PCRE_INCLUDE_DIRS ${dir}/src/pcre2-build ${dir}/src/pcre2/src)
```
LinuxJedi, line 8: This renaming will break ColumnStore, best to flag this to that team.

LinuxJedi (approved):
In general I approve, but please notify the ColumnStore team prior to merging, they use the server's `PCRE_INCLUDES` in their submodule.

cmake/pcre.cmake
grooverdan: Turns out 10.6 (develop-6) was already out of date - https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3304
develop-5 doesn't appear to be getting updated.

Appended commit for it d80ce3d905

include/my_cpu.h
```diff
@@ -81,7 +94,13 @@ static inline void MY_RELAX_CPU(void)
...
 #endif
 #elif defined(_ARCH_PWR8)
-  __ppc_get_timebase();
+#ifdef __FreeBSD__
+  uint64_t __tb;
+  __asm__ volatile ("mfspr %0, 268" : "=r" (__tb));
+#else
+  /* Changed from __ppc_get_timebase for musl compatibility */
+  __builtin_ppc_get_timebase();
+#endif
```
dr-m, line 103: Is there a reason why we use inline assembler on FreeBSD and a built-in function elsewhere? On https://godbolt.org there only is a clang trunk for POWER. It would translate `__builtin_ppc_get_timebase()` to the same instruction, with 268 as the second operand.

This came across my head while I was working on #3752.

grooverdan: I think I was just copying the FreeBSD patch. With it in LLVM it seems unlikely it was required, or if so it was a very long time ago.

https://github.com/llvm/llvm-project/commit/65e1e4dbe0f1fff23eceeb0fcadd93e8f0a302aa

PR #3485, opened 2024-08-28
Extract some of #3360 fixes to 10.5.x

ParadoxV5:
  * Update: *The Jira issue number for this PR is: [MDEV-35429](https://jira.mariadb.org/browse/MDEV-35429)*
* To clarify, my coding period for Google Summer of Code (GSoC) 2024 ended.
  While this code originates from my GSoC project, this publication (i.e., commit and PR descriptions) isn’t.

## Description

#3360 uncovered countless *potential minor security vulnerabilities* on `my_snprintf` uses.
This commit ports a squashed subset of their fixes according to the bug-fixing process to 10.5, our oldest maintained version.
(My GSoC mentor said that they aren’t much of a problem and I can open a PR normally.)

Most `size_t`s were still simple `uint`s back here, though I didn’t include their changes from `%d` to `%u` so this commit doesn’t conflict when merging up. The exception is `sql/table.cc` with its large chuncks of diffs from reformatting.
After all, I expect users to upgrade to the newer versions sooner or later.
Similarly, while 10.5 may have had other issues in this category that are obsolete in 11.6 (the branch from which the superset *was* based on), I am not goïng to hunt them down.

## Release Notes
~~Nothing?
These internal corrections shouldn’t be user-visible (other than perhaps one fewer crashes).~~

* Fixed data size mismatches that were garbling outputs (or possibly even crashes) on *problematic* platforms – mostly on error messages and debug logs

## How can this PR be tested?
I’m not certain how we could catch these human errors besides testing #3360 (MDEV-21978) and syncing this PR with it.

## PR quality check

### DIFF
--- a/plugin/server_audit/server_audit.c
+++ b/plugin/server_audit/server_audit.c
@@ -1398,7 +1398,7 @@ static size_t log_header(char *message, size_t message_len,
   if (output_type == OUTPUT_SYSLOG)
     return my_snprintf(message, message_len,
         "%.*s,%.*s,%.*s,%d,%lld,%s",
-        (unsigned int) serverhost_len, serverhost,
+        (int) serverhost_len, serverhost,
         username_len, username,
         host_len, host,
         connection_id, query_id, operation);
@@ -1408,7 +1408,7 @@ static size_t log_header(char *message, size_t message_len,
       "%04d%02d%02d %02d:%02d:%02d,%.*s,%.*s,%.*s,%d,%lld,%s",
       tm_time.tm_year+1900, tm_time.tm_mon+1, tm_time.tm_mday,
       tm_time.tm_hour, tm_time.tm_min, tm_time.tm_sec,
-      serverhost_len, serverhost,
+      (int) serverhost_len, serverhost,
       username_len, username,
       host_len, host,
       connection_id, query_id, operation);
@@ -1477,7 +1477,7 @@ static int log_connection_event(const struct mysql_event_connection *event,
                     event->ip, event->ip_length,
                     event->thread_id, 0, type);
   csize+= my_snprintf(message+csize, sizeof(message) - 1 - csize,
-    ",%.*s,,%d", event->database.length, event->database.str, event->status);
+    ",%.*s,,%d", (int) event->database.length, event->database.str, event->status);
   message[csize]= '\n';
   return write_log(message, csize + 1, 1);
 }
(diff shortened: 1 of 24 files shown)


ParadoxV5:
Riddle me this, some runners doesn’t understand `PRIdPTR` while others do.
This problem has come up before too, at https://github.com/MariaDB/server/pull/3360#discussion_r1703552883.

ParadoxV5:
The Buildbot failure is `innodb.innodb-lock-inherit-read_commited` timing out.

plugin/server_audit/server_audit.c
```diff
@@ -1398,7 +1398,7 @@ static size_t log_header(char *message, size_t message_len,
   if (output_type == OUTPUT_SYSLOG)
     return my_snprintf(message, message_len,
         "%.*s,%.*s,%.*s,%d,%lld,%s",
-        (unsigned int) serverhost_len, serverhost,
+        (int) serverhost_len, serverhost,
```
grooverdan, line 1401: why are we casting to int  - a size_t is `%zu`

grooverdan (commented):
Like the avoiding of casting in printf like output and the adjustment of the format string.

Where the attributes around format strings for the applicable functions going to be ported too?

plugin/server_audit/server_audit.c
ParadoxV5: Because per [the C/C++ standard](https://en.cppreference.com/w/c/io/fprintf#Parameters), the `*`s in format specifiers take `signed int` rather than `unsigned int` or `size_t`… ¯\\\_🤨_/¯

grooverdan: oops. Yes of course.

ParadoxV5: 🤔 The `my_snprintf` family is our own variant of `snprintf`, after all.
Is it worth to take `size_t` from `*`s, or is breaking from the Standard unjust?

ParadoxV5:
Wow, we are agile. I *just* rebased an hour ago and it’s now out-of-date already.

bnestere (approved):

ParadoxV5:
Sure, I’ll rebase as well.

@grooverdan merged the others without requesting commit message changes, though.

grooverdan:
Thank you very much for your patience and efforts in cleaning this up. Just a merge up to go...

PR #3486, opened 2024-08-28
MDEV-34823 Invalid arguments in ib_push_warning()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34823*
## Description
In the bug report MDEV-32817 it occurred that the function `row_mysql_get_table_status()` is outputting a `fil_space_t*` as if it were a numeric tablespace identifier. A closer inspection reveals more errors.

`ib_push_warning()`: Remove. Let us invoke `push_warning_printf()` directly.

`innodb_decryption_failed()`: Report a decryption failure and set the `dict_table_t::file_unreadable` flag. This code was being duplicated in very many places.

`innodb_fk_error()`: Report a `FOREIGN KEY` error.

`dict_foreign_def_get_fields()`: Remove. This code was being used in `dict_create_add_foreign_to_dictionary()` in an apparently uncovered code path. That `ib_push_warning()` call would pass the integer `i+1` instead of a pointer to NUL terminated string (`"%s"`), and therefore the call should have resulted in a crash.

`dict_print_info_on_foreign_key_in_create_format()`, `innobase_quote_identifier()`: Add `const` qualifiers.

`row_mysql_get_table_status()`: Display no message on `DB_CORRUPTION`; it should be properly reported at the SQL layer anyway.
## Release Notes
Some diagnostic output on error handling was simplified and corrected.
## How can this PR be tested?
Stress testing MariaDB Server 10.5 is somewhat challenging, because DDL is not crash safe and InnoDB would crash easily on corrupted data. That said, there are some `./mtr --suite=encryption` tests that do cover the `innodb_decryption_failed()`. One test also covers the `innodb_fk_error()`.
## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/innodb-fk-warnings.result
+++ b/mysql-test/suite/innodb/r/innodb-fk-warnings.result
@@ -16,7 +16,7 @@ CONSTRAINT test FOREIGN KEY (b) REFERENCES t2 (id)
 ERROR HY000: Can't create table `test`.`t2` (errno: 121 "Duplicate key on write or update")
 show warnings;
 Level	Code	Message
-Warning	121	Create or Alter table `test`.`t2` with foreign key constraint failed. Foreign key constraint `test`.`test` already exists on data dictionary. Foreign key constraint names need to be unique in database. Error in foreign key definition: CONSTRAINT `test` FOREIGN KEY (`b`) REFERENCES `test`.`t2` (`id`).
+Warning	121	CREATE or ALTER TABLE `test`.`t2` failed: duplicate name, CONSTRAINT `test` FOREIGN KEY (`b`) REFERENCES `t2` (`id`)
 Error	1005	Can't create table `test`.`t2` (errno: 121 "Duplicate key on write or update")
 Warning	1022	Can't write; duplicate key in table 't2'
 drop table t1;
--- a/storage/innobase/dict/dict0crea.cc
+++ b/storage/innobase/dict/dict0crea.cc
@@ -1727,104 +1727,6 @@ dict_create_add_foreign_field_to_dictionary(
 }
 
 /********************************************************************//**
-Construct foreign key constraint defintion from data dictionary information.
-*/
-UNIV_INTERN
-char*
-dict_foreign_def_get(
-/*=================*/
-	dict_foreign_t*	foreign,/*!< in: foreign */
-	trx_t*		trx)	/*!< in: trx */
-{
-	char* fk_def = (char *)mem_heap_alloc(foreign->heap, 4*1024);
-	const char* tbname;
-	char tablebuf[MAX_TABLE_NAME_LEN + 1] = "";
-	unsigned i;
-	char* bufend;
-
-	tbname = dict_remove_db_name(foreign->id);
-	bufend = innobase_convert_name(tablebuf, MAX_TABLE_NAME_LEN,
-				tbname, strlen(tbname), trx->mysql_thd);
-	tablebuf[bufend - tablebuf] = '\0';
-
-	sprintf(fk_def,
-		(char *)"CONSTRAINT %s FOREIGN KEY (", (char *)tablebuf);
-
-	for(i = 0; i < foreign->n_fields; i++) {
-		char	buf[MAX_TABLE_NAME_LEN + 1] = "";
-		innobase_convert_name(buf, MAX_TABLE_NAME_LEN,
-				foreign->foreign_col_names[i],
-				strlen(foreign->foreign_col_names[i]),
-				trx->mysql_thd);
-		strcat(fk_def, buf);
-		if (i < static_cast<unsigned>(foreign->n_fields-1)) {
-			strcat(fk_def, (char *)",");
-		}
-	}
-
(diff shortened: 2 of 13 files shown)


mysql-test/suite/innodb/r/innodb-fk-warnings.result
```diff
@@ -16,7 +16,7 @@ CONSTRAINT test FOREIGN KEY (b) REFERENCES t2 (id)
 ERROR HY000: Can't create table `test`.`t2` (errno: 121 "Duplicate key on write or update")
 show warnings;
 Level	Code	Message
-Warning	121	Create or Alter table `test`.`t2` with foreign key constraint failed. Foreign key constraint `test`.`test` already exists on data dictionary. Foreign key constraint names need to be unique in database. Error in foreign key definition: CONSTRAINT `test` FOREIGN KEY (`b`) REFERENCES `test`.`t2` (`id`).
+Warning	121	CREATE or ALTER TABLE `test`.`t2` failed, CONSTRAINT `test` FOREIGN KEY (`b`) REFERENCES `t2` (`id`)
```
mariadb-DebarunBanerjee, line 19: Looks like the precision of the error messages degraded. The reason for this failure is existing constraint named "test" and the current error message at least explains this part.

> Foreign key constraint `test`.`test` already exists on data dictionary. Foreign key constraint names need to be unique in database.

It usually is annoying when an error is thrown and I cannot make out what is the problem. Perhaps along with the general error/warning we can preserve the important information for each specific code path by pushing mote warning to the stack.

storage/innobase/dict/dict0crea.cc
```diff
@@ -1865,29 +1831,8 @@ dict_create_add_foreign_to_dictionary(
...
-
-			fk_def = dict_foreign_def_get((dict_foreign_t*)foreign, trx);
-
-			ib_push_warning(trx, error,
-				"Create or Alter table %s with foreign key constraint"
-				" failed. Foreign key constraint %s"
-				" already exists on data dictionary."
-				" Foreign key constraint names need to be unique in database."
-				" Error in foreign key definition: %s.",
-				tablename, buf, fk_def);
```
mariadb-DebarunBanerjee, line 1888: Can we preserver the important information in this warning (the same is reflected in the test result) along with the general error message in innodb_fk_error ? I think it is ok to have warnings pushed from each specific path with more details.

mariadb-DebarunBanerjee (changes requested):
I agree with the changes. Please see if we can do something about the lost details in the error message in the innodb_fk_error() paths.

mysql-test/suite/innodb/r/innodb-fk-warnings.result
dr-m: I thought that the information was redundant, because also the following errors would be emitted for the same statement:
```
Error	1005	Can't create table `test`.`t2` (errno: 121 "Duplicate key on write or update")
Warning	1022	Can't write; duplicate key in table 't2'
```
But, I now realize that it can be a little confusing. I think that we can improve this by adding `: duplicate name` right before the dump that starts with `, CONSTRAINT`.

storage/innobase/dict/dict0crea.cc
dr-m: This appears to be the only path that can actually output the foreign key error. The other call should always have crashed, because we are passing an integer to a `"%s"` format parameter.

But, I agree that it would be useful to somehow mention that the constraint name is already being used in the current schema.

storage/innobase/dict/dict0crea.cc
```diff
@@ -1865,29 +1831,8 @@ dict_create_add_foreign_to_dictionary(
...
-			char	tablename[MAX_TABLE_NAME_LEN + 1] = "";
-			char*	fk_def;
-
-			innobase_convert_name(tablename, MAX_TABLE_NAME_LEN,
-				name, strlen(name), trx->mysql_thd);
-
-			innobase_convert_name(buf, MAX_TABLE_NAME_LEN,
-				foreign->id, strlen(foreign->id), trx->mysql_thd);
-
-			fk_def = dict_foreign_def_get((dict_foreign_t*)foreign, trx);
```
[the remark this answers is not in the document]
dr-m: It is called by `dict_create_add_foreign_to_dictionary()`.

storage/innobase/row/row0mysql.cc
```diff
@@ -1259,49 +1259,25 @@ row_lock_table(row_prebuilt_t* prebuilt)
...
 @retval	DB_CORRUPTION		table is corrupted
 @retval	DB_TABLESPACE_NOT_FOUND	tablespace .ibd file not found */
-static
-dberr_t
-row_mysql_get_table_status(
-	const dict_table_t*	table,
-	trx_t*			trx,
-	bool 			push_warning = true)
+ATTRIBUTE_COLD
+static dberr_t row_mysql_get_table_status(trx_t *trx, dict_table_t *table)
```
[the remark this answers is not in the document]
dr-m: I agree. I will improve the comment too.

mariadb-DebarunBanerjee (approved):
Thanks. The changes look good to me.

PR #3487, opened 2024-08-29
MDEV-24813 Signal full scan to storage engines.

mariadb-YuchenPei:
This is an initial patch for discussion.

Will work in the following minimal example

CREATE TABLE t (a INT PRIMARY KEY) ENGINE=InnoDB;
insert into t values (42);

BEGIN;
SELECT * FROM t LOCK IN SHARE MODE; # will acquire S lock 
COMMIT;

DROP TABLE t;


storage/innobase/include/row0mysql.h
```diff
@@ -686,6 +686,8 @@ struct row_prebuilt_t {
 	/** The MySQL table object */
 	TABLE*		m_mysql_table;
 
+    bool full_table_lock=false;
```
dr-m, line 689: The declaration of an initializer expression would introduce an implicit constructor `row_prebuilt_t::row_prebuilt_t()`. That in turn could lead into trouble with the GCC 6 `-flifetime-dse`, because the rest of the fields would not be zero-initialized.

All fields will be zero-initialized in `row_create_prebuilt()` by invoking `mem_heap_zalloc()`.

The field name is misleading: it should be `full_table_scan`. It is also missing a preceding Doxygen comment that explains when the flag would be set and what its impact is.

mariadb-YuchenPei: Done.

sql/records.cc
```diff
@@ -338,8 +338,8 @@ bool init_read_record(READ_RECORD *info,THD *thd, TABLE *table,
   }
   /* Condition pushdown to storage engine */
   if ((table->file->ha_table_flags() & HA_CAN_TABLE_CONDITION_PUSHDOWN) &&
-      select && select->cond && 
-      (select->cond->used_tables() & table->map) &&
+      select && // select->cond &&
+      // (select->cond->used_tables() & table->map) &&
```
spetrunia: I wonder what was the need to comment out 
```
// (select->cond->used_tables() & table->map) &&
```
Why would a table $TBL have `select->cond` that has something that doesn't depend on $TBL? 
The only thing that comes to mind is:

```sql
select * from t1 where RAND() < 0.5
```
IIRC conditions with RAND_TABLE_BIT are attached to the last table in the join order...

spetrunia:
Also, I've tried this testcase
```sql
create table t10 (a int, b varchar(100), index(a));
insert into t10 select seq, uuid() from seq_1_to_10000;
create table t11(a int);
explain insert into t11 select a from t10;
insert into t11 select a from t10;
```
and init_read_record()  isn't invoked for it.

The index scan on t10 is initialized here:

```
#0  ha_innobase::index_init (this=0x7fffb80b6eb0, keynr=0)   at /home/psergey/dev-git/12.3-no-lock/storage/innobase/handler/ha_innodb.cc:8767
#1  0x0000555555f0a5a0 in handler::ha_index_init (this=0x7fffb80b6eb0,     idx=0, sorted=true)    at /home/psergey/dev-git/12.3-no-lock/sql/handler.h:3527
#2  0x0000555556125194 in join_read_first (tab=0x7fffb80cae40)    at /home/psergey/dev-git/12.3-no-lock/sql/sql_select.cc:24768
```

spetrunia:
With this patch, we have "Using where with pushed condition" all over the place:

```
MariaDB [test]> explain select * from t1 where b <1;
+------+-------------+-------+------+---------------+------+---------+------+-------+-----------------------------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows  | Extra                             |
+------+-------------+-------+------+---------------+------+---------+------+-------+-----------------------------------+
|    1 | SIMPLE      | t1    | ALL  | NULL          | NULL | NULL    | NULL | 10157 | Using where with pushed condition |
+------+-------------+-------+------+---------------+------+---------+------+-------+-----------------------------------+
1 row in set (0.002 sec)
```

sql/records.cc
mariadb-YuchenPei: No longer commenting out these in the new patch

mariadb-YuchenPei:
> Also, I've tried this testcase
> 
> ```sql
> create table t10 (a int, b varchar(100), index(a));
> insert into t10 select seq, uuid() from seq_1_to_10000;
> create table t11(a int);
> explain insert into t11 select a from t10;
> insert into t11 select a from t10;
> ```
> 
> and init_read_record() isn't invoked for it.
> 
> The index scan on t10 is initialized here:
> 
> ```
> #0  ha_innobase::index_init (this=0x7fffb80b6eb0, keynr=0)   at /home/psergey/dev-git/12.3-no-lock/storage/innobase/handler/ha_innodb.cc:8767
> #1  0x0000555555f0a5a0 in handler::ha_index_init (this=0x7fffb80b6eb0,     idx=0, sorted=true)    at /home/psergey/dev-git/12.3-no-lock/sql/handler.h:3527
> #2  0x0000555556125194 in join_read_first (tab=0x7fffb80cae40)    at /home/psergey/dev-git/12.3-no-lock/sql/sql_select.cc:24768
> ```

You are right. I had forgotten to make the call in join_read_first (and join_read_last). Added the calls to ha_extra in a new commit now.

mariadb-YuchenPei:
> With this patch, we have "Using where with pushed condition" all over the place:

No longer the case with the new patch that does not comment out the cond_push if condition

mysql-test/main/innodb_full_scan.test
```diff
@@ -0,0 +1,211 @@
...
+SELECT * FROM t LOCK IN SHARE MODE;
+COMMIT;
+--echo ## +0, was +1
+eval $print_lock_rec_lock_created;
+
+BEGIN;
+SELECT * FROM t for update;
+COMMIT;
+--echo ## +0, was +1
+eval $print_lock_rec_lock_created;
```
iMineLink: I wonder if this can be changed so that it captures the delta from the previous "call"

mysql-test/main/innodb_full_scan.test
```diff
@@ -0,0 +1,211 @@
+--source include/have_sequence.inc
+--source include/have_innodb.inc
+--source include/no_view_protocol.inc
+--echo #
+--echo # MDEV-24813 Locking full table scan fails to use table-level locking
+--echo #
+
+set @old_innodb_table_lock_on_full_scan=@@innodb_table_lock_on_full_scan;
+set innodb_table_lock_on_full_scan=1;
```
iMineLink: Can this be tested with combinations instead, so that the `.rdiff` shows the decrease in `lock_rec_lock_created` counts?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15940,6 +15944,31 @@ ha_innobase::extra(
 	return(0);
 }
 
+int
+ha_innobase::extra_opt(
+/*===============*/
+	enum ha_extra_function operation,
+			   /*!< in: HA_EXTRA_FLUSH or some other flag */
```
iMineLink: Maybe stale comment?

sql/sql_select.cc
```diff
@@ -25338,6 +25338,7 @@ int test_if_use_dynamic_range_scan(JOIN_TAB *join_tab)
 int join_init_read_record(JOIN_TAB *tab)
 {
   bool need_unpacking= FALSE;
+  /* TODO: s/tab->join/join/g in this function */
```
iMineLink, line 25741: What is this about?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -864,6 +864,10 @@ static MYSQL_THDVAR_BOOL(ft_enable_stopword, PLUGIN_VAR_OPCMDARG,
   NULL, NULL,
   /* default */ TRUE);
 
+static MYSQL_THDVAR_BOOL(table_lock_on_full_scan, PLUGIN_VAR_OPCMDARG,
+  "Whether to acquire table lock when SQL layer advises full scan",
```
iMineLink: Maybe add a reference to the fact that table lock is taken in place of row-locks for each row, so there is performance gain at the cost of less granular locking?

iMineLink (commented):
Please check my few minor comments, especially the ones around `innodb_full_scan.test`, thanks.

sql/sql_select.cc
mariadb-YuchenPei: just a minor cleanup todo, nothing to do with innodb

iMineLink:
Thanks for addressing my review comments, LGTM.

iMineLink (approved):

mysql-test/suite/innodb/r/autoinc_debug,table_lock.rdiff
```diff
@@ -0,0 +1,70 @@
+--- /home/ycp/source/mariadb-server/push-main/src/mysql-test/suite/innodb/r/autoinc_debug.result	2026-06-17 14:13:34.462383376 +1000
++++ /home/ycp/source/mariadb-server/push-main/src/mysql-test/suite/innodb/r/autoinc_debug,table_lock.reject	2026-06-17 16:11:38.709757127 +1000
```
dr-m: Please follow the style of 88d9348dfc494105046451beef15a8e68d7f51be by omitting timestamps as well as absolute path names.

dr-m (changes requested):

mysql-test/suite/innodb/r/autoinc_debug,table_lock.rdiff
mariadb-YuchenPei: sure that can be done, but why?

mariadb-YuchenPei: done

PR #3494, opened 2024-09-02
MDEV-34855  Bootstrap hangs while shrinking the system tablespace 

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV- 34855*

## Description
Reason:
 - During bootstrap, InnoDB shrinks the system tablespace while purge is
 still active on system tablespace. This could lead to deadlock.
 
Fix:
   - Avoid System tablespace shrinking during bootstrap.
   -  Move the shrinking logic of system tablespace after purge gets shutdown.
 
## How can this PR be tested?
Patch to test this scenario during bootstrap
```
diff --git a/storage/innobase/fsp/fsp0fsp.cc b/storage/innobase/fsp/fsp0fsp.cc
index 64f22c1cbe8..ec604a81315 100644
--- a/storage/innobase/fsp/fsp0fsp.cc
+++ b/storage/innobase/fsp/fsp0fsp.cc
@@ -40,6 +40,7 @@ Created 11/29/1995 Heikki Tuuri
 #include "dict0mem.h"
 #include "fsp0types.h"
 #include "log.h"
+#include "trx0purge.h"
 
 /** Returns the first extent descriptor for a segment.
 We think of the extent lists of the segment catenated in the order
@@ -3598,6 +3599,7 @@ void fsp_system_tablespace_truncate()
     /* Tablespace is being used within fixed size */
     return;
 
+  ut_ad(!purge_sys.enabled());
   /* Set fixed size as threshold to truncate */
   if (fixed_size > last_used_extent)
     last_used_extent= fixed_size;
```

Test case to check whether the purge system is active during shrinking of system tablespace.
```
.opt file:
--innodb_sys_tablespaces
--innodb_purge_threads=1
--innodb_page_size=64k
--innodb_data_file_path=ibdata1:3M:autoextend
```
Test case:
========
```
create table t1(f1 int not null)engine=innodb;
drop table t1;
```

## PR quality check

### DIFF
--- a/storage/innobase/fsp/fsp0fsp.cc
+++ b/storage/innobase/fsp/fsp0fsp.cc
@@ -40,6 +40,9 @@ Created 11/29/1995 Heikki Tuuri
 #include "dict0mem.h"
 #include "fsp0types.h"
 #include "log.h"
+#ifndef DBUG_OFF
+# include "trx0purge.h"
+#endif
 
 /** Returns the first extent descriptor for a segment.
 We think of the extent lists of the segment catenated in the order
@@ -3576,6 +3579,8 @@ dberr_t fsp_sys_tablespace_validate()
 
 void fsp_system_tablespace_truncate()
 {
+  ut_ad(!purge_sys.enabled());
+  ut_ad(!srv_undo_sources);
   uint32_t last_used_extent= 0;
   fil_space_t *space= fil_system.sys_space;
   mtr_t mtr;
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -4283,12 +4283,6 @@ innobase_end(handlerton*, ha_panic_function)
 		 	}
 		}
 
-		/* Do system tablespace truncation during slow shutdown */
-		if (!srv_fast_shutdown && !high_level_read_only
-		    && srv_operation == SRV_OPERATION_NORMAL) {
-			fsp_system_tablespace_truncate();
-		}
-
 		innodb_shutdown();
 		mysql_mutex_destroy(&log_requests.mutex);
 	}
--- a/storage/innobase/srv/srv0srv.cc
+++ b/storage/innobase/srv/srv0srv.cc
@@ -1577,5 +1577,8 @@ void srv_purge_shutdown()
     }
     purge_sys.coordinator_shutdown();
     srv_shutdown_purge_tasks();
+    if (!srv_fast_shutdown && !high_level_read_only && srv_was_started &&
+        !opt_bootstrap && srv_operation == SRV_OPERATION_NORMAL)
+      fsp_system_tablespace_truncate();
   }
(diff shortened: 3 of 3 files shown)


storage/innobase/srv/srv0start.cc
```diff
@@ -1994,6 +1994,11 @@ void innodb_preshutdown()
   srv_shutdown_bg_undo_sources();
   srv_purge_shutdown();
 
+  /* Do system tablespace truncation during slow shutdown */
+  if (!srv_fast_shutdown && !high_level_read_only && srv_was_started
+      && srv_operation == SRV_OPERATION_NORMAL)
+    fsp_system_tablespace_truncate();
```
dr-m, line 2000: It is unclear to me how things are improved when this piece of code is executed _earlier_. The pull request mentions that the test `innodb.sys_truncate_shutdown` covers this code, but the test was not extended to demonstrate the impact of this. Is there any (even nondeterministic) way of demonstrating the impact?

Lines should customarily be split _after_, not _before_ binary operators, such as `&&`. Only the archaic InnoDB formatting style followed that convention.

dr-m: I now see that the original code was misplaced, invoked before `innodb_shutdown()`, which would invoke `logs_empty_and_mark_files_at_shutdown()`, which would seem to be a safer place to invoke the code.

An even more natural place for this code might be in `srv_purge_shutdown()`, which is handling `!srv_fast_shutdown` as a special case due to the logic in `srv_purge_should_exit()`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3598,6 +3601,7 @@ void fsp_system_tablespace_truncate()
     /* Tablespace is being used within fixed size */
     return;
 
+  ut_ad(!purge_sys.enabled());
```
dr-m, line 3604: Shouldn’t this assertion be located at the start of this function?
Can we also assert `!srv_undo_sources` and that no rollback is in progress?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -40,6 +40,9 @@ Created 11/29/1995 Heikki Tuuri
 #include "dict0mem.h"
 #include "fsp0types.h"
 #include "log.h"
+#ifdef UNIV_DEBUG
+#include "trx0purge.h"
+#endif
```
dr-m, line 45: I’d suggest to indent conditional preprocessor directives for better readability:
```c++
#ifndef DBUG_OFF
# include "trx0purge.h"
#endif
```

dr-m (commented):
I confirmed that the added assertion would fail on server bootstrap if the fix is not present.

Because there are two calls to `fsp_system_tablespace_truncate()`, I think that it does make sense to have that assertion.

dr-m (approved):

PR #3500, opened 2024-09-04
MDEV-34869 ssl_cipher parameter cannot configure TLSv1.3 and TLSv1.2 ciphers at the same time

tonychen2001:
- [x] *The Jira issue number for this PR is: MDEV-34869*

## Description

Allow users to configure exact ciphers supported for all TLS versions including TLSv1.3 ciphers through the `ssl_cipher` server system variables. This configuration ability is supported by other databases and it may be necessary to meet compliance for certain security requirements/regulations.

## Release Notes

ssl_cipher supports configuring TLSv1.3 cipher suites. Note that if no TLSv1.3 ciphers are specified to `--ssl-cipher`, then the TLSv1.3 ciphers enabled by default will still be supported.

Pages that needs updates:
- https://mariadb.com/kb/en/ssltls-system-variables/#ssl_cipher
- https://mariadb.com/kb/en/using-tlsv13/

## How can this PR be tested?

Test cases were added to the main.ssl_cipher MTR.

## PR quality check

---

### DIFF
--- a/mysql-test/include/have_tlsv13.inc
+++ b/mysql-test/include/have_tlsv13.inc
@@ -1,0 +1,10 @@
+--disable_query_log
+connect (ssl_connection,localhost,root,,,,,SSL);
+
+if (`SELECT VARIABLE_VALUE NOT LIKE 'TLSv1.3' FROM information_schema.SESSION_STATUS WHERE VARIABLE_NAME = 'ssl_version'`) {
+  skip Needs TLSv1.3;
+}
+
+disconnect ssl_connection;
+connection default;
+--enable_query_log
--- a/mysql-test/main/tlsv13.test
+++ b/mysql-test/main/tlsv13.test
@@ -1,0 +1,43 @@
+--source include/have_ssl_communication.inc
+--source include/require_openssl_client.inc
+--source include/have_tlsv13.inc
+
+#
+# BUG - SSL_CIPHER SYSTEM VARIABLE CANNOT CONFIGURE TLSV1.3 AND TLSV1.2 CIPHERS AT THE SAME TIME
+#
+# If users specify TLSv1.3 and TLSv1.2 ciphers, it will only configure the
+# TLSv1.3 ciphers correctly but then it will keep all TLSv1.2 ciphers enabled.
+#
+# This is a potential security vulnerability because users trying to restrict
+# secure TLSv1.3 and TLSv1.2 ciphers will unintentionally have insecure TLSv1.2
+# ciphers enabled on the database.
+#
+let $restart_parameters=--ssl-cipher=TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384;
+source include/restart_mysqld.inc;
+
+--exec $MYSQL --host=localhost --ssl-cipher=TLS_AES_128_GCM_SHA256 --tls-version=TLSv1.3 -e "SHOW STATUS LIKE 'Ssl_cipher%';"
+--exec $MYSQL --host=localhost --ssl-cipher=ECDHE-RSA-AES256-GCM-SHA384 --tls-version=TLSv1.2 -e "SHOW STATUS LIKE 'Ssl_cipher%';"
+
+# Check that other ciphers are correctly not supported by the server
+--replace_regex /sslv3 alert handshake failure/ssl\/tls alert handshake failure/
+--error 1
+--exec $MYSQL --host=localhost --ssl-cipher=TLS_AES_256_GCM_SHA384 --tls-version=TLSv1.3 -e "SHOW STATUS LIKE 'Ssl_cipher';" 2>&1
+
+--replace_regex /sslv3 alert handshake failure/ssl\/tls alert handshake failure/
+--error 1
+--exec $MYSQL --host=localhost --ssl-cipher=ECDHE-RSA-AES128-GCM-SHA256 --tls-version=TLSv1.2 -e "SHOW STATUS LIKE 'Ssl_cipher';" 2>&1
+
+# TLSv1.3 ciphers are still supported if only TLSv1.2 ciphers are passed to --ssl-cipher
+let $restart_parameters=--ssl-cipher=ECDHE-RSA-AES256-GCM-SHA384;
+source include/restart_mysqld.inc;
+--exec $MYSQL --host=localhost --ssl-cipher=TLS_AES_128_GCM_SHA256 --tls-version=TLSv1.3 -e "SHOW STATUS LIKE 'Ssl_cipher';"
+--exec $MYSQL --host=localhost --ssl-cipher=ECDHE-RSA-AES256-GCM-SHA384 --tls-version=TLSv1.2 -e "SHOW STATUS LIKE 'Ssl_cipher';"
+--error 1
+--exec $MYSQL --host=localhost --ssl-cipher=ECDHE-RSA-AES128-GCM-SHA256 --tls-version=TLSv1.2 -e "SHOW STATUS LIKE 'Ssl_cipher';"
+
+# TLSv1.2 ciphers are not supported if only TLSv1.3 ciphers are passed to --ssl-cipher
+let $restart_parameters=--ssl-cipher=TLS_AES_128_GCM_SHA256;
+source include/restart_mysqld.inc;
+--exec $MYSQL --host=localhost --ssl-cipher=TLS_AES_128_GCM_SHA256 --tls-version=TLSv1.3 -e "SHOW STATUS LIKE 'Ssl_cipher%';"
+--error 1
+--exec $MYSQL --host=localhost --ssl-cipher=ECDHE-RSA-AES128-GCM-SHA256 --tls-version=TLSv1.2 -e "SHOW STATUS LIKE 'Ssl_cipher';"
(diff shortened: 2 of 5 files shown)


tonychen2001:
`amd64-debian-11-msan` and `amd64-debian-11-msan-clang-16` is failing with:
```
ERROR 2026 (HY000): TLS/SSL error: No or insufficient priorities were set.
mysqltest: At line 21: exec of '/home/buildbot/amd64-debian-11-msan/build/client//mariadb --defaults-file=/home/buildbot/amd64-debian-11-msan/build/mysql-test/var/9/my.cnf --host=localhost --ssl-cipher=TLS_AES_128_GCM_SHA256 --tls-version=TLSv1.3 -e "SHOW STATUS LIKE 'Ssl_cipher%';"' failed, error: 256, status: 1, errno: 11
Output from before failure:
# restart: --ssl-cipher=TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384
```

Not exactly sure what is causing this. Probably related to how OpenSSL is configured on the system.

---

`amd64-ubuntu-2004-debug` is failing with:
```
mysqltest: At line 27: command "$MYSQL --host=localhost --ssl-cipher=TLS_AES_256_GCM_SHA384 --tls-version=TLSv1.3 -e "SHOW STATUS LIKE 'Ssl_cipher';" 2>&1" succeeded - should have failed with errno 1...
The result from queries just before the failure was:
# restart: --ssl-cipher=TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384
Variable_name	Value
Ssl_cipher	TLS_AES_128_GCM_SHA256
Ssl_cipher_list	TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384
Variable_name	Value
Ssl_cipher	ECDHE-RSA-AES256-GCM-SHA384
Ssl_cipher_list	TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384
```

Seems that configuring TLSv13 ciphers has no effect on this worker.

---

Looks like some systems allow configuring TLSv1.3 cipher suites and some don't.

LinuxJedi:
We of course want good TLSv1.3 support. The Debian errors are GNUTLS by the look of it, not OpenSSL. These issues are going to take some time to go through.

mysql-test/main/tlsv13.test
```diff
@@ -0,0 +1,46 @@
+--source include/have_ssl_communication.inc
+--source include/have_tlsv13.inc
```
vaintroub, line 2: Also add a line 
```
source include/require_openssl_client.inc
```
gnutls or schannel clients do not work like openssl ones, they will not use cipher with TLSv1.3

mysql-test/main/tlsv13.test
```diff
@@ -0,0 +1,46 @@
...
+# If users specify TLSv1.3 and TLSv1.2 ciphers, it will only configure the
+# TLSv1.3 ciphers correctly but then it will keep all TLSv1.2 ciphers enabled.
+#
+# This is a potential security vulnerability because users trying to restrict
+# secure TLSv1.3 and TLSv1.2 ciphers will unintentionally have insecure TLSv1.2
+# ciphers enabled on the database.
+#
+let $restart_parameters=--ssl-cipher=TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384;
+source include/restart_mysqld.inc;
+# The builder, amd64-ubuntu-2004-debug, doesn't allow configuring TLSv1.3 cipher suites
```
vaintroub, line 16: I think this comment is misleading. And the SQL below, too. I can't figure out what this test is doing, however I see that ssl_cipher_list uses doublequotes, not singlequotes which one would expect.

Can you explain the intention? It still fails on amd64-ubuntu-2004-debug, and I'm sure why, but I see it uses some older OpenSSL 1.1.1f

vaintroub (changes requested):
I commented on places that need fixing, and also asked a question about the "TLSv1.3 is not configurable", in the test.
Could you please fix it and answer that question?
Thanks!

mysql-test/main/tlsv13.test
tonychen2001: Added, thanks!

mysql-test/main/tlsv13.test
```diff
@@ -0,0 +1,46 @@
...
+if (`SELECT VARIABLE_VALUE LIKE '%TLS_AES_256_GCM_SHA384%' from information_schema.GLOBAL_STATUS WHERE VARIABLE_NAME="ssl_cipher_list"`) {
+  skip TLSv1.3 is not configurable;
+}
+
+--exec $MYSQL --host=localhost --ssl-cipher=TLS_AES_128_GCM_SHA256 --tls-version=TLSv1.3 -e "SHOW STATUS LIKE 'Ssl_cipher%';"
```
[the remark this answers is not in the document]
tonychen2001: I've added `--source include/require_openssl_client.inc` to this test.

Correct me if I'm wrong but I believe this should disable this test for Schannel based clients.

mysql-test/include/have_tlsv13.inc
```diff
@@ -0,0 +1,11 @@
+--disable_query_log
+create table test.tls_version(version VARCHAR(20));
```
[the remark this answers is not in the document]
tonychen2001: I think before MariaDB 10.11, the connections are non-TLS by default which resulted in the `ssl_version` status variable being empty in the conditional query.

Not sure why, but I didn't think of switching to current connection to TLS and instead opted for this workaround.

I've changed this to no longer create a table.

mysql-test/main/tlsv13.test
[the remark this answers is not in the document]
vaintroub: require_openssl_client.inc disables this for schannel based clients, and for gnutls based clients

mysql-test/include/have_tlsv13.inc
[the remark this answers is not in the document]
vaintroub: Yes, thanks. Right, in the older versions connections were not TLS by default.

mysql-test/main/tlsv13.test
tonychen2001: I was observing that the TLSv1.3 cipher suites were not being modified. I thought it might have been something specific with the environment of that particular worker, so this was an attempt at skipping the test on that worker.

> but I see it uses some older OpenSSL 1.1.1f

Thanks for the info. I tested with MariaDB on OpenSSL 1.1.1.
It seems that on OpenSSL 1.1.1, if `ssl_cipher` includes a TLSv1.2 cipher in the list passed to `SSL_CTX_set_ciphersuites()`, it will result in an error and won't modify the available TLSv1.3 cipher list for the context.

So with OpenSSL 1.1.1, setting `--ssl-cipher=TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384` will error on the call to `SSL_CTX_set_ciphersuites()` but succeed on the call to `SSL_CTX_set_cipher_list()`.
Thus the resulting `Ssl_cipher_list` will be `TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384` instead of `TLS_AES_128_GCM_SHA256:ECDHE-RSA-AES256-GCM-SHA384`.

vaintroub:
Now , now the tests look better.

I understood the intention of testing on Ubuntu 20.04. Not sure what went wrong with it, however it uses OpenSSL version which is already EOL, i.e 1.1.1. Tests can be skipped also outside of the test itself, and we have several of examples of that in mysql-test/suite.pm

I'd suggest to go this way, since it is easier. Maybe you can add the snippet below to your patch, and let buildbot test it ,wat do you think?


```diff
diff --git a/mysql-test/suite.pm b/mysql-test/suite.pm
index d5c4855ce6e..15734cb6007 100644
--- a/mysql-test/suite.pm
+++ b/mysql-test/suite.pm
@@ -71,6 +71,9 @@ sub skip_combinations {
   $skip{'main/ssl_7937.combinations'} = [ 'x509v3' ]
     unless $ssl_lib =~ /WolfSSL/ or $openssl_ver ge "1.0.2";

+  $skip{'main/tlsv13.test'} = 'OpenSSL too old'
+    unless $ssl_lib =~ /WolfSSL/ or $openssl_ver ge "3.0.0";
+
   $skip{'main/ssl_verify_ip.test'} = 'x509v3 support required'
     unless $openssl_ver ge "1.0.2";
```

mysql-test/include/have_tlsv13.inc
```diff
@@ -0,0 +1,10 @@
+--disable_query_log
+connect (ssl_connection,localhost,root,,,,,SSL);
+
+if (`SELECT VARIABLE_VALUE NOT LIKE 'TLSv1.3' FROM information_schema.GLOBAL_STATUS WHERE VARIABLE_NAME = 'ssl_version'`) {
```
vaintroub, line 4: Would information_schema.SESSION_STATUS work? If yes, I'd prefer that.
It *is* a session level status variable, so I would not use GLOBAL_STATUS, unless necessary.

tonychen2001: Got it, updated it to `information_schema.SESSION_STATUS`. Thanks

vaintroub (approved):

PR #3503, opened 2024-09-06
MDEV-36000 - Add major version upgrade testing from previous versions to mysql_upgrade.test

anson1014:
### **Add major version upgrade testing from previous versions to mysql_upgrade.test**

Previously, only upgrades from 10.0-10.4 dumps were included in the mysql_upgrade.test file. Include dumps from versions 10.5-11.3 as 11.4 is the current LTS and upgrades from more recent versions are arguably more common/important and have a clear need to continually be validated via this MTR.

For the new versions added (10.5,10.6,10.11,11.1,11.2,11.3), they are either prior LTS versions or their difference in schemas is not insignificant and therefore is valuable to be included for test coverage. Certain intermediate versions are omitted in the interest of avoiding bloating the size of the git repository. For versions 10.7,10.8,10.9 and 10.10, the difference in their schemas in comparison to 10.11 are extremely minimal if present at all. Therefore, the inclusion of these dumps are redundant. Testing of these versions will not provide significant additional marginal value. For these same reasons, 11.0 is also not included as there is no major difference between it and 11.1.

As done previously in f67522e, these dumps were generated with an empty root password as to avoid using unix_socket. Furthermore, locks on the log tables were also omitted. Both of these changes were done in order to prevent errors failing the upgrade during the MTR. Help tables are also omitted from the dump file as they do not affect the upgrade and more importantly, for the sake of minimizing the size of the file.

This PR merely fills the apparent gaps of using the existing functionality of loading a dump from a different version of MariaDB to ensure that an upgrade completes successfully.

Suggestions on ways to further extend, improve or augment major version upgrade testing would also be appreciated.

## Release Notes

N/A

## How can this PR be tested?

Only changes are to mysql_upgrade.test, which passes successfully.

```shellscript

worker[01] mysql-test-run: WARNING: running this script as _root_ will cause some tests to be skipped
main.mysql_upgrade                       [ pass ]  52384
The servers were restarted 0 times
Spent 52.384 of 58 seconds executing testcases

Completed: All 1 tests were successful.
```

- [x] _This is a change in testing that is targeted towards the current MariaDB LTS._

## PR quality check

---


LinuxJedi:
The 11.3 one is nearly 3x larger than the rest. It is difficult to review on GitHub, so is there any insight into why?

PR #3504, opened 2024-09-09
 MDEV-33894 / 34825 FreeBSD - few InnoDB things are innodb, cmake/os/FreeBSD lost in merge

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-33894/ MDEV-34825*

## Description

FreeBSD on merge to 10.11 got warnings and merge errors:

https://buildbot.mariadb.org/#/builders/619/builds/6641/steps/5/logs/stdio

## Release Notes

merge fix - nothing required.

## How can this PR be tested?

FreeBSD BB worker will build without InnoDB warnings. Other Linux/Windows builder will also build.

## PR quality check

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -1095,13 +1095,11 @@ static ATTRIBUTE_COLD void os_file_log_buffered()
   log_sys.log_maybe_unbuffered= false;
   log_sys.log_buffered= true;
 }
-# endif
 
 /** @return whether the log file may work with unbuffered I/O. */
 static ATTRIBUTE_COLD bool os_file_log_maybe_unbuffered(const struct stat &st)
 {
   MSAN_STAT_WORKAROUND(&st);
-# ifdef __linux__
   char b[20 + sizeof "/sys/dev/block/" ":" "/../queue/physical_block_size"];
   if (snprintf(b, sizeof b, "/sys/dev/block/%u:%u/queue/physical_block_size",
                major(st.st_dev), minor(st.st_dev)) >=
@@ -1133,13 +1131,11 @@ static ATTRIBUTE_COLD bool os_file_log_maybe_unbuffered(const struct stat &st)
   if (s > 4096 || s < 64 || !ut_is_2pow(s))
     return false;
   log_sys.set_block_size(uint32_t(s));
-# else
-  constexpr unsigned long s= 4096;
-# endif
 
   return !(st.st_size & (s - 1));
 }
-#endif
+# endif /* __linux__ */
+#endif /* O_DIRECT */
 
 /** NOTE! Use the corresponding macro os_file_create(), not directly
 this function!
@@ -1192,7 +1188,9 @@ os_file_create_func(
 	}
 
 #ifdef O_DIRECT
+# ifdef __linux__
 	struct stat st;
+# endif
 	ut_a(type == OS_LOG_FILE
 	     || type == OS_DATA_FILE || type == OS_DATA_FILE_NO_O_DIRECT);
 	int direct_flag = 0;
--- a/cmake/os/FreeBSD.cmake
+++ b/cmake/os/FreeBSD.cmake
@@ -1,0 +1,19 @@
+# Copyright (C) 2024 MariaDB Foundation
+# 
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; version 2 of the License.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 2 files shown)


storage/innobase/os/os0file.cc
```diff
@@ -1095,13 +1095,11 @@ static ATTRIBUTE_COLD void os_file_log_buffered()
...
-# endif
 
 /** @return whether the log file may work with unbuffered I/O. */
 static ATTRIBUTE_COLD bool os_file_log_maybe_unbuffered(const struct stat &st)
 {
   MSAN_STAT_WORKAROUND(&st);
-# ifdef __linux__
   char b[20 + sizeof "/sys/dev/block/" ":" "/../queue/physical_block_size"];
   if (snprintf(b, sizeof b, "/sys/dev/block/%u:%u/queue/physical_block_size",
                major(st.st_dev), minor(st.st_dev)) >=
```
dr-m, line 1105: Do all operating systems that support `O_DIRECT` provide a Linux-like interface for determining the physical block size of a storage device?

dr-m (changes requested):
Thank you, it is great to have FreeBSD back.

The first commit message needs to identify the incorrect merge d002b1f503a13ed654184cf91c61833657a4b60a. I wonder if everything else in that merge was correct. @mariadb-YuchenPei probably forgot to check that identified conflict in detail:
```
	deleted by us:   cmake/os/FreeBSD.cmake
	both modified:   mysql-test/main/mysqld--help.result
	both modified:   sql/sql_delete.cc
	both modified:   sql/sql_select.cc
	both modified:   sql/sql_update.cc
```
For the other files, there are several commits by others that may be behind the conflicts. Those may have to be reviewed as well.
```sh
git log d002b1f503a13ed654184cf91c61833657a4b60a~..d002b1f503a13ed654184cf91c61833657a4b60a^2 mysql-test/main/mysqld--help.result sql/sql_delete.cc sql/sql_select.cc sql/sql_update.cc
```

storage/innobase/os/os0file.cc
mariadb-YuchenPei: > It would be good to mention why https://github.com/MariaDB/server/commit/5369df741b3a87d7b262aafe95438b33ed124295 was not applied to 10.6 or 10.5.

This looks more likely to be the problem than the merge itself, as it is natural to keep a file deleted in a conflict of merging from a lower version to a higher version, if the lower version modifies the file and the file is deleted in a higher version.

dr-m: Even if a cherry-pick of 5369df741b3a87d7b262aafe95438b33ed124295 had been included in the merge, `git merge` would have flagged the conflict as "deleted by us". Such conflicts can be understood by reading the output of `git log HEAD..MERGE_HEAD filename` or `git log MERGE_HEAD..HEAD filename` and resolved by studying the identified commits in more detail.

storage/innobase/os/os0file.cc
```diff
@@ -1095,13 +1095,11 @@ static ATTRIBUTE_COLD void os_file_log_buffered()
   log_sys.log_maybe_unbuffered= false;
   log_sys.log_buffered= true;
 }
-# endif
 
 /** @return whether the log file may work with unbuffered I/O. */
 static ATTRIBUTE_COLD bool os_file_log_maybe_unbuffered(const struct stat &st)
 {
   MSAN_STAT_WORKAROUND(&st);
-# ifdef __linux__
```
dr-m, line 1104: Finally, I understood it. The first `#endif` removal makes the function `os_file_log_maybe_unbuffered()` defined only `#ifdef __linux__`. That is acceptable, and I see that the FreeBSD run passed.

storage/innobase/os/os0file.cc
```diff
@@ -1133,13 +1131,11 @@ static ATTRIBUTE_COLD bool os_file_log_maybe_unbuffered(const struct stat &st)
...
   log_sys.set_block_size(uint32_t(s));
-# else
-  constexpr unsigned long s= 4096;
-# endif
 
   return !(st.st_size & (s - 1));
 }
-#endif
+#endif /* __linux__ */
+#endif /* O_DIRECT */
```
dr-m, line 1138: I would suggest to add the missing indentation, either instead or in addition to adding those comments:
```c
# endif /* __linux__ */
#endif /* O_DIRECT */
```

dr-m (approved):
Sorry, I misread the initial fix.

The first line in the revised first commit message is somewhat backwards. I think that both the 10.5 commit and the incorrect 10.11 merge commit need to be mentioned in the same sentence. There is also a stray backtick at the end of that commit.

PR #3505, opened 2024-09-09
MDEV-34879 InnoDB fails to merge the change buffer to ROW_FORMAT=COMPRESSED tables

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34879*
## Description
`buf_page_t::read_complete()`: Fix an incorrect condition that had been added in commit aaef2e1d8c843d1e40b1ce0c5199c3abb3c5da28 (MDEV-27058). Also for compressed-only pages we must remember that buffered changes may exist.

`buf_read_page()`: Correct the function comment; this is for a synchronous and not asynchronous read. Pass the parameter `unzip=true` to `buf_read_page_low()`, because each of our callers will be interested in the uncompressed page frame.
## Release Notes
InnoDB could corrupt secondary index pages of `ROW_FORMAT=COMPRESSED` tables unless `innodb_change_buffering=none` had never been used.
## How can this PR be tested?
Stress test with `innodb_change_buffering=all`, including server restarts. This fix was tested on a copy of a data directory on which the failure was originally reproduced.

By its design, the change buffer is very difficult to cover with tests. It is also disabled by default (5c46751f238ee8dcef1e718ac5f63952bff5d09d) and has been removed in later major versions (f27e9c894779a4c7ebe6446ba9aa408f1771c114). That said, the test `innodb_zip.wl5522-zip` kind of does cover these changes. It would hit an infinite loop due to a bug that was revealed in `buf_pool_t::page_fix()`: If the page exists in the buffer pool in compressed-only format but needs a change buffer merge, the current thread must proceed to decompress the block.

The change to `buf_read_page()` is reflected by the test `encryption.innodb-compressed-blob`, where we would start to emit more errors due to `page_zip_decompress()` failing after mis-decrypting the page.
## PR quality check

### DIFF
--- a/mysql-test/suite/encryption/r/innodb-compressed-blob.result
+++ b/mysql-test/suite/encryption/r/innodb-compressed-blob.result
@@ -1,6 +1,8 @@
 call mtr.add_suppression("InnoDB: The page \\[page id: space=[1-9][0-9]*, page number=[1-9][0-9]*\\] in file '.*test.t[123]\\.ibd' cannot be decrypted; key_version=1");
 call mtr.add_suppression("InnoDB: Recovery failed to read page");
-call mtr.add_suppression("InnoDB: Unable to decompress ..test.t[1-3]\\.ibd\\[page id: space=[1-9][0-9]*, page number=[0-9]+\\]");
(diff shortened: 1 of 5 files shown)


mariadb-DebarunBanerjee (approved):
Looks good to me.

PR #3506, opened 2024-09-09
MDEV-33500 (part 2): rpl.rpl_parallel_sbm can still fail

bnestere:
The failing test case validates Seconds_Behind_Master for a delayed slave, while STOP SLAVE is executed during a delay. The test fixes initially added to the test (commit b04c8575967) added a table lock to ensure a transaction could not finish before validating the Seconds_Behind_Master field after SLAVE START, but did not address a possibility that the transaction could finish before running the STOP SLAVE command, which invalidates the validations for the rest of the test case. Specifically, this would result in 1) a timeout in “Waiting for table metadata lock” on the replica, which expects the transaction to retry after slave restart and hit a lock conflict on the locked tables (added in b04c8575967), and 2) that Seconds_Behind_Master should have increased, but did not.

The failure can be reproduced by synchronizing the slave to the master before the MDEV-32265 echo statement (i.e. before the SLAVE STOP).

This patch fixes the test by adding a mechanism to use DEBUG_SYNC to synchronize a MASTER_DELAY, rather than continually increase the duration of the delay each time the test fails on buildbot. This is to ensure that on slow machines, a delay does not pass before the test gets a chance to validate results. Additionally, it decreases overall test time because the test can continue immediately after validation, thereby bypassing the remainder of a full delay for each transaction.


sql/slave.cc
```diff
@@ -3794,13 +3796,26 @@ sql_delay_event(Log_event *ev, THD *thd, rpl_group_info *rgi)
                         sql_delay, (long)ev->when,
                         rli->mi->clock_diff_with_master,
                         (long)now, (ulonglong)sql_delay_end, (long)nap_time));
-
-    if (sql_delay_end > now)
+    /* if using debug_sync for sql_delay, only delay once per event */
+    if (DBUG_EVALUATE_IF("sql_delay_by_debug_sync", type == GTID_EVENT,
+                         sql_delay_end > now))
```
knielsen, line 3801: Do you mean here "only delay once per event group" ?

In any case, thanks Brandon for taking this and fixing the random failures in the test suite.

mysql-test/suite/rpl/t/rpl_parallel_sbm.test
```diff
@@ -16,22 +17,42 @@
...
 create table t1 (a int);
+--source include/save_master_gtid.inc
+--connection slave
+set debug_sync= "now WAIT_FOR at_sql_delay";
+set debug_sync= "now SIGNAL continue_sql_thread";
+--source include/sync_with_master_gtid.inc
+
+--connection master
 create table t2 (a int);
---source include/sync_slave_sql_with_master.inc
```
andrelkin, line 34: .. as mtr was used to time out on this line?

mysql-test/suite/rpl/t/rpl_parallel_sbm.test
```diff
@@ -16,22 +17,42 @@
...
+--source include/save_master_gtid.inc
+--connection slave
+set debug_sync= "now WAIT_FOR at_sql_delay";
+set debug_sync= "now SIGNAL continue_sql_thread";
+--source include/sync_with_master_gtid.inc
```
andrelkin, line 55: Consider if it makes sense (rather for future) to encapsulate the 3 lines above into a `suite/rpl/include/` 's new macro, to invoke as
` --source include/sync_with_master_gtid_after_delay.inc`.

andrelkin (approved):
The patch looks good.
The test controlled SQL_DELAY is a promising direction for future.

mysql-test/suite/rpl/t/rpl_parallel_sbm.test
```diff
@@ -77,6 +110,11 @@ LOCK TABLES t1 WRITE;
 --source include/start_slave.inc
 
 --connection slave
+--echo # Wait for transaction to begin SQL delay...
+set debug_sync= "now WAIT_FOR at_sql_delay";
+--echo # ...and immediately release it because we don't care about delaying at this point
+set debug_sync= "now SIGNAL continue_sql_thread";
+
 --echo # Waiting for replica to get blocked by the table lock
 --let $wait_condition= SELECT count(*) FROM information_schema.processlist WHERE state LIKE 'Waiting for table metadata lock';
```
bnestere, line 119: @andrelkin MTR would time out on this line

mysql-test/suite/rpl/t/rpl_parallel_sbm.test
bnestere: Right. Perhaps to allow ignoring in batches too. E.g. here we have 2 DDLs that we can effectively ignore the SQL delay for.

mysql-test/suite/rpl/t/rpl_parallel_sbm.test
andrelkin: Then I did not get the idea of this synchronization conversion.
In the OLD version l.34 reads l.50 has been executed. There's no more to apply on slave.
The new version is different in that the GTID of the CREATE-TABLE gets in sync additionally which I don't see anyhow relevant to the rest of the test.

mysql-test/suite/rpl/t/rpl_parallel_sbm.test
bnestere: Updated, see my latest commit. I opted for an explicit number of event groups to skip, rather than the implicit option is suggested earlier.

PR #3508, opened 2024-09-11
MDEV-34650 main.having_cond_pushdown test failure - crash server (s390x)

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34650*

## Description

Code added in 10.5 as fix for MDEV-32608.

During merge in f071b7620b2e the new flag name was used.

Unfortunately missed in the merge was the refactoring performed in c76eabfb5e39a1b0831159decd3e0b544b9ad2e3, that changed the flag types to uint16.

While on little-endian processors, the cast of uint16 and int are the same, this is not true on big-endian processors (cast performed in set_extraction_flag_processor).

The result is big-endian processors cleared the immutable flag rather than set the flag, resulting in MDEV-32608 being unfixed on big-endian processors.

## Release Notes

Correct MDEV-34650  for big-endian processors.

## How can this PR be tested?

main.having_cond_pushdown test on big-endian processors

## PR quality check

### DIFF
--- a/sql/item.cc
+++ b/sql/item.cc
@@ -1295,7 +1295,7 @@ Item *Item::multiple_equality_transformer(THD *thd, uchar *arg)
       This flag will be removed at the end of the pushdown optimization by
       remove_immutable_flag_processor processor.
     */
-    int new_flag= MARKER_IMMUTABLE;
+    int16 new_flag= MARKER_IMMUTABLE;
     this->walk(&Item::set_extraction_flag_processor, false,
                (void*)&new_flag);
   }


mariadb-YuchenPei:
LGTM. Can you update the cause description in the commit message to something like?

```
The 10.5->10.6 merge commit 3bc98a4ec4e casts the arg to an int16 pointer in set_extraction_flag_processor(). The commit a5e4c34991e for MDEV-29363 added a call to set_extraction_flag_processor on IMMUTABLE_FL (MARKER_IMMUTABLE in 10.6). The subsequent 10.5->10.6 merge did not cast the flag to int16 when merging this change.
```

PR #3512, opened 2024-09-11
MDEV-34721: Make debian/autobake-debs.sh change compat level when needed

illuusio:
- [x] *The Jira issue number for this PR is: MDEV-34721*

## Description
Debian Deb Compat level version 12 brings more Systemd confort to Debian package installation handling.
Althought it's not working correctly on certain Debian and Ubuntu version which MariaDB supports.

Commit add change to compat level to version 12 when it's should be working correctly.

Currently these distros that are changed:
  * Debian Bullseye (11)
  * Debian Bookwork (12)
  * Debian Trixie (13)
  * Debian Sid
  * Ubuntu Lunar (23.04)
  * Ubuntu Mantic (23.10)
  * Ubuntu Noble (24.04)
  * Ubuntu Oracular (24.10)

## Release Notes
Bit tricky one as this changes compat level but should not be seen to user anyhow

## How can this PR be tested?
Manually on docker one can pull repo and run `AUTOBAKE_PREP_CONTROL_RULES_ONLY=1 debian/autobuild-debs.sh` and see with `git diff` if changes have been made or not. 
One can also download debs from this Salsa-CI build to test with https://salsa.debian.org/illuusio/mariadb-server/-/commit/5efd075a00b80d916cf25f92b0bc433ef9cb4d46

and see if that build is all green.

## PR quality check

### DIFF
--- a/debian/autobake-deb.sh
+++ b/debian/autobake-deb.sh
@@ -77,6 +77,14 @@ disable_libfmt()
   sed '/libfmt-dev/d' -i debian/control
 }
 
+change_compat_level()
+{
+  echo "${1}" > debian/compat
+  sed -e 's/ --fail-missing//' \
+      -e 's/#override_dh_missing/override_dh_missing/' \
+      -e 's/#\tdh_missing/\tdh_missing/' -i debian/rules
+}
+
 architecture=$(dpkg-architecture -q DEB_BUILD_ARCH)
 uname_machine=$(uname -m)
 
@@ -111,9 +119,10 @@ in
   "buster")
     disable_libfmt
     replace_uring_with_aio
-    ;&
+    ;;
   "bullseye")
     add_lsb_base_depends
+    change_compat_level 12
     ;&
   "bookworm")
     # mariadb-plugin-rocksdb in control is 4 arches covered by the distro rocksdb-tools
@@ -122,10 +131,12 @@ in
     then
       replace_uring_with_aio
     fi
+    change_compat_level 12
     ;&
   "trixie"|"sid")
     # The default packaging should always target Debian Sid, so in this case
     # there is intentionally no customizations whatsoever.
+    change_compat_level 12
     ;;
   # Ubuntu
   "focal")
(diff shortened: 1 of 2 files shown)


debian/autobake-deb.sh
```diff
@@ -111,9 +119,10 @@ in
   "buster")
     disable_libfmt
     replace_uring_with_aio
-    ;&
+    ;;
```
grooverdan, line 122: probably safer to remove "buster" (EOL) as all the previous following cases are now removed because of `;;`.

debian/autobake-deb.sh
```diff
@@ -111,9 +119,10 @@ in
...
   "bullseye")
     add_lsb_base_depends
+    change_compat_level 12
```
grooverdan, line 125: The general pattern here is the config has the most recent version, and that early versions push this back.

So:

1. remove `change_compat_level 12` from existing locations.
1. add `change_compat_level 11` to `jammy` branch

debian/autobake-deb.sh
```diff
@@ -134,12 +145,13 @@ in
     ;&
   "jammy"|"kinetic")
     add_lsb_base_depends
-    ;&
+    ;;
```
grooverdan, line 148: this is changing the behaviour of the fall throughs.

Recommend:
1. leave this as `;&`. Remove `kinetic` as its EOL. Add `change_compat 11`.

https://mariadb.com/kb/en/mariadb-platform-deprecation-policy/

debian/rules
```diff
@@ -192,6 +192,11 @@ override_dh_installinit-arch:
 override_dh_gencontrol:
 	dh_gencontrol -- -Tdebian/substvars
 
+# When running compat level 12 or higher this is needed
```
grooverdan, line 195: uncomment this.

Change the debian/compat to contain 12.

debian/autobake-deb.sh
```diff
@@ -77,6 +77,14 @@ disable_libfmt()
   sed '/libfmt-dev/d' -i debian/control
 }
 
+change_compat_level()
+{
+  echo "${1}" > debian/compat
+  sed -e 's/ --fail-missing//' \
+      -e 's/#override_dh_missing/override_dh_missing/' \
+      -e 's/#\tdh_missing/\tdh_missing/' -i debian/rules
```
grooverdan, line 85: To make this a backwards moving to compat 11, keep the `echo` and replace with a simpler sed to remove the paragraph of `compat level 12` for when `change_compat_level 11` is specified.
```
sed -e '/compat level 12/,/^$/d' -i debian/rules
```
removes the paragraph.

grooverdan (changes requested):

PR #3514, opened 2024-09-12
Update mini-benchmark to use constant transactions

robinnewhouse:
## Description

Mini-benchmark was using total time instead of total transactions
(events) to measure performance. This could give inconsistent results,
particulary when used to count CPU cycles. This changes the sysbench
command to use events x threads in each benchmark job, making the
benchmark more informative.

When `perf` [1] is available and enabled, mini-benchmark will measure how
many CPU cycles are required to do a fixed amount of work/queries. If
`perf` is unavailable, mini-benchmark will measure the peak queries per
second (QPS) during the execution of this work, regardless of the
duration.

[1] https://perf.wiki.kernel.org

## Release Notes
No release notes necessary.

## How can this PR be tested?

Testing is done in GitLab CI. A passing example can be seen at https://salsa.debian.org/robinnewhouse/mariadb-server-mirror/-/pipelines.

## PR quality check


LinuxJedi:
Hi @robinnewhouse,

In general this looks good, I'm not sure about 10.11 for it. This is because it is a behaviour change and that script is installed as part of packages.

robinnewhouse:
> Hi @robinnewhouse,
> 
> In general this looks good, I'm not sure about 10.11 for it. This is because it is a behaviour change and that script is installed as part of packages.

True, but I would call this undesirable behaviour since it gives confusing results as is. But true, it is a change in behaviour so I'm happy to rebase to main.

ottok (approved):
LGTM!

ottok:
Can you please rebase on latest 'main' so we can see if CI passed and this can be merged?

robinnewhouse:
> Can you please rebase on latest 'main' so we can see if CI passed and this can be merged?

Thanks for approving Otto. This is currently based on the latest `main`. The last commit before this on `main` was 8478a06cdbb on Sep 19, 2024.

PR #3520, opened 2024-09-13
MDEV-35151 DDL recovery for high-level indexes

svoj:
- [] *The Jira issue number for this PR is: MDEV-35151*

## Description
Various fixes to make DDL recovery work with high-level indexes.

## Release Notes
None.

## How can this PR be tested?
mtr

- [x] This is a bug fix to a new feature and the PR is based against the feature tree.

## PR quality check


sql/ddl_log.cc
```diff
@@ -1112,6 +1113,24 @@ static void execute_rename_table(DDL_LOG_ENTRY *ddl_log_entry, handler *file,
...
+    char *idx_to_end= strmov(idx_to, to_path);
+
+    for (uint i= keys; i < total_keys; i++)
+    {
+      my_snprintf(idx_from_end, HLINDEX_BUF_LEN, HLINDEX_TEMPLATE, i);
+      my_snprintf(idx_to_end, HLINDEX_BUF_LEN, HLINDEX_TEMPLATE, i);
+      file->ha_rename_table(idx_from, idx_to);
+    }
+  }
+
```
vuvova, line 1133: I'm not sure this is a good approach. It feels like you put too much inside one `DDL_RENAME_PHASE_TABLE` entry. There can be crashes between renames of individual vector indexes (presuming there will be many some day).

Perhaps, `DDL_RENAME_PHASE_TABLE` should be one `ha_rename_table` only? That is, a table with hlindexes would generate one `DDL_RENAME_PHASE_TABLE` entry **per index** (and one for the main table, of course)?

vuvova: Same for DROP, etc.

mysql-test/main/AAA.test
```diff
@@ -0,0 +1,38 @@
+source include/have_innodb.inc;
```
vuvova, line 1: erroneously committed file :)

_if this would've been the only issue, I'd removed it myself, but see below about rename_

vuvova (changes requested):

sql/ddl_log.cc
svoj: > There can be crashes between renames of individual vector indexes (presuming there will be many some day).
Yes, sure. And my code handles it inline with what original code have been doing, that is ignores `non_existing_table_errors()`, e.g. if handler files were moved and .frm not. Or as you suggested above crash between different high-level indexes rename.

> ... a table with hlindexes would generate one DDL_RENAME_PHASE_TABLE entry *per index* (and one for the main table, of course)?
1. `PHASE` is just an `enum` member of an `ACTION`, it can't have stuff like table name or whatever attached.
2. `ACTION` is more logical (or high-level) entity, which may do many things like binlogging, handling stat tables, triggers, handler files and of course .frm.

Here're options:
1. We could add special phases to handle high-level indexes, but there's going to be no much difference between my solution. And it will introduce compatibility issues. Though it is very true that older versions won't be able to handle high-level indexes properly even with this patch.
2. We can't re-use existing actions like `DDL_LOG_RENAME_TABLE_ACTION`, as I mentioned before they're way to sophisticated for high-level indexes.
3. In theory we could try re-using special actions like `DDL_LOG_RENAME_ACTION` for high level indexes. That is issue one `DDL_LOG_RENAME_TABLE_ACTION` and then `DDL_LOG_RENAME_ACTION` for each high-level index. It may work, it has no compatibility concerns, but it is a question if we're going to be able to integrate it with e.g. `ALTER`.

Your thoughts?

mysql-test/main/AAA.test
svoj: Grrr... sorry, must've mistakenly added during rebase.

storage/maria/ha_maria.cc
```diff
@@ -3317,6 +3317,11 @@ int ha_maria::create(const char *name, TABLE *table_arg,
   THD *thd= current_thd;
   DBUG_ENTER("ha_maria::create");
 
+  if (share->total_keys > share->keys)
+  {
+    my_error(ER_ILLEGAL_HA_CREATE_OPTION, MYF(0), "Aria", "VECTOR");
+    DBUG_RETURN(HA_ERR_UNSUPPORTED);
+  }
```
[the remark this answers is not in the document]
svoj: Thanks!

vuvova:
could you please rebase it on top of the latest bb-11.6-MDEV-32887-vector ? It's impossible to review now, it looks like it includes basically everything

sql/ddl_log.cc
vuvova: I don't think I'm much of an expert in ddl logging. Let's use your approach if it works.

vuvova:
cherry-picked into the vector branch, thanks! closing.

PR #3523, opened 2024-09-17
MDEV-31636 Memory leak in Sys_var_gtid_binlog_state::do_check()

DaveGosselin-MariaDB:
Move memory allocations performed during Sys_var_gtid_binlog_state::do_check to Sys_var_gtid_binlog_state::global_update where they will be freed before the latter method returns.

### DIFF
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -2016,15 +2016,6 @@ struct gtid_binlog_state_data { rpl_gtid *list; uint32 list_len; };
 bool
 Sys_var_gtid_binlog_state::do_check(THD *thd, set_var *var)
 {
-  String str, *res;
-  struct gtid_binlog_state_data *data;
-  rpl_gtid *list;
-  uint32 list_len;
-
-  DBUG_ASSERT(var->type == OPT_GLOBAL);
-
-  if (!(res= var->value->val_str(&str)))
-    return true;
   if (thd->in_active_multi_stmt_transaction())
   {
     my_error(ER_CANT_DO_THIS_DURING_AN_TRANSACTION, MYF(0));
@@ -2040,6 +2031,31 @@ Sys_var_gtid_binlog_state::do_check(THD *thd, set_var *var)
     my_error(ER_BINLOG_MUST_BE_EMPTY, MYF(0));
     return true;
   }
+  return false;
+}
+
+
+bool
+Sys_var_gtid_binlog_state::global_update(THD *thd, set_var *var)
+{
+  DBUG_ASSERT(var->type == OPT_GLOBAL);
+
+  if (!var->value)
+  {
+    my_error(ER_NO_DEFAULT, MYF(0), var->var->name.str);
+    return true;
+  }
+
+  bool result;
+  String str, *res;
+  struct gtid_binlog_state_data *data;
+  rpl_gtid *list;
+  uint32 list_len;
+
+  DBUG_ASSERT(var->type == OPT_GLOBAL);
+
+  if (!(res= var->value->val_str(&str)))
+    return true;
   if (res->length() == 0)
   {
     list= NULL;
@@ -2061,31 +2077,13 @@ Sys_var_gtid_binlog_state::do_check(THD *thd, set_var *var)
   data->list= list;
   data->list_len= list_len;
   var->save_result.ptr= data;
-  return false;
-}
-
-
-bool
-Sys_var_gtid_binlog_state::global_update(THD *thd, set_var *var)
-{
-  bool res;
-
-  DBUG_ASSERT(var->type == OPT_GLOBAL);
-
-  if (!var->value)
-  {
-    my_error(ER_NO_DEFAULT, MYF(0), var->var->name.str);
-    return true;
-  }
-
-  struct gtid_binlog_state_data *data=
-    (struct gtid_binlog_state_data *)var->save_result.ptr;
+  
   mysql_mutex_unlock(&LOCK_global_system_variables);
-  res= (reset_master(thd, data->list, data->list_len, 0) != 0);
+  result= (reset_master(thd, data->list, data->list_len, 0) != 0);
   mysql_mutex_lock(&LOCK_global_system_variables);
   my_free(data->list);
   my_free(data);
-  return res;
+  return result;
 }
 
 
--- a/mysql-test/main/mdev-31636.result
+++ b/mysql-test/main/mdev-31636.result
@@ -1,0 +1,8 @@
+RESET MASTER;
+SET
+@@global.gtid_binlog_state='1-1-101,2-1-2002',
+@@global.slave_parallel_mode=x;
+ERROR 42000: Variable 'slave_parallel_mode' can't be set to the value of 'x'
+SELECT @@global.gtid_binlog_state;
+@@global.gtid_binlog_state
+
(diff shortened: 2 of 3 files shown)


sql/sys_vars.cc
```diff
@@ -2013,6 +2013,20 @@ Sys_gtid_strict_mode(
 
 struct gtid_binlog_state_data { rpl_gtid *list; uint32 list_len; };
 
+void
+Sys_var_gtid_binlog_state::on_error(set_var *var)
+{
+  if (!var->save_result_exists)
+    return;
+  auto data= (gtid_binlog_state_data *)var->save_result.ptr;
```
andrelkin, line 2021: `ptr` is allocated in [`do_check()`](https://github.com/MariaDB/server/blob/10.5-mdev-31636/sql/sys_vars.cc#L2068) somewhat quite surprisingly 'cos it all looks to me this could be done inside
[`var->update()](https://github.com/MariaDB/server/blob/review-10.5-mdev-31636/sql/set_var.cc#L750).

I'd consider that first. If turns true indeed, we naturally shall carefully check all existing class' `do_check()`  (to my browsing there's none but ours replication's).

andrelkin (commented):
Thanks Dave for a good piece. It all seems to provide a solution. However let's look at the issue at a different angle I am suggesting in a note.

DaveGosselin-MariaDB:
> Thanks Dave for a good piece. It all seems to provide a solution. However let's look at the issue at a different angle I am suggesting in a note.

Thank you Andrei 😄 I will investigate the other angle.

sql/sys_vars.cc
bnestere: Hmm, though it logically makes sense for the actual validation of the input value to be in `do_check()`. `Sys_var_gtid_slave_pos` has a slightly different, yet still consistent to `check` semantics pattern, which in `Sys_var_gtid_slave_pos::do_check()` validates the value and saves the string on `thd`, and then uses that string in `Sys_var_gtid_slave_pos::global_update()` in a load. Can we follow that pattern here? (Though @DaveGosselin-MariaDB I recall in our discussion you mentioned trying to save something on the thread, and it not working. Is this what you tried?).

Generally speaking though, I' say whichever pattern we pick, `Sys_var_gtid_binlog_state` and `Sys_var_gtid_slave_pos` should probably follow the same pattern in the source code for consistency.

DaveGosselin-MariaDB: Hey Brandon, I wasn't saving something on the thread but rather marking variables such that an `on_error` phase could be executed over all affected variables so that they may handle any cleanup in their own implementation-dependent way (see https://github.com/MariaDB/server/commit/90eab642c6583744ea6d3e09554fbac04a27f84a )

andrelkin (commented):
Looks good to me.

What do you think if we cover a lesser but still concerning [`Sys_var_gtid_slave_pos::do_check()](https://github.com/MariaDB/server/blob/10.5-mdev-31636/sql/sys_vars.cc#L1905)
similarly as well?
(This time is just a memroot allocation which would be consistent with this patch too to do in the update time)

montywi:
Hi!

On Wed, 18 Sept 2024, 20:27 'andrelkin' via Michael Widenius, <
***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In sql/sys_vars.cc
> <https://github.com/MariaDB/server/pull/3523#discussion_r1765447552>:
>
> > @@ -2013,6 +2013,20 @@ Sys_gtid_strict_mode(
>
>  struct gtid_binlog_state_data { rpl_gtid *list; uint32 list_len; };
>
> +void
> +Sys_var_gtid_binlog_state::on_error(set_var *var)
> +{
> +  if (!var->save_result_exists)
> +    return;
> +  auto data= (gtid_binlog_state_data *)var->save_result.ptr;
>
>
Please so not use auto. Better to write explicit types to make it more
human readable.


Regards,
Monty

> Please
>
> .
>

sql/sys_vars.cc
andrelkin: @bnestere, a good reminder point `Sys_var_gtid_slave_pos::do_check`!
Indeed both variables need be consistent wrt `do_check()` actions.
They both allocate and are different in the style of allocation.  [`Sys_var_gtid_slave_pos` 's](https://github.com/MariaDB/server/blob/10.5-mdev-31636/sql/sys_vars.cc#L1930)

 ```
 if (!(var->save_result.string_value.str=
        thd->strmake(res->ptr(), res->length())))
  {
    my_error(ER_OUT_OF_RESOURCES, MYF(0));
    return true;
  }

```
is comparable with a block the current patch relocated.

I suggest we do the same. @DaveGosselin-MariaDB, your thoughts?

DaveGosselin-MariaDB: Hey @bnestere and @andrelkin!  `Sys_var_gtid_slave_pos::do_check` leaks memory in the sense that the result of `strmake` will live on the THD memroot indefinitely in the case of a failure like the one in the attached test.  Consequently, it seems that making `Sys_var_gtid_binlog_state` behave similarly won't fix the leak but will hide it instead.  For the `handle_*` replication threads, the memroot's lifetime extends beyond the current query, correct?  If so, we could potentially have leaks build up for the lifetime of the process, not just for the query's lifetime.
I think we could either (1) deliver my changes from https://github.com/MariaDB/server/commit/90eab642c6583744ea6d3e09554fbac04a27f84a or (2) keep the current changes and modify `Sys_var_gtid_slave_pos::do_check` to avoid leaking memory in the case of a failure by applying to it the same idea that we applied to `Sys_var_gtid_binlog_state::do_check` in the current commit.

bnestere: Ah, great point, we probably shouldn't hide the leak either. I think the `on_error()` approach offers a good amount of flexibility. Though before we make a decision, I'll throw a third option into the mix: 3) add a `MEM_ROOT` to `set_var` so we can allocate within the context of one `SET` (and automatically free it in `set_var::~set_var()`)

DaveGosselin-MariaDB: Good idea @bnestere , I will try that out and report back.

andrelkin: I agree (1) is more flexible. (2) is just simpler as so far it has been sufficient. I would pick (2) to hold on with (1) until any first incident that leaves no other, incl (1), choices is met.
To (3), I am reluctant. Indeed the `THD::main_mem_root` that is used here gets reset at the end of the statement (incl a SET one with a failing sub-SET).

I would propose (4) to keep (2) fool-proof with consting `do_check( ..., const set_var *var)` the 2nd arg. But I am not really proposing, as the signature change looks super-pervasive.
Also this measure of course would not block altering other objects (in replication uses case `mysql_bin_log` or `rpl_slave_state` could become victims of that). That's why I think such altering actions should be a part of `global_update()` :s rather than a part of (1) 's `on_error`.

DaveGosselin-MariaDB: If we keep (2) but without adding the `const` then the change as it is should be fine to merge because of my reasoning from  https://github.com/MariaDB/server/pull/3523#discussion_r1775450108 .  If that's the case, then let's deliver this change as it is.  We could create a ticket to address broader architectural changes as a separate effort.

DaveGosselin-MariaDB: (I see I have an OK from Andrei in a separate comment on this PR, can I take that to mean that it's OK to push?)

andrelkin: Amen!

andrelkin (approved):

PR #3525, opened 2024-09-18
MDEV-34940: Fix Item_func_collect inheritance

DaveGosselin-MariaDB:
Creates new parent Item_sum_str for Item_func_group_concat and Item_func_collect, migrating shared methods from each of those latter classes to the former.  Simultaneously, tighten up the method scopes.

### DIFF
--- a/sql/item_sum.cc
+++ b/sql/item_sum.cc
@@ -3553,6 +3553,39 @@ String *Item_sum_udf_str::val_str(String *str)
 #endif /* HAVE_DLOPEN */
 
 
+bool
+Item_sum_str::fix_fields(THD *thd, Item **ref)
+{
+  DBUG_ASSERT(fixed() == 0);
+
+  if (init_sum_func_check(thd))
+    return TRUE;
+
+  set_maybe_null();
+
+  /*
+    Fix fields for select list and ORDER clause
+  */
+
+  for (uint i=0 ; i < arg_count ; i++)
+  {
+    if (args[i]->fix_fields_if_needed_for_scalar(thd, &args[i]))
+      return TRUE;
+    /* We should ignore FIELD's in arguments to sum functions */
+    with_flags|= (args[i]->with_flags & ~item_with_t::FIELD);
+  }
+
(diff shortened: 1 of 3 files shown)


sql/item_sum.cc
```diff
@@ -3553,6 +3553,57 @@ String *Item_sum_udf_str::val_str(String *str)
...
+Item_sum_str::Item_sum_str(THD *thd, Item_sum_str *item)
+  : Item_sum(thd, item)
+{
+}
+
+Item_sum_str::Item_sum_str(THD *thd, Item *item_par)
+  : Item_sum(thd, item_par)
+{
+}
+
```
abarkov, line 3570: The implementations of the constructors above could probably go directly to item_sum.h

sql/item_sum.cc
```diff
@@ -3553,6 +3553,57 @@ String *Item_sum_udf_str::val_str(String *str)
...
+      return TRUE;
+    /* We should ignore FIELD's in arguments to sum functions */
+    with_flags|= (args[i]->with_flags & ~item_with_t::FIELD);
+  }
+
+  bool ffi_result= fix_fields_impl(thd, ref);
+
+  if (ffi_result)
+      return ffi_result;
+
```
abarkov, line 3598: The variable ffi_result does not seem to be needed.

```
if (fix_fields_impl(thd, ref))
  return true;
```
Wouldn't this work instead?

sql/item_sum.cc
```diff
@@ -4637,7 +4664,7 @@ void Item_func_collect::clear() {
 void Item_func_collect::cleanup() {
   List_iterator<String> geometries_iterator(geometries);
   geometries.delete_elements();
-  Item_sum_int::cleanup();
+  Item_sum::cleanup();
 }
```
abarkov, line 4669: I think it should be `Item_sum_str::cleanup()` here.

abarkov (changes requested):
The patch looks ok for me. I only found some minor issues (see comments).

PR #3528, opened 2024-09-19
MDEV-25611: RESET MASTER causes the server to hang

knielsen:
MDEV-25611: RESET MASTER causes the server to hang

RESET MASTER waits for storage engines to reply to a binlog checkpoint
requests. If this response is delayed for a long time for some reason, then
RESET MASTER can hang.

Fix this by forcing a log sync in all engines just before waiting for the
checkpoint reply.

(Waiting for old checkpoint responses is needed to preserve durability of
any commits that were synced to disk in the to-be-deleted binlog but not yet
synced in the engine.)


andrelkin (approved):
Looks good. Thanks!

PR #3538, opened 2024-09-21
Add missing `LEX_STRING::str`s for `my_snprintf`

ParadoxV5:
  * Udate: *The Jira issue number for this PR is: [MDEV-35434](https://jira.mariadb.org/browse/MDEV-35434)*

## Description
This commit backports a fix group from #3360 for 11.5.
These are the only mistakes introduced during 11.5.x.

When these members switched from plain `char*` to `LEX_CSTRING`, not all usages were converted. Specifically, in this commit are args of `my_snprintf` derivatives.
Because until MDEV-21978 (#3360), automated type checks were unavailable for those functions due to their incompatibility, so these tools didn’t catch them.

### If some output changed that is not visible in a test case, what was it looking like before the change and how it's looking with this patch applied?
`str` is the initial member of `LEX_CSTRING` (`typedef` of `struct st_mysql_const_lex_string`), so this difference did not cause incompatibility on most platforms with seamless alignments;
howëver, [my co-mentor notes that Visual Studio reads garbage](https://github.com/MariaDB/server/pull/3360#discussion_r1712667580).

## Release Notes
* Fixed data size mismatches that were garbling error and debug outputs (or possibly even crashes) on problematic platforms
  * (I adapted this line from the mould I used for previous iterations of https://github.com/MariaDB/server/pull/3360#issue-2374107080 backports.)

## How can this PR be tested?
This difference only manifests when `LEX_CSTRING::str` does not align on the `0`th index of the overall `LEX_CSTRING`.
Besides using Visual Studio as noted, another deception to disregard this leniency is to shift the `str` member away from the `0`th alignment, such as adding a padding member before it or moving it after member `length`.

## PR quality check

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -5732,7 +5732,7 @@ static void test_lc_time_sz()
         (*loc)->max_day_name_length != max_day_len)
     {
       DBUG_PRINT("Wrong max day name(or month name) length for locale:",
-                 ("%s", (*loc)->name));
+                 ("%s", (*loc)->name.str));
       DBUG_ASSERT(0);
     }
   }
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -6063,7 +6063,7 @@ static bool check_locale(sys_var *self, THD *thd, set_var *var)
     {
       push_warning_printf(thd, Sql_condition::WARN_LEVEL_WARN, ER_UNKNOWN_ERROR,
                           "Can't process error message file for locale '%s'",
-                          locale->name);
+                          locale->name.str);
       return true;
     }
   }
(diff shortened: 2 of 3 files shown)


ParadoxV5:
Oh great, the sporadic failures came back to haunt us.

grooverdan:
Looks all right. Thanks. 11.5 is now EOL - can you rebase to 11.6 please.

ParadoxV5:
👀 Oh right, 11.5 is a rolling release and has been superceded by 11.6.
The [Maintenance Policy](https://mariadb.org/about/#maintenance-policy) links to the [rolling release system announcement](https://mariadb.org/adjusting-release-model/) but doesn’t (explicitly) say much itself.

grooverdan (approved):

PR #3541, opened 2024-09-22
Fix a couple of `my_snprintf` arg mismatches

ParadoxV5:
* Sibling of #3485, #3493 and #3518

## Description
This commit ports two fixes from #3360 according to the bug-fixing process to 11.4, covering cases introduced during 11.4.x.
* `client/mysqldump.cc`: Error message template had `%u` for a `long` arg (`MYSQL::thread_id`)
* `sql/log.cc`: Log message template had a `%s` but received no arg.

## Release Notes
* Fixed data size mismatches that were garbling error and debug outputs (or possibly even crashes) on problematic platforms
  * This should  match #3538.

## How can this PR be tested?
* `client/mysqldump.cc`: I’m not sure which platforms has `sizeof(long)` different from `sizeof(int)`.
* `sql/log.cc`: The dangling `%s` could pick up a garbage byte sequence in *ideal* scenarios.

* *This is a new feature or a refactoring, and the PR is based against the `main` branch.*~~

## PR quality check

### DIFF
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -7285,7 +7285,7 @@ int main(int argc, char **argv)
       goto err;
     connection_pool.for_each_connection([](MYSQL *c) {
       if (start_transaction(c))
-        maybe_die(EX_MYSQLERR, "Failed to start transaction on connection ID %u", mysql->thread_id);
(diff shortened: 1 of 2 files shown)


LinuxJedi (approved):
Many thanks for this.

PR #3542, opened 2024-09-23
MDEV-34983: Remove x86 asm work-arounds from InnoDB

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34983*
## Description
Starting with GCC 7 and clang 15, single-bit operations such as `fetch_or(1) & 1` are translated into 80386 instructions such as LOCK BTS, instead of using the generic translation pattern of emitting a loop around `LOCK CMPXCHG`.

Given that the oldest currently supported GNU/Linux distributions ship GCC 7, and that older versions of GCC are out of support, let us remove some work-arounds that are not actually necessary.

`srw_mutex_impl::HOLDER`: Changed from `1U<<31` to `1` in order to work around llvm/llvm-project#37322 which is related to the most significant bit.

`srw_mutex_impl::WAITER`: A multiplier of waiting requests. This used to be `1`, which would now collide with `HOLDER`.

`fil_space_t::set_stopping()`: Remove this unused function.

In MSVC we need `_interlockedbittestandset()` for `LOCK BTS`.
## Release Notes
No mention needed. This should not affect performance; it is only cleaning up some code.
## How can this PR be tested?
The regression test suite covers this low-level InnoDB code. Additionally, the function `srw_mutex_impl<false>::wait_and_lock` in `sql/mariadbd` or in the unit test program `storage/innobase/unittest/innodb_sync-t` may be disassembled to confirm that no `lock cmpxchg` is being used.

We can also see how `fil_space_t::clear_flush()` is inlined at the end of `fil_space_t::flush_low()`:
```asm
lock andl $0xefffffff,0x9c(%r14)
lock subq $0x1,0x1215a39(%rip)        # 0x202af88 <fil_n_pending_tablespace_flushes>
```
Or how `fil_space_t::clear_closing()` is translated in `fil_space_t::prepare_acquired()`:
```asm
lock andl $0xdfffffff,0x9c(%rbx)
```
According to https://stackoverflow.com/questions/64904176/is-there-any-way-to-compile-x86-64-arm64-universal-binaries-in-macos-10-14 this could allow more efficient code to be compiled for Apple macOS 10.14 (apparently the last version that supports x86-64). Someone who has access to such an environment might be interested in testing that. The latest XCode release that supports x86 ought to emit loops around `lock cmpxchg`.
## PR quality check

### DIFF
--- a/storage/innobase/sync/srw_lock.cc
+++ b/storage/innobase/sync/srw_lock.cc
@@ -269,44 +269,10 @@ template void ssux_lock_impl<false>::wake();
 template void srw_mutex_impl<true>::wake();
 template void ssux_lock_impl<true>::wake();
 
-/*
-
-Unfortunately, compilers targeting IA-32 or AMD64 currently cannot
-translate the following single-bit operations into Intel 80386 instructions:
-
-     m.fetch_or(1<<b) & 1<<b       LOCK BTS b, m
-     m.fetch_and(~(1<<b)) & 1<<b   LOCK BTR b, m
-     m.fetch_xor(1<<b) & 1<<b      LOCK BTC b, m
-
-Hence, we will manually translate fetch_or() using GCC-style inline
-assembler code or a Microsoft intrinsic function.
-
-*/
-
-#if defined __clang_major__ && __clang_major__ < 10
-/* Only clang-10 introduced support for asm goto */
-#elif defined __APPLE__
-/* At least some versions of Apple Xcode do not support asm goto */
-#elif defined __GNUC__ && (defined __i386__ || defined __x86_64__)
-# define IF_FETCH_OR_GOTO(mem, bit, label)				\
-  __asm__ goto("lock btsl $" #bit ", %0\n\t"				\
-               "jc %l1" : : "m" (mem) : "cc", "memory" : label);
-# define IF_NOT_FETCH_OR_GOTO(mem, bit, label)				\
-  __asm__ goto("lock btsl $" #bit ", %0\n\t"				\
-               "jnc %l1" : : "m" (mem) : "cc", "memory" : label);
-#elif defined _MSC_VER && (defined _M_IX86 || defined _M_X64)
-# define IF_FETCH_OR_GOTO(mem, bit, label)				\
-  if (_interlockedbittestandset(reinterpret_cast<volatile long*>(&mem), bit)) \
-    goto label;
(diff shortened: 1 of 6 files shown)


storage/innobase/sync/srw_lock.cc
```diff
@@ -269,44 +269,10 @@ template void ssux_lock_impl<false>::wake();
...
-# define IF_NOT_FETCH_OR_GOTO(mem, bit, label)				\
-  if (!_interlockedbittestandset(reinterpret_cast<volatile long*>(&mem), bit))\
-    goto label;
-#endif
-
 template<bool spinloop>
 void srw_mutex_impl<spinloop>::wait_and_lock()
 {
-  uint32_t lk= 1 + lock.fetch_add(1, std::memory_order_relaxed);
+  uint32_t lk= WAITER + lock.fetch_add(WAITER, std::memory_order_relaxed);
```
mariadb-DebarunBanerjee, line 275: Now we have HOLDER as the first BIT, and WAITER is 2. Hopefully increment by 2 is not going to cost more than atomic increment but perhaps something to watch out.

mariadb-DebarunBanerjee (approved):
Nice work! Looks good to me.

dr-m:
For the record, according to [godbolt.org](https://godbolt.org/z/W4KrPvejT) it does not seem to make any difference if we atomically increment by 1 or 2. All of ARM, POWER, s390x would by default use a loop around lock-loaded/store-conditional. In the above link, I specified `-march=armv8.1-a` so that `ldadd` would be used instead. In case the link expires, my test program was as follows:
```c++
#include <atomic>
template<uint32_t N>
uint32_t a(std::atomic<uint32_t> &a) { return a.fetch_add(N, std::memory_order_relaxed); }
template uint32_t a<1>(std::atomic<uint32_t>&);
template uint32_t a<2>(std::atomic<uint32_t>&);
```
In any case, I think it would be good if @vaintroub tested this on Microsoft Windows. While I checked some things on https://godbolt.org, I can’t be completely sure if more complex code gets translated optimally. Specifically, we do not want to see any increase in `lock cmpxchg` operations.

dr-m:
[Here is a more complex example](https://godbolt.org/z/nqv8a85rM) that is based on `void srw_mutex_impl<false>::wait_and_lock()`. It is being translated on both MSVC and GCC 7.1 as expected, using no `lock cmpxchg` but just `lock xadd` and `lock bts`.

PR #3543, opened 2024-09-23
reformat galera sst error messages

vuvova:
put the command line at the end. so that when a very long command line
is truncated, it doesn't take the actual error message with it

### DIFF
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -597,8 +597,8 @@ static void* sst_joiner_thread (void* a)
       if (!tmp || strlen(tmp) < (magic_len + 2) ||
           strncasecmp (tmp, magic, magic_len))
       {
-        WSREP_ERROR("Failed to read '%s <addr>' from: %s\n\tRead: '%s'",
-                    magic, arg->cmd, tmp);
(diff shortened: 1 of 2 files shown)


vuvova:
@janlindstrom , I've changed a couple of error messages and I'm not sure if they're important — e.g. some tool might grep for them or look for exact wording, so here's a review request

janlindstrom (approved):
Looks good to me.

sql/wsrep_sst.cc
```diff
@@ -610,8 +610,8 @@ static void* sst_joiner_thread (void* a)
     else
     {
       err= proc.error();
-      WSREP_ERROR("Failed to execute: %s : %d (%s)",
-                  arg->cmd, err, strerror(err));
+      WSREP_ERROR("Failed to execute (%M): %s",
```
ParadoxV5, line 613: As a part of #3360, I noted that `WSREP_ERROR` does *not* use `my_vsnprintf` and thus will not recognize `%M`:
https://github.com/MariaDB/server/blob/f644b9945b4286d44c64948160a4a7b6a10a13a5/include/wsrep.h#L27-L33
https://github.com/MariaDB/server/blob/f644b9945b4286d44c64948160a4a7b6a10a13a5/sql/wsrep_mysqld.cc#L289-L326

PR #3544, opened 2024-09-23
MDEV-34122: Assertion entry failed in Active_tranx::assert_thd_is_waiter

bnestere:
In between the binlogging of a transaction and its wait, it is
possible that its entry in Active_tranx was removed if semi-sync was 
switched off and on. This fires an assertion that checks before
awaiting an ACK that the entry must exist in Active_tranx.

The fix is to ensure that the entry exists before awaiting the ACK,
and if there is no entry, write an informative message to the user
explaining that the transaction is skipping its wait. Additionally,
debug-build only logic is added to ensure that the cause of the 
missing entry is due to semi-sync being turned off and on.

Initial PR organized as three commits:
 1. Regression MTR test to show the problem
 2. Fix
 3. Extension to the test/result to ensure informational note is provided

A general note, I'm not thrilled about moving `is_thd_waiter()` to be in non-debug builds, but I also can't think of another way without adding overhead to every thread. I thought about making it part of `clear_active_tranx_nodes()` call to `signal_waiting_transaction()`, and reusing existing variables (e.g. by resetting `thd->semisync_info`) but it doesn't work well group commit.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_semi_sync_cond_var_per_thd.test
+++ b/mysql-test/suite/rpl/t/rpl_semi_sync_cond_var_per_thd.test
@@ -19,9 +19,12 @@
 # References:
 #   MDEV-33551: Semi-sync Wait Point AFTER_COMMIT Slow on Workloads with Heavy
 #               Concurrency
+#   MDEV-34122: Assertion `entry' failed in Active_tranx::assert_thd_is_waiter
 #
 --source include/have_binlog_format_row.inc
+--source include/have_innodb.inc
 --source include/have_debug.inc
+--source include/have_debug_sync.inc
 --source include/master-slave.inc
 
 --connection master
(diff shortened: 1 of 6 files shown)


sql/semisync_master.cc
```diff
@@ -922,6 +915,34 @@ int Repl_semi_sync_master::commit_trx(const char *trx_wait_binlog_name,
...
+       */
+      if (unlikely(!m_active_tranxs->is_thd_waiter(thd, trx_wait_binlog_name,
+                                                   trx_wait_binlog_pos)))
+      {
+        sql_print_information(
+            "Skipping semi-sync wait for transaction at pos %s, %lu. This is "
+            "likely because semi-sync turned off and on during the lifetime "
+            "of this transaction.",
+            trx_wait_binlog_name, trx_wait_binlog_pos);
+
```
knielsen, line 936: I don't think this sql_print_information() is needed (just make it a DBUG_PRINT() instead).
If semi-sync is disabled during the transaction, it is expected that it will not wait for ack. And a message like this triggered by a specific timing between re-enabling and commit processing is more confusing than helpful.

mysql-test/suite/rpl/t/rpl_semi_sync_cond_var_per_thd.test
```diff
@@ -77,4 +80,116 @@ set @@global.binlog_commit_wait_usec=@save_bgc_usec;
...
+
+--connection server_1
+--reap
+
+--let $assert_text= Ensure error log shows that transaction is skipping its semi-sync wait
+--let $assert_select=Skipping semi-sync wait for transaction
+--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_count= 1
+--let $assert_only_after=CURRENT_TEST
+--source include/assert_grep.inc
```
knielsen, line 152: This part of the test is not necessary since the information message should be removed.
If you want, and to not waste the great effort on the test case, feel free to use DBUG injection or similar to get a message through in the test case to verify this anyway, if you find it meaningful, up to you.

knielsen (approved):
Thanks, Brandon, I think the patch looks generally ok

As described in detail comment, the sql_print_information() is not needed I think and should be omitted.

Thanks for writing good commit messages and comments in general, that is really appreciated and I think a mark of high quality development work. One suggestion for improvement, related to this great comment from the test case:
```
+--echo # MDEV-34122
+--echo # If semi-sync is switched off then on while a transaction is
+--echo # in-between binlogging and waiting for an ACK, ensure that the
+--echo # transaction skips the wait altogether (otherwise it would time-out).
+--echo # Note that prior to MDEV-34122, there was a debug assertion that would
+--echo # trigger if the transaction tried to wait but could not receive an ACK
+--echo # signal.
```
This is a great explanation of what the actual problem is, and better than what is in the commit comment and in the MDEV-34122 issue (this is natural, often happens as the real problem is discovered during debugging/development).

So I suggest to update the commit comment (and MDEV-34122 description perhaps) to be along the lines of the explanation in the test case, describing the possible hang.

Good work also with adding the DBUG check for that the cause of missing entry is really the turning off+on.

PR #3545, opened 2024-09-24
MDEV-34340 netbsd fix crash on MariaDB-backup

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34340*

## Description

Without a get_exe implementation there was a null pointer exception my_realpath.

Refactor to mysys library as requested by @vuvova.

## Release Notes

Fix NetBSD compatibility in MariaDB backup thanks to Thomas Klausner

## How can this PR be tested?

Confirmed by contributor. (or we setup a NetBSD test)

## PR quality check


vuvova:
I thought it should also replace similar code in `mysys/my_init.c` — and _that's why_ this function could be in mysys, because it's universal enough to be used in at least two places.

mysys/my_getexe.c
```diff
@@ -0,0 +1,66 @@
+/* Copyright (c) 2017, MariaDB Plc
```
LinuxJedi, line 1: Is it though?

LinuxJedi (commented):
One comment in the review.

Also, if Sergei is referring to the symlink check in `my_init`. With recent discussions that might end up going away anyway.

mysys/my_getexe.c
grooverdan: Vlad was the plc author of the entire my_get_exepath function that was copied here and its authored year of 2017 was used.

grooverdan:
> I thought it should also replace similar code in `mysys/my_init.c` — and _that's why_ this function could be in mysys, because it's universal enough to be used in at least two places.

Regardless, this is a 10.5 fix. To adjust my_init.c for checking replacing a symlink check would be in the 11.x branch (if discussions of its obsoletes got ahead). Regardless it becomes a common library to use.

PR #3546, opened 2024-09-24
MDEV-34994: sql/mysqld: stop accept() loop after the first EAGAIN

MaxKellermann:
- [x] *The Jira issue number for this PR is: MDEV-34994*

## Description
Each time a listener socket becomes ready, MariaDB calls accept() ten
times (MAX_ACCEPT_RETRY), even if all but the first one return EAGAIN
because there are no more connections.  This causes unnecessary CPU
usage - on our server, the CPU load of that thread, which does nothing
but accept(), saturates one CPU core by ~45%.  The loop should stop
after the first EAGAIN.

Perf report:

```
    11.01%  mariadbd  libc.so.6          [.] accept4
     6.42%  mariadbd  [kernel.kallsyms]  [k] finish_task_switch.isra.0
     5.50%  mariadbd  [kernel.kallsyms]  [k] _raw_spin_unlock_irqrestore
     5.50%  mariadbd  [kernel.kallsyms]  [k] syscall_enter_from_user_mode
     4.59%  mariadbd  [kernel.kallsyms]  [k] __fget_light
     3.67%  mariadbd  [kernel.kallsyms]  [k] kmem_cache_alloc
     2.75%  mariadbd  [kernel.kallsyms]  [k] fput
     2.75%  mariadbd  [kernel.kallsyms]  [k] mod_objcg_state
     1.83%  mariadbd  [kernel.kallsyms]  [k] __inode_wait_for_writeback
     1.83%  mariadbd  [kernel.kallsyms]  [k] __sys_accept4
     1.83%  mariadbd  [kernel.kallsyms]  [k] _raw_spin_unlock_irq
     1.83%  mariadbd  [kernel.kallsyms]  [k] alloc_inode
     1.83%  mariadbd  [kernel.kallsyms]  [k] call_rcu
```

## Release Notes
Reduce number of accept() system calls.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -6360,7 +6360,9 @@ void handle_connections_sockets()
                                     &length);
       if (mysql_socket_getfd(new_sock) != INVALID_SOCKET)
         handle_accepted_socket(new_sock, sock);
-      else if (socket_errno != SOCKET_EINTR && socket_errno != SOCKET_EAGAIN)
+      else if (socket_errno == SOCKET_EAGAIN || socket_errno == SOCKET_EWOULDBLOCK)
+        break;
+      else if (socket_errno != SOCKET_EINTR)
       {
         /*
           accept(2) failed on the listening port.


sql/mysqld.cc
```diff
@@ -6360,7 +6360,9 @@ void handle_connections_sockets()
                                     &length);
       if (mysql_socket_getfd(new_sock) != INVALID_SOCKET)
         handle_accepted_socket(new_sock, sock);
-      else if (socket_errno != SOCKET_EINTR && socket_errno != SOCKET_EAGAIN)
+      else if (socket_errno == SOCKET_EAGAIN)
```
vaintroub, line 6363: could you change that test to `(socket_errno == SOCKET_EAGAIN || socket_errno == SOCKET_EWOULDBLOCK)` ?
Linux and FreeBSD documentation mention both of the codes for the case where non-blocking server socket does not have connections that can be accepted, while macOS only mentions EWOULDBLOCK

vaintroub (changes requested):
Good catch, thanks. 
Makes sense to test for EWOULDBLOCK as well (I commented on the code line)

sql/mysqld.cc
MaxKellermann: Actually, it's just `EWOULDBLOCK` on Linux, FreeBSD and all the others - but on these, `EAGAIN==EWOULDBLOCK`.
Only on Windows, it's documented as `WSAEINPROGRESS` (for blocking sockets) or `WSAEWOULDBLOCK` (for non-blocking sockets). Windows is weird.
See https://github.com/CM4all/libcommon/blob/master/src/net/SocketError.hxx#L90 (and the surrounding code) for all the weird corner cases....
In any case, adding `EWOULDBLOCK` doesn't hurt - it costs nothing, it's redundant and the compiler will optimize it away.

MaxKellermann:
Force-push: rebase and added `SOCKET_EWOULDBLOCK` as requested.

sql/mysqld.cc
vaintroub: Thanks!

Do not worry about Windows, it does not run this code. The Windows specific code had been factored out into handle_connections_win.cc, which has named pipes, Windows events, and asynchronous APIs, and special stuff like AcceptEx,  more than one "poll" loop like that would offer.

Thanks for mentioning EAGAIN and EWOULDBLOCK are be the same, I never actually checked, alas it is not guaranteed to be everywhere,  there is at least a "historic" Unix (AIX), where we try to build, where those constants are different, according to https://stackoverflow.com/a/7003379/547065.

vaintroub (approved):
Looks good to me now.

PR #3548, opened 2024-09-24
MDEV-32631 : galera_2_cluster: before_rollback(): Assertion `0' failed

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-32631*

## Description
Test case changes only. Add wait_conditions to make sure nodes rejoin the cluster. Assertion itself should not be possible anymore as we do not allow sequences on Aria tables.

## Release Notes

## How can this PR be tested?

- [x ] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/disabled.def
+++ b/mysql-test/suite/galera/disabled.def
@@ -12,7 +12,6 @@
 
 galera_as_slave_ctas : MDEV-28378 timeout
 galera_pc_recovery : MDEV-25199 cluster fails to start up
-galera_sequences : MDEV-32561 WSREP FSM failure: no such a transition REPLICATING -> COMMITTED
 galera_concurrent_ctas : MDEV-32779 galera_concurrent_ctas: assertion in the galera::ReplicatorSMM::finish_cert()
 galera_as_slave_replay : MDEV-32780 galera_as_slave_replay: assertion in the wsrep::transaction::before_rollback()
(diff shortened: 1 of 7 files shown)


sysprg:
@janlindstrom Thanks for fixes for complex tests, but the fix does not include changes to the wsrep inserts or library code, it only includes changes to the tests - yes, this will solve the problems with the tests by bypassing the problematic scenario, but it will not fix anything for the user - if the user performs actions as in the original test, he will get a crash on assertion instead of a diagnostic telling him what he did wrong. The user should always get diagnostics clearly showing him what he is doing wrong, without crashing on assertion. Therefore, I am returning the pull request to you for revision. Please return it to me after adding some checks in the code that give diagnostics and avoid subsequent crash in the assertion form, or an explanation why it is impossible/very problematic to add such a check.

janlindstrom:
@sysprg Problem is that this test does not reproduce the crash not when it is unchanged. I did small changes to make it pass as there was one timeout. If I correctly remember assertion can happen if you have MyISAM changes inside a transaction, this can't anymore happen because (1) we run so that mysql.gtid_slave_pos table is InnoDB as it is on release binaries and (2) because multi-engine changes containing non-transactional engines are now restricted on Galera. Therefore, there is nothing to be fixed on server code or in library code in this case.

sysprg:
Thanks, the fix has been merged with the head revision:  https://github.com/MariaDB/server/commit/94ef07d61edd21d81eaa788c07a15e5508aacbe0

PR #3550, opened 2024-09-25
MDEV-31888 : galera.galera_wan, galera.galera_vote_rejoin_* fail

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-31888*

## Description

Clean up configuration and tests. Add wait conditions to make sure test continues from clean state. Test case changes only.

## Release Notes

## How can this PR be tested?

- [x ] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check


sysprg:
@janlindstrom Thanks, but it's unclear why the generic galera_4nodes.cnf will now include log-bin for all tests that use it, including versioning_trx_id.test and GCF-360.test. Is this really always necessary when testing a 4-node configuration? The second question is increasing the gcache size to 20M - what is the need for that? I'm concerned that this may mask some bugs in other tests (other than the ones this fix is ​​intended to fix).

sysprg:
I have temporarily re-assigned this pull request back to you - please assign it back to me after you have finalized the configuration change so that it does not affect unrelated tests.

janlindstrom:
@sysprg Ok, I will restrict configuration change to this test only. In general case we should run almost all tests using both log-bin and binlog-off cases but that would make suites slower. I changed suspect and evict timeouts because in these tests starting of servers takes longer than normally.

janlindstrom:
Timeout change is needed for slow machines.

sysprg:
Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/7ffa7b6b0193738d011e514b66d875ce29bc11d2

PR #3552, opened 2024-09-26
MDEV-32780 : galera_as_slave_replay: assertion in the wsrep::transact…

janlindstrom:
…ion::before_rollback()

- [x] *The Jira issue number for this PR is: MDEV-32780*

## Description

Test changes only because assertion reported is not reproducible anymore.

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/disabled.def
+++ b/mysql-test/suite/galera/disabled.def
@@ -14,5 +14,3 @@ galera_as_slave_ctas : MDEV-28378 timeout
 galera_pc_recovery : MDEV-25199 cluster fails to start up
 galera_sequences : MDEV-32561 WSREP FSM failure: no such a transition REPLICATING -> COMMITTED
 galera_concurrent_ctas : MDEV-32779 galera_concurrent_ctas: assertion in the galera::ReplicatorSMM::finish_cert()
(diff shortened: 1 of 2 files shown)


sysprg:
@janlindstrom Thanks, but the fix does not include changes to the wsrep inserts or library code, it only includes changes to the test - yes, this will solve the problem with the test by bypassing the problematic place, but it will not fix anything for the user. The user should get diagnostics clearly showing him what he is doing wrong, without crashing on assertion. Therefore, I am returning the pull request to you for revision.

janlindstrom:
@sysprg Problem is that this test does not reproduce the crash not when it is unchanged. I did small changes to fortify it to make it pass as there was one result difference and one timeout. If I correctly remember assertion can happen if you have MyISAM changes inside a transaction, this can't anymore happen because (1) we run so that mysql.gtid_slave_pos table is InnoDB as it is on release binaries and (2) because multi-engine changes containing non-transactional engines are now restricted on Galera. Therefore, there is nothing to be fixed on server code or in library code in this case.

sysprg:
Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/d598ee3cf952e528a8b42261959360de637add2e

PR #482, opened 2017-11-01
MDEV-14531 - 10.3 fused madd ppc64

grooverdan:
From #303, roundings aren't required when -ffp-contract=off is used. This. This is enabled on ppc64 and not x86_64 based on tests. Extensive discussion https://github.com/mysql/mysql-server/pull/158.

FYI @akopytov.

My contributions are under the MCA.

PR #3553, opened 2024-09-26
MDEV-34969: test fail main.spatial_utility_function_simplify

DaveGosselin-MariaDB:
The commented=out tests are known to have different results on different architectures due to floating-point tolerance calculations being slightly different; higher tolerance answers for the same geometries match.


grooverdan:
Ah, like #482.  Good POC fix. I don't think it will take compilers long to started fusing a split bit of code.  Is `_attribute__((target("fp-contract=off"))` on a inline function for this secion?

DaveGosselin-MariaDB:
> Ah, like #482. Good POC fix. I don't think it will take compilers long to started fusing a split bit of code. Is `_attribute__((target("fp-contract=off"))` on a inline function for this secion?

Thank you 😄 I wasn't aware of `fp-contract=off`, I will update the patch accordingly.

DaveGosselin-MariaDB:
When I supply `__attribute__((target("fp-contract=off"))`, the compiler complains:
```
'+fp-contract=off' is not a recognized feature for this target (ignoring feature)
```
Googling a bit doesn't seem to turn up much useful.  Maybe I'm holding it wrong.  I looked at #482 and didn't see a use of it in that commit.  Additionally, `cmake/floating_point.cmake` is gone from our codebase, was there a problem with it?

grooverdan:
Was assuming it would behave like other compile options already used in the codebase.

#482 never got merged. There was something missing. Forget exactly what.

Think removing the couple of non-predictable result test cases is acceptable.

grooverdan (approved):

PR #3557, opened 2024-09-29
MDEV-34880 Incorrect result for query with derived table having TEXT …

Olernov:
…field

- [x] *The Jira issue number for this PR is: MDEV-34880*

## Description
When a derived table which has distinct values and BLOB fields is materialized, an index is created over all columns to ensure only distinct values are placed to the result.
This index is created in a special mode HA_UNIQUE_HASH to support BLOBs. Later in `best_access_path` the optimizer may incorrectly choose this index to retrieve values from the derived table, although such type of index cannot be used for data retrieval.

This commit excludes HA_UNIQUE_HASH indexes from consideration in `best_access_path`

## Release Notes

## How can this PR be tested?

A test case is added to `main/derived.test`

## PR quality check

### DIFF
--- a/mysql-test/main/derived.result
+++ b/mysql-test/main/derived.result
@@ -1862,3 +1862,19 @@ pk	a
 deallocate prepare stmt;
 drop table t1,t2,t3;
 # End of MariaDB 11.1 tests
+#
+# MDEV-34880: Incorrect result for query with derived table having TEXT field
+#
+CREATE TABLE t1 (id int NOT NULL  PRIMARY KEY, notes TEXT NOT NULL);
+INSERT INTO t1 VALUES (1, 'test1'), (2, 'test2');
+SELECT dt.* FROM (SELECT * FROM t1 UNION SELECT * FROM t1) dt WHERE id = 1;
+id	notes
+1	test1
+SELECT dt.* FROM (SELECT * FROM t1 UNION SELECT * FROM t1) dt WHERE id = 1 AND
+notes = 'test1';
+id	notes
+1	test1
+DROP TABLE t1;
+#
+# End of 11.2 tests
+#
--- a/mysql-test/main/derived.test
+++ b/mysql-test/main/derived.test
@@ -1514,3 +1514,20 @@ deallocate prepare stmt;
 drop table t1,t2,t3;
 
 --echo # End of MariaDB 11.1 tests
+
+--echo #
+--echo # MDEV-34880: Incorrect result for query with derived table having TEXT field
+--echo #
+CREATE TABLE t1 (id int NOT NULL  PRIMARY KEY, notes TEXT NOT NULL);
+INSERT INTO t1 VALUES (1, 'test1'), (2, 'test2');
+
+SELECT dt.* FROM (SELECT * FROM t1 UNION SELECT * FROM t1) dt WHERE id = 1;
+
+SELECT dt.* FROM (SELECT * FROM t1 UNION SELECT * FROM t1) dt WHERE id = 1 AND
+  notes = 'test1';
+
+DROP TABLE t1;
+
+--echo #
+--echo # End of 11.2 tests
+--echo #
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -3737,6 +3737,28 @@ int handler::ha_rnd_pos(uchar *buf, uchar *pos)
   DBUG_RETURN(result);
 }
 
+
+int handler::ha_index_init(uint idx, bool sorted)
+{
+  DBUG_EXECUTE_IF("ha_index_init_fail", return HA_ERR_TABLE_DEF_CHANGED;);
+  int result;
+  DBUG_ENTER("ha_index_init");
+  DBUG_ASSERT(inited==NONE);
+  if (!(result= index_init(idx, sorted)))
+  {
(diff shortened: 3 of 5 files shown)


spetrunia:
If the goal was to disallow using the index for lookups, one can still do that:
```
create table t0 (a int, b varchar(100));
insert into t10 values (1, 'aa'),(2,'bb');
insert into t0 values (1, 'aa'),(2,'bb');
```

```
explain SELECT dt.* FROM (SELECT id,notes FROM t1 UNION SELECT * FROM t1) dt, t0 WHERE dt.id=t0.a and dt.notes=t0.b;
Query OK, 0 rows affected (0.000 sec)
```
```
+------+--------------+------------+--------+---------------+--------------+---------+---------------------+------+-------------+
| id   | select_type  | table      | type   | possible_keys | key          | key_len | ref                 | rows | Extra       |
+------+--------------+------------+--------+---------------+--------------+---------+---------------------+------+-------------+
|    1 | PRIMARY      | t0         | ALL    | NULL          | NULL         | NULL    | NULL                | 2    | Using where |
|    1 | PRIMARY      | <derived2> | eq_ref | distinct_key  | distinct_key | 16      | test.t0.a,test.t0.b | 1    | Using where |
|    2 | DERIVED      | t1         | ALL    | NULL          | NULL         | NULL    | NULL                | 2    |             |
|    3 | UNION        | t1         | ALL    | NULL          | NULL         | NULL    | NULL                | 2    |             |
| NULL | UNION RESULT | <union2,3> | ALL    | NULL          | NULL         | NULL    | NULL                | NULL |             |
+------+--------------+------------+--------+---------------+--------------+---------+---------------------+------+-------------+
```
that's because you have handled only one branch in best_access_path.

But trying to take a wider look at this:

* If the index is not usable at all for best_access_path, perhaps we should avoid creating KEYUSE() objects describing use of this index?

* In the example in your testcase, the index is on `id, notes`.  And we try to construct a ref access using id=1.  Is it really not possible to use index in this case?  Why ?

spetrunia (changes requested):
Need to clarify the above. And add my testcase in either case.

Olernov:
Pushed a revised version where HA_UNIQUE_HASH keys do not create KEYUSE() objects. 
I am not sure whether it makes sense to add a similar exclusion in this path:
```
int collect_statistics_for_index(THD *thd, TABLE *table, uint index)
{
...
  /* No statistics for FULLTEXT indexes. */
  if (key_info->flags & (HA_FULLTEXT|HA_SPATIAL))
    DBUG_RETURN(rc);
```

> And we try to construct a ref access using id=1. Is it really not possible to use index in this case? Why ?
I believe that type of index obeys the rules for [regular hash indexes](https://mariadb.com/kb/en/storage-engine-index-types/#hash-indexes) which do not permit leftmost prefixing. I wonder why the index is not usable when all fields are specified (`id=1 AND notes='test1'`). Probably it is just not implemented at Maria storage engine.

Olernov:
Another point is should we skip `HA_UNIQUE_HASH` at  `SQL_SELECT::test_quick_select()` (opt_range.cc) as this is already done for `HA_FULLTEXT`:
```
if (key_info->flags & HA_FULLTEXT)
{
    trace_idx_details.add("usable", false).add("cause", "fulltext");
    continue;    // ToDo: ft-keys in non-ft ranges, if possible   SerG
}
```

spetrunia:
Reviewer note:  Is there any sense at all that indexes with HA_UNIQUE_HASH are present in `table-> keys_in_use_for_query` ? 

It is tempting that somewhere inside create_tmp_table() we just remove the unique-constraint index from `keys_in_use_for_query`...

.. and that won't work, as there is code elsewhere that clears `keys_in_use_for_query` and re-populates it.

This is one of the places:
```cpp
bool TABLE_LIST::process_index_hints(TABLE *tbl)
{   
  /* initialize the result variables */
  tbl->keys_in_use_for_query= tbl->keys_in_use_for_group_by=
    tbl->keys_in_use_for_order_by= tbl->s->usable_indexes(tbl->in_use);
```

Can one have TABLE_SHARE::usable_indexes to not return HA_UNIQUE_HASH indexes? (can limit  this logic table_category==TABLE_CATEGORY_TEMPORARY ... it is not EXACTLY "temporary work tables" but it will allow to avoid looping through indexes for regular tables...)

spetrunia:
@Olernov , https://gist.github.com/spetrunia/7852faf7d6344e4b48fedee54bebaf89  - implementation of suggestion from the call. What do you think?

Olernov:
@spetrunia , frankly I'm not sure I like this idea. `TABLE_SHARE::usable_indexes()` is called only by `TABLE_LIST::process_index_hints()`. It looks a bit strange to exclude HA_UNIQUE_HASH indexes in that code path given that the logic is not related to index hints. At the same time, `add_key_part()` from the previous solution is called by `update_refs_and_keys()` which is annotated as `Update keyuse array with all possible keys we can use to fetch rows`. This looks like a more appropriate place to add a new exclusion, especially taking into account that some similar exclusions are already there:
```
if (form->key_info[key].flags & (HA_FULLTEXT | HA_SPATIAL))
	continue; 
```

spetrunia (approved):
Looks good to me

PR #3560, opened 2024-09-30
MDEV-34450 os_file_write_func() is an overkill for ib_logfile0

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34450*
## Description
`log_file_t::read()`, `log_file_t::write()`: Invoke `pread()` or `pwrite()` directly, so that we can give more accurate diagnostics in case of a failure, and so that we will avoid the overhead of setting up 5(!) stack frames and related objects.
## Release Notes
This could be a minor performance improvement, probably not worth mentioning.
## How can this PR be tested?
The log writing and crash recovery are exercised by the existing test suite, in particular by the `innodb`, `encryption`, and `mariabackup` suites.

This could be applicable to other versions as well. 10.11 is the earliest maintained branch after the last major refactoring of the log reading and writing in 685d958e38b825ad9829be311f26729cccf37c46.
## PR quality check

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -46,6 +46,7 @@ Created 12/9/1995 Heikki Tuuri
 #include "buf0dump.h"
 #include "log0sync.h"
 #include "log.h"
+#include "tpool.h"
 
 /*
 General philosophy of InnoDB redo-logs:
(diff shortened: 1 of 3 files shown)


vaintroub (approved):
Looks OK to me, even though the loop around pread/pwrite seems overcautious.

Maybe this needs a followup to remove other synchronous reads and writes, if they are not need to be reflected by performance schema.

PR #21, opened 2015-03-11
Add Master_gtid_wait_{count,time,timeouts} status

grooverdan:
MASTER_GTID_WAIT function needs some status to evaluate its use.

master_gtid_wait_count indicates how many times the function is called.
master_gtid_wait_time indicates how much time in microseconds occurred
waiting (or timing out)
master_gtid_timeouts indicates how many time times this function timed
out rather than all successful gtids events being available.

From MDEV-6981

PR #20, opened 2015-03-11
Add Slave_skipped_errors to global status

grooverdan:
This counts the number of times a replication event is ignored due to slave_skip_errors. from mdev-7198

PR #19, opened 2015-03-11
MDEV-7384 Add --persistent option for mysqlcheck

grooverdan:
10.0 has an "analyze table .. persistent for all" syntax. This adds
--persistent to mysqlcheck(mysqlanalyize) to perform this extended
analyze table option.

From MDEV-7384

PR #18, opened 2015-02-28
Added regression test for MDEV-5871

pastcompute:
This is a regression test for MDEV-5871 that I never got merged into BZR
I had to untangle it whilst dealing with the fork/push for MDEV-6282

PR #3562, opened 2024-10-01
MDEV-35049: Improve adaptive hash index scalability

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35049, MDEV-35312*
## Description
This also includes #3574 and #3627 until they have been merged to 10.6.

I suggest to read the individual [commit mesages|https://github.com/MariaDB/server/pull/3562/commits] (`git log origin/10.6..origin/10.6-MDEV-35190`).

This is a rewrite of the InnoDB adaptive hash index, as well as a rewrite of record comparisons so that heap memory allocation as well as calls to `rec_get_offsets()` will be avoided.

`MEM_HEAP_BTR_SEARCH` and `mem_block_info_t::free_block` are removed. We implement a custom memory allocator for the adaptive hash index entries (`ahi_node`, renamed from `ha_node_t`). The `mem_heap` subsystem is simplfiied.

We also make `rec_fold()` compute CRC-32C on a contiguous range of bytes, from the start of the record to the last included byte of the last field. In `dtuple_fold()` we compute the checksum field by field.

The function `page_cur_search_with_match_bytes()`, which can benefit `INSERT` into the end of an index page, will be enabled even when `innodb_adaptive_hash_index=OFF`.

`btr_search_guess_on_hash()` will return `BTR_CUR_HASH_FAIL` in fewer cases. This will lead to less frequent rebuild of the adaptive hash index and therefore better performance in a read-only workload.

Many fields related to comparing records and the adaptive hash index are consolidated into a single `n_bytes_fields` or `left_bytes_fields`. A redundant set of AHI related parameters is removed from `buf_block_t`. The `buf_block_t::ahi_left_bytes_fields` will use `Atomic_relaxed<uint32_t>`.
## Release Notes
The performance of `innodb_adaptive_hash_index=ON` was improved at high concurrency.
## How can this PR be tested?
```sh
./mtr --parallel=auto --big-test --force --mysqld=--loose-innodb-adaptive-hash-index=ON --max-test-fail=0
```
I tested this with and without `cmake -DWITH_INNODB_AHI=OFF` as well as with `cmake -DWITH_INNODB_ROOT_GUESS=OFF`.

With the `innodb-hashtest.sh` that @montywi attached to [MDEV-35049](https://jira.mariadb.org/browse/MDEV-35049) I no longer see any waits for the `btr_sea::partition::latch`, and there are no calls to `btr_sea::partition::insert` during the workload. This means that the adaptive hash index remains constant during a constant read-only workload where the data fits in the buffer pool.

This needs to be tested with a larger workload that originally reproduced the scalability issue.

Part of this could be applicable to 10.5, but the patch is currently is based on the 10.6 branch, because it depends on many performance improvements that are only available starting with 10.6.
## PR quality check


dr-m:
Sorry, this needs some more work:
```
#17 0x00007f78d3e42592 in __assert_fail (assertion=0x5617ede15ee8 "(table->freed_indexes).count == 0", file=0x5617ede15da8 "/mariadb/10.6/storage/innobase/dict/dict0mem.cc", line=0xcf, function=0x5617ede15ea0 "void dict_mem_table_free(dict_table_t*)") at ./assert/assert.c:103
#18 0x00005617ed7cc4fc in dict_mem_table_free (table=0x7f7894290990) at /mariadb/10.6/storage/innobase/dict/dict0mem.cc:207
#19 0x00005617ed5e127f in row_import_cleanup (prebuilt=0x7f789428d480, err=DB_SUCCESS, fts_table=0x7f7894290990) at /mariadb/10.6/storage/innobase/row/row0import.cc:2277
#20 0x00005617ed5e9fde in row_import_for_mysql (table=0x7f7894290990, prebuilt=0x7f789428d480) at /mariadb/10.6/storage/innobase/row/row0import.cc:4991
#21 0x00005617ed428669 in ha_innobase::discard_or_import_tablespace (this=0x7f7894293d40, discard=0x0) at /mariadb/10.6/storage/innobase/handler/ha_innodb.cc:13403
```
However, this should not prevent any performance testing.

storage/innobase/btr/btr0sea.cc
```diff
@@ -167,21 +167,19 @@ will not guarantee success.
...
 static void btr_search_check_free_space_in_heap(const dict_index_t *index)
 {
-  /* Note that we peek the value of heap->free_block without reserving
-  the latch: this is ok, because we will not guarantee that there will
-  be enough free space in the hash table. */
-
-  buf_block_t *block= buf_block_alloc();
-  auto part= btr_search_sys.get_part(*index);
-
-  part->latch.wr_lock(SRW_LOCK_CALL);
```
dr-m, line 177: On a closer thought, we may need `rd_lock(SRW_LOCK_CALL)` here in order to prevent a race condition with `mem_heap_free(part->heap)` or `mem_heap_empty(part->heap)` (which I think should be covered by an exclusive latch). I am not sure if there can be such calls, so I must check this, and add a source code comment, with or without the latch acquisition, as appropriate.

dr-m: We indeed must check `btr_search_enabled` while holding `part->latch` in order to avoid reintroducing a race condition with a concurrent `btr_search_disable()`, which had been fixed in ad2bf1129cfa85c00072b46e0355fe14bf69ee54.

dr-m:
> Sorry, this needs some more work:
> 
> ```
> #17 0x00007f78d3e42592 in __assert_fail (assertion=0x5617ede15ee8 "(table->freed_indexes).count == 0", file=0x5617ede15da8 "/mariadb/10.6/storage/innobase/dict/dict0mem.cc", line=0xcf, function=0x5617ede15ea0 "void dict_mem_table_free(dict_table_t*)") at ./assert/assert.c:103
> #18 0x00005617ed7cc4fc in dict_mem_table_free (table=0x7f7894290990) at /mariadb/10.6/storage/innobase/dict/dict0mem.cc:207
> #19 0x00005617ed5e127f in row_import_cleanup (prebuilt=0x7f789428d480, err=DB_SUCCESS, fts_table=0x7f7894290990) at /mariadb/10.6/storage/innobase/row/row0import.cc:2277
> #20 0x00005617ed5e9fde in row_import_for_mysql (table=0x7f7894290990, prebuilt=0x7f789428d480) at /mariadb/10.6/storage/innobase/row/row0import.cc:4991
> #21 0x00005617ed428669 in ha_innobase::discard_or_import_tablespace (this=0x7f7894293d40, discard=0x0) at /mariadb/10.6/storage/innobase/handler/ha_innodb.cc:13403
> ```
> 
> However, this should not prevent any performance testing.

This turned out to be an independent bug, which I fixed in cc70ca7eab449912b6e2f99c56265eab826e5645.

dr-m:
ee9174350d7d324662981f8729e662902ba0537f switches from `srw_spin_lock` to `srw_lock` without spin loops, and 62dca98f5c800bff62ab3c02feac7ac6fb6f8875 implements a further optimization to `btr_search_sys_t::partition::erase()`, which is renamed from `ha_search_and_delete_if_found()`.

dr-m:
994a740027107b93b6bfd4c3f08487146b3e0bb2 reverts another attempt to use lock upgrade again. It would lead to occasional deadlocks when running `./mtr --mysqld=--loose-innodb-adaptive-hash-index=ON`.

storage/innobase/srv/srv0srv.cc
```diff
@@ -811,11 +811,10 @@ srv_printf_innodb_monitor(
 	for (ulint i = 0; i < btr_ahi_parts && btr_search_enabled; ++i) {
 		const auto part= &btr_search_sys.parts[i];
 		part->latch.rd_lock(SRW_LOCK_CALL);
-		ut_ad(part->heap->type == MEM_HEAP_FOR_BTR_SEARCH);
 		fprintf(file, "Hash table size " ULINTPF
 			", node heap has " ULINTPF " buffer(s)\n",
 			part->table.n_cells,
-			part->heap->base.count - !part->heap->free_block);
+			part->heap->base.count - !part->heap->ahi_block);
 		part->latch.rd_unlock();
```
dr-m, line 818: If `btr_search_enabled` had been cleared between our check and the acquisition of `part->latch`, we’d be dereferencing `part->heap=nullptr` here.

A possible fix could be to ensure that `part->heap` will always remain allocated. In that way, we would not have to check for `btr_search_enabled` here at all, and `btr_search_check_free_space_in_heap()` could safely use atomic memory access instead of acquiring `part->latch`.

storage/innobase/include/btr0sea.h
```diff
@@ -240,111 +239,73 @@ struct btr_search_sys_t
...
 
   /** Partitions of the adaptive hash index */
   partition *parts;
 
   /** Get an adaptive hash index partition */
-  partition *get_part(index_id_t id, ulint space_id) const
+  partition *get_part(index_id_t id, ulint space_id) const noexcept
   {
     return parts + ut_fold_ulint_pair(ulint(id), space_id) % btr_ahi_parts;
   }
```
dr-m, line 285: This had better be just
```c++
    return parts + ulint(id) % btr_ahi_parts;
```

dr-m:
The revert 36a8b44ebd96ec9a8d449c83248109d5e893f534 was an attempt to see if it would fix a race condition between buffer pool resizing and the adaptive hash index. Unfortunately, while 2 runs of the mandatory amd64-debian-11-msan-clang-16 builder were successful, the 3rd attempt reproduced the assertion failures in `innodb.innodb_buffer_pool_resize`. I will investigate that further.

dr-m (commented):
It looks like I figured out the cause of some trouble. Hopefully addressing these will allow some stress tests to pass.

storage/innobase/btr/btr0sea.cc
```diff
@@ -1034,1066 +1041,797 @@ TRANSACTIONAL_TARGET
...
+  modified or evicted (buf_page_t::can_relocate() will not hold) while
+  we validate the guessed rec. Thus, releasing part.latch is safe. */
+  part.latch.rd_unlock();
 
-	while (const rec_t* next_rec = page_rec_get_next_const(rec)) {
-		if (page_rec_is_supremum(next_rec)) {
+  const auto state= block->page.state();
 
-			if (!left_side) {
+  if (UNIV_UNLIKELY(state < buf_page_t::UNFIXED))
```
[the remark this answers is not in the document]
dr-m: I got some further thoughts regarding this. What if instead of just reading `block->page.state()` we invoked `block->page.fix()` already here? We will be doing that a little later anyway, unless the `state` was `buf_page_t::REMOVE_HASH`. The only question is how to make sure that a possible concurrent execution of `buf_page_t::REMOVE_HASH` will not run into any trouble. A possible solution could be to protect also this increment under the `buf_pool.page_hash` latch, so that we can back off without damage if we had incremented `buf_page_t::REMOVE_HASH` to `buf_page_t::FREED`. If the `buf_page_t::lock` and the `buf_page_t::state()` field are located in different cache line, this would seem to be a bad idea: adding 1 more cache line access (a read-modify-write access at that) and conditional branch inside the critical section of a `buf_pool.page_hash` latch.

vlad-lesin (approved):
I did not find obvious errors during code review. Fixed-size allocator, crc32-c as hash function for hash index, hashing contiguous array of bytes instead of separate fields, `part.blocks_mutex` along with `part.spare` and other changes should improve hash index performance.

dr-m:
These changes were applied to the main branch in the commits between 9c8bdc6c15a89ffe9a9bd64d518fe90eb50edcf5 and 18dbeae1b86f1b6c0c29acd02e7e0dfbca91181c. The development base was 10.6, so that it would be easier to update the branch for fixing any unrelated bugs that were encountered during testing.

dr-m:
I had also created the branch bb-10.11-MDEV-35049-rebase, whose current head revision is a253f935fa4cf3a0843a43a2ba411c683b1473d5.

PR #2170, opened 2022-06-24
MDEV-35984 - Add MTR tests for MySQL (5.7) to MariaDB migration

PhysicsTing:
## Description

Currently upgrades from MySQL 5.7 to MariaDB fail on certain edge
cases. This commit adds MTR tests to reproduce those issues. The purpose
of this is to make the problems visible and facilitate the development
of the proper upgrade code in MariaDB. The MTR tests are added to skip
list as they will fail for all branches until these incompatibilities
are addressed.

Populated a MySQL 5.7 data directory from official MySQL docker image.
The scripts are stored in mysql-test/lib, and the data directory is
stored in std_data to be consumed by MTR directly.

Add MTR tests for the usual migration procedure from MySQL to MariaDB.
  1. Start MariaDB from the MySQL data directory
  2. Run mariadb-upgrade
  3. Verify that the upgrade succeed

Also highlight the incompatibilities that are either not documented
before or documented but the detailed scenario are not clarified. These
tests will fail for now but can be used as references for future to
improve these incompatibilities.

Focus on MySQL 5.7 migration for now.

## How can this PR be tested?

Newly created MTR tests failed but that is intentional.
Existing MTR tests succeed as expected.

## Basing the PR against the correct MariaDB version
- [x] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced: 10.5*

## Backward compatibility
Backward compatibility is not compromised, and will be improved based on the newly added tests.

## Copyright
All new code of the whole pull request, including one or several files
that are either new files or modified ones, are contributed under the
BSD-new license. I am contributing on behalf of my employer Amazon Web
Services, Inc.

PR #3563, opened 2024-10-01
WIP: MDEV-36007 - Implement major version upgrade testing with binary files

anson1014:
### **Implement major version upgrade testing with binary files**

In order to increase the confidence, validation and coverage of major version upgrade testing, this approach similarly follows the pre-existing method of loading dumps from prior versions into the data-directory and attempting an upgrade. Instead of using the dumps of entire schemas, we do the same with specific binary files of tables.

The PR to expand the versions and testing for the upgrading from dump approach can be found here: #3503.

**Currently, only binary files from 10.11 are used. Furthermore, only a subset of all available binary files of tables are being included. The intent is to validate the approach of this method first and then further extend the scope of what is covered by including binary files from other versions and the binary files for the other mysql tables.**

The act of loading binary files from prior version into the data directory and then performing operations upon them was inspired by [#2170](https://github.com/MariaDB/server/pull/2170). Feedback this regarding this approach and ways to best implement testing for major version upgrades are welcomed and greatly appreciated.

## Release Notes

N/A

## How can this PR be tested?

Changes are only to testing, no functionality is changed. Introduced tests are passing.

```
main.mysql_binary_upgrade                w8 [ pass ]   1355
```

- [x] _This is a change in testing that is targeted towards the current MariaDB LTS._

## PR quality check

---


anson1014:
Similar to #3503

LinuxJedi:
Whilst I get it, these tests are way more useful than the original tests behind #3503, I don't think this is a good idea here for several reasons. My arguments won't count for much, I know people will try and counter them, but it will be why I will abstain from reviewing it:

1. The on-disk binary format _might_ be platform / endian dependent.
2. This should cover all engines and data types, not just the auth tables.
3. The matrix for this test will end up being so wide that it is better to do it as part of a CI system that installs the old version, sets up some tables and executes the upgrade, running some post-upgrade tests on the data.
**Note:** If this is difficult to do, then that screams that there is a usability issue for users too, who are trying to automate these things as well.
4. Both this and the original tests behind #3503 don't really test data integrity. More "yea, that looks like the right number of rows in a table, we'll call that good".
5. As little binary data as possible should be in the repo. I'd prefer scripted generators instead (which I know will not be easy in this case).

As it is, I think these kind of tests are as good as a chocolate fireguard without a wider overhaul of them.

LinuxJedi:
> The reason we are not doing it in the official CIs is that the matrix is actually so big that it doesn't fit even there, at least not on per-push basis. We need a solution for scheduled runs to do it. I mean, it can be done for default installations, we have stubs for that in package upgrade tests, it would only need a bit more meaningful data. But it won't be a comprehensive matrix either.

I assume by this you mean there is finite compute resources available? This is something that should be easy to parallelise across hundreds of VMs or containers so it all runs in minutes. I can even think of several optimisations that can be made to make it quicker, particularly with containers (although it would probably work with VMs too).

PR #3570, opened 2024-10-10
MDEV-34919 - Aria crashes with high-level (vector) indexes

svoj:
- [x] *The Jira issue number for this PR is: MDEV-34919*

## Description
Fixes aria crash with high-level indexes.

## Release Notes
None.

## How can this PR be tested?
mtr vector

## PR quality check


vuvova:
Looks good. I've generalized it a bit, which allowed me to remove my InnoDB hack (that was needed because of missing store_lock).

Pushed into the vector branch. Closing.

svoj:
@vuvova, fine with me. Just one thing re 07b74febdcd16b5bb1c72ed9015bab51493a4e08: `store_lock()` may return multiple `THR_LOCK_DATA`, but you moved it out of `if (lock_count() == 1)` guard. Do you feel like it is not going to be relevant for this code ever? May be at least `DBUG_ASSERT(lock_count() <= 1)`?

vuvova:
I have an assert below, and I believe effect is the same. But I'll move it up indeed, it'll be a bit more obvious

PR #3571, opened 2024-10-10
MDEV-34830: LSN in the future is not being treated as serious corruption

dr-m:
This is a 10.11 version of #3488.
- [x] *The Jira issue number for this PR is: MDEV-34830*
## Description
The invariant of write-ahead logging is that before any change to a page is written to the data file, the corresponding log record must must first have been durably written.

On crash recovery, there were some sloppy checks for this. Let us implement accurate checks and flag an inconsistency as a hard error, so that we can avoid further corruption of a corrupted database. For data extraction from the corrupted database, `innodb_force_recovery` can be used.

Before recovery is reading any data pages or invoking `buf_dblwr_t::recover()` to recover torn pages from the doublewrite buffer, InnoDB will have parsed the log until the final LSN and updated `log_sys.lsn` to that. So, we can rely on `log_sys.lsn` at all times. The doublewrite buffer recovery has been refactored in such a way that the `recv_sys.dblwr.pages` may be consulted while discovering files and their page sizes, but nothing will be written back to data files before `buf_dblwr_t::recover()` is invoked.

`recv_max_page_lsn`, `recv_lsn_checks_on`: Remove.

`recv_sys_t::validate_checkpoint()`: Validate the write-ahead-logging condition at the end of the recovery.

`recv_dblwr_t::validate_page()`: Keep track of the maximum LSN (if we are checking a non-doublewrite copy of a page) but do not complain LSN being in the future. The doublewrite buffer is a special case, because it will be read early during recovery. Besides, starting with commit 762bcb81b5bf9bbde61fed59afb26417f4ce1e86 the dblwr=true copies of pages may legitimately be "too new".

`recv_dblwr_t::find_page()`: Find a valid page with the smallest `FIL_PAGE_LSN` that is in the valid range for recovery.

`recv_dblwr_t::restore_first_page()`: Replaced by `find_page()`. Only `buf_dblwr_t::recover()` will write to data files.

`buf_dblwr_t::recover()`: Simplify the message output. Do attempt doublewrite recovery on user page read error. Ignore doublewrite pages whose `FIL_PAGE_LSN` is outside the usable bounds. Previously, we could wrongly recover a too new page from the doublewrite buffer. It is unlikely that this could have lead to an actual error. Write back all recovered pages from the doublewrite buffer here, including for the first page of any tablespace.

`buf_page_is_corrupted()`: Distinguish the return values `CORRUPTED_FUTURE_LSN` and `CORRUPTED_OTHER`.

`buf_page_check_corrupt()`: Return the error code `DB_CORRUPTION` in case the LSN is in the future.

`Datafile::read_first_page()`: Handle `FSP_SPACE_FLAGS=0xffffffff` in the same way on both 32-bit and 64-bit architectures.

`Datafile::read_first_page_flags()`: Split from `read_first_page()`. Take a copy of the first page as a parameter.

`recv_sys_t::free_corrupted_page()`: Take the file as a parameter and return whether a message was displayed. This avoids some duplicated and incomplete error messages.

`buf_page_t::read_complete()`: Remove some redundant output and always display the name of the corrupted file. Never return `DB_FAIL`; use it only in internal error handling.

`IORequest::read_complete()`: Assume that `buf_page_t::read_complete()` will have reported any error.

`fil_space_t::set_corrupted()`: Return whether this is the first time the tablespace had been flagged as corrupted.

`Datafile::validate_first_page()`, `fil_node_open_file_low()`, `fil_node_open_file()`, `fil_space_t::read_page0()`,
`fil_node_t::read_page0()`: Add a parameter for a copy of the first page, and a parameter to indicate whether the `FIL_PAGE_LSN` check should be suppressed. Before `buf_dblwr_t::recover()` is invoked, we cannot validate the `FIL_PAGE_LSN`, but we can trust the `FSP_SPACE_FLAGS` and the tablespace ID that may be present in a potentially too new copy of a page.
## Release Notes
InnoDB was silently ignoring some inconsistent data pages (with log sequence number in the future). Now it is being treated in a similar way as a page checksum mismatch.
## How can this PR be tested?
The regression test `innodb.corrupted_during_recovery` will try to assign a future page LSN.

We also need stress tests with RQG and analyze any cases where recovery is failing.
## PR quality check


storage/innobase/buf/buf0buf.cc
```diff
@@ -4041,15 +4077,21 @@ dberr_t buf_page_t::read_complete(const fil_node_t &node)
...
+                      " read in from %s are "
+                      "[page id: space=" UINT32PF ", page number=" UINT32PF
+                      "], should be "
+                      "[page id: space=" UINT32PF ", page number=" UINT32PF
+                      "]",
+                      node.name,
+                      read_id.space(), read_id.page_no(),
+                      expected_id.space(), expected_id.page_no());
+      err= DB_FAIL;
       goto release_page;
```
mariadb-DebarunBanerjee, line 4095: I see we are setting err to DB_FAIL here to indicate we have already reported the error and changing later to DB_PAGE_CORRUPTED as the function is no longer supposed to return DB_FAIL. A bit round about logic but should be fine.

mariadb-DebarunBanerjee (approved):
Looks fine to me.

storage/innobase/buf/buf0buf.cc
dr-m: Right, with that we are signaling that we already reported a more specific error and do not want a generic error to be output:
```c++
release_page:
    if (recv_sys.free_corrupted_page(expected_id, node));
    else if (err == DB_FAIL)
      err= DB_PAGE_CORRUPTED;
    else
    {
      sql_print_error("InnoDB: Failed to read page " UINT32PF
                      " from file '%s': %s", expected_id.page_no(),
                      node.name, ut_strerr(err));
```
We must call `recv_sys.free_corrupted_page()` in every case, and I did not want to duplicate that call.

PR #3573, opened 2024-10-14
MDEV-35122  Incorrect NULL value handling for instantly dropped BLOB columns

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35122*

## Description
Problem:
=======
- Redundant table fails to insert into the table after instant drop blob column. Instant drop column only marking the column as hidden and consecutive insert statement tries to insert NULL value for the dropped BLOB column and returns the fixed length of the blob type as 65535. This lead to row size too large error.

Fix:
====
 For redundant table, if the non-fixed dropped column can be null
then set the length of the field type as 0.

## How can this PR be tested?
./mtr innodb.instant_alter_bugs

## PR quality check


mysql-test/suite/innodb/t/instant_alter_bugs.test
```diff
@@ -565,4 +565,13 @@ ALTER TABLE t1 DROP COLUMN a;
 ALTER TABLE t1 DROP COLUMN c;
 DROP TABLE t1;
 
+--echo #
+--echo #  MDEV-35122  Incorrect NULL value handling for instantly
+--echo #                 dropped BLOB columns
+--echo #
+CREATE TABLE t1 (c1 INT, c2 BLOB) ROW_FORMAT=REDUNDANT,ENGINE=InnoDB;
+ALTER TABLE t1 DROP c2;
```
dr-m, line 573: Could we also add and drop a `c3 BLOB NOT NULL` column to exercise that code path as well?

storage/innobase/row/row0row.cc
```diff
@@ -250,6 +250,14 @@ row_build_index_entry_low(
 			dict_col_copy_type(f.col, &dfield->type);
 			if (f.col->is_nullable()) {
 				dfield_set_null(dfield);
+				if (!dict_table_is_comp(index->table) &&
+				    f.col->mtype == DATA_BINARY) {
```
dr-m, line 254: `f.col` was just recently accessed, so it ought to be cheaper to test the condition on `f.col->mtype` first, before dereferencing two pointers:
```c++
				if (f.col->mtype == DATA_BINARY
				    && !index->table->not_redundant())
```
Also note that because this function was formatted in the original InnoDB style, we should follow that style.

dr-m (approved):

PR #3578, opened 2024-10-15
MDEV-35019 Provide a way to enable "rollback XA on disconnect" behavi…

knielsen:
…or we had before 10.5.2

Implement variable legacy_xa_rollback_at_disconnect to support backwards compatibility for applications that rely on the pre-10.5 behavior for connection disconnect, which is to rollback the transaction (in violation of the XA specification).

Andrei: This is a patch to support the pre-10.5 behavior that disconnect rolls back pending XA PREPARE, as was needed for some users with applications that relied on that behaviour; like we discussed briefly last week.

### DIFF
--- a/mysql-test/main/mysqld--help.result
+++ b/mysql-test/main/mysqld--help.result
@@ -433,6 +433,13 @@ The following specify which files/extra groups are read (specified before remain
  --lc-time-names=name 
  Set the language used for the month names and the days of
  the week.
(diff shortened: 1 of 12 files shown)


andrelkin (approved):
Looks good.
Thanks!

dr-m:
This was reverted in 5ebda30ccc7616443a5e29bca806a1095b7d90cc by @vuvova.

PR #3582, opened 2024-10-16
MDEV-35174 Possible hang in trx_undo_prev_version()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34174*
## Description
In commit b7b9f3ce825a08f16ad851c16f603365ae473195 (MDEV-34515) we accidentally made the InnoDB MVCC code acquire a shared purge_sys.latch twice. Recursive shared latch acquisition may cause a deadlock of InnoDB threads if another thread in between will start waiting for an exclusive latch.

`purge_sys_t::latch`: In debug builds, use `srw_lock_debug` instead of `srw_spin_lock`, so that bugs like this will result in debug assertion failures.

`trx_undo_report_row_operation()`: Pass the `view_guard` to `trx_undo_prev_version()` and the rest of the arguments in the same order, so that the work to permute argument registers is minimized.
## Release Notes
This does not need to be mentioned, because there is no release that would include the bug but not this fix.
## How can this PR be tested?
Stress test with any workload that includes transaction rollbacks.
## PR quality check

### DIFF
--- a/storage/innobase/include/trx0purge.h
+++ b/storage/innobase/include/trx0purge.h
@@ -125,7 +125,8 @@ class purge_sys_t
 
 public:
   /** latch protecting view, m_enabled */
-  alignas(CPU_LEVEL1_DCACHE_LINESIZE) mutable srw_spin_lock latch;
+  alignas(CPU_LEVEL1_DCACHE_LINESIZE)
+  mutable IF_DBUG(srw_lock_debug,srw_spin_lock) latch;
 private:
   /** Read view at the start of a purge batch. Any encountered index records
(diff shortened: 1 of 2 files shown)


mariadb-DebarunBanerjee (approved):
Looks fine to me. 
We should also check other callers of `purge_sys.is_purgeable`  and assess similar risk.

dr-m:
> Looks fine to me. We should also check other callers of `purge_sys.is_purgeable` and assess similar risk.

Thank you. I actually did that. Each one is outside any MVCC code path.

PR #3584, opened 2024-10-17
MDEV-35189: Updating cache for INNODB_LOCKS et al is suboptimal

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35189*
## Description
There are some performance issues with the cache that is underlying the `INFORMATION_SCHEMA` tables `INNODB_LOCKS`, `INNODB_LOCK_WAITS` and `INNODB_TRX`.

`ha_storage_put_memlim()`: Invoke `my_crc32c()` just once to "fold".

`fold_lock()`: Remove some redundant conditions and use `my_crc32c()` instead of `ut_fold_ulint_pair()`.

`add_lock_to_cache()`: Invoke `fold_lock()` only once.

TODO (MDEV-35190): Combine `HASH_SEARCH` and `HASH_INSERT` similar to `buf_pool.page_hash.cell_get()`.
## Release Notes
There performance of InnoDB hash table operations was slightly improved.
## How can this PR be tested?
```sh
./mtr --parallel=auto innodb.innodb_i_s_innodb_trx innodb.innodb_i_s_innodb_locks
```

Some of this would be applicable to 10.5 as well, but we target 10.6 because of the MDEV-35190 dependency.
## PR quality check

### DIFF
--- a/storage/innobase/trx/trx0i_s.cc
+++ b/storage/innobase/trx/trx0i_s.cc
@@ -59,47 +59,6 @@ now, then 39th chunk would accommodate 1677416425 rows and all chunks
 would accommodate 3354832851 rows. */
 #define MEM_CHUNKS_IN_TABLE_CACHE	39
 
-/** The following are some testing auxiliary macros. Do not enable them
-in a production environment. */
-/* @{ */
-
-#if 0
-/** If this is enabled then lock folds will always be different
-resulting in equal rows being put in a different cells of the hash
-table. Checking for duplicates will be flawed because different
-fold will be calculated when a row is searched in the hash table. */
-#define TEST_LOCK_FOLD_ALWAYS_DIFFERENT
-#endif
-
-#if 0
-/** This effectively kills the search-for-duplicate-before-adding-a-row
-function, but searching in the hash is still performed. It will always
-be assumed that lock is not present and insertion will be performed in
-the hash table. */
-#define TEST_NO_LOCKS_ROW_IS_EVER_EQUAL_TO_LOCK_T
-#endif
-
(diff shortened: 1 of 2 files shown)


storage/innobase/trx/trx0i_s.cc
```diff
@@ -819,112 +785,18 @@ static
...
-
-	if (hash_chain == NULL) {
-
-		return(NULL);
-	}
-	/* else */
-
-	return(hash_chain->value);
+  ut_ad((heap_no == 0xFFFF) == lock.is_table());
+  if (heap_no == 0xFFFF)
```
mariadb-DebarunBanerjee, line 794: I would prefer the condition to be `if (lock.is_table())` which should be used uniformly in all places. heap_no is not relevant for table lock and to have the value as "0xFFFF" is good but we better not create functional dependency on it.

mariadb-DebarunBanerjee (approved):
The improvement looks good. Added one minor suggestion.

On a independent note, it does seem to be a very good idea to get rid of the has table abstraction and pull in the search/insert code in different part of the code. This comment is not specific to this patch ... just a note.

storage/innobase/trx/trx0i_s.cc
dr-m: The `0xFFFF` is special only within this subsystem, returned by the function `wait_lock_get_heap_no()`. We have debug assertions in place (including the one that you quoted) to document this special rule. In non-debug builds, referring to `heap_no == 0xFFFF` should yield better performance than dereferencing `lock`, because that parameter is likely to reside in a register.

PR #3585, opened 2024-10-18
Enable Interprocedural Optimization

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-______* (first a draft though CI to see what impacts it has)

## Description

Interprocedural Optimization (LTO) provides generally good benefits for performance. Lets enable this where supported.

To achieve this gain we need to be strict and accurate with function prototypes and casting and not rely on undefined behaviour.

## Release Notes

Enable Interprocedural Optimization (LTO) for most of the codebase.

## How can this PR be tested?

Though all the test in CI.

- [X] *This is an optimization used by some distributions. To make MariaDB ready for them we test it here first.

## PR quality check

### DIFF
--- a/configure.cmake
+++ b/configure.cmake
@@ -22,7 +22,7 @@ INCLUDE (CheckFunctionExists)
 INCLUDE (CheckCCompilerFlag)
 INCLUDE (CheckCSourceRuns)
 INCLUDE (CheckSymbolExists)
-
+INCLUDE (CheckIPOSupported)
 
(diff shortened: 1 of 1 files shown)


vaintroub:
Does not make sense to enable it on Debug builds, and even on Release builds, it should be optional, as it (usually) takes longer time. It makes sense to enable it on release builds on CI

grooverdan:
This was largely a test of what fails and where. Thinking maybe a full test CI. I'm trying keep "options" to a managed set we can actually test.

PR #3587, opened 2024-10-21
MDEV-35149 Race condition around SET GLOBAL innodb_lru_scan_depth

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35149*
## Description
A debug assertion in `buf_LRU_get_free_block()` could fail if `SET GLOBAL innodb_lru_scan_depth` is being executed during a workload that involves allocating buffer pool pages.

`buf_pool_t::LRU_scan_depth`: Replaces `srv_LRU_scan_depth`.

`buf_pool_t::flush_neighbors`: Replaces `srv_flush_neighbors`.

`innodb_buf_pool_update<T>()`: Update a parameter of `buf_pool` while holding `buf_pool.mutex`.
## Release Notes
This is not worth mentioning, because `SET GLOBAL` is a rarely executed operation that requires SUPER privileges, and because there likely is no bad impact outside `CMAKE_BUILD_TYPE=Debug`.
## How can this PR be tested?
The variables are covered by the regression test suite. In addition, the stress test that reproduced the bug in the first place should be run.

This bug affects 10.5 as well, but because the code has been refactored between 10.5 and 10.6 and the estimated impact of the bug is very low, we will only fix it starting with 10.6 in order to avoid merge conflicts.
## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -18360,6 +18360,16 @@ innobase_fts_find_ranking(FT_INFO* fts_hdl, uchar*, uint)
 	return(fts_retrieve_ranking(result, ft_prebuilt->fts_doc_id));
 }
 
+/** Update a field that is protected by buf_pool.mutex */
+template<typename T>
+static void innodb_buf_pool_update(THD *thd, st_mysql_sys_var *,
+                                   void *val, const void *save)
+{
+  mysql_mutex_lock(&buf_pool.mutex);
+  *static_cast<T*>(val)= *static_cast<const T*>(save);
+  mysql_mutex_unlock(&buf_pool.mutex);
+}
+
 #ifdef UNIV_DEBUG
 static my_bool	innodb_log_checkpoint_now = TRUE;
 static my_bool	innodb_buf_flush_list_now = TRUE;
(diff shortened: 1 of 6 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -18360,6 +18360,16 @@ innobase_fts_find_ranking(FT_INFO* fts_hdl, uchar*, uint)
...
 
+/** Update a field that is protected by buf_pool.mutex */
+template<typename T>
+static void innodb_buf_pool_update(THD *thd, st_mysql_sys_var *,
+                                   void *val, const void *save)
+{
+  mysql_mutex_lock(&buf_pool.mutex);
+  *static_cast<T*>(val)= *static_cast<const T*>(save);
+  mysql_mutex_unlock(&buf_pool.mutex);
+}
```
mariadb-DebarunBanerjee, line 18371: Looks correct. Some things to note here. I would not recommend any change right now but let's just note in case any user comes back with similar observation.

1. We could possibly get blocked for some time here waiting for buffer pool mutex. User might observe the difference here as these modifications are generally considered asynchronous.

2. A related point is that we are also likely holding some configuration mutex from server here and if we block here it might block update to other configuration variables. I don't see a possibility of deadlock unless we ask back for configuration information while holding the buffer pool mutex.

mariadb-DebarunBanerjee (approved):
The changes looks safe and correct to me. I have added a note.

storage/innobase/handler/ha_innodb.cc
dr-m: Right, we would be holding `LOCK_global_system_variables` while waiting for the `buf_pool.mutex`. In some places, where the wait is expected to be much longer, we are releasing `LOCK_global_system_variables` while waiting. This could open an opportunity of further race conditions. Here I do not see the need for it; `buf_pool.mutex` should never be held for an extensive period of time. The same can’t currently be said about `dict_sys.latch`, for instance.

PR #3591, opened 2024-10-23
MDEV-29015/MDEV-29260/MDEV-34938: os_file_get_size() WSL work-around

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-34938*
## Description
When MariaDB Server is run in a container under Windows Subsystem for Linux, the `fstat(2)` system calls that InnoDB invokes in `os_file_set_size()` or `os_file_get_size()` are causing a failure in case the file had been renamed in the past. This could happen at least on `ALTER TABLE` or `OPTIMIZE TABLE`.

`os_file_get_size()`: Invoke `lseek(2)` or `SetFilePointerEx()` instead of `fstat(2)`. We do not mind if the file pointer is moving to the end of the file, because InnoDB exclusively invokes positioned reads and writes, or in some rare cases, appends to an existing file.

`os_file_set_size()`: Invoke `os_file_get_size()` instead of `fstat()`.
## Release Notes
Under Windows Subsystem for Linux, MariaDB Server should no longer crash when executing some DDL statements, such as `ALTER TABLE` or `OPTIMIZE TABLE`.
## How can this PR be tested?
This is a low-level change that should be very well covered by the existing regression test suite that creates or extends InnoDB data files.
## PR quality check

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -46,9 +46,6 @@ Created 10/21/1995 Heikki Tuuri
 #include "srv0start.h"
 #include "fil0fil.h"
 #include "fsp0fsp.h"
-#ifdef HAVE_LINUX_UNISTD_H
-#include "unistd.h"
-#endif
 #include "os0event.h"
 #include "os0thread.h"
 
@@ -74,10 +71,12 @@ Created 10/21/1995 Heikki Tuuri
 #ifdef _WIN32
 #include <winioctl.h>
 #else
+#include <unistd.h>
 // my_test_if_atomic_write()
 #include <my_sys.h>
 #endif
 
+#include "log.h"
 #include "buf0dblwr.h"
 
 #include <thread>
@@ -1505,16 +1504,9 @@ bool os_file_close_func(os_file_t file)
   return false;
 }
 
-/** Gets a file size.
-@param[in]	file		handle to an open file
-@return file size, or (os_offset_t) -1 on failure */
-os_offset_t
-os_file_get_size(os_file_t file)
+os_offset_t os_file_get_size(os_file_t file) noexcept
 {
-  struct stat statbuf;
-  if (fstat(file, &statbuf)) return os_offset_t(-1);
-  MSAN_STAT_WORKAROUND(&statbuf);
-  return statbuf.st_size;
+  return lseek(file, 0, SEEK_END);
 }
 
 /** Gets a file size.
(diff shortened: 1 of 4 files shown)


storage/innobase/os/os0file.cc
```diff
@@ -2712,31 +2799,16 @@ bool os_file_set_sparse_win32(os_file_t file, bool is_sparse)
...
-os_file_change_size_win32(
+os_file_set_size(
 	const char*	pathname,
 	os_file_t	file,
 	os_offset_t	size)
 {
 	LARGE_INTEGER	length;
 
 	length.QuadPart = size;
+	ut_ad(!(size & 4095));
```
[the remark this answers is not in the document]
dr-m: I added the assertion because there was one in the common `os_file_set_size()` function. I agree that we can do without this one; d09426f9e60fd93296464ec9eb5f9d85566437d3 claims that the issue does not affect Microsoft Windows.

storage/innobase/os/os0file.cc
```diff
@@ -4180,7 +4109,7 @@ bool fil_node_t::read_page0()
 {
 	ut_ad(mutex_own(&fil_system.mutex));
 	const unsigned psize = space->physical_size();
-#ifndef _WIN32
+#ifdef __linux__
```
vaintroub, line 4112: This "linux-only" optimization does not make sense, in terms of performance (I'd like to see benchmark, where it helped), and uses unreliable fstat call (subject of this patch). Can this be removed?

dr-m: The bug that we are trying to address is microsoft/WSL#8443, to my understanding in a Microsoft specific DrvFs driver. You successfully tested a version of the patch where we would invoke the `fstat()` not only on Linux, but on any non-Windows platform. With the above change, we’d invoke `os_file_get_size()` instead of `fstat()` everywhere else but Linux.

With your suggestion, there would be an extra `lseek` system call before the `fstat`, which we will need anyway. This function `fil_node_t::read_page0()` should only be invoked when InnoDB is accessing a tablespace for the very first time, and it needs to determine things like the page size and the file size and whether the file is located on SSD (to disable `innodb_flush_neighbors`, which made sense on HDD storage).

The DrvFs bug scenario that we are trying to prevent is that a file handle was opened with a name, then the file was renamed, and after that, `fstat(2)` would be invoked on the previously invoked file handle. The function `fil_node_t::read_page0()` would typically invoked right after the file was opened for the first time. The caller would set `fil_space_t::size` and therefore prevent any further calls to `read_page0()` for that file. The `fil_space_t` and `fil_node_t` objects would remain in memory until the file is deleted or the database server is shut down. In InnoDB, any `RENAME`, `ALTER`, `OPTIMIZE` or `TRUNCATE` operation that would involve renaming files must be preceded by a call to `fil_node_t::read_page0()`, so that we know that the table is accessible to begin with. There should not be any issue if `fstat` was invoked *before* `rename`. The issue was when `fstat` was invoked *after* `rename`.

vaintroub: We do not know much about the nature of this bug https://github.com/microsoft/WSL/issues/8443,

My own testing is limited (I remain a reviewer, not tester in this case), mtr does not run on WSL on Windows partitions, claiming something about "world-writable file my.cnf", so I had to manually try single scenario, and it happened to work, but I can't tell whether it is working OK generally. 

What we know is that MySQL does not use fstat, and that it is working ok. We know that fstat failed in the scenario this patch is aimed to fix.  I do not claim to predict where drvfs will fail. 

We also know, and I measured, that lseek is probably the fastest syscall, and its latency does not count (10 million lseeks in 0.4 sec I measured,  makes 400ns per file which, called once per instance lifetime per file).

lseek is not known to fail. fstat is known to fail. fstat in find_metadata can be completely avoided, if there people are not interested in page_compression (they usually are not), or set innodb_flush_neighbors=0 themselves, rather than relying on helpful innodb introspection. With that , latency of lseek is only 25% of fstat, and if file opening is a performance problem, that Linux-specific handling was aimed to fix, it is best fixed by not doing fstat at all.

dr-m: I see that `fil_node_t::find_metadata()` was already tolerating failures of the `fstat` system call. That is a valid argument for unifying the interface and invoking an extra `lseek` system call. Both `read_page0()` nor `find_metadata()` should be invoked very rarely, when opening or creating a file for the first time.

vaintroub (approved):
Looks good, thank you!

PR #3592, opened 2024-10-23
MDEV-35130 - Assertion fails in trx_t::check_bulk_buffer upon CREATE.. SELECT with vector key

svoj:
- [x] *The Jira issue number for this PR is: MDEV-35130*

## Description
Similarly to 4b814066dcd, don't enable bulk insert when issuing create ...
insert into a table containing vector index. InnoDB can't handle situation
when bulk insert is enabled for one table but disabled for another. We can't
do bulk insert on vector index as it does table updates currently.
## Release Notes
N/A
## How can this PR be tested?
mtr vector_innodb

## PR quality check

### DIFF
--- a/mysql-test/main/vector_innodb.result
+++ b/mysql-test/main/vector_innodb.result
@@ -179,3 +179,8 @@ create table t (v vector(1), key(v)) engine=innodb;
 select v from t where v = vec_fromtext('[1]');
 v
 drop table t;
+#
+# MDEV-35130 Assertion fails in trx_t::check_bulk_buffer upon CREATE.. SELECT with vector key
+#
+CREATE TABLE t (pk INT PRIMARY KEY, v VECTOR(5) NOT NULL, VECTOR(v)) ENGINE=InnoDB SELECT 1 AS pk, x'f09baa3ea172763f123def3e0c7fe53e288bf33e' AS v;
+DROP TABLE t;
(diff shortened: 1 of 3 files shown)


vuvova:
merged, but I removed a couple of checks. If bulk insert is enabled under `!table->s->long_unique_table` but disabled without this check, then I suppose it should be safe to add `!table->s->hlindexes()` only when long_unique check is done, that is, no need to do it when disabling

vuvova:
and a minor nitpit, don't refer to commit hashes in comments yet. the tree will still be rebased more than onces, all hashes becomes invalid quickly

svoj:
@vuvova, unsure re `long_unique_table`, either I miss something or 65b3c89430c38fec0d5adb78755a80b3e6600dce is incomplete?
You're very right re commit hashes.

vuvova:
I mean, the code is
```c
   if (!table->s->long_unique_table)
       table->file->start_bulk_insert();
```
and later
```c
   table->file->end_bulk_insert();
```
that is, `end` is not protected. I conclude, there is no need to protect it, and it's allowed to invoke `end_bulk_insert` without `start`. Allowed, because that's what the code does now.

So, it's enough to add `!table->s->hlindexes()` only where `!table->s->long_unique_table` is checked.

svoj:
Uhm, is that a good convention? And then why other callers of `end_bulk_insert()` do care?
Anyway, I don't mind if it works and you feel like it is alright.

vuvova:
Good or not, I think it's confusing and inconsistent to check both `!table->s->long_unique_table` and `!table->s->hlindexes()`  when starting bulk insert, but only the latter — when ending it. I don't see how that could ever be correct, we either need to check all conditions everywhere, or all conditions only at start time. It seems that only at start time is sufficient. If not — it's a bug and we'll need to have all checks (not just `hlindexes`) at the end time too.

PR #3595, opened 2024-10-24
MDEV-35062 Assertion failed in Binary_string::q_append

DaveGosselin-MariaDB:
Reserve sufficient space in the result for q_append'ed data, as q_append does not itself reserve space like it's append counterpart.


grooverdan (approved):
`reserve` maps to following two `q_append` statements. Nice simple fix. Approved.

PR #3596, opened 2024-10-25
MDEV-35257 Backup fails during an ALTER TABLE with FULLTEXT INDEX

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35257*
-->
## Description
In commit 1c55b845e0fe337e647ba230288ed13e966cb7c7 (MDEV-32932) the test `mariabackup.innodb_ddl_on_intermediate_table` was introduced but disabled.

`xb_load_single_table_tablespace()`: Properly handle missing `FTS_` tables.
## Release Notes
`mariadb-backup --backup` could fail if an `ALTER TABLE` or `OPTIMIZE TABLE` statement is concurrently being executed on an InnoDB table that contains `FULLTEXT INDEX`.
## How can this PR be tested?
```sh
./mtr mariabackup.innodb_ddl_on_intermediate_table
```
## PR quality check

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -1185,27 +1185,25 @@ static void backup_file_op_fail(uint32_t space_id, int type,
 	const byte* new_name, ulint new_len)
 {
 	bool fail = false;
-	switch(type) {
+	const static std::string spacename{filename_to_spacename(name, len)};
+	switch (type) {
 	case FILE_CREATE:
 		msg("DDL tracking : create %u \"%.*s\"", space_id, int(len), name);
-		fail = !check_if_skip_table(
-				filename_to_spacename(name, len).c_str());
+		fail = !check_if_skip_table(spacename.c_str());
 		break;
 	case FILE_MODIFY:
 		break;
 	case FILE_RENAME:
 		msg("DDL tracking : rename %u \"%.*s\",\"%.*s\"",
 			space_id, int(len), name, int(new_len), new_name);
-		fail = !check_if_skip_table(
-				filename_to_spacename(name, len).c_str())
+		fail = !check_if_skip_table(spacename.c_str())
 		       || !check_if_skip_table(
 				filename_to_spacename(new_name, new_len).c_str());
 		break;
 	case FILE_DELETE:
-		fail = !check_if_skip_table(
-                                filename_to_spacename(name, len).c_str())
-                       && !check_if_fts_table(reinterpret_cast<const char*>(name));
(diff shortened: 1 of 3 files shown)


extra/mariabackup/xtrabackup.cc
```diff
@@ -3892,16 +3892,18 @@ static void xb_load_single_table_tablespace(const char *dirname,
...
+			goto cant_open;
 		}
-		file = rf;
 	} else {
 		file = new Datafile();
 		file->make_filepath(".", n, IBD);
 	}
 
 	if (file->open_read_only(true) != DB_SUCCESS) {
+	cant_open:
```
dr-m, line 3905: Apparently something around here is broken, because the test would fail in MemorySanitizer as well as on Microsoft Windows. I’ll debug that next week.

dr-m: I am unable to reproduce the failure with clang 19. [Luckily the test fails also on Microsoft Windows](https://buildbot.mariadb.org/#/builders/234/builds/33862/steps/12/logs/stdio), and we actually get a stack trace from the `mariadb-backup --backup` process:
```
mariabackup!ut_dbg_assertion_failed
mariabackup!backup_file_op_fail
mariabackup!recv_sys_t::parse<recv_buf,1>
mariabackup!recv_sys_t::parse_mtr<1>
mariabackup!xtrabackup_copy_logfile
mariabackup!log_copying_thread
```
The relevant part of `backup_file_op_fail()` is the following:
```c++
	if (fail) {
		ut_a(opt_no_lock);
		die("DDL operation detected in the late phase of backup."
			"Backup is inconsistent. Remove --no-lock option to fix.");
	}
```
I see that `check_if_skip_table()` invokes `check_if_fts_table()` for `FILE_DELETE`, which the `type` should be here, unless I am misreading the output. I do not see the file name, but the length would match `./test/FTS_000000000000002c_DELETED_CACHE.ibd`, for example. An obvious problem would seem to be the following:
```c++
	case FILE_DELETE:
		fail = !check_if_skip_table(
                                filename_to_spacename(name, len).c_str())
                       && !check_if_fts_table(reinterpret_cast<const char*>(name));
                msg("DDL tracking : delete %u \"%.*s\"", space_id, int(len), name);
		break;
```
The function `check_if_fts_table()` expects a NUL terminated string. Here, the string should be followed by the end-of-mini-transaction indicator `\x00` or `\x01`, depending on the value of the sequence bit. 63a7e4c96b6ef0a3a37a711328644056c32bb68d should address this.

vlad-lesin (approved):

PR #3598, opened 2024-10-27
MDEV-35237  Bulk insert fails to apply buffered operation during CREATE..SELECT statement

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35237*

## Description

Problem:
=======
- InnoDB fails to write the buffered insert operation during
create..select operation. This happens when bulk_insert
in transaction is reset to false while unlocking a source table.

Fix:
===
- InnoDB should apply the previous buffered changes to
all tables if we encounter any statement other than
pure INSERT or INSERT..SELECT statement in
ha_innobase::external_lock() and start_stmt().

- Remove the function bulk_insert_apply_for_table()

start_stmt(), external_lock(): Assert that trx->duplicates
should be enabled during bulk insert operation

## How can this PR be tested?
./mtr innodb.alter_copy_bulk

## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -15902,20 +15902,17 @@ ha_innobase::start_stmt(
 		if (trx->is_bulk_insert()) {
 			/* Allow a subsequent INSERT into an empty table
 			if !unique_checks && !foreign_key_checks. */
+			ut_ad(!trx->duplicates);
 			break;
 		}
 		/* fall through */
 	default:
-		trx->bulk_insert_apply_for_table(m_prebuilt->table);
+		trx->bulk_insert_apply();
+		trx->end_bulk_insert();
 		if (!trx->bulk_insert) {
 			break;
 		}
 
-		/* Trigger could've initiated another stmt.
-		So apply all bulk operation and mark as
-		end bulk insert for all tables */
-		trx->bulk_insert_apply();
-		trx->end_bulk_insert();
 		trx->bulk_insert = false;
 		trx->last_sql_stat_start.least_undo_no = trx->undo_no;
 	}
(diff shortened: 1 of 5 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -16091,6 +16091,8 @@ ha_innobase::external_lock(
 
 	reset_template();
 	switch (sql_command) {
+	case SQLCOM_CREATE_TABLE:
+		break;
```
dr-m, line 16095: A safer alternative might be to modify the `default:` case so that we will apply and buffered bulk insert to *all* of `trx->mod_tables` (and not only to `m_prebuilt->table`) when we observe the next non-INSERT statement. However, if we do that, I think that we should double check what would happen on a `CREATE TABLE…SELECT` on a partitioned table. (I checked from the grammar that it should be supported.) If that would lead to trouble, then this `case SQLCOM_CREATE_TABLE:` can be preserved, with a comment that explains why we need it.

dr-m: The intention is to finish any previously buffered bulk insert operation at the start of anything else than a plain `INSERT` statement. This had better be done also on `INSERT…ON DUPLICATE KEY UPDATE` and on `REPLACE` statements, which I assume might be overlooked in the existing check.

dr-m (approved):
The commit message is a little misleading, because mentions "copy alter" while the buggy operation was `CREATE TABLE…SELECT`. By definition, the source tables of that operation must be different from the being-created table.

The point of this fix is that when we encounter the start of any other statement than a pure `INSERT` or `INSERT…SELECT`, we will apply any previously buffered changes to the tables, so that for example an `UPDATE` or `REPLACE` will use the regular row-level-logged-and-locked code path.

PR #3600, opened 2024-10-29
[MDEV-34614] mysqlbinlog warn on EOF before GTID in --stop-position

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-34614](https://jira.mariadb.org/browse/MDEV-34614)*
  * Resolves https://github.com/MariaDB/server/pull/3400#discussion_r1681030120
* Follow-up of [MDEV-27037](https://jira.mariadb.org/browse/MDEV-27037)
  * Follow-up of #3400

## Description

This PR adds `mariadb-binlog` warnings for `--stop-position` GTIDs that were not reached at EOF, mainly as a follow-up to MDEV-27037 which added warnings for unreached `--stop-position` file indices and `--stop-datetime`s.

`--stop-position` warnings inform possible mistakes in the input; they’re helpful for both DBAs on CLI and scripts/wrappers that check/log the output status.

MDEV-34614 enchances MDEV-27037 with per-domain GTID validation.
MDEV-27037 was considered a bug fix and so built upon the oldest supported branch (at the time) for forwardporting, but GTID range selection wasn’t a feature of its ancient base version.

<details><summary>

Unlike #3400, the code behind this feature is in the `rpl_gtid.cc` class `Binlog_gtid_state_validator` rather than directly in the `mysqlbinlog.cc` program itself via its filter inputs.
There are two ~~limitations~~ _current situations_ that lead to this decision:

</summary>

* `Binlog_gtid_state_validator` is already prompting unreached *`--start-position`s* in `--gtid-strict-mode` (errors) or `-vvv` mode (warnings).
  This may rather be a side effect. (not certain; I’m new 👶)
* Both MDEV-34614 and MDEV-27037 are considered bug fixes that will apply on older supported versions. Such tickets should have minimal changes and no refactorings.
  The GITD range filter doesn’t have an API for enumerating statuses of individual `--stop-position` GTIDs, whereas modifying `Binlog_gtid_state_validator` only requires one function to gain some optional parameters.

A “new feature” PR with unrestrained rewrites would have a more ideal Separation of Concerns, such as:
* Filters preselect which entries (row or transaction depending on the format) are processed.
* The GTID validator doesn’t care about ranges; it only validates entries (e.g., confirming GITD order) within the range given by the filters.
* The program uses the GTID filters – not the GTID validator – to warn out-of-range GTIDs to match the logic and error message of position & time ([MDEV-27037](https://jira.mariadb.org/browse/MDEV-27037)).
* For parity, the program warns out-of-range `--start-position`/`datetime`s regardless of `--gtid-strict-mode`/`-vvv`, not just `--stop-position`/`datetime`s.

</details>

<details><summary>

To enable the `Binlog_gtid_state_validator`-based design, I had to (temporarily?) remove the *optimization* that recycles the validator object at the *first* (encounter order) `--start-position` GTID outside of `--gtid-strict-mode`/`-vvv`.

</summary>

https://github.com/MariaDB/server/blob/0e8fb977b00983d98c4c35e39bc1f36463095938/client/mysqlbinlog.cc#L1130-L1151

Note that despite this, this latter (runtime order) code queries its states again (if it’s not deleted), specifically to confirm the expectation of reaching (start) GTID states.
https://github.com/MariaDB/server/blob/0e8fb977b00983d98c4c35e39bc1f36463095938/client/mysqlbinlog.cc#L3732-L3741

Consequently, outside of `--gtid-strict-mode` and `-vvv`, the only case that enters this line is when EOF is found before __every__ *`--start-position`* GTID (rather than _any_).
(Recall that sequences of each domain have individual `--start-position`s, just that file logs are concurrent.)
One of these two designs has to be erroneous.

</details>

Optional parameters preserve the current (non-)behaviors of (non-)reporting.
Early exits from `--gtid-strict-mode` exceptions also do (should?) not prompt unreached `--start-position`s since these states may be natural for those. (We’re logging unreachability, not reachability.)

## Release Notes

*What does #3400 say for its Release Notes?… Nothing? 🤷*

Here’s a summary that covers both this and the previous PR:
> `mariadb-binlog` now warns about any `--stop-position`s and `--stop-datetime`s that didn’t apply after reading the entire log.

## How can this PR be tested?

The new test `binlog_mysqlbinlog_warn_stop_gtid` covers positive expectations (but not negative testing yet), though QA should run the entire `binlog` suite to also confirm no regressions.

## PR quality check


ParadoxV5:
PR # 60²

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_gtid.test
```diff
@@ -0,0 +1,49 @@
+--echo
+--echo # MDEV-34614 mysqlbinlog warn on EOF before GTID in --stop-position
+--echo
+
+--source include/have_binlog_format_row_or_statement.inc
```
ParadoxV5, line 5: I originally had
```suggestion
--source include/have_log_bin.inc
```
But (at least for me,) the Mixed format would rotate `0-1-2` and latter to `master-bin.000002`.
`mariadb-binlog path/to/master-bin.000002` somehow even alerted
> ERROR: Found out of order GTID. Got 0-1-1 after 0-1-3

sql/rpl_gtid.cc
```diff
@@ -3343,6 +3344,7 @@ static my_bool report_audit_findings(void *entry, void *report_ctx_arg)
     }
 
     /* Report any out of order GTIDs */
+    if (report_ctx->report_out_of_order)
```
ParadoxV5, line 3347: Instead of stashing `late_gtids` until the end, `Binlog_gtid_state_validator::record` could instead warn/error out-of-order GTIDs itself live.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_gtid.test
bnestere: Did your test look different when you saw those errors? I can't reproduce them.

_BTW I didn't know about the suggestion code block feature, that is nice :)_

sql/rpl_gtid.cc
```diff
@@ -3361,13 +3363,26 @@ static my_bool report_audit_findings(void *entry, void *report_ctx_arg)
...
+  {
+    rpl_gtid *stop_gtid= &stop_gtids[i];
+    struct audit_elem *audit_elem= (struct audit_elem *)my_hash_search(
+      &m_audit_elem_domain_lookup,
+      (const uchar *)&(stop_gtid->domain_id), 0
+    );
+    if (!(audit_elem && audit_elem->last_gtid.seq_no >= stop_gtid->seq_no)) // `>` case should not occur
+      Binlog_gtid_state_validator::warn(
+        out,
+        "Did not reach stop position %u-%u-%llu before end of input",
```
bnestere, line 3382: I'd say to add a comment where MDEV-27037 / #3400 added its warnings, to mention that when `--stop-position` is provided a GTID, that a warning is also emitted (and where/how)

bnestere (changes requested):
Hi @ParadoxV5 

Thanks for a well done PR! Please see my notes.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_gtid.test
ParadoxV5: I can’t reproduce by running the test by itself either, but it occurs when running the complete `binlog` suite (`mtr --suite=binlog`).

sql/rpl_gtid.cc
```diff
@@ -3361,13 +3363,26 @@ static my_bool report_audit_findings(void *entry, void *report_ctx_arg)
...
-  report_ctx.is_strict_mode= is_strict_mode;
-  report_ctx.contains_err= FALSE;
+  struct gtid_report_ctx report_ctx= {out, is_strict_mode, report_out_of_order, FALSE};
   my_hash_iterate(&m_audit_elem_domain_lookup, report_audit_findings, &report_ctx);
+  for (size_t i= 0; i < n_stop_gtids; ++i)
+  {
+    rpl_gtid *stop_gtid= &stop_gtids[i];
+    struct audit_elem *audit_elem= (struct audit_elem *)my_hash_search(
+      &m_audit_elem_domain_lookup,
+      (const uchar *)&(stop_gtid->domain_id), 0
```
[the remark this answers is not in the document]
ParadoxV5: Thanks; my background was in C and I’m unfamiliar with C++ trickeries like this.

Note that I ~~copied~~ _based_ this section from
https://github.com/MariaDB/server/blob/0e8fb977b00983d98c4c35e39bc1f36463095938/sql/rpl_gtid.cc#L3188-L3190
which used the C-cast.

sql/rpl_gtid.cc
ParadoxV5: Hello, please review my code comments.

sql/rpl_gtid.cc
```diff
@@ -1,5 +1,5 @@
 /* Copyright (c) 2013, Kristian Nielsen and MariaDB Services Ab.
-   Copyright (c) 2020, 2022, MariaDB Corporation.
+   Copyright (c) 2020, 2024, MariaDB Corporation.
```
ParadoxV5, line 2: Hello seniors, new member question: Who, when, and what conventions are on the copyright years?

By the way, perhaps devs should ditch them altogether: https://hynek.me/til/copyright-years

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_datetime.test
```diff
@@ -23,10 +23,6 @@ flush logs;
 --echo # MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:50' MYSQLD_DATADIR/master-bin.000001 --result-file=ignored_output_file
 --exec $MYSQL_BINLOG --short-form --stop-datetime='2001-09-08 21:46:50' $MYSQLD_DATADIR/master-bin.000001 --result-file=$ignored_output_file 2>&1
 
---echo Case: Stop datetime between records, must not see warning.
```
DaveGosselin-MariaDB, line 26: This only looks like a duplicate superficially, but the behavior is different because there's a binlog entry now from the prior case.  Please restore this case.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_position.test
```diff
@@ -2,25 +2,25 @@
 --echo # MDEV-27037 mysqlbinlog emits a warning when reaching EOF before stop-condition
 --echo
 
---let assert_file= $MYSQLTEST_VARDIR/tmp/warn_pos_test_file.out
```
DaveGosselin-MariaDB, line 5: This renaming is fine, but IMHO not worth the churn.

sql/rpl_gtid.cc
```diff
@@ -3361,13 +3364,28 @@ static my_bool report_audit_findings(void *entry, void *report_ctx_arg)
...
-  report_ctx.out_file= out;
-  report_ctx.is_strict_mode= is_strict_mode;
-  report_ctx.contains_err= FALSE;
+  struct gtid_report_ctx report_ctx= {out, is_strict_mode, report_out_of_order, FALSE};
   my_hash_iterate(&m_audit_elem_domain_lookup, report_audit_findings, &report_ctx);
+  // Because we don't want to add a new struct gtid_report_ctx member just to index the stop GTIDs,
+  // we can't check them in the report_audit_findings iteration and have to look them up one by one.
+  for (size_t i= 0; i < n_stop_gtids; ++i)
+  {
+    rpl_gtid *stop_gtid= &stop_gtids[i];
```
DaveGosselin-MariaDB, line 3377: It would be cleaner if you could move this behind a method like `is_expected_gtid()` and then the condition at 3382 could be:
```
if (is_expected_gtid(stop_gtid[i])) ...
```
In general, I like the code to read as "what is done" and not "how it's done", unless I want to drop lower in the abstraction.

DaveGosselin-MariaDB (changes requested):

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_datetime.test
DaveGosselin-MariaDB: I retract that, it is correct to remove this case, thanks!

sql/rpl_gtid.cc
bnestere: That's probably a question for @vuvova (or maybe @dbart ?)

And FWIW, speaking for myself, I've only actually set a year when creating a new file. I never felt it necessary to update one in an existing file.

mysql-test/suite/binlog/t/binlog_mysqlbinlog_warn_stop_gtid.test
bnestere: Ah, the test just isn't initializing its state properly. You should be able to do
```suggestion
--source include/have_log_bin.inc
RESET MASTER;
```

Note we don't `RESET MASTER` at the end of a test, see the following comment in `rpl_end.inc` for details:
```
# It does not execute RESET MASTER or RESET SLAVE, because that would
# remove binlogs which are possibly useful debug information in case
# the test case later fails with a result mismatch.
```

bnestere (changes requested):
Thanks @ParadoxV5 ! See my latest notes (though I can't help but feel I lead you in a circle earlier, to arrive back where you had already started looking :sweat_smile:)

client/mysqlbinlog.cc
```diff
@@ -3490,6 +3491,8 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
       /*
         Emit a warning in the event that we finished processing input
         before reaching the boundary indicated by --stop-position.
+        Due to design limitations, Binlog_gtid_state_validator::report
```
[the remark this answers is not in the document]
ParadoxV5: > *Though I wonder if that is already what you had planned before I interfered in the fist place 😇*

Yes, it was similar.
I originally considered adding an API to `Domain_gtid_event_filter *position_gtid_filter` that exposes the completion states of each domain-specific subfilter.
I was ultimately convinced that hiding the code in `gtid_state_validator` instead is a lesser change to APIs. After all,
> *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

I’d definitely choose the former, more consistent strategy if

> *This is a new feature or a refactoring, and the PR is based against the main branch.*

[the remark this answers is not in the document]
ParadoxV5: > For more straightforward reading, it'd be better to highlight the inconsistency at the start of the sentence. Also, "due to design limitations" doesn't really provide context into why (which is good to have when going back through time :), so it would be good to expand upon that).

I have my rationals in the commit description. I’ll copy them over.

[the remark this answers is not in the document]
ParadoxV5: Here’s what we could do: Apply these introverted changes for `10.11` branch, but also follow-up with a refactor (see top post) on the `main` branch.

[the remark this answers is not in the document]
bnestere: > I originally considered adding an API to Domain_gtid_event_filter *position_gtid_filter that exposes the completion states of each domain-specific subfilter.

You could also do the warnings in the filters. There is already `has_finished()`, which you could extend with an optional err-mode parameter (with some noop default) , and then call `has_finished(<err_mode_config>)` here (or something like that. One idea could be a function-pointer to an output function (like `report_f` in `report_audit_findings` (just an idea, you could do it how you like, if you agree with the overall direction)). _Though you'll also have to extend the filters a little to support this :)_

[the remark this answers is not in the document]
ParadoxV5: Utilizing `has_finished` requires modifying it for the entire `Gtid_event_filter` family, or at least both the `Domain_gtid_event_filter` and the `Window_gtid_event_filter`s it’s comprised of.
(I don’t know if C++ subclasses can override parent methods with optional params.)

This scale is no less than creating a new method (which is a new feature) outright, which is why I’ve been hesitant to do these in `10.11` but would feel unrestrained on `main`.

Moreover, although the location would be right, the code itself would still be inconsistent with the surrounding #3400 content – this would be `has_finished(/* warnings */ TRUE)` while those were `if (has_finished()) warn("…")`.
Coercing #3400’s logic to be similar to this would somewhat undo them entirely.
In comparison, my original idea looked like:
```c++
for(Window_gtid_event_filter stop_filter: position_gtid_filter.m_stop_filters) {
  if(!stop_filter.has_finished())
    warn("…");
}
```

sql/rpl_gtid.cc
ParadoxV5: I looked around but found that all other instances with similar search and compare has different actions for “domain not found” and “sequence unexpected”.
Therefore, unless we replace all those actions with lambdas (our C++ standard just happens to have those), I don’t find it enough value to add a single-use function, especially that the condition is already readable (“drop lower in the abstraction”) before or after [`@bnestere`’s inversion](#discussion_r1824594249).

client/mysqlbinlog.cc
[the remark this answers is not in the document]
bnestere: Right, I see your point. Although looking deeper through our current API, we actually already can (basically) do what you suggest, with minimal changes (and through the `state_validator`). We can use `Binlog_gtid_state_validator::verify_gtid_state()` and iterate through our stop-position GTIDs to use as the `domain_state_gtid`, and modify the code so if `out` is `NULL`, then we don't print the warnings within the function, but we just return `FALSE`. So pseudo-codically:

```
rpl_gtid *stop_gtids= position_gtid_filter.get_stop_gtids(); // or maybe to cache them at a previous use?
for(rpl_gtid *stop_gtid: stop_gtids) {
  if(!gtid_state_validator.verify_gtid_state(NULL, stop_gtid))
    warn("…", stop_gtid)
}
my_free(stop_gtids);
```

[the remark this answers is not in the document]
bnestere: I think we'd be ok extending the filter API in 10.11, so long as we do it minimally (i.e. I'm not a huge fan of exposing the filters directly as `position_gtid_filter.m_stop_filters`). Perhaps we can add a function to `Domain_gtid_event_filter` of signature `rpl_gtid *Domain_gtid_event_filter::get_stop_gtids_for_active_filters(uint n)` which abstracts the loop and `has_finished()` call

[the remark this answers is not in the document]
ParadoxV5: I have reëvaluated that it’s not worth to implement suboptimally and inconsistentlly (and also tamper the `delete gtid_state_validator` optimization) just to minimalize API differences.

> I think we'd be ok extending the filter API in 10.11, so long as we do it minimally.

I am going to reïmplement this and #3400 with my ideal concept imagined in the top post:
Move the warning feature to within the (entire?) `Gtid_event_filter` family.
This will probably be a new method (rather than repurposing `has_finished`) so it futureproofs for other filters such as `--start-position` and even `--ignore-server-ids` ([MDEV-20119](https://jira.mariadb.org/browse/MDEV-20119)).

If this turns out to be too much change for 10.11, I might rebrand this ticket & PR to a new feature on `main`.
(I won’t refactor the validator, at least for now.)

---

Speaking of – according to `git grep`, this is an unimplemented prototype found only in its header:
https://github.com/MariaDB/server/blob/a37f71bd10fea224b8f198cb42ca0320dffaf8eb/sql/rpl_gtid.h#L666-L674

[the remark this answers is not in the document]
bnestere: Sounds good (though I could never find what you were referring to with):
> with my ideal concept imagined in the top post:

And please feel free to remove `verify_gtid_is_expected()` while you're at it :) (though as a separate cleanup commit would be best)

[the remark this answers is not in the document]
ParadoxV5: > with my ideal concept imagined in the top post:

oh, it’s [collapsed](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections) under “▶ Unlike 3400”

sql/rpl_gtid.cc
```diff
@@ -3607,6 +3634,31 @@ my_bool Id_delegating_gtid_event_filter<T>::has_finished()
...
+
+template <typename T>
+my_bool Id_delegating_gtid_event_filter<T>::verify_completed_state()
+{
+  if (has_finished()) // fast happy path
+    return FALSE;
+  // If a user-defined filters is not deactivated, it may not be complete.
+  my_bool is_any_incomplete= FALSE;
+  my_hash_iterate(&m_filters_by_id_hash,
+                  verify_subfilter_completed_state, &is_any_incomplete);
```
ParadoxV5, line 3658: Instead of iterating the hash ~~(and segfault myself from the hash’s untypedness)~~, I could’ve directly iterated the `Domain_gtid_event_filter::m_stop_filters`.
Iterating `Id_delegating_gtid_event_filter::m_filters_by_id_hash`
* future-proofs for `--start-position` warnings – if we want to.
* is possible in the superclass, so `Server_gtid_event_filter` (used by `--ignore-server-ids`) can inherit the method override.

Any preference?

sql/rpl_gtid.cc
```diff
@@ -4067,3 +4119,17 @@ my_bool Intersecting_gtid_event_filter::has_finished()
   }
   return FALSE;
 }
+
+my_bool Intersecting_gtid_event_filter::verify_completed_state()
+{
+  my_bool is_any_incomplete= FALSE;
+  Gtid_event_filter *subfilter;
+  for (size_t i= 0; i < m_filters.elements; i++)
```
ParadoxV5, line 4127: This method was based on `::has_finished()`.
Besides formatting, a difference is that this uses `size_t` to match the type of `m_filters.elements`.
The former had a `ulong`; it’s a potential bug, but I’m yet to review which earliest (supported) version this fix should apply on.

sql/rpl_gtid.cc
```diff
@@ -3525,6 +3526,32 @@ my_bool Window_gtid_event_filter::has_finished()
...
+  */
+  if (m_has_stop && !m_has_passed)
+  {
+    /*XXX:
+      `mysqlbinlog.cc`'s `static warning()` family is all duplicate
+      code of the Binlog_gtid_state_validator::warn() family, except
+      it specifies stderr and has a `fflush(result_file)` step.
+    */
+    Binlog_gtid_state_validator::warn(stderr,
+      "Did not reach stop position %u-%u-%llu before end of input",
```
ParadoxV5, line 3547: `%u-%u-%llu` is used throughout this file (and likely more); but GTID struct fields are actually fixed-size, whereas these formats varies by platform.
I’ll file this mismatch as a separate subissue of https://github.com/MariaDB/server/pull/3360#pullrequestreview-2421203312 ([MDEV-21978](https://jira.mariadb.org/browse/MDEV-21978)).

ParadoxV5: [MDEV-35451](https://jira.mariadb.org/browse/MDEV-35451)

sql/rpl_gtid.cc
ParadoxV5: Or can I just leave this to [[MDEV-26896] Enable -Wconversion globally](https://jira.mariadb.org/browse/MDEV-26896)?

client/mysqlbinlog.cc
```diff
@@ -3489,7 +3489,7 @@ static Exit_status dump_local_log_entries(PRINT_EVENT_INFO *print_event_info,
```
ParadoxV5, line 3489: Status: I fell for the same trap that causes [MDEV-35528](https://jira.mariadb.org/browse/MDEV-35528). Its PR #3670 is planning to move this section to a more orderly place.

ParadoxV5: Matched.
See #3757 re. consistency with `--stop-position` indices

sql/rpl_gtid.cc
```diff
@@ -4029,3 +4065,18 @@ my_bool Intersecting_gtid_event_filter::has_finished()
...
 }
+
+bool Intersecting_gtid_event_filter::verify_final_state()
+{
+  bool is_any_not_final= false;
+  Gtid_event_filter *subfilter;
+  for (size_t i= 0; i < m_filters.elements; ++i)
+  {
+    subfilter=
+      *reinterpret_cast<Gtid_event_filter **>(dynamic_array_ptr(&m_filters, i));
```
ParadoxV5, line 4076: I usually use
```suggestion
      *static_cast<Gtid_event_filter **>(dynamic_array_ptr(&m_filters, i));
```
, but `dynamic_array_ptr()` appearently returns `uchar*` rather than `void*`.

client/mysqlbinlog.cc
```diff
@@ -3687,11 +3687,23 @@ int main(int argc, char** argv)
...
     warning("Did not reach stop datetime '%s' before end of input",
             stop_datetime_str);
 
+  /*
+    Emit warning(s) (in Gtid_event_filter::verify_completed_state())
+    in the event that we finished processing input
+    before reaching all boundaries indicated by --stop-position GTID(s).
+  */
+  if (position_gtid_filter)
+    position_gtid_filter->verify_final_state();
```
ParadoxV5, line 3705: When implemented,
```suggestion
  if (gtid_event_filter)
    gtid_event_filter->verify_final_state();
```
will additionally warn if we finished processing before
* reaching all `--start-position` GTID(s), and
* touching all `--ignore-domain`/`server-ids`.

sql/rpl_gtid.cc
```diff
@@ -3487,6 +3488,17 @@ my_bool Window_gtid_event_filter::has_finished()
   return m_has_stop ? m_has_passed : FALSE;
 }
 
+bool Window_gtid_event_filter::verify_final_state()
+{
+  bool is_not_final= m_has_stop && !m_has_passed;
+  if (is_not_final)
+    Binlog_gtid_state_validator::warn(stderr,
```
ParadoxV5, line 3495: `mysqlbinlog.cc`'s `static warning()` family is all duplicate code of the `Binlog_gtid_state_validator::warn()` family, except the former specifies `stderr` and has a `fflush(result_file)` step.

ParadoxV5 (commented):
Force-push: updated to 10.11.11 (https://github.com/MariaDB/server/pull/3600/files#r1872203713)

sql/rpl_gtid.cc
ParadoxV5: Mmm, now that I’m thinking about a larger-scale refactoring, I should use `Domain_gtid_event_filter::m_stop_filters` instead to reign in this patch’s scope.

sql/rpl_gtid.cc
```diff
@@ -3417,21 +3418,27 @@ my_bool Window_gtid_event_filter::exclude(rpl_gtid *gtid)
 
     if (!m_has_start && is_gtid_at_or_before(&m_stop, gtid))
     {
+      DBUG_ASSERT(m_has_stop);
```
bnestere: This feels a bit late to assert **after** checking `is_gtid_at_or_before(&m_stop, gtid)`. Perhaps to instead make the check
```
    if (!m_has_start && m_stop && is_gtid_at_or_before(&m_stop, gtid))
```

bnestere (approved):
Thanks @ParadoxV5, this looks good! I left one small note.

sql/rpl_gtid.cc
ParadoxV5: 👍

Reminds me how silly it is that `Window_gtid_event_filter` “can” have neither start nor stop points.

ParadoxV5:
I messed up (and fixed) the conflict resolution 😅…
I’m glad that I merged first instead of directly rebasing, or else the mess would’ve gotten lost.

Wishing good luck to myself that the squash will be less error-prone.

ParadoxV5:
> c41efde47953603f6ba7cd67a2f527049bdecf28 and 36786541e4d6f36b68ef60f41cc2d32e2aeaf7d0 are identical.

Gottem at last

PR #3603, opened 2024-10-31
MDEV-35301: Bump minimum cmake version to 3.12.0

cvicentiu:
Libmariadb introduced a cmake change that required 3.12 or newer.

Since none of the platforms we currently support have anything lower than 3.12 available, we should make this a proper requirement for the Server as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1335 USA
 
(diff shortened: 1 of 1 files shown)


vuvova:
`s390x-sles-12` uses 3.10.3. Seems to be the only builder with pre-3.12.0

grooverdan:
> `s390x-sles-12` uses 3.10.3. Seems to be the only builder with pre-3.12.0

Thankfully now [EOL](https://mariadb.com/kb/en/mariadb-platform-deprecation-policy/) (though wasn't in last release notes)

PR #3605, opened 2024-11-02
MDEV-35109: Semi-sync Replication stalling Primary using wait point=AFTER_SYNC

bnestere:
For a primary configured with wait_point=AFTER_SYNC, if two threads
T1 (binlogging through MYSQL_BIN_LOG::write()) and T2 were
binlogging at the same time, T1 could accidentally wait for its
semi-sync ACK using the binlog coordinates of T2. Prior to
MDEV-33551, this only resulted in delayed transactions, because all
transactions shared the same condition variable for ACK signaling.
However, with the MDEV-33551 changes, each thread has its own
condition variable to signal. So T1 could wait indefinitely when
either:

  1) T1's ACK is received but not T2's when T1 goes into
wait_after_sync(), because the ACK receiver thread has already
notified about the T1 ACK, but T1 was _actually_ waiting on T2's
ACK, and therefore tries to wait (in vain).

  2) T1 goes to wait_after_sync() before any ACKs have arrived. When
T1's ACK comes in, T1 is woken up; however, sees it needs to wait
more (because it was actually waiting on T2's ACK), and goes to wait
again (this time, in vain).

Note that the actual cause of T1 waiting on T2's binlog coordinates
is when MYSQL_BIN_LOG::write() would call
Repl_semisync_master::wait_after_sync(), the binlog offset parameter
was read as the end of MYSQL_BIN_LOG::log_file, which is shared
among transactions. So if T2 had updated the binary log _after_ T1
had released LOCK_log, but not yet invoked wait_after_sync(), it
would use the end of the binary log file as the binlog offset, which
was that of T2 (or any future transaction). 

The fix in this patch ensures consistency between the binary log
coordinates a transaction uses between report_binlog_update() and
wait_after_sync().

The commits in this PR are organized as follows:
 1. A preparatory commit to make the debug_dbug point "simulate_delay_semisync_slave_reply" use debug_sync to coordinate an ACK reply, rather than sleeps.
 2. A regression to show the issue of MDEV-35109
 3. The fix

### DIFF
--- a/mysql-test/include/search_pattern_in_file.inc
+++ b/mysql-test/include/search_pattern_in_file.inc
@@ -18,6 +18,23 @@
 #    Optionally, SEARCH_ABORT can be set to "FOUND" or "NOT FOUND" and this
 #    will abort if the search result doesn't match the requested one.
 #
+#    Optionally, SEARCH_ABORT_IS_SUCCESS can be set to inverse the error logic
+#    when the SEARCH_ABORT condition is met. The specific cases and behaviors
+#    become:
+#
+#      1) In the normal case (SEARCH_ABORT_IS_SUCCESS is NOT set), when the
+#         matching result (FOUND N or NOT FOUND) matches SEARCH_ABORT, exit in
+#         error (via die); otherwise, we assume success and print the matching
+#         result. This is used in the normal search_pattern_in_file.inc case,
+#         as well as wait_for_pattern_in_file.inc when searching that some
+#         pattern exists at all (i.e. a result of NOT FOUND will trigger die,
+#         so the script can continue waiting until the result is found, and
+#         finally print the result).
+#
+#      2) If SEARCH_ABORT_IS_SUCCESS is set, then we want to inverse the logic
+#         from case (1). That is, if the match result is SEARCH_ABORT, this
+#         is the success case, and we die for all other matching results.
+#
 #    Optionally, SEARCH_OUTPUT can be set to control the format of output.
 #    Supported formats:
 #     - (default)  : "FOUND n /pattern/ in FILE " or "NOT FOUND ..."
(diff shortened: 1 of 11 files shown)


knielsen (approved):
Thanks, Brandon!

I didn't spend a lot of time to understand all your test changes in detail. But the fix in the code looks obviously correct.

 \- Kristian.

mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
```diff
@@ -78,19 +88,72 @@ let $status_var= Rpl_semi_sync_master_wait_sessions;
...
+{
+  --connection server_3
+  set debug_sync= "now wait_for io_thd_at_slave_reply";
+}
+
+# We are using wait_for_pattern_in_file to detect the shutdown is being
+# delayed, but to not re-detect a pattern, we note the size of the error log
+# before the shutdown and use it as the SEARCH_RANGE parameter to start the
+# search at.
+#--let $err_log_size_before_shutdown= `SELECT LENGTH(LOAD_FILE('$log_error_file'))`
```
andrelkin, line 106: Experimental block still remains?

andrelkin (commented):
Looks good though I found two spots to ask about.

mysql-test/suite/rpl/t/rpl_semi_sync_shutdown_await_ack.inc
bnestere: Nice find, thanks!

mysql-test/include/search_pattern_in_file.inc
```diff
@@ -82,14 +82,41 @@ perl;
...
     $ENV{SEARCH_FILE} =~ s{^.*?([^/\\]+)$}{$1};
 
+    my $print_func= sub { print($_[0]); };
+    my $die_func= sub { die($_[0]); };
+    my $abort_func;
+    my $match_func;
+
+    # The outcome of matching is dependent on the context that we are called.
+    #
+    # 1) In the normal case (SEARCH_ABORT_IS_SUCCESS is NOT set), when the
```
andrelkin, line 92: I could not locate any example of SEARCH_ABORT_IS_SUCCESS.. Did I overlook?

At any rate this env variable suits for mentioning in the header of this file. Please consider that.

knielsen (approved):

mysql-test/include/search_pattern_in_file.inc
bnestere: It is used in `wait_for_pattern_in_file.inc` [here](https://github.com/MariaDB/server/pull/3605/commits/f46a34e77b80bd48f7b60a13c2ff9fcd3a9262ef#diff-40d389db838c160e05682e40361fda980a38af1e9f8c46b7e6eb4e38c6d222b6R39)

As far as mentioning in the header file, agreed, will add that.

andrelkin (commented):
Looks good.

PR #3606, opened 2024-11-04
MDEV-35115 Inconsistent Replace behaviour when multiple unique index exist

Thirunarayanan:
## Description
- Replace statement fails with duplicate key error when multiple unique key conflict happens. Reason is that Server expects the InnoDB engine to store the confliciting keys in ascending order. But the InnoDB doesn't store the conflicting keys
in ascending order.

Fix:
===
- Enable HA_DUPLICATE_KEY_NOT_IN_ORDER for InnoDB storage engine to indicate that multiple key conflicts reported is not in the ascending order
## How can this PR be tested?
./mtr innodb.innodb-index

## PR quality check


storage/innobase/handler/ha_innodb.cc
```diff
@@ -2580,6 +2580,7 @@ ha_innobase::ha_innobase(
                           | HA_CAN_ONLINE_BACKUPS
 			  | HA_CONCURRENT_OPTIMIZE
 			  |  (srv_force_primary_key ? HA_REQUIRE_PRIMARY_KEY : 0)
+			  | HA_DUPLICATE_KEY_NOT_IN_ORDER
```
dr-m, line 2583: Instead of setting this flag for all InnoDB table handles, could we execute `m_int_table_flags|= HA_DUPLICATE_KEY_NOT_IN_ORDER` in `check_index_consistency()` only when the InnoDB unique indexes are defined in a different order from the `TABLE_SHARE`? It seems that we should also rename that function to `ha_innobase::check_index_consistency(const dict_table_t&) noexcept`.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -5593,8 +5593,8 @@ innobase_build_v_templ(
 @param[in]	table	table object formed from .frm
 @param[in]	ib_table	InnoDB table definition
 @retval	true if not errors were found */
-static bool
-check_index_consistency(const TABLE* table, const dict_table_t* ib_table)
+bool
+ha_innobase::check_index_consistency(const dict_table_t* ib_table) noexcept
 {
```
dr-m, line 5598: There is no parameter `table` anymore.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -5633,6 +5633,20 @@ check_index_consistency(const TABLE* table, const dict_table_t* ib_table)
...
+		if (index->is_unique()) {
+			ulint n_index= 0;
+			for (dict_index_t *idx =
+				UT_LIST_GET_FIRST(ib_table->indexes);
+			     idx; idx= UT_LIST_GET_NEXT(indexes, idx)) {
+
+				if (idx == index && count != n_index) {
+					m_int_table_flags|=
+						HA_DUPLICATE_KEY_NOT_IN_ORDER;
+				}
```
dr-m, line 5646: This is missing a `break;` or `return;`. The check is too strict as well. Any mismatch regarding the positions of non-unique indexes should not matter.

mysql-test/suite/innodb/t/innodb-index.test
```diff
@@ -1202,3 +1202,17 @@ call mtr.add_suppression("Plugin 'InnoDB' init function returned error");
...
+--echo #
+CREATE TABLE t1 (c1 NUMERIC UNSIGNED NOT NULL,
+		 c2 INT3 UNIQUE,
+		 c3 BIT(2) PRIMARY KEY)ENGINE=InnoDB;
+CREATE UNIQUE INDEX i1 ON t1(c1);
+INSERT INTO t1 (c1,c2,c3) VALUES (0,0,b'01');
+INSERT INTO t1 (c1,c2,c3) VALUES (1,1,b'10');
+REPLACE INTO t1 (c1,c2,c3) VALUES (0,1,b'11');
+SELECT * FROM t1;
+DROP TABLE t1;
```
dr-m, line 1218: Before `DROP TABLE`, can we run `ALTER TABLE t1 ALGORITHM=COPY` to force a rebuild, and then execute the same? Can we check some "rows deleted" counter before and after the `REPLACE` to show what is going on.

Also, can we test adding the `UNIQUE INDEX` in the same order, but having non-unique indexes in between? In that case the `REPLACE` must fall back to `UPDATE` rather than `DELETE` and `INSERT`.

mysql-test/suite/innodb/r/innodb-index.result
```diff
@@ -1947,3 +1947,63 @@ Warnings:
...
+HANDLER_DELETE	2
+HANDLER_READ_KEY	2
+HANDLER_READ_RND_NEXT	28
+HANDLER_TMP_WRITE	51
+HANDLER_WRITE	3
+ALTER TABLE t1 ALGORITHM=COPY;
+FLUSH STATUS;
+SELECT * FROM INFORMATION_SCHEMA.SESSION_STATUS WHERE VARIABLE_NAME LIKE 'HANDLER_%' AND VARIABLE_VALUE > 0;
+VARIABLE_NAME	VARIABLE_VALUE
+HANDLER_TMP_WRITE	24
```
dr-m, line 1977: It looks like `HANDLER_TMP_WRITE` is totally irrelevant for this test. I’d suggest to filter it out in the `WHERE` condition. Perhaps we should enumerate the conditions that we are really interested in, namely, `WHERE VARIABLE_NAME IN ('HANDLER_DELETE','HANDLER_WRITE','HANDLER_READ_KEY')`. In that way, even if later branches change something around the irrelevant counters, the results of this tests will not change.

mysql-test/suite/innodb/r/innodb-index.result
```diff
@@ -1947,3 +1947,63 @@ Warnings:
...
+INSERT INTO t1 (c1,c2,c3) VALUES (1,1,b'10');
+FLUSH STATUS;
+SELECT * FROM INFORMATION_SCHEMA.SESSION_STATUS WHERE VARIABLE_NAME LIKE 'HANDLER_%' AND VARIABLE_VALUE > 0;
+VARIABLE_NAME	VARIABLE_VALUE
+HANDLER_TMP_WRITE	24
+REPLACE INTO t1 (c1,c2,c3) VALUES (0,1,b'11');
+SELECT * FROM INFORMATION_SCHEMA.SESSION_STATUS WHERE VARIABLE_NAME LIKE 'HANDLER_%' AND VARIABLE_VALUE > 0;
+VARIABLE_NAME	VARIABLE_VALUE
+HANDLER_COMMIT	1
+HANDLER_DELETE	2
```
dr-m, line 1968: I had expected 0 or 1 `HANDLER_DELETE` for the `REPLACE` statement, instead of 2 or 1 like this test shows. What is the other `HANDLER_DELETE` for? Can you add an `--echo #` comment about it?

mysql-test/suite/innodb/t/innodb-index.test
```diff
@@ -1202,3 +1202,51 @@ call mtr.add_suppression("Plugin 'InnoDB' init function returned error");
...
+--echo #
+--echo #  MDEV-35115 Inconsistent Replace behaviour when multiple
+--echo #               unique index exist
+--echo #
+let $get_handler_status_counts= SELECT * FROM INFORMATION_SCHEMA.SESSION_STATUS WHERE VARIABLE_NAME LIKE 'HANDLER_%' AND VARIABLE_VALUE > 0;
+
+--disable_ps2_protocol
+CREATE TABLE t1 (c1 NUMERIC UNSIGNED NOT NULL,
+		 c2 INT3 UNIQUE,
+		 c3 BIT(2) PRIMARY KEY)ENGINE=InnoDB;
```
dr-m, line 1215: Do we really have to disable `./mtr --ps-protocol` for the entire test, or would it suffice to do that only around the `eval $get_handler_status_counts`? It should affect `SELECT` only.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -5633,8 +5633,21 @@ check_index_consistency(const TABLE* table, const dict_table_t* ib_table)
 			ret = false;
 			goto func_exit;
 		}
-	}
 
+		if (index->is_unique()) {
+			uint32_t idx_num = index->get_num();
```
dr-m, line 5638: Do we really need a new member function for this?
```c++
if (index->is_unique()) {
	ulint i;
	while ((index = UT_LIST_GET_PREV(indexes, index))) i++;
	if (last_unique > i) {
		m_int_table_flags
			|= HA_DUPLICATE_KEY_NOT_IN_ORDER;
	}
	last_unique = i;
}
```
I’d use the same type for both; in your version, we are dealing with mismatching types `ulint` and `uint32_t`.

storage/innobase/handler/ha_innodb.h
```diff
@@ -437,6 +437,12 @@ class ha_innobase final : public handler
 			  const KEY_PART_INFO& old_part,
 			  const KEY_PART_INFO& new_part) const override;
 
+	/** Check consistency between .frm indexes and InnoDB indexes
+	Set HA_DUPLICATE_KEY_NOT_IN_ORDER if multiple unique index
+	are not in the correct order.
+	@param[in] ib_table InnoDB table definition
```
dr-m, line 443: Let’s avoid `[in]` in new comments.

dr-m (commented):
The algorithm now looks correct to me. I have a few comments about the details.

dr-m (approved):

PR #3610, opened 2024-11-05
MDEV-28378: Don't hang trying to peek log event past the end of log

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-28378*

## Description

While applying CTAS log event, we peek the relay log to see if CTAS contains inserted rows or if it's empty.
The peek function didn't check for end-of-file condition when tried to get the next event from the log, and thus it hanged.

The fix includes checking for end-of-file while peeking for log events and considering returned XID_EVENT value as a sign of an empty CTAS.

## How can this PR be tested?

Re-enabled an MTR test which is now passing.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/disabled.def
+++ b/mysql-test/suite/galera/disabled.def
@@ -10,7 +10,6 @@
 #
 ##############################################################################
 
-galera_as_slave_ctas : MDEV-28378 timeout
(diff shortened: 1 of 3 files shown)


sysprg:
Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/6d5fe9ed0d64eac8d20712123046974642e7f99a

PR #3612, opened 2024-11-05
MDEV-35350: Consolidate MTR wait_for_pattern_in_file.inc and SEARCH_WAIT in search_pattern_in_file.inc

bnestere:
In 10.11, the SEARCH_WAIT extension added to search_pattern_in_file.inc duplicates the wait_for_pattern_in_file.inc logic (mostly). It'd be good to consolidate these uses, and the new additions to search_pattern_in_file.inc seem favorable because it puts all search-in-file logic in one place to make it easier to find for future developers, and is more robust. It does need a slight extension though to handle matching a certain number of times.


mysql-test/include/search_pattern_in_file.inc
```diff
@@ -71,54 +59,64 @@ perl;
...
+                }
+              }
+            }
+            close(FILE);
+            $content.= $file_content;
+        }
+        @matches= ($content =~ /$search_pattern/gs);
+        $res=@matches ? "FOUND " . scalar(@matches) : "NOT FOUND";
+
+        if ($ENV{SEARCH_WAIT} and not $res =~ /^$ENV{SEARCH_WAIT}/) {
```
knielsen, line 94: I think there is a bug here. If $res is "FOUND 10" and SEARCH_WAIT is "FOUND 1", this will terminate the wait (but "FOUND 9" will not for example).

More generally, it is not a good idea to interpolate user input directly into a regexp (unless specifically intended as a user regexp search facility), it is prone to problems like this. Probably better to provide a separate SEARCH_COUNT (for example) parameter, which specifies a specific count to wait for or abort on, if given, that is simpler and more robust.

knielsen (changes requested):
One request for a more robust "FOUND n" matching.
Otherwise looks great, thanks Brandon!

mysql-test/include/search_pattern_in_file.inc
vuvova: On the contrary, when I suggested something similar earlier (for `wait_for_pattern_in_file.inc`) I thought that this provides the desirable flexibility. One can search for `/FOUND 1\b/` or for `/FOUND 10\b/` or for `/FOUND [4-9]/` etc.

Normally `/FOUND 1/` will work though. I can invent a rather artificial case where one needs to wait for "1" but keep waiting if he number is "10", but i don't think it'll happen practically. And if it will, one can always add `\b`

knielsen: Sure, this is fine too, this is "specifically intended as a user regexp search facility".

So then just update the file comments to explain that SEARCH_WAIT is a regexp that waits until the search result (which will be "FOUND <n>" or "NOT FOUND") matches that regexp.

And also document that the regexp will be prepended with start-of-string-anchor "^".

knielsen (approved):

PR #3613, opened 2024-11-06
MDEV-35299

dmitryshulga:
- [x] *The Jira issue number for this PR is: MDEV-35299*

## Description
Server build failure on MacOS Monterey with clang version 14.0.0

## How can this PR be tested?

Try to build server on MacOS with XCode 14.2

## PR quality check


plugin/type_inet/sql_type_inet.cc
```diff
@@ -493,8 +493,13 @@ size_t Inet6::to_string(char *dst, size_t dstsize) const
       //
       // If it is not the last field, append closing ':'.
 
-      p += sprintf(p, "%x", ipv6_words[i]);
+      int ret= snprintf(p, dstsize_available, "%x", ipv6_words[i]);
```
grooverdan, line 496: With Vista EOL this could be `inet_ntop` now. Btw good pattern, a recent security conference shows the dangers of `p+= snprintf(p` because it doesn't mean that the full length was written to the buffer.

Is it really safe to return without a SQL warning pushed?

sql/log_event.cc
```diff
@@ -1409,7 +1409,14 @@ code_name(int code)
   case Q_MASTER_DATA_WRITTEN_CODE: return "Q_MASTER_DATA_WRITTEN_CODE";
   case Q_HRNOW: return "Q_HRNOW";
   }
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
grooverdan, line 1414: notably never reached as `CHECK_SPACE`, the only caller of `code_name` always has a valid code. Changing to a snprint would be less verbose.

sql/log_event.h
```diff
@@ -3177,22 +3188,11 @@ inline char *serialize_xid(char *buf, long fmt, long gln, long bln,
     c+= 2;
   }
   c[0]= '\'';
-  sprintf(c+1, ",%lu", fmt);
+  snprintf(c + 1, ser_buf_size - 1, ",%lu", fmt);
```
grooverdan, line 3191: technically `ser_buf_size - 1 - (2 + (gln + bln) * 2 + 4 + 1)`

sql/log_event_client.cc
```diff
@@ -648,7 +655,14 @@ log_event_print_value(IO_CACHE *file, PRINT_EVENT_INFO *print_event_info,
 
       float8get(dbl, ptr);
       char tmp[320];
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
       sprintf(tmp, "%-.20g", dbl); /* strmake doesn't support %-20g */
+#ifdef __clang__
```
grooverdan, line 663: I think compilers could determine if this is safe with a sizeof rather than warning suppression.

Could expand on safe vs not safe in commit message.

sql/log_event.cc
dr-m: I made the exact same comment in [MDEV-35299](https://jira.mariadb.org/browse/MDEV-35299) a couple of days ago. Basically, after a proper fix, `git grep -w sprintf` should not return anything.

storage/innobase/ut/ut0ut.cc
```diff
@@ -104,7 +104,8 @@ Sprintfs a timestamp to a buffer, 13..14 chars plus terminating NUL. */
 void
 ut_sprintf_timestamp(
 /*=================*/
-	char*	buf) /*!< in: buffer where to sprintf */
+	char*	buf, /*!< in: buffer where to sprintf */
+	size_t len)
 {
 #ifdef _WIN32
 	SYSTEMTIME cal_tm;
```
dr-m, line 111: The parameter `len` is undocumented and not used `#ifdef _WIN32`.

dr-m (commented):
I think that all occurrences of `sprintf()` must be replaced. This should be easy to check with `git grep -w sprintf`.

sql/password.c
```diff
@@ -122,7 +122,14 @@ void my_make_scrambled_password_323(char *to, const char *password,
...
   hash_password(hash_res, password, (uint) pass_len);
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
   sprintf(to, "%08lx%08lx", hash_res[0], hash_res[1]);
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
 }
```
dr-m, line 133: At least one caller (`execute_commands()` in `client/mysqladmin.cc`) is invoking this with `to` pointing to a buffer of 64 bytes. Based on the format expression, the length should be 16 bytes. Based on the `ulong` data type, whose size is 64 bits on LP64 systems and 32 bits elsewhere (including the LLP64 Microsoft Windows), the length could be up to 32 bytes. It seems to me that the `ulong` should be replaced with `uint32_t`, because `SCRAMBLE_LENGTH_323` is defined as 8.

In `Item_func_password::val_str_ascii()`, the argument is a pointer to the buffer `tmp_value[SCRAMBLED_PASSWORD_CHAR_LENGTH+1]`. It seems that this is the length that should be passed to `snprintf` here, and should be mentioned in the function documentation as well.

sql/password.c
```diff
@@ -270,7 +277,14 @@ void get_salt_from_password_323(ulong *res, const char *password)
...
 {
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
   sprintf(to,"%08lx%08lx", salt[0], salt[1]);
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
 }
```
dr-m, line 288: This function seems to have been orphaned in 1ff476b415cacc616f68af542b75793d064367ae. Is it really needed?

dr-m:
I was wondering why the `sprintf` in `dict_table_rename_in_cache()` was not replaced. The explanation seems to be that the file `storage/innobase/dict/dict0dict.cc` contains `#include "sql_class.h"`, which in turn contains `#include "violite.h"`, which starting with 60589aeee03949033c66da5c1eae70d4342179fc back in 2012 disabled all deprecation warnings on macOS for other reasons:
```diff
diff a/include/violite.h b/include/violite.h
@@ -111,6 +129,10 @@ int vio_getnameinfo(const struct sockaddr *sa,
 #define DES_set_key_unchecked(k,ks) des_set_key_unchecked((k),*(ks))
 #define DES_ede3_cbc_encrypt(i,o,l,k1,k2,k3,iv,e) des_ede3_cbc_encrypt((i),(o),
(l),*(k1),*(k2),*(k3),(iv),(e))
 #endif
+/* apple deprecated openssl in MacOSX Lion */
+#ifdef __APPLE__
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
 
 #define HEADER_DES_LOCL_H dummy_something
 #define YASSL_MYSQL_COMPATIBLE
```
I think that this change needs to be reverted, so that all use of `sprintf` can be flagged. Alternatively, all uses of `sprintf` need to be replaced, so that `git grep -w sprintf` will not report anything.

I could review a well defined change, such as "replace all use of `sprintf` with `snprintf`", possibly limited to a particular subsystem, such as InnoDB or `mariadb-backup`. If I understood correctly, macOS Monterey has been superseded by 3 releases already; the current one appears to be called Sequoia. The two macOS targets on https://buildbot.mariadb.org may run something newer than macOS Monterey; https://buildbot.mariadb.org/#/builders/708/builds/5331 and https://buildbot.mariadb.org/#/builders/590/builds/10325 for the currently latest commit 689a1905cb171a227feab14864c364712f1d6af5 are displaying a number of warnings about deprecated `sprintf` and identifying the compiler as `AppleClang 15.0.0.15000100`.

plugin/type_inet/sql_type_inet.cc
dmitryshulga: The goal of this task is to modify the server source code to make it compilable. That is, replace sprintf to snprintf at every place where building of core server without extra plugin (e.g. CONNECT engine) fails. Any other modifications is out of scope. If such modifications are absolutely required they should be filed in JIRA as separate tasks.

sql/log_event.cc
dmitryshulga: the latest commit doesn't contain any #pragma and ifdef __clang__

sql/password.c
dmitryshulga: I removed this function in the latest commit

dmitryshulga:
> I was wondering why the `sprintf` in `dict_table_rename_in_cache()` was not replaced. The explanation seems to be that the file `storage/innobase/dict/dict0dict.cc` contains `#include "sql_class.h"`, which in turn contains `#include "violite.h"`, which starting with [60589ae](https://github.com/MariaDB/server/commit/60589aeee03949033c66da5c1eae70d4342179fc) back in 2012 disabled all deprecation warnings on macOS for other reasons:
> 
> ```diff
> diff a/include/violite.h b/include/violite.h
> @@ -111,6 +129,10 @@ int vio_getnameinfo(const struct sockaddr *sa,
>  #define DES_set_key_unchecked(k,ks) des_set_key_unchecked((k),*(ks))
>  #define DES_ede3_cbc_encrypt(i,o,l,k1,k2,k3,iv,e) des_ede3_cbc_encrypt((i),(o),
> (l),*(k1),*(k2),*(k3),(iv),(e))
>  #endif
> +/* apple deprecated openssl in MacOSX Lion */
> +#ifdef __APPLE__
> +#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
> +#endif
>  
>  #define HEADER_DES_LOCL_H dummy_something
>  #define YASSL_MYSQL_COMPATIBLE
> ```
> 
> I think that this change needs to be reverted, so that all use of `sprintf` can be flagged. Alternatively, all uses of `sprintf` need to be replaced, so that `git grep -w sprintf` will not report anything.
> 
> I could review a well defined change, such as "replace all use of `sprintf` with `snprintf`", possibly limited to a particular subsystem, such as InnoDB or `mariadb-backup`. If I understood correctly, macOS Monterey has been superseded by 3 releases already; the current one appears to be called Sequoia. The two macOS targets on https://buildbot.mariadb.org may run something newer than macOS Monterey; https://buildbot.mariadb.org/#/builders/708/builds/5331 and https://buildbot.mariadb.org/#/builders/590/builds/10325 for the currently latest commit [689a190](https://github.com/MariaDB/server/commit/689a1905cb171a227feab14864c364712f1d6af5) are displaying a number of warnings about deprecated `sprintf` and identifying the compiler as `AppleClang 15.0.0.15000100`.

The scope of the current task MDEV-35299 is to modify the source code so that it be compilable with Apple clang version 14.0.0. Changes introducing by this patch makes the source code better since a number of places, where the C function sprintf used, has been decreased. Replacing of all the places where the sprintf function is invoked with snprintf function is not the purpose of this task.

dr-m:
I pushed 615340fa01c0846589ddec0a21b5e69bdbd7e09d so that I could see if it would suppress deprecation warnings for any other functions than `sprintf`. It turns out that it did not: https://buildbot.mariadb.org/#/builders/708/builds/6074/steps/3/logs/warnings__869_ includes 869 warnings, of which 188 are `-Wdeprecated-declarations`, all for the function `sprintf`. The build log indicates that we actually use OpenSSL, but not the one that used to be bundled in macOS, but a third-party package (`/opt/homebrew/lib/libcrypto.dylib`).

I think that it would be best to replace all use of the inherently unsafe `sprintf()` function and not to disable any deprecation warnings.

dr-m:
f8e70a4bc9d2714cd47ee5146f4645d33ea8f1a0 was a semi-related experiment (to see if there are any additional deprecation warnings for macOS in that code). I revised it due to some build failures and pushed to 10.5 as 9a25f2a5bbc084df725538b8b3e9d2019cbacd31.

dr-m:
I see that @DaveGosselin-MariaDB had filed [MDEV-33714](https://jira.mariadb.org/browse/MDEV-33714), whose description aligns better with what I’d like to be done: replace all `sprintf()`. (While doing that, remove any CMake magic to detect the availability of `snprintf`; it was introduced in C99, which we explicitly use.)

PR #3614, opened 2024-11-07
MDEV-35363   Avoid cloning of table statistics while saving the InnoDB table stats

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-32667*

## Description
 - While saving the InnoDB table statistics, InnoDB does clone
 the table object and its statistics to protect the metadata
 in case if table is being dropped. From 10.6 onwards, any
 background task inside InnoDB on the table takes MDL.
 So metadata is protected by MDL of the table. Avoid the
 cloning of the table and its associated index statistics.

## How can this PR be tested?
./mtr --suite=innodb

## PR quality check

### DIFF
--- a/storage/innobase/dict/dict0stats.cc
+++ b/storage/innobase/dict/dict0stats.cc
@@ -550,178 +550,6 @@ dberr_t dict_stats_exec_sql(pars_info_t *pinfo, const char* sql, trx_t *trx)
   return que_eval_sql(pinfo, sql, trx);
 }
 
-/*********************************************************************//**
-Duplicate a table object and its indexes.
-This function creates a dummy dict_table_t object and initializes the
-following table and index members:
-dict_table_t::id (copied)
-dict_table_t::heap (newly created)
-dict_table_t::name (copied)
-dict_table_t::corrupted (copied)
-dict_table_t::indexes<> (newly created)
-dict_table_t::magic_n
-for each entry in dict_table_t::indexes, the following are initialized:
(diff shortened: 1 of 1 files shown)


mysql-test/suite/innodb/r/stat_tables.result
```diff
@@ -100,4 +100,19 @@ SET GLOBAL innodb_stats_persistent=@save_persistent;
...
+BEGIN;
+SELECT COUNT(*)>=0 FROM mysql.innodb_index_stats LOCK IN SHARE MODE;
+COUNT(*)>=0
+1
+INSERT INTO t VALUES(0),(0);
+SELECT sleep(5);
+sleep(5)
+0
+COMMIT;
+DROP TABLE t;
```
dr-m, line 117: I don’t think we want to add this test case. Any `sleep` should ideally be forbidden in the regression test suite. It makes sense to add this test in #3616, but here I think we had better remove the test case to reduce unnecessary delay on our CI systems.

dr-m (changes requested):
Please push #3614 first, and rebase this once that fix has been merged to 10.6.

For clarity, please file this cleanup as a MDEV. It is not actually fixing what the title says, but removing some code that I had forgotten to remove in 45a05fda27dc7058ce8a89f14b1daa56352adf6b.

dr-m (approved):
Thank you, this looks good. I would also suggest removing the test that was added in #3616.

PR #3615, opened 2024-11-07
MDEV-17856 Port binlog_error_action from MySQL

ParadoxV5:
This is a work-in-progress draft pushed for feedback.

---

* [x] *The Jira issue number for this PR is: MDEV-17856*

An amazing description should answer some questions like:
## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

* *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check


sql/log.cc
```diff
@@ -2908,6 +2901,33 @@ Please consider archiving some logs.", next, (MAX_LOG_UNIQUE_FN_EXT - next));
...
+      previous errors so that a new critical error message can be pushed
+      to the client side.
+    */
+    thd->clear_error();
+    static const char *abort_error=
+      "Either disk is full or file system is read only while opening the binlog. "
+      "Aborting the server";
+    //XXX: MySQL needs both error calls. mysql/mysql-server@3b6b4bf8c5
+    my_error(ER_BINLOG_LOGGING_IMPOSSIBLE, MYF(0), abort_error);
+    sql_print_error("Could not use %s for logging (error %d). %s", name, error_id, abort_error);
```
ParadoxV5, line 2922: mysql/mysql-server@3b6b4bf8c5 amended the second error call.

I wonder, what’s the difference between `my_error` and `sql_print_error`, and why did they (if not *us*) need to error twice?

I know that `my_error` receives its format template from `errmsg-utf8.txt` while `sql_print_error` uses a given template. (I know this from `#3360`.) This is why they have different error messages here.

https://github.com/MariaDB/server/blob/ebbbe9d96023ed4f12409abf9c363d4ea9113089/sql/share/errmsg-utf8.txt#L8311-L8317
(Yes, `ER_BINLOG_LOGGING_IMPOSSIBLE` already exists.)

sql/mysqld.h
```diff
@@ -938,6 +938,15 @@ extern bool max_user_connections_checking;
 extern ulong opt_binlog_dbug_fsync_sleep;
 static const int SERVER_UID_SIZE= 29;
 extern char server_uid[SERVER_UID_SIZE+1];
+extern ulong binlog_error_action;
+
```
ParadoxV5, line 942: I referred MySQL’s locations for changes in `sys_vars.cc` and `mysqld.cc`, but `mysqld.h`… I don’t understand how the sections were divided.

sql/sys_vars.cc
```diff
@@ -1570,6 +1570,15 @@ Sys_log_bin_compress_min_len(
   GLOBAL_VAR(opt_bin_log_compress_min_len),
   CMD_LINE(OPT_ARG), VALID_RANGE(10, 1024), DEFAULT(256), BLOCK_SIZE(1));
 
+static Sys_var_enum Sys_binlog_error_action(
+       "binlog_error_action",
+       "When statements cannot be written to the binary log due to a fatal "
+       "error, the server can either ignore the error and let the primary "
+       "continue, or abort",
+       GLOBAL_VAR(binlog_error_action), CMD_LINE(REQUIRED_ARG),
+       (const char *[]){"IGNORE_ERROR", "ABORT_SERVER", NullS},
```
ParadoxV5, line 1579: I felt that it’s excessive to exile the enum list at a different place but still only use it here.

sql/sys_vars.cc
```diff
@@ -1570,6 +1570,15 @@ Sys_log_bin_compress_min_len(
...
+       DEFAULT(IGNORE_ERROR));
+
 static Sys_var_on_access_global<Sys_var_mybool,
```
ParadoxV5, line 1582: What’s a `Sys_var_on_access_global`?

sql/log.cc
```diff
@@ -3048,7 +3069,7 @@ bool MYSQL_LOG::open(
   DBUG_RETURN(0);
 
 err:
-  sql_print_error(fatal_log_error, name, errno);
```
ParadoxV5, line 3051: MySQL had an additional `log_type == LOG_BIN` requirement for this condition:
```c++
  if (log_type == LOG_BIN && binlogging_impossible_mode == ABORT_SERVER)
  {
    …
    my_error(ER_BINLOG_LOGGING_IMPOSSIBLE …
```
Why is it there?

ParadoxV5 (commented):
a draft is a draft

sql/sys_vars.cc
ParadoxV5: Darn. Compount literals is a GCC extension for C++, even though it’s [standardized in C99](https://en.cppreference.com/w/c/language/compound_literal), [our regular-C version](https://github.com/MariaDB/server/blob/main/CODING_STANDARDS.md#language-standards).

sql/mysqld.cc
```diff
@@ -454,6 +454,7 @@ uint opt_binlog_gtid_index_span_min= 65536;
 my_bool opt_master_verify_checksum= 0;
 my_bool opt_slave_sql_verify_checksum= 1;
 const char *binlog_format_names[]= {"MIXED", "STATEMENT", "ROW", NullS};
+ulong binlog_error_action;
```
ParadoxV5, line 457: A passerby colleague questioned why a variable for a single file requires modifying 4 files to add.

I’ll move this like to `log.cc` and then, yes, why is a variable split among variables`.cc`, implementation`.cc` and a bridging `.h` 🤷?

sql/mysqld.h
```diff
@@ -938,6 +938,15 @@ extern bool max_user_connections_checking;
 extern ulong opt_binlog_dbug_fsync_sleep;
 static const int SERVER_UID_SIZE= 29;
 extern char server_uid[SERVER_UID_SIZE+1];
+extern ulong binlog_error_action;
```
ParadoxV5, line 941: An enum variable only has so many valid options, but they’re `unsigned long`s in RAM. The old design left so much unused RAM space 🤷.

mysql-test/suite/binlog/t/binlog_error_action.test
```diff
@@ -0,0 +1,173 @@
...
+## Simulate diskfull during opening of binlog and check for diskfull error
+## behaviour where binlog error action is to continue the server
+## after the error.
+#SET SESSION debug="+d,simulate_file_write_error";
+#--replace_regex /\.[\\\/]master/master/ /Errcode: 28 ".*"\)/Errcode: 28 "No space left on device")/
+## error EE_WRITE with error code number 3 is expected.
+#--error 3
+#flush logs;
+#SET SESSION debug="-d,simulate_file_write_error";
+#--source include/restart_mysqld.inc
```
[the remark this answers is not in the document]
ParadoxV5: [MDEV-20796](https://jira.mariadb.org/browse/MDEV-20796) notes that MariaDB would continuously retry writing every minute until it no longer encounters a full disk error. It reminded me that I saw this retry message in my logs; they confirm this “hanging” behavior.

sql/log.cc
```diff
@@ -6033,7 +6054,7 @@ int MYSQL_BIN_LOG::new_file_impl(bool commit_by_rotate)
        - ...
```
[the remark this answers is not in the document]
ParadoxV5: > - kill mysql (safety);

would be the above commit (change the default to `ABORT_SERVER`), while [MDEV-20796](https://jira.mariadb.org/browse/MDEV-20796) will add a (possibly default) option of periodically retrying.

sql/log.cc
bnestere: Good question, I'm actually not sure.. I imagine we should only need `my_error` , though we should extend `abort_error` with the 

> Could not use %s ..

details.

sql/log.cc
bnestere: It ensures we only apply the error logic to the binary log, and not other logs. That is, the binlog (of type `MYSQL_BIN_LOG`) is derived from `MYSQL_LOG`, and this is the general `MYSQL_LOG::open` function.

sql/log.cc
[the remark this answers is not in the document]
bnestere: I think just having `binlog_error_action` satisfies the `TODO` :) So we can probably re-write the comment to reflect the real behavior (and eventually MDEV-20796)

sql/sys_vars.cc
```diff
@@ -1570,6 +1570,15 @@ Sys_log_bin_compress_min_len(
   GLOBAL_VAR(opt_bin_log_compress_min_len),
   CMD_LINE(OPT_ARG), VALID_RANGE(10, 1024), DEFAULT(256), BLOCK_SIZE(1));
 
+static Sys_var_enum Sys_binlog_error_action(
+       "binlog_error_action",
+       "When statements cannot be written to the binary log due to a fatal "
+       "error, the server can either ignore the error and let the primary "
```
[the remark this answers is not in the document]
bnestere: No this is good! (As of now...) the convention is that outward-facing references (e.g. in messages and SQL tokens) use `primary` and `replica`, but internally in the code-base and tests, we use `master` and `slave` for consistency reasons (e.g. searching is easier).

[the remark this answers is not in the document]
ParadoxV5: Our real reason is that nobody would bother `/slave/replica/`ing every internal reference. 😁

sql/sys_vars.cc
bnestere: Its basically an authentication wrapper to secure the variable s.t. its modification is protected by some role (see `set_var::check()` and for more details on the variable checking process).

mysql-test/suite/binlog/t/binlog_error_action.test
```diff
@@ -0,0 +1,173 @@
...
+# will be turned off and server will continue. 4 different test cases are
+# added to check both the behaviours.
+###############################################################################
+--source include/have_log_bin.inc
+#--source include/have_binlog_format_statement.inc
+--source include/have_debug.inc
+# Don't test this under valgrind, memory leaks will occur
+--source include/not_valgrind.inc
+# Avoid CrashReporter popup on Mac
+--source include/not_crashrep.inc
```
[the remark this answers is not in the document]
bnestere: Have you tested without this line? It might already be (hopefully is!) fixed

mysql-test/suite/binlog/t/binlog_error_action.test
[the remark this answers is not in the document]
bnestere: Well, in general, we don't have to, but if we _can_ conform to `MySQL` options, it is preferred, as it makes migration more straightforward. Another option would be to include MDEV-20796 in the scope of this PR (perhaps a separate commit though) as a third option that allows for consistency between the option names and behaviors.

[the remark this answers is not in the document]
ParadoxV5: > Another option would be to include MDEV-20796 in the scope of this PR

Nay, the option MDEV-20796 wants to make other cases continuously retry as well; it’s a distinct feature.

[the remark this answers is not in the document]
bnestere: Right, but it would allow each option to be consistent to its meaning. That is, we could slightly change `IGNORE_ERROR` behavior so it wouldn't "retry later" on full disk errors, but it would always turn off binlogging and move on. And we wouldn't lose the functionality because it would be nested in the `RETRY` configuration.

sql/log.cc
```diff
@@ -2908,6 +2901,33 @@ Please consider archiving some logs.", next, (MAX_LOG_UNIQUE_FN_EXT - next));
...
+  {
+    THD *thd= current_thd;
+    /*
+      On fatal error when code enters here we should forcefully clear the
+      previous errors so that a new critical error message can be pushed
+      to the client side.
+    */
+    thd->clear_error();
+    static const char *abort_error=
+      "Either disk is full or file system is read only while opening the binlog. "
```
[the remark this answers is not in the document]
ParadoxV5: It’s not immediately obvious to non-technical folk that saving a file also requires opening it, but it’s alright 👌.

mysql-test/suite/binlog/t/binlog_error_action.test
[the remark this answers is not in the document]
ParadoxV5: MySQL `IGNORE_ERROR` seems to actually mean `DISABLE_LOGGING`. 🤨🤔

mysql-test/suite/binlog/t/binlog_error_action.test
```diff
@@ -0,0 +1,173 @@
...
+# continue. This can lead to out of sync slaves and improper backups.
+#
+# Test:
+# =====
+# A new option "binlog_error_action" has been introduced whose values
+# are "IGNORE" or "ABORT". When binlogging becomes impossible if user sets
+# the variable to "ABORT" server will stop if user sets it to "IGNORE" binlog
+# will be turned off and server will continue. 4 different test cases are
+# added to check both the behaviours.
+###############################################################################
```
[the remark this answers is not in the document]
ParadoxV5: The file is filling with copy-pastes. By reformatting, we might as well rewrite the whole test with `--source`s.

sql/sys_vars.cc
```diff
@@ -1569,6 +1569,21 @@ Sys_log_bin_compress_min_len(
...
+static Sys_var_on_access_global<Sys_var_enum,
+  PRIV_SET_SYSTEM_GLOBAL_VAR_BINLOG_ERROR_ACTION
+> Sys_binlog_error_action("binlog_error_action",
+  "When binary logging encounters a fatal error, "
+  "this option determines whether the server "/*ignores this error, "*/
+  /*"retries in 60 secs, */"closes the binlog, or directly aborts. "
+  "IGNORE_ERROR is a MySQL compatibility option that is the same as "
+  "CLOSE_BINLOG except it retries in 60 secs when disk full, and " // i.e. RETRY
+  "UNSET is a MariaDB compatibility option that is the same as IGNORE_ERROR "
+  "except the binlog doesn't close when rotation cannot generate a unique name", // i.e. IGNORE
```
ParadoxV5, line 1583: MySQL thought that this non-behavior was a bug and patched it into their `IGNORE_ERROR`. WDYT, seniors?

ParadoxV5, line 1583: MySQL thought that this non-behavior was a bug and patched it into their `IGNORE_ERROR`. WDYT, seniors?

sql/log.cc
```diff
@@ -2921,6 +2920,40 @@ bool MYSQL_LOG::init_and_set_log_file_name(const char *log_name,
   return FALSE;
 }
 
+/** Shut down logging and complain part 2 of the original fatal_log_error msg */
+void MYSQL_LOG::exec_error_action()
```
ParadoxV5, line 2924: I see we have many types of logs.
I wonder how many of them benefits from a config similar to binary logs’ `binlog_error_action`?

sql/sys_vars.cc
ParadoxV5: Nevermind. We had [MDEV-22451](https://jira.mariadb.org/browse/MDEV-22451) for that, but somehow my changes broke the (incomprehensible?) call flow.

PR #3616, opened 2024-11-07
MDEV-32667   dict_stats_save_index_stat() reads uninitialized index->stats_error_printed

Thirunarayanan:
## Description
Problem:
========
- dict_stats_table_clone_create() does not initialize the flag stats_error_printed in either dict_table_t or dict_index_t. Because dict_stats_save_index_stat() is operating on a copy of a dict_index_t object, it appears that
dict_index_t::stats_error_printed will always be false for actual metadata objects, and uninitialized in
dict_stats_save_index_stat().

Solution:
=========
dict_stats_table_clone_create(): Assign stats_error_printed for table and index while copying the statistics

## How can this PR be tested?
./mtr innodb.table_stats

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/stat_tables.test
+++ b/mysql-test/suite/innodb/t/stat_tables.test
@@ -15,3 +15,16 @@ select * from t1;
 drop table t1;
 rename table mysql.table_stats_save to mysql.table_stats;
 flush tables;
+
+--echo #
+--echo # MDEV-32667 dict_stats_save_index_stat() reads uninitialized
+--echo #              index->stats_error_printed
+--echo #
+call mtr.add_suppression("InnoDB: Cannot save index statistics for table");
+CREATE TABLE t1(a INT) ENGINE=InnoDB STATS_PERSISTENT=1 STATS_AUTO_RECALC=1;
+BEGIN;
+SELECT COUNT(*)>=0 FROM mysql.innodb_index_stats LOCK IN SHARE MODE;
+INSERT INTO t1 VALUES(0),(0);
+SELECT sleep(1);
+COMMIT;
+DROP TABLE t1;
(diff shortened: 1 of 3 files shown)


mysql-test/suite/innodb/t/stat_tables.test
```diff
@@ -15,3 +15,16 @@ select * from t1;
...
+--echo # MDEV-32667 dict_stats_save_index_stat() reads uninitialized
+--echo #              index->stats_error_printed
+--echo #
+call mtr.add_suppression("InnoDB: Cannot save index statistics for table");
+CREATE TABLE t (a INT) ENGINE=InnoDB STATS_PERSISTENT=1 STATS_AUTO_RECALC=1;
+BEGIN;
+SELECT COUNT(*)>=0 FROM mysql.innodb_index_stats LOCK IN SHARE MODE;
+INSERT INTO t VALUES(0),(0);
+SELECT sleep(5);
+COMMIT;
```
dr-m, line 29: I’m not a friend of `sleep` in regression tests, but I don’t think that there is a better way of reproducing this bug in 10.5. With the minimum `innodb_lock_wait_timeout=1`, the complex InnoDB timeout handling mechanism before 10.6 makes the minimum timeout more like 2 seconds, and we need some safety margin as well. Luckily, if the delay is too short for a slow builder, the test will not fail, but it may merely catch the bug, should it reappear.

storage/innobase/dict/dict0stats.cc
```diff
@@ -423,6 +423,8 @@ dict_stats_table_clone_create(
 	UT_LIST_INIT(t->freed_indexes, &dict_index_t::indexes);
 #endif /* BTR_CUR_HASH_ADAPT */
 
+	t->stats_error_printed = table->stats_error_printed;
```
dr-m, line 426: I think that you mentioned that you’re going to remove this function from 10.6, where `dict_stats_save()` should always be protected by a metadata lock and therefore the `table_orig` parameter can just be renamed to `table`.

Let’s push this one quickly to 10.5 and merge to 10.6, so that you can file a 10.6 pull request for this follow-up cleanup ASAP. That pull request should remove also the test that is added here, to avoid wasting CI resources.

dr-m (approved):

PR #3620, opened 2024-11-07
MDEV-35046 SIGSEGV in list_delete in optimized builds when using pseu…

DaveGosselin-MariaDB:
…do_slave_mode

This patch disallows OPTIMIZE TABLE when (1) in pseudo slave mode and (2) after the session has run XA PREPARE, matching the behavior of the system when pseudo slave mode is disabled.

This patch fixes a memory corruption issue where we would tell innodb to delete a table, which it correctly dropped, but then the SQL layer would not remove it from the correct trx_t structure.  Later, when the session closed and we iterated tables on trx_t (specifically mod_tables), we would attempt to access memory associated with the dropped table and crash.

When in pseudo slave mode, we "forget" that we're within an XA transaction during XA PREPARE, and we can't reject statements like OPTIMIZE TABLE. XA PREPARE with pseudo-slave mode enabled resets the XA transaction sim state to one-phase commit by clearing the cached XA transaction state (that is, by setting to NULL the xid_state.xid_cache_element in slave_applier_reset_xa_trans.  So we need a way to remember that we're in pseudo slave mode for a particular transaction and also remember whatever transaction ID is in flight.

With that information, we can lookup the XA transaction ID when dropping a table and, if it exists, use that instead of the XA state information associated between the handler and THD.  Of course, as a fallback, we will still look at the XA state (trx_t) associated between the handler and THD when the XA transaction doesn't exist (because we're not in XA, or it rolled-back, etc).  This fixes the memory corruption by removing the table from the correct trx_t structure when we tell innodb to delete it, keeping consistency.

### DIFF
--- a/include/mysql/plugin.h
+++ b/include/mysql/plugin.h
@@ -733,6 +733,7 @@ void mysql_query_cache_invalidate4(MYSQL_THD thd,
                                    const char *key, unsigned int key_length,
                                    int using_trx);
 
+void *thd_pseudo_slave_and_get_prepared_xid(MYSQL_THD thd);
 
 /**
   Provide a handler data getter to simplify coding
--- a/include/mysql/plugin_audit.h.pp
+++ b/include/mysql/plugin_audit.h.pp
@@ -672,6 +672,7 @@ void thd_get_xid(const THD* thd, MYSQL_XID *xid);
 void mysql_query_cache_invalidate4(THD* thd,
                                    const char *key, unsigned int key_length,
                                    int using_trx);
+void *thd_pseudo_slave_and_get_prepared_xid(THD* thd);
 void *thd_get_ha_data(const THD* thd, const struct handlerton *hton);
 void thd_set_ha_data(THD* thd, const struct handlerton *hton,
                      const void *ha_data);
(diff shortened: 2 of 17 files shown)


andrelkin (changes requested):
@DaveGosselin-MariaDB , to follow up what we discussed yesterday, it's clear now for myself that the problem at hand is the identity of the temporary table owner.

The owner is by def [<db,table_name,connection_id>](https://github.com/MariaDB/server/blob/11.2-mdev-35046-n_ref_count_based/sql/temporary_tables.cc#L921).
In `pseudo_slave_mode` context the connection id should be specified by `THD::pseudo_thread_id`.
Obviously the test does not do that, unlike a legal script produced by `mysqlbinlog`.

Ideally we should consider how to validate the input to the server from such a script.
I am not ready to discuss how it can be implemented the most general way though.
Yet a minimal form of validation seems to exist as checking of `THD::variables::pseudo_thread_id` value in `slave_applier_reset_xa_trans()` with to
1. warn when it's found not being set, and
2. reset (to zero?) anyway.
The p.2 guarantees the regression test's DDL won't find the table.

PR #3621, opened 2024-11-08
MDEV-35088 main.timezone failing -  MEST vs CET time zone difference

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35088*

## Description

Reported in Debian bug #1084293, from the tzdata changelog:

  * Upstream obsoleted the System V names CET, CST6CDT, EET, EST*, HST, MET, MST*, PST8PDT, and WET. They are symlinks now. Move those zones to tzdata-legacy and update /etc/localtime on package update to the new names. Please use Etc/GMT* in case you want to avoid DST changes.

As such the timezone output started to output CET (or CEST) as the current timezone. Due to the way the test was written, its only possible to hit this error when running mtr from a package. The internals of MTR fix the timezone so this will never be hit in a build.

As such, added Europe/Brussels as the MET same timezone hard fixed by timezone.opt file so it will always run.

As replace result to the CET output and treat MET/MEST as the same while its on its way out.

Thanks Santiago Vila for the bug report and Otto for forwarding it.

## Release Notes

## How can this PR be tested?

## PR quality check


ottok:
Note that a fix suggestion and patch was posted in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1084293 as well. I asked the person to contribute directly upstream.

Side note: If not even MariaDB core devs like you find it challenging to fill out this section, should it be removed from the template as it could be challenging for random contributors as well?

image

abarkov (approved):

PR #3627, opened 2024-11-12
MDEV-35190 HASH_SEARCH duplicates effort before HASH_INSERT or HASH_DELETE

dr-m:
This supercedes #3593, which was accidentally closed and can’t be resurrected.

- [x] *The Jira issue number for this PR is: MDEV-35190 and MDEV-35247*
## Description
The `HASH_` macros are unnecessarily obfuscating the logic, so we had better replace them.

`hash_cell_t::remove()`: Implement most of the `HASH_DELETE` logic. This is based on `buf_pool_t::page_hash_table::remove()`.

`xb_filter_hash_free()`: Avoid any hash table lookup; just traverse the hash bucket chains and free each element.

`rm_if_not_found()`: Make use of `find_filter_in_hashtable()`.

`dict_sys_t::acquire_temporary_table()`, `dict_sys_t::find_table()`: Define non-inline to avoid unnecessary code duplication.

`dict_sys_t::add(dict_table_t *table)`, `dict_table_rename_in_cache()`: Look for duplicate while finding the insert position.

`dict_table_change_id_in_cache()`: Merged to the only caller `row_discard_tablespace()`.

`hash_insert()`: Helper function of `dict_sys_t::resize()`.

`fil_space_t::create()`: Look for a duplicate (and crash if found) when searching for the insert position.

`lock_rec_discard()`: Take the hash array cell as a parameter to avoid a duplicated lookup.

`lock_rec_free_all_from_discard_page()`: Remove a parameter.

`ut_fold_ull()`: For `SIZEOF_SIZE_T < 8`, we simulate universal hashing (Carter and Wegman, 1977) by pretending that `SIZE_T_MAX + 1` is a prime. In other words, we implement a Rabin–Karp rolling hash algorithm similar to `java.lang.String.hashCode()`. This is used for representing 64-bit `dict_index_t::id` or `dict_table_t::id` in the native word size.

If `SIZEOF_SIZE_T >= 8`, we just use an identity mapping instead of `ut_fold_ull()`.

`ut_hash_ulint()`: Remove. The exclusive OR before a modulus operation does not serve any useful purpose; it is only obfuscating and wasting some code. mysql/mysql-server@b11a175924194d574238f42068f09b15924ae2f8 included a similar change.
## Release Notes
Some InnoDB hash table operations were simplified to slightly improve performance.
## How can this PR be tested?
This low level code should be well covered by the regression test suite. Some additional performance and stress testing would be helpful.

This depends on #3584 which was merged to 10.6.
## PR quality check

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -2642,25 +2642,19 @@ my_bool regex_list_check_match(
 	return(FALSE);
 }
 
-static
-my_bool
-find_filter_in_hashtable(
-	const char* name,
-	hash_table_t* table,
-	xb_filter_entry_t** result
-)
+static bool find_filter_in_hashtable(const char *name, hash_table_t *table,
+                                     xb_filter_entry_t **result) noexcept
 {
-	xb_filter_entry_t* found = NULL;
-	const ulint fold = my_crc32c(0, name, strlen(name));
-	HASH_SEARCH(name_hash, table, fold,
-		    xb_filter_entry_t*,
-		    found, (void) 0,
-		    !strcmp(found->name, name));
-
-	if (found && result) {
-		*result = found;
-	}
-	return (found != NULL);
+  const ulint fold= my_crc32c(0, name, strlen(name));
+  if (auto found= table->cell_get(fold)->
+      find(&xb_filter_entry_t::name_hash,[name](xb_filter_entry_t *f)
+      { return !strcmp(f->name, name); }))
+    {
+      if (result)
+        *result= found;
+      return true;
+    }
+  return false;
 }
 
 /************************************************************************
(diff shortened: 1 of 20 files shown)


storage/innobase/dict/dict0dict.cc
```diff
@@ -1113,59 +1153,34 @@ void dict_table_t::add_to_cache()
...
+    {
+      if (!t) return true;
+      ut_ad(t->cached);
+      ut_a(strcmp(t->name.m_name, name));
+      return false;
+    });
+  *prev= table;
+  prev= (table->is_temporary() ? temp_id_hash : table_id_hash).
+    cell_get(ut_fold_ull(table->id))->
+    search<dict_table_t>(&dict_table_t::name_hash, [&table](dict_table_t* t)
```
dr-m, line 1174: Sorry, the `name_hash` should of course have been `id_hash` here. I will also see if I can let the compiler deduce the template parameter of `search` or `find`.

extra/mariabackup/xtrabackup.cc
```diff
@@ -4204,11 +4198,11 @@ xb_register_filter_entry(
...
-					db_entry, (void) 0,
-					!strcmp(db_entry->name, dbname));
+			db_entry = databases_hash->cell_get(fold)
+				->find<xb_filter_entry_t>(
+					&xb_filter_entry_t::name_hash,
+					[&dbname](xb_filter_entry_t* f)
+					{ return !strcmp(f->name, dbname); });
 		}
 		if (!db_entry) {
 			db_entry = xb_add_filter(dbname, databases_hash);
```
dr-m, line 4208: This logic is still too complex, and apparently incorrect after this change, because many `--suite=mariabackup` tests would fail. `databases_hash` can never be `nullptr` here; `xb_add_filter()` would have crashed if it were.

I will rewrite this so that we do a `search` followed by a conditional append (in case no entry was found).

mariadb-DebarunBanerjee (approved):
Thanks @dr-m for keeping the hash table abstraction and still keeping the refactoring scope limited. Looks good to me.

PR #3629, opened 2024-11-13
MDEV-36010 - fix myrocks_hotbackup for python3

i18nsite:
see https://peps.python.org/pep-0469/
CleanShot 2024-11-13 at 10 56 21

### DIFF
--- a/storage/rocksdb/myrocks_hotbackup.py
+++ b/storage/rocksdb/myrocks_hotbackup.py
@@ -268,7 +268,7 @@ class RocksDBBackup():
     start_time= time.time()
     last_progress_time= start_time
     progress_size= 0
-    for fname, size in self.target_sst.iteritems():
+    for fname, size in self.target_sst.items():
       self.do_backup_sst(fname, size)
       progress_size= progress_size + size
       elapsed_seconds = time.time() - start_time
@@ -319,11 +319,13 @@ class MySQLUtil:
     if socket:
       dbh = MySQLdb.Connect(user=user,
                             passwd=password,
+                            charset='utf8',
                             unix_socket=socket)
     else:
       dbh = MySQLdb.Connect(user=user,
                             passwd=password,
                             port=port,
+                            charset='utf8',
                             host="127.0.0.1")
     return dbh
 
@@ -645,8 +647,8 @@ def move_back():
     print_move_back_usage()
     sys.exit()
   create_moveback_dir(opts.datadir)
-  create_moveback_dir(opts.rocksdb_datadir)
   create_moveback_dir(opts.rocksdb_waldir)
+  create_moveback_dir(opts.rocksdb_datadir)
 
   os.chdir(opts.backupdir)
   for f in os.listdir(opts.backupdir):


storage/rocksdb/myrocks_hotbackup.py
```diff
@@ -1,4 +1,4 @@
-#!@PYTHON_SHEBANG@
```
grooverdan, line 1: why are you changing this? The change is the default unless overwritten by RPM/Deb builds.

storage/rocksdb/myrocks_hotbackup.py
```diff
@@ -268,7 +271,7 @@ def do_backup_until(self, time_limit):
     start_time= time.time()
     last_progress_time= start_time
     progress_size= 0
-    for fname, size in self.target_sst.iteritems():
+    for fname, size in self.target_sst.items():
```
grooverdan, line 274: ok. This is mentioned in the PEP

storage/rocksdb/myrocks_hotbackup.py
```diff
@@ -26,8 +26,8 @@
 rocksdb_files = ['MANIFEST', 'CURRENT', 'OPTIONS']
 rocksdb_data_suffix = '.sst'
 rocksdb_wal_suffix = '.log'
-exclude_files = ['master.info', 'relay-log.info', 'worker-relay-log.info',
-                 'auto.cnf', 'gaplock.log', 'ibdata', 'ib_logfile', '.trash']
+exclude_files = [b'master.info', b'relay-log.info', b'worker-relay-log.info',
+                 b'auto.cnf', b'gaplock.log', b'ibdata', b'ib_logfile', b'.trash']
```
grooverdan, line 30: Can you describe the significance of this? It looks like a backwards/forwards utf8 coding/decoding.

grooverdan (commented):
Thanks for the contribution.

Can you rebase this to 10.5.

Also first commit should be titled more like the pr - 'fix myrocksdb_hostbackup for python3`. Can you also mention the python version it is now finally failing with.

Can the commit be squashed up too. Micro changes aren't adding value to the commit history.

storage/rocksdb/myrocks_hotbackup.py
i18nsite: In python3, I don't know why,  db and filename is bytes , not str. 

Otherwise, an error will be reported.
After I modified it like this, it can run.

storage/rocksdb/myrocks_hotbackup.py
i18nsite: Because of the problems with bytes and str, I suspect that my modified script cannot be run with python2

storage/rocksdb/myrocks_hotbackup.py
i18nsite: see this https://github.com/oppia/oppia/wiki/Bytes-and-string-handling-in-Python-3

i18nsite: I originally wanted to use mariadb-backup to back up myrocks, but found that it was impossible to back up myrocks without InnoDB.

For details, see:
https://lists.mariadb.org/hyperkitty/list/discuss@lists.mariadb.org/thread/7Z64ZTVLUESRSTU5QIED4OLNOCMT5DON/

I tried to use myrocks_hotbackup to backup and restore, but got stuck again. A .frm file exists for table i18n.uidHost, but that table is not registered in RocksDB

this is frustrating

i18nsite:
CleanShot 2024-11-13 at 15 10 28
fixed , I use charset='utf8' for mysql connection , fix  is return bytes in python3

grooverdan (approved):

storage/rocksdb/myrocks_hotbackup.py
grooverdan: Merged finally. I got distracted on the time I looked back and thinking there was more work to be done.

Sorry I haven't had time to resolve/investigate/debug "RocksDB: Schema mismatch". This is worthy of a bug report on its own.

On using `mariadb-backup`, this does sound like a preferred way to get something maintainable. As it was based on a InnoDB backup exclusive its a major part of its functionality. Its not too impossible to make a code contribution that adds `--skip-innodb` or otherwise excludes innodb components at compile from mariadb-backup.

Thanks for fixing the python3 compatibility.

i18nsite:
@grooverdan can merge this into main branch ？

grooverdan:
It will be, its been merged as far as the bb-11.8-release branch  - 9220be8866a00788d11b08418498a704d04cc2b4 (annotation on url). It will reach `main`, and the large merge is in progress.

PR #3631, opened 2024-11-13
MDEV-35069 - IMPORT TABLESPACE does not work for tables with vector, although allowed

svoj:
- [x] *The Jira issue number for this PR is: MDEV-35069*

## Description
Propagate discard/import tablespace request to hlindexes.

Let FLUSH TABLES ... FOR EXPORT open/lock hlindexes, so that InnoDB prepares hlindexes for export.

Moved reset_hlindexes() to external_lock(F_UNLCK), so that hlindexes are available for export until UNLOCK TABLES.

## Release Notes
Probably update "Copying Transportable Tablespaces" with information about VECTOR indexes. Their tablespaces have to be copied along with parent table.

## How can this PR be tested?
mtr vector_innodb

## PR quality check

### DIFF
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -7667,6 +7667,9 @@ int handler::ha_external_lock(THD *thd, int lock_type)
     }
   }
 
+  if (lock_type == F_UNLCK)
+    (void) table->unlock_hlindexes();
+
   /*
     We cache the table flags if the locking succeeded. Otherwise, we
     keep them as they were when they were fetched in ha_open().
@@ -7741,8 +7744,6 @@ int handler::ha_reset()
     delete lookup_handler;
     lookup_handler= this;
   }
-  if (table->reset_hlindexes())
-    return 1;
   DBUG_RETURN(reset());
 }
 
--- a/mysql-test/main/vector_innodb.result
+++ b/mysql-test/main/vector_innodb.result
@@ -252,4 +252,39 @@ d
 0.00000
 0.00000
 drop table t;
+#
+# MDEV-35069 - IMPORT TABLESPACE does not work for tables with vector,
+#              although allowed
+#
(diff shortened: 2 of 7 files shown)


vuvova:
Where does InnoDB "prepares hlindexes for export" (or tables, FWIW), `REFRESH_FOR_EXPORT` isn't passed down to the engine, as far as I can see.

I don't understand "reset_hlindexes" move either, why hlindexes aren't "available for export until UNLOCK TABLES" now?

svoj:
@vuvova 

> Where does InnoDB "prepares hlindexes for export" (or tables, FWIW), `REFRESH_FOR_EXPORT` isn't passed down to the engine, as far as I can see.

`ha_innobase::external_lock()` / `row_quiesce_table_start()` / `row_quiesce_write_cfg()`

> I don't understand "reset_hlindexes" move either, why hlindexes aren't "available for export until UNLOCK TABLES" now?

InnoDB creates `.cfg` file on `FTFE` and keeps it until `UNLOCK TABLES` (`external_lock(F_UNLCK)`). This file has to be copied along with `.ibd` to destination.

Currently hlindexes are open until `handler::ha_reset()`, that is end of statement. It means `.cfg` will be removed before `FTFE` itself completes.

sql/handler.cc
```diff
@@ -7639,6 +7639,9 @@ int handler::ha_external_lock(THD *thd, int lock_type)
     }
   }
 
+  if (lock_type == F_UNLCK)
+    (void) table->reset_hlindexes();
```
vuvova, line 7643: may be, rename it to `unlock_hlindexes()` ? it's only used here, and you've move it from the reset to the unlock code path.

vuvova (approved):
ok to push, thanks

sql/handler.cc
svoj: Fine with me. I've updated PR and rebased on top of 11.7, but apparently you've already merged it to bb-11.7-release.

PR #3632, opened 2024-11-13
MDEV-35398  Improve shrinking of system tablespace

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35398*

## Description
Problem:
========
From 10.6.13(86767bcc0f121db3ad83a74647a642754a0ce57f) version, InnoDB purge thread does free the TRX_UNDO_CACHED undo segment. Pre-10.6.13 version data directory can contain TRX_UNDO_CACHED undo segment in system tablespace even though it has external undo tablespace.

During slow shutdown, InnoDB collects the segment from tables exist in system tablespace and cached undo segment in the system tablespace as used segment exist in system tablespace. While shrinking the system tablespace, last used extent can be used by undo cached segment. This extent blocks the shrinking of system tablespace in a effective way.

Solution:
========
While freeing the unused segment, InnoDB should free the cached undo segment header page exists in system tablespace and reset the TRX_RSEG_UNDO_SLOTS to 0xff for the rollback segment header page exists in system tablespace. This could improve the shrinking of system tablespace further.

## How can this PR be tested?
./mtr innodb.undo_leak --mem

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/shrink_cached_undo.test
+++ b/mysql-test/suite/innodb/t/shrink_cached_undo.test
@@ -1,0 +1,47 @@
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+--source include/have_debug.inc
+--source include/not_embedded.inc
+--source include/big_test.inc
+--echo #
+--echo # MDEV-35398 Improve shrinking of system tablespace
+--echo #
+# Make pre-10.6.13 version data directory which skips the
+# purging of cached undo log records
+let $restart_parameters=--debug_dbug=d,skip_cached_undo;
+--source include/restart_mysqld.inc
+SET GLOBAL innodb_file_per_table= 0;
+CREATE TABLE t1(f1 INT PRIMARY KEY)ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT PRIMARY KEY)ENGINE=InnoDB;
+
+SET STATEMENT unique_checks=0,foreign_key_checks=0 FOR
+INSERT INTO t1 SELECT seq FROM seq_1_to_24576;
+SET STATEMENT unique_checks=0,foreign_key_checks=0 FOR
+INSERT INTO t2 SELECT seq FROM seq_1_to_24576;
+
+--echo # Insert 34 transaction which has undo cached records
+--disable_query_log
+let $c = 34;
+while ($c)
+{
+  connect (con$c,localhost,root,,);
+  START TRANSACTION;
+  eval INSERT INTO t1 SELECT seq+$c*32768 FROM seq_1_to_3753;
+  dec $c;
+}
+
+connection default;
+let $c = 34;
+while ($c)
+{
+  disconnect con$c;
+  dec $c;
+}
+--enable_query_log
+
+DROP TABLE t2, t1;
+SET GLOBAL innodb_fast_shutdown=0;
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+let $restart_parameters=;
+--source include/restart_mysqld.inc
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
--- a/storage/innobase/fsp/fsp0fsp.cc
+++ b/storage/innobase/fsp/fsp0fsp.cc
@@ -3851,22 +3851,6 @@ static dberr_t fsp_get_sys_used_segment(inode_info *inodes, mtr_t *mtr)
                               TRX_RSEG_FSEG_HEADER))
         err= DB_CORRUPTION;
       block->page.unfix();
-
-      /* Even after slow shutdown, there is a possiblity that
-      cached undo log can exist. So store the segment as used one */
-      for (trx_undo_t *undo= UT_LIST_GET_FIRST(rseg->undo_cached);
-           undo && err == DB_SUCCESS;
-           undo= UT_LIST_GET_NEXT(undo_list, undo))
-      {
-        block= buf_pool.page_fix(page_id_t{0, undo->hdr_page_no}, &err,
-                                 buf_pool_t::FIX_WAIT_READ);
-        if (!block)
-	  return err;
-	if (!inodes->insert_seg(block->page.frame + TRX_UNDO_SEG_HDR +
-                                TRX_UNDO_FSEG_HEADER))
-          err= DB_CORRUPTION;
-        block->page.unfix();
-      }
     }
   }
 
@@ -4009,6 +3993,33 @@ dberr_t fil_space_t::garbage_collect(bool shutdown)
                     "%s %s ", ut_strerr(err), ctx);
     return err;
   }
+
+  /* Reset the undo log segments slots in the rollback segment header
+  which exist in system tablespace. Undo cached segment will be
+  treated as unused file segment. These segments will be freed as a
+  part of inode_info::free_segs  */
+  mtr.start();
+  mtr.x_lock_space(fil_system.sys_space);
+  for (trx_rseg_t &rseg : trx_sys.rseg_array)
+  {
+    if (rseg.space == fil_system.sys_space &&
+        UT_LIST_GET_LEN(rseg.undo_cached))
+    {
+      buf_block_t *block=
+        buf_page_get_gen(page_id_t{0, rseg.page_no}, 0,
+                         RW_X_LATCH, nullptr, BUF_GET, &mtr, &err);
+      if (!block)
+      {
+        mtr.commit();
+        return err;
+      }
+
+      mtr.memset(block, TRX_RSEG_UNDO_SLOTS + TRX_RSEG,
+                 TRX_RSEG_N_SLOTS * TRX_RSEG_SLOT_SIZE, 0xff);
+      rseg.reinit(rseg.page_no);
+    }
+  }
+  mtr.commit();
 
   return unused_inodes.free_segs();
 }
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -497,6 +497,7 @@ loop:
       if (UNIV_UNLIKELY(mach_read_from_4(TRX_RSEG + TRX_RSEG_FORMAT +
                                          rseg_hdr->page.frame)))
         trx_rseg_format_upgrade(rseg_hdr, &mtr);
+      DBUG_EXECUTE_IF("skip_cached_undo", goto skip_purge_free;);
       if (UNIV_LIKELY(undo != nullptr))
       {
         UT_LIST_REMOVE(rseg.undo_cached, undo);
--- a/mysql-test/suite/innodb/r/shrink_cached_undo.result
+++ b/mysql-test/suite/innodb/r/shrink_cached_undo.result
@@ -1,0 +1,23 @@
+#
+# MDEV-35398 Improve shrinking of system tablespace
+#
+# restart: --debug_dbug=d,skip_cached_undo
+SET GLOBAL innodb_file_per_table= 0;
+Warnings:
+Warning	1287	'@@innodb_file_per_table' is deprecated and will be removed in a future release
+CREATE TABLE t1(f1 INT PRIMARY KEY)ENGINE=InnoDB;
+CREATE TABLE t2(f1 INT PRIMARY KEY)ENGINE=InnoDB;
+SET STATEMENT unique_checks=0,foreign_key_checks=0 FOR
+INSERT INTO t1 SELECT seq FROM seq_1_to_24576;
+SET STATEMENT unique_checks=0,foreign_key_checks=0 FOR
+INSERT INTO t2 SELECT seq FROM seq_1_to_24576;
+# Insert 34 transaction which has undo cached records
+DROP TABLE t2, t1;
+SET GLOBAL innodb_fast_shutdown=0;
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
+NAME	FILE_SIZE
+innodb_system	79691776
+# restart
+SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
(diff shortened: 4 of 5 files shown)


storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -4010,6 +3994,32 @@ dberr_t fil_space_t::garbage_collect(bool shutdown)
...
+  mtr.x_lock_space(fil_system.sys_space);
+  for (ulint rseg_id= 0; rseg_id < TRX_SYS_N_RSEGS; rseg_id++)
+  {
+    trx_rseg_t *rseg= &trx_sys.rseg_array[rseg_id];
+    if (rseg->space->id == 0)
+    {
+      dberr_t err= DB_SUCCESS;
+      buf_block_t *block=
+        buf_page_get_gen(page_id_t{0, rseg->page_no}, 0,
+                         RW_X_LATCH, nullptr, BUF_GET, &mtr, &err);
```
dr-m, line 4009: It is not necessary to initialize `err` in the caller. It will be initialized in every code path of `buf_page_get_gen()`.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -4010,6 +3994,32 @@ dberr_t fil_space_t::garbage_collect(bool shutdown)
...
   }
 
+  /* Remove cached undo log segments in the rollback segment
+  header page exist in system tablespace */
+  mtr.start();
+  mtr.x_lock_space(fil_system.sys_space);
+  for (ulint rseg_id= 0; rseg_id < TRX_SYS_N_RSEGS; rseg_id++)
+  {
+    trx_rseg_t *rseg= &trx_sys.rseg_array[rseg_id];
+    if (rseg->space->id == 0)
```
dr-m, line 4004: We don’t need `rseg_id` for anything here. I’d use a C++11 style enumeration:
```c++
  for (trx_rseg_t &rseg : trx_sys.rseg_array)
  {
    if (rseg.space == fil_system.sys_space)
```
We can also avoid dereferencing `trx_rseg_t::space` by comparing it to `fil_system.sys_space`, which we already have accessed in this function.

I was wondering if it is safe to access 128 pages in a single mini-transaction. It seems to be safe, because the `trx_rseg_t::reinit()` call is not accessing any further pages in the buffer pool. The minimum size of a buffer pool should be larger than 128 pages.

But do we actually need to access anything than the first rollback segment? Isn’t this code only being executed when multiple undo tablespaces exist? In that case, shouldn’t the only rollback segment 0 reside in the system tablespace?

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -4010,6 +3994,32 @@ dberr_t fil_space_t::garbage_collect(bool shutdown)
     return err;
   }
 
+  /* Remove cached undo log segments in the rollback segment
+  header page exist in system tablespace */
```
dr-m, line 3998: This is not valid English grammar. At the very least, a word such as "that" or "which" seems to be missing.

storage/innobase/fsp/fsp0fsp.cc
Thirunarayanan: We can shrink the system tablespace even though we don't have external undo tablespace. (innodb.undo_leak test case does it). We're resetting TRX_RSEG_N_SLOTS in all rollback segments. Treat the cached undo segment
as unused one and free it. This is done only after ensuring that no undo logs exist to purge.

mysql-test/suite/innodb/t/undo_leak.test
```diff
@@ -24,3 +24,58 @@ SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPAC
...
+}
+
+let $c = 63;
+while ($c)
+{
+  connection con$c;
+  START TRANSACTION;
+  eval INSERT INTO t1 select seq from seq_1_to_4096;
+  dec $c;
+}
```
dr-m, line 57: These two loops can be combined. `connect` implies a change of the current `connection`.

mysql-test/suite/innodb/t/undo_leak.test
```diff
@@ -24,3 +24,58 @@ SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPAC
...
+  ROLLBACK;
+  dec $c;
+}
+
+let $c = 63;
+while ($c)
+{
+  disconnect con$c;
+  dec $c;
+}
```
dr-m, line 72: The first of these loops can be omitted. But, we will need `wait_until_count_sessions.inc` to ensure that the server processed the `disconnect` and rolled back the transactions.

mysql-test/suite/innodb/t/undo_leak.test
```diff
@@ -24,3 +26,47 @@ SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPAC
...
+  eval INSERT INTO t1 select seq from seq_1_to_4096;
+  dec $c;
+}
+
+let $c = 63;
+while ($c)
+{
+  connection con$c;
+  ROLLBACK;
+  disconnect con$c;
```
dr-m, line 59: `disconnect` will imply rollback. The test will execute a little faster if we just disconnect the connections without sending a statement and waiting for response. The server would be free to roll back al the 63 transactions in whatever order it pleases. The test as it is now is forcing a particular order of rolling back.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -4010,6 +3994,30 @@ dberr_t fil_space_t::garbage_collect(bool shutdown)
...
+        buf_page_get_gen(page_id_t{0, rseg.page_no}, 0,
+                         RW_X_LATCH, nullptr, BUF_GET, &mtr, &err);
+      if (!block)
+      {
+        mtr.commit();
+        return err;
+      }
+
+      mtr.memset(block, TRX_RSEG_UNDO_SLOTS + TRX_RSEG,
+                 TRX_RSEG_N_SLOTS * 4, 0xff);
```
dr-m, line 4015: The 4 needs to be written as `TRX_RSEG_SLOT_SIZE`. I think that this needs a source code comment. Why is it necessary to overwrite this array? If this was linking to some undo pages, how and where will that garbage be collected?
We might also avoid this operation if it is a no-op, but I think that the size of the `MEMSET` record is rather small.

storage/innobase/trx/trx0purge.cc
```diff
@@ -497,6 +497,7 @@ inline dberr_t purge_sys_t::iterator::free_history_rseg(trx_rseg_t &rseg) const
       if (UNIV_UNLIKELY(mach_read_from_4(TRX_RSEG + TRX_RSEG_FORMAT +
                                          rseg_hdr->page.frame)))
         trx_rseg_format_upgrade(rseg_hdr, &mtr);
+      DBUG_EXECUTE_IF("skip_cached_undo", goto skip_purge_free;);
```
dr-m, line 500: This indeed corresponds to reverting 86767bcc0f121db3ad83a74647a642754a0ce57f. Well done.

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -4010,6 +3994,33 @@ dberr_t fil_space_t::garbage_collect(bool shutdown)
     return err;
   }
 
+  /* Reset the undo log segment slots in the rollback segment header
+  which exists in system tablespace. Undo cached segment will be
+  treated as unused file segment. These segment will be freed as a
+  part of inode_info::free_segs  */
```
dr-m, line 4000: Grammar nitpick: "which exist" ("undo log segments lots" is plural) and "These segments".

In the commit message there is a similar grammatical error: "… InnoDB collects the segment from tables exist in system tablespace …"
There can’t be multiple predicates ("collects" and "exist") in the same sentence. We need a "that" or "which" in between: "tables that exist".

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -3851,22 +3851,6 @@ static dberr_t fsp_get_sys_used_segment(inode_info *inodes, mtr_t *mtr)
...
-           undo && err == DB_SUCCESS;
-           undo= UT_LIST_GET_NEXT(undo_list, undo))
-      {
-        block= buf_pool.page_fix(page_id_t{0, undo->hdr_page_no}, &err,
-                                 buf_pool_t::FIX_WAIT_READ);
-        if (!block)
-	  return err;
-	if (!inodes->insert_seg(block->page.frame + TRX_UNDO_SEG_HDR +
-                                TRX_UNDO_FSEG_HEADER))
-          err= DB_CORRUPTION;
```
dr-m, line 3867: I checked that if this code is not removed, the test would fail like this:
```diff
CURRENT_TEST: innodb.undo_leak
--- /home/marko/11.2/mysql-test/suite/innodb/r/undo_leak.result	2024-11-15 06:37:39.933929352 -0800
+++ /home/marko/11.2/mysql-test/suite/innodb/r/undo_leak.reject	2024-11-15 06:52:50.204918531 -0800
@@ -43,4 +43,5 @@
 # restart
 SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
 NAME	FILE_SIZE
-innodb_system	12582912
+Warnings:
+Warning	1012	InnoDB: SELECTing from INFORMATION_SCHEMA.innodb_sys_tablespaces but the InnoDB storage engine is not installed
```

storage/innobase/fsp/fsp0fsp.cc
```diff
@@ -4010,6 +3994,33 @@ dberr_t fil_space_t::garbage_collect(bool shutdown)
...
+        buf_page_get_gen(page_id_t{0, rseg.page_no}, 0,
+                         RW_X_LATCH, nullptr, BUF_GET, &mtr, &err);
+      if (!block)
+      {
+        mtr.commit();
+        return err;
+      }
+
+      mtr.memset(block, TRX_RSEG_UNDO_SLOTS + TRX_RSEG,
+                 TRX_RSEG_N_SLOTS * TRX_RSEG_SLOT_SIZE, 0xff);
```
dr-m, line 4018: I checked that without this, the test would fail like this:
```diff
CURRENT_TEST: innodb.undo_leak
--- /home/marko/11.2/mysql-test/suite/innodb/r/undo_leak.result	2024-11-15 06:37:39.933929352 -0800
+++ /home/marko/11.2/mysql-test/suite/innodb/r/undo_leak.reject	2024-11-15 06:50:26.151920244 -0800
@@ -43,4 +43,4 @@
 # restart
 SELECT NAME, FILE_SIZE FROM information_schema.innodb_sys_tablespaces WHERE SPACE = 0;
 NAME	FILE_SIZE
-innodb_system	12582912
+innodb_system	18874368
```

dr-m (approved):
Great work! I only have a grammar nitpick about a source code comment and the commit message.

mysql-test/suite/innodb/t/shrink_cached_undo.test
```diff
@@ -0,0 +1,43 @@
...
+--echo #
+# Make pre-10.6.13 version data directory which skips the
+# purging of cached undo log records
+let $restart_parameters=--debug_dbug=d,skip_cached_undo;
+--source include/restart_mysqld.inc
+set global innodb_file_per_table= 0;
+create table t1(f1 int not null)engine=innodb;
+create table t2(f2 int not null)engine=innodb;
+INSERT INTO t1 SELECT seq FROM seq_1_to_32768;
+INSERT INTO t2 SELECT seq FROM seq_1_to_32768;
```
dr-m, line 16: Can these be any smaller? Can we prepend each of these with
```sql
```
so that there will be less undo log written? Or is this undo log important for the test?

Moreover, since we are using debug instrumentation anyway, could we use `SET GLOBAL innodb_limit_optimistic_insert_debug` to bloat the tables while inserting much fewer rows?

mysql-test/suite/innodb/t/shrink_cached_undo.test
```diff
@@ -0,0 +1,43 @@
...
+--echo # Insert 63 transaction which has undo cached records
+--disable_query_log
+let $c = 63;
+while ($c)
+{
+  connect (con$c,localhost,root,,);
+  START TRANSACTION;
+  INSERT INTO t1 select seq from seq_1_to_4096;
+  dec $c;
+}
```
dr-m, line 27: Can we use fewer transactions?

dr-m (commented):
The test is taking an excessive amount of time (sometimes under, sometimes over 900 seconds or 15 minutes) on the mandatory Windows debug builder. This is most likely because on Windows, no RAM disk is being used. On one s390x box, I observed the test to consume 6 minutes. That is way too much as well. On x86-64 and ARMv8 the execution time would be more acceptable, but still over 30 seconds. On the MSAN builder, it was 120 seconds, which is a bit too much.

If we can’t shrink the test enough by other means, adding the following would seem to be a good work-around:
```sql
--source include/big_test.inc
```
We’d still have some coverage as long as there is at least one environment that covers debug instrumented executables with `./mtr --big-test`. (This is not a given; we have seldomly had such builders. Currently the MSAN builders are such ones.)

mysql-test/suite/innodb/t/shrink_cached_undo.test
Thirunarayanan: yes, we can use fewer transaction

mysql-test/suite/innodb/t/shrink_cached_undo.test
Thirunarayanan: For some reason, it is not repeating with `innodb_limit_optimistic_insert_debug`. But I added `include/big_test.inc`

dr-m (approved):

eworm-de:
I have seen missing tablespace with MariaDB version `11.4.3` several times now (in a `galera` cluster, if that matters). I *think* it happened on server shutdown, and I guess setting `innodb_fast_shutdown = 1` does work around this - still to be proven.

Now that I see this commit... Are you aware of a situation where tablespace could corrupt, and is this a possible candidate to fix it?

PR #3633, opened 2024-11-14
MDEV-35046 SIGSEGV in list_delete in optimized builds when using pseu…

DaveGosselin-MariaDB:
…do_slave_mode

slave_applier_reset_xa_trans() should clear the THD::pseudo_thread_id when called to reset XA transaction state completely.

### DIFF
--- a/sql/xa.cc
+++ b/sql/xa.cc
@@ -1196,5 +1196,10 @@ static bool slave_applier_reset_xa_trans(THD *thd)
   thd->has_waiter= false;
   MYSQL_COMMIT_TRANSACTION(thd->m_transaction_psi); // TODO/Fixme: commit?
   thd->m_transaction_psi= NULL;
+  if (thd->variables.pseudo_slave_mode && thd->variables.pseudo_thread_id == 0)
+    push_warning(thd, Sql_condition::WARN_LEVEL_WARN,
+		 ER_PSEUDO_THREAD_ID_OVERWRITE,
+		 ER_THD(thd, ER_PSEUDO_THREAD_ID_OVERWRITE));
+  thd->variables.pseudo_thread_id= 0;
   return thd->is_error();
 }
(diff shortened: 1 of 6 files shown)


DaveGosselin-MariaDB:
Relates to #3620 but clears the pseudo thread id instead of trying to sort out the trx_t confusion, which shouldn't happen in this case to begin with.

sql/xa.cc
```diff
@@ -1196,5 +1196,6 @@ static bool slave_applier_reset_xa_trans(THD *thd)
   thd->has_waiter= false;
   MYSQL_COMMIT_TRANSACTION(thd->m_transaction_psi); // TODO/Fixme: commit?
   thd->m_transaction_psi= NULL;
+  thd->variables.pseudo_thread_id= 0;
```
andrelkin, line 1199: I think a warning in case the pseudo-thread-id is not set is worth of giving, like
```
modified   sql/xa.cc
@@ -1196,6 +1196,10 @@ static bool slave_applier_reset_xa_trans(THD *thd)
   thd->has_waiter= false;
   MYSQL_COMMIT_TRANSACTION(thd->m_transaction_psi); // TODO/Fixme: commit?
   thd->m_transaction_psi= NULL;
+  if (thd->variables.pseudo_slave_mode && thd->variables.pseudo_thread_id == 0)
+    push_warning(thd, Sql_condition::WARN_LEVEL_WARN,
+		 ER_INCORRECT_GLOBAL_LOCAL_VAR,
+		 ER_THD(thd, ER_INCORRECT_GLOBAL_LOCAL_VAR));
   thd->variables.pseudo_thread_id= 0;
   return thd->is_error();
 }

```

Never mind `ER_INCORRECT_GLOBAL_LOCAL_VAR`. I picked only for compile, a new one seems necessary.

andrelkin (changes requested):
Looks quite good. One note is made anyway.

Let's also add a detail to
>  should clear the THD::pseudo_thread_id 

to convey the following.
The policy of clearing when `pseudo_slave_mode` is `true` is adopted from binlog applier that handles BASE64-encoded events possibly containing `pseudo_thread_id`  to restore the pre-event's state
of the connection's respective session var.

andrelkin (approved):
Reads good. Thanks Dave!

PR #3635, opened 2024-11-15
MDEV-35413 InnoDB: Cannot load compressed BLOB

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35413*
## Description
A race condition was observed between two `buf_page_get_zip()` for a page. One of them had proceeded to `buf_read_page()`, allocating and x-latching a `buf_block_t` that initially comprises only an uncompressed page frame. While that thread was waiting inside `buf_block_alloc()`, another thread would try to access the same page. Without acquiring a page latch, it would wrongly conclude that there is corruption because no compressed page frame exists for the block.

`buf_page_get_zip()`: Simplify the logic and correct the documentation. Always acquire a shared latch to prevent any race condition with a concurrent read operation. No longer increment a buffer-fix; the latch is sufficient for preventing page relocation or eviction.

`buf_read_page()`: Add the parameter `bool unzip=true`. In `buf_page_get_zip()` there is no need to allocate an uncompressed page frame for reading a compressed BLOB page. We only need that for other `ROW_FORMAT=COMPRESSED` pages, or for writing compressed BLOB pages.

`btr_copy_zblob_prefix()`: Remove the message "Cannot load compressed BLOB" because `buf_page_get_zip()` will already have reported a more specific error whenever it returns `nullptr`.

`row_merge_buf_add()`: Do not crash on BLOB corruption, but return an error instead. (In debug builds, an assertion will fail if this corruption is noticed.)
## Release Notes
`InnoDB: Cannot load compressed BLOB` could be reported due to a race condition between two concurrent accesses for the same record in a `ROW_FORMAT=COMPRESSED` table, which would need to load data into the buffer pool.

The message will be removed as part of this fix.

Furthermore, InnoDB could crash if this scenario occurred during an `ALTER TABLE` operation.
## How can this PR be tested?
Our regression test suite is not suitable for reproducing this type of low-level race conditions in the InnoDB buffer pool. This is best tested by the Random Query Generator, which is what was used in the first place to reproduce this.
## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -6850,14 +6850,9 @@ btr_copy_zblob_prefix(
 		buf_page_t*	bpage;
 		uint32_t	next_page_no;
 
-		/* There is no latch on bpage directly.  Instead,
-		bpage is protected by the B-tree page latch that
-		is being held on the clustered index record, or,
-		in row_merge_copy_blobs(), by an exclusive table lock. */
 		bpage = buf_page_get_zip(id);
 
 		if (UNIV_UNLIKELY(!bpage)) {
-			ib::error() << "Cannot load compressed BLOB " << id;
 			goto func_exit;
 		}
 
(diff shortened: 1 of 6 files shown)


mariadb-DebarunBanerjee (approved):
Looks good.

PR #3636, opened 2024-11-15
MDEV-35394  Innochecksum misinterprets freed undo pages

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35394*

## Description
 Innochecksum misinterprets the freed pages as active one.
This leads the user to think there are too many valid
pages exist.

- To avoid this confusion, innochecksum introduced one
more option --skip-freed-pages and -r to avoid the freed
pages while dumping or printing the summary of the tablespace.

- Innochecksum can safely assume the page is freed if
the respective extent doesn't belong to a segment and marked as
freed in XDES_BITMAP in extent descriptor page.

- Innochecksum shouldn't assume that zero-filled page as extent
descriptor page.

## How can this PR be tested?
./mtr innodb.innochecksum_undo_page

## PR quality check

### DIFF
--- a/extra/innochecksum.cc
+++ b/extra/innochecksum.cc
@@ -75,6 +75,8 @@ static my_bool do_leaf;
 static my_bool per_page_details;
 static ulint n_merge;
 static ulint physical_page_size;  /* Page size in bytes on disk. */
+static ulint extent_size;
+static ulint xdes_size;
 ulong srv_page_size;
 ulong srv_page_size_shift;
 /* Current page number (0 based). */
@@ -99,7 +101,7 @@ char*				log_filename = NULL;
 FILE*				log_file = NULL;
 /* Enabled for log write option. */
 static bool			is_log_enabled = false;
-
+static bool			skip_freed_pages;
 static byte field_ref_zero_buf[UNIV_PAGE_SIZE_MAX];
 const byte *field_ref_zero = field_ref_zero_buf;
 
@@ -267,6 +269,8 @@ static void init_page_size(const byte* buf)
 		srv_page_size_shift = UNIV_ZIP_SIZE_SHIFT_MIN - 1 + ssize;
 		srv_page_size = 512U << ssize;
 		physical_page_size = srv_page_size;
+		extent_size = FSP_EXTENT_SIZE;
+		xdes_size = XDES_SIZE;
 		return;
 	}
 
@@ -278,6 +282,8 @@ static void init_page_size(const byte* buf)
 
 	srv_page_size = fil_space_t::logical_size(flags);
 	physical_page_size = fil_space_t::physical_size(flags);
+	extent_size = FSP_EXTENT_SIZE;
+	xdes_size = XDES_SIZE;
 }
 
 #ifdef _WIN32
@@ -555,8 +561,8 @@ bool
 is_page_doublewritebuffer(
 	const byte*	page)
 {
-	if ((cur_page_num >= FSP_EXTENT_SIZE)
-		&& (cur_page_num < FSP_EXTENT_SIZE * 3)) {
+	if ((cur_page_num >= extent_size)
+		&& (cur_page_num < extent_size * 3)) {
 		/* page is doublewrite buffer. */
 		return (true);
 	}
@@ -757,8 +763,8 @@ static inline bool is_page_free(const byte *xdes, ulint physical_page_size,
 {
   const byte *des=
       xdes + XDES_ARR_OFFSET +
-      XDES_SIZE * ((page_no & (physical_page_size - 1)) / FSP_EXTENT_SIZE);
-  return xdes_is_free(des, page_no % FSP_EXTENT_SIZE);
+      xdes_size * ((page_no & (physical_page_size - 1)) / extent_size);
+  return xdes_is_free(des, page_no % extent_size);
 }
 
 /*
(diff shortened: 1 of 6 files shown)


extra/innochecksum.cc
```diff
@@ -873,6 +873,17 @@ parse_page(
...
+			page_no = mach_read_from_4(page + FIL_PAGE_OFFSET);
+			const byte *des= xdes + XDES_ARR_OFFSET +
+				XDES_SIZE *
+				((page_no & (physical_page_size - 1))
+				 / FSP_EXTENT_SIZE);
+			if (mach_read_from_4(des) != XDES_FSEG &&
+		            xdes_is_free(des, page_no % FSP_EXTENT_SIZE)) {
+				break;
+			}
+		}
```
dr-m, line 886: I don’t think that it is a good idea to make this a special case for a particular page type. In older InnoDB data files, the `FIL_PAGE_TYPE` could be garbage. The only thing that held was that if the page is an index page, then `FIL_PAGE_TYPE` would be written as `FIL_PAGE_INDEX`. The opposite was not necessarily true: any page could carry a `FIL_PAGE_TYPE` of `FIL_PAGE_INDEX`.

It would seem better to implement this logic for any page type, and enabled by a special command line option of `innochecksum`, say, `--skip-freed-pages`. In that way, this feature would also be opt-in.

extra/innochecksum.cc
```diff
@@ -1861,8 +1872,14 @@ int main(
...
 
-			if (page_get_page_no(buf) % physical_page_size == 0) {
+			/* Since we read the page0 already. We can
+			safely ignore if the page number is 0 as
+			extent descriptor page */
+			if (uint32_t page_no = page_get_page_no(buf)) {
+			    if (page_no > 0 &&
+			        page_no % physical_page_size == 0) {
 				memcpy(xdes, buf, physical_page_size);
+			    }
```
dr-m, line 1882: The condition `page_no > 0 &&` is redundant, because the outer `if` already guarantees that (it is nonzero and an unsigned type).

We seem to be trusting the page number that is written to the page frame, instead of consulting `cur_page_num`. If we are not validating `page_get_page_no()` against `cur_page_num` anywhere, it would be good to add another command line option that checks that the page number and tablespace ID within the file are sensible, except when we get a page that is full of NUL bytes.

mysql-test/suite/innodb/t/innochecksum_undo_page.test
```diff
@@ -0,0 +1,47 @@
...
+let $c = 16;
+while ($c)
+{
+  connection con$c;
+  START TRANSACTION;
+  eval INSERT INTO t1 select seq from seq_1_to_128;
+  dec $c;
+}
+
+--disable_query_log
```
dr-m, line 25: The two first `while` loops can be combined: `connect` implies a change of `connection`.

Do we really need this many connections and rows? If yes, then I think that we need `big_test.inc` to avoid overloading our CI system.

extra/innochecksum.cc
```diff
@@ -1288,6 +1303,9 @@ innochecksum_get_one_option(
 	case 'l':
 		is_log_enabled = true;
 		break;
+	case 'r':
+		skip_freed_pages = true;
+		break;
```
dr-m, line 1308: Are these assignments really needed? If yes, what is the purpose of `&skip_freed_pages` in the parameter declaration?

Also, `argument MY_ATTRIBUTE((unused))` could be removed from the function definition.

extra/innochecksum.cc
```diff
@@ -786,6 +788,16 @@ parse_page(
...
 	str = skip_page ? "Double_write_buffer" : "-";
+	page_no = mach_read_from_4(page + FIL_PAGE_OFFSET);
+	if (skip_freed_pages) {
+		const byte *des= xdes + XDES_ARR_OFFSET +
+			XDES_SIZE * ((page_no & (physical_page_size - 1))
+				     / FSP_EXTENT_SIZE);
+		if (mach_read_from_4(des) != XDES_FSEG &&
+		    xdes_is_free(des, page_no % FSP_EXTENT_SIZE)) {
+			return;
+		}
```
dr-m, line 799: `FSP_EXTENT_SIZE` is a rather complex macro that depends on `srv_page_size_shift`. Can we introduce `static size_t extent_size` and assign `extent_size = FSP_EXTENT_SIZE` right after we have determined the page size?

extra/innochecksum.cc
```diff
@@ -100,6 +100,8 @@ FILE*				log_file = NULL;
 /* Enabled for log write option. */
 static bool			is_log_enabled = false;
 
+/* Skip freed pages */
+static bool			skip_freed_pages= false;
```
dr-m, line 104: `static` and global variables are zero-initialized by the loader by default, by the virtue of assigning them to BSS. Modern compilers can optimize away the assignment, but older compilers might generate some code for the redundant zero-initialization.

extra/innochecksum.cc
```diff
@@ -100,6 +101,8 @@ FILE*				log_file = NULL;
 /* Enabled for log write option. */
 static bool			is_log_enabled = false;
 
+/* Skip freed pages */
+static bool			skip_freed_pages;
```
dr-m, line 105: The comment before the variable declaration is not adding any value, and it is not in Doxygen format either.

In my previous review, I quoted the redundant initialization of `is_log_enabled` with the thought that you’d remove all redundant zero-initialization of any global variables in this compilation unit.

I also quoted some code in `innochecksum_get_one_option()` that seemed to be redundant. You addressed that comment by not adding further redundant code there, but you did not remove other redundant code:
```c++
	case 'n':
		no_check = true;
		break;
	case 'a':
	case 'S':
		break;
	case 'w':
		do_write = true;
		break;
	case 'l':
		is_log_enabled = true;
		break;
```
Note: the following is not redundant in the same way, because the parameter that is bound to the `D` option is `&page_dump_filename`. For now, I am not suggesting to remove the variable `page_type_dump`, although I would assume that it is redundant too: `page_type_dump == !!page_dump_filename`.
```c++
	case 'D':
		page_type_dump = true;
		break;
```

Thirunarayanan: My intention is that I don't want to touch the old code. That's the reason, I didn't remove it. Since you insist, I will remove it in the upcoming patch.

dr-m (approved):

dr-m (approved):

PR #3639, opened 2024-11-18
MDEV-34534 main.plugin_load(daemon_example) - AddressSanitizer: Joining already joined thread, aborting

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34534*

## Description

Joining with a thread that has previously been joined results in undefined behavior.

This example plugin performs the same join to the same thread a few lines later. ASAN keeps track of this and fails.

Make the behaviour defined by joining only once.

## Release Notes

Fixed test plugin daemon_example used by main.plugin_load test.

## How can this PR be tested?

Under ASAN (clang-18 or 19+) run the main.plugin_load test

## PR quality check


vaintroub (approved):
Looks good to me

plugin/daemon_example/daemon_example.cc
```diff
@@ -77,14 +77,14 @@ pthread_handler_t mysql_heartbeat(void *p)
...
-    Starts up heartbeatbeat thread
+    Starts up heartbeat thread (mysql_heartbeat)
 
   RETURN VALUE
     0                    success
     1                    failure (cannot happen)
 */
 
-static int daemon_example_plugin_init(void *p __attribute__ ((unused)))
+static int daemon_example_plugin_init(void *)
```
dr-m, line 87: This should be cleaned up in `daemon_example_plugin_deinit()` as well. The `#define __attribute__` can then be removed.

plugin/daemon_example/daemon_example.cc
```diff
@@ -175,11 +175,9 @@ static int daemon_example_plugin_deinit(void *p __attribute__ ((unused)))
   my_write(con->heartbeat_file, (uchar*) buffer, strlen(buffer), MYF(0));
 
   /*
-    Need to wait for the hearbeat thread to terminate before closing
-    the file it writes to and freeing the memory it uses.
+    As thread is joined, we can close the file it writes to and
+    freeing the memory it uses.
```
dr-m, line 179: "We can freeing" does not compute. I’d move this comment before the `my_write()` call, because also that call could race with `mysql_heartbeat()`.

dr-m (approved):
Thank you. I have a couple minor suggestions.

PR #3651, opened 2024-11-21
MDEV-35478 part 2: Redo `space_id` format fix

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-35478](https://jira.mariadb.org/browse/MDEV-35478)*

## Description
`table->space_id` is `ulint` on 10.6 and `uint32_t` on 10.11+, but I included its format specifier change in 10.6 rather than 10.11.

PR #3650 reverts the change from 10.6; this commit reapplies it on 10.11 as a follow up on its batch (MDEV-35431, merged #3518).

## Release Notes
N/A

Note that this’ll be a merge conflict if this merges before #3518 merges to 10.11.

## How can this PR be tested?
N/A

## PR quality check

### DIFF
--- a/storage/innobase/dict/dict0load.cc
+++ b/storage/innobase/dict/dict0load.cc
@@ -2298,8 +2298,8 @@ dict_load_tablespace(
 		table->file_unreadable = true;
 
 		if (!(ignore_err & DICT_ERR_IGNORE_RECOVER_LOCK)) {
-			sql_print_error("InnoDB: Failed to load tablespace "
-					ULINTPF " for table %s",
+			sql_print_error("InnoDB: Failed to load tablespace %"
+					PRIu32 " for table %s",
 					table->space_id, table->name);
 		}
 	}


storage/innobase/dict/dict0load.cc
```diff
@@ -2298,8 +2298,8 @@ dict_load_tablespace(
 		table->file_unreadable = true;
 
 		if (!(ignore_err & DICT_ERR_IGNORE_RECOVER_LOCK)) {
-			sql_print_error("InnoDB: Failed to load tablespace "
-					ULINTPF " for table %s",
+			sql_print_error("InnoDB: Failed to load tablespace %"
+					PRIu32 " for table %s",
 					table->space_id, table->name);
```
dr-m, line 2303: I think that we could at the same time correct the last argument to `table->name.str`, like it was done in #3493. In any case, when or after that change has been merged from 10.6 to 10.11, we will have to pay attention to this.

dr-m (approved):

storage/innobase/dict/dict0load.cc
ParadoxV5: Sure, if it helps with conflict resolution

ParadoxV5: I suppose this thread is closed now that @grooverdan’ve merged the PR as is.

PR #3655, opened 2024-11-22
MDEV-35465 Async replication stops working on Galera async replica no…

temeo:
…de when parallel replication is enabled

- [x] *The Jira issue number for this PR is: MDEV-35465*

## Description

Parallel slave failed to retry in retry_event_group() with error

    WSREP: Parallel slave worker failed at wsrep_before_command() hook

Fix wsrep transaction cleanup/restart in retry_event_group() to properly clean up previous transaction by calling wsrep_after_statement(). Also move call to reset error after call to wsrep_after_statement() to make sure that it remains effective.

Add a MTR test galera_as_slave_parallel_retry to reproduce the error when the fix is not present.

Other issues which were detected when testing with sysbench:

Check if parallel slave is killed for retry before waiting for prior commits in THD::wsrep_parallel_slave_wait_for_prior_commit(). This is required with slave-parallel-mode=optimistic to avoid deadlock when a slave later in commit order manages to reach prepare phase before a lock conflict is detected.

Suppress wsrep applier specific warning for slave threads.

## How can this PR be tested?

This PR contains a new test `galera.galera_as_slave_parallel_retry` which was used to reproduce the issue. The test must pass when the fix is present.

## PR quality check


sysprg:
Thanks, the fix has been merged with the main branch:
https://github.com/MariaDB/server/commit/a2575a0703406f633659d9b9c8e0ff9750c888bf

PR #3656, opened 2024-11-23
MDEV-35475  Assertion `!rec_offs_nth_extern(offsets1, n)' failed in cmp_rec_rec_simple_field

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35475*

## Description
Problem:
=======
InnoDB wrongly stores the primary key field in externally stored off page during bulk insert operation. This leads to assert failure.

Solution:
========
row_merge_buf_blob(): Should store the primary key fields inline. Store the variable length field data externally based on the row format of the table.
g.

## How can this PR be tested?
./mtr innodb.alter_copy_bulk innodb.insert_into_empty

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/alter_copy_bulk.result
+++ b/mysql-test/suite/innodb/r/alter_copy_bulk.result
@@ -66,4 +66,28 @@ SELECT COUNT(*) FROM t;
 COUNT(*)
 2
 DROP TABLE t1, t2, t;
+#
+#  MDEV-35475 Assertion `!rec_offs_nth_extern(offsets1, n)'
+#            failed in cmp_rec_rec_simple_field
+#
+CREATE TABLE t1(a BLOB, b VARCHAR(2048), PRIMARY KEY (b)) ENGINE=InnoDB;
+INSERT INTO t1 VALUES
+(REPEAT('x',4805),'a'), (REPEAT('x',16111),'b'),
+(REPEAT('x',65535),'c'), (REPEAT('x',11312),'d'),
+(REPEAT('x',35177),'e'), (REPEAT('x',65535),'f'),
+(REPEAT('x',1988),'g'), (NULL,REPEAT('x',2048)),
+(REPEAT('x',2503),'h'), (REPEAT('x',33152),'i'),
+(REPEAT('x',65535),'j'), (REPEAT('x',1988),'k'),
+(REPEAT('x',65535),'l'), (REPEAT('x',65535),'m'),
+(REPEAT('x',65535),'n'), (REPEAT('x',65535),'o'),
+(REPEAT('x',1988),'p'), (REPEAT('x',2503),'q'),
+(REPEAT('x',65535),'r'), (REPEAT('x',65535),'s'),
+(REPEAT('x',65535),'t'), (REPEAT('x',3169),'u'),
+(REPEAT('x',7071),'v'), (REPEAT('x',16111),'w'),
+(REPEAT('x',2325),'x'), (REPEAT('x',33152),'y'),
+(REPEAT('x',65535),'z'), (REPEAT('x',65535),'aa'),
+(REPEAT('x',16111),'bb'), (REPEAT('x',4805),'cc'),
+(REPEAT('x',65535),'dd');
+ALTER TABLE t1 FORCE, ALGORITHM=COPY;
+DROP TABLE t1;
 SET GLOBAL innodb_stats_persistent=@default_stats_persistent;
--- a/storage/innobase/row/row0merge.cc
+++ b/storage/innobase/row/row0merge.cc
@@ -1057,7 +1057,8 @@ length in the field data
 @param	heap		heap to store the blob offset and length
 @return DB_SUCCESS if successful */
 static dberr_t row_merge_write_blob_to_tmp_file(
-   dfield_t *field, merge_file_t *blob_file,mem_heap_t **heap)
+   dfield_t *field, uint32_t local_len,
+   merge_file_t *blob_file, mem_heap_t **heap)
 {
   if (blob_file->fd == OS_FILE_CLOSED)
   {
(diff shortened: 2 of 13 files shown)


mysql-test/suite/innodb/r/alter_copy_bulk.result
```diff
@@ -66,4 +66,28 @@ SELECT COUNT(*) FROM t;
...
+(REPEAT('x',65535),'l'), (REPEAT('x',65535),'m'),
+(REPEAT('x',65535),'n'), (REPEAT('x',65535),'o'),
+(REPEAT('x',1988),'p'), (REPEAT('x',2503),'q'),
+(REPEAT('x',65535),'r'), (REPEAT('x',65535),'s'),
+(REPEAT('x',65535),'t'), (REPEAT('x',3169),'u'),
+(REPEAT('x',7071),'v'), (REPEAT('x',16111),'w'),
+(REPEAT('x',2325),'x'), (REPEAT('x',33152),'y'),
+(REPEAT('x',65535),'z'), (REPEAT('x',65535),'aa'),
+(REPEAT('x',16111),'bb'), (REPEAT('x',4805),'cc'),
+(REPEAT('x',65535),'dd');
```
dr-m, line 90: Can you add this `INSERT` statement also to the test `innodb.insert_into_empty`? I suspect that a similar code path will be invoked also in that case.

dr-m (commented):
Is any part of this applicable to `ALTER TABLE…ALGORITHM=INPLACE` in 10.5 or 10.6? Did you try to modify the test case to exercise that code path?

storage/innobase/row/row0merge.cc
```diff
@@ -5186,7 +5208,7 @@ dberr_t row_merge_bulk_t::bulk_insert_buffered(const dtuple_t &row,
     {
       /* Tuple data size is greater than srv_sort_buf_size */
       dtuple_t *big_tuple= row_merge_buf_large_tuple(
-        row, &m_blob_file, &large_tuple_heap);
+        row, &m_blob_file, &ind, &large_tuple_heap);
       if (row_merge_bulk_buf_add(buf, *ind.table, *big_tuple))
```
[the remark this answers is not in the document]
Thirunarayanan: Fixed

mysql-test/suite/innodb/r/alter_copy_bulk.result
Thirunarayanan: Done

storage/innobase/row/row0merge.cc
```diff
@@ -1060,83 +1061,75 @@ static dberr_t row_merge_write_blob_to_tmp_file(
...
-
-
-/** Write the field data whose length is more than 2000 bytes
-into blob temporary file and write offset, length into the
-tuple field
-@param entry     index fields to be encode the blob
+/** Store the variable length field data or fixed length
+character greater than 255 can be stored externally. Instead
+of storing it externally, InnoDB should store it in temporary
+file and write offset, length into the tuple field
```
dr-m, line 1094: The first sentence of this description is grammatically incorrect. Can you clarify the wording?

storage/innobase/row/row0merge.cc
```diff
@@ -1166,16 +1159,21 @@ dberr_t row_merge_buf_write(const row_merge_buf_t *buf,
...
 		if (blob_file) {
 			ut_ad(buf->index->is_primary());
-			err = row_merge_buf_blob(
-				entry, n_fields, &blob_heap, blob_file);
+			if (!blob_heap) {
+				blob_heap = mem_heap_create(100);
+			}
+
+			err = row_merge_buf_blob(entry->fields, n_fields,
+						 blob_file, buf->index,
```
dr-m, line 1167: Nice: `blob_file` is the 3rd argument, which matches the parameters of this function `#ifdef DBUG_OFF`. That means that the compiler has a chance to pass through this parameter without having to insert an instructions for copying the parameter register to another one.

storage/innobase/row/row0merge.cc
```diff
@@ -1166,16 +1159,21 @@ dberr_t row_merge_buf_write(const row_merge_buf_t *buf,
...
 			}
 		}
 
 		ulint rec_size= row_merge_buf_encode(
 				&b, index, entry, n_fields);
-		if (blob_file && rec_size > srv_page_size) {
+		if (blob_file && rec_size > BTR_PAGE_MAX_REC_SIZE) {
 			err = DB_TOO_BIG_RECORD;
 			goto func_exit;
 		}
```
dr-m, line 1179: This constant looks bogus to me and had better be replaced (in a separate bug ticket, targeting 10.5). Isn’t the maximum size of a record limited to 16383 bytes? In `ROW_FORMAT=REDUNDANT`, we reserve 7 or 14 bits per index field to store the end offset of each field. The most significant bit is the NULL flag, and in case of 2-byte offsets we have another flag `REC_2BYTE_EXTERN_MASK=1<<14`. In other formats, the limit is 16383 bytes per record field, for variable-length fields. See 9284e8b2c64aba1ac108512ecc415a2ea563c387.

dr-m (commented):
I think that this had better be tested also with `ROW_FORMAT=REDUNDANT` and `ROW_FORMAT=COMPACT`, possibly by means of `innodb_default_row_format.inc`.

storage/innobase/row/row0merge.cc
```diff
@@ -1060,83 +1061,75 @@ static dberr_t row_merge_write_blob_to_tmp_file(
       return DB_OUT_OF_MEMORY;
   }
   uint64_t val= blob_file->offset;
-  uint32_t len= field->len;
   dberr_t err= os_file_write(
     IORequestWrite, "(bulk insert)", blob_file->fd,
-    field->data, blob_file->offset, len);
+    (byte*)field->data + local_len, blob_file->offset,
```
dr-m, line 1066: `static_cast<const byte*>(field->data) + local_len` would be safer.

storage/innobase/row/row0merge.cc
```diff
@@ -1060,83 +1061,75 @@ static dberr_t row_merge_write_blob_to_tmp_file(
...
   /* Write length of the blob in 4 bytes */
-  mach_write_to_4(data + 16, len);
-  blob_file->offset+= field->len;
+  mach_write_to_4(data + local_len + 16, field->len - local_len);
+  blob_file->offset+= (field->len - local_len);
   blob_file->n_rec++;
-  dfield_set_data(field, data, BTR_EXTERN_FIELD_REF_SIZE);
+  dfield_set_data(field, data,
+                  local_len + BTR_EXTERN_FIELD_REF_SIZE);
   dfield_set_ext(field);
```
dr-m, line 1087: We could invoke `dfield_set_data()` and `dfield_set_ext()` at the start and then do `data += local_len` to reduce the diff and to possibly simplify the code, in case the common subexpression elimination (CSE) phase of the compiler might miss the repeated `data + local_len`.

storage/innobase/row/row0merge.cc
```diff
@@ -1060,83 +1061,75 @@ static dberr_t row_merge_write_blob_to_tmp_file(
...
+                             ? blob_prefix + FIELD_REF_SIZE
+                             : 2 * FIELD_REF_SIZE;
 
   for (ulint i= 0; i < n_fields; i++)
   {
-    dfield_t *field= &entry->fields[i];
-    if (dfield_is_null(field) || field->len <= 2000)
+    dfield_t *field= &fields[i];
+    if (i < index->n_uniq || dfield_is_null(field))
       continue;
```
dr-m, line 1118: Could we just adjust the initial value of the loop variable?
```c++
for (auto i= index->first_user_field(); i < n_fields; i++)
```

storage/innobase/row/row0merge.cc
```diff
@@ -1166,18 +1159,31 @@ dberr_t row_merge_buf_write(const row_merge_buf_t *buf,
 
 		if (blob_file) {
 			ut_ad(buf->index->is_primary());
-			err = row_merge_buf_blob(
-				entry, n_fields, &blob_heap, blob_file);
+			if (!blob_heap) {
+				blob_heap = mem_heap_create(100);
+			}
```
dr-m, line 1164: How was the initial size 100 bytes determined? If it is too small, a subsequent `mem_heap_alloc()` might just waste that block and allocate a new one. If it is too large, the end of the block would go wasted.

storage/innobase/row/row0merge.cc
```diff
@@ -3549,9 +3555,14 @@ in field data for the tuple
...
-static dberr_t row_merge_copy_blob_from_file(dtuple_t *tuple, mem_heap_t *heap,
+static dberr_t row_merge_copy_blob_from_file(dtuple_t *tuple,
+                                             dict_index_t *index,
+                                             mem_heap_t *heap,
                                              merge_file_t *blob_file)
 {
+  const uint blob_prefix= dict_table_has_atomic_blobs(index->table)
+                          ? 0
+                          : REC_ANTELOPE_MAX_INDEX_COL_LEN;
   for (ulint i = 0; i < dtuple_get_n_fields(tuple); i++)
```
dr-m, line 3566: Could the loop start as follows?
```c++
  for (auto i= index->first_user_field(); i < tuple->n_fields; i++)
```
No clustered index field before that may be stored externally.

storage/innobase/row/row0merge.cc
```diff
@@ -1066,83 +1067,79 @@ static dberr_t row_merge_write_blob_to_tmp_file(
...
+                             ? blob_prefix + FIELD_REF_SIZE
+                             : 2 * FIELD_REF_SIZE;
+
+  for (ulint i= index->first_user_field(); i < n_fields; i++)
   {
-    dfield_t *field= &entry->fields[i];
-    if (dfield_is_null(field) || field->len <= 2000)
+    dfield_t *field= &fields[i];
+    if (dfield_is_null(field))
       continue;
```
dr-m, line 1126: `dtuple_convert_big_rec()` is also checking `index->fields[i].fixed_len`, which we fail to check here.

storage/innobase/row/row0merge.cc
```diff
@@ -1066,83 +1067,79 @@ static dberr_t row_merge_write_blob_to_tmp_file(
...
 @param heap      heap to store the blob offset and blob length
-@param blob_file file to store the blob data */
-static dberr_t row_merge_buf_blob(const mtuple_t *entry, ulint n_fields,
-                                  mem_heap_t **heap, merge_file_t *blob_file)
+@return error code */
+static
+dberr_t row_merge_buf_blob(dfield_t *fields, ulint n_fields,
+                           merge_file_t *blob_file,
+                           const dict_index_t *index,
+                           mem_heap_t **heap)
```
dr-m, line 1113: Is the parameter `n_fields` redundant and always equal to `index->n_fields`?

storage/innobase/row/row0merge.cc
```diff
@@ -5197,8 +5212,17 @@ dberr_t row_merge_bulk_t::bulk_insert_buffered(const dtuple_t &row,
     if (buf->n_tuples == 0)
     {
       /* Tuple data size is greater than srv_sort_buf_size */
-      dtuple_t *big_tuple= row_merge_buf_large_tuple(
-        row, &m_blob_file, &large_tuple_heap);
+      dtuple_t *big_tuple;
+      if (!large_tuple_heap)
+        large_tuple_heap= mem_heap_create(DTUPLE_EST_ALLOC(row.n_fields));
+
+      big_tuple= dtuple_copy(&row, large_tuple_heap);
```
dr-m, line 5219: We seem to be missing `ut_ad(i == 0)` here. There can’t be any BLOBs in other indexes than the clustered index, right?

I’d declare and initialize `big_tuple` in a single statement.

At the start of this function, can we assert `ut_ad(row.n_fields == ind.n_fields)`?

storage/innobase/row/row0merge.cc
```diff
@@ -1066,83 +1067,79 @@ static dberr_t row_merge_write_blob_to_tmp_file(
...
-  if (!*heap)
-    *heap= mem_heap_create(100);
-
-  for (ulint i= 0; i < n_fields; i++)
+  const uint blob_prefix= dict_table_has_atomic_blobs(index->table)
+                          ? 0
+                          : REC_ANTELOPE_MAX_INDEX_COL_LEN;
+  const uint min_local_len = blob_prefix
+                             ? blob_prefix + FIELD_REF_SIZE
+                             : 2 * FIELD_REF_SIZE;
```
dr-m, line 1120: The `?` and `:` lines should be indented only by 2 extra spaces.

storage/innobase/row/row0merge.cc
```diff
@@ -3561,10 +3568,15 @@ in field data for the tuple
...
-static dberr_t row_merge_copy_blob_from_file(dtuple_t *tuple, mem_heap_t *heap,
+static dberr_t row_merge_copy_blob_from_file(dtuple_t *tuple,
+                                             dict_index_t *index,
+                                             mem_heap_t *heap,
                                              merge_file_t *blob_file)
 {
-  for (ulint i = 0; i < dtuple_get_n_fields(tuple); i++)
+  const uint blob_prefix= dict_table_has_atomic_blobs(index->table)
+                          ? 0
+                          : REC_ANTELOPE_MAX_INDEX_COL_LEN;
```
dr-m, line 3578: The `?` and `:` lines should be indented only by 2 extra spaces.

Can we assert `ut_ad(tuple->n_fields == index->n_fields)`?

PR #3660, opened 2024-11-25
MDEV-30469 Support ORDER BY and LIMIT for multi-table DELETE, index h…

DaveGosselin-MariaDB:
…ints for single-table DELETE.

We now allow multitable queries with order by and limit, such as:
  delete t1.*, t2.* from t1, t2 order by t1.id desc limit 3;
To predict what rows will be deleted, run the equivalent select:
  select t1.*, t2.* from t1, t2 order by t1.id desc limit 3;
Additionally, index hints are now supported with single table delete statements:
  delete from t2 use index(xid) order by (id) limit 2;

This approach changes the multi_delete SELECT result interceptor to use a temporary table to collect row ids pertaining to the rows that will be deleted, rather than directly deleting rows from the target table(s).  Row ids are collected during send_data, then read during send_eof to delete target rows.  In the event that the temporary table created in memory is not big enough for all matching rows, it is converted to an aria table.

This patch includes a number of changes that begin to make the code easier to maintain
  - Renamed SQL_I_List::link_in_list to SQL_I_List::insert.  link_in_list was ambiguous as it could refer to a link or it could refer to a node
  - Remove field_name local variable in multi_update::initialize_tables because it is not used when creating the temporary tables
  - multi_update changes:
    - Move temp table callocs to init, a more natural location for them, and moved tables_to_update to const member variable so we don't recompute it.
    - Filter out jtbm tables and tables not in the update map, pushing those that will be updated into an update_targets container.  This simplifies checks and loops in initialize_tables.

Other changes:

- Deleting from a sequence now affects zero rows instead of emitting an error

Limitations:
  - The federated connector does not create implicit row ids, so we to use a key when conditionally deleting.  See the change in federated_maybe_16324629.test

### DIFF
--- a/sql/sql_delete.cc
+++ b/sql/sql_delete.cc
@@ -41,6 +41,7 @@
 #include "filesort.h"
 #include "uniques.h"
 #include "sql_derived.h"                        // mysql_handle_derived
+#include "key.h"
                                                 // end_read_record
 #include "sql_insert.h"          // fix_rownum_pointers
 #include "sql_partition.h"       // make_used_partitions_str
@@ -1094,12 +1095,23 @@ extern "C" int refpos_order_cmp(void *arg, const void *a, const void *b)
 }
 
 
-multi_delete::multi_delete(THD *thd_arg, TABLE_LIST *dt, uint num_of_tables_arg):
-    select_result_interceptor(thd_arg), delete_tables(dt), deleted(0), found(0),
-    num_of_tables(num_of_tables_arg), error(0),
-    do_delete(0), transactional_tables(0), normal_tables(0), error_handled(0)
+multi_delete::multi_delete(THD *thd_arg,
+                           TABLE_LIST *dt,
+                           uint num_of_tables_arg)
+  : select_result_interceptor(thd_arg),
+    delete_tables(dt),
+    deleted(0),
+    found(0),
+    table_count(num_of_tables_arg),
+    error(0),
+    do_delete(0),
+    transactional_tables(0),
+    normal_tables(0),
+    error_handled(0)
 {
-  tempfiles= thd_arg->calloc<Unique*>(num_of_tables);
+  tempfiles= thd_arg->calloc<Unique*>(table_count);
+  tmp_tables = thd->calloc<TABLE*>(table_count);
+  tmp_table_param = thd->calloc<TMP_TABLE_PARAM>(table_count);
 }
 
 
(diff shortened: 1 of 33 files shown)


mysql-test/main/mdev-30469.test
```diff
@@ -0,0 +1,131 @@
+#
```
spetrunia, line 1: I think the name 'MDEV-30460.test' will not ring any bells for nearly anyone.  Please rename to something like 
'`delete_multi_order_by.test` ?

sql/sql_delete.cc
```diff
@@ -1182,25 +1237,46 @@ multi_delete::initialize_tables(JOIN *join)
...
-             walk == delete_tables)
-    {
-      /*
-        We are not deleting from the table we are scanning. In this
-        case send_data() shouldn't delete any rows a we may touch
-        the rows in the deleted table many times
-      */
-      delete_while_scanning= false;
+
+      // maybe don't do this if join->need_tmp is false, dunno
```
spetrunia, line 1241: The code for new features should have more confidence :-)

sql/sql_delete.cc
```diff
@@ -1182,25 +1237,46 @@ multi_delete::initialize_tables(JOIN *join)
...
+      group.item= (Item**) temp_fields.head_ref();
+      TMP_TABLE_PARAM *tmp_param;
+      prior->shared = index;
+      tmp_param= &tmp_table_param[prior->shared];
+      tmp_param->init();
+      tmp_param->tmp_name="update";
+      tmp_param->field_count= temp_fields.elements;
+      tmp_param->func_count=  temp_fields.elements - 1;
+      calc_group_buffer(tmp_param, &group);
+      /* small table, ignore @@big_tables */
```
spetrunia, line 1265: This looks cryptic for a casual reader. Is this "ignore @@big_tables" introduced by this patch? If yes  please elaborate about the reason...

sql/sql_delete.cc
```diff
@@ -1423,29 +1490,57 @@ int multi_delete::do_deletes()
...
+    goto err;
+  }
+  table->file->extra(HA_EXTRA_NO_CACHE);
+  if (unlikely((local_error= tmp_table->file->ha_rnd_init(1))))
   {
+    err_table= tmp_table;
+    goto err;
+  }
+
+  while (!thd->killed) {
```
spetrunia, line 1515: Coding style here and in code around:  `{` should be on its own line...

sql/sql_update.cc
```diff
@@ -1772,25 +1772,47 @@ bool Multiupdate_prelocking_strategy::handle_end(THD *thd)
...
-                           bool ignore_arg):
-   select_result_interceptor(thd_arg),
-   all_tables(table_list), leaves(leaves_list), update_tables(0),
-   tmp_tables(0), updated(0), found(0), fields(field_list),
-   values(value_list), table_count(0), copy_field(0),
-   handle_duplicates(handle_duplicates_arg), do_update(1), trans_safe(1),
-   transactional_tables(0), ignore(ignore_arg), error_handled(0), prepared(0),
-   updated_sys_ver(0)
+multi_update::multi_update(THD *thd_arg
+                           , TABLE_LIST *table_list
```
spetrunia, line 1776: starting lines from comma doesn't match the coding style.

spetrunia:
>  - DELETE can send empty results, so don't assert(0) on DA_EMPTY.

Please explain why?  I thought its result would be always "N rows deleted"  (where N can be 0).  What did change with this patch?

mysql-test/main/mdev-30469.test
DaveGosselin-MariaDB: Fixed.

sql/sql_delete.cc
DaveGosselin-MariaDB: Removed incorrect comment.

sql/sql_delete.cc
DaveGosselin-MariaDB: Fixed.

sql/sql_update.cc
DaveGosselin-MariaDB: Fixed.

DaveGosselin-MariaDB:
> > * DELETE can send empty results, so don't assert(0) on DA_EMPTY.
> 
> Please explain why? I thought its result would be always "N rows deleted" (where N can be 0). What did change with this patch?

My change was incorrect.  Sanja explained to me what to do on the MDEV, so I fixed my code accordingly.

sql/sql_delete.cc
DaveGosselin-MariaDB: It seems that this is no longer needed as of 11.4.  It was deprecated in 10.5 according to [this KB page](https://mariadb.com/kb/en/server-system-variables/).  Our KB has [another page](https://mariadb.com/docs/server/ref/mdb/system-variables/big_tables/) which says that this is an "old variable, which if set to 1, allows large result sets by saving all temporary sets to disk, avoiding 'table full' errors. No longer needed, as the server now handles this automatically."  I saw this when I was reading the `multi_update` code and preserved it in my own implementation for `multi_delete` as it seems reasonable, but allowing the system to decide automatically is better.

PR #3661, opened 2024-11-25
MDEV-34815 SIGILL error when executing mariadbd compiled for RISC-V with Clang

grooverdan:

## Description

RISC-V and Clang produce rdcycle for __builtin_readcyclecounter.

Since Linux kernel 6.6 this is a privileged instruction not available to userspace programs.

The use of __builtin_readcyclecounter is excluded from RISCV falling back to the rdtime/rdtimeh instructions provided in MDEV-33435.

Author: BINSZ on JIRA

## Release Notes

Correct SIGILL on RISC-V for Linux kernels 6.6+ when compiled with clang.

## How can this PR be tested?

Environment as above.

## PR quality check


include/my_rdtsc.h
```diff
@@ -97,7 +97,7 @@ C_MODE_START
 /**
   A cycle timer.
 
-  On clang we use __builtin_readcyclecounter(), except for AARCH64.
+  On clang we use __builtin_readcyclecounter(), except for AARCH64 and RISC-V.
```
dr-m, line 100: It would be nice to see a justification for these exceptions, such as links to clang bug reports. I found that the RISC-V implementation was added in https://reviews.llvm.org/D64125 but I found no bug report about this. Can you please file one?

For ARMv8 there already is a bug report https://github.com/llvm/llvm-project/issues/43652 where it is suggested that the instruction could work depending on the operating system kernel configuration.

These optionally privileged instructions could be usable if on the affected ISA we made `my_timer_cycles()` a function pointer. An initialization function could try to run the instruction under a SIGILL wrapper.

grooverdan: Quite right, was thinking the same. Clang isn't just for kernels and root processes. I did compiler export the latest version and the behaviour is still there.

include/my_rdtsc.h
```diff
@@ -152,7 +152,7 @@ C_MODE_START
 */
 static inline ulonglong my_timer_cycles(void)
 {
-# if __has_builtin(__builtin_readcyclecounter) && !defined (__aarch64__)
+# if __has_builtin(__builtin_readcyclecounter) && !defined (__aarch64__) && !defined(__riscv)
```
arichardson, line 155: This should be gated on __linux__, other operating systems such as FreeBSD don't make this inaccessible. Quite surprising to see that x86 rdtsc is permitted but the aarch64 counters are not.

grooverdan: Noted, adjusted. Seems the Aarch64 was removed for performance reasons instead rather than Kernel limitation. Oops mea cupla.

knielsen (approved):
LGTM, thanks, Daniel

PR #3662, opened 2024-11-25
MDEV-11675 fixup: MDEV-35474 Start Alter GTID Error Message Can Use Wrong Server_Id

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-35474](https://jira.mariadb.org/browse/MDEV-35474)*

## Description
Store a whole GTID in `start_alter_info` entries.

While Two-Phase ALTER only cares about the `domain_id` and `seq_no`, logs/warnings that show full GTIDs also need the ALTER’s `server_id`.

The Two-Phase ALTER work ([MDEV-11675](https://jira.mariadb.org/browse/MDEV-11675)) left an error message with an incomplete (and out-of-order), TODO GTID report.
The fix https://github.com/MariaDB/server/pull/3518/commits/8c18af91bf4bfe58d7ac02e04d0648af75d6886f also incorrectly filled it in with the *current primary* `server_id`, which isn’t necessarily the `server_id` of the ALTER (which varies on replication setups or `server_id` configs).

Instead of filling in a `struct start_alter_info::server_id` member, I decided to replace its current `sa_seq_no` & `domain_id` with a whole GTID (inline-memory) member.
I find this design more intuitively structural.

I normally wouldn’t refactor APIs in a bug fix, but it looks like only Two-Phase ALTER uses `struct`, so I made this exception.
Of course, this’d still conflict with collegues’ related work, if any.

## Release Notes
(We can nest this under #3518.)
Fixed the GTID in the warning message of when replication encounters a two-phase ALTER query that failed to complete in the primary.

## How can this PR be tested?
MDEV-11675 includes the test `rpl_start_alter_restart_master` for testing this “failed to complete” behavior.
I included the transaction server ID in its warning suppression to check for the current session server ID.

## PR quality check

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_start_alter_restart_master.test
+++ b/mysql-test/suite/rpl/t/rpl_start_alter_restart_master.test
@@ -26,7 +26,12 @@ set global gtid_strict_mode=1;
 start slave;
 
 --connection master
-call mtr.add_suppression("ALTER query started at .+ could not be completed");
+# MDEV-35474 Start Alter GTID Error Message Can Use Wrong Server_Id
+#
+# When replicating, the GTID the slave warns should be of this sessional
+# server_id rather than 1 from reconnecting to the restarted master.
+SET @@session.server_id= 11;
+call mtr.add_suppression("ALTER query started at 0-11-\\d+ could not be completed");
 
 SET @old_debug_master= @@global.debug;
 --let $binlog_alter_two_phase= `select @@binlog_alter_two_phase`
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -626,8 +626,9 @@ bool write_bin_log_start_alter(THD *thd, bool& partial_alter,
     start_alter_info *info= thd->rgi_slave->sa_info;
     bool is_shutdown= false;
 
-    info->sa_seq_no= start_alter_id;
-    info->domain_id= thd->variables.gtid_domain_id;
+    info->gtid.domain_id= thd->variables.gtid_domain_id;
+    info->gtid.server_id= thd->variables.server_id;
+    info->gtid.seq_no= start_alter_id;
     mysql_mutex_lock(&mi->start_alter_list_lock);
     // possible stop-slave's marking of the whole alter state list is checked
     is_shutdown= mi->is_shutdown;
--- a/sql/log_event_server.cc
+++ b/sql/log_event_server.cc
@@ -1682,8 +1682,9 @@ static start_alter_info *get_new_start_alter_info(THD *thd)
     sql_print_error("Failed to allocate memory for ddl log free list");
     return 0;
   }
-  info->sa_seq_no= 0;
-  info->domain_id= 0;
+  info->gtid.domain_id= 0;
+  info->gtid.server_id= 0;
+  info->gtid.seq_no= 0;
   info->direct_commit_alter= false;
   info->state= start_alter_state::INVALID;
   mysql_cond_init(0, &info->start_alter_cond, NULL);
@@ -1769,8 +1770,8 @@ int Query_log_event::handle_split_alter_query_log_event(rpl_group_info *rgi,
     List_iterator<start_alter_info> info_iterator(mi->start_alter_list);
     while ((info= info_iterator++))
     {
-      if(info->sa_seq_no == rgi->gtid_ev_sa_seq_no &&
-         info->domain_id == rgi->current_gtid.domain_id)
+      if(info->gtid.seq_no == rgi->gtid_ev_sa_seq_no &&
+         info->gtid.domain_id == rgi->current_gtid.domain_id)
       {
         info_iterator.remove();
         break;
@@ -2705,10 +2706,9 @@ static void check_and_remove_stale_alter(Relay_log_info *rli)
   {
     DBUG_ASSERT(info->state == start_alter_state::REGISTERED);
 
-    sql_print_warning("ALTER query started at %u-%lu-%llu could not "
+    sql_print_warning("ALTER query started at %u-%u-%llu could not "
                       "be completed because of unexpected master server "
-                      "or its binlog change", info->domain_id,
-                      mi->master_id, info->sa_seq_no);
+                      "or its binlog change", PARAM_GTID(info->gtid));
     info_iterator.remove();
     mysql_mutex_lock(&mi->start_alter_lock);
     info->state= start_alter_state::ROLLBACK_ALTER;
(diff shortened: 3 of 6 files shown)


sql/log.cc
```diff
@@ -626,8 +626,9 @@ bool write_bin_log_start_alter(THD *thd, bool& partial_alter,
     start_alter_info *info= thd->rgi_slave->sa_info;
     bool is_shutdown= false;
 
-    info->sa_seq_no= start_alter_id;
-    info->domain_id= thd->variables.gtid_domain_id;
+    info->gtid.domain_id= thd->variables.gtid_domain_id;
+    info->gtid.server_id= thd->variables.server_id;
+    info->gtid.seq_no= start_alter_id;
```
ParadoxV5, line 631: `start_alter_id` originates from `get_start_alter_id()`, which gives `thd->variables.gtid_seq_no` in a (configured) replica and `0` otherwise.
And there’re code that branches on this `0` or proceeds otherwise.

Perhaps the replica-checking logic should be split away from `start_alter_id`, so `THD::variables` can hold a GTID multi-member like `start_alter_info` in this PR?
Implementation’s out of this ticket’s scope, of course.

sql/rpl_rli.h
```diff
@@ -654,8 +654,7 @@ struct start_alter_info
   /*
     ALTER id is defined as a pair of GTID's seq_no and domain_id.
   */
-  decltype(rpl_gtid::seq_no) sa_seq_no; // key for searching (SA's id)
-  uint32 domain_id;
+  rpl_gtid gtid; // rpl_gtid::seq_no is the key for searching (SA's id)
   bool   direct_commit_alter; // when true CA thread executes the whole query
```
ParadoxV5, line 658: This (as opposed to `rpl_gtid *gtid`) will put the GTID in the memory next to sibling members like `direct_commit_alter` (at least in C). This also matches `sa_seq_no` and `domain_id`.

I’m not sure about this “key” designation in the comments, though – I didn’t encounter any “keying” while migrating affected code.

sql/log_event_server.cc
```diff
@@ -2705,10 +2706,9 @@ static void check_and_remove_stale_alter(Relay_log_info *rli)
   {
     DBUG_ASSERT(info->state == start_alter_state::REGISTERED);
 
-    sql_print_warning("ALTER query started at %u-%lu-%llu could not "
+    sql_print_warning("ALTER query started at %u-%u-%llu could not "
                       "be completed because of unexpected master server "
-                      "or its binlog change", info->domain_id,
-                      mi->master_id, info->sa_seq_no);
+                      "or its binlog change", PARAM_GTID(info->gtid));
```
ParadoxV5, line 2711: Maybe real reason I swapped to a nested struct is so I can write this `PARAM_GTID`? 🙃

ParadoxV5 (commented):
Oh yeah, I forgot to rebase my branch before pushing 🧒.

ParadoxV5:
🤔 Would it make sense to query this `server_id` by looking up the entry in the primary’s logs?
🤔🤔 Can an entry disappear from the primary logs inbetween the two phase of an ALTER?

sql/log.cc
```diff
@@ -626,8 +626,9 @@ bool write_bin_log_start_alter(THD *thd, bool& partial_alter,
     start_alter_info *info= thd->rgi_slave->sa_info;
     bool is_shutdown= false;
 
-    info->sa_seq_no= start_alter_id;
-    info->domain_id= thd->variables.gtid_domain_id;
+    info->gtid.domain_id= thd->variables.gtid_domain_id;
```
andrelkin, line 629: Shall we make the assignment more compact as `info->gtid= { /* copy-list initialization */ ... }`

mysql-test/suite/rpl/t/rpl_start_alter_restart_master.test
```diff
@@ -26,7 +26,12 @@ set global gtid_strict_mode=1;
 start slave;
 
 --connection master
-call mtr.add_suppression("ALTER query started at .+ could not be completed");
+# MDEV-35474 Start Alter GTID Error Message Can Use Wrong Server_Id
+#
+# When replicating, the GTID the slave warns should be of this sessional
+# server_id rather than 1 from reconnecting to the restarted master.
+SET @@session.server_id= 11;
+call mtr.add_suppression("ALTER query started at 0-11-\\d+ could not be completed");
```
andrelkin, line 34: This of course does not prove `11` is indeed in the warning.
Consider search in the error, like [here](https://github.com/MariaDB/server/blob/mdev-35474/mysql-test/suite/rpl/t/rpl_checksum.test#L321).

andrelkin (approved):
The patch looks good. Approved though
a couple of small notes would be good to address.

mysql-test/suite/rpl/t/rpl_start_alter_restart_master.test
ParadoxV5: The MTR suppression would still have to be here …🤔

sql/log.cc
ParadoxV5: Ooh, that wasn’t a thing in C. https://stackoverflow.com/a/18325112

mysql-test/suite/rpl/t/rpl_start_alter_restart_master.test
andrelkin: Right, the suppression stays.

sql/log.cc
andrelkin: I noticed you also found useful [cppreference](https://en.cppreference.com/w/cpp/language/aggregate_initialization) which in this case spells out the copy-list assignment is good.

mysql-test/suite/rpl/t/rpl_start_alter_restart_master.test
ParadoxV5: How’s this? I’ll rebase, squash and merge if it’s good.

sql/log.cc
```diff
@@ -626,9 +626,9 @@ bool write_bin_log_start_alter(THD *thd, bool& partial_alter,
     start_alter_info *info= thd->rgi_slave->sa_info;
     bool is_shutdown= false;
 
-    info->gtid.domain_id= thd->variables.gtid_domain_id;
-    info->gtid.server_id= thd->variables.server_id;
-    info->gtid.seq_no= start_alter_id;
+    info->gtid= {thd->variables.gtid_domain_id,
+                 static_cast<uint32>(thd->variables.server_id),
```
ParadoxV5, line 630: Shall we start nuking the existing `ulong`s, @vaintroub?

PR #3664, opened 2024-11-26
MDEV-35207 ignored error at binlogging by CREATE-TABLE-SELECT leads to assert + MDEV-35499 + MDEV-35502

andrelkin:

## Description
The PR branch consists of 4 commits.
The 1st one was reviewed, its edited commit message follows.
The 2nd one addresses the Kristian's notes.
The 3rd, with the summary line
```
MDEV-35502 incremental commit to address a special case of
```
makes sure CREATE-TABLE-SELECT handler chooses the correct cache and by that
eliminates possibility of failing in binlog_commit().
The 4th one extends tests to involve stored functions with side effects.

MDEV-35207 ignored error at binlogging by CREATE-TABLE-SELECT leads to assert

MDEV-35499 Errored-out CREATE-or-REPLACE-SELECT does not log DROP table into binlog
MDEV-35502 Failed at ROW-format binlogging CREATE-TABLE-SELECT should
           not generate Incident event

CREATE-TABLE-SELECT can face an error near the end of its execution
select_create::send_eof() but the error was never checked, therefore no
cleanup action were undertaken which led to various assert inside
binlogging.
E.g when binlog_commit() of ha_commit_one_phase() that
CREATE-TABLE-SELECT employs errored out because of a limited cache size
(binlog_commit may try writing to a transactional cache) the cache
was not flushed to binlog. The missed error check allowed further
execution down to trans_commit_implicit() in whose stack

  DBUG_ASSERT(!(entry->using_trx_cache && !mngr->trx_cache.empty() &&
                mngr->get_binlog_cache_log(TRUE)->error));

fired.

The fixes need and install the error checking in select_create::send_eof().
That prevents from any further execution when ha_commit_one_phase() fails
for any reason (typically due to binlog_commit()).
Note the patch did not address an immediate disconnect after
binlog_commit() error which is a MDEV-35506 subject.

This commit also covers CREATE-or-REPLACE-SELECT that additionally had
a specific issue in that DROP TABLE was not logged the binary log, MDEV-35499.
See select_create::abort_result_set().

The current commit also corrects an unnecessary Incident event
logging when CREATE-TABLE-SELECT encounters a binloging issue, MDEV-35502.
The Incident was actually only harmful in this case as the table was
never going to be created, therefore replicated, in such a case.

NOTE: a new test does not verify a binlog cache emptiness upon disconnect
which is the MDEV-35506 subject.

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_create_select_row.result
+++ b/mysql-test/suite/rpl/r/rpl_create_select_row.result
@@ -6,7 +6,7 @@ set @binlog_cache_size             = @@global.binlog_cache_size;
 set @@global.max_binlog_cache_size = 4096;
 set @@global.    binlog_cache_size = 4096;
 #
-# MDEV-35207
+# MDEV-35207 ignored error at binlogging by CREATE-TABLE-SELECT leads to assert
 #
 connect  conn_err,localhost,root,,;
 call mtr.add_suppression("Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage");
@@ -22,7 +22,7 @@ disconnect conn_err;
 connection master;
 
 #
-# TODO: MDEV-35499 errored CoRS does not DROP table in binlog
+# MDEV-35499 errored CREATE-OR-REPLACE-SELECT does not DROP table in binlog
 #
 #
 # Engine = innodb
--- a/mysql-test/suite/rpl/r/rpl_create_select_row.result
+++ b/mysql-test/suite/rpl/r/rpl_create_select_row.result
@@ -1,0 +1,109 @@
+include/master-slave.inc
+[connection master]
+connection master;
+set @max_binlog_cache_size         = @@global.max_binlog_cache_size;
+set @binlog_cache_size             = @@global.binlog_cache_size;
+set @@global.max_binlog_cache_size = 4096;
+set @@global.    binlog_cache_size = 4096;
+#
+# MDEV-35207
+#
+connect  conn_err,localhost,root,,;
+call mtr.add_suppression("Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage");
+create table t engine=myisam select repeat ('a',4096*3) AS a;
+ERROR HY000: Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage; increase this mariadbd variable and try again
+create table t engine=innodb select repeat ('a',4096*3) AS a;
+ERROR HY000: Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage; increase this mariadbd variable and try again
+create table t (a int unique, b char) select 1 AS a, 'b' as b union select 1 as a, 'c' as b;
+ERROR 23000: Duplicate entry '1' for key 'a'
+select * from t;
+ERROR 42S02: Table 'test.t' doesn't exist
+disconnect conn_err;
+connection master;
+
+#
+# TODO: MDEV-35499 errored CoRS does not DROP table in binlog
+#
+#
+# Engine = innodb
+#
+set statement binlog_format=statement for create table t (a int) select 1 as a;
+set statement binlog_format=row for create or replace table t (a int primary key, b char) engine=innodb select 1 AS a, 'b' as b union select 1 as a, 'c' as b;
+ERROR 23000: Duplicate entry '1' for key 'PRIMARY'
+select * from t;
+ERROR 42S02: Table 'test.t' doesn't exist
+#
+# Prove an expected lonely `DROP table t'
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
+master-bin.000001	#	Query	#	#	use `test`; DROP TABLE IF EXISTS `test`.`t`/* Generated to handle failed CREATE OR REPLACE */
+master-bin.000001	#	Query	#	#	ROLLBACK
+set statement binlog_format=statement for create table t (a int) select 1 as a;
+set statement binlog_format=row for create or replace table t (a text) engine=innodb select repeat ('a',1024) AS a union select repeat ('a',3*4096) AS a union select repeat ('a',3*4096) AS a;
+ERROR HY000: Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage; increase this mariadbd variable and try again
+select * from t;
+ERROR 42S02: Table 'test.t' doesn't exist
+#
+# Prove an expected lonely `DROP table t'
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
+master-bin.000001	#	Query	#	#	use `test`; DROP TABLE IF EXISTS `test`.`t`/* Generated to handle failed CREATE OR REPLACE */
+master-bin.000001	#	Query	#	#	ROLLBACK
+set statement binlog_format=statement for create table t (a int) select 1 as a;
+set statement binlog_format=row for create or replace table t (a text) engine=innodb select repeat ('a',4096*3) AS a;;
+ERROR HY000: Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage; increase this mariadbd variable and try again
+select * from t;
+ERROR 42S02: Table 'test.t' doesn't exist
+#
+# Prove an expected lonely `DROP table t'
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
+master-bin.000001	#	Query	#	#	use `test`; DROP TABLE IF EXISTS `test`.`t`/* Generated to handle failed CREATE OR REPLACE */
+master-bin.000001	#	Query	#	#	ROLLBACK
+#
+# Engine = myisam
+#
+set statement binlog_format=statement for create table t (a int) select 1 as a;
+set statement binlog_format=row for create or replace table t (a int primary key, b char) engine=myisam select 1 AS a, 'b' as b union select 1 as a, 'c' as b;
+ERROR 23000: Duplicate entry '1' for key 'PRIMARY'
+select * from t;
+ERROR 42S02: Table 'test.t' doesn't exist
+#
+# Prove an expected lonely `DROP table t'
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
+master-bin.000001	#	Query	#	#	use `test`; DROP TABLE IF EXISTS `test`.`t`/* Generated to handle failed CREATE OR REPLACE */
+master-bin.000001	#	Query	#	#	ROLLBACK
+set statement binlog_format=statement for create table t (a int) select 1 as a;
+set statement binlog_format=row for create or replace table t (a text) engine=myisam select repeat ('a',1024) AS a union select repeat ('a',3*4096) AS a union select repeat ('a',3*4096) AS a;
+ERROR HY000: Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage; increase this mariadbd variable and try again
+select * from t;
+ERROR 42S02: Table 'test.t' doesn't exist
+#
+# Prove an expected lonely `DROP table t'
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
+master-bin.000001	#	Query	#	#	use `test`; DROP TABLE IF EXISTS `test`.`t`/* Generated to handle failed CREATE OR REPLACE */
+master-bin.000001	#	Query	#	#	ROLLBACK
+set statement binlog_format=statement for create table t (a int) select 1 as a;
+set statement binlog_format=row for create or replace table t (a text) engine=myisam select repeat ('a',4096*3) AS a;;
+ERROR HY000: Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage; increase this mariadbd variable and try again
+select * from t;
+ERROR 42S02: Table 'test.t' doesn't exist
+#
+# Prove an expected lonely `DROP table t'
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
+master-bin.000001	#	Query	#	#	use `test`; DROP TABLE IF EXISTS `test`.`t`/* Generated to handle failed CREATE OR REPLACE */
+master-bin.000001	#	Query	#	#	ROLLBACK
+SET @@global.max_binlog_cache_size = @max_binlog_cache_size;
+SET @@global.    binlog_cache_size = @binlog_cache_size;
+connection slave;
+End of the tests
+include/rpl_end.inc
--- a/mysql-test/suite/rpl/r/rpl_create_select_row.result
+++ b/mysql-test/suite/rpl/r/rpl_create_select_row.result
@@ -102,6 +102,55 @@ Log_name	Pos	Event_type	Server_id	End_log_pos	Info
 master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
 master-bin.000001	#	Query	#	#	use `test`; DROP TABLE IF EXISTS `test`.`t`/* Generated to handle failed CREATE OR REPLACE */
 master-bin.000001	#	Query	#	#	ROLLBACK
+create table ti_pk (a int primary key) engine=innodb;
+create table ta (a int) engine=aria;
+create function f_ia(arg int)
+returns integer
+begin
+insert into ti_pk set a=1;
+insert into ta set a=1;
+insert into ti_pk set a=arg;
+return 1;
+end |
+set statement binlog_format = ROW for create table t_y (a int) engine=aria select f_ia(1 /* err */) as a;
+ERROR 23000: Duplicate entry '1' for key 'PRIMARY'
+select * from t_y;
+ERROR 42S02: Table 'test.t_y' doesn't exist
+# correct execution: `ta` is modified and its new record is binlogged
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
+master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
+master-bin.000001	#	Table_map	#	#	table_id: # (test.ta)
+master-bin.000001	#	Write_rows_v1	#	#	table_id: # flags: STMT_END_F
+master-bin.000001	#	Query	#	#	COMMIT
+select * from ta;
+a
+1
+select * from ti_pk;
+a
+connection slave;
+include/diff_tables.inc [master:ta,slave:ta]
+connection master;
+delete from ta;
+connection slave;
+connection master;
+set statement binlog_format = STATEMENT for create table t_y (a int) engine=aria select f_ia(1 /* err */) as a;
+ERROR 23000: Duplicate entry '1' for key 'PRIMARY'
+select * from t_y;
+ERROR 42S02: Table 'test.t_y' doesn't exist
+# MDEV-36027: `ta` is modified but that's not binlogged
+include/show_binlog_events.inc
+select *,'on_master' from ta;
+a	on_master
+1	on_master
+select * from ti_pk;
+a
+connection slave;
+select *,'on_slave' from ta;
+a	on_slave
+connection master;
+drop function f_ia;
+drop table ti_pk, ta;
 SET @@global.max_binlog_cache_size = @max_binlog_cache_size;
 SET @@global.    binlog_cache_size = @binlog_cache_size;
 connection slave;
--- a/mysql-test/suite/rpl/t/rpl_create_select_row.test
+++ b/mysql-test/suite/rpl/t/rpl_create_select_row.test
@@ -9,9 +9,9 @@ set @@global.max_binlog_cache_size = 4096;
 set @@global.    binlog_cache_size = 4096;
 
 --echo #
---echo # MDEV-35207
+--echo # MDEV-35207 ignored error at binlogging by CREATE-TABLE-SELECT leads to assert
 --echo #
-# fixate the current (write) binlog position
+# fix the current (write) binlog position
 --let $binlog_file_0= query_get_value(SHOW MASTER STATUS, File, 1)
 --let $binlog_start_0 = query_get_value(SHOW MASTER STATUS, Position, 1)
 
@@ -45,7 +45,7 @@ if (!$cmp)
 
 --echo
 --echo #
---echo # TODO: MDEV-35499 errored CoRS does not DROP table in binlog
+--echo # MDEV-35499 errored CREATE-OR-REPLACE-SELECT does not DROP table in binlog
 --echo #
 --let $i = 2
 while ($i)
@@ -95,7 +95,7 @@ while ($i)
 SET @@global.max_binlog_cache_size = @max_binlog_cache_size;
 SET @@global.    binlog_cache_size = @binlog_cache_size;
 
-# prove binlog is sane for slave as well
+# test that binlog replicates correctly to slave
 # --connection slave
 --sync_slave_with_master
 
--- a/mysql-test/suite/rpl/t/rpl_create_select_row.test
+++ b/mysql-test/suite/rpl/t/rpl_create_select_row.test
@@ -92,6 +92,64 @@ while ($i)
 --dec $i
 }
 
+# Tests of mixed engines to demonstrate non-transaction table updates
+# are binlogged or otherwise MDEV-36027.
+create table ti_pk (a int primary key) engine=innodb;
+create table ta (a int) engine=aria;
+delimiter |;
+create function f_ia(arg int)
+returns integer
+begin
+  insert into ti_pk set a=1;
+  insert into ta set a=1;
+  insert into ti_pk set a=arg;
+  return 1;
+end |
+delimiter ;|
+
+--let $binlog_file= query_get_value(SHOW MASTER STATUS, File, 1)
+--let $binlog_start = query_get_value(SHOW MASTER STATUS, Position, 1)
+
+--error ER_DUP_ENTRY
+set statement binlog_format = ROW for create table t_y (a int) engine=aria select f_ia(1 /* err */) as a;
+--error ER_NO_SUCH_TABLE
+select * from t_y;
+
+--echo # correct execution: `ta` is modified and its new record is binlogged
+--source include/show_binlog_events.inc
+select * from ta;
+select * from ti_pk;
+
+--sync_slave_with_master
+--let  $diff_tables=master:ta,slave:ta
+--source include/diff_tables.inc
+
+--connection master
+delete from ta;
+--sync_slave_with_master
+
+--connection master
+# MDEV-36027 Errored-out CREATE-SELECT does not binlog results of any function modifying non-transactional table
+--let $binlog_file= query_get_value(SHOW MASTER STATUS, File, 1)
+--let $binlog_start = query_get_value(SHOW MASTER STATUS, Position, 1)
+--error ER_DUP_ENTRY
+set statement binlog_format = STATEMENT for create table t_y (a int) engine=aria select f_ia(1 /* err */) as a;
+--error ER_NO_SUCH_TABLE
+select * from t_y;
+
+--echo # MDEV-36027: `ta` is modified but that's not binlogged
+--source include/show_binlog_events.inc
+select *,'on_master' from ta;
+select * from ti_pk;
+
+--sync_slave_with_master
+select *,'on_slave' from ta;
+
+# Cleanup
+--connection master
+drop function f_ia;
+drop table ti_pk, ta;
+
 SET @@global.max_binlog_cache_size = @max_binlog_cache_size;
 SET @@global.    binlog_cache_size = @binlog_cache_size;
 
--- a/mysql-test/suite/rpl/t/rpl_create_select_row.test
+++ b/mysql-test/suite/rpl/t/rpl_create_select_row.test
@@ -1,0 +1,103 @@
+--source include/have_binlog_format_row.inc
+--source include/have_innodb.inc
+--source include/master-slave.inc
+
+--connection master
+set @max_binlog_cache_size         = @@global.max_binlog_cache_size;
+set @binlog_cache_size             = @@global.binlog_cache_size;
+set @@global.max_binlog_cache_size = 4096;
+set @@global.    binlog_cache_size = 4096;
+
+--echo #
+--echo # MDEV-35207
+--echo #
+# fixate the current (write) binlog position
+--let $binlog_file_0= query_get_value(SHOW MASTER STATUS, File, 1)
+--let $binlog_start_0 = query_get_value(SHOW MASTER STATUS, Position, 1)
+
+# use a separate connection also to validate its close will be clean
+connect (conn_err,localhost,root,,);
+
+call mtr.add_suppression("Multi-statement transaction required more than 'max_binlog_cache_size' bytes of storage");
+--error ER_TRANS_CACHE_FULL
+create table t engine=myisam select repeat ('a',4096*3) AS a;
+
+--error ER_TRANS_CACHE_FULL
+create table t engine=innodb select repeat ('a',4096*3) AS a;
+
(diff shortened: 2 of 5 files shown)


cvicentiu:
@andrelkin You probably want to rebase your branch on top of most recent `main` branch and force-push. Or perhaps the PR should be targeting a different branch than `main`. It's not 100% clear to me, but right now it looks like you are trying to merge your change together with others`

andrelkin:
Thanks, @cvicentiu ! (And sorry to force you replicating your alarms).

knielsen:

> @andrelkin requested your review on: MariaDB/server#3664 MDEV-35207
> ignored error at binlogging by CREATE-TABLE-SELECT leads to assert +
> MDEV-35499 + MDEV-35502.

Here is my review of the patch.

I found the commit description very hard to read, to the point that it makes it
hard to be sure I understand what is meant. Please check below suggestions,
and if you meant something different correct them:

> MDEV-35207 ignored error at binlogging by CREATE-TABLE-SELECT leads to assert

"Binlogging of CREATE-SELECT ignores error, causing assertion" ?

> MDEV-35499 Errored-out CREATE-or-REPLACE-SELECT does not DROP table in binlog

"Error during CREATE-OR-REPLACE drops table on master without binlogging DROP TABLE" ?

> MDEV-35502 Failed at ROW-format binlogging CREATE-TABLE-SELECT should
> not cause Incident event

"Error during row-based binlogging of CREATE-SELECT should not generate INCIDENT event" ?

> ha_commit_one_phase() path to commit CREATE-TABLE-SELECT features
> a possibility of a failure within binlog_commit() before to flush
> the transaction's caches to binlog.

What is a "ha_commit_one_phase() path to commit" ?

Do you mean a particular scenario of CREATE-SELECT where the code ends up
calling ha_commit_one_phase(), and if so, how is this scenario different
from others?

Or do you mean that CREATE-SELECT _always_ calls into ha_commit_one_phase(),
and that ha_commit_one_phase() in turn callse binlog_commit() which may
fail?

> In such case the "autocommit" statement is supposed to roll back.

But CREATE-SELECT is not affected by auto-commit state, is it? DDL commits
any active transaction, and then itself runs non-transactionally, doesn't
it? Is this TEMPORARY table, which I think (not sure) doesn't commit an
active transaction? I don't think it can ever roll back any existing
transaction. With "roll back", do you mean that the CREATE TABLE is aborted
and any partially created table deleted?

> However CREATE-TABLE-SELECT handler skipped checking of an error
> presence so execution proceeded to either implicit commit or send OK
> to hit an assert in ether branch.

The "skipped checking of an error presence" is cryptic, better just mention
what is concretely wrong. Eg. "function foo() did not check the error return
when calling bar() and allowed the create operation to succeed. This later
asserts because an error is left set in the diagnostics_area".

What do you mean by "implicit commit"? Normally, "implicit commit" is the
commit of any active transaction that happens when a DDL is run (for
example), but I'm not sure that is what you mean here? Also, when you say
"either implicit commit or send OK", it sounds as if it does either one or
the other (if so, when does it do one and when the other), but again I am
not sure that is what you mean.

> The fixes install the checking in select_create::send_eof().

"install the checking", that is very cryptic and hard to understand. Do you
simply mean something like "The patch adds the missing error checking in
function select_create::send_eof()" ?

> This commit also covers CREATE-or-REPLACE-SELECT that additionally had
> a "personal" issue to miss DROP TABLE logging into the binary log, MDEV-35499.

I do not understand what you mean with "a 'personal' issue". What does it
mean for an SQL statement to have a personal issue? It is much better to
describe briefly what actually is going on. For example:

  "This commit also fixes a case where CREATE-OR-REPLACE-AS-SELECT could
  fail on the master and drop the old table, but no corresponding DROP TABLE
  was binlogged"

, if that is what you mean?

> The current commit also corrects from unnecessary Incident event

"also corrects from unnecessary" -> "also corrects an unnecessary" ?

> logging when it's C-T-S that suffers from a binloging issue, MDEV-35502.

It is best to be careful with using abbreviations like "C-T-S", as it makes
things harder to read, and definitely never use them without explaining
them.

Using words like "suffers" also makes it harder to understand the text, as
it is imprecise - in this case a word like "encounter" for example might be
better?

I want to be clear that I'm not trying to pick on not using perfect English
here, most of us are non-native speakers. But I'm pointing out these things
because they disrupt the reading to the point where the text is almost
incomprehensible, even to me. And I hope I gave concrete suggestions on how
using simpler, but more concrete and precise wording can easily help make
the text easier to understand.

> diff --git a/sql/log.cc b/sql/log.cc
> index 42c8dc31383..4747e3b4fc0 100644
> --- a/sql/log.cc
> +++ b/sql/log.cc
> @@ -1957,9 +1957,13 @@ binlog_truncate_trx_cache(THD *thd, binlog_cache_mngr *cache_mngr, bool all)
> */
> if (ending_trans(thd, all))
> {
> -    if (cache_mngr->trx_cache.has_incident())
> -      error= mysql_bin_log.write_incident(thd);
> +    DBUG_ASSERT(!cache_mngr->trx_cache.has_incident() ||
> +		thd->lex->sql_command != SQLCOM_CREATE_TABLE ||
> +		thd->is_current_stmt_binlog_format_row());
> 
> +    if (cache_mngr->trx_cache.has_incident() &&
> +	thd->lex->sql_command != SQLCOM_CREATE_TABLE)
> +      error= mysql_bin_log.write_incident(thd);

This cannot be right. If SQLCOM_CREATE_TABLE does not want an incident event
written, then it should arrange for has_incident() to not get set. With
this, the semantics of has_incident() becomes something like "There is a
pending incident event which must be written to the binlog, unless the
current command is SQLCOM_CREATE_TABLE, in which case has_incident() has no
meaning". That is much too complex.

How does the incident event get set in the first place? If the incident
event is meaningful, should it be binlogged before binlogging the DROP TABLE
(if that is where the problem arises)? Or maybe the incident event is not
meaningful (if I understood correctly from commit comments), and then
preferably it should never be generated, or if that is complex, it could
simply be cleared (with a good comment explaining why) at the point in the
code where the situation arises?


> if (create_info->table_was_deleted)
> {
> -    /* Unlock locked table that was dropped by CREATE. */
> -    (void) trans_rollback_stmt(thd);
> +    if (has_drop_table_logged)
> +    {
> +      /* for DROP binlogging the error status has to be canceled first */
> +      Diagnostics_area new_stmt_da(thd->query_id, false, true);
> +      Diagnostics_area *old_stmt_da= thd->get_stmt_da();
> +
> +      thd->set_stmt_da(&new_stmt_da);
> +      (void) trans_rollback_stmt(thd);
> +      thd->set_stmt_da(old_stmt_da);
> +    }
> +    else
> +    {
> +      /* Unlock locked table that was dropped by CREATE. */
> +      (void) trans_rollback_stmt(thd);
> +    }
> thd->locked_tables_list.unlock_locked_table(thd, create_info->mdl_ticket);
> }

Agree.

It seems the DROP TABLE, which is in the statement cache, is binlogged by
trans_rollback_stmt(), through the function binlog_rollback().
I'm wondering if it is actually appropriate to call binlog_rollback() here,
we are not rolling back a transaction, we are aborting with an error.
But this patch is for 10.6, whether or not the code could be refactored to
be better, that would not be appropriate for 10.6 and is outside of the
scope of this patch.

One suggestion could be to add assertions that the stmt cache is non-empty
in the has_drop_table_logged case, and that it is empty in the
!has_drop_table_logged case (if those assertions are actually correct?)


> @@ -5286,6 +5292,7 @@ void select_create::abort_result_set()
> /* possible error of writing binary log is ignored deliberately */
> (void) thd->binlog_flush_pending_rows_event(TRUE, TRUE);
> 
> +  bool has_drop_table_logged= false;

Clearer name: "did_binlog_drop_table" or "binlogged_drop_table" (but just a
suggestion, up to you).

> diff --git a/sql/sql_insert.cc b/sql/sql_insert.cc
> index 0837449071d..70894c06c7f 100644
> --- a/sql/sql_insert.cc
> +++ b/sql/sql_insert.cc
> @@ -4450,7 +4450,8 @@ void select_insert::abort_result_set()
> query_cache_invalidate3(thd, table, 1);
> }
> DBUG_ASSERT(transactional_table || !changed ||
> -		thd->transaction->stmt.modified_non_trans_table);
> +		(thd->transaction->stmt.modified_non_trans_table ||
> +                 thd->transaction->all.modified_non_trans_table));
> 
> table->s->table_creation_was_logged|= binary_logged;
> table->file->ha_release_auto_increment();

I couldn't understand why you need to extend the DBUG_ASSERT condition here.
Why can the old assert now fail with your patch, and why is it ok anyway?
Is it because we now call abort_result_set() after trans_commit_stmt(thd) ?


> diff --git a/mysql-test/suite/rpl/t/rpl_create_select_row.test b/mysql-test/suite/rpl/t/rpl_create_select_row.test
> new file mode 100644
> index 00000000000..62d946da79d
> --- /dev/null
> +++ b/mysql-test/suite/rpl/t/rpl_create_select_row.test

> +# fixate the current (write) binlog position
> +--let $binlog_file_0= query_get_value(SHOW MASTER STATUS, File, 1)
> +--let $binlog_start_0 = query_get_value(SHOW MASTER STATUS, Position, 1)

I do not understand "fixate", what do you mean?

> +--echo # TODO: MDEV-35499 errored CoRS does not DROP table in binlog

I suggest to spell it out ("CREATE-OR-REPLACE-SELECT" ?) instead of the
somewhat cryptic abbreviation "CoRS".

> +# prove binlog is sane for slave as well
> +# --connection slave
> +--sync_slave_with_master

"Test that binlog replicates correctly to slave".

 - Kristian.

knielsen (changes requested):
See previous comments

andrelkin:
Thanks for your feedback Kristian!
Sorry for the troubles to understand the commit messages that you faced.
It's a bit too late to answer to your rather generous list of notes at my current later hour,
I am going to do that comprehensively tomorrow.

Let me attend an "english" related (that's how I perceive it) pair

> I found the commit description very hard to read, to the point that it makes it hard to be sure I understand what is meant. Please check below suggestions, and if you meant something different correct them:
>> MDEV-35207 ignored error at binlogging by CREATE-TABLE-SELECT leads to assert

> "Binlogging of CREATE-SELECT ignores error, causing assertion" ?

Indeed the original formation simply states there was an error at certain stage of C-T-S execution, namely at its binlogging, and that led to an assert.

I hope you did not mean not perfect english line confused you..

andrelkin:
> andrelkin ***@***.***> writes:
> @andrelkin requested your review on: [MariaDB/server#3664](https://github.com/MariaDB/server/pull/3664) 

Answering to the rest of the "english" block of notes. The technical notes follow after a short break.


>>  MDEV-35499 Errored-out CREATE-or-REPLACE-SELECT does not DROP table in binlog

> "Error during CREATE-OR-REPLACE drops table on master without binlogging DROP TABLE" ?

One can say that though I preferred 'drop table in binlog' because from the 2pc
point of view, which is relevant to the current bug's context,
the Binlog is a resource and in this way similar to the Engine.


>>  MDEV-35502 Failed at ROW-format binlogging CREATE-TABLE-SELECT should not cause Incident event

> "Error during row-based binlogging of CREATE-SELECT should not generate INCIDENT event" ?

One can say that.

>> ha_commit_one_phase() path to commit CREATE-TABLE-SELECT features  a possibility of a failure within binlog_commit() before to flush  the transaction's caches to binlog.

> What is a "ha_commit_one_phase() path to commit" ? Do you mean ...

You were thinking in the right direction here
> the code ends up calling ha_commit_one_phase(), and if so, how is this scenario different from others?

C-T-S has a way too specific path of execution commit. It does so via ha_commit_one_phase() which you know of course
what is.
Unlike all other types of objects that execute 1pc, the C-T-S does not flush the pending Rows_log_event at the end of
insertion. Therefore binlog_commit() has to complete, and it has been ready to do so since the ROW format inception,
the caching in which phase the caching error may happen too.


>> In such case the "autocommit" statement is supposed to roll back.

> But CREATE-SELECT is not affected by auto-commit state, is it?..
> With "roll back", do you mean that the CREATE TABLE is aborted and any partially created table deleted?

Right.
One can sense a rollback in the abort of CREATE-SELECT. That makes it autocommit at least that consistent with
a def of autocommit, e.g in [https://docs.oracle.com/javadb/10.5.1.1/devguide/cdevconcepts29416.html]

```Auto-commit mode means that when a statement is completed, the method commit is called on that statement automatically. Auto-commit in effect makes every SQL statement a transaction.```


> The "skipped checking of an error presence" is cryptic, better just mention what is concretely wrong.

I agree, that might've been specific. Much much better if the commit messages simply states
that any further actions of normal execution after the error binlog_commit() might lead to asserts or worse.

Details would be clearly irrelevant, including this question

> What do you mean by "implicit commit"?
what I was referring is

```@@ -5164,9 +5165,14 @@ bool select_create::send_eof()
     /* Remember xid's for the case of row based logging */
     ddl_log_update_xid(&ddl_log_state_create, thd->binlog_xid);
     ddl_log_update_xid(&ddl_log_state_rm, thd->binlog_xid);
-    trans_commit_stmt(thd);
-    if (!(thd->variables.option_bits & OPTION_GTID_BEGIN))
-      trans_commit_implicit(thd);
+    if (trans_commit_stmt(thd) ||
+	(!(thd->variables.option_bits & OPTION_GTID_BEGIN) &&
+	 trans_commit_implicit(thd)))
```

> "install the checking", that is very cryptic and hard to understand... "The patch adds the missing error checking in
function select_create::send_eof()"

Fine with me.


> What does it mean for an SQL statement to have a personal issue?

Well, notice in your question an important part of the double quotes *disappeared*. Of course it was used figuratively,
to underline the REPLACE kind of CREATE-SELECT has something really peculiar (a "whim" if I may) unlike the simple one.
(I almost had a nightmare dream last night in which I thought the quotes were forgotten :crying_cat_face:).

> "This commit also fixes a case where CREATE-OR-REPLACE-AS-SELECT ...
> if that is what you mean?

Yes.

> "also corrects from unnecessary" -> "also corrects an unnecessary" ?

Accepted. This was clumsy. Thanks for pointing to it!

> using abbreviations like "C-T-S"

I'd sympathize with this note. However, it'd be to much to ask from the commit message.
Naturally operating with undefined abbr:s (abbreviations) should be careful.
I hoped C-T-S would read ambiguously when it is put after CREATE-TABLE-SELECT.
Did it do btw to you?

> in this case a word like "encounter" for example might be better?

I am not sure. 'To suffer' is more emotional, I feel it consistent with `"personal"`.
However it is generally used in technical writing.
I'd say it's a freedom of expression in practice as long as unambiguous.

>  I'm not trying to pick on not using perfect English ...  the text is almost incomprehensible, even to me.

I believe we all can do the description part better. I'll certainly be trying.
However I believe the ultimate objective should be the description should reflect key parts of
the coding work (ideally refer to pertinent patch lines with one click).
So if it speaks clumsily but those hunks are identifiable, it would pass to me.


> I hope I gave concrete suggestions

Thank you indeed!
I learned something from that and I do hope my explanations shed some light
on the immediate (and future, sorry :smiling_face_with_tear: !)  obscurities.

We sure need to let each other known if anything is unreadable or worse, and do necessary homework.

andrelkin:
> andrelkin ***@***.***> writes:

To the tech part,


``` +    if (cache_mngr->trx_cache.has_incident() &&
    +        thd->lex->sql_command != SQLCOM_CREATE_TABLE)
    +      error= mysql_bin_log.write_incident(thd);
```

> How does the incident event get set in the first place?

E.g CREATE ... SELECT f(), where has a side effect of modifying a transactional table.
At time of f() to my checking there's no way to find out what is the top-level statement.
If such method existed then of course the Incident error could be prevented that early.

> it could simply be cleared

Again that can be done only at the top level.

With your clearance suggestion we might be forced to deploy a number of checks of a possible Incident error
once nested routines execution returns to the top-level. Remotely something like this we do
with regard to `stmt_modified_non_trans_table`  (when the top level merges leaves' values).

To my judgement, the ignore policy of the current patch is the only feasible option.
Apparently it's correct and I don't see a  practical reason to have such a purist stance
in this case.
After all, in the patch it is the binlog actions time, where we compute a number of various
decisions **pertinent** to binlogging, also with analyzing `lex` environment.
The above is just yet another one.


> why you need to extend the DBUG_ASSERT condition here

It's extended 'cos C-T-S statement gets reset but before that it's
`stmt.modified_non_trans_table` is propagated into `all. ...`.
In BASE (of the patch) execution could not reach this point.

> I do not understand "fixate" ... in
```
+# fixate the current (write) binlog position
+--let $binlog_file_0= query_get_value(SHOW MASTER STATUS, File, 1)
+--let $binlog_start_0 = query_get_value(SHOW MASTER STATUS, Position, 1)
```
A dictionary reads: make fixed, stable or stationary.

I see uses of an expression like 'fixate the value of' a variable in math related contexts.
I would beg for help from @bnestere  to cut this tie.

I am fine with the rest of the notes to be addressed in next commit.

bnestere:
Specifically to the fixate question, though I could extrapolate what you meant, it reads as if you are making the binlog write position unchangeable for the test using those commands. I think it would be more accurate to say that you are saving the current binlog coordinates to ensure the binlog doesn't change after executing the following DDLs which err-to-rollback (or something of the like, per my understanding of an earlier read-through of the patch).

knielsen:

> ``` +    if (cache_mngr->trx_cache.has_incident() &&
>     +        thd->lex->sql_command != SQLCOM_CREATE_TABLE)
>     +      error= mysql_bin_log.write_incident(thd);
> ```

>> it could simply be cleared
>
> Again that can be done only at the top level.
>
> With your clearance suggestion we might be forced to deploy a number
> of checks of a possible Incident error

I think it can simply be cleared.
This patch makes your test case pass, simply by clearing the incident in the
select_create::abort_result_set() function instead of the above special
case:

-----------------------------------------------------------------------
diff --git a/sql/log.cc b/sql/log.cc
index 4747e3b4fc0..7ad2b9129b2 100644
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -320,6 +320,11 @@ class binlog_cache_data
     incident= TRUE;
   }
   
+  void clear_incident(void)
+  {
+    incident= FALSE;
+  }
+
   bool has_incident(void)
   {
     return(incident);
@@ -1957,13 +1962,9 @@ binlog_truncate_trx_cache(THD *thd, binlog_cache_mngr *ca
che_mngr, bool all)
   */
   if (ending_trans(thd, all))
   {
-    DBUG_ASSERT(!cache_mngr->trx_cache.has_incident() ||
-               thd->lex->sql_command != SQLCOM_CREATE_TABLE ||
-               thd->is_current_stmt_binlog_format_row());
-
-    if (cache_mngr->trx_cache.has_incident() &&
-       thd->lex->sql_command != SQLCOM_CREATE_TABLE)
+    if (cache_mngr->trx_cache.has_incident())
       error= mysql_bin_log.write_incident(thd);
+
     thd->reset_binlog_for_next_statement();
 
     cache_mngr->reset(false, true);
@@ -2361,6 +2362,18 @@ void binlog_reset_cache(THD *thd)
 }
 
 
+void binlog_clear_incident(THD *thd)
+{
+  binlog_cache_mngr *const cache_mngr= opt_bin_log ? 
+    (binlog_cache_mngr*) thd_get_ha_data(thd, binlog_hton) : 0;
+  if (cache_mngr)
+  {
+    cache_mngr->stmt_cache.clear_incident();
+    cache_mngr->trx_cache.clear_incident();
+  }
+}
+
+
 void MYSQL_BIN_LOG::set_write_error(THD *thd, bool is_transactional)
 {
   DBUG_ENTER("MYSQL_BIN_LOG::set_write_error");
diff --git a/sql/log.h b/sql/log.h
index 130b701378d..739f2fdbe32 100644
--- a/sql/log.h
+++ b/sql/log.h
@@ -1186,6 +1186,7 @@ File open_binlog(IO_CACHE *log, const char *log_file_name,
 
 void make_default_log_name(char **out, const char* log_ext, bool once);
 void binlog_reset_cache(THD *thd);
+void binlog_clear_incident(THD *thd);
 bool write_annotated_row(THD *thd);
 
 extern MYSQL_PLUGIN_IMPORT MYSQL_BIN_LOG mysql_bin_log;
diff --git a/sql/sql_insert.cc b/sql/sql_insert.cc
index 70894c06c7f..7dc931f5418 100644
--- a/sql/sql_insert.cc
+++ b/sql/sql_insert.cc
@@ -5291,8 +5291,13 @@ void select_create::abort_result_set()
 
   /* possible error of writing binary log is ignored deliberately */
   (void) thd->binlog_flush_pending_rows_event(TRUE, TRUE);
+  /*
+    In the error case, we remove any partially created table. So clear any
+    incident event generates due to cache error, as it no longer applies.
+  */
+  binlog_clear_incident(thd);
 
-  bool has_drop_table_logged= false;
+  bool drop_table_was_logged= false;
   if (table)
   {
     bool tmp_table= table->s->tmp_table;
@@ -5339,7 +5344,7 @@ void select_create::abort_result_set()
                          create_info->db_type == partition_hton,
                          &create_info->tabledef_version,
                          tmp_table);
-          has_drop_table_logged= true;
+          drop_table_was_logged= true;
           debug_crash_here("ddl_log_create_after_binlog");
           thd->binlog_xid= 0;
         }
@@ -5364,7 +5369,7 @@ void select_create::abort_result_set()
 
   if (create_info->table_was_deleted)
   {
-    if (has_drop_table_logged)
+    if (drop_table_was_logged)
     {
       /* for DROP binlogging the error status has to be canceled first */
       Diagnostics_area new_stmt_da(thd->query_id, false, true);
-----------------------------------------------------------------------
(I also renamed your has_drop_table_logged) to be more consistent with the
similar existing flag table_creation_was_logged).

This way, the incident event handling happens at a natural place in the
code, where the specific case is handled. It is very hard to understand how
it could be correct to ignore any incident event for _all_ CREATE TABLE
statements, but not for any other statements, as in your original patch.

> To my judgement, the ignore policy of the current patch is the only feasible option.

Wouldn't the above patch solve the problem as well?

> Apparently it's correct and I don't see a  practical reason to have such a purist stance
> in this case.

But "apparently it is correct" is not a valid reason to approve a patch!

It is either "I have analyzed the code and determined that the patch is
correct because so-and-so" (now I, as the reviewer, had to spend many hours
to analyze it by myself :-/ ).

Or else it is "it passes the test, but I have not yet had the time to fully
analyze and understand the code to determine if it is the correct solution.
That is fair, but we don't push changes to releases until we fully
understand them!


BTW, while analysing, I found there is still a problem here, with this small
change to your test case it asserts because there is still stuff left in the
binlog caches at connection close:

-----------------------------------------------------------------------
diff --git a/mysql-test/suite/rpl/t/rpl_create_select_row.test b/mysql-test/suit
e/rpl/t/rpl_create_select_row.test
index 62d946da79d..0e11def558c 100644
--- a/mysql-test/suite/rpl/t/rpl_create_select_row.test
+++ b/mysql-test/suite/rpl/t/rpl_create_select_row.test
@@ -21,10 +21,14 @@ connect (conn_err,localhost,root,,);
 call mtr.add_suppression("Multi-statement transaction required more than 'max_b
inlog_cache_size' bytes of storage");
 --error ER_TRANS_CACHE_FULL
 create table t engine=myisam select repeat ('a',4096*3) AS a;
+--disconnect conn_err
 
+connect (conn_err,localhost,root,,);
 --error ER_TRANS_CACHE_FULL
 create table t engine=innodb select repeat ('a',4096*3) AS a;
+--disconnect conn_err
 
+connect (conn_err,localhost,root,,);
 --error ER_DUP_ENTRY
 create table t (a int unique, b char) select 1 AS a, 'b' as b union select 1 as
 a, 'c' as b;
 --error ER_NO_SUCH_TABLE
-----------------------------------------------------------------------

So it seems clear that the error handling around binlog cache write errors
is still lacking, but perhaps this is a separate issue from this particular
patch of yours.


> I hoped C-T-S would read ambiguously when it is put after CREATE-TABLE-SELECT.
> Did it do btw to you?

No it did not, I had to think about it and guess.

The point is, I found the description _very_ hard to understand, to the
point that it was almost impossible. The reason is that the problem you are
solving has so many details that are very complex, and when too many
statements are ambigous, cryptic, and require guessing at their meaning, it
becomes too hard to understand.

I persevere for your patches, because I know that you have a gift for
understanding the complexities of replication problems, and there is usually
a valid reasoning behind. That is why I tried to point out all the places
where I had a hard time understanding - to try and explain how you can help
the reviewer to better understand your work, whether it is me reviewing or
someone else.

I had to spend many hours on this relatively small patch to understand the
problem, probably repeating a lot of the analysis you already did. It
shouldn't really be like that to review.

I hope that makes sense, it is getting late here, but my comments really are
an attempt to be constructive to improve further reviews, not to pick on
specific language in any particular sentense.

 - Kristian.

knielsen:
I pushed the above referenced patch to branch knielsen_mdev35207_review

andrelkin:
@knielsen, 
> I think it can simply be cleared.

I have to agree. Thanks for dwelling into this subject. Indeed I overlooked/underestimated `select_create::abort_result_set()` as a  top-level that has been always for this sort purpose.
Many thanks! (And I owe you also time you spent to write  the code part while I would have been fully content with a reference to the function).

> But "apparently it is correct" is not a valid reason to approve a patch!
It was not assertive enough, I agree too. It does not make much of sense to provide a math proof now, but
I'll be remembering doing any next time.

Once more I highly appreciate your effort to improve on this matter and do regret it caused more work on your part
than it might have been. This lesson is learned, we'll be having better times moving forward!

andrelkin:
> One suggestion could be to add assertions that the stmt cache is non-empty in the has_drop_table_logged case

The idea is clear and reasonable. Yet let me skip it for now as it's not really as simple as it looks.
Let me delve into that specifically.
First, it's not the stmt cache as it's the trx cache, surprisingly for me too.
I believe you'd agree that the resultfulness of `DROP` logging (which is the quoted part of the suggestion is about) is better off be checked at the nd/inside-of [`log_drop_table`](https://github.com/MariaDB/server/blob/bb-10.6-MDEV-35207/sql/sql_insert.cc#L5337).
And similarly to the opposite case of  `DROP` need not to be logged, it should be (should've been) done in unimplemented [`if (table_creation_was_logged) ... else` ](https://github.com/MariaDB/server/blob/10.6-MDEV-35207/sql/sql_insert.cc#L5331) branch.
And what makes this matter more "funny" is the fact of no result checking from `log_drop_table()`. Unfortunately that's all the way this function is invoked in two more contexts which should apologize  me merely report another [MDEV-35918](https://jira.mariadb.org/browse/MDEV-35918).

andrelkin:
> [@knielsen]: BTW, while analysing, I found there is still a problem here, with this small
change to your test case it asserts .. but perhaps this is a separate issue

And indeed [MDEV-35506](https://jira.mariadb.org/browse/MDEV-35506)! The p.1 of the description.

andrelkin:
@knielsen, the review notes are addressed in the updated branch. It is also extended to 4 commits, please read on the purpose of each in the Description.
Note the 3rd commit refers to MDEV-35506 whose branch is based on the current one and being reviewed as PR-3665.
Without it MDEV-35506 tests showed the Incident event binlogged. The 3rd commit's message also explains why it was for no reason.

knielsen (approved):
Thanks, Andrei!

mysql-test/suite/rpl/r/rpl_create_select_row.result
```diff
@@ -0,0 +1,158 @@
...
+insert into ti_pk set a=arg;
+return 1;
+end |
+set statement binlog_format = ROW for create table t_y (a int) engine=aria select f_ia(1 /* err */) as a;
+ERROR 23000: Duplicate entry '1' for key 'PRIMARY'
+select * from t_y;
+ERROR 42S02: Table 'test.t_y' doesn't exist
+# correct execution: `ta` is modified and its new record is binlogged
+include/show_binlog_events.inc
+Log_name	Pos	Event_type	Server_id	End_log_pos	Info
```
bnestere, line 121: I'm a bit confused here, this looks like correct behavior to me, but the JIRA description in MDEV-36027 says this is a bug. That is, it looks like all data-modifying information is binlogged. The `create table t_y...` isn't binlogged, but that seems correct, as the query errors and the table creation is rolled back. The updates to `ti_pk` are rolled back, and thereby not binlogged. The only thing binlogged is the update to `ta`, which is the only thing that changed data (by my understanding).

mysql-test/suite/rpl/t/rpl_create_select_row.test
```diff
@@ -0,0 +1,161 @@
...
+--sync_slave_with_master
+--let  $diff_tables=master:ta,slave:ta
+--source include/diff_tables.inc
+
+--connection master
+delete from ta;
+--sync_slave_with_master
+
+--connection master
+# MDEV-36027 Errored-out CREATE-SELECT does not binlog results of any function modifying non-transactional table
```
bnestere, line 132: Here, I'd have the test actually prove the correct behavior, but fail, and just comment it out and prior to that, echo some TODO mentioning MDEV-26027. Then in the MDEV-36027 JIRA description, mention the test already exists in the code base and just needs to be uncommented. IMO, that'd make it easier to resume this work to immediately see the failure.

bnestere (approved):
Thanks @andrelkin ! I left a few minor notes, and have one general request as well. Can you add a new paragraph at the start of the git commit message which describes the broken behavior(s) that users would see? E.g, When a `CREATE TABLE .. SELECT` errors while inserting data, a user would expect that all changes are rolled back and the table would not exist after executing the query; however, the error is ignored in the code, and the table would still exist (and in what state, etc).

sql/sql_insert.cc
```diff
@@ -4264,7 +4264,11 @@ bool select_insert::store_values(List<Item> &values)
 bool select_insert::prepare_eof()
 {
   int error;
-  bool const trans_table= table->file->has_transactions_and_rollback();
+  // make sure any ROW format pending event is logged in the same binlog cache
+  bool const trans_table= (thd->is_current_stmt_binlog_format_row() > 0 &&
```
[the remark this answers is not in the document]
andrelkin: Indeed your comment actually exposes the function's rather sloppy `int` type return while defacto it returns a bool value
[current_stmt_binlog_format == BINLOG_FORMAT_ROW](https://github.com/MariaDB/server/blob/10.6/sql/sql_class.h#L3107).

I'd rather to cast implicitly that is to remove `> 0` as it's the case everywhere in the sources and change the func's signature one day in a refactoring commit.

mysql-test/suite/rpl/r/rpl_create_select_row.result
andrelkin: In case `--error ER_DUP_ENTRY` MDEV-36027 requires the statement format (as the description exemplifies). Here it is ROW.

bnestere: Hmm, but we set `set statement binlog_format = ROW`. And with this configuration, don't we re-write the DDL to exclude the `SELECT` component from the user input `CREATE TABLE ... SELECT` query at binlogging, and then provide the data changes as row events? And thus, when the query fails and everything that can be rolled back, is, and the items that are permanent are logged in `ROW` format.

andrelkin: Just in case, I should've said ^ that YES at this point the execution is correct.

Now to your followup, I don't see any disagreement with what I tried to convey..

In the ROW the `ER_DUP_ENTRY` failed **at "selecting"** `C..S` always knows how to clean up.
So we have only those
>  the items that are permanent are logged in ROW format
```include/show_binlog_events.inc
Log_name	Pos	Event_type	Server_id	End_log_pos	Info
master-bin.000001	#	Gtid	#	#	BEGIN GTID #-#-#
master-bin.000001	#	Table_map	#	#	table_id: # (test.ta)
master-bin.000001	#	Write_rows_v1	#	#	table_id: # flags: STMT_END_F
```

mysql-test/suite/rpl/t/rpl_create_select_row.test
bnestere: >  I believe it's better to freeze the current buggy status,

Hmm, interesting idea. I could see how that could give us more insight into ripple effects of various patches, and ensure existing bugs don't get worse.

My first thought is that it disguises broken behavior as something that is good. But perhaps if we had a clear convention/policy on how writing acceptance tests for bugs (e.g., a special prefix in the filename or something), it could make that easier.

It also leaves additional work to re-write the test when fixing the issue, to differentiate between the good and bad pieces of the test (which, if someone else is fixing the issue, leaves more room for mistakes, e.g. through misinterpretation of the test). I'd find it better to have one clear workflow that highlights broken behavior, such that when the behavior is fixed, the test wouldn't need to be re-written. But perhaps we could combine the ideas, and have both, where the test would freeze the existing behavior, and have comments/commented code on modifying the test to create the expected behavior.


> Would you agree?

Yeah I think so, though I think we should make it more clear that this is currently bugged behavior, have a clearer explanation of the expected behavior once the bug is fixed, and ideally some commented code that can be un-commented to "swap-out" this test case with one that validates correct behavior.

andrelkin:
@knielsen, the review notes are addressed in the updated branch. It is also extended to 4 commits, please read on the purpose of each in the Description.
Note the 3rd commit refers to MDEV-35506 whose branch is based on the current one and being reviewed as PR-3665.
Without it MDEV-35506 tests showed the Incident event binlogged. The 3rd commit's message also explains why it was for no reason.

PR #3665, opened 2024-11-26
MDEV-35506 commit policy of one-phase-commit even at errored-out binlogging leads to assert

andrelkin:

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check


sql/handler.cc
```diff
@@ -1910,6 +1910,8 @@ int ha_commit_trans(THD *thd, bool all)
     if (run_wsrep_hooks && !error)
       error= wsrep_after_commit(thd, all) ? 2 : 0;
 #endif /* WITH_WSREP */
+    if (error)
+      goto err;
```
vuvova, line 1914: This seems to be wrong, a failure of `wsrep_after_commit()` does not mean a failure of the transaction commit.

sql/handler.cc
```diff
@@ -1942,6 +1944,8 @@ int ha_commit_trans(THD *thd, bool all)
   if (!is_real_trans)
   {
     error= commit_one_phase_2(thd, all, trans, is_real_trans);
+    if (error)
+      goto err;
```
vuvova, line 1948: why is this? this isn't a transaction commit at all, `is_real_trans` is false

sql/handler.cc
```diff
@@ -2145,14 +2149,16 @@ commit_one_phase_2(THD *thd, bool all, THD_TRANS *trans, bool is_real_trans)
...
+    int err= 0;
 
     if (has_binlog_hton(ha_info) &&
         (err= binlog_commit(thd, all,
                             is_ro_1pc_trans(thd, ha_info, all, is_real_trans))))
     {
       my_error(ER_ERROR_DURING_COMMIT, MYF(0), err);
       error= 1;
+
+      goto err;
```
vuvova, line 2161: okay, this actually makes sense.

Although I'm not sure on what code path we can get here (in `commit_one_phase_2()` with `has_binlog_hton()` being non-zero and `binlog_commit()` being non-trivial)

But let's have it just in case, sure

vuvova (changes requested):

sql/handler.cc
andrelkin: Indeed. It does not look to relate to the current fixes.
Probably I aimed at uniformity with the above `if (error) goto err`.
Although it makes sense in this context to prefer the new `goto err`  over the current `goto done` I would not rush with the change.

Let me drop the hunk.

sql/handler.cc
andrelkin: >  what code path we can get here

[binlog_commit_fail.test](https://github.com/MariaDB/server/blob/bb-10.6-MDEV-35506/mysql-test/suite/binlog/t/binlog_commit_fail.test#L73`) provides one. And [also](https://github.com/MariaDB/server/blob/bb-10.6-MDEV-35506/mysql-test/suite/binlog/t/binlog_commit_fail.test#L60).

PR #3666, opened 2024-11-27
MDEV-35512 remove all RPM compat packages

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35512*

## Description

The end of life of SLES 12 combined with the
EOL for Centos/RHEL 7 marks a happy time in
our distribution maintenance where we can drop
all RPM compat packages.

They where only needed for distros that
we released at the time of MariaDB-10.1.

All those have finally gone EOL. Hooray!

## Release Notes

There are no more MariaDB-compat packages. Applications that linked to libmysqlclient.so.18.0.0 should have been updated by virtue of obsolescence.

## How can this PR be tested?

Good logic on why they existing to start with.

We shouldn't have been releasing them for RHEL8. And nothing else was carrying them.

## PR quality check


vuvova (approved):

vuvova:
actually, code-wise it's, obviously, fine. But let's wait a bit with this. Users could have a closed-source app linked against an old MySQL/MariaDB version. And 10.5 was GA for 4 years, may be it's not a good idea to start removing packages there

cvicentiu:
@vuvova So then let's target this change to 10.6? That way we can close the issue.

vuvova:
we will close this issue. Eventually. See MDEV

vuvova (approved):

PR #3668, opened 2024-11-27
MDEV-35508 Race condition between purge and secondary index INSERT or UPDATE

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35508*
## Description
`row_purge_remove_sec_if_poss_leaf()`: If the `PAGE_MAX_TRX_ID` belongs to an active transaction, return the impossible value `~0` so that `row_purge_remove_sec_if_poss_tree()` is guaranteed to recheck if the record needs to be purged. It could be the case that the active transaction would insert this record between the time this check completed and `row_purge_remove_sec_if_poss_tree()` acquired a latch on the secondary index leaf page again.

This fixes a regression that had been introduced in #3448.

## Release Notes
InnoDB could issue bogus messages into the error log:
```
[ERROR] InnoDB: tried to purge non-delete-marked record in index …
```
## How can this PR be tested?
I tested this on `CMAKE_BUILD_TYPE=RelWithDebInfo` and `CMAKE_BUILD_TYPE=Debug` builds by repeatedly running the following test case:
```sql

CREATE TABLE t(a INT PRIMARY KEY, b INT NOT NULL, INDEX(b)) ENGINE=InnoDB;

INSERT INTO t SELECT seq, seq from seq_1_to_10000;

--disable_query_log
let $N=30;
while ($N)
{
UPDATE t SET b=a+1; UPDATE t SET b=a;
dec $N;
}
--enable_query_log

SET GLOBAL innodb_max_purge_lag_wait=0;
DROP TABLE t;
```
Notably, the RelWithDebInfo build would complete more than 20 times faster than the  Debug build. I think that a test like this is too slow to be included in the regression test suite. Using `DEBUG_SYNC` is not an option either, because it can’t easily be used with the purge tasks.
## PR quality check

### DIFF
--- a/storage/innobase/include/trx0trx.h
+++ b/storage/innobase/include/trx0trx.h
@@ -598,10 +598,25 @@ public:
   Cleared in commit_in_memory() after commit_state(),
   trx_sys_t::deregister_rw(), release_locks(). */
   trx_id_t id;
-  /** The largest encountered transaction identifier for which no
-  transaction was observed to be active. This is a cache to speed up
-  trx_sys_t::find_same_or_older(). */
-  trx_id_t max_inactive_id;
+  union
+  {
+    /** The largest encountered transaction identifier for which no
+    transaction was observed to be active. This is a cache to speed up
+    trx_sys_t::find_same_or_older().
+
+    This will be zero-initialized in Pool::Pool() and not initialized
+    when a transaction object in the pool is freed and reused. The
+    idea is that new transactions can reuse the result of
+    an expensive trx_sys_t::find_same_or_older_low() invocation that
+    was performed in an earlier transaction that used the same
+    memory area. */
+    trx_id_t max_inactive_id;
+    /** Same as max_inactive_id, for purge_sys.query->trx which may be
+    accessed by multiple concurrent threads in in
+    trx_sys_t::find_same_or_older_in_purge(). Writes are protected by
+    trx_t::mutex. */
+    Atomic_relaxed<trx_id_t> max_inactive_id_atomic;
+  };
 
 private:
   /** mutex protecting state and some of lock
--- a/storage/innobase/row/row0purge.cc
+++ b/storage/innobase/row/row0purge.cc
@@ -746,6 +746,22 @@ bool row_purge_poss_sec(purge_node_t *node, dict_index_t *index,
   return can_delete;
 }
 
+/** Report an error about not delete-marked secondary index record
+that was about to be purged.
+@param cur   cursor on the secondary index record
+@param entry search key */
+ATTRIBUTE_COLD ATTRIBUTE_NOINLINE
+static void row_purge_del_mark_error(const btr_cur_t &cursor,
+                                     const dtuple_t &entry)
+{
+  const dict_index_t *index= cursor.index();
+  ib::error() << "tried to purge non-delete-marked record in index "
+              << index->name << " of table " << index->table->name
+              << ": tuple: " << entry
+              << ", record: " << rec_index_print(cursor.page_cur.rec, index);
+  ut_ad(0);
+}
+
 __attribute__((nonnull, warn_unused_result))
 /** Remove a secondary index entry if possible, by modifying the index tree.
 @param node             purge node
@@ -765,6 +781,16 @@ static bool row_purge_remove_sec_if_poss_tree(purge_node_t *node,
 	mtr_t			mtr;
 
 	log_free_check();
+#ifdef ENABLED_DEBUG_SYNC
+	DBUG_EXECUTE_IF("enable_row_purge_sec_tree_sync",
+		debug_sync_set_action(current_thd, STRING_WITH_LEN(
+			"now SIGNAL "
+			"purge_sec_tree_begin"));
+		debug_sync_set_action(current_thd, STRING_WITH_LEN(
+			"now WAIT_FOR "
+			"purge_sec_tree_execute"));
+	);
+#endif
 	mtr.start();
 	index->set_modified(mtr);
 	pcur.btr_cur.page_cur.index = index;
@@ -812,26 +838,13 @@ found:
 
 		/* Remove the index record, which should have been
 		marked for deletion. */
-		if (!rec_get_deleted_flag(btr_cur_get_rec(
-						btr_pcur_get_btr_cur(&pcur)),
-					  dict_table_is_comp(index->table))) {
-			ib::error()
-				<< "tried to purge non-delete-marked record"
-				" in index " << index->name
-				<< " of table " << index->table->name
-				<< ": tuple: " << *entry
-				<< ", record: " << rec_index_print(
-					btr_cur_get_rec(
-						btr_pcur_get_btr_cur(&pcur)),
-					index);
-
-			ut_ad(0);
-
+		if (!rec_get_deleted_flag(btr_pcur_get_rec(&pcur),
+					  index->table->not_redundant())) {
+			row_purge_del_mark_error(pcur.btr_cur, *entry);
 			goto func_exit;
 		}
 
-		btr_cur_pessimistic_delete(&err, FALSE,
-					   btr_pcur_get_btr_cur(&pcur),
+		btr_cur_pessimistic_delete(&err, FALSE, &pcur.btr_cur,
 					   0, false, &mtr);
 		switch (UNIV_EXPECT(err, DB_SUCCESS)) {
 		case DB_SUCCESS:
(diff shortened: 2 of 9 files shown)


storage/innobase/row/row0purge.cc
```diff
@@ -943,10 +944,23 @@ static trx_id_t row_purge_remove_sec_if_poss_leaf(purge_node_t *node,
...
 
 			if (btr_cur_optimistic_delete(btr_cur, 0, &mtr)
-                            == DB_FAIL) {
+			    == DB_FAIL) {
 				page_max_trx_id = page_get_max_trx_id(
 					btr_cur_get_page(btr_cur));
-                        }
+				if (trx_sys.find(nullptr, page_max_trx_id,
+						 false)) {
+					/* Because this transaction is
```
mariadb-DebarunBanerjee, line 952: This is not enough. We would require to check also for other active transactions having ID smaller than page_max_trx_id.

mariadb-DebarunBanerjee (changes requested):
@dr-m  Thanks for detecting the problem. I think it is a critical one. We could either disable this optimization for now or provide a more comprehensive fix. I have suggested/uploaded a patch in MDEV with a test

storage/innobase/row/row0purge.cc
dr-m: Good point. Such transactions would not update `PAGE_MAX_TRX_ID`.

storage/innobase/row/row0purge.cc
```diff
@@ -943,10 +954,27 @@ static trx_id_t row_purge_remove_sec_if_poss_leaf(purge_node_t *node,
...
+						may modify the secondary index
+						but not PAGE_MAX_TRX_ID,
+						row_purge_poss_sec() must be
+						invoked again after re-latching
+						the index page. Let us return
+						an impossible ID. */
+						page_max_trx_id = ~trx_id_t{0};
+					}
+				}
+			}
```
dr-m, line 977: Come to think of it, all of this code should maybe placed in an `ATTRIBUTE_COLD ATTRIBUTE_NOINLINE static` function. I’ll check the generated code once more.

Moreover, I think that `1` should be a more convenient "impossible" return value, because then the value could be assigned from a register that holds the value of `hot`. Transaction 1 would be the following during InnoDB bootstrap:
```
#0  trx_sys_t::get_new_trx_id_no_refresh (this=0x557174a4c180 <trx_sys>) at /mariadb/10.6/storage/innobase/include/trx0sys.h:1257
#1  trx_sys_t::register_rw (this=0x557174a4c180 <trx_sys>, trx=trx@entry=0x7f0fb0e023c0) at /mariadb/10.6/storage/innobase/include/trx0sys.h:1114
#2  0x000055717316ce15 in trx_assign_rseg_low (trx=trx@entry=0x7f0fb0e023c0) at /mariadb/10.6/storage/innobase/trx/trx0trx.cc:811
#3  0x000055717316e27f in trx_start_low (trx=trx@entry=0x7f0fb0e023c0, read_write=read_write@entry=0x1) at /mariadb/10.6/storage/innobase/trx/trx0trx.cc:953
#4  0x000055717316e988 in trx_start_internal_low (trx=trx@entry=0x7f0fb0e023c0, read_write=read_write@entry=0x1) at /mariadb/10.6/storage/innobase/trx/trx0trx.cc:2263
#5  0x000055717316e9c5 in trx_start_for_ddl_low (trx=trx@entry=0x7f0fb0e023c0) at /mariadb/10.6/storage/innobase/trx/trx0trx.cc:2273
#6  0x00005571732a721c in dict_sys_t::create_or_check_sys_tables (this=0x557174abf100 <dict_sys>) at /mariadb/10.6/storage/innobase/dict/dict0crea.cc:1375
#7  0x000055717310b00b in srv_start (create_new_db=<optimized out>) at /mariadb/10.6/storage/innobase/srv/srv0start.cc:1909
#8  0x0000557172d1ce86 in innodb_init (p=<optimized out>) at /mariadb/10.6/storage/innobase/handler/ha_innodb.cc:4316
```
In MariaDB Server 10.6, this very first transaction with id 1 would write 27 undo log records (all of them `TRX_UNDO_INSERT_REC`, which does not trigger this code). The transaction would not invoke `btr_page_split_and_insert()` either.

storage/innobase/include/trx0trx.h
```diff
@@ -601,7 +601,7 @@ struct trx_t : ilist_node<>
   /** The largest encountered transaction identifier for which no
   transaction was observed to be active. This is a cache to speed up
   trx_sys_t::find_same_or_older(). */
-  trx_id_t max_inactive_id;
+  Atomic_relaxed<trx_id_t> max_inactive_id;
```
mariadb-DebarunBanerjee, line 604: Can you please help me understand how this atomic variable in trx_t is initialized ?

mariadb-DebarunBanerjee (approved):
I am fine with the patch as long as we ensure the atomic variable max_inactive_id in trx_t is initialized correctly. Added one comment.

storage/innobase/include/trx0trx.h
dr-m: It is a bit obfuscated. In `storage/innobase/include/ut0pool.h` you will find that `Pool::Pool()` will invoke `aligned_malloc()` and `memset_aligned` to zero-initialize the memory (4 MiB). There is a partial initialization of some fields in `Factory::init()` a.k.a. `TrxFactory::init()`. The `max_inactive_id` does not need to be one of them; in fact, we actually benefit from having this field cached in a previously used instance of `trx_t`.

It looks like `trx_create()` would dereference a null pointer in case we have more concurrent transactions than `MAX_TRX_BLOCK_SIZE / sizeof(trx_t)`. According to a comment in `Deadlock::find_cycle()`, the maximum number of concurrent transactions is `innodb_page_size / 16 * 128`, which would be at most 524,288 when using `innodb_page_size=64k`. It looks like we’d only have to exceed 381 concurrently active transactions to hit that condition, if `sizeof(trx_t)` is 2752.

mariadb-DebarunBanerjee: ok, memset/zero initialization is good strategy for POD. I think it is safer to initialize atomic variables explicitly. Can we do that in trx_init() ?

dr-m: Yesterday, I missed the fact that `trx_pools` uses one more class template. `trx_create()` is actually invoking `PoolManager::get()`, which may create additional `Pool` or wait for a transaction object to become available.

If we zero-initialized `trx_t::max_inactive_id`, we would reintroduce a performace regression that this field was designed to prevent in 4105017a5832c3486002c4ec0c67005ceb5ab88b ([MDEV-30357](https://jira.mariadb.org/browse/MDEV-30357)).

As far as I understand, [relaxed memory ordering](https://en.cppreference.com/w/cpp/atomic/memory_order#Relaxed_ordering) for loads and stores is [typically implemented](https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html) mostly as a compiler barrier, with no difference whatsoever to non-atomic loads or stores. For anything stronger than `std::memory_order_relaxed`, special instructions are needed, more so on ISA implementations that do not follow Total Store Ordering. On ARM before LSE and on POWER that construct would be a loop around load-acquire/store-conditional. But the relaxed loads and stores are just like regular loads and stores.

Only on 32-bit targets there could be a difference, because we are dealing with a 64-bit quantity.

In fact, it might be better to use an anonymous `union` between the non-atomic field and the `Atomic_relaxed` one in order to avoid introducing a regression on 32-bit systems. We only need the load and the store in `row_purge_check()` to be atomic; `lock_sec_rec_some_has_impl()` had better use regular access, because the `trx_t` there is almost exclusively accessed by the current thread. (The main exception is the lock conversion from implicit to explicit by other transactions that are registering a conflicting lock.)

On 32-bit processors the 64-bit atomic loads and stores are special; using regular instructions could theoretically result in torn reads. For example, if `trx_t::max_inactive_id` was being concurrently modified from 0x1ffffffff to 0x200000001, a concurrent reader might read it as 0x100000001 or 0x2ffffffff. However, I don’t think that this is likely in practice, because `trx_t` in  `trx_pools` is aligned to the cache line size, which is at least 512 bits (64 bytes). Even without that, on 32-bit systems the minimum alignment of structs should be 64 bits, for compatibility with the IEEE 754 `double` precision 64-bit floating point type. Okay, if some code was interrupted between two 32-bit loads and the context switch was long enough, then we could get a torn read.

On IA-32, I would expect a 80387 floating point register to come to the rescue, but on https://godbolt.org I checked that it’s actually using an XMM register of the Pentium MMX vintage. That makes sense, for years already, we depend on `lock cmpxchg8b`, which was introduced in the Pentium. In fact, our "x86" is `-march=i686`. There might not be a 64-bit 80387 load or store either; the register file width is 80 bits.
```c++
#include <atomic>
uint64_t f(std::atomic<uint64_t> &num)
{ return num.load(std::memory_order_relaxed); }
```
On ARMv2 through ARMv7, the atomic load would invoke a runtime library function `__atomic_load_8()`. I tested the above program on an `armhf` target, and it turns out that such a function is not available unless I specify `-latomic`. In `libatomic`. That seems to be a rather complex function, because it is also invoking `std::operator&(std::memory_order, std::__memory_order_modifier)` on an `armhf` system that I checked. On 32-bit systems other than `i686` we might in fact want to just use the `trx_t::mutex` instead of using atomic memory access here.

I will revise this accordingly.

dr-m: 19acb0257eec0956e3a0dd3b11bbb701fa7ea648 adds source code comments about the initialization and the reuse of this cached value across transactions. I checked the generated code on all 32-bit and 64-bit targets I could find on https://godbolt.org. Most 32-bit targets would invoke a library function `__atomic_load_8()` for an atomic 64-bit load. The Super-H would emit a loop around two 32-bit loads, and `--march=i686` would use a 64-bit load via an XMM register instead of two regular 32-bit loads. On all 64-bit targets, the relaxed and regular loads used the same machine instructions.

PR #3671, opened 2024-12-01
MDEV-34699 - mhnsw: support aarch64 SIMD instructions

svoj:
- [x] *The Jira issue number for this PR is: MDEV-34699*

## Description
SIMD implementations of bloom filters and dot product calculation.

## Release Notes
None.

## How can this PR be tested?
mtr

## PR quality check

### DIFF
--- a/sql/bloom_filters.h
+++ b/sql/bloom_filters.h
@@ -28,9 +28,18 @@ SOFTWARE.
 #include <cmath>
 #include <vector>
 #include <algorithm>
+
+/*
+  Use gcc function multiversioning to optimize for a specific CPU with run-time
+  detection. Works only for x86, for other architectures we provide only one
+  implementation for now.
+*/
+#define DEFAULT_IMPLEMENTATION
+#if __GNUC__ > 7
+#ifdef __x86_64__
 #ifdef HAVE_IMMINTRIN_H
 #include <immintrin.h>
-#if __GNUC__ > 7 && defined __x86_64__
+#undef DEFAULT_IMPLEMENTATION
 #define DEFAULT_IMPLEMENTATION    __attribute__ ((target ("default")))
 #define AVX2_IMPLEMENTATION __attribute__ ((target ("avx2,avx,fma")))
 #if __GNUC__ > 9
@@ -38,8 +47,11 @@ SOFTWARE.
 #endif
 #endif
 #endif
-#ifndef DEFAULT_IMPLEMENTATION
-#define DEFAULT_IMPLEMENTATION
+#ifdef __aarch64__
+#include <arm_neon.h>
+#undef DEFAULT_IMPLEMENTATION
+#define NEON_IMPLEMENTATION
+#endif
 #endif
 
 template <typename T>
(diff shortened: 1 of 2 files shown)


sql/bloom_filters.h
```diff
@@ -28,18 +28,25 @@ SOFTWARE.
...
 #if __GNUC__ > 9
 #define AVX512_IMPLEMENTATION __attribute__ ((target ("avx512f,avx512bw")))
 #endif
 #endif
 #endif
-#ifndef DEFAULT_IMPLEMENTATION
-#define DEFAULT_IMPLEMENTATION
+#ifdef __aarch64__
+#include <arm_neon.h>
+#undef DEFAULT_IMPLEMENTATION
```
vuvova, line 47: I think a default implementation must always exist, that's why it's "default", for a case when CPU doesn't support necessary SIMD commands.

sql/bloom_filters.h
```diff
@@ -177,9 +184,153 @@ struct PatternedSimdBloomFilter
...
+  {
+    static constexpr uint64_t prime_mx2= 0x9FB21C651E98DF25ULL;
+    static constexpr uint64_t bitflip= 0xC73AB174C5ECD5A2ULL;
+    uint64x2_t step1= veorq_u64(vecData, vdupq_n_u64(bitflip));
+    uint64x2_t step2= veorq_u64(vshrq_n_u64(step1, 48), vshlq_n_u64(step1, 16));
+    uint64x2_t step3= veorq_u64(vshrq_n_u64(step1, 24), vshlq_n_u64(step1, 40));
+    uint64x2_t step4= veorq_u64(step1, veorq_u64(step2, step3));
+    uint64x2_t step5;
+    step5= vsetq_lane_u64(vgetq_lane_u64(step4, 0) * prime_mx2, step4, 0);
+    step5= vsetq_lane_u64(vgetq_lane_u64(step4, 1) * prime_mx2, step5, 1);
```
vuvova, line 198: strange. it doesn't have 64-bit vector multiplication at all?

sql/bloom_filters.h
```diff
@@ -177,9 +184,153 @@ struct PatternedSimdBloomFilter
...
+  uint64x2_t ConstructMask(uint64x2_t vecHash)
+  {
+    uint64x2_t vecMaskIdxMask= vdupq_n_u64((1 << mask_idx_bits) - 1);
+    uint64x2_t vecMaskMask= vdupq_n_u64((1ull << bits_per_mask) - 1);
+
+    uint64x2_t vecMaskIdx= vandq_u64(vecHash, vecMaskIdxMask);
+    uint64x2_t vecMaskByteIdx= vshrq_n_u64(vecMaskIdx, 3);
+    int64x2_t vecMaskBitIdx=
+      vsubq_s64(vdupq_n_s64(0),
+                vreinterpretq_s64_u64(vandq_u64(vecMaskIdx, vdupq_n_u64(0x7))));
```
vuvova, line 224: could you explain this one, please?

sql/bloom_filters.h
```diff
@@ -177,9 +184,153 @@ struct PatternedSimdBloomFilter
...
+    uint64x2_t vecRawMasks= vdupq_n_u64(*reinterpret_cast<const uint64_t*>
+                              (masks + vgetq_lane_u64(vecMaskByteIdx, 0)));
+    vecRawMasks= vsetq_lane_u64(*reinterpret_cast<const uint64_t*>
+                   (masks + vgetq_lane_u64(vecMaskByteIdx, 1)), vecRawMasks, 1);
```
vuvova, line 228: hmm, does this neon specific implementation even make sense? only for two 64-bit values at a time, multiplication of array lookups are done per value, not simd. Perhaps a default implementation of the bloom filter will be just as fast? It'd make the code simpler.

sql/vector_mhnsw.cc
```diff
@@ -188,7 +188,56 @@ struct FVector
...
 
+
+/*
+  ARM NEON implementation. A microbenchmark shows 1.7x dot_product() performance
+  improvement compared to regular -O2/-O3 builds and 2.4x compared to builds
+  with auto-vectorization disabled.
+
+  There seem to be no performance difference between vmull+vmull_high and
+  vmull+vmlal2_high implementations. The second implementation is preserved
+  just for reference.
```
vuvova, line 199: let's drop the other implementation :) just keep the comment "There seem to be ... implementation"

sql/vector_mhnsw.cc
```diff
@@ -188,7 +188,56 @@ struct FVector
...
+#if 0
+      d+= vaddlvq_s32(vmlal_high_s16(vmull_s16(vget_low_s16(p1),
+                                               vget_low_s16(p2)), p1, p2));
+#else
+      d+= vaddlvq_s32(vmull_s16(vget_low_s16(p1), vget_low_s16(p2))) +
+          vaddlvq_s32(vmull_high_s16(p1, p2));
+#endif
+      v1+= NEON_dims;
+      v2+= NEON_dims;
+    }
```
vuvova, line 222: did you look at the assembly code for this function? Can you paste it here, please? (not in the code, in the comment on a PR).

sql/bloom_filters.h
svoj: Function multiversioning works only for x86, there's no `default` target on other platforms:
https://gcc.gnu.org/wiki/FunctionMultiVersioning
```
This support is available in GCC 4.8 and later. Support is only available in C++ for i386 targets.
```

It is under `#ifdef __aarch64__`, which guarantees NEON presence. It presumes ARMv8 at least. If we need to support this optimisation on earlier platforms (ARMv7?), it can be tricky.

There are also more advanced SVE/SVE2, but they don't seem to be widely available, e.g. there's none on our bb machine. And if they're available, they don't seem to offer registers larger than NEON does.

sql/bloom_filters.h
svoj: Correct, I couldn't find any. And I don't seem to be alone, google for "neon multiply 64".

sql/bloom_filters.h
svoj: That's easy: it is equivalent to `-(int64_t) (maskIdx & 7)`, almost what default implementation does.

Negative is to turn left shift into right shift (`vshlq_u64(uint64x2_t a, int64x2_t b)`). There's no direct right shift call, it only supports shift by scalar: `vshrq_n_u64(uint64x2_t a, const int n)`.

sql/bloom_filters.h
svoj: I did ask myself the same question in this commit comment:
> Performance improvement (microbenchmark) for bloom filters is less exciting,
> within 10-30% ballpark depending on compiler options and load.

I could try shuffling intrinsics to gain better improvement if you feel it is worthy (considering how BF are used and if they're bottleneck anyway). Generally I tend to agree we should put it on a shelve, unless you see good reason to have it.

sql/vector_mhnsw.cc
svoj: Alright, I'll let the other implementation gather dust as jira comment.

sql/bloom_filters.h
vuvova: add a comment please, something like
```c
/* shift right in NEON is implemented as shift left by a negative value. do the negation here */
```

sql/bloom_filters.h
vuvova: 10-30% isn't that bad and the change is highly local, one header file only. may be it's worth it after all

sql/vector_mhnsw.cc
vuvova: another question, would be good to compare with the code that compiler would generate without explicit use of intrinsics, but with explicit vectorization. That is, with something like
```c
typedef float v128i __attribute__((vector_size(NEONbytes)));
v128i *p1= (v128i*)v1, *p2=(v128i*)v2, d=0;
for (size_t i= 0; i < (len + NEON_dims - 1) / NEON_dims; i++, p1++, p2++)
  d+=*p1 * *p2;
```
This compiles into (https://godbolt.org/z/GjnsqoEhq)
```
dot_product(void const*, void const*, int):
        add     w4, w2, 14
        adds    w3, w2, 7
        csel    w4, w4, w3, mi
        asr     w4, w4, 3
        cmp     w2, 0
        ble     .L4
        movi    v0.4s, 0
        mov     x2, 0
        mov     w3, 0
.L3:
        ldr     q2, [x0, x2]
        add     w3, w3, 1
        ldr     q1, [x1, x2]
        add     x2, x2, 16
        mla     v0.4s, v2.4s, v1.4s
        cmp     w3, w4
        blt     .L3
        dup     s1, v0.s[1]
        add     v0.2s, v1.2s, v0.2s
        fmov    w0, s0
        ret
.L4:
        mov     w0, 0
        ret
```
and your explicit code is
```
dot_product(short const*, short const*, int):
        add     w5, w2, 14
        adds    w3, w2, 7
        csel    w5, w5, w3, mi
        mov     x6, x0
        asr     w5, w5, 3
        cmp     w2, 0
        ble     .L4
        mov     x3, 0
        mov     w4, 0
        mov     w0, 0
.L3:
        ldr     q0, [x6, x3]
        add     w4, w4, 1
        ldr     q2, [x1, x3]
        add     x3, x3, 16
        smull   v1.4s, v0.4h, v2.4h
        smull2  v0.4s, v0.8h, v2.8h
        saddlv  d1, v1.4s
        saddlv  d0, v0.4s
        add     d0, d1, d0
        fmov    x2, d0
        add     w0, w0, w2
        sxth    w0, w0
        cmp     w4, w5
        blt     .L3
        ret
.L4:
        mov     w0, 0
        ret
```
the `#ifdef 0` implementation differs in
```
        smull   v0.4s, v2.4h, v1.4h
        smlal2  v0.4s, v2.8h, v1.8h
        saddlv  d0, v0.4s
        fmov    x2, d0
```

vuvova:
just to clarify, "approved" means (see https://mariadb.com/kb/en/mariadb-quality-development-rules/) that you push into, say, bb-11.8-MDEV-34699-vector-arm and change the MDEV status to in-testing

sql/vector_mhnsw.cc
svoj: Dunno, I tried adding `dot_product()` w/o explicit vectorization and adding `-ftree-vectorize`. It produced something very complex that had vector instructions. Probably it couldn't find a way to get your example vectorized due to widening or whatnot.

svoj:
@vuvova I think I resolved all your requests. I don't think development rules apply to me, since I have no write access (which I'm very happy about in moments like this).

ottok:
Feel free to assign @HugoWenTD as an additional reviewer. It is not possible right now due to permissions, but I believe @cvicentiu  is looking into how to add long-term contributors with minimal permissions into the project so they can contribute reviews with full review tooling.

PR #3676, opened 2024-12-04
MDEV-24035 Failing assertion: UT_LIST_GET_LEN(lock.trx_locks) == 0 causing disruption and replication failure

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-24035*
## Description
The SQL layer may wrongly disregard an invocation of `thd_mark_transaction_to_rollback()` when an InnoDB transaction had been aborted (rolled back) due to one of the following errors:
* `HA_ERR_LOCK_DEADLOCK`
* `HA_ERR_RECORD_CHANGED` (if `innodb_snapshot_isolation=ON`)
* `HA_ERR_LOCK_WAIT_TIMEOUT` (if `innodb_rollback_on_timeout=ON`)

Such an error used to cause a crash of InnoDB during transaction commit. These changes aim to catch and report the error earlier, so that not only this crash can be avoided but also the original root cause be found and fixed more easily.

The idea of this fix as well as the changes to the tests and the code outside InnoDB are from @montywi.

`HA_ERR_ROLLBACK`: A new error code signalling that the current transaction had been aborted.

`trx_t::state`: Add `TRX_STATE_ABORTED` that is like `TRX_STATE_NOT_STARTED`, but noting that the transaction had been rolled back and aborted.

`trx_t::is_started()`: Replaces `trx_is_started()`.

`DB_NO_SAVEPOINT`: Remove. This is now handled directly in `ha_innobase`.

`ha_innobase`: Check the transaction state in various places. Simplify the logic around `SAVEPOINT`.

`ha_innobase::is_valid_trx()`: Replaces `ha_innobase::is_read_only()`.

`ha_innobase::store_lock()`: Remove a redundant assignment of `trx->will_lock` that should be performed in `ha_innobase::external_lock()` already.

`ha_innobase::external_lock()`: Reset `TRX_STATE_ABORTED` to `TRX_STATE_NOT_STARTED`.

`convert_error_code_to_mysql()`: If the transaction was aborted, write to the error log, and set the state to `TRX_STATE_ABORTED`.
## Release Notes
An error code `HA_ERR_ROLLBACK` as well as error messages fo the form
```
[ERROR] InnoDB: Transaction was aborted due to Deadlock
```
were added in order to better catch cases where a transaction had been aborted by the storage engine.
## How can this PR be tested?
Extensive stress testing by RQG is needed. Hopefully some assertions `ut_ad("invalid state" == 0)` will fail, leading us to find some root causes of the original crash.

This bug very likely also affects MariaDB Server 10.5, but it is an extensive change and the majority of users who have encountered this bug should be running 10.6 or later.
## PR quality check


storage/innobase/handler/ha_innodb.cc
```diff
@@ -16078,6 +16147,23 @@ ha_innobase::external_lock(
...
 
+	bool not_started = false;
+	switch (trx->state) {
+	default:
+	case TRX_STATE_PREPARED:
+		ut_ad("invalid state" == 0);
+		DBUG_RETURN(HA_ERR_WRONG_COMMAND);
+	case TRX_STATE_ABORTED:
+		/* Reset the state if the transaction had been aborted. */
+		trx->state = TRX_STATE_NOT_STARTED;
```
dr-m, line 16158: This logic was in the original patch by @montywi. If I understand correctly, this is intended only for `autocommit=1` transactions. In multi-statement transactions, we’d probably better return an error instead.

sql/handler.cc
```diff
@@ -4676,6 +4676,17 @@ void handler::print_error(int error, myf errflag)
...
     my_error(ER_VERS_NOT_ALLOWED, errflag, table->s->db.str, table->s->table_name.str);
     DBUG_VOID_RETURN;
+  case HA_ERR_ROLLBACK:
+    /* Crash if we run with --debug-assert-on-error */
+    DBUG_ASSERT(!debug_assert_if_crashed_table);
+    SET_FATAL_ERROR;
+    my_printf_error(HA_ERR_ROLLBACK,
+                    "Transaction was rolled back internally for table %`s.%`s. "
+                    "This error is wrong non fatal error that should be "
+                    "reported so it can get fixed.",
```
vuvova, line 4686: the text cannot say "non fatal" after `SET_FATAL_ERROR`. In fact, there is already `ER_ROLLBACK_ONLY` which works fine here, so this can be changed to
```c++
case HA_ERR_ROLLBACK:
  SET_FATAL_ERROR;
  textno= ER_ROLLBACK;
  break;
```

storage/innobase/handler/ha_innodb.cc
dr-m: This was addressed already in 41839d2d3d4af71a12336b108fda2f54098828c5.

sql/handler.cc
dr-m: 840d19e4a60edec79c6ab6e7275c8bfedabfc985 addresses this.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -16639,14 +16714,6 @@ ha_innobase::store_lock(
...
-
-	if (!trx_is_started(trx)
-	    && (m_prebuilt->select_lock_type != LOCK_NONE
-	        || m_prebuilt->stored_select_lock_type != LOCK_NONE)) {
-
-		trx->will_lock = true;
-	}
-
 	return(to);
 }
```
dr-m, line 16718: Before I forget: @montywi made a comment about this last week.

I will need to check how `ha_innobase::store_lock()` and `ha_innobase::external_lock()` are being invoked in
```sql
insert into t values(1);
insert into t values(2);
```
Both with `autocommit=0` and `autocommit=1` as well as with explicit `START TRANSACTION` and `COMMIT` and additionally inside `LOCK TABLES`/`UNLOCK TABLES`. Possibly this `trx->will_lock=true` assignment needs to be restored. I had been under the impression that this code is just duplicating something that `ha_innobase::external_lock()` is executing.

dr-m: I tested the following SQL with `./mtr --rr`:
```sql
create table t(a int primary key) engine=innodb;
insert into t values(1);
insert into t values(2);
set autocommit=0;
insert into t values(3);
insert into t values(4);
commit;
lock tables t write;
insert into t values(5);
insert into t values(6);
commit;
unlock tables;
set autocommit=1;
begin;
insert into t values(7);
insert into t values(8);
commit;
lock tables t write;
begin;
insert into t values(9);
insert into t values(10);
commit;
unlock tables;
start transaction with consistent snapshot;
insert into t values(11);
insert into t values(12);
commit;
start transaction with consistent snapshot;
lock tables t write;
insert into t values(13);
insert into t values(14);
commit;
begin;
lock tables t write;
insert into t values(15);
insert into t values(16);
commit;
drop table t;
```
For each `lock tables` and `unlock tables` as well as for each `insert` statement except (5), (6), (13), (14), (15), (16), the function `ha_innobase::store_lock()` will be invoked once, followed by 2 calls of `ha_innobase::external_lock()`, with `F_WRLCK` and `F_UNLCK`. The first `ha_innobase::external_lock()` would execute the following:
```c++
		if (!trx_is_started(trx)
		    && (m_prebuilt->select_lock_type != LOCK_NONE
			|| m_prebuilt->stored_select_lock_type != LOCK_NONE)) {

			trx->will_lock = true;
		}
```
This basically duplicates the code that I would remove from `ha_innobase::store_lock()`. I added watchpoints on `trx->state` as well as `trx->will_lock` to find out if we might neglect setting the latter. The state changes are as follows:
* (1): `trx->will_lock=true` in `ha_innobase::external_lock()`, transaction started with `ha_innobase::write_row()`, committed in `mysql_execute_command()` (`trx_t::commit_cleanup()` will clear `will_lock=false`)
* (2): ditto
* (3): starts like (1), but there is an explicit `commit` after (4)
* (5) and (6): like (3) and (4), but `trx->will_lock=true` and `trx->state=TRX_STATE_ACTIVE` will be set `lock tables t write` by `row_lock_table()`, invoked by `ha_innobase::external_lock(F_WRLCK)`
* (7) and (8): like (3) and (4)
* (9) and (10): like (3) and (4); `will_lock` will be set and cleared during `lock tables`, which will be auto-committed
* (11) and (12): During the execution `START TRANSACTION WITH CONSISTENT SNAPSHOT`, `trx->will_lock = true` will be set by `trx_start_low()`.
* (13) and (14): Like (1) and (2), that is, 2 separate autocommit `insert`! This was a real surprise. The `LOCK TABLE` implicitly commit the `START TRANSACTION WITH CONSISTENT SNAPSHOT` and then starts and auto-commit itself. For each `insert`, the `trx->will_lock` will be set in `ha_innobase::start_stmt()` inside `open_and_lock_tables()`, and the transaction started in `handler::ha_write_row()` and autocommitted in `mysql_execute_command()`, like (1) and (2).
* (15) and (16): Like (13) and (14), but the initial `BEGIN` will not start any transaction inside InnoDB.

In each and every case we do have `trx->will_lock=true` when the transaction is being started. So, the code in `ha_innobase::store_lock()` is redundant.

The only somewhat surprising finding (unrelated to these changes) is that `LOCK TABLES` can basically only work when it is executed before `BEGIN` or `START TRANSACTION`. It could be good to issue a warning (or an error if some particular `sql_mode` is set) when there `LOCK TABLES` is executed inside a transaction.

storage/innobase/fts/fts0fts.cc
```diff
@@ -5645,16 +5618,19 @@ fts_savepoint_rollback(
...
-	if (i != ULINT_UNDEFINED) {
-		fts_savepoint_t*	savepoint;
+	if (i == 0) {
+		/* fts_trx_create() must have been invoked after
+		this savepoint had been created, and we must roll back
+		everything. */
+		i = 1;
+	}
 
-		ut_a(i > 0);
```
dr-m, line 5653: I intend to extend the regression test suite to cover the `i = 1` case:
```diff
diff --git a/mysql-test/suite/innodb_fts/t/savepoint.test b/mysql-test/suite/innodb_fts/t/savepoint.test
index 09ccb383bb9..ae83c23d20c 100644
--- a/mysql-test/suite/innodb_fts/t/savepoint.test
+++ b/mysql-test/suite/innodb_fts/t/savepoint.test
@@ -366,12 +366,28 @@ TRUNCATE TABLE articles;
 
 INSERT INTO articles(id, title) VALUES(1, 'mysql');
 
+CREATE TABLE t(a INT PRIMARY KEY) ENGINE=InnoDB;
 BEGIN;
+INSERT INTO t SET a=1;
+SAVEPOINT t;
+INSERT INTO articles(id, title) VALUES(2, 'mysql');
+SELECT * FROM articles WHERE MATCH(title) AGAINST('mysql');
+ROLLBACK TO SAVEPOINT t;
+SELECT * FROM articles WHERE MATCH(title) AGAINST('mysql');
+SELECT * FROM t;
+COMMIT;
 
+SELECT * FROM articles WHERE MATCH(title) AGAINST('mysql');
+
+BEGIN;
+INSERT INTO t SET a=2;
 INSERT INTO articles(id, title) VALUES(2, 'mysql');
 
 ROLLBACK;
 
+SELECT * FROM t;
+DROP TABLE t;
+
 INSERT INTO articles(id, title) VALUES(3, 'mysql');
 
 SELECT * FROM articles WHERE MATCH(title) AGAINST('mysql');
```

vlad-lesin (approved):
Looks good to me.

PR #3680, opened 2024-12-05
MDEV-35577  online log file resizing corrupts redo log

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35577*
## Description
If InnoDB is killed in such a way that there had been no writes to a newly resized `ib_logfile101` after it replaced `ib_logfile0` in `log_t::write_checkpoint()`, it is possible that recovery will accidentally interpret some garbage at the end of the log as valid.

`log_t::write_buf()`: To prevent the corruption, write an extra NUL byte at the end of `log_sys.resize_buf`, like we always did for the main `log_sys.buf`. To remove some conditional branches from a time critical code path, we instantiate a separate template for the rare case that the log is being resized. Define as `__attribute__((always_inline))` so that this will be inlined also in the rare case the log is being resized.

`log_t::writer`: Pointer to the current implementation of `log_write_up_to()`. For quick access, this is located in the same cache line with `log_sys.latch`.

`log_writer()`: Replaces `log_write_up_to()`.

`log_t::write_update()`: Update the `log_write_up_to()` implementation. We have a separate implementations for memory-mapped writes and for log resizing.

`log_t::resize_write_buf()`: Remove `ATTRIBUTE_NOINLINE ATTRIBUTE_COLD`. Now that `log_writer()` will be instantiated separately for the rare case of log resizing being in progress, there is no need to forbid this code from being inlined.
## How can this PR be tested?
```sh
```
when the log is *not* memory-mapped, that is, `cmake -DWITH_INNODB_PMEM=OFF` or `mysql-test/var` does not point to `/dev/shm` or a `mount -o dax` file system.

## PR quality check

### DIFF
--- a/storage/innobase/include/log0log.h
+++ b/storage/innobase/include/log0log.h
@@ -60,7 +60,7 @@ wait and check if an already running write is covering the request.
 @param durable  whether the write needs to be durable
 @param callback log write completion callback */
 void log_write_up_to(lsn_t lsn, bool durable,
-                     const completion_callback *callback= nullptr);
+                     const completion_callback *callback= nullptr) noexcept;
 
 /** Write to the log file up to the last log entry.
 @param durable  whether to wait for a durable write to complete */
@@ -249,6 +249,8 @@ public:
 
   /** latest completed checkpoint (protected by latch.wr_lock()) */
   Atomic_relaxed<lsn_t> last_checkpoint_lsn;
+  /** The log writer (protected by latch.wr_lock()) */
+  lsn_t (*writer)() noexcept;
   /** next checkpoint LSN (protected by latch.wr_lock()) */
   lsn_t next_checkpoint_lsn;
 
@@ -270,7 +272,8 @@ private:
   @return the value of buf_free */
   size_t lock_lsn() noexcept;
 
-  /** log sequence number when log resizing was initiated, or 0 */
+  /** log sequence number when log resizing was initiated;
+  0 if the log is not being resized, 1 if resize_start() is in progress */
   std::atomic<lsn_t> resize_lsn;
   /** the log sequence number at the start of the log file */
   lsn_t first_lsn;
@@ -358,7 +361,8 @@ public:
   inline lsn_t get_write_target() const;
 
   /** @return LSN at which log resizing was started and is still in progress
-      @retval 0 if no log resizing is in progress */
+      @retval 0 if no log resizing is in progress
+      @retval 1 if resize_start() is in progress */
   lsn_t resize_in_progress() const noexcept
   { return resize_lsn.load(std::memory_order_relaxed); }
 
@@ -387,7 +391,6 @@ private:
   /** Write resize_buf to resize_log.
   @param b       resize_buf or resize_flush_buf
   @param length  the used length of b */
-  ATTRIBUTE_COLD ATTRIBUTE_NOINLINE
   void resize_write_buf(const byte *b, size_t length) noexcept;
 public:
 
@@ -489,6 +492,9 @@ public:
 #endif
 
 private:
+  /** Update writer and mtr_t::finisher */
+  void writer_update() noexcept;
+
   /** Wait in append_prepare() for buffer to become available
   @tparam spin  whether to use the spin-only lock_lsn()
   @param b      the value of buf_free
@@ -551,10 +557,20 @@ public:
   @param end_lsn    start LSN of the FILE_CHECKPOINT mini-transaction */
   inline void write_checkpoint(lsn_t end_lsn) noexcept;
 
-  /** Write buf to ib_logfile0.
-  @tparam release_latch whether to invoke latch.wr_unlock()
+  /** Variations of write_buf() */
+  enum resizing_and_latch {
+    /** skip latch.wr_unlock(); log resizing may or may not be in progress */
+    RETAIN_LATCH,
+    /** invoke latch.wr_unlock(); !(resize_in_progress() > 1) */
+    NOT_RESIZING,
+    /** invoke latch.wr_unlock(); resize_in_progress() > 1 */
+    RESIZING
+  };
+
+  /** Write buf to ib_logfile0 and possibly ib_logfile101.
+  @tparam resizing whether to release latch and whether resize_in_progress()>1
   @return the current log sequence number */
-  template<bool release_latch> inline lsn_t write_buf() noexcept;
+  template<resizing_and_latch resizing> inline lsn_t write_buf() noexcept;
 
   /** Create the log. */
   void create(lsn_t lsn) noexcept;
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -101,6 +101,7 @@ void log_t::create()
   ut_ad(!checkpoint_buf);
   ut_ad(!buf);
   ut_ad(!flush_buf);
+  ut_ad(!writer);
   max_buf_free= 1;
 
   latch.SRW_LOCK_INIT(log_latch_key);
@@ -330,6 +331,7 @@ bool log_t::attach(log_file_t file, os_offset_t size)
 
   ut_ad(!buf);
   ut_ad(!flush_buf);
+  ut_ad(!writer);
 #ifdef HAVE_INNODB_MMAP
   if (size)
   {
@@ -352,7 +354,7 @@ bool log_t::attach(log_file_t file, os_offset_t size)
 # endif
       buf= static_cast<byte*>(ptr);
       max_buf_free= 1;
-      mtr_t::finisher_update();
+      writer_update();
 # ifdef HAVE_PMEM
       if (is_pmem)
         return true;
@@ -395,7 +397,7 @@ bool log_t::attach(log_file_t file, os_offset_t size)
   TRASH_ALLOC(buf, buf_size);
   TRASH_ALLOC(flush_buf, buf_size);
   max_buf_free= buf_size / LOG_BUF_FLUSH_RATIO - LOG_BUF_FLUSH_MARGIN;
-  mtr_t::finisher_update();
+  writer_update();
   memset_aligned<512>(checkpoint_buf, 0, write_size);
 
 #ifdef HAVE_INNODB_MMAP
@@ -507,6 +509,8 @@ void log_t::close_file()
     aligned_free(checkpoint_buf);
     checkpoint_buf= nullptr;
   }
+
+  writer= nullptr;
 
 #ifdef HAVE_INNODB_MMAP
   if (really_close)
@@ -671,6 +675,8 @@ log_t::resize_start_status log_t::resize_start(os_offset_t size) noexcept
              (lsn_t{write_size - 1} + start_lsn - first_lsn));
         else if (!is_opened())
           resize_log.close();
+
+        writer_update();
       }
       resize_lsn.store(start_lsn, std::memory_order_relaxed);
       status= success ? RESIZE_STARTED : RESIZE_FAILED;
@@ -721,6 +727,7 @@ void log_t::resize_abort() noexcept
     resize_lsn.store(0, std::memory_order_relaxed);
   }
 
+  writer_update();
   log_resize_release();
 }
 
@@ -924,7 +931,6 @@ void log_t::persist(lsn_t lsn, bool holding_latch) noexcept
 }
 #endif
 
-ATTRIBUTE_COLD ATTRIBUTE_NOINLINE
 void log_t::resize_write_buf(const byte *b, size_t length) noexcept
 {
   const size_t block_size_1= write_size - 1;
@@ -959,20 +965,24 @@ void log_t::resize_write_buf(const byte *b, size_t length) noexcept
                           b, offset, length) == DB_SUCCESS);
 }
 
-/** Write buf to ib_logfile0.
-@tparam release_latch whether to invoke latch.wr_unlock()
+/** Write buf to ib_logfile0 and possibly ib_logfile101.
+@tparam resizing whether to release latch and whether resize_in_progress()>1
 @return the current log sequence number */
-template<bool release_latch> inline lsn_t log_t::write_buf() noexcept
+template<log_t::resizing_and_latch resizing>
+inline __attribute__((always_inline))
+lsn_t log_t::write_buf() noexcept
 {
   ut_ad(latch_have_wr());
   ut_ad(!is_mmap());
   ut_ad(!srv_read_only_mode);
+  ut_ad(resizing == RETAIN_LATCH ||
+        (resizing == RESIZING) == (resize_in_progress() > 1));
 
   const lsn_t lsn{get_lsn(std::memory_order_relaxed)};
 
   if (write_lsn >= lsn)
   {
-    if (release_latch)
+    if (resizing != RETAIN_LATCH)
       latch.wr_unlock();
     ut_ad(write_lsn == lsn);
   }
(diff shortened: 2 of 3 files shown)


storage/innobase/log/log0log.cc
```diff
@@ -1002,6 +1002,8 @@ template<bool release_latch> inline lsn_t log_t::write_buf() noexcept
         MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) - length);
 # endif
       buf[length]= 0; /* allow recovery to catch EOF faster */
+      if (UNIV_LIKELY_NULL(re_write_buf))
+        resize_buf[length]= 0;
```
dr-m, line 1006: Instead of extending rather frequently executed low-level code, I think that we had better fix this bug in `log_t::write_checkpoint()`.

There is a small probability that `buf[length]` would already contain a 0 byte. That is not actually a problem for recovery. The `log_sys.buf` will always end at the end of a mini-transaction, that is, a byte 0 or 1 followed by a 4-byte checksum and a possible 8-byte encryption nonce. The log parser does not allow a 0 or 1 byte right after the end of a mini-transaction.

dr-m: I realized that changing this in `log_t::write_checkpoint()` would introduce a new system call and may require duplicating a lot of logic.

A better solution might be to replace the function `log_write_up_to()` with a function pointer, pointing to 3 or 4 different implementations:
 * one that invokes `log_t::persist()`, with or without log resizing being active
 * one that invokes `pwrite` to `ib_logfile0` while log resizing is not active
 * one that invokes `pwrite` to `ib_logfile0` and `ib_logfile101` while log resizing is active

The call in `log_write_and_flush()` should be infrequent and should not matter.

I will file a separate ticket for this refactoring.

dr-m: I ended up implementing this refactoring in 7ba1f87b5d128902e484f31852cfa6ac91ab1004. Surprisingly, the generated code for `log_writer<false>` would be larger than `log_writer<true>`, although I would expect it to be vice versa. I am investigating that.

dr-m: The explanation is that in `log_writer<true>`, the call to `log_t::write_buf<1>()` is not inline, while in the common case `log_writer<false>` it is. There is no non-`inline` instantiation of `template<int resizing> log_t::write_buf` with the template argument `0` in the executable that I checked, only with `-1` and `1`.

dr-m: I figured out that `__attribute__((always_inline))` does the trick: 9dd4ab5fc3fecf8da99ac40232ca8a8eef0b0037
For `log_t::resize_write_buf()` I will avoid forcing that, because the function is rather large (about 1/6 of the size of the caller) and would be invoked from two places: `log_t::write_buf<-1>` inlined in `log_writer<true>` and `log_t::write_buf<1>` inlined in `log_write_and_flush()`.

storage/innobase/log/log0log.cc
```diff
@@ -1165,6 +1180,18 @@ void log_write_up_to(lsn_t lsn, bool durable,
...
 
+void log_t::writer_update() noexcept
+{
+  ut_ad(latch_have_wr());
+  ut_ad(resize_in_progress() != 1);
+  writer=
+#ifdef HAVE_PMEM
+    is_mmap() ? log_writer<false,true> :
+#endif
+    resize_in_progress() ? log_writer<true> : log_writer<false>;
```
dr-m, line 1191: We may actually have `resize_in_progress() == 1` here: https://buildbot.mariadb.org/#/builders/234/builds/35089/steps/11/logs/stdio
The reason is simply that the caller did not assign `log_sys.resize_lsn` yet; it would after this call returns. I think that we must simply remove this assertion.

We do have an assertion on `resize_in_progress() > 1` in `log_t::write_buf()`, which would be inlined in `log_writer<false>` and `log_writer<true>`.

storage/innobase/log/log0log.cc
```diff
@@ -1002,13 +1012,14 @@ template<bool release_latch> inline lsn_t log_t::write_buf() noexcept
...
-      if (UNIV_LIKELY_NULL(re_write_buf))
-        MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) - length);
-# endif
-      buf[length]= 0; /* allow recovery to catch EOF faster */
+      buf[length]= 0; /* ensure that recovery catches EOF */
 #endif
+      if (resizing < 0 ? UNIV_UNLIKELY(!!re_write_buf) : bool(resizing))
+      {
+        MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) - length);
+        re_write_buf[length]= 0;
```
dr-m, line 1021: For the record: I tested commenting out both `re_write_buf[…]= 0` in this function, and then running the test `innodb.log_file_size_online`. The test passed 140×10 repetitions on a powerful server. I double checked that this interface was being used, instead of `cmake -DWITH_INNODB_PMEM=ON`. Finally, on another attempt (140×50 tests), I got a failure (not the one that I was looking for):
```
innodb.log_file_size_online 'encrypted'  w127 [ 8 fail ]
        Test ended at 2024-12-10 04:13:03

CURRENT_TEST: innodb.log_file_size_online
mysqltest: At line 51: query 'reap' failed: <Unknown> (2013): Lost connection to server during query
…
#6  0x00005bb38395248c in memcpy (__len=0x1c0, __src=0x200, __dest=<optimized out>) at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:29
#7  memcpy_aligned<16ul> (n=0x1c0, src=0x200, dest=<optimized out>) at /home/marko/10.6/include/assume_aligned.h:44
#8  log_t::write_buf<1> (this=0x5bb384ca2080 <log_sys>) at /home/marko/10.6/storage/innobase/log/log0log.cc:1045
#9  log_writer<true> (lsn=<optimized out>, durable=0x1, callback=<optimized out>) at /home/marko/10.6/storage/innobase/log/log0log.cc:1166
#10 0x00005bb3839e5f8f in log_t::write_up_to (callback=0x7436697673c0, durable=<optimized out>, lsn=0x3680df5, this=<optimized out>) at /home/marko/10.6/storage/innobase/include/log0log.h:569
#11 trx_flush_log_if_needed (lsn=0x3680df5, trx=trx@entry=0x74366ac01680) at /home/marko/10.6/storage/innobase/trx/trx0trx.cc:1275
#12 0x00005bb3839e7196 in trx_flush_log_if_needed (trx=0x74366ac01680, lsn=<optimized out>) at /home/marko/10.6/storage/innobase/trx/trx0trx.cc:1262
#13 0x00005bb383908ce3 in innobase_commit (hton=<optimized out>, thd=0x7435ec000c68, commit_trx=0x0) at /home/marko/10.6/storage/innobase/handler/ha_innodb.cc:4570
#14 0x00005bb3836cb200 in commit_one_phase_2 (thd=thd@entry=0x7435ec000c68, all=0x0, trans=0x7435ec004518, is_real_trans=0x1) at /home/marko/10.6/sql/handler.cc:2154
#15 0x00005bb3836cb3a9 in ha_commit_one_phase (thd=thd@entry=0x7435ec000c68, all=all@entry=0x0) at /home/marko/10.6/sql/handler.cc:2107
#16 0x00005bb3836cc347 in ha_commit_trans (thd=thd@entry=0x7435ec000c68, all=all@entry=0x0) at /home/marko/10.6/sql/handler.cc:1901
#17 0x00005bb3835b8385 in trans_commit_stmt (thd=thd@entry=0x7435ec000c68) at /home/marko/10.6/sql/transaction.cc:497
#18 0x00005bb38348d93c in mysql_execute_command (thd=thd@entry=0x7435ec000c68, is_called_from_prepared_stmt=is_called_from_prepared_stmt@entry=0x0) at /home/marko/10.6/sql/sql_parse.cc:6216
#19 0x00005bb38349287b in mysql_parse (thd=0x7435ec000c68, rawbuf=<optimized out>, length=<optimized out>, parser_state=<optimized out>) at /home/marko/10.6/sql/sql_parse.cc:8179
#20 0x00005bb38349488a in dispatch_command (command=command@entry=COM_QUERY, thd=thd@entry=0x7435ec000c68, packet=packet@entry=0x7435ec008579 "UPDATE t SET b='' WHERE a<10", packet_length=packet_length@entry=0x1c, blocking=blocking@entry=0x1) at /home/marko/10.6/sql/sql_parse.cc:2006
```
We have `log_sys.resize_flush_buf=nullptr` and `re_write_buf=nullptr` here, as well as `log_sys.resize_lsn=0`. Apparently, log resizing had been completed.

I realize that this version of the code is dereferencing `log_t::writer` while not holding `log_sys.latch`. It looks like it would only be safe to put the following part of `log_write_up_to()` behind a function pointer:
```c++
pending_write_lsn= write_lock.release(log_sys.write_buf<resizing>());
```

dr-m: After I made `log_sys.writer` wrap instantiations of the function `log_sys.write_buf`, the test `innodb.log_file_size_onilne` passed 140×50 rounds both with and without the following patch:
```diff
diff --git a/storage/innobase/log/log0log.cc b/storage/innobase/log/log0log.cc
index 9a851c41e67..a0b99211597 100644
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1018,7 +1018,6 @@ lsn_t log_t::write_buf() noexcept
       if (resizing < 0 ? UNIV_UNLIKELY(!!re_write_buf) : bool(resizing))
       {
         MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) - length);
-        re_write_buf[length]= 0;
       }
       length= write_size_1 + 1;
     }
@@ -1044,7 +1043,6 @@ lsn_t log_t::write_buf() noexcept
                            new_buf_free);
           memcpy_aligned<16>(resize_flush_buf, re_write_buf + length,
                              (new_buf_free + 15) & ~15);
-          re_write_buf[length + new_buf_free]= 0;
         }
         length+= write_size_1 + 1;
       }
```
I tested this patch also on my local system. In both cases, I failed to reproduce the problem that this patch aims to fix. @Thirunarayanan can you please test my fix with and without the above?

Thirunarayanan: yes, After applying the above patch(removing 0 in re_write_buf), I can repeat the scenario.

storage/innobase/log/log0log.cc
```diff
@@ -1023,27 +1034,28 @@ template<bool release_latch> inline lsn_t log_t::write_buf() noexcept
...
         buf[length]= 0; /* allow recovery to catch EOF faster */
         length&= ~write_size_1;
         memcpy_aligned<16>(flush_buf, buf + length, (new_buf_free + 15) & ~15);
-        if (UNIV_LIKELY_NULL(re_write_buf))
+        if (resizing < 0 ? UNIV_UNLIKELY(!!re_write_buf) : bool(resizing))
+        {
+          MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) -
+                           new_buf_free);
           memcpy_aligned<16>(resize_flush_buf, re_write_buf + length,
                              (new_buf_free + 15) & ~15);
```
mariadb-DebarunBanerjee, line 1046: > if (resizing < 0 ? UNIV_UNLIKELY(!!re_write_buf) : bool(resizing))

In all three places we have this condition, it should be safe and better to keep the old condition i.e. checking re_write_buf directly.
1. resizing < 0 : already has UNIV_UNLIKELY(!!re_write_buf)
2. resizing >=0 : byte *const re_write_buf{resizing ? resize_buf : nullptr};
Also, if re_write_buf is nullptr, certainly the code below is going to crash. So, I am not sure why not use
> if (UNIV_LIKELY_NULL(re_write_buf))

mariadb-DebarunBanerjee (commented):
The writer_update after checkpoint looks to be the key here. We need to be careful to have writer_update() called every time resize operation starts/stops and the code seems to be handling the cases nicely.

I have two suggestions  on the modified code for your consideration,

storage/innobase/log/log0log.cc
dr-m: Your suggestion to just check `re_write_buf` without any additional template condition would cause an unnecessary runtime condition for the case `resizing=1` where we know that resizing is in progress. This is because the local variable `re_write_buf` is being initialized as `resizing ? resize_buf : nullptr` and the compiler can’t know that the pointer is not supposed to be `nullptr`.

Earlier in `log_t::write_buf()` there are two debug assertions for ensuring that the compile-time and run-time parameters match:
```c++
  ut_ad(resizing < 0 || bool(resizing) == (resize_in_progress() > 1));
// …
    byte *const re_write_buf{resizing ? resize_buf : nullptr};
    ut_ad(resizing < 0 || bool(resizing) == !!re_write_buf);
```
I just realized that we could possibly use precursors of the [C++23 `assume`](https://en.cppreference.com/w/cpp/language/attributes/assume), like we do elsewhere:
```c++
#ifdef _MSC_VER
template <std::size_t Alignment, class T>
static inline T my_assume_aligned(T ptr)
{
  assert(reinterpret_cast<size_t>(ptr) % Alignment == 0);
  __assume(reinterpret_cast<size_t>(ptr) % Alignment == 0);
  return ptr;
}
#else
template <std::size_t Alignment, class T>
static inline T my_assume_aligned(T ptr)
{
  assert(reinterpret_cast<size_t>(ptr) % Alignment == 0);
  return static_cast<T>(__builtin_assume_aligned(ptr, Alignment));
}
#endif
```
I will try that.

dr-m: The `__builtin_unreachable()` based solution that I implemented in c391fb1ff122adbfb9a7f98d3239aa08311918d6 seems to work, both in `CMAKE_BUILD_TYPE=RelWithDebInfo` and in optimized `CMAKE_BUILD_TYPE=Debug`:
```asm
   call   0xd69270 <log_write_buf(byte const*, size_t, lsn_t)>
   mov    %rbx,%rdx
   mov    %r14,%rsi
   lea    0x1ae4750(%rip),%rdi        # 0x2186d00 <log_sys>
   call   0xd69ca0 <_ZN5log_t16resize_write_bufEPKhm>
```
There is no conditional branch between the two `call` near the end of the function `log_writer_resizing()`. The `if` condition in the following is essentially being replaced with `true`:
```c++
    log_write_buf(write_buf, length, offset);

    if (UNIV_LIKELY_NULL(re_write_buf))
      resize_write_buf(re_write_buf, length);
```

storage/innobase/include/log0log.h
```diff
@@ -551,10 +557,10 @@ struct log_t
   @param end_lsn    start LSN of the FILE_CHECKPOINT mini-transaction */
   inline void write_checkpoint(lsn_t end_lsn) noexcept;
 
-  /** Write buf to ib_logfile0.
-  @tparam release_latch whether to invoke latch.wr_unlock()
+  /** Write buf to ib_logfile0 and possibly ib_logfile101.
+  @tparam resizing -1=skip latch.wr_unlock(); else, resize_in_progress()>1
   @return the current log sequence number */
-  template<bool release_latch> inline lsn_t write_buf() noexcept;
+  template<int resizing> inline lsn_t write_buf() noexcept;
```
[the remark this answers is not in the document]
dr-m:
```c++
  /** Variations of write_buf() */
  enum resizing_and_latch {
    /** skip latch.wr_unlock(); log resizing may or may not be in progress */
    RETAIN_LATCH,
    /** invoke latch.wr_unlock(); !(resize_in_progress() > 1) */
    NOT_RESIZING,
    /** invoke latch.wr_unlock(); resize_in_progress() > 1 */
    RESIZING
  };

  /** Write buf to ib_logfile0 and possibly ib_logfile101.
  @tparam resizing whether to release latch and whether resize_in_progress()>1
  @return the current log sequence number */
  template<resizing_and_latch resizing> inline lsn_t write_buf() noexcept;
```
I also removed the implicit conversion from `bool` to this template parameter.

mariadb-DebarunBanerjee (approved):
@dr-m  Thanks for the changes. Looks good me.

PR #3681, opened 2024-12-05
MDEV-35583 Tests failing on macOS

DaveGosselin-MariaDB:
These tests, unit.my_json_writer and unit.dynstring, rely on THR_KEY_mysys but it is not initialized.  On Linux, the corresponding thread variable is null, but on macOS it has a nonzero value.  In all cases, initialize the variable explicitly by calling MY_INIT and my_end appropriately.


grooverdan:
https://buildbot.mariadb.org/#/builders/590/builds/10938/steps/4/logs/stdio

unit.dynstring                           w7 [ pass ]     85

The my_json_writer didn't get executed as the max-test-fail limit was reached before it could be run. It compiled ok though. Seems a good enough pattern of a fix.

grooverdan (approved):

PR #3682, opened 2024-12-05
MDEV-35585 unit.json_normalize crashes on mac

DaveGosselin-MariaDB:
Like MDEV-35583, initialize THR_KEY_mysys explicitly by calling MY_INIT and my_end.


cvicentiu:
LGTM!

PR #3685, opened 2024-12-09
MDEV-35602: Wrong table column count mentioning corruption harmful

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35602*

## Description

ER_COL_COUNT_DOESNT_MATCH_CORRUPTED_V2 mentioning the table is possibly corrupted leaves users with no avenue as to what action to take.

The simpler version of just saying the column count isn't matching leaves a manual reconciliation and mariadb-upgrade paths as something that can be attempted.

## Release Notes

"Possible corruption" notices around system table with the wrong column count was unlikely to be corruption, just a non-standard upgrade path.

## How can this PR be tested?

## PR quality check


cvicentiu (changes requested):
Please explain the problem you are fixing in the MDEV properly.

* What is the user story behind this?
* Where does this problem come from?
* What kind of use cases lead to this error?

Because if users run into this issue, there must be a "tutorial" somewhere that tells them to shoot themselves in the foot.

If you want users to have an easier time, make sure that the errors clearly explain what the problem is and what to look at in order to resolve things. With your suggested changes you do not explain to the user anything extra.

PR #3686, opened 2024-12-09
MDEV-35607 - gcc-15 fixes - use void for sig_handler returns

heitbaum:
- [x] *The Jira issue number for this PR is: [MDEV-35607](https://jira.mariadb.org/browse/MDEV-35607)*

## Description
fixes compile issue with gcc-15
```
../mysys/stacktrace.c: In function 'default_handle_fatal_signal':
../mysys/stacktrace.c:53:3: error: 'return' with no value, in function 
returning non-void [-Wreturn-mismatch]
    53 |   return;
       |   ^~~~~~
...
```

## Release Notes
Fixes build with gcc-15 when combined with fixes from:
- https://jira.mariadb.org/browse/MDEV-35605
- https://jira.mariadb.org/browse/MDEV-35606

## How can this PR be tested?
Testing can be performed using gcc-15 as a compiler

## PR quality check

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -1168,10 +1168,10 @@ static void print_warnings(void);
 static void print_last_query_cost(void);
 static void end_timer(ulonglong start_time, char *buff);
 static void nice_time(double sec,char *buff,bool part_second);
-extern "C" sig_handler mysql_end(int sig) __attribute__ ((noreturn));
-extern "C" sig_handler handle_sigint(int sig);
+extern "C" void mysql_end(int sig) __attribute__ ((noreturn));
+extern "C" void handle_sigint(int sig);
 #if defined(HAVE_TERMIOS_H) && defined(GWINSZ_IN_SYS_IOCTL)
-static sig_handler window_resize(int sig);
+static void window_resize(int sig);
 #endif
 
 static void end_in_sig_handler(int sig);
@@ -1409,7 +1409,7 @@ int main(int argc,char *argv[])
 #endif
 }
 
-sig_handler mysql_end(int sig)
+void mysql_end(int sig)
 {
 #ifndef _WIN32
   /*
@@ -1594,7 +1594,7 @@ bool kill_query(const char *reason)
   If 'source' is executed, abort source command
   no query in process, regenerate prompt.
 */
-sig_handler handle_sigint(int sig)
+void handle_sigint(int sig)
 {
   /*
      On Unix only, if no query is being executed just clear the prompt,
@@ -1634,7 +1634,7 @@ sig_handler handle_sigint(int sig)
 
 
 #if defined(HAVE_TERMIOS_H) && defined(GWINSZ_IN_SYS_IOCTL)
-sig_handler window_resize(int sig)
+void window_resize(int sig)
 {
   struct winsize window_size;
 
--- a/client/mysqladmin.cc
+++ b/client/mysqladmin.cc
@@ -62,7 +62,7 @@ static my_bool sql_connect(MYSQL *mysql, uint wait);
 static int execute_commands(MYSQL *mysql,int argc, char **argv);
 static char **mask_password(int argc, char ***argv);
 static int drop_db(MYSQL *mysql,const char *db);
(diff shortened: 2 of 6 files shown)


grooverdan:
Interesting - note it looks like it was intended to autodetect this - https://github.com/MariaDB/server/blob/f5821aaf7731e1f74b970fdaf186b945ae012c9d/configure.cmake#L678 - I haven't looked into the history as to why.

heitbaum:
> Interesting - note it looks like it was intended to autodetect this -
> 
> https://github.com/MariaDB/server/blob/f5821aaf7731e1f74b970fdaf186b945ae012c9d/configure.cmake#L678
> 
> - I haven't looked into the history as to why.

Yes. I'll have a look why once I get all the parts cleanly compiling. I thought it was going to be a quick one liner... :-(

dr-m:
> Interesting - note it looks like it was intended to autodetect this -
> 
> https://github.com/MariaDB/server/blob/f5821aaf7731e1f74b970fdaf186b945ae012c9d/configure.cmake#L678
> - I haven't looked into the history as to why.

If that auto-detection is obsolete by now, maybe we could remove it and thereby reduce the CMake execution time. `man 2 signal` on my system mentions C11 (ISO/IEC 9899:2011). In https://en.cppreference.com/w/c/program/signal there is no mention of `signal()` returning `void`; it always was supposed to return a function pointer. I would expect that any currently supported compiler would support at least C11.

heitbaum:
> > Interesting - note it looks like it was intended to autodetect this -
> > https://github.com/MariaDB/server/blob/f5821aaf7731e1f74b970fdaf186b945ae012c9d/configure.cmake#L678
> > 
> > 
> > * I haven't looked into the history as to why.
> 
> If that auto-detection is obsolete by now, maybe we could remove it and thereby reduce the CMake execution time. `man 2 signal` on my system mentions C11 (ISO/IEC 9899:2011). In https://en.cppreference.com/w/c/program/signal there is no mention of `signal()` returning `void`; it always was supposed to return a function pointer. I would expect that any currently supported compiler would support at least C11.

That article concurs with the changes I have made:

```
void (*signal( int sig, void (*handler) (int))) (int);
                        ^^^^
```
In that the handler is suppose to be `void` not `sig_handler` which happens to be `int` in this instance.

vuvova:
as @grooverdan wrote, we need to understand why the check isn't working before we decide to remove or ignore it.
Here's how you can find it out:
* run cmake with whatever options you normally do it
* then run `cmake -USIGNAL_RETURN_TYPE_IS_VOID --debug-trycompile`
* then `cd CMakeScratch/TryCompile-* && make`

heitbaum:
This has been fixed upstream in 11.4.5 - thanks

grooverdan:
> This has been fixed upstream in 11.4.5 - thanks

Really? I can't see any code changes that would make behaviour different?

Have you had a gcc update that performs the try_compile outcome differently?

heitbaum:
with 15-20250202 - mariadb-11.4.4 compile fails with
```
mysys/thr_timer.c
In file included from ../include/my_global.h:1033,
                 from ../mysys/mysys_priv.h:19,
                 from ../mysys/thr_timer.c:22:
../mysys/thr_timer.c: In function 'process_timers':
../include/my_dbug.h:172:46: error: 'return' with no value, in function returning non-void [-Wreturn-mismatch]
  172 | #define DBUG_VOID_RETURN                do { return; } while(0)
      |                                              ^~~~~~
../mysys/thr_timer.c:289:3: note: in expansion of macro 'DBUG_VOID_RETURN'
  289 |   DBUG_VOID_RETURN;
      |   ^~~~~~~~~~~~~~~~
../mysys/thr_timer.c:240:20: note: declared here
  240 | static sig_handler process_timers(struct timespec *now)
      |                    ^~~~~~~~~~~~~~
[124/1940] Building C object mysys/CMakeFiles/mysys.dir/thr_lock.c.o
ninja: build stopped: subcommand failed.
```
and mariadb-11.4.5 compiles successfully

PR #3689, opened 2024-12-10
Change my_umask{,_dir} to mode_t and remove os_innodb_umask

grooverdan:

## Description

The mode_t type is required by the standard for
umask and open libc calls.

This has only been changed inside innodb as the global my_umask (which should be the same type), but is an int, is exposed as a defined plugin accessable variable and therefore would break ABI compatibility if this was changed.

It also isn't creating compile warnings in the global server, only within innodb.

Clang-19 reports these was warnings:

[55/123] Building CXX object storage/innobase/CMakeFiles/innobase.dir/os/os0file.cc.o storage/innobase/os/os0file.cc:1075:46: warning: implicit conversion loses integer precision: 'ulint' (aka 'unsigned long') to 'mode_t' (aka 'unsigned int') [-Wshorten-64-to-32]
 1075 |                 file = open(name, create_flag | O_CLOEXEC, os_innodb_umask);
      |                        ~~~~                                ^~~~~~~~~~~~~~~
storage/innobase/os/os0file.cc:1249:46: warning: implicit conversion loses integer precision: 'ulint' (aka 'unsigned long') to 'mode_t' (aka 'unsigned int') [-Wshorten-64-to-32]
 1249 |                 file = open(name, create_flag | O_CLOEXEC, os_innodb_umask);
      |                        ~~~~                                ^~~~~~~~~~~~~~~
storage/innobase/os/os0file.cc:1381:45: warning: implicit conversion loses integer precision: 'ulint' (aka 'unsigned long') to 'mode_t' (aka 'unsigned int') [-Wshorten-64-to-32]
 1381 |         file = open(name, create_flag | O_CLOEXEC, os_innodb_umask);
      |                ~~~~                                ^~~~~~~~~~~~~~~

## Release Notes

nothing - No user exposed functionality

## How can this PR be tested?

compile with clang 19+ - no warnings is good.

## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -4024,8 +4024,6 @@ static int innodb_init(void* p)
 			    test_filename));
 #endif /* DBUG_OFF */
 
-	os_file_set_umask(my_umask);
-
 	/* Setup the memory alloc/free tracing mechanisms before calling
 	any functions that could possibly allocate memory. */
 	ut_new_boot();
--- a/include/my_sys.h
+++ b/include/my_sys.h
@@ -246,9 +246,9 @@ extern void (*my_sigtstp_cleanup)(void),
 					/* Executed before jump to shell */
 	    (*my_sigtstp_restart)(void);
 					/* Executed when coming from shell */
-extern MYSQL_PLUGIN_IMPORT int my_umask;		/* Default creation mask  */
-extern int my_umask_dir,
-	   my_recived_signals,	/* Signals we have got */
+extern MYSQL_PLUGIN_IMPORT mode_t my_umask;	/* Default creation mask  */
+extern mode_t my_umask_dir;
+extern int my_recived_signals,	/* Signals we have got */
 	   my_safe_to_handle_signal, /* Set when allowed to SIGTSTP */
 	   my_dont_interrupt;	/* call remember_intr when set */
 extern MYSQL_PLUGIN_IMPORT my_bool my_use_symdir;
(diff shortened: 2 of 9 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -4024,7 +4024,7 @@ static int innodb_init(void* p)
 			    test_filename));
 #endif /* DBUG_OFF */
 
-	os_file_set_umask(my_umask);
+	os_file_set_umask((mode_t) my_umask);
```
dr-m, line 4027: What if we just remove this entire function as well as `os_innodb_umask`, and change the type of `my_umask` to `mode_t`?

I believe that the reason for this kind of data structure duplication is that originally, the intention was to embed InnoDB in multiple different databases, or to develop it as a stand-alone product outside MySQL.

dr-m (commented):
Coincidentally, just 1 or 2 weeks ago I encountered `os_innodb_umask` and was thinking of filing a bug to remove the code duplication.

storage/innobase/handler/ha_innodb.cc
grooverdan: Ype, can remove function and go straight though.

The definition of `my_umask` looks like an exposed ABI so I was avoiding changing it:
`include/my_sys.h:extern MYSQL_PLUGIN_IMPORT int my_umask;               /* Default creation mask  */`. 

From the clang warnings this is only a sign difference as opposed to innodb that has the type longer.

grooverdan: Ack, updated to remove innodb implementation and normalize the mode_t types elsewhere in the codebase.

dr-m (approved):

PR #3690, opened 2024-12-11
MDEV-35626 Race condition between buf_page_create_low() and read comp…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-35626*

## Description
This regression is introduced in 10.6.20 by MDEV-34453 Trying to read 16384 bytes at 70368744161280. The page state could change after being buffer-fixed and needs to be read again after locking the page.

## Release Notes
TBD

## How can this PR be tested?

It could not be repeated by mtr test. The duplicate bug MDEV-35181 describes RQG test that could repeat the issue.

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -3322,6 +3322,8 @@ retry:
         ut_ad(!bpage->is_io_fixed(state));
         ut_ad(bpage->buf_fix_count(state));
       }
+      else
+        state= bpage->state();
 
       ut_ad(state >= buf_page_t::FREED);
       ut_ad(state < buf_page_t::READ_FIX);


dr-m (approved):
This is the fix that I suggested in [MDEV-35626](https://jira.mariadb.org/browse/MDEV-35626). 

But, in that ticket I also pointed out that there is some redundant code in the rare case that we have a `ROW_FORMAT=COMPRESSED` block in the buffer pool, without an uncompressed page (`bpage->frame == nullptr`). In that branch, we are declaring another local variable `state`, which should have the same value as this one, with the possible exception of the buffer-fix count.

I would like to point out that when this fix is present and after the following has been executed:
```c++
      if (state < buf_page_t::UNFIXED)
        bpage->set_reinit(buf_page_t::FREED);
      else
      {
        bpage->set_reinit(state & buf_page_t::LRU_MASK);
        ibuf_exist= (state & buf_page_t::LRU_MASK) == buf_page_t::IBUF_EXIST;
      }
```
the `bpage->state()` will be `buf_page_t::UNFIXED` plus at least one buffer-fix count. That cannot change even after we release `buf_pool.mutex`, because other state changes than the buffer-fix count are mostly protected by `bpage->lock`, which this thread will be holding in exclusive mode. Therefore, it should be safe to remove the following code:
```diff
@@ -3345,24 +3347,12 @@ static buf_block_t *buf_page_create_low(page_id_t page_id, ulint zip_size,
       }
       else
       {
-        auto state= bpage->state();
-        ut_ad(state >= buf_page_t::FREED);
-        ut_ad(state < buf_page_t::READ_FIX);
-
         page_hash_latch &hash_lock= buf_pool.page_hash.lock_get(chain);
         /* It does not make sense to use transactional_lock_guard here,
         because buf_relocate() would likely make the memory transaction
         too large. */
         hash_lock.lock();
 
-        if (state < buf_page_t::UNFIXED)
-          bpage->set_reinit(buf_page_t::FREED);
-        else
-        {
-          bpage->set_reinit(state & buf_page_t::LRU_MASK);
-          ibuf_exist= (state & buf_page_t::LRU_MASK) == buf_page_t::IBUF_EXIST;
-        }
-
         mysql_mutex_lock(&buf_pool.flush_list_mutex);
         buf_relocate(bpage, &free_block->page);
         free_block->page.lock.x_lock();
```
Here, the value of the new variable `state` should always be `buf_page_t::UNFIXED` plus some unfix count. I now realize that this duplicated code may lead to another corruption:  `ibuf_exist` could initially have the value `true`, but this duplicated logic would reset it to `false`. As a result of this, we would fail to invoke `ibuf_merge_or_delete_for_page()` and eventually end up applying some "ghost records" to the newly created page, similar to what happened in another scenario before the fix 34c283ba1b3b0b51488746416a737c1456df2fe9.

Please check this analysis and file a new bug about this. We already had [MDEV-34879](https://jira.mariadb.org/browse/MDEV-34879), which but this is something else. I did not find any open bug about this potential `ROW_FORMAT=COMPRESSED` table corruption.

PR #3691, opened 2024-12-11
MDEV-34898   Doublewrite recovery of innodb_checksum_algorithm=full_crc32 encrypted pages does not work

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34989*

## Description
- InnoDB fails to recover the full crc32 encrypted page from doublewrite buffer. The reason is that buf_dblwr_t::recover() fails to identify the space id from the page because the page has been encrypted from FIL_PAGE_FILE_FLUSH_LSN_OR_KEY_VERSION bytes.

Fix:
===
buf_dblwr_t::recover(): preserve any pages whose space_id does not match a known tablespace. These could be encrypted pages of tablespaces that had been created with
innodb_checksum_algorithm=full_crc32.

buf_page_t::read_complete(): If the page looks corrupted and the tablespace is encrypted and in full_crc32 format, try to restore the page from doublewrite buffer.

recv_dblwr_t::recover_encrypted_page(): Find the page which has the same page number and try to decrypt the page using space->crypt_data. After decryption, compare the space id. Write the recovered page back to the file.

## How can this PR be tested?
./mtr encryption.doublewrite_debug

## PR quality check

### DIFF
--- a/mysql-test/suite/encryption/t/debug_key_management.test
+++ b/mysql-test/suite/encryption/t/debug_key_management.test
@@ -3,6 +3,7 @@
 -- source include/innodb_undo_tablespaces.inc
 -- source include/not_embedded.inc
 
+call mtr.add_suppression("InnoDB: Encrypted page \\[page id: space=[1-9][0-9]*, page number=[0-9]*\\] in file");
 if (`select count(*) = 0 from information_schema.plugins
      where plugin_name = 'debug_key_management' and plugin_status='active'`)
 {
--- a/mysql-test/suite/encryption/t/doublewrite_debug.test
+++ b/mysql-test/suite/encryption/t/doublewrite_debug.test
@@ -1,0 +1,223 @@
+--source include/have_innodb.inc
+--source include/have_debug.inc
+--source include/not_embedded.inc
+--source include/have_example_key_management_plugin.inc
+call mtr.add_suppression("InnoDB: Encrypted page \\[page id: space=[1-9][0-9]*, page number=3\\] in file .*test.t[12]\\.ibd looks corrupted");
+call mtr.add_suppression("InnoDB: Unable to apply log to corrupted page ");
+call mtr.add_suppression("InnoDB: Plugin initialization aborted");
+call mtr.add_suppression("Plugin 'InnoDB' init function returned error");
+call mtr.add_suppression("Plugin 'InnoDB' registration as a STORAGE ENGINE failed");
+
+let INNODB_PAGE_SIZE=`select @@innodb_page_size`;
+let MYSQLD_DATADIR=`select @@datadir`;
+let ALGO=`select @@innodb_checksum_algorithm`;
+
+create table t1 (f1 int primary key, f2 blob)page_compressed = 1 engine=innodb stats_persistent=0;
+create table t2(f1 int primary key, f2 blob)engine=innodb stats_persistent=0;
+
+start transaction;
+insert into t1 values(1, repeat('#',12));
+insert into t1 values(2, repeat('+',12));
+insert into t1 values(3, repeat('/',12));
+insert into t1 values(4, repeat('-',12));
+insert into t1 values(5, repeat('.',12));
+insert into t2 select * from t1;
+commit work;
+
+# Slow shutdown and restart to make sure ibuf merge is finished
+SET GLOBAL innodb_fast_shutdown = 0;
+let $shutdown_timeout=;
+let $restart_parameters=--debug_dbug=+d,ib_log_checkpoint_avoid_hard --innodb_flush_sync=0;
+--source include/restart_mysqld.inc
+--source ../../suite/innodb/include/no_checkpoint_start.inc
+
+select space into @t1_space_id from information_schema.innodb_sys_tablespaces where name='test/t1';
+select space into @t2_space_id from information_schema.innodb_sys_tablespaces where name='test/t2';
+
+begin;
+insert into t1 values (6, repeat('%', 400));
+insert into t2 values (6, repeat('%', 400));
+
+# Copy the t1.ibd, t2.ibd file
+let $targetdir=$MYSQLTEST_VARDIR/tmp/backup_1;
+--disable_result_log
+exec $XTRABACKUP --defaults-file=$MYSQLTEST_VARDIR/my.cnf --backup --target-dir=$targetdir;
+--enable_result_log
+
+echo # xtrabackup prepare;
+--disable_result_log
+exec $XTRABACKUP  --prepare --target-dir=$targetdir;
+
+set global innodb_saved_page_number_debug = 3;
+set global innodb_fil_make_page_dirty_debug = @t1_space_id;
+
+set global innodb_saved_page_number_debug = 3;
+set global innodb_fil_make_page_dirty_debug = @t2_space_id;
+
+set global innodb_buf_flush_list_now = 1;
+--let CLEANUP_IF_CHECKPOINT=drop table t1, t2, unexpected_checkpoint;
+--source ../../suite/innodb/include/no_checkpoint_end.inc
+# Corrupt the page 3 in t1.ibd, t2.ibd file
+perl;
+use IO::Handle;
+do "$ENV{MTR_SUITE_DIR}/include/crc32.pl";
+my $polynomial = 0x82f63b78; # CRC-32C
+my $algo = $ENV{ALGO};
+die "Unsupported innodb_checksum_algorithm=$algo\n" unless $algo =~ /crc32/;
+
+my $fname= "$ENV{'MYSQLD_DATADIR'}test/t1.ibd";
+my $page_size = $ENV{INNODB_PAGE_SIZE};
+my $page;
+do "$ENV{MTR_SUITE_DIR}/../innodb/include/crc32.pl";
+open(FILE, "+<", $fname) or die;
+sysseek(FILE, 3*$page_size, 0);
+sysread(FILE, $page, $page_size)==$page_size||die "Unable to read $name\n";
+sysseek(FILE, 3*$page_size, 0)||die "Unable to seek $fname\n";
+my $corrupted = $page;
+# Set FIL_PAGE_LSN to the maximum
+substr($corrupted, 16, 8) = chr(255) x 8;
+substr($corrupted, $page_size - 8, 8) = chr(255) x 8;
+if ($algo =~ /full_crc32/)
+{
+    my $ck = mycrc32(substr($corrupted, 0, $page_size - 4), 0, $polynomial);
+    substr($corrupted, $page_size - 4, 4) = pack("N", $ck);
+}
+else
+{
+    # Replace the innodb_checksum_algorithm=crc32 checksum
+    my $ck= pack("N",
+                 mycrc32(substr($corrupted, 4, 22), 0, $polynomial) ^
+                 mycrc32(substr($corrupted_, 38, $page_size - 38 - 8), 0,
+                         $polynomial));
+    substr ($corrupted, 0, 4) = $ck;
+    substr ($corrupted, $page_size - 8, 4) = $ck;
+}
+syswrite(FILE, $corrupted);
+close FILE;
+
+# Zero the complete page
+my $fname= "$ENV{'MYSQLD_DATADIR'}test/t2.ibd";
+open(FILE, "+<", $fname) or die;
+FILE->autoflush(1);
+binmode FILE;
+sysseek(FILE, 3*$page_size, 0);
+print FILE chr(0) x ($ENV{'INNODB_PAGE_SIZE'});
+close FILE;
+EOF
+
+# Successful recover from doublewrite buffer
+let $restart_parameters=;
+--source include/start_mysqld.inc
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+let SEARCH_PATTERN=InnoDB: Recovered page \\[page id: space=[1-9]*, page number=3\\];
+--source include/search_pattern_in_file.inc
+
+check table t1;
+check table t2;
+select f1, f2 from t1;
+select f1, f2 from t2;
+
+SET GLOBAL innodb_fast_shutdown = 0;
+let $shutdown_timeout=;
+let $restart_parameters=--debug_dbug=+d,ib_log_checkpoint_avoid_hard --innodb_flush_sync=0;
+--source ../../mariabackup/include/restart_and_restore.inc
+--source ../../suite/innodb/include/no_checkpoint_start.inc
+select space into @t1_space_id from information_schema.innodb_sys_tablespaces where name='test/t1';
+
+begin;
+insert into t1 values (6, repeat('%', 400));
+
+set global innodb_saved_page_number_debug = 3;
+set global innodb_fil_make_page_dirty_debug = @t1_space_id;
+
+set global innodb_buf_flush_list_now = 1;
+--let CLEANUP_IF_CHECKPOINT=drop table t1, unexpected_checkpoint;
+--source ../../suite/innodb/include/no_checkpoint_end.inc
+
+--echo # Corrupt the page 3 in t1.ibd file
+--echo # Assign the maximum value to lsn in doublewrite buffer page
+perl;
+use IO::Handle;
+do "$ENV{MTR_SUITE_DIR}/include/crc32.pl";
+my $polynomial = 0x82f63b78; # CRC-32C
+my $algo = $ENV{ALGO};
+die "Unsupported innodb_checksum_algorithm=$algo\n" unless $algo =~ /crc32/;
+
+my $fname= "$ENV{'MYSQLD_DATADIR'}test/t1.ibd";
+my $page_size = $ENV{INNODB_PAGE_SIZE};
+my $page;
+do "$ENV{MTR_SUITE_DIR}/../innodb/include/crc32.pl";
+open(FILE, "+<", $fname) or die;
+sysseek(FILE, 3*$page_size, 0);
+sysread(FILE, $page, $page_size)==$page_size||die "Unable to read $name\n";
+sysseek(FILE, 3*$page_size, 0)||die "Unable to seek $fname\n";
+my $corrupted = $page;
+# Set FIL_PAGE_LSN to the maximum
+substr($corrupted, 16, 8) = chr(255) x 8;
+substr($corrupted, $page_size - 8, 8) = chr(255) x 8;
+if ($algo =~ /full_crc32/)
+{
+    my $ck = mycrc32(substr($corrupted, 0, $page_size - 4), 0, $polynomial);
+    substr($corrupted, $page_size - 4, 4) = pack("N", $ck);
+}
+else
+{
+    # Replace the innodb_checksum_algorithm=crc32 checksum
+    my $ck= pack("N",
+                 mycrc32(substr($corrupted, 4, 22), 0, $polynomial) ^
+                 mycrc32(substr($corrupted_, 38, $page_size - 38 - 8), 0,
+                         $polynomial));
+    substr ($corrupted, 0, 4) = $ck;
+    substr ($corrupted, $page_size - 8, 4) = $ck;
+}
+syswrite(FILE, $corrupted);
+close FILE;
+
+# Change the page lsn to maximum value
+open(FILE, "+<", "$ENV{MYSQLD_DATADIR}ibdata1")||die "cannot open ibdata1\n";
+sysseek(FILE, 6 * $page_size - 190, 0)||die "Unable to seek ibdata1\n";
+sysread(FILE, $_, 12) == 12||die "Unable to read TRX_SYS\n";
+my($magic,$d1,$d2)=unpack "NNN", $_;
+die "magic=$magic, $d1, $d2\n" unless $magic == 536853855 && $d2 >= $d1 + 64;
+sysseek(FILE, $d1 * $page_size, 0)||die "Unable to seek ibdata1\n";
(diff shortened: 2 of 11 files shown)


storage/innobase/srv/srv0start.cc
```diff
@@ -1507,8 +1507,6 @@ dberr_t srv_start(bool create_new_db)
 			: recv_recovery_from_checkpoint_start(flushed_lsn);
 		recv_sys.close_files();
 
-		recv_sys.dblwr.pages.clear();
```
dr-m, line 1510: `recv_sys.dblwr.pages` would seem to waste memory until `recv_sys.close()` is invoked on server shutdown. The collection serves no useful purpose after recovery has finished. Can we clear this near the end of `recv_sys_t::apply()` inside the `if (last_batch)` section while we are holding `recv_sys.mutex`?

storage/innobase/log/log0recv.cc
```diff
@@ -4841,6 +4841,53 @@ bool recv_dblwr_t::validate_page(const page_id_t page_id, lsn_t max_lsn,
...
+    const ulint physical_size= space->physical_size();
+    fil_io_t fio= space->io(IORequestWrite,
+                            os_offset_t{page_no} * physical_size,
+                            physical_size, const_cast<byte*>(page));
+    if (fio.err == DB_SUCCESS)
+    {
+      sql_print_information("InnoDB: Recovered page [page id: space="
+                            UINT32PF ", page number=" UINT32PF "] "
+			    "to '%s' from the doublewrite buffer.",
+                            space->id, page_no, fio.node->name);
```
dr-m, line 4884: Unfortunately, ca501ffb04246dcaa1f1d433d916d8436e30602e is not present in 10.6. Probably the easiest solution that would avoid trouble on a merge to 10.11 would be to replace `space->id` with `uint32_t(space->id)`.

storage/innobase/buf/buf0buf.cc
```diff
@@ -3777,6 +3777,16 @@ dberr_t buf_page_t::read_complete(const fil_node_t &node)
   if (err == DB_PAGE_CORRUPTED || err == DB_DECRYPTION_FAILED)
   {
 release_page:
+    if (node.space->full_crc32() && node.space->crypt_data
+        && recv_recovery_is_on())
+    {
+      /* Recover from doublewrite buffer */
+      err= recv_sys.dblwr.recover_encrypted_page(
+             node.space, id().page_no(),
+             const_cast<byte*>(read_frame));
```
dr-m, line 3786: When this is not returning `DB_SUCCESS`, it is discarding the previous `err`, which could be `DB_PAGE_CORRUPTED` or `DB_DECRYPTION_FAILED`. Maybe the function call should just return `bool` and we should preserve the current `err` across the call?

Also, for an accurate file name in `sql_print_information` reporting about successful recovery, we should pass `node` instead of `node.space` to the function. That would work for a multi-page system tablespace.

storage/innobase/buf/buf0buf.cc
```diff
@@ -3777,6 +3777,16 @@ dberr_t buf_page_t::read_complete(const fil_node_t &node)
   if (err == DB_PAGE_CORRUPTED || err == DB_DECRYPTION_FAILED)
   {
 release_page:
+    if (node.space->full_crc32() && node.space->crypt_data
+        && recv_recovery_is_on())
```
dr-m, line 3781: Lines should be split after binary operators such as `&&`, not before them.

It looks like `recv_sys_t::recovery_on` had better be defined as `Atomic_relaxed<bool>`, and the assignment to `false` should be documented as being protected by `mutex`.

Not all assignments to `false` are currently protected by `recv_sys.mutex`. Before the call to `recv_recovery_read_max_checkpoint()` there is an assignment that is not protected. This assignment seems to be redundant, because `recv_recovery_from_checkpoint_start()` should not have been invoked yet.

storage/innobase/log/log0recv.cc
```diff
@@ -4841,6 +4841,53 @@ bool recv_dblwr_t::validate_page(const page_id_t page_id, lsn_t max_lsn,
...
+
+    if (space->is_compressed())
+    {
+      ulint write_size= fil_page_decompress(slot->crypt_buf, read_buf,
+                                            space->flags);
+      slot->release();
+      if (write_size == 0)
+        continue;
+    }
+    else slot->release();
```
dr-m, line 4870: This could be simplified:
```c++
    buf_tmp_buffer_t *slot= buf_pool.io_buf_reserve(false);
    ut_a(slot);
    slot->allocate();
    if (!fil_space_decrypt(space, slot->crypt_buf, read_buf) ||
        (space->is_compressed() && 
         !fil_page_decompress(slot->crypt_buf, read_buf, space->flags))
    {
      slot->release();
      continue;
    }
    slot->release();
```
Because `slot->reserve()` is just a store of a constant, it does not seem to be useful to avoid duplicating that, by doing something like
```c++
slot->allocate();
const bool invalid= …;
slot->release();
if (invalid)
  continue;
```

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -385,8 +386,11 @@ void buf_dblwr_t::recover()
     fil_space_t *space= fil_space_t::get(space_id);
 
     if (!space)
+    {
       /* The tablespace that this page once belonged to does not exist */
+      encrypted_pages.push_back(*i);
       continue;
+    }
```
dr-m, line 393: It would be good to add a comment that the `FIL_PAGE_LSN` of these pages will be validated in `recover_encrypted_page()`.

storage/innobase/include/log0recv.h
```diff
@@ -150,6 +150,8 @@ struct recv_dblwr_t
                         const fil_space_t *space= nullptr,
                         byte *tmp_buf= nullptr) const noexcept;
 
+  dberr_t recover_encrypted_page(fil_space_t *space, uint32_t page_no,
+                                 byte *tmp_buf) const noexcept;
```
dr-m, line 154: The parameters and the usage of this function need to be documented.

mysql-test/suite/encryption/t/doublewrite_debug.test
```diff
@@ -0,0 +1,75 @@
...
+
+my $fname= "$ENV{'MYSQLD_DATADIR'}test/t2.ibd";
+open(FILE, "+<", $fname) or die;
+FILE->autoflush(1);
+binmode FILE;
+seek(FILE, 3 * $ENV{'INNODB_PAGE_SIZE'}, SEEK_SET);
+print FILE chr(0) x ($ENV{'INNODB_PAGE_SIZE'});
+close FILE;
+EOF
+
```
dr-m, line 64: I checked that `FIL_PAGE_LSN` is not encrypted on the page. Therefore, we should be able to corrupt just that (for example, write `chr(255) x 8`, which will be newer than end of the log) and recompute the checksum. I think that it would be more interesting than this "totally corrupted" page.

Can you please also do a test variant that corrupts this page in the doublewrite buffer, with an LSN that is before the checkpoint? That should lead to InnoDB crash recovery failing. Better make it two attempts: also with an LSN that is after the end of the log.

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -385,8 +386,13 @@ void buf_dblwr_t::recover()
     fil_space_t *space= fil_space_t::get(space_id);
 
     if (!space)
-      /* The tablespace that this page once belonged to does not exist */
+    {
+      /* The tablespace that this page once belonged to does not
+      exist and validate the lsn for these pages in
+      find_encrypted_page(). */
```
dr-m, line 392: The comment could be improved:
>The tablespace that this page once belonged to does not seem to exist. However, this could be an encrypted page in full_crc32 format, to be checked by find_encrypted_page().

In fact, I checked and the `FIL_PAGE_LSN` will be stored in clear in `fil_encrypt_buf_for_full_crc32()`. So, we can and should validate the `FIL_PAGE_LSN` already here and avoid preserving invalid pages for `find_encrypted_page()`. In this way, `find_encrypted_page()` will be able to simply invoke `buf_page_is_corrupted(true, …)` without any fear of a false alarm regarding doublewrite buffer pages.

storage/innobase/srv/srv0start.cc
```diff
@@ -324,8 +324,6 @@ static dberr_t create_log_file(bool create_new_db, lsn_t lsn,
 		srv_startup_is_before_trx_rollback_phase = false;
 	}
 
-	/* Enable checkpoints in buf_flush_page_cleaner(). */
-	recv_sys.recovery_on = false;
```
dr-m, line 328: This had better be replaced with `ut_ad(!recv_sys.recovery_on)` to avoid any surprise.

storage/innobase/srv/srv0start.cc
```diff
@@ -1351,9 +1349,6 @@ dberr_t srv_start(bool create_new_db)
 				return(srv_init_abort(DB_ERROR));
 			}
 
-			/* Enable checkpoints in the page cleaner. */
-			recv_sys.recovery_on = false;
```
dr-m, line 1355: `ut_ad(!recv_sys.recovery_on)`

storage/innobase/log/log0recv.cc
```diff
@@ -3811,6 +3811,8 @@ void recv_sys_t::apply(bool last_batch)
 
   mysql_mutex_lock(&mutex);
 
+  if (last_batch)
+    dblwr.pages.clear();
```
dr-m, line 3815: I thought that we could do this inside the existing `if (last_batch)` earlier in this function. There is no need to introduce one more condition. Already at that point we should have finished applying log records (and reading pages).

storage/innobase/log/log0recv.cc
```diff
@@ -4841,6 +4843,50 @@ bool recv_dblwr_t::validate_page(const page_id_t page_id, lsn_t max_lsn,
...
+  const lsn_t max_lsn{log_sys.get_lsn()};
+  for (list::iterator page_it= pages.begin(); page_it != pages.end();
+       page_it++)
+  {
+    const byte *page= *page_it;
+    if (page_get_page_no(page) != page_no)
+      continue;
+    const lsn_t lsn= mach_read_from_8(page + FIL_PAGE_LSN);
+    if (buf_page_is_corrupted(lsn <= max_lsn, page, node.space->flags))
+      continue;
```
dr-m, line 4862: If `lsn` is in the future, we would seem to wrongly approve the copy of the page in the doublewrite buffer. We must only approve pages that are between the latest checkpoint LSN and the end of the log. This filtering can actually be done already in `buf_dblwr_t::recover()`.

storage/innobase/srv/srv0start.cc
```diff
@@ -1668,6 +1665,10 @@ dberr_t srv_start(bool create_new_db)
 
 			srv_log_file_size = srv_log_file_size_requested;
 
+			mysql_mutex_lock(&log_sys.mutex);
+			recv_sys.recovery_on = false;
+			mysql_mutex_unlock(&log_sys.mutex);
```
dr-m, line 1670: The comment in `log0recv.h` claims that `recv_sys.recovery_on` is protected by `recv_sys.mutex`.

I don’t think that we need any mutex here. When this log resizing is being executed on startup, the buffer pool should be completely idle (no reads or writes pending). You should omit the mutex acquisition here and amend the comment.

Is this assignment really needed here either? If it is, I would suggest to add the assignment near the following statements earlier in this code path:
```c++
			ut_d(recv_no_log_write = true);
			ut_ad(!os_aio_pending_reads());
```
Also, add a comment why the assignment is needed. Normally the assignment is part of `recv_sys.debug_free()`.

storage/innobase/log/log0recv.cc
```diff
@@ -4841,6 +4842,47 @@ bool recv_dblwr_t::validate_page(const page_id_t page_id, lsn_t max_lsn,
...
+    sql_print_information("InnoDB: Recovered page [page id: space="
+                          UINT32PF ", page number=" UINT32PF "] "
+                          "to '%s' from the doublewrite buffer.",
+                          uint32_t(node.space->id), page_no,
+                          node.name);
+    mysql_mutex_unlock(&recv_sys.mutex);
+    return true;
+  }
+  mysql_mutex_unlock(&recv_sys.mutex);
+  return false;
```
dr-m, line 4883: Can we avoid duplicating the `mysql_mutex_unlock()` call?

storage/innobase/log/log0recv.cc
```diff
@@ -4841,6 +4842,47 @@ bool recv_dblwr_t::validate_page(const page_id_t page_id, lsn_t max_lsn,
...
+    if (invalidate ||
+        mach_read_from_4(buf + FIL_PAGE_SPACE_ID) != node.space->id)
+      continue;
+
+    pages.erase(page_it);
+    sql_print_information("InnoDB: Recovered page [page id: space="
+                          UINT32PF ", page number=" UINT32PF "] "
+                          "to '%s' from the doublewrite buffer.",
+                          uint32_t(node.space->id), page_no,
+                          node.name);
```
dr-m, line 4878: This could be invoked after releasing the mutex. We could assign `buf= nullptr` when no match is found, and do `if (buf) sql_print_information(…)` after releasing the mutex.

storage/innobase/include/log0recv.h
```diff
@@ -150,6 +150,17 @@ struct recv_dblwr_t
...
+  /** Find the doublewrite copy of an encrypted page with the
+  smallest FIL_PAGE_LSN that is large enough for recovery.
+  @param space    tablespace object
+  @param page_no  page number to find
+  @param buf      unencrypted page
+  @retval true if encrypted page found in doublewrite buffer
+  @retval false otherwise */
+  bool find_encrypted_page(const fil_node_t &space,
+                           uint32_t page_no,
+                           byte *buf);
```
dr-m, line 162: I’d suggest the following:
```c++
/** …
@param buf    buffer for unencrypted page
@return buf
@retval nullptr  if the page was not found in the doublewrite buffer */
byte *find_encrypted_page(const fil_node_t &space, uint32_t page_no, byte *buf) noexcept;
```

storage/innobase/log/log0recv.cc
```diff
@@ -3785,6 +3785,7 @@ void recv_sys_t::apply(bool last_batch)
       /* We skipped this in buf_page_create(). */
       mlog_init.mark_ibuf_exist();
     mlog_init.clear();
+    dblwr.pages.clear();
```
dr-m, line 3788: Is there any reason why we couldn’t assign `recovery_on= false` here?

storage/innobase/log/log0recv.cc
```diff
@@ -4841,6 +4842,47 @@ bool recv_dblwr_t::validate_page(const page_id_t page_id, lsn_t max_lsn,
...
+  mysql_mutex_lock(&recv_sys.mutex);
+  for (list::iterator page_it= pages.begin(); page_it != pages.end();
+       page_it++)
+  {
+    const byte *page= *page_it;
+    if (page_get_page_no(page) != page_no ||
+        buf_page_is_corrupted(true, page, node.space->flags))
+      continue;
+    memcpy(buf, page, node.space->physical_size());
+    buf_tmp_buffer_t* slot= buf_pool.io_buf_reserve(false);
```
dr-m, line 4860: The space should be before the unary `*`, not after it.

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -465,6 +474,8 @@ void buf_dblwr_t::recover()
   }
 
   recv_sys.dblwr.pages.clear();
+  for (auto it : encrypted_pages)
+    recv_sys.dblwr.pages.push_back(it);
```
dr-m, line 478: Usually the variable `it` refers to `iterator`, while here it is referring to a data item. I would suggest to write
```c++
  for (byte *page : encrypted_pages)
    recv_sys.dblwr.pages.push_back(page);
```

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -385,8 +386,16 @@ void buf_dblwr_t::recover()
     fil_space_t *space= fil_space_t::get(space_id);
 
     if (!space)
-      /* The tablespace that this page once belonged to does not exist */
+    {
+      /* These pages doesn't belong to existing tablespace.
+      There is a possibility that these pages could be
```
dr-m, line 391: We’re processing one page at a time, so I would suggest to write in singular form:
```c
/* This page does not appear to belong to any tablespace.
There is a possibility that this page … */
```

dr-m (commented):
I didn’t check the test case yet. I see that `buf_dblwr_t::recover()` is already validating the unencrypted `FIL_PAGE_LSN`.

storage/innobase/log/log0recv.cc
Thirunarayanan: In case of log file resizing during restart, InnoDB expects the `recv_sys.recovery_on` to be true in `srv_prepare_to_delete_redo_log_file()`

mysql-test/suite/encryption/t/doublewrite_debug.test
```diff
@@ -0,0 +1,223 @@
...
+    close(FILE);
+    exit 0;
+}
+die "Did not find the page in the doublewrite buffer ($d1,$d2)\n";
+EOF
+
+let $restart_parameters=;
+--source include/start_mysqld.inc
+let SEARCH_FILE= $MYSQLTEST_VARDIR/log/mysqld.1.err;
+let SEARCH_PATTERN=InnoDB: Encrypted page \\[page id: space=[1-9]*, page number=3\\] in file .*test.t[1].ibd looks corrupted;
```
dr-m, line 215: `[1]` could be written as just `1`.

dr-m (commented):
I have now reviewed the test case as well. It is only testing with `FIL_PAGE_LSN=~0ULL` and not with an LSN that is earlier than the checkpoint LSN. I think that it is sufficient, given that the same `if` statement will be testing for both too small and too large LSN.

storage/innobase/log/log0recv.cc
dr-m: Based on the assertion failure during log resizing in https://buildbot.mariadb.org/#/builders/640/builds/8172/steps/7/logs/stdio
it looks like we need `recv_sys.recovery_on=false` in order to prevent any log to be written after a page flush batch has completed. So, we can’t add the assignment here. The assignment at a later point needs to be accompanied with a comment that explains this.

dr-m: Sorry, I meant that we need `recv_sys.recovery_on=true` to disable any checkpoint related log writes in `buf_flush_page_cleaner`. If we are rebuilding the redo log during startup not because of a size change, but because of a format change, we want to avoid any log corruption (6e5333fc8cf9efaee2abc7e34982a862c2268284) due to writing new-format log records to an old log file.

mysql-test/suite/encryption/t/debug_key_management.test
```diff
@@ -3,6 +3,7 @@
 -- source include/innodb_undo_tablespaces.inc
 -- source include/not_embedded.inc
 
+call mtr.add_suppression("InnoDB: Encrypted page \\[page id: space=[1-9][0-9]*, page number=[0-9]*\\] in file .*");
```
dr-m, line 6: Trailing `.*` in the regexp is redundant.

mysql-test/suite/encryption/t/doublewrite_debug.test
```diff
@@ -0,0 +1,223 @@
+--source include/have_innodb.inc
+--source include/have_debug.inc
+--source include/not_embedded.inc
+--source include/have_example_key_management_plugin.inc
+call mtr.add_suppression("InnoDB: Encrypted page \\[page id: space=[1-9][0-9]*, page number=3\\] in file .*test.t[12].ibd looks corrupted");
```
dr-m, line 5: Actually it should be `\\.ibd` instead of `.ibd` because we are looking for a literal `.` in the file name, instead of just any character. But, I don’t feel strongly about this little bit of sloppiness.

mysql-test/suite/encryption/t/doublewrite_debug.test
```diff
@@ -0,0 +1,223 @@
...
+call mtr.add_suppression("InnoDB: Unable to apply log to corrupted page ");
```
dr-m, line 6: This is always page 3, right?

storage/innobase/srv/srv0start.cc
dr-m: The pre-existing comment is better, but it could be amended to be more descriptive:
```c++
	/* Enable log_checkpoint() in buf_flush_page_cleaner().
	If we are upgrading or resizing the log at startup, we must not
	write any FILE_MODIFY or FILE_CHECKPOINT records to the old log file
	because the ib_logfile0 could be in an old format that we can only
	check for emptiness, or the write could lead to an overflow. */
```

dr-m (approved):

storage/innobase/srv/srv0start.cc
dr-m: It looks like my request to improve this comment was disregarded in f8cf4932903aa3e14076357cdfc9abed01c390b6.

dr-m: I would revise this comment as well as a comment about `recv_sys_t::recovery_on` on a merge 435c6a399cae175f2e75c9ab594865408928e923.

dr-m: I revised the comment in the merge to 10.11: 420d9eb27fca033079ddf0465eb91e4e4f2b7609

PR #3694, opened 2024-12-12
MDEV-35614 json unquote utf8mb4

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35614*

## Description

No smiley emojis in JSON_UNQUOTE.

## Release Notes

Smilieys are back.

## How can this PR be tested?

basic test case there. Need to do more (and probably split out non-json commits)

## PR quality check


sql/item_jsonfunc.cc
```diff
@@ -3931,13 +3979,23 @@ int Arg_comparator::compare_json_str_basic(Item *j, Item *s)
...
                                je.value + je.value_len,
-                               &my_charset_utf8mb3_general_ci,
+                               &my_charset_utf8mb4_bin,
                                (uchar *) value2.ptr(),
                                (uchar *) (value2.ptr() + je.value_len))) < 0)
+       {
+         if (current_thd)
+           push_warning_printf(current_thd, Sql_condition::WARN_LEVEL_WARN,
+                               ER_JSON_BAD_CHR, ER_THD(current_thd, ER_JSON_BAD_CHR),
+                               0, "comparison", (int)((const char *) je.s.c_str - js->ptr()));
```
vuvova, line 3996: is there a test case for that?

sql/item_jsonfunc.cc
```diff
@@ -3980,13 +4038,23 @@ int Arg_comparator::compare_e_json_str_basic(Item *j, Item *s)
...
                               (uchar *) value+value_len,
-                              &my_charset_utf8mb3_general_ci,
+                              &my_charset_utf8mb4_bin,
                               (uchar *) value1.ptr(),
                               (uchar *) (value1.ptr() + value_len))) < 0)
+    {
+      if (current_thd)
+        push_warning_printf(current_thd, Sql_condition::WARN_LEVEL_WARN,
+                            ER_JSON_BAD_CHR, ER_THD(current_thd, ER_JSON_BAD_CHR),
+                            0, "equality comparison", 0);
```
vuvova, line 4055: is there a test case for that?

strings/json_lib.c
```diff
@@ -1562,15 +1562,7 @@ int json_unescape(CHARSET_INFO *json_cs,
...
-      /*
-        Result charset doesn't support the json's character.
-        Let's replace it with the '?' symbol.
-      */
-      if ((c_len= my_ci_wc_mb(res_cs, '?', res, res_end)) > 0)
-      {
-        res+= c_len;
-        continue;
-      }
+      return -1;
```
vuvova, line 1565: it looks like there **is** a test case for that, but, please, confirm

vuvova (changes requested):
questions re. test cases.
and please, rebase it on top of 10.11

sql/item_jsonfunc.cc
grooverdan: No. In looking for one, it has to:
* be comparison like `SELECT JSON_EXTRACT('{"n": "aaaa\\ud834\\udd1e"}', '$.n') = 'value' as invalid_utf8mb4;`
* pass the current connection character set. this is validated in the `j->val_str` at beginning of the function, easy - set to binary
* pass the escape syntax which is valid utf16 which the two code points above are.
* by the time it passes to my_wc_mb_utf8mb4 to validate the utf8mb4, the byte seqeunce has been converted its high/low surrages have been converted down to the valid utf8mb4:
```
Thread 4 "mariadbd" hit Breakpoint 9, json_handle_esc (s=0x7f043a1ea8a8) at /home/dan/repos/mariadb-server-rebase/strings/json_lib.c:322
322	    uchar code[4]= {0,0,0,0};
(gdb) c
Continuing.

Thread 4 "mariadbd" hit Breakpoint 8, my_wc_mb_utf8mb4 (cs=0x6aa4db8 <my_charset_utf8mb4_bin>, wc=119070, r=0x7f0424029f1c "𝄞", e=0x7f0424029f28 "")
    at /home/dan/repos/mariadb-server-rebase/strings/ctype-utf8.c:3056
3056	  if (r >= e)
(gdb) p/x wc
$38 = 0x1d11e
```

So my chatgpt ai says that there isn't a high/low surrogate when resolve that forms a character that isn't valid utf8.  So at the moment, code unreachable.

strings/json_lib.c
grooverdan: nope, there doesn't appear to be, at least in 10.11. based on above, I'm not sure it reachable.

sql/item_jsonfunc.cc
grooverdan: To clarify further the test case I added would have failed, but since utf8mb4 was set as the character set it didn't fail.

The number of times a validation occurs could be optimised and if it is, this codepath might be hit. It could be replaced with a debug assert, but its a test and function call so there's not much gain to optimise this further.

PR #3695, opened 2024-12-12
MDEV-35633 - Setting s3_protocol_version on MariaDB Server 10.5 has no effect

RazvanLiviuVarzaru:
- [x] *The Jira issue number for this PR is: MDEV-35633*

## Description
Bump libmarias3 submodule to libmarias3-3.1.2-37-g289108f.

In **libmarias3 3.1.2 (3846890)**, `MS3_OPT_FORCE_PROTOCOL_VERSION` overwrites the wrong variable hence `s3_protocol_version` has no effect, and will default to protocol_version 2 when using a hostname instead of an IP. This could lead to wrong request headers when using S3 providers that require the version 1 of the protocol.

**This commit is only applicable to 10.5 branch and one should not apply it to newer versions because it will cause a downgrade of libmarias3.**

## Release Notes
Fixed a bug where s3_protocol_version had no effect on forcing the version.

## How can this PR be tested?

Run MTR with `--suite=s3`, where protocol_version 1 (`S3_PROTOCOL_VERSION=Original`) is needed and use a hostname instead of an IP (for example: `S3_HOST_NAME=minio.mariadb.org`)

## PR quality check


grooverdan:
Definately too late. Won't fix per MDEV.

PR #3696, opened 2024-12-12
MDEV-35638 wsrep_last_{written,seen}_gtid call my_error(ER_OUTOFMEMORY) incorrectly

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35638*

## Description

The my_error has a flags argument before the varargs size, which in ER_OUTOFMEMORY is a %d. Pass in the size, corrected for what was attempted to be allocated.

For wsrep_sync_wait_upto we error a size roughly of a gtid, though the allocation that failed in wait_gtid_upto was doing:

   "m_wait_map.insert(std::make_pair(seqno, &wait_cond))"

which fails. Rather than working out the exact size, which won't be large and not actionably by the user, doing an approximation instead.

## Release Notes

Correct OOM messages if they occured during wsrep_last_{writtent,seen}_gtid functions.

## How can this PR be tested?

Not easily, manual memory limited, but the calling convention on the error message is quite clear.

## PR quality check

### DIFF
--- a/sql/item_strfunc.cc
+++ b/sql/item_strfunc.cc
@@ -5327,7 +5327,7 @@ String *Item_func_wsrep_last_written_gtid::val_str_ascii(String *str)
 {
   if (gtid_str.alloc(WSREP_MAX_WSREP_SERVER_GTID_STR_LEN+1))
   {
-    my_error(ER_OUTOFMEMORY, WSREP_MAX_WSREP_SERVER_GTID_STR_LEN);
+    my_error(ER_OUTOFMEMORY, MYF(0), WSREP_MAX_WSREP_SERVER_GTID_STR_LEN+1);
     null_value= TRUE;
     return 0;
   }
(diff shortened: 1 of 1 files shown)


janlindstrom (approved):
LGTM.

sql/item_strfunc.cc
```diff
@@ -5414,7 +5414,8 @@ longlong Item_func_wsrep_sync_wait_upto::val_int()
       }
       else if (wait_gtid_ret == ENOMEM)
       {
-        my_error(ER_OUTOFMEMORY, MYF(0), func_name());
+        /* length is bogus here, no idea how much it tried */
+        my_error(ER_OUTOFMEMORY, MYF(0), WSREP_GTID_STR_LEN);
         ret= 0;
       }
     }
```
knielsen, line 5421: You could also use here instead the error ER_OUT_OF_RESOURCES (which doesn't take the size).

knielsen (approved):
The patch is ok, just one comment you can consider.

But this is too minor for old GA release, let's do 11.4 or latest development branch.

Ok to push after changing destination branch to main/11.4.

vuvova:
@grooverdan could you please rebase it on the latest 10.5 and resolve the conflicts?

grooverdan:
updated, but was fixed in 765458c93cd5 applied at a very different date to commit.

PR #3698, opened 2024-12-13
MDEV-35639 func sha2/char - handle OOM

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35639*

## Description

The SQL functions sha2 and char perform string allocations without returning
a SQL error if these allocations fail.

Corrected so SQL errors are returned if the reallocation failed.

## Release Notes

Handle sha2/char function out of memory errors on the chance reallocation of memory fails.

## How can this PR be tested?

With difficulty. The error handling is pretty straight forward.

## PR quality check

### DIFF
--- a/sql/item_strfunc.cc
+++ b/sql/item_strfunc.cc
@@ -271,7 +271,12 @@ String *Item_func_sha2::val_str_ascii(String *str)
     Since we're subverting the usual String methods, we must make sure that
     the destination has space for the bytes we're about to write.
   */
-  str->alloc((uint) digest_length*2 + 1); /* Each byte as two nybbles */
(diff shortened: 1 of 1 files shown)


cvicentiu (changes requested):
Hi Daniel!

Good catch for sha2. You can test this (at least manually) by just faking an OOM for str->alloc. Like so:

 ```C++
  if (str->alloc((uint) digest_length*2 + 1) || true) /* Each byte as two nybbles */
 ```
 
 For the other realloc in `Item_func_char`, however, at least for this function you need to fix append_char too. Here's an idea how I would optimize the function:
 
What if we prealloc: `arg_count * 4 + 1` at the start of val_str, as that's the maximum length the str can end up storing. In that case you just need to check *that* function's error code.

PR #3700, opened 2024-12-13
MDEV-28649 json_array(false) in the definition of view return 0 inste…

grooverdan:
…ad of "true"

- [x] *The Jira issue number for this PR is: MDEV-28649*

## Description

Using any boolean argument to a function in a view resulted in a view definition like:

  select json_array(0) AS `json_array(false)`

This was imply the Item::print function used was from Item_int::print rather than a specific Item_bool::print function.

Simplification to STRING_WITH_LEN noted from Alexey Botchkov's patch after almost equivalent implementation.
[6835d2ce268b001a9a4f85d92c132c58e33d490e](https://github.com/MariaDB/server/commit/6835d2ce268b001a9a4f85d92c132c58e33d490e)

Explain FORMAT=JSON output to take boolean output rather than printing "true"/"false" in quotes.

## Release Notes

View statements created with functions that take booleans now are accurately represented as booleans.

Also json output containing booleans are displayed as booleans. 

## How can this PR be tested?

test included

## PR quality check

### DIFF
--- a/mysql-test/main/analyze_engine_stats.result
+++ b/mysql-test/main/analyze_engine_stats.result
@@ -138,7 +138,7 @@ ANALYZE
     "select_id": 1,
     "r_loops": 1,
     "r_total_time_ms": "REPLACED",
-    "const_condition": "true",
+    "const_condition": true,
     "nested_loop": [
       {
         "table": {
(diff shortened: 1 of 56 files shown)


grooverdan:
note in #3708 `"const_condition": true,` aka `"const_condition": "1"` goes away creating less test changes, and making `sql/sql_explain.cc` from something to make the JSON valid, to an optional extra that could be put in a new MDEV feature request.

PR #3706, opened 2024-12-15
MDEV-35338 - Non-copying ALTER does not pad VECTOR column, vector sea…

svoj:
…rch further does not work

Since VECTOR data type is based on VARCHAR, ALTER TABLE expanding VECTOR column was allowed to go with ALGORITHM=INPLACE.

However InnoDB sees such columns as VARCHAR and thus it doesn't pad updated columns data to new length. In contrast to ALGORITHM=COPY, which goes with field copy routines.

With this patch ALGORITHM=INPLACE is not allowed for VECTOR columns.

- [x] *The Jira issue number for this PR is: MDEV-35338*

## How can this PR be tested?
mtr test added

## PR quality check

### DIFF
--- a/mysql-test/main/vector_innodb.result
+++ b/mysql-test/main/vector_innodb.result
@@ -227,3 +227,17 @@ connection con1;
 ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
 drop table t;
 disconnect con1;
+connection default;
+#
+# MDEV-35338 - Non-copying ALTER does not pad VECTOR column,
+#              vector search further does not work
+#
+create or replace table t (a int, v vector(1) not null, primary key (a)) engine=InnoDB;
+insert into t values (1,0x38383838),(2,0x37373737),(3,0x31313131);
+alter table t modify v vector(2) not null;
+select hex(v) from t order by a;
+hex(v)
+3838383800000000
+3737373700000000
+3131313100000000
+drop table t;
--- a/mysql-test/main/vector_innodb.test
+++ b/mysql-test/main/vector_innodb.test
@@ -231,3 +231,14 @@ rollback;
 --reap
 drop table t;
 --disconnect con1
+connection default;
+
+--echo #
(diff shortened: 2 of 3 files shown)


sql/sql_table.cc
```diff
@@ -6904,7 +6905,8 @@ static bool fill_alter_inplace_info(THD *thd, TABLE *table,
 
       if (!is_equal)
       {
-        if (field->table->file->can_convert_nocopy(*field, *new_field))
+        if (!dynamic_cast<const Field_vector *>(field) &&
+            field->table->file->can_convert_nocopy(*field, *new_field))
```
vuvova, line 6909: I'd say it's InnoDB limitation. Technically, it can implement INPLACE with extending and padding of vectors. But it doesn't.

Meaning, the check that disallows vectors should be inside `ha_innobase::can_convert_nocopy`, not here.

svoj: @vuvova we feed `VARCHAR` to InnoDB and expect it to behave as if it was `CHAR`. To perform padding in this case InnoDB (as well as any other engine) will have to have extra code.

I'd say it is a special case of SQL layer, and it looks consistent when SQL layer handles it.

I could alternatively allow `INPLACE` and fix it by implementing `Field_vector::val_str()` or adding special code to InnoDB so that it can distinguish `VECTOR` type and pad it. So we have four options, either is fine with, just confirm which one you like more.

Btw, given that it is supposed to be fixed length, why it inherits from `Field_varstring` rather than from `Field_string`? I assume length limit?

vuvova: Yes, length limit.

`Field_vector::val_str()` would mean to pay the performance price for every row during normal operations — I think your approach of disabling INPLACE is better, as it means that only `ALTER TABLE` will get slower.

Agree, InnoDB could have special code to handle `VECTOR`, but as long as it doesn't it (that is, InnoDB) should not attempt to alter them in-place. The server cannot know if InnoDB implements this special code or not, so the server should not disallow that.

vuvova (approved):
looks good, thanks!

PR #3707, opened 2024-12-16
MDEV-35519 Populate thread_id of Backward Compatibility Query event

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-35519](https://jira.mariadb.org/browse/MDEV-35519)*

## Description
This commit bubbles `info->fdev` to `Query_log_event::begin_event` so it can reconstruct the GTID event from binlog for *real* conversion.

For replicas that don't support GTIDs, a MariaDB primary will “convert” (read: replace) `Gtid_log_event`s into `Query_log_event`s of BEGIN queries.
The two events’ subheaders had no common fields previously, so this “conversion” simply zeroes all fields, `thread_id` among them.
That background changed when [MDEV-7850](https://jira.mariadb.org/browse/MDEV-7850) added `thread_id` records to GTID events as well, and now this “conversion” is dropping this field.

## Release Notes
Compatibility replication to pre-GTID replicas now include Thread IDs.

## How can this PR be tested?
**TODO**

> And note too, that we don't have any MTR way to test backwards compatibility to Mysql.
> ⸺ @bnestere on Slack DM

Testing this conversion (in general) in MTR would require somehow tricking a GTID-oriented primary to think it’s replicating to a old-Maria/MySQL replica.

## PR quality check


sql/log_event.cc
```diff
@@ -1906,16 +1906,18 @@ Query_log_event::dummy_event(String *packet, ulong ev_offset,
...
 */
 int
 Query_log_event::begin_event(String *packet, ulong ev_offset,
+                             Format_description_log_event* fdev,
                              enum_binlog_checksum_alg checksum_alg)
 {
   uchar *p= (uchar *)packet->ptr() + ev_offset;
   uchar *q= p + LOG_EVENT_HEADER_LEN;
   size_t data_len= packet->length() - ev_offset;
+  Gtid_log_event original_event= Gtid_log_event(p, data_len, fdev);
```
ParadoxV5, line 1920: `Query_log_event::begin_event` occurs during pre-send, which is layers away from where the GTID event originates.
Persumably, this is because replication works by sending primaries’ already-serialized binary logs to replicas as relay logs.

The GTID event has a fixed-size format comprising two variable-sized sections, the thread_id being the very last optional entry.
This means, if I don’t leverage the existing constructor, I’ll have to duplicate its parsing code.

Let’s pretend that, instead of overwriting the buffer with a blank `BEGIN` query, we’re finally starting to *convert* the GTID event for real.

sql/log_event.cc
```diff
@@ -1932,7 +1934,7 @@ Query_log_event::begin_event(String *packet, ulong ev_offset,
   int2store(p + FLAGS_OFFSET, flags);
 
   p[EVENT_TYPE_OFFSET]= QUERY_EVENT;
```
ParadoxV5, line 1936: Is `LOG_EVENT_OFFSET` a duplicate of `EVENT_TYPE_OFFSET`?

sql/log_event.cc
ParadoxV5, line 1920: `packet->length()` is a `uint32`, `ev_offset` is a `ulong`, their difference is stashed into `size_t 
 data_len` and it’s then passed to `Gtid_log_event(const uchar, uint, const Format_description_log_event*)`.
[What is this madness?](https://jira.mariadb.org/browse/MDEV-26896)

sql/log_event.cc
bnestere: I think it is actually meant to refer to the [binlog magic header](https://mariadb.com/kb/en/1-binlog-events/#log-file-structure), though it seems that has somewhere along the way been confused with `EVENT_TYPE_OFFSET`. I wonder if `LOG_EVENT_OFFSET` should be removed, and its usages either changed to (depending on the usage) `EVENT_TYPE_OFFSET` (for extracting event type)  or `BIN_LOG_HEADER_SIZE` (for usage in `Rotate_log_event` construction).

@andrelkin , @knielsen any thoughts?

sql/log_event.cc
```diff
@@ -1932,7 +1934,7 @@ Query_log_event::begin_event(String *packet, ulong ev_offset,
...
-  int4store(q + Q_THREAD_ID_OFFSET, 0);
+  int4store(q + Q_THREAD_ID_OFFSET, original_event.thread_id);
```
bnestere, line 1937: Let's add an extra condition here, to ensure `original_events.flags_extra & FL_EXTRA_THREAD_ID` must be set, in order to use `original_event.thread_id`, and otherwise `0`. Where the current state of the code _should_ guarantee that, let's make this check just in case, as we don't have any MTR testing of this, and we don't want this to silently break from future changes.

This is so we don't overwrite the `thread_id` for events that originate before MDEV-7850, e.g. from old binary logs, or in chain replication setups like `MariaDB 11.4` -> `MariaDB 11.5+` -> `MySQL 5.7`.

sql/log_event.cc
ParadoxV5:
```console
$ git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
$ git grep -n LOG_EVENT_OFFSET
sql/log.cc:5904:    Rotate_log_event r(new_name + dirname_length(new_name), 0, LOG_EVENT_OFFSET,
sql/log_event.h:95:#define LOG_EVENT_OFFSET 4
sql/sql_repl.cc:2272:  DBUG_PRINT("info", ("log event code %d", (*packet)[LOG_EVENT_OFFSET+1] ));
sql/sql_repl.cc:2539:  event_type= (Log_event_type)((uchar)(*packet)[LOG_EVENT_OFFSET+ev_offset]);
sql/sql_repl.cc:2674:  event_type= (Log_event_type)((uchar)(*packet)[LOG_EVENT_OFFSET + ev_offset]);
sql/sql_repl.cc:2943:        (Log_event_type)((uchar)(*packet)[LOG_EVENT_OFFSET+ev_offset]);
```

andrelkin: > LOG_EVENT_OFFSET should be removed ..?

Right.

bnestere:
> [knielsen](https://github.com/knielsen) commented [last week](https://github.com/MariaDB/server/pull/3707#issuecomment-2567147425) via email

I didn't realize email replies are added into the conversation chain here, nice!

But back on topic:

> First, I wonder if this change is really worth it?

That's a fair question. No one has requested the feature, but I was thinking it would be an easy win (and learning exercise) while interoperability is in the light. I had the thought that Mariadb -> MySQL 8 might still use this feature too, but now having looked into that, it wouldn't, as MySQL uses `Anonymous_Gtid` events to begin its transactions without their GTIDs.

Though I think the `LOG_EVENT_OFFSET` cleanup would be good to have regardless (perhaps as a separate PR then).

I also think this brings up another question -- how long do we want to keep this pre-GTID support code around, where all such servers are EoL?

knielsen:

>> [knielsen](https://github.com/knielsen) commented [last

>> First, I wonder if this change is really worth it?
>
> That's a fair question. No one has requested the feature, but I was
> thinking it would be an easy win (and learning exercise) while

> interoperability is in the light. I had the thought that Mariadb ->
> MySQL 8 might still use this feature too, but now having looked into
> that, it wouldn't, as MySQL uses `Anonymous_Gtid` events to begin its
> transactions without their GTIDs.

I think that MariaDB -> MySQL replication would use this, and it will
provide the thread_id in the BEGIN event for the MySQL slave (in fact that
is the only realistic case I can think of where it would be used?). So if
anyone wants the thread_id available there on MySQL replicating from
MariaDB, that would be a use-case.

> Though I think the `LOG_EVENT_OFFSET` cleanup would be good to have
> regardless (perhaps as a separate PR then).

Yes, agree.

> I also think this brings up another question -- how long do we want to
> keep this pre-GTID support code around, where all such servers are
> EoL?

Well, it will need to be around for as long as we want to support
replication to MySQL? Because MariaDB always logs a GTID event at the start
of an event group, but MySQL expects a BEGIN query event. And this pre-GTID
support code provides exactly that, replacing the GTID event (which is
unknown to MySQL) with the BEGIN event that it expects.

 - Kristian.

ParadoxV5:
> I think that MariaDB -> MySQL replication would use this, and it will provide the thread_id in the BEGIN event for the MySQL slave (in fact that is the only realistic case I can think of where it would be used?). So if anyone wants the thread_id available there on MySQL replicating from MariaDB, that would be a use-case.

Does this apply?
https://github.com/MariaDB/server/blob/74331a48dd3dadfe5041f3440b81a7183e9fd4a8/sql/log_event_server.cc#L1009-L1025

ParadoxV5:
TL;DR: No bug, but a feature that has no value currently. Archived

PR #3709, opened 2024-12-16
MDEV-35103 CREATE TABLE AS SELECT ST_VALIDATE creates a column with d…

DaveGosselin-MariaDB:
…atatype point, not geometry

Changes the type handler for geometry validate from 'point' to 'geometry'


grooverdan (commented):
For consistency I'd rather the test cases ended up in mysql-test/main/spatial_utility_function_isvalid.test.

Otherwise free to merge/push this simple fix.

PR #3713, opened 2024-12-18
MDEV-35683: add basic unit test for DYNAMIC_ARRAY

ericherman:
- [x] *The Jira issue number for this PR is: MDEV-35683*

## Description

This demonstrates and tests the basic usage of DYNAMIC_ARRAY.

Additional tests could be added to demonstrate and test the remaining functions and capabilities.

## Release Notes

No release notes are needed for this change, as this only adds a test.

## How can this PR be tested?

build and then run:

unittest/mysys/dynamic_array-t

## PR quality check

### DIFF
--- a/unittest/mysys/dynamic_array-t.c
+++ b/unittest/mysys/dynamic_array-t.c
@@ -1,0 +1,246 @@
+/* Copyright (c) 2024 Eric Herman and MariaDB Foundation.
+
+   This program is free software; you can redistribute it and/or
+   modify it under the terms of the GNU General Public License as
+   published by the Free Software Foundation; version 2 of the License.
+
+   This program is distributed in the hope that it will be useful, but
+   WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1335  USA */
+
+#include <my_global.h>
+#include <my_sys.h>
+#include <tap.h>
+
+struct thing {
+  unsigned id;
+  char name[40];
+};
+
+#define THING_STACK_BUF_SIZE 3
+#define ROUND_TRIP_NO_GROW_SIZE (THING_STACK_BUF_SIZE - 1)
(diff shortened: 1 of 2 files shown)


grooverdan:
Thanks @ericherman. Please correct the Windows compile error. I've taking note of the other unrelated tests failures.

unittest/mysys/dynamic_array-t.c
```diff
@@ -0,0 +1,235 @@
...
+  ok(grow_to_size == things_array.elements, "size expect: %zu, actual: %zu",
+     grow_to_size, things_array.elements);
+
+  for (size_t i= 0; i < things_array.elements; ++i) {
+    struct thing retrieved;
+    memset(&retrieved, 0x00, sizeof(struct thing));
+
+    get_dynamic(&things_array, &retrieved, i);
+    ok(retrieved.id == i, "thing %zu: id: %u, name: '%s'",
+       i, retrieved.id, retrieved.name);
```
grooverdan, line 73: trusting it retrieved the name correctly?

grooverdan (changes requested):

unittest/mysys/dynamic_array-t.c
```diff
@@ -0,0 +1,235 @@
...
+     0, things_array.elements);
+
+  return EXIT_SUCCESS;
+}
+
+int main(void)
+{
+  int err= 0;
+
+  plan((3 + (2 * (THING_STACK_BUF_SIZE - 1)) + 1)
```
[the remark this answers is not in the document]
ericherman: I think above the calls to the functions rather than above the functions themselves, since the plan values are variable, but yes, I can think I can make this easier to follow.

unittest/mysys/dynamic_array-t.c
ericherman: I view the intent of the test is to check the order.

Do you think there is a test-worthy scenario to guard against where it would have the id, but not the name?
Like perhaps where a struct could be partially copied in to the DYNAMIC_ARRAY?

grooverdan: Thanks for addressing review comments to make it a easier to change later.

Noted on order.

Yes, I was considering a bounds check on the contents of the array element copied back/forward. Checking bounds is going to be expanding the contents etc so while trivial, it can be added later.

While I'm not on leave, and I remember the good status of this is, I'm going to merge this now. If you want to expand this to include bounds checking feel free to do so in a new PR under same MDEV.

ericherman:
thanks @grooverdan !

PR #3718, opened 2024-12-19
MDEV-34924 : gtid_slave_pos table neven been deleted on non replica n…

janlindstrom:
…odes (wsrep_gtid_mode = 1)

- [x] *The Jira issue number for this PR is: MDEV-34924*

## Description

Problem was caused by MDEV-31413 commit 277968aa where mysql.gtid_slave_pos table was replicated by Galera. However, as not all nodes in Galera cluster are replica nodes, rows were not deleted from table.

In this fix this is corrected so that mysql.gtid_slave_pos table is not replicated by Galera. Instead when Galera node receives GTID event and wsrep_gtid_mode=1, this event is stored to mysql.gtid_slave_pos table

Added test case galera_2primary_replica for 2 async primaries replicating to galera cluster.

Added test case galera_circular_replication where async primary replicates to galera cluster and
one of the galera cluster nodes is masterto async replica.

Modified test case galera_restart_replica to monitor gtid positions and rows in mysql.gtid_pos_table

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/t/galera_2primary_replica.cnf
+++ b/mysql-test/suite/galera/t/galera_2primary_replica.cnf
@@ -1,0 +1,23 @@
+!include ../galera_2nodes_as_replica_2primary.cnf
+
+[mysqld]
+wsrep-debug=1
+
+[mysqld.1]
+server_id=15
+wsrep_gtid_mode=1
+wsrep_gtid_domain_id=16
+gtid_domain_id=11
+gtid_strict_mode=1
+wsrep-slave-threads=4
+slave-parallel-threads=2
+
+[mysqld.2]
+skip-slave-start=OFF
+server_id=15
+wsrep_gtid_mode=1
+wsrep_gtid_domain_id=16
+gtid_domain_id=11
+gtid_strict_mode=1
+wsrep-slave-threads=4
+
--- a/mysql-test/suite/galera/t/galera_restart_replica.test
+++ b/mysql-test/suite/galera/t/galera_restart_replica.test
@@ -1,77 +1,115 @@
 #
-# Test Galera as a replica to a MySQL async replication
+# Test Galera as a replica to a MariaDB async replication
+#
+#   MariaDB
+#    primary ---async replication--->galera node_2 (replica)<----galera replication---> galera node1
+#
+#  Test outline:
+#
+#  - Create user for async replication
+#  - Create table and some data in primary
+#  - Verify that table and data is replicated to galera nodes
+#  - Verify that mysql.gtid_slave_pos has some rows in all Galera nodes
+#  - Verify that gtid_slave_pos, gtid_binlog_pos and gtid_current_pos are
+#    same in all Galera nodes
+#  - Verify that we can shutdown and restart Galera replica (node #2)
+#  - Verify that gtid_slave_pos, gtid_binlog_pos and gtid_current_pos are
+#    same in all Galera nodes
+#  - Verify that mysql.gtid_slave_pos table has limited amount of rows
+#  - Veruft that ddl works (drop table)
+#
+#  Similar test is done so that new rows are added to table in
+#  primary while async replica (node #2) is down.
 #
 # The galera/galera_2node_slave.cnf describes the setup of the nodes
 #
 --source include/force_restart.inc
 --source include/galera_cluster.inc
 --source include/have_innodb.inc
---source include/have_sequence.inc
+
+# In this test we mark node #2 as replica
+--connect replica, 127.0.0.1, root, , test, $NODE_MYPORT_2
 
 # As node #3 is not a Galera node, and galera_cluster.inc does not open connetion to it
-# we open the node_3 connection here
---connect node_3, 127.0.0.1, root, , test, $NODE_MYPORT_3
+# we open the primary connection her
+--connect primary, 127.0.0.1, root, , test, $NODE_MYPORT_3
 
 create user repl@'%' identified by 'repl';
 grant all on *.* to  repl@'%';
 ALTER TABLE mysql.gtid_slave_pos ENGINE=InnoDB;
 
 --let $node_1 = node_1
---let $node_2 = node_2
+--let $node_2 = replica
 --source include/auto_increment_offset_save.inc
 
---connection node_2
+--connection replica
 --disable_query_log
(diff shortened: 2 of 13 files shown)


sql/wsrep_schema.cc
```diff
@@ -1535,3 +1537,54 @@ int Wsrep_schema::recover_sr_transactions(THD *orig_thd)
...
 }
+
+int Wsrep_schema::store_gtid_event(THD* thd,
+                                   const Gtid_log_event *gtid)
+{
+  int error=0;
+  void *hton= NULL;
+  const bool in_transaction= (gtid->flags2 & Gtid_log_event::FL_TRANSACTIONAL);
+  const bool in_ddl= (gtid->flags2 & Gtid_log_event::FL_DDL);
+  Wsrep_schema_impl::wsrep_off  wsrep_off(thd);
```
temeo, line 1548: Turning wsrep off here causes the first statement to run with wsrep OFF. If `mysql.gtid_slave_pos` table is InnoDB table, this causes the transaction to be started with wsrep OFF. As the wsrep status is synced only at the beginning of the transaction in InnoDB, this causes the rest of the transaction to run with wsrep OFF too.

temeo (changes requested):

sql/wsrep_schema.cc
```diff
@@ -1535,3 +1555,61 @@ int Wsrep_schema::recover_sr_transactions(THD *orig_thd)
...
+
+  if (in_ddl)
+  {
+    // Commit transaction if this GTID is part of DDL-clause because
+    // DDL causes implicit commit assuming there is no multi statement
+    // transaction ongoing.
+    if((error= trans_commit_stmt(thd)))
+      goto out;
+
+    (void)trans_commit(thd);
```
bnestere, line 1602: I'm not really familiar with the Galera code - when in the transaction life-cycle is `store_gtid_event()`/`wsrep_apply_events()` called? It looks to me like this is before any of the transaction events have actually been applied. If so, and then we update & commit `mysql.gtid_slave_pos` here before actually applying the DDL, then if something happens while the DDL is running and we restart, `gtid_slave_pos` would be ahead of the actual database state.

sql/wsrep_schema.h
```diff
@@ -133,6 +133,16 @@ class Wsrep_schema
   */
   int recover_sr_transactions(THD* orig_thd);
 
+  /**
+     Store GTID-event to mysql.gtid_slave_pos table.
```
bnestere, line 137: I think it would be good to extend this comment to explain why it is needed.

mysql-test/suite/galera/t/galera_restart_replica.test
```diff
@@ -1,102 +1,146 @@
...
+#
+#  Test outline:
+#
+#  - Create user for async replication
+#  - Create table and some data in primary
+#  - Verify that table and data is replicated to galera nodes
+#  - Verify that mysql.gtid_slave_pos has some rows in all Galera nodes
+#  - Verify that gtid_slave_pos, gtid_binlog_pos and gtid_current_pos are
+#    same in all Galera nodes
+#  - Verify that we can shutdown and restart Galera replica (node #2)
```
bnestere, line 15: In our previous email exchange, you noted an extended use case for this, where (adapting it to this test) `node_2` would go down, and there would be an SST from `node1`->`node_2` , and then `node_2` would need to restart at the correct slave position. I think it would be good to extend this (or one of your other) test cases to handle that.

mysql-test/suite/galera/t/galera_2primary_replica.cnf
```diff
@@ -0,0 +1,23 @@
...
+wsrep-debug=1
+
+[mysqld.1]
+server_id=15
+wsrep_gtid_mode=1
+wsrep_gtid_domain_id=16
+gtid_domain_id=11
+gtid_strict_mode=1
+wsrep-slave-threads=4
+slave-parallel-threads=2
```
bnestere, line 13: I think it would be good to set `gtid-cleanup-batch-size` differently on the mariadb nodes, to add a layer of verification on the reason of a lesser `gtid_slave_pos` table size. I' dsuggest, on one node, made it _really_ big, so the cleanup never happens; and on another node, keep it small. Then your validations would be more strict (rather than just `< 1000`).

On the node with a really big `gtid-cleanup-batch-size`, you could validate that `SELECT COUNT(*) FROM mysql.gtid_slave_pos == <domain_3_seqno> + <domain_4_seqno>` (note the `gtid-cleanup-batch-size` value should be much more than `<domain_3_seqno> + <domain_4_seqno>` (`2006`), as previously run tests can impact the cleanup time).

And on the node with a small `gtid-cleanup-batch-size`, you should be able to assert `SELECT COUNT(*) FROM mysql.gtid_slave_pos < @@gtid-cleanup-batch-size`

bnestere (commented):
Hi @janlindstrom !

In general I think the approach works. See my comments for some specific suggestions/questions.

Also, where your git commit message does a good job explaining _what_ the patch does, it doesn't cover _why_/_how_ this approach works (which I think it should).

mysql-test/suite/galera/t/galera_restart_replica.test
janlindstrom: See line 59, we intentionally force SST (here it does not matter which Galera node is donor) and at line 70 node_2 is restarted and because it has skip-slave-restart=0 slave thread is started.

sql/wsrep_schema.cc
janlindstrom: In case of DDL, it is problematic. If GTID event is executed before DDL, it will be committed anyway as DDL does implicit commit before start. IF GTID event is executed after DDL it is also not atomic with DDL.

sysprg:
@janlindstrom Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/133e26fd7dc454c5b154495ab876c76bbcfd1715
@bnestere I accept this patch as the current working version after your review which generally finds no problems here in light of Jan's explanations - since without fixes from this commit we will not close the current issues for customers and will not be able to move on to solving the next critical issue: https://jira.mariadb.org/browse/MDEV-20065 since the Codership team needs this fix as a prerequisite for a future PR on MDEV-20065. I propose that all future improvements, if we deem them necessary, be made as separate MDEVs.

PR #3720, opened 2024-12-20
MDEV-35701 trx_t::autoinc_locks causes unnecessary dynamic memory allocation

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35701*
## Description
`trx_t::autoinc_locks`: Use `small_vector<lock_t*,4>` in order to avoid any dynamic memory allocation in the most common case (a statement is holding `AUTO_INCREMENT` locks on at most 4 tables or partitions).

`lock_cancel_waiting_and_release()`: Instead of removing elements from the middle, simply assign `nullptr`, like `lock_table_remove_autoinc_lock()`.
## Release Notes
This is a minor performance improvement.
## How can this PR be tested?
A simple Sysbench is expected to show some impact. I did not test this yet.
## PR quality check


storage/innobase/lock/lock0lock.cc
```diff
@@ -3646,7 +3646,7 @@ lock_t *lock_table_create(dict_table_t *table, unsigned type_mode, trx_t *trx,
 			ut_ad(!table->autoinc_trx);
 			table->autoinc_trx = trx;
 
-			ib_vector_push(trx->autoinc_locks, &lock);
+			trx->autoinc_locks.emplace_back(std::move(lock));
 			goto allocated;
```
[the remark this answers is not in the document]
dr-m: I initially had `lock` there, but then assertions were failing because we failed to release auto-inc locks. However, `&*lock` does seem to work on those tests that failed for the [very first revision](https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F3720%2Fmerge).

storage/innobase/lock/lock0lock.cc
```diff
@@ -3695,79 +3695,49 @@ lock_t *lock_table_create(dict_table_t *table, unsigned type_mode, trx_t *trx,
...
-				ib_vector_set(trx->autoinc_locks, s, &null_var);
-				return;
-			}
-		}
+  if (*--i == lock)
+  {
+    /* This is the default fast case. */
+    while (begin != i && !*i) i--;
+    trx->autoinc_locks.erase(i, end);
+  }
```
mariadb-DebarunBanerjee, line 3724: IIUC, the logic earlier was to unconditionally pop the element (that equals lock) and then remove the followings if NULL.

```
do {
		ib_vector_pop(trx->autoinc_locks);

		if (ib_vector_is_empty(trx->autoinc_locks)) {
			return;
		}

	} while (*(lock_t**) ib_vector_get_last(trx->autoinc_locks) == NULL);
```
Here, however, it seems ...

`i `points to lock and immediately we check for `!*i` which would always fail and we might end-up erasing only one element and not the other NULL elements that follow.

Can you please confirm ? It is possible I missed something here.

mariadb-DebarunBanerjee (commented):
A am done with the review. Please see the comments.

storage/innobase/lock/lock0lock.cc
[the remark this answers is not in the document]
dr-m: That could work as well. I will try that tomorrow.

storage/innobase/include/small_vector.h
```diff
@@ -49,24 +48,44 @@ class small_vector : public small_vector_base
 
   using small_vector_base::set_size;
 
-  void grow_if_needed()
+  void grow_if_needed() noexcept
   {
-    if (unlikely(size() >= capacity()))
+    if (unlikely(size() >= N))
       grow_by_1(small, sizeof *small);
```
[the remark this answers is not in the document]
dr-m: You may very well be right. I will debug this tomorrow.

storage/innobase/lock/lock0lock.cc
dr-m: You seem to be right. In other words, the `while` line could as well be removed, because the condition `!*i` cannot hold after `*--i == lock` was verified to hold.

This does not cause a correctness issue, because somewhere else we should be ignore `nullptr` entries elsewhere. Still, it would be good to erase the trailing `nullptr` here in order to avoid heap memory allocation.

storage/innobase/include/small_vector.h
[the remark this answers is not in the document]
dr-m: Somehow I had overlooked the `Capacity*= 2` in `small_vector_base::grow_by_1()`. We indeed must retain this data member so that the logic for exponential reallocation sizes can work.

storage/innobase/lock/lock0lock.cc
dr-m: The condition `*--i == lock` always seems to hold in all regression tests that are expected to cover this code, and there never are any preceding `nullptr` entries (because the `else` branch, which would produce them, is never taken). I will try to come up with some improved test coverage.

dr-m: In e07c6806a3116db646222caf301e2880e165444e there is a hint how this code might have been covered in the past. It seems that implementing proper `TEMPORARY TABLE` in InnoDB ([WL#6560](https://dev.mysql.com/worklog/task/?id=6560), [WL#6469](https://dev.mysql.com/worklog/task/?id=6469)) made this test case irrelevant, because no locks would be acquired on temporary tables (because they’re private to the current connection). With the following patch:
```diff
diff --git a/storage/innobase/handler/ha_innodb.cc b/storage/innobase/handler/ha_innodb.cc
index 4fedd79790b..e2ea21b6f17 100644
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -11274,7 +11274,7 @@ create_table_info_t::create_options_are_invalid()
 
 	const char*	ret = NULL;
 	enum row_type	row_format	= m_create_info->row_type;
-	const bool	is_temp 	= m_create_info->tmp_table();
+	const bool	is_temp 	= false&&m_create_info->tmp_table();
 
 	ut_ad(m_thd != NULL);
 
@@ -11706,7 +11706,7 @@ bool create_table_info_t::innobase_table_flags()
 	enum row_type	row_type;
 	rec_format_t	innodb_row_format =
 		get_row_format(m_default_row_format);
-	const bool	is_temp = m_create_info->tmp_table();
+	const bool	is_temp = false && m_create_info->tmp_table();
 	bool		zip_allowed = !is_temp;
 
 	const ulint	zip_ssize_max =
```
the test would fail as follows:
```
CURRENT_TEST: innodb.innodb-autoinc-56228
mysqltest: At line 29: query 'SELECT bug56228()' failed: ER_BAD_TABLE_ERROR (1051): Unknown table 'test.t1_56228'
```
It looks like this is reported because `ha_innobase::delete_table()` returned `HA_ERR_LOCK_WAIT_TIMEOUT`, converted from `DB_LOCK_WAIT_TIMEOUT`. Only before 10.6 1bd681c8b3c5213ce1f7976940a7dc38b48a0d39 the original scenario would have a chance of repeating. If I changed the test to use a non-`TEMPORARY` table, it would fail in a clearer way:
```
mysqltest: At line 12: query 'CREATE FUNCTION bug56228() RETURNS INT DETERMINISTIC
BEGIN
INSERT INTO t1_56228 VALUES(NULL);
INSERT INTO t2_56228 VALUES(NULL);
INSERT INTO t1_56228 VALUES(NULL);
INSERT INTO t2_56228 VALUES(NULL);
DROP TABLE t1_56228;
RETURN 42;
END ' failed: ER_COMMIT_NOT_ALLOWED_IN_SF_OR_TRG (1422): Explicit or implicit commit is not allowed in stored function or trigger
```
Because of all this, I think that this actually is dead code that needs to be simplified further. I’ll try to figure out a proof that the `AUTO_INCREMENT` locks will always be released in a last-in-first-out manner.

dr-m: With quite a few attempts, 48414299f603d948164285e97212c3820da13292 covered an out-of-order lock release scenario. I have been unable to cover any `AUTO_INCREMENT` lock release in `lock_cancel_waiting_and_release()`.

storage/innobase/lock/lock0lock.cc
mariadb-DebarunBanerjee, line 3650: Please remove std::move(). We would like the new version of emplace_back to be called.
`trx->autoinc_locks.emplace_back(lock);`

mariadb-DebarunBanerjee (approved):
All good. Thanks for the changes. 
Please check one related comment.

PR #3723, opened 2024-12-25
10.6 mdev 34877

vlad-lesin:
The pool request contains fixes for MDEV-35708 and 
MDEV-34877. See commit messages for detailed description of the changes.

### DIFF
--- a/mysql-test/suite/innodb/t/avoid_deadlock_with_blocked.test
+++ b/mysql-test/suite/innodb/t/avoid_deadlock_with_blocked.test
@@ -16,14 +16,15 @@ START TRANSACTION WITH CONSISTENT SNAPSHOT;
 --connection default
 CREATE TABLE t1 (id INT PRIMARY KEY) ENGINE=InnoDB STATS_PERSISTENT=0;
 INSERT INTO t1 (id) VALUES (1);
-# Simplest scenario:
-# <con1, S, granted>,
-# <con1, S, granted>, <con2, X, waiting for con1>,
-# Before MDEV-34877:
-# <con1, S, granted>, <con2, X, waiting for con1>, <con1, X, waiting for con1>
-# After MDEV-34877:
-# <con1, S, granted>, <con1, X, granted>, <con2, X, waiting for con1>
-# Expected: instead of deadlocking, the con1's request should ingore con2's
+
+--echo # Simplest scenario:
+--echo # <con1, S, granted>,
+--echo # <con1, S, granted>, <con2, X, waiting for con1>,
+--echo # Before MDEV-34877:
+--echo # <con1, S, granted>, <con2, X, waiting for con1>, <con1, X, waiting for con1>
+--echo # After MDEV-34877:
+--echo # <con1, S, granted>, <con1, X, granted>, <con2, X, waiting for con1>
+--echo # Expected: instead of deadlocking, the con1's request should ingore con2's
 
 --connection con1
   BEGIN;
@@ -43,11 +44,46 @@ INSERT INTO t1 (id) VALUES (1);
   --reap
   COMMIT;
 
-# A variant of the above scenario:
-# <con1, X REC_NOT_GAP, granted>,
-# <con1, X REC_NOT_GAP, granted>, <con2, S, waiting for con1>,
-# <con1, X REC_NOT_GAP, granted>, <con2, S, waiting for con1>, <con1, INSERT INTENTION, waiting for con1>
-# Expected: a deadlock, as INSERT INTENTION should not overtake locks on gap, to not slice them
+--echo # The scenario when we bypass X<-S pair:
+--echo # <con1, S, granted>,
+--echo # <con1, S, granted>, <con2, X, waiting for con1>,
+--echo # <con1, S, granted>, <con2, X, waiting for con1>, <con3, S, waiting for con2>
+--echo # <con1, S, granted>, <con1, X, granted>, <con2, X, waiting for con1>, <con3, S, waiting for con2>
+
+--connection con1
+  BEGIN;
+  SELECT * FROM t1 LOCK IN SHARE MODE;
+
+--connection con2
+  BEGIN;
+  SET DEBUG_SYNC = 'lock_wait_before_suspend SIGNAL con2_will_wait';
+  --send SELECT * FROM t1 FOR UPDATE
+
+--connection con3
+  SET DEBUG_SYNC = 'now WAIT_FOR con2_will_wait';
+  BEGIN;
+  SET DEBUG_SYNC = 'lock_wait_before_suspend SIGNAL con3_will_wait';
+  --send SELECT * FROM t1 LOCK IN SHARE MODE;
+
+--connection con1
+  SET DEBUG_SYNC = 'now WAIT_FOR con3_will_wait';
+  SELECT * FROM t1 FOR UPDATE;
+  COMMIT;
+
+--connection con2
+  --reap
+  COMMIT;
+
+--connection con3
+  --reap
+  COMMIT;
+
+#
+--echo # A variant of the above scenario:
+--echo # <con1, X REC_NOT_GAP, granted>,
+--echo # <con1, X REC_NOT_GAP, granted>, <con2, S, waiting for con1>,
+--echo # <con1, X REC_NOT_GAP, granted>, <con2, S, waiting for con1>, <con1, INSERT INTENTION, waiting for con1>
+--echo # Expected: a deadlock, as INSERT INTENTION should not overtake locks on gap, to not slice them
 --connection con1
   BEGIN;
   SELECT * FROM t1 WHERE id=1 FOR UPDATE;
@@ -67,13 +103,13 @@ INSERT INTO t1 (id) VALUES (1);
   --reap
   COMMIT;
 
-# More complicated scenario:
-# <con1, S, granted>,
-# <con1, S, granted>, <con2, S REC_NOT_GAP, granted>,
-# <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con2>
-# <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con1>, <con1, INSERT_INTENTION, waiting for con3>
-# <con1, S, granted>, <con3, X, waiting for con1>, <con1, INSERT_INTENTION, waiting for con3>
-# Expected: a deadlock, as INSERT INTENTION should not overtake locks on gap, to not slice them
+--echo # More complicated scenario:
+--echo # <con1, S, granted>,
+--echo # <con1, S, granted>, <con2, S REC_NOT_GAP, granted>,
+--echo # <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con2>
+--echo # <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con1>, <con1, INSERT_INTENTION, waiting for con3>
+--echo # <con1, S, granted>, <con3, X, waiting for con1>, <con1, INSERT_INTENTION, waiting for con3>
+--echo # Expected: a deadlock, as INSERT INTENTION should not overtake locks on gap, to not slice them
 
 --connection con1
   BEGIN;
@@ -102,20 +138,19 @@ INSERT INTO t1 (id) VALUES (1);
   --reap
   ROLLBACK;
 
-
 --connection con3
   --error ER_LOCK_DEADLOCK
   --reap
 
-# More complicated scenario.
-# <con1, S, granted>,
-# <con1, S, granted>, <con2, S REC_NOT_GAP, granted>,
-# <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con1>
-# <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con1>, <con1, X REC_NOT_GAP, waiting for con2>
-# Before MDEV-34877:
-# <con1, S, granted>, <con3, X, waiting for con1>, <con1, X REC_NOT_GAP, waiting for con3>
-# After MDEV-34877:
-# <con1, S, granted>, <con1, X REC_NOT_GAP, granted>, <con3, X, waiting for con1>
+--echo # More complicated scenario.
+--echo # <con1, S, granted>,
+--echo # <con1, S, granted>, <con2, S REC_NOT_GAP, granted>,
+--echo # <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con1>
+--echo # <con1, S, granted>, <con2, S REC_NOT_GAP, granted>, <con3, X, waiting for con1>, <con1, X REC_NOT_GAP, waiting for con2>
+--echo # Before MDEV-34877:
+--echo # <con1, S, granted>, <con3, X, waiting for con1>, <con1, X REC_NOT_GAP, waiting for con3>
+--echo # After MDEV-34877:
+--echo # <con1, S, granted>, <con1, X REC_NOT_GAP, granted>, <con3, X, waiting for con1>
 
 
 --connection con1
@@ -149,14 +184,14 @@ INSERT INTO t1 (id) VALUES (1);
   --reap
   COMMIT;
 
-# A secenario, where con1 has to bypass two transactions:
-# <con1, S, granted>
-# <con1, S, granted> <con2, X, waiting>
-# <con1, S, granted> <con2, X, waiting> <con3, X, waiting>
-# Before MDEV-34877:
-# <con1, S, granted> <con2, X, waiting> <con3, X, waiting> <con1, X REC_NOT_GAP, waiting for con2>
-# After MDEV-34877:
-# <con1, S, granted> <con1, X REC_NOT_GAP, granted> <con2, X, waiting> <con3, X, waiting>
+--echo # A secenario, where con1 has to bypass two transactions:
+--echo # <con1, S, granted>
+--echo # <con1, S, granted> <con2, X, waiting>
+--echo # <con1, S, granted> <con2, X, waiting> <con3, X, waiting>
+--echo # Before MDEV-34877:
+--echo # <con1, S, granted> <con2, X, waiting> <con3, X, waiting> <con1, X REC_NOT_GAP, waiting for con2>
+--echo # After MDEV-34877:
+--echo # <con1, S, granted> <con1, X REC_NOT_GAP, granted> <con2, X, waiting> <con3, X, waiting>
 --connection con1
   BEGIN;
   SELECT * FROM t1 LOCK IN SHARE MODE;
--- a/mysql-test/suite/innodb/t/avoid_deadlock_with_blocked.test
+++ b/mysql-test/suite/innodb/t/avoid_deadlock_with_blocked.test
@@ -1,0 +1,194 @@
+--source include/have_innodb.inc
+--source include/have_debug_sync.inc
+--source include/count_sessions.inc
+
+--disable_query_log
+call mtr.add_suppression("InnoDB: Transaction was aborted due to ");
+--enable_query_log
+
+connect stop_purge,localhost,root;
+START TRANSACTION WITH CONSISTENT SNAPSHOT;
+
+--connect (con1,localhost,root,,)
+--connect (con2,localhost,root,,)
+--connect (con3,localhost,root,,)
+
+--connection default
+CREATE TABLE t1 (id INT PRIMARY KEY) ENGINE=InnoDB STATS_PERSISTENT=0;
+INSERT INTO t1 (id) VALUES (1);
+# Simplest scenario:
+# <con1, S, granted>,
+# <con1, S, granted>, <con2, X, waiting for con1>,
+# Before MDEV-34877:
+# <con1, S, granted>, <con2, X, waiting for con1>, <con1, X, waiting for con1>
+# After MDEV-34877:
+# <con1, S, granted>, <con1, X, granted>, <con2, X, waiting for con1>
+# Expected: instead of deadlocking, the con1's request should ingore con2's
+
+--connection con1
+  BEGIN;
+  SELECT * FROM t1 LOCK IN SHARE MODE;
+
+--connection con2
+  BEGIN;
+  SET DEBUG_SYNC = 'lock_wait_before_suspend SIGNAL con2_will_wait';
+  --send SELECT * FROM t1 FOR UPDATE
+
+--connection con1
+  SET DEBUG_SYNC = 'now WAIT_FOR con2_will_wait';
+  SELECT * FROM t1 FOR UPDATE;
+  COMMIT;
+
+--connection con2
+  --reap
+  COMMIT;
+
+# A variant of the above scenario:
+# <con1, X REC_NOT_GAP, granted>,
+# <con1, X REC_NOT_GAP, granted>, <con2, S, waiting for con1>,
+# <con1, X REC_NOT_GAP, granted>, <con2, S, waiting for con1>, <con1, INSERT INTENTION, waiting for con1>
+# Expected: a deadlock, as INSERT INTENTION should not overtake locks on gap, to not slice them
+--connection con1
+  BEGIN;
+  SELECT * FROM t1 WHERE id=1 FOR UPDATE;
+
+--connection con2
+  BEGIN;
+  SET DEBUG_SYNC = 'lock_wait_start SIGNAL con2_will_wait';
+  --send SELECT * FROM t1 LOCK IN SHARE MODE
+
+--connection con1
+  SET DEBUG_SYNC = 'now WAIT_FOR con2_will_wait';
+  INSERT INTO t1 VALUES (0);
+  ROLLBACK;
+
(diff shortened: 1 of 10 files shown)


mysql-test/suite/innodb/t/avoid_deadlock_with_blocked.test
```diff
@@ -0,0 +1,194 @@
...
+  COMMIT;
+
+--connection con1
+  --reap
+  ROLLBACK;
+
+
+--connection con3
+  --error ER_LOCK_DEADLOCK
+  --reap
```
mariadb-DebarunBanerjee, line 108: This is a very nice test ensuring II (INSERT INTENTION) lock doesn't try to bypass other locks and deadlock is expected here. 

One thing about deadlock detection is we never want to guarantee the deadlock victim that would be chosen in case of a deadlock. What do you think ? Is it fine to assume `con3` as deadlock victim here or should we try to handle the case if `con1` is chosen as victim ?

mariadb-DebarunBanerjee (commented):
Hi @vlad-lesin,

This is a very nice improvement patch. I agree with the design. I have one key point about the type of locks we are bypassing and few other minor comments. Please have look.

-Deb

storage/innobase/include/hash0hash.h
```diff
@@ -111,6 +111,38 @@ struct hash_cell_t
...
+  @param next    the next-element pointer in T */
+  template<typename T>
+  void remove(const T &remove, T *T::*next)
+  {
+    T *prev;
+    for (prev= static_cast<T *>(node); prev && prev->*next != &remove;
+         prev= prev->*next);
+    ut_a(prev);
+    prev->*next= remove.*next;
+  }
```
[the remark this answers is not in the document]
vlad-lesin: Yes, you are right, I missed the existing function somehow.

storage/innobase/include/lock0types.h
```diff
@@ -237,6 +237,26 @@ struct ib_lock_t
...
 
+        static inline bool is_rec_granted_X_not_ii_gap(unsigned type_mode)
+        {
+          return (type_mode & (LOCK_INSERT_INTENTION | LOCK_GAP |
+                               LOCK_MODE_MASK)) == LOCK_X;
+        }
+
+        bool is_rec_granted_X_not_ii_gap() const {
+          return is_rec_granted_X_not_ii_gap(type_mode);
+        }
```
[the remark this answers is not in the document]
vlad-lesin: 1. Yes, you are right about "granted" word. The function is called for different cases, when the lock can or can't be waiting.

2. Yes, logically it does not matter if II and Gap locks are X or S, and it's reflected in lock_rec_has_to_wait() function. But as well as we have X and S bits along with II and Gap bits in the same bits set, there would be some kind of ambiguity if we missed "not_ii_not_gap" suffix in the function name.

3. Good point about "ut_ad(!is_table())". I will add it. II-lock is also a gap lock, the correspondent code can be found in lock_rec_insert_check_and_lock(). It's enough just to check gap bit. Thus we can use existing is_rec_granted_exclusive_not_gap() function removing "granted" from the function's name. I would also add debug check for simultaneous II and GAP bits set. I wouldn't  add is_supremum parameter to reuse existing function.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1200,32 +1215,71 @@ static void wsrep_handle_lock_conflict(trx_t *trx)
...
+  /* There is no need to lock lock_sys.wait_mutex to check
+  trx->lock.wait_trx here because the current function is executed under
+  the cell latch, and trx->lock.wait_trx transaction can change wait_trx
+  field only under the cell latch, wait_trx trx_t object can not be
+  deinitialized before releasing all its locks, and during releasing the
+  locks the cell latch will also be requested. So while the cell latch
+  is held, lock->trx->lock.wait_trx can't be changed. */
+  return has_s_lock_or_stronger && is_waiting() &&
+         trx->lock.wait_trx == blocking_trx && is_rec_granted_X_not_ii_gap();
+}
```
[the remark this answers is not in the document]
vlad-lesin: 1. `trx->lock.wait_trx == blocking_trx` means that `trx` is waiting for `blocking_trx`. We must not bypass locks which belong to the transaction, which is not blocked by bypassing transaction. The logic is described in the original MySQL commit message: 
 
> Option 2. avoids a deadlock in this simple case (to solve this in general we'd have to look at the full transitive closure of the waits-for relation) and is justifiable:
>     - trx2 already had to wait for trx1,
>     - in theory, trx1 could have taken that stronger lock earlier which would lead to equivalent result
>     - our code correctness doesn't really depend on honouring waiting lock requests - for example High Priority transactions overtake them, and CATS ignores them during lock granting logic
>     - this will not lead to starvation: trx2 already waited for trx1, so trx2 is not "a new comming transaction", it was in the queue before trx2, and such old transactions should drain eventually
> 

I.e. the idea is the following. If trx2 is waiting for trx1, then it anyway must wait for trx1, and bypassing its lock does not change anything in waiting logic.

But you might be right.

Suppose trx2 does not wait for trx1, but waits for any other transaction trxW in the list. If trxW lock was X-lock, then `has_s_lock_or_stronger` would be false and we would not bypass trx2 lock. So trxW is S-lock. In this case trx1 X-lock would conflict with trxW S-lock, and we would finish locks list iterating before we would reach trx2 X-lock. I depicted this case on [slides 6-7](https://docs.google.com/presentation/d/1eUzNpEL71ttv0Arza5lfOOaVoe4bjlSXhwrwLhM2mLM/edit?usp=sharing).

So if we reached not-gap waiting X-lock, and there were not conflicting locks during the locks iterating, then those not-gap waiting X-lock is waiting for our transaction. I will add the correspondent debug assertion, remove unnecessary condition check and test it with RQG.

2. is_rec_granted_X_not_ii_gap().
> We want to restrict bypassed lock only to X ? 

If bypassed lock is S-lock, then that S-lock conflicts with X-lock. So `has_s_lock_or_stronger` must be X-lock. But in this case new X-lock would not be created, as our transaction already holds X-lock for that heap_no, see lock_rec_has_expl() calls. So strictly speaking, we must check only gap flag there. I will check it with RQG testing.

dr-m (commented):
Sorry, I have been neglecting this too long. I have now read and I think understood the commit message of the larger change, but I didn’t read any of its changes yet.

storage/innobase/lock/lock0lock.cc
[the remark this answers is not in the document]
vlad-lesin: I have checked both (1) and (2) in my last comment.

My claim in (2) is wrong due to the case when we validate lock queue in debug build and don't create new lock but set bit in bitmap of suitable existing lock.

My claim in (1) is wrong due to `S1 X2(waiting for S1) S3(waiting for X2)` case, where newcoming X1 lock seems must bypass both X2 S3 locks. Both (1) and (2) arguments don't make any sense if we are going to implement such bypassing.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4696,8 +4899,10 @@ static bool lock_release_on_prepare_try(trx_t *trx, bool unlock_unmodified)
     {
       ut_ad(!lock->index->table->is_temporary());
       bool supremum_bit= lock_rec_get_nth_bit(lock, PAGE_HEAP_NO_SUPREMUM);
+      /* if XA is being prepared, it must not own waiting locks */
+      ut_ad(!lock->is_waiting());
       bool rec_granted_exclusive_not_gap=
-        lock->is_rec_granted_exclusive_not_gap();
+        lock->is_rec_exclusive_not_gap();
```
dr-m, line 4905: An assertion `!(type_mode & 2048U) || (type_mode & 512U)` is failing here during the execution of the test `rpl.rpl_parallel_xa_same_xid,mix`: https://buildbot.mariadb.org/#/builders/677/builds/12597
The `type_mode` is 2051, or `LOCK_INSERT_INTENTION | LOCK_S`. The assertion expression expects also `LOCK_GAP` to be set if `LOCK_INSERT_INTENTION` is set.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1200,32 +1215,63 @@ static void wsrep_handle_lock_conflict(trx_t *trx)
...
+@return conflicting lock, lock after which new lock should be inserted
+in lock queue in the case when the conflicting lock must be bypassed and
+bypassed lock */
+static conflicting_lock_info
+lock_rec_other_has_conflicting(unsigned mode, const hash_cell_t &cell,
+                               const page_id_t id, ulint heap_no,
+                               const trx_t *trx)
+{
+  bool is_supremum= (heap_no == PAGE_HEAP_NO_SUPREMUM);
+  bool bypass_mode= !is_supremum && lock_t::is_rec_exclusive_not_gap(mode);
```
dr-m, line 1243: These should be `const`.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1200,32 +1215,63 @@ static void wsrep_handle_lock_conflict(trx_t *trx)
...
-                                              ulint heap_no, const trx_t *trx)
-{
-	bool	is_supremum = (heap_no == PAGE_HEAP_NO_SUPREMUM);
+@return conflicting lock, lock after which new lock should be inserted
+in lock queue in the case when the conflicting lock must be bypassed and
+bypassed lock */
+static conflicting_lock_info
+lock_rec_other_has_conflicting(unsigned mode, const hash_cell_t &cell,
+                               const page_id_t id, ulint heap_no,
+                               const trx_t *trx)
```
dr-m, line 1240: This could be `noexcept`.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1294,6 +1340,68 @@ lock_number_of_tables_locked(
...
+  (trx1 S) (trx1 X) (trx2 X) (trx3 X)
+Note the above locks are record or next-key locks.
+If wrong sequence is found, the function will crash with failed assertion.
+@param checked_lock the lock up to which the queue to check
+@param heap_no heap_no of the queue to check */
+static void lock_rec_queue_validate_bypass(const lock_t *checked_lock,
+                                           ulint heap_no)
+{
+  /* "do_lock_reverse_page_reorganize" causes lock queue reversing during page
+  reorganizing, what causes validation failure. Skip the validation for such
```
dr-m, line 1358: s/what/which/. The word is referring to a whole clause or sentence. https://dictionary.cambridge.org/grammar/british-grammar/which

storage/innobase/lock/lock0lock.cc
```diff
@@ -1294,6 +1340,68 @@ lock_number_of_tables_locked(
...
+  case. */
+  DBUG_EXECUTE_IF("do_lock_reverse_page_reorganize", return;);
+  if (!checked_lock || checked_lock->is_waiting())
+    return;
+  page_id_t page_id= checked_lock->un_member.rec_lock.page_id;
+  hash_cell_t *cell= lock_sys.rec_hash.cell_get(page_id.fold());
+  auto mode = checked_lock->type_mode;
+  const trx_t *trx = checked_lock->trx;
```
dr-m, line 1366: Monty wants to omit the space before `=`.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1354,6 +1466,8 @@ lock_rec_create_low(
 	ut_ad(!(type_mode & LOCK_TABLE));
 	ut_ad(trx->state != TRX_STATE_NOT_STARTED);
 	ut_ad(!trx->is_autocommit_non_locking());
+	ut_ad(!c_lock_info.insert_after || !(type_mode & LOCK_WAIT));
+	ut_ad(!c_lock_info.bypassed || c_lock_info.insert_after);
```
dr-m, line 1470: These could be written as
```c++
ut_ad(c_lock_info.insert_after ? !(type_mode & LOCK_WAIT) : c_lock_info.bypassed);
```
Maybe this could be a member function?
```c++
struct conflicting_lock_info {
// ...
  bool some_predicate_name(unsigned type_mode) const noexcept
  { return insert_after ? (type_mode & LOCK_WAIT) : bypassed; }
};
// ...
  ut_ad(c_lock_info.some_predicate_name(type_mode));
```

storage/innobase/lock/lock0lock.cc
```diff
@@ -1525,18 +1647,20 @@ lock_rec_enqueue_waiting(
...
+@param trx       the transaction which lock we are looking for
 @return lock or NULL */
-static inline
-lock_t*
-lock_rec_find_similar_on_page(
-	ulint           type_mode,      /*!< in: lock type_mode field */
-	ulint           heap_no,        /*!< in: heap number of the record */
-	lock_t*         lock,           /*!< in: lock_sys.get_first() */
-	const trx_t*    trx)            /*!< in: transaction */
+static inline lock_t *lock_rec_find_similar_on_page(ulint type_mode,
```
dr-m, line 1659: Elsewhere we are using `unsigned type_mode`. The function could also be declared `noexcept`.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1636,21 +1761,38 @@ static void lock_rec_add_to_queue(unsigned type_mode, const hash_cell_t &cell,
 	if (type_mode & LOCK_WAIT) {
 		goto create;
 	} else if (lock_t *first_lock = lock_sys_t::get_first(cell, id)) {
+		bool bypass_mode= !is_supremum
+			&& lock_t::is_rec_exclusive_not_gap(type_mode);
+		bool has_s_lock_or_stronger= false;
 		for (lock_t* lock = first_lock;;) {
```
dr-m, line 1767: In the old InnoDB formatting style, which this function is following, there always was a space before `=`.

Almost the entire function is being modified here, so we might as well reformat the function in the common style.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1636,21 +1761,38 @@ static void lock_rec_add_to_queue(unsigned type_mode, const hash_cell_t &cell,
...
-			if (lock->is_waiting()
-			    && lock_rec_get_nth_bit(lock, heap_no)) {
-				goto create;
+			if (!lock_rec_get_nth_bit(lock, heap_no))
+				goto cont;
+			if (bypass_mode && lock_rec_can_be_bypassing(trx, lock))
```
dr-m, line 1770: We can avoid the `goto` by defining an empty body and adding `else´ before the subsequent `if`.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1636,21 +1761,38 @@ static void lock_rec_add_to_queue(unsigned type_mode, const hash_cell_t &cell,
...
+			{
+				has_s_lock_or_stronger= true;
 			}
+			/* There can be several locks suited for bypassing,
+			skip them all */
+			else if (lock->is_waiting() &&
+				 (!bypass_mode || !lock->can_be_bypassed(
+				    has_s_lock_or_stronger)))
+					goto create;
```
dr-m, line 1779: `can_be_bypassed()` is checking `is_waiting()` too. Can we simplify this so that `is_waiting()` will be checked only once?

storage/innobase/lock/lock0lock.cc
```diff
@@ -1636,21 +1761,38 @@ static void lock_rec_add_to_queue(unsigned type_mode, const hash_cell_t &cell,
...
+cont:
 			if (!(lock = lock_rec_get_next_on_page(lock))) {
 				break;
 			}
 		}
 
+		const lock_t *bypassed = c_lock_info.insert_after ?
+			lock_rec_get_next(heap_no, c_lock_info.insert_after) :
+			nullptr;
+		ut_ad(bypassed == c_lock_info.bypassed);
```
dr-m, line 1789: Lines are usually split before `?` or `:`, not after them.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1806,21 +1950,26 @@ lock_rec_lock(
...
-        else if (!impl)
+          err= lock_rec_enqueue_waiting(c_lock_info, mode, id,
+                                        block->page.frame, heap_no, index, thr,
+                                        nullptr);
+        /* If some lock was bypassed, we need to create explicit lock to avoid
+        conflicting lock search on every try to convert implicit to explicit
+        lock. */
+        else if (!impl || c_lock_info.insert_after)
         {
           /* Set the requested lock on the record. */
```
dr-m, line 1970: I’d place the comment inside the `else if` body.

storage/innobase/lock/lock0lock.cc
```diff
@@ -4253,24 +4442,38 @@ static void lock_rec_rebuild_waiting_queue(
...
-    if (!lock->is_waiting())
+    /* Store pointer to the next element, because if some lock is
+    bypassed, the pointer to the next lock in the current lock
+    object will be changed, as the current lock will change
+    its position in lock queue. */
+    lock_t *next= lock_rec_get_next(heap_no, lock);
+    if (!lock->is_waiting()) {
+      lock= next;
       continue;
+    }
```
dr-m, line 4455: We could declare `next` in the `for` statement to simplify this a little:
```c++
for (lock_t *lock= first_lock, *next; lock; lock= next)
{
  next= lock_rec_get_next(heap_no, lock);
  if (!lock->is_waiting())
    continue;
  /* If a lock is bypassed, … */
```
This would also allow the `lock= next` statement to be omitted at the end of the body.

storage/innobase/lock/lock0lock.cc
```diff
@@ -2376,10 +2549,14 @@ static void lock_rec_dequeue_from_page(lock_t *in_lock, bool owns_wait_mutex)
...
+	     lock != NULL;) {
+		/* Store pointer to the next element, because if some lock is
+		bypassed, the pointer to the next lock in the current lock
+		object will be changed, as the current lock will change
+		its position in lock queue. */
+		lock_t *next= lock_rec_get_next_on_page(lock);
 		if (!lock->is_waiting()) {
+			lock= next;
 			continue;
 		}
```
dr-m, line 2561: Please apply my suggestion for `lock_rec_rebuild_waiting_queue()`.

In fact, these two functions look rather similar. Would it be feasible to unify them somehow, maybe using some `template` parameter to reduce source code duplication?

storage/innobase/lock/lock0lock.cc
```diff
@@ -1853,46 +2002,70 @@ lock_rec_lock(
...
-/*********************************************************************//**
-Checks if a waiting record lock request still has to wait in a queue.
+/** Checks if a waiting record lock request still has to wait in a queue.
+@param cell      record locks hash table cell for waiting lock
+@param wait_lock waiting lock
 @return lock that is causing the wait */
-static
-const lock_t*
+static conflicting_lock_info
 lock_rec_has_to_wait_in_queue(const hash_cell_t &cell, const lock_t *wait_lock)
```
dr-m, line 2016: The `@return` comment should be adjusted.

This function looks very similar to `lock_rec_other_has_conflicting()`. Would it be feasible to unify these somehow, perhaps by defining some `template` parameter?

dr-m (commented):
Here are some rather low-level observations. I intend to perform a higher-level review, trying to correlate the code changes with the thorough commit message.

mysql-test/suite/innodb/t/avoid_deadlock_with_blocked.test
```diff
@@ -0,0 +1,229 @@
+--source include/have_innodb.inc
+--source include/have_debug_sync.inc
```
dr-m, line 2: It would be good to check for `have_debug.inc` as well, to ensure that the test is skipped at the start of an `mtr` run, rather than in the middle, of a `CMAKE_BUILD_TYPE≠Debug` build.

mysql-test/suite/innodb/t/avoid_deadlock_with_blocked.test
```diff
@@ -0,0 +1,229 @@
...
+--source include/count_sessions.inc
+
+--disable_query_log
+call mtr.add_suppression("InnoDB: Transaction was aborted due to ");
+--enable_query_log
```
dr-m, line 7: After a2d37705ca5a79149b1b064fe3a0be0bbf7eaf73 this should not be strictly necessary. But it will not hurt either.

dr-m (commented):
As far as I can tell, at the core of the patch is the `conflicting_lock_info::insert_after` and how it is used in `lock_rec_create()`, `lock_rec_add_to_queue()`, `lock_rec_lock()` (to upgrade from implciit to explicit locking), `lock_rec_dequeue_from_page()`, and `lock_rec_rebuild_waiting_queue()`. The logic looks clever and sound to me.

When it comes to the assertion failure regaring `LOCK_INSERT_INTENTION` without `LOCK_GAP`, I see that the `LOCK_GAP` flag is being cleared for the `supremum` pseudo-record in `lock_rec_create()` and `lock_rec_add_to_queue()`. Maybe you should add an assertion to those places to check that it is not an insert intention lock.

storage/innobase/include/lock0types.h
```diff
@@ -232,12 +232,36 @@ struct ib_lock_t
...
+        bool is_rec_exclusive_not_gap() const
+        {
+          ut_ad(!is_table());
+          return is_rec_exclusive_not_gap(type_mode);
+        }
+
+        /** Checks if a lock suits for bypassing.
+        @param blocking_trx           transaction for which the lock is checked
+        @return true if lock suits, false otherwise */
+        inline bool can_be_bypassed(bool has_s_lock_or_stronger) const
```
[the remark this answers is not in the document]
vlad-lesin: I would leave the parameter. The only reason of this function is not to duplicate the code. If we return `!has_s_lock_or_stronger` in the callers, then, the first, this check will be duplicated, and the second, there will be only `is_waiting() && !is_gap()` in the function, what can be also treated as not deserving separate function.

storage/innobase/lock/lock0lock.cc
vlad-lesin: The clause is used only once for debug build, and `conflicting_lock_info` contains only two members with public access for release build. I don't see any sense in member function for it.

storage/innobase/lock/lock0lock.cc
vlad-lesin: The comment explains the presence of `c_lock_info.insert_after` in the `else if` block. I think the comment should be before the code it explains.

dr-m: I agree that this is a matter of taste. Because this formatting style does not force us to have the `}` of the preceding `if` block in the same line with the `else if`, the comment is not placed misleadingly.

storage/innobase/lock/lock0lock.cc
vlad-lesin: I would not do such changes in the current task, as they are potentially bug-prone, what can require more time for testing, debugging, delaying the release of the solution for unpredictable amount of time.

storage/innobase/lock/lock0lock.cc
```diff
@@ -984,13 +987,14 @@ lock_rec_get_prev(
   ut_ad(!in_lock->is_table());
   const page_id_t id{in_lock->un_member.rec_lock.page_id};
   hash_cell_t *cell= lock_sys.hash_get(in_lock->type_mode).cell_get(id.fold());
+  lock_t *prev_lock= nullptr;
 
   for (lock_t *lock= lock_sys_t::get_first(*cell, id); lock != in_lock;
        lock= lock_rec_get_next_on_page(lock))
     if (lock_rec_get_nth_bit(lock, heap_no))
-      return lock;
+      prev_lock= lock;
```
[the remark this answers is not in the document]
vlad-lesin: It was decided to create separate pull request for this fix.

storage/innobase/lock/lock0lock.cc
dr-m: I agree. It could make sense to file a follow-up ticket for that. After some review, we might also end concluding that it is not feasible, for some specific reasons that should be written to the ticket.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1200,32 +1215,63 @@ static void wsrep_handle_lock_conflict(trx_t *trx)
...
 #endif /* WITH_WSREP */
 
-/*********************************************************************//**
-Checks if some other transaction has a conflicting explicit lock request
+static inline bool lock_rec_can_be_bypassing(const trx_t *trx,
+                                             const lock_t *lock)
+{
+  ut_ad(!lock->is_insert_intention() || lock->is_gap());
+  return lock->trx == trx && !(lock->type_mode & (LOCK_WAIT | LOCK_GAP)) &&
+         lock_mode_stronger_or_eq(lock->mode(), LOCK_S);
```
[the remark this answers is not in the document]
dr-m: It looks like this comment was not addressed yet?

dr-m (approved):
All in all, this looks OK to me.

You had marked my previous comment about `lock_rec_can_be_bypassing()` as resolved, even though neither the suggested optimization was applied nor rejected with an explanation.

We’re still waiting for clarification which major version this should be pushed to: 10.6 or 11.4.

dr-m:
@montywi requested this to be merged to the 11.4 branch, to reduce any potential risk in earlier GA versions.

vlad-lesin:
This pull request is closed because it was decided to merge the changes into 11.4 branch. See [this](https://github.com/MariaDB/server/pull/3785) pull request for 11.4.

PR #3724, opened 2024-12-30
MDEV-35599: Replace (`u`)`llstr()` with `%llu`/`d`

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-35599](https://jira.mariadb.org/browse/MDEV-35599)*

Nearly all of (`u`)`llstr()` calls were preparations for calls to the `printf` extended family; but both the C/C++ `printf` family and our `my_snprintf()` & descendants already support `long long`s through the `ll` size specifier.
Inlining `llstr()` saves an additional function call each *everywhere*, which
* reduces cognitive load
* potentially improves performance
* mitigates buffer overruns from *if*[^if] the `long long` is too large for the buffer given to `llstr()`

#### Real motivation
I wanted to modernize our ancient code.[^fun]
In addition, I visited lots of files across components during project-wide cleanups like this and #3360, which helps me familiarize with the repository (and its state of maintenance) as a new recruit of the Corp.

#### Non-goals
* Check the correctness of the types with or without this rewrite
  * Particularly, I feel that many `llstr()` calls (replaced with `"%lld"`) should rather use `ullstr()` (which would be `"%llu"`)
    * There’s only a few `ullstr` calls currently.
    * Turning `-Wformat-signedness` on will warn[^Werror] these mistakes.
    * `"%"PRId64`, `"%"PRIu64` and `%zu` are likely more accurate.
  * Many also explicitly casted their values to `longlong`
  * I’ve only corrected cases that emitted `-Wformat`[^Wformat] complaints, such as `sql/sql_test.cc`
    * Some latter files used `llstr` to convert `int`s and `long`s to strings… ¯\\\_(ツ)\_/¯
* Extensive reformat our ancient files
  * You can tell a file’s gone through a history from its mix of tab-space and space-only indents.
    * Speaking of indents – I used a mix of preferences in my diffs… I should chat with others on our preferences vs. practicality.

## Release Notes
TODO: *Ask QA engineers for their impressions on this housekeeping*

## How can this PR be tested?
* Manual: Check the deletion of each `llstr()` usage that it’s either
  * inlined into its outer `printf` by replacing `%s` with `%lld`
  * or, seldomly (a dozen out of a few hundred), entirely replaced with `snprintf(…, "%lld", …)`
* Automatic:
  1. Optional: Create to a temporary branch to merge #3360 with this commit[^Wformat]
  2. Build the server and verify there’re no compile warnings[^Werror]
  3. Run the tests to verify that no output or behavior were affected

* [x] *This is a new feature or __a refactoring__, and the PR is based against the `main` branch.*
* [ ] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check

[^if]: I don’t think we have one right now… :face_in_clouds:
[^fun]: See also: https://justforfunnoreally.dev
[^Werror]: We also have `-Werror` active.
[^Wformat]: `-Wformat` currently only supports the C/C++ `printf` family; #3360 of [MDEV-21978](https://jira.mariadb.org/browse/MDEV-21978) will add support for `my_vsnprintf` and descendants.


storage/maria/ma_recovery_util.c
```diff
@@ -140,8 +139,8 @@ my_bool _ma_redo_not_needed_for_page(uint16 shortid, LSN lsn,
     if ((dirty_page == NULL) ||
         cmp_translog_addr(lsn, dirty_page->rec_lsn) < 0)
     {
-      tprint(tracef, ", ignoring page %s because of dirty_pages list\n",
-             llstr((ulonglong) page, llbuf));
+      tprint(tracef, ", ignoring page %lld because of dirty_pages list\n",
+             (ulonglong) page);
```
vuvova, line 143: do you need a cast here?

storage/myisam/mi_check.c
```diff
@@ -335,21 +332,19 @@ int chk_size(HA_CHECK *param, register MI_INFO *info)
     {
       error=1;
       mi_check_print_error(param,
-			   "Size of indexfile is: %-8s        Should be: %s",
-			   llstr(size,buff), llstr(skr,buff2));
+        "Size of indexfile is: %-8lld        Should be: %lld", size, skr);
```
vuvova, line 335: `%-8lld` -> `%lld` ?

same below and in Aria

storage/myisam/mi_check.c
```diff
@@ -335,21 +332,19 @@ int chk_size(HA_CHECK *param, register MI_INFO *info)
...
   if (!(param->testflag & T_VERY_SILENT) &&
       ! (info->s->options & HA_OPTION_COMPRESS_RECORD) &&
       ulonglong2double(info->state->key_file_length) >
       ulonglong2double(info->s->base.margin_key_file_length)*0.9)
-    mi_check_print_warning(param,"Keyfile is almost full, %10s of %10s used",
-			   llstr(info->state->key_file_length,buff),
-			   llstr(info->s->base.max_key_file_length-1,buff));
+    mi_check_print_warning(param,
+      "Keyfile is almost full, %10lld of %10lld used",
+      info->state->key_file_length, info->s->base.max_key_file_length-1);
```
vuvova, line 347: I'll stop commenting on the width in `%lld`. I think it make sense to specify the width in
* status output printfs, that print many rows and we want numbers to be aligned
* a warning/error that can be reported many times (e.g. some link is broken) and we want all numbers there to be aligned

if it's a warning/error that is reported only once per file "file length is wrong", something in the header is wrong, etc — it doesn't need align to numbers, and there's nothing to align them with

storage/myisam/mi_check.c
```diff
@@ -1225,9 +1205,9 @@ int chk_data_link(HA_CHECK *param, MI_INFO *info, my_bool extend)
                            SEARCH_SAME, info->s->state.key_root[key]);
               if (search_result)
               {
-                mi_check_print_error(param,"Record at: %10s  "
+                mi_check_print_error(param, "Record at: %10lld  "
                                      "Can't find key for index: %2d",
-                                     llstr(start_recpos,llbuff),key+1);
+                                     start_recpos, key+1);
```
vuvova, line 1210: this is one of the errors, I suppose, that can be reported many times for different records. So it kind of makes sense to align the offset here

storage/myisam/myisamchk.c
```diff
@@ -1347,14 +1342,10 @@ static void descript(HA_CHECK *param, register MI_INFO *info, char * name)
 
     printf("%-4d%-6ld%-3d %-8s%-21s",
 	   key+1,(long) keyseg->start+1,keyseg->length,text,buff);
-    if (share->state.key_root[key] != HA_OFFSET_ERROR)
-      llstr(share->state.key_root[key],buff);
-    else
-      buff[0]=0;
     if (param->testflag & T_VERBOSE)
-      printf("%11lu %12s %10d",
+      printf("%11lu %12.0lld %10d", // `.0`: print nothing if zero (HA_OFFSET_ERROR)
```
vuvova, line 1346: same comment as above about `HA_OFFSET_ERROR`

vuvova (changes requested):
Looks great! these *llstr() annoyed me for quite a while. thanks!
```
 35 files changed, 656 insertions(+), 1030 deletions(-)
 ```
And -400 lines, no bad.

See comments inline.

**Review by ParadoxV5 — COMMENTED** [review]
I haven’t checked CI and the feedback’s already in.
Thank you, boss!

File: client/mysqlbinlog.cc [hunk 6]
```diff
@@ -821,8 +820,7 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
   if (is_stmt_end && !result)
   {
     if (print_event_info->print_row_count)
-      fprintf(result_file, "# Number of rows: %s\n",
-              llstr(print_event_info->row_events, ll_buff));
+      fprintf(result_file, "# Number of rows: %lld\n", print_event_info->row_events);
```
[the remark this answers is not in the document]
ParadoxV5: 🤔 Turning on `-Wconversion` (MDEV-26896) will complain about cases like these before this replacement.
🧐 It will also complain about `char *ullstr(longlong value,char *buff)` because it delegates a `ulonglong` arg to `longlong10_to_str(value,buff,10);` which is sign-agnostic.

sql/sql_prepare.cc
```diff
@@ -3483,8 +3483,9 @@ void mysqld_stmt_fetch(THD *thd, char *packet, uint packet_length)
   if (!(stmt= find_prepared_statement(thd, stmt_id)))
   {
     char llbuf[22];
-    my_error(ER_UNKNOWN_STMT_HANDLER, MYF(0), static_cast<int>(sizeof(llbuf)),
-             llstr(stmt_id, llbuf), "mysqld_stmt_fetch");
+    my_error(ER_UNKNOWN_STMT_HANDLER, MYF(0),
+             snprintf(llbuf, sizeof(llbuf), "%lu", stmt_id), stmt_id,
+             "mysqld_stmt_fetch");
```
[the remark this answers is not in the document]
vuvova: ouch. okay. it's not String, it's LEX_STRING, but I got the point

storage/maria/aria_chk.c
```diff
@@ -1717,14 +1704,10 @@ static void descript(HA_CHECK *param, register MARIA_HA *info, char *name)
 
     printf("%-4d%-6ld%-3d %-9s%-23s",
 	   key+1,(long) keyseg->start+1,keyseg->length,text,buff);
-    if (share->state.key_root[key] != HA_OFFSET_ERROR)
-      llstr(share->state.key_root[key],buff);
-    else
-      buff[0]=0;
     if (param->testflag & T_VERBOSE)
-      printf("%9.0f %12s %10d",
+      printf("%9.0f %12.0lld %10d", // `.0`: print nothing if zero (HA_OFFSET_ERROR)
```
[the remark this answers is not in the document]
vuvova: yes, and here I'm saying that while in general you should keep the current behavior, in this particular place you **can** deviate from it, if
* new behavior still shows HA_OFFSET_ERROR different from any valid value
* new behavior is not confusing
* new code is much simpler if you don't keep the current behavior.

or keep the current one, as you prefer

sql/item_func.cc
```diff
@@ -4282,7 +4282,7 @@ longlong Item_func_get_lock::val_int()
     if (args[1]->null_value)
       strmov(buf, "NULL");
     else
-      llstr(((longlong) timeout), buf);
+      snprintf(buf, sizeof(buf), "%lld", (longlong) timeout);
```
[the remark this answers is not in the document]
vuvova: Nope. It's confusing to have so many of them, I agree. But at least some of those were carefully hand-coded for max performance under tons of microbenchmarks. That's why, for example, there are generic number-to-int and int-to-string conversions that work for any base and faster specialized versions for base=10, which is the most commonly needed case.

It's ok to remove a function if you can prove it's obsolete — this is totally possible, the approach that was the fastest 15-20 years ago isn't necessarily the one now. But analyzing/benchmarking all these functions can take a while, so, please, don't do it in this PR/MDEV.

sql/sql_prepare.cc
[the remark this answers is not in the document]
ParadoxV5: By ‘struct String’ I meant ‘struct strings’ (lower case 😁) which includes all except plain `char*` – `String`, `LEX_STRING`, C++ `std::string`…
`std::string` has `c_str()` while `LEX_STRING` is plain struct with no methods. Ouch indeed.

tests/mysql_client_test.c
```diff
@@ -1538,8 +1537,7 @@ static void test_prepare()
       fprintf(stdout, "\n\t tiny   : %d (%lu)", tiny_data, length[0]);
       fprintf(stdout, "\n\t short  : %d (%lu)", small_data, length[3]);
       fprintf(stdout, "\n\t int    : %d (%lu)", int_data, length[2]);
```
ParadoxV5, line 1539: The size specifier for `char` is `hh`.
But since varargs smaller than `int` get auto-promoted to `int`, I won’t judge it here (or press for `h` support in `my_vsnprintf`).

This “`long`” type applied on an `int` data itches me, though –
They probably meant Maria-SQL `INT`, which is better defined with `int32_t` (and format with `"%"PRId32`) to strictly match `INT4`.

storage/maria/ma_check.c
```diff
@@ -5452,28 +5330,26 @@ static int sort_get_next_record(MARIA_SORT_PARAM *sort_param)
       {
 	if (! searching)
 	  _ma_check_print_info(param,
-                               "Found block with wrong recordlength: %lu "
-                               "at %s\n",
-                               block_info.rec_len,
-                               llstr(sort_param->pos,llbuff));
+            "Found block with wrong recordlength: %lu at %lld\n",
```
ParadoxV5, line 5333: Is this `\n` extraneous?
Also applies to the counterpart in `storage/myisam/mi_check.c`

storage/maria/ma_check.c
```diff
@@ -1152,7 +1123,7 @@ static char * record_pos_to_txt(MARIA_HA *info, my_off_t recpos,
                                 char *buff)
 {
   if (info->s->data_file_type != BLOCK_RECORD)
-    llstr(recpos, buff);
+    sprintf(buff, "%lld", recpos);
```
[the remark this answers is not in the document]
ParadoxV5: On the flip side, the code in the `else` branch seems suitable for `sprintf`.

storage/maria/ma_check.c
vuvova: Yes, it is.

https://github.com/MariaDB/server/commit/adefb7cc81404fb70202fd84f614f3d38daa1173#diff-fa3f6010d2b9c0bd15e824385f2f59933a886d13bd44ade44ea679805a65ba03R5431-R5432

storage/maria/aria_chk.c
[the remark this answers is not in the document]
ParadoxV5: I’ve determined that the only “simpler” code is to make use of `.0` by adding one to `my_off_t share->state.key_root[key]`, which is questionable.

sql/log_event_client.cc
```diff
@@ -2013,9 +2009,8 @@ bool Query_log_event::print_query_header(IO_CACHE* file,
       (unlikely(print_event_info->sql_mode != sql_mode ||
                 !print_event_info->sql_mode_inited)))
   {
-    char llbuff[22];
-    if (my_b_printf(file,"SET @@session.sql_mode=%s%s\n",
-                    ullstr(sql_mode, llbuff), print_event_info->delimiter))
+    if (my_b_printf(file, "SET @@session.sql_mode=%llu%s\n",
+                    sql_mode, print_event_info->delimiter))
```
ParadoxV5, line 2013: `my_b_printf` does not support and ignores `%llu`, causing segfaults here and likely erroneous outputs in other places.

`my_b_printf` is *another* one of our `printf` alternative that formats to `my_b_write`.
It only supports format specifiers `%s`, `%c`, `%d`/`%u` and `%ld`/`%lu` (yes, they’re specifiers in its lexicon) and ignores others.

---

Notably, `my_b_printf` also supports `%b` and the `` ` `` flag for `%s` similar to `my_snprintf`; but [that’s switching to `%sB` and `%sQ`](https://jira.mariadb.org/browse/MDEV-21978), whereas `my_b_printf` is a separate implementation independent of that.

When I migrated `my_snprintf` users, I recognized that `my_b_printf` isn’t part of the family and skipped over it.
That meant `my_b_printf` would remain incompatible with `-Wformat` and be “out of date” with the modified `my_snprintf`.

Except `my_b_printf` is causing trouble right now.
@vuvova, what shall we do about it?

vuvova: suggestion — rewrite `my_b_vprintf` copying the logic from `my_vfprintf`

ParadoxV5: ~~and they will be consistent until the next time we update the `my_snprintf` family.~~

🧐 I’m not sure about that. `my_vfprintf` works by resizing its allocation until the result fits[^1], whereas `my_b_write` writes to a limited buffer that only flushes when full.

---

Either case, I should mark this PR as on hold.

[^1]: … which is not efficient; unfortunately, unlike C/C++ `snprintf`, `my_snprintf` doesn’t tell the number of bytes it *would’ve* written if the output buffer’s sufficient.

vuvova: Yes. I mean, `my_b_vprintf` could write to a buffer, resizing it like `my_vfprintf` does. Agree, it's not the best approach, but fixing it would be a separate task. At the end `my_b_vprintf` would use `my_b_write` to write the buffer to `IO_CACHE`

ParadoxV5: I have another idea.

We can modify the implementation of `my_vsnprintf` to instead build around a writer callback, say `bool (*writer)(void *reference, char *string, int string_length)` that returns whether the loop should halt.
When respective buffers fill up, regular `my_vsnprintf`’s callback would return a halt request, while both `my_vfprintf`’s and `my_b_vprintf`’s would flush and carry on.

Of course, because `my_vsnprintf` is divided to many subfunction components, this approach is intrusive and very complex.

vuvova: Sure. That's why it wasn't implemented yet. In particular, it wasn't implemented, when we needed `my_vfprintf`.

A compromise approach — simpler than yours, more complex than what I suggested earlier — would be to make `my_vsnprint_ex` to return, as you wrote, number of bytes it would have written if the buffer were big enough. In this approach `my_vsnprint` can still return what it does now, it's a public API function and if we want to be conservative, we won't change it. But the internal `my_vsnprint_ex` can be changed to return whatever we want, and `my_vfprintf` can call it directly.

ParadoxV5: 👌 Let’s put then on the backburner for now; I have higher-priority assignments.

PR #3725, opened 2024-12-30
MDEV-34119  Assertion `page_dir_get_n_heap(new_page) == 2U' failed in dberr_t PageBulk::init()

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34119*

## Description

Problem:
=======
- insert..select statement on partition table fails to use bulk insert for the transaction.

Solution:
========
- Enable the bulk insert operation for insert..select statement for partition table.

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/row/row0ins.cc
+++ b/storage/innobase/row/row0ins.cc
@@ -2562,6 +2562,22 @@ row_ins_index_entry_big_rec(
 	return(error);
 }
 
+/** Check whether the executed sql command is from insert
+statement
+@param thd thread information
+@return true if it is insert statement */
+static bool thd_sql_is_insert(const THD *thd) noexcept
+{
+  switch(thd_sql_command(thd))
+  {
+    case SQLCOM_INSERT:
+    case SQLCOM_INSERT_SELECT:
+      return true;
+    default:
+      return false;
+  }
+}
+
 #if defined __aarch64__&&defined __GNUC__&&__GNUC__==4&&!defined __clang__
 /* Avoid GCC 4.8.5 internal compiler error due to srw_mutex::wr_unlock().
 We would only need this for row_ins_clust_index_entry_low(),
@@ -2712,7 +2728,7 @@ err_exit:
 	    && block->page.id().page_no() == index->page
 	    && !index->table->is_native_online_ddl()
 	    && (!dict_table_is_partition(index->table)
-	        || thd_sql_command(trx->mysql_thd) == SQLCOM_INSERT)) {
+		|| thd_sql_is_insert(trx->mysql_thd))) {
 
 		if (!index->table->n_rec_locks
 		    && !index->table->versioned()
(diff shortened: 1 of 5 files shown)


storage/innobase/row/row0ins.cc
```diff
@@ -2710,14 +2710,16 @@ row_ins_clust_index_entry_low(
...
 
+		int sql_com= thd_sql_command(trx->mysql_thd);
 		if (!index->table->n_rec_locks
 		    && !index->table->versioned()
 		    && !index->table->is_temporary()
-		    && !index->table->has_spatial_index()) {
+		    && !index->table->has_spatial_index()
+		    && (!dict_table_is_partition(index->table)
+		        || sql_com == SQLCOM_INSERT
+			|| sql_com == SQLCOM_INSERT_SELECT)) {
```
dr-m, line 2722: We must not call any non-`inline` function `thd_sql_command()` unless absolutely necessary. I think that we just need to add a Boolean predicate:
```c++
static bool thd_sql_is_insert(const THD *thd) noexcept
{
  switch (thd_sql_command(thd)) {
  case SQLCOM_INSERT:
  case SQLCOM_INSERT_SELECT:
    return false;
  default:
    return true;
  }
}
```
and replace `thd_sql_command(trx->mysql_thd) == SQLCOM_INSERT` with `thd_sql_is_insert(trx->mysql_thd)`.

dr-m: On a second thought, maybe this needs to cover `ALTER TABLE…ALGORITHM=COPY`, possibly by checking for `SQLCOM_ALTER_TABLE`. Can you add a test for that?

Thirunarayanan: Alter table ..copy algorithm is already handled by the following if condition:
```
if (flags == (BTR_NO_UNDO_LOG_FLAG | BTR_NO_LOCKING_FLAG)
                   && !index->table->n_rec_locks) {

                ut_ad(index->table->skip_alter_undo);
                ut_ad(!entry->is_metadata());
```

dr-m: I see that there is some test coverage in `innodb.alter_copy_bulk`.

storage/innobase/row/row0ins.cc
```diff
@@ -2711,8 +2727,8 @@ row_ins_clust_index_entry_low(
 	    && !trx->dict_operation
 	    && block->page.id().page_no() == index->page
 	    && !index->table->is_native_online_ddl()
-	    && (!dict_table_is_partition(index->table)
-	        || thd_sql_command(trx->mysql_thd) == SQLCOM_INSERT)) {
+            && (!dict_table_is_partition(index->table)
+                || thd_sql_is_insert(trx->mysql_thd))) {
```
dr-m, line 2731: It seems that some TAB was inadvertently replaced with spaces in the indentation.

dr-m (approved):

PR #3726, opened 2024-12-31
Convert comment to compile-time check, fix grammar

ericherman:

There is no JIRA for this simple refactoring.

## Description

By converting the text describing the constraints of the constant to a compile-time check, this enables us to make IO_SIZE configurable.

It should be noted that this #define is duplicated in the submodule libmariadb in the include/ma_global.h file.

## Release Notes

No release notes are needed

## How can this PR be tested?

Alter the source to #define IO_SIZE to be a non-valid value, then re-compile

At this time there is no configure options allowing IO_SIZE to beconfigurable, as such, this does not lend itself to automated testing.

## PR quality check


ericherman:
This #define is duplicated in the submodule `libmariadb` in the `include/ma_global.h` file.

How are changes in `mariadb-server`'s includes propagated to `libmariadb`'s includes?

Is there an existing process?

Should I open a parallel PR for that?

vuvova:
`libmariadb` needs a special PR there's no automatic propagation

ericherman:
> `libmariadb` needs a special PR there's no automatic propagation

I've created a draft PR: https://github.com/mariadb-corporation/mariadb-connector-c/pull/265

I think I will create an MDEV for this work, as it crosses repository boundaries.

include/my_global.h
```diff
@@ -672,12 +672,17 @@ typedef SOCKET_SIZE_TYPE size_socket;
...
-  smaller what the disk page size. This influences the speed of the
-  isam btree library. eg to big to slow.
+  I/O buffer size. May be smaller than the disk page size.
+  This influences the speed of the isam btree library. E.g.: too big too slow.
   4096 is a common block size on SSDs.
 */
+#ifndef IO_SIZE
 #define IO_SIZE			4096U
+#endif
+#if ((IO_SIZE <= 0) || ((IO_SIZE % 512) != 0) || ((IO_SIZE & (IO_SIZE-1)) != 0))
```
dr-m, line 682: This can be shortened to
```c
#if IO_SIZE < 512 || (IO_SIZE & (IO_SIZE - 1))
```

ericherman: Yes. And with the text in the `#error`, I agree that it remains clear-enough. I shall adjust it.

cvicentiu:
@ericherman Is this PR still a draft? How would you like to proceed with this code change given the input you've received thus far?

My thoughts on the code change:
While It would seem that this should be a safe change, I wonder if there aren't any hidden gotchas where a server compiled with a particular IO_SIZE can not actually read the data directory from a server compiled with a different IO_SIZE. I don't have any particular good ideas on answering this other than investigating all code that uses it.

ericherman:
This is still draft because https://github.com/mariadb-corporation/mariadb-connector-c/pull/267 should be merged first.

svoj:
@ericherman now that https://github.com/mariadb-corporation/mariadb-connector-c/pull/267 got merged, should we proceed?

ericherman:
rebased and ready for review

CMakeLists.txt
```diff
@@ -191,6 +191,10 @@ ELSE()
   SET (SKIP_COMPONENTS "N-O-N-E")
 ENDIF()
 
+SET(IO_SIZE "" CACHE STRING "Specify the I/O buffer size")
+IF(IO_SIZE)
+  ADD_DEFINITIONS(-DIO_SIZE=${IO_SIZE})
```
gkodinov: I'd avoid adding a global definition to the command line. I'd add a 

`#cmakedefine IO_SIZE @IO_SIZE@ `

to config.h.cmake. It's arelady included by my_global.h.

CMakeLists.txt
```diff
@@ -191,6 +191,10 @@ ELSE()
   SET (SKIP_COMPONENTS "N-O-N-E")
 ENDIF()
 
+SET(IO_SIZE "" CACHE STRING "Specify the I/O buffer size")
```
gkodinov: I'd put the default here and move this to configure.cmake.

include/my_global.h
```diff
@@ -665,12 +665,17 @@ typedef SOCKET_SIZE_TYPE size_socket;
...
 
 /*
-  Io buffer size; Must be a power of 2 and a multiple of 512. May be
-  smaller what the disk page size. This influences the speed of the
-  isam btree library. eg to big to slow.
+  I/O buffer size. May be smaller than the disk page size.
+  This influences the speed of the isam btree library. E.g.: too big too slow.
   4096 is a common block size on SSDs.
 */
+#ifndef IO_SIZE
```
gkodinov: I'd remove the while @ifndef block and make sure the variable is always defined by cmake.

include/my_global.h
```diff
@@ -665,12 +665,17 @@ typedef SOCKET_SIZE_TYPE size_socket;
...
 #define IO_SIZE			4096U
+#endif
+#if (IO_SIZE < 512) || (IO_SIZE & (IO_SIZE-1))
```
gkodinov: I'd move this check into configure.cmake. Here's it's too late to check and kind of a gotcha.

ericherman:
Thank you for taking a look at this. I am not likely to have a nice chance to look at this in the next week.

CMakeLists.txt
ericherman: I'm not sure I understand why you would avoid this. While perhaps unlikely to be used by people other than those doing performance comparisons, a command-line option makes it easy to build with other values than the default, yes?

ericherman:
The latest push moves the check to CMake
(and rebased)

CMakeLists.txt
```diff
@@ -191,6 +191,15 @@ ELSE()
   SET (SKIP_COMPONENTS "N-O-N-E")
 ENDIF()
 
+# I/O buffer size. May be smaller than the disk page size.
```
gkodinov, line 194: I'd still have it in configure.cmake. But here's is OK too.

libmariadb
gkodinov, line 1: do you really need to change these? It looks like there's a conflict on these in buildbot.

wsrep-lib
gkodinov, line 1: ditto. please focus on the change at hand!

CMakeLists.txt
```diff
@@ -191,6 +191,15 @@ ELSE()
...
+# 4096 is a common block size on SSDs.
+# This influences the speed of the isam btree library.
+# E.g.: too big may become too slow.
+SET(IO_SIZE 4096 CACHE STRING "Specify the I/O buffer size")
+math(EXPR is_power_of_two "(${IO_SIZE} & (${IO_SIZE} - 1))")
```
gkodinov: this is more of a nitpicking, so please feel free to just ignore it: most of the CMake commands in this file are written in UPCASE. Please consider changing math() and message()

gkodinov (changes requested):
Thanks for taking my review notes into account! Please address the below (the changed submodules really, the rest is optional). But, most of all, please work towards getting buildbot green.

CMakeLists.txt
gkodinov: The command line is only that long (on most OSes). So you risk of running out of it if you put too much. In this case this is a rarely (?) option that will have ended up on all command lines for the compiler.

dr-m: I think that also @vaintroub would agree on using `#cmakedefine`.

CMakeLists.txt
```diff
@@ -191,6 +191,15 @@ ELSE()
   SET (SKIP_COMPONENTS "N-O-N-E")
 ENDIF()
 
+# I/O buffer size. May be smaller than the disk page size.
+# 4096 is a common block size on SSDs.
+# This influences the speed of the isam btree library.
+# E.g.: too big may become too slow.
+SET(IO_SIZE 4096 CACHE STRING "Specify the I/O buffer size")
+MATH(EXPR IS_POWER_OF_TWO "(${IO_SIZE} & (${IO_SIZE} - 1))")
```
vaintroub, line 199: this expression should be called IS_NOT_POWER_OF_TWO . !=0 is TRUE, 0 is FALSE, going after usual logic. and the test should be OR IS_NOT_POWER_OF_TWO below

CMakeLists.txt
```diff
@@ -191,6 +191,15 @@ ELSE()
...
+IF(IO_SIZE LESS 512 OR NOT IS_POWER_OF_TWO EQUAL 0)
```
vaintroub, line 200: You also might want to add MATCHES "^[0-9]+$" , to ensure it is a number rather than say "512KB"

CMakeLists.txt
vaintroub: #cmakedefine, yes, otherwise verbose compilation will look even less pretty than it is now.

PR #3729, opened 2025-01-03
MDEV-35000 dict_table_close() breaks innodb_stats_auto_recalc

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35000*
## Description
Every time `dict_table_close()` is decrementing the table reference count to 0, it is performing some expensive logic that aims to ensure that during `FLUSH TABLES`, the InnoDB persistent statistics will be deinitialized, so that during a subsequent `ha_innobase::open()` the statistics will be reloaded. The purpose of this was to ensure that `FLUSH TABLES` will make any changes to `mysql.innodb_table_stats` and `mysql.innodb_index_stats` accessible. This is an overkill, because `FLUSH TABLES` is executed very rarely. It was also breaking the logic of `innodb_stats_auto_recalc=ON` by making frequent resets of `dict_table_t::stat_modified_counter` very likely.

An alternative interface to force the reload of InnoDB persistent statistics will continue to work:
```sql
RENAME TABLE t TO tmp, tmp TO t;
```
This also includes some preparation to simplify the initialization and clearing of the statistics:

`innodb_stats_transient_sample_pages`, `innodb_stats_persistent_sample_pages`: Change the type to `UNSIGNED`, because the number of pages in a table is limited to 32 bits by the InnoDB file formats.

`btr_get_size_and_reserved()`, `fseg_get_n_frag_pages()`, `fseg_n_reserved_pages_low()`, `fseg_n_reserved_pages()`: Return `uint32_t`. The file format limits page numbers to 32 bits.

`dict_table_t::stat`: An `Atomic_relaxed<uint32_t>` that combines a number of metadata fields.

`innodb_copy_stat_flags()`: Copy the statistics flags from `TABLE_SHARE` or `HA_CREATE_INFO`.

`dict_table_t::stats_initialized()`, `dict_table_t::stats_is_persistent()`: Accessors to `dict_table_t::stat`.
## Release Notes
`innodb_stats_transient_sample_pages`, `innodb_stats_persistent_sample_pages`: Change the type to `BIGINT UNSIGNED` to `UNSIGNED`, because the number of pages in a table is limited to 32 bits by the InnoDB file formats.

`FLUSH TABLES` will no longer cause InnoDB persistent statistics to be reloaded from `mysql.innodb_index_stats` and `mysql.innodb_table_stats`. `RENAME TABLE` will continue to trigger a reload of the statistics, before and after this change.
## How can this PR be tested?
This code is rather well covered by the regression test suite.

The `FLUSH TABLES` functionality with respect to InnoDB persistent statistics is covered by the following tests, which were adjusted to use `RENAME TABLE`:
- `innodb.innodb_stats_auto_recalc_on_nonexistent`
- `innodb.innodb_stats_fetch` 

This is a bug fix, but not deemed important enough to be fixed in 10.6 or earlier releases.
## PR quality check


storage/innobase/handler/ha_innodb.cc
```diff
@@ -1116,6 +1116,42 @@ static SHOW_VAR innodb_status_variables[]= {
   {NullS, NullS, SHOW_LONG}
 };
 
+/** Deinitialize InnoDB statistics when closing the last table handle. */
+class InnoDB_share final : public Handler_share
+{
+  /** InnoDB table that contains persistent statistics */
+  dict_table_t &table;
```
dr-m, line 1123: The data member needs to be a pointer and not a reference. On table-rebuilding operations or when a `DROP TABLE` operation is committed, we must update this to point to the new table or invalidate this to `nullptr`. Currently, there are plenty of `heap-use-after-free` errors in table-rebuilding tests.

storage/innobase/dict/dict0stats.cc
```diff
@@ -3005,17 +2977,19 @@ dict_stats_save(
 		"END;", trx);
 
 	if (UNIV_UNLIKELY(ret != DB_SUCCESS)) {
-		ib::error() << "Cannot save table statistics for table "
-			<< table->name << ": " << ret;
+		sql_print_error("InnoDB: Cannot save table statistics for"
+				" table %`.*s.%`s: %s",
+				int(table->name.dblen()), table->name.m_name,
+				table->name.basename(), ut_strerr(ret));
```
dr-m, line 2983: This change turns out to have caused a crash in the following test:
```sh
./mtr --embedded innodb_fts.misc_debug,orig
```
The cause is that the implementation of `vprint_msg_to_log()` in `libmysqld/lib_sql.cc` invokes the standard library function `vsnprintf()`, which cannot deal with the backtick modifier. The result is a crash when something is trying to compute `strlen()` on the length 4 (the length of the first component of `test/t1`). That is, the first parameter of the format is being misinterpreted as a pointer instead of a length.

I figured out that this can be checked by `#ifdef EMBEDDED_SERVER`.

Thirunarayanan (approved):

PR #3732, opened 2025-01-07
MDEV-35785 innodb_log_file_mmap is not defined on 32-bit systems

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35785*
## Description
`HAVE_INNODB_MMAP`: Remove, and unconditionally enable this code.

`log_mmap()`: On 32-bit systems, ensure that the size fits in 32 bits.

`log_t::resize_start()`, `log_t::resize_abort()`: Only handle memory-mapping if `HAVE_PMEM` is defined. The generic memory-mapped interface is only for reading the log in recovery. Writable memory mappings are only for persistent memory, that is, Linux file systems with `mount -o dax`.
## Release Notes
The configuration parameter `innodb_log_file_mmap` will be available in all target platforms.
## How can this PR be tested?
This is covered by the regression test suite. The test `sys_vars.sysvars_innodb` will ensure that the parameter is present.

I tested this by running the regression test suite on IA-32 and AMD64 in my local development environment.
## PR quality check

### DIFF
--- a/mysql-test/suite/sys_vars/t/sysvars_innodb.test
+++ b/mysql-test/suite/sys_vars/t/sysvars_innodb.test
@@ -11,7 +11,6 @@ select VARIABLE_NAME, SESSION_VALUE, DEFAULT_VALUE, VARIABLE_SCOPE, VARIABLE_TYP
     'innodb_numa_interleave',           # only available WITH_NUMA
     'innodb_evict_tables_on_commit_debug', # one may want to override this
     'innodb_use_native_aio',            # default value depends on OS
-    'innodb_log_file_mmap',             # only available on 64-bit
     'innodb_log_file_buffering',        # only available on Linux and Windows
     'innodb_buffer_pool_load_pages_abort')            # debug build only, and is only for testing
   order by variable_name;
--- a/storage/innobase/include/univ.i
+++ b/storage/innobase/include/univ.i
@@ -169,9 +169,6 @@ using the call command. */
 #define UNIV_INLINE static inline
 
 #define UNIV_WORD_SIZE		SIZEOF_SIZE_T
-#if SIZEOF_SIZE_T == 8
-# define HAVE_INNODB_MMAP
-#endif
 
 /** The following alignment is used in memory allocations in memory heap
 management to ensure correct alignment for doubles etc. */
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -1331,9 +1331,7 @@ enum options_xtrabackup
   OPT_INNODB_BUFFER_POOL_FILENAME,
   OPT_INNODB_LOCK_WAIT_TIMEOUT,
   OPT_INNODB_LOG_BUFFER_SIZE,
-#ifdef HAVE_INNODB_MMAP
   OPT_INNODB_LOG_FILE_MMAP,
-#endif
 #if defined __linux__ || defined _WIN32
   OPT_INNODB_LOG_FILE_BUFFERING,
 #endif
@@ -1895,13 +1893,11 @@ struct my_option xb_server_options[] =
    (G_PTR*) &log_sys.buf_size, (G_PTR*) &log_sys.buf_size, 0,
    GET_UINT, REQUIRED_ARG, 2U << 20,
    2U << 20, log_sys.buf_size_max, 0, 4096, 0},
-#ifdef HAVE_INNODB_MMAP
   {"innodb_log_file_mmap", OPT_INNODB_LOG_FILE_SIZE,
    "Whether ib_logfile0 should be memory-mapped",
    (G_PTR*) &log_sys.log_mmap,
    (G_PTR*) &log_sys.log_mmap, 0, GET_BOOL, NO_ARG,
    log_sys.log_mmap_default, 0, 0, 0, 0, 0},
-#endif
 #if defined __linux__ || defined _WIN32
   {"innodb_log_file_buffering", OPT_INNODB_LOG_FILE_BUFFERING,
    "Whether the file system cache for ib_logfile0 is enabled during --backup",
@@ -3370,7 +3366,6 @@ skip:
 	return(FALSE);
 }
 
-#ifdef HAVE_INNODB_MMAP
 static int
 xtrabackup_copy_mmap_snippet(ds_file_t *ds, const byte *start, const byte *end)
 {
@@ -3470,7 +3465,6 @@ static bool xtrabackup_copy_mmap_logfile()
     msg(">> log scanned up to (" LSN_PF ")", recv_sys.lsn);
   return false;
 }
-#endif
 
 /** Copy redo log until the current end of the log is reached
 @return whether the operation failed */
@@ -3482,10 +3476,9 @@ static bool xtrabackup_copy_logfile()
   ut_a(dst_log_file);
   ut_ad(recv_sys.is_initialised());
 
-#ifdef HAVE_INNODB_MMAP
   if (log_sys.is_mmap())
     return xtrabackup_copy_mmap_logfile();
-#endif
+
   const size_t sequence_offset{log_sys.is_encrypted() ? 8U + 5U : 5U};
   const size_t block_size_1{log_sys.write_size - 1};
 
--- a/mysql-test/suite/mariabackup/innodb_redo_log_overwrite.combinations
+++ b/mysql-test/suite/mariabackup/innodb_redo_log_overwrite.combinations
@@ -1,0 +1,4 @@
+[pread]
+--innodb-log-file-mmap=OFF
+[mmap]
+--innodb-log-file-mmap=ON
--- a/mysql-test/suite/sys_vars/r/sysvars_innodb.result
+++ b/mysql-test/suite/sys_vars/r/sysvars_innodb.result
@@ -4,7 +4,6 @@ variable_name not in (
 'innodb_numa_interleave',           # only available WITH_NUMA
 'innodb_evict_tables_on_commit_debug', # one may want to override this
 'innodb_use_native_aio',            # default value depends on OS
-'innodb_log_file_mmap',             # only available on 64-bit
 'innodb_log_file_buffering',        # only available on Linux and Windows
 'innodb_buffer_pool_load_pages_abort')            # debug build only, and is only for testing
 order by variable_name;
@@ -1027,6 +1026,18 @@ NUMERIC_MAX_VALUE	NULL
 NUMERIC_BLOCK_SIZE	NULL
 ENUM_VALUE_LIST	OFF,ON
 READ_ONLY	NO
+COMMAND_LINE_ARGUMENT	OPTIONAL
+VARIABLE_NAME	INNODB_LOG_FILE_MMAP
+SESSION_VALUE	NULL
+DEFAULT_VALUE	ON
+VARIABLE_SCOPE	GLOBAL
+VARIABLE_TYPE	BOOLEAN
+VARIABLE_COMMENT	Whether ib_logfile0 resides in persistent memory (when supported) or should initially be memory-mapped
+NUMERIC_MIN_VALUE	NULL
+NUMERIC_MAX_VALUE	NULL
+NUMERIC_BLOCK_SIZE	NULL
+ENUM_VALUE_LIST	OFF,ON
+READ_ONLY	YES
 COMMAND_LINE_ARGUMENT	OPTIONAL
 VARIABLE_NAME	INNODB_LOG_FILE_SIZE
 SESSION_VALUE	NULL
--- a/storage/innobase/CMakeLists.txt
+++ b/storage/innobase/CMakeLists.txt
@@ -49,7 +49,7 @@ IF(UNIX)
       LINK_LIBRARIES(numa)
     ENDIF()
     IF(CMAKE_SIZEOF_VOID_P EQUAL 8)
-      IF(CMAKE_SYSTEM_PROCESSOR MATCHES "(aarch|AARCH|p(ower)?pc|x86_|amd)64")
+      IF(CMAKE_SYSTEM_PROCESSOR MATCHES "(aarch|AARCH|p(ower)?pc|x86_|amd|loongarch|riscv)64")
         OPTION(WITH_INNODB_PMEM "Support memory-mapped InnoDB redo log" ON)
       ELSE() # Disable by default on ISA that are not covered by our CI
         OPTION(WITH_INNODB_PMEM "Support memory-mapped InnoDB redo log" OFF)
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -19475,18 +19475,14 @@ static MYSQL_SYSVAR_UINT(log_buffer_size, log_sys.buf_size,
   "Redo log buffer size in bytes.",
   NULL, NULL, 16U << 20, 2U << 20, log_sys.buf_size_max, 4096);
 
-#ifdef HAVE_INNODB_MMAP
   static constexpr const char *innodb_log_file_mmap_description=
     "Whether ib_logfile0"
-# ifdef HAVE_PMEM
-    " resides in persistent memory or"
-# endif
+    " resides in persistent memory (when supported) or"
     " should initially be memory-mapped";
 static MYSQL_SYSVAR_BOOL(log_file_mmap, log_sys.log_mmap,
   PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_READONLY,
   innodb_log_file_mmap_description,
   nullptr, nullptr, log_sys.log_mmap_default);
-#endif
 
 #if defined __linux__ || defined _WIN32
 static MYSQL_SYSVAR_BOOL(log_file_buffering, log_sys.log_buffered,
@@ -19982,9 +19978,7 @@ static struct st_mysql_sys_var* innobase_system_variables[]= {
   MYSQL_SYSVAR(deadlock_report),
   MYSQL_SYSVAR(page_size),
   MYSQL_SYSVAR(log_buffer_size),
-#ifdef HAVE_INNODB_MMAP
   MYSQL_SYSVAR(log_file_mmap),
-#endif
 #if defined __linux__ || defined _WIN32
   MYSQL_SYSVAR(log_file_buffering),
 #endif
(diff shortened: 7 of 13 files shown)


dr-m:
4268d64587b10e4b8dce2b506ad4bbe66e92163f needs to be tested on the Debian CI by @ottok, because we have neither RISC-V nor LoongArch on https://buildbot.mariadb.org.

storage/innobase/include/univ.i
```diff
@@ -169,9 +169,6 @@ using the call command. */
 #define UNIV_INLINE static inline
 
 #define UNIV_WORD_SIZE		SIZEOF_SIZE_T
-#if SIZEOF_SIZE_T == 8
-# define HAVE_INNODB_MMAP
-#endif
```
mariadb-DebarunBanerjee, line 174: Now we are removing HAVE_INNODB_MMAP, we should also remove from extra/mariabackup/xtrabackup.cc.

mariadb-DebarunBanerjee (changes requested):
Good that we are getting rid of some conditional paths `HAVE_INNODB_MMAP`. Please check the backup source file too. I wonder if some backup test should have caught it,

storage/innobase/include/univ.i
dr-m: Good catch. Unfortunately, due to the design of `mariadb-backup`, unknown command line options (such as `innodb_log_file_mmap` in this case) will be silently ignored; I checked it by creating two combinations of the test `mariabackup.innodb_redo_log_overwrite`; we had `log_sys.log_mmap=false` in both, and the test passed. At least the test amended in such a way has a chance of covering both log parsers on all targets.

dr-m: 6adffcd96873f57a2b4fd84e51a0829ff8c79f1d extends the test coverage, but it has no way of ensuring that both values of the option are actually covered.

mariadb-DebarunBanerjee (approved):

ottok:
> [4268d64](https://github.com/MariaDB/server/commit/4268d64587b10e4b8dce2b506ad4bbe66e92163f) needs to be tested on the Debian CI by @ottok, because we have neither RISC-V nor LoongArch on https://buildbot.mariadb.org.

I am in the process of testing with riscv64, but builds are very slow..
I don't have access to LoongArch but let's assume they are close enough to riscv64 in this regard.

ottok:
I converted this to a draft so it does not get accidentally merged before it has been finalized and rebased so the fixup commit is squashed into the first one.

ottok:
By the way, have you noticed tha ton s390x the description in the `mariadbd --help --verbose` is:

>    --innodb-log-file-mmap 
>                      Whether ib_logfile0 should initially be memory-mapped

On all other archs it seems to be:

>   --innodb-log-file-mmap 
>                      Whether ib_logfile0 resides in persistent memory or
>                      should initially be memory-mapped

(and on some archs `--innodb-log-file-mmap` is completely missing, which this PR is about to fix)

dr-m:
> By the way, have you noticed tha ton s390x the description in the `mariadbd --help --verbose` is:
> 
> > --innodb-log-file-mmap
> > Whether ib_logfile0 should initially be memory-mapped
> 
> On all other archs it seems to be:
> 
> > --innodb-log-file-mmap
> > Whether ib_logfile0 resides in persistent memory or
> > should initially be memory-mapped
> 
> (and on some archs `--innodb-log-file-mmap` is completely missing, which this PR is about to fix)



> By the way, have you noticed tha ton s390x the description in the `mariadbd --help --verbose` is:
> 
> > --innodb-log-file-mmap
> > Whether ib_logfile0 should initially be memory-mapped
> 
> On all other archs it seems to be:
> 
> > --innodb-log-file-mmap
> > Whether ib_logfile0 resides in persistent memory or
> > should initially be memory-mapped
> 
> (and on some archs `--innodb-log-file-mmap` is completely missing, which this PR is about to fix)

I pointed this out [a week ago](https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/message/KRH7TUT4B7VEBLBGOUVCZ54F2QGQBKY3/), including some reasoning why we can’t support persistent memory on IBM System Z (s390x), or on any 32-bit architecture.

dr-m:
> I converted this to a draft so it does not get accidentally merged before it has been finalized and rebased so the fixup commit is squashed into the first one.

I intend this to be merged as 2 commits, addressing separate issues in the same area.

mysql-test/suite/sys_vars/t/sysvars_innodb.test
```diff
@@ -5,13 +5,13 @@
 
 --vertical_results
 --replace_regex /^\/\S+/PATH/ /\.\//PATH/
+--replace_result 'resides in persistent memory or ' ''
```
ottok, line 8: Alright, I see this unifies the variation in the help text:

```
   --innodb-log-file-mmap 
-                      Whether ib_logfile0 resides in persistent memory or
-                      should initially be memory-mapped
+                      Whether ib_logfile0 should initially be memory-mapped
                       (Defaults to on; use --skip-innodb-log-file-mmap to disable.)
```

mysql-test/suite/sys_vars/t/sysvars_innodb.test
```diff
@@ -5,13 +5,13 @@
...
 select VARIABLE_NAME, SESSION_VALUE, DEFAULT_VALUE, VARIABLE_SCOPE, VARIABLE_TYPE, VARIABLE_COMMENT, NUMERIC_MIN_VALUE, NUMERIC_MAX_VALUE, NUMERIC_BLOCK_SIZE, ENUM_VALUE_LIST, READ_ONLY, COMMAND_LINE_ARGUMENT from information_schema.system_variables
   where variable_name like 'innodb%' and
   variable_name not in (
     'innodb_numa_interleave',           # only available WITH_NUMA
```
ottok, line 12: ..and this seems to be a collection of all variables that appear/disappear based on platform capabilities, even though the source code and dependency versions would be the same.

mysql-test/suite/sys_vars/t/sysvars_innodb.test
ottok: Typically places that have documentation to users show one version per software version. For example https://manpages.debian.org/unstable/mariadb-server-core/mariadbd.8.en.html has 3 versions for the 3 versions of MariaDB in Debian currently. Now it seems that the text will be different depending on architecture, so documentation should maybe have its own version per architecture.. obviously that would not make sense in any online collection of documentation, but for binaries outputting different help texts on different platforms it is less clear that may be the case

mysql-test/suite/sys_vars/t/sysvars_innodb.test
ottok: Typically places that have documentation to users show one version per software version. For example https://manpages.debian.org/unstable/mariadb-server-core/mariadbd.8.en.html has 3 versions for the 3 versions of MariaDB in Debian currently. Now it seems that the text will be different depending on architecture, so documentation should maybe have its own version per architecture.. obviously that would not make sense in any online collection of documentation, but for binaries outputting different help texts on different platforms it is less clear that may be the case

ottok:
My riscv64 test build is still running.. so slow, and also host/queue seems to be busy

ottok:
> > I converted this to a draft so it does not get accidentally merged before it has been finalized and rebased so the fixup commit is squashed into the first one.
> 
> I intend this to be merged as 2 commits, addressing separate issues in the same area.

Thanks for clarification. I misunderstood that "fixup" keyword as it (or "!fixup") is typically used as command words for `git rebase --interactive --autosquash` but seems in this case it is meant for humans to parse together with the MDEV number.

dr-m:
I prefer to push this manually (first rebasing and force-pushing the source branch, then pushing it to the target branch as soon as the tests on the mandatory builders have completed), because it is not clear to me how and when the GitHub auto-merge is being triggered.
I’m eagerly looking forward to the LoongArch and RISC-V test results.
I could also make the parameter documentation string the same on all platforms, replacing the compile-time condition with a comprehension-time condition, something like this:
>Whether ib_logfile0 resides in persistent memory (when supported) or should initially be memory-mapped

Would that be acceptable?

ottok:
Unfortunately the riscv64 build is now failing on unrelated error:
```
[ 39%] Building CXX object storage/perfschema/CMakeFiles/perfschema.dir/table_esgs_by_account_by_event_name.cc.o
cd /<<PKGBUILDDIR>>/builddir/storage/perfschema && /usr/bin/c++ -DHAVE_CONFIG_H -DMYSQL_SERVER -D_FILE_OFFSET_BITS=64 -I/<<PKGBUILDDIR>>/wsrep-lib/include -I/<<PKGBUILDDIR>>/wsrep-lib/wsrep-API/v26 -I/<<PKGBUILDDIR>>/builddir/include -I/<<PKGBUILDDIR>>/include/providers -I/<<PKGBUILDDIR>>/include -I/<<PKGBUILDDIR>>/sql -I/<<PKGBUILDDIR>>/builddir/sql -I/<<PKGBUILDDIR>>/builddir/storage/perfschema -g -O2 -fno-omit-frame-pointer -ffile-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -fno-stack-clash-protection -fdebug-prefix-map=/<<PKGBUILDDIR>>=/usr/src/mariadb-1:11.4.4-3~bpo25.04.1~1736454963.86e3d18b7fc+dev.otto -Wdate-time -D_FORTIFY_SOURCE=3 -Wdate-time -D_FORTIFY_SOURCE=3 -pie -fPIC -fstack-protector --param=ssp-buffer-size=4 -O2 -g -fno-omit-frame-pointer -fno-strict-aliasing -Wno-uninitialized -fno-omit-frame-pointer -DDBUG_OFF -Wall -Wenum-compare -Wenum-conversion -Wextra -Wformat-security -Wmissing-braces -Wno-format-truncation -Wno-init-self -Wno-nonnull-compare -Wno-unused-parameter -Wnon-virtual-dtor -Woverloaded-virtual -Wvla -Wwrite-strings -std=gnu++11   -Wdate-time -D_FORTIFY_SOURCE=3 -DHAVE_OPENSSL -DOPENSSL_API_COMPAT=0x10100000L  -fvisibility=hidden -MD -MT storage/perfschema/CMakeFiles/perfschema.dir/table_esgs_by_account_by_event_name.cc.o -MF CMakeFiles/perfschema.dir/table_esgs_by_account_by_event_name.cc.o.d -o CMakeFiles/perfschema.dir/table_esgs_by_account_by_event_name.cc.o -c /<<PKGBUILDDIR>>/storage/perfschema/table_esgs_by_account_by_event_name.cc
In file included from /usr/include/unicode/uenum.h:25,
                 from /usr/include/unicode/ucnv.h:52,
                 from /usr/include/libxml2/libxml/encoding.h:31,
                 from /usr/include/libxml2/libxml/parser.h:812,
                 from /<<PKGBUILDDIR>>/storage/connect/libdoc.cpp:8:
/usr/include/unicode/localpointer.h:561:26: error: parameter declared ‘auto’
  561 | template <typename Type, auto closeFunction>
      |                          ^~~~
/usr/include/unicode/localpointer.h:573:76: error: template argument 2 is invalid
  573 |     explicit LocalOpenPointer(std::unique_ptr<Type, decltype(closeFunction)> &&p)
      |                                                                            ^
/usr/include/unicode/localpointer.h:583:78: error: template argument 2 is invalid
  583 |     LocalOpenPointer &operator=(std::unique_ptr<Type, decltype(closeFunction)> &&p) {
      |                                                                              ^
/usr/include/unicode/localpointer.h:599:59: error: template argument 2 is invalid
  599 |     operator std::unique_ptr<Type, decltype(closeFunction)> () && {
      |                                                           ^
/usr/include/unicode/uenum.h:69:1: note: invalid template non-type parameter
   69 | U_DEFINE_LOCAL_OPEN_POINTER(LocalUEnumerationPointer, UEnumeration, uenum_close);
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/unicode/ucnv.h:597:1: note: invalid template non-type parameter
  597 | U_DEFINE_LOCAL_OPEN_POINTER(LocalUConverterPointer, UConverter, ucnv_close);
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~
make[4]: *** [storage/connect/CMakeFiles/connect.dir/build.make:796: storage/connect/CMakeFiles/connect.dir/libdoc.cpp.o] Error 1
make[4]: Leaving directory '/<<PKGBUILDDIR>>/builddir'
make[3]: *** [CMakeFiles/Makefile2:6589: storage/connect/CMakeFiles/connect.dir/all] Error 2
```

Trying to debug it..

ottok:
Finally https://launchpadlibrarian.net/769798422/buildlog_ubuntu-oracular-riscv64.mariadb_1%3A11.4.4-3~bpo24.10.1~1736527158.86e3d18b7fc+dev.otto_BUILDING.txt.gz

`innodb-log-file-mmap` is now behaving identical to amd64 and also all main suite tests pass so no regressions.

```
The servers were restarted 230 times
Spent 6990.452 of 1577 seconds executing testcases
Completed: All 1108 tests were successful.
``

Looks good to me!

**Comment by ottok** [comment]
> I could also make the parameter documentation string the same on all platforms, replacing the compile-time condition with a comprehension-time condition, something like this:
> 
> > Whether ib_logfile0 resides in persistent memory (when supported) or should initially be memory-mapped
> 
> Would that be acceptable?

Sound good.The text from --help and first sentence of wiki (https://mariadb.com/kb/en/innodb-system-variables/#innodb_log_file_mmap) should perhaps be identical so that they are aligned and later easy to keep aligned?

**Comment by dr-m** [comment]
@ottok Thank you for the testing. Did you have a chance to run this on LoongArch? This is about covering the following function:
```c++
void pmem_persist(const void *buf, size_t size)
{
# if defined __riscv && __riscv_xlen == 64
  __asm__ __volatile__("fence w,w" ::: "memory");
# elif defined __loongarch64
  __asm__ __volatile__("dbar 0" ::: "memory");
# else
#  error "Missing implementation; recompile with cmake -DWITH_INNODB_PMEM=OFF"
# endif
}
```
According to https://godbolt.org, `fence w,w` compiles into 0x0110000f, which matches
https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/20240411/insts/fence.html
with the following parameters:
* fm=0 (fence mode: normal)
* pred=1 (PW)
* succ=1 (SW)
* rs1=0, rd=0 (ignored)
If I understand the reference correctly, implementing this instruction is mandatory.

https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN#_dbar
https://gcc.gnu.org/onlinedocs/gcc/LoongArch-Base-Built-in-Functions.html mentions that `__builtin_loongarch_dbar()` is available in all ISA versions. I’m using inline assembly, because older compilers might not implement that built-in function. For clang, it was added in [D136906](https://reviews.llvm.org/D136906). The oldest that https://godbolt.org offers is clang-17, which generates the same binary encoding (0x38720000) for both `__builtin_loongarch_dbar(0)` and `dbar 0`. Likewise, all GCC versions available there generate the same binary encoding.

Hence, it seems to me that it is not strictly necessary to wait for the test results for RISC-V. Curiously, https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Built-in-Functions.html does not mention any barrier instructions, only `__builtin_riscv_pause()`, which `MY_RELAX_CPU()` currently fails to put into use. The `my_atomic_cas32_strong_explicit()` fallback ought to be much more expensive. It could be the topic of another patch (which would be nice to get tested):
```diff
diff --git a/include/my_cpu.h b/include/my_cpu.h
index 003087dd94d..0aa3c748d27 100644
--- a/include/my_cpu.h
+++ b/include/my_cpu.h
@@ -104,6 +104,8 @@ static inline void MY_RELAX_CPU(void)
 #elif defined __GNUC__ && (defined __arm__ || defined __aarch64__)
   /* Mainly, prevent the compiler from optimizing away delay loops */
   __asm__ __volatile__ ("":::"memory");
+#elif defined __riscv
+  __builtin_riscv_pause();
 #else
   int32 var, oldval = 0;
   my_atomic_cas32_strong_explicit(&var, &oldval, 1, MY_MEMORY_ORDER_RELAXED,
```
Curiously, `__builtin_riscv_pause()` uses almost the same encoding as `fence w,w` (0x0100000f, only differing in the parameter succ=0). [The patch for implementing `__builtin_riscv_pause()`](https://gcc.gnu.org/pipermail/gcc-patches/2021-January/562936.html) is accompanied with a statement that the `fence` instruction encoding is valid for all ISA versions.

ottok:
> Did you have a chance to run this on LoongArch?

No, I don't have access to a LoongArch machine other than as an official Debian builder. I will upload the contents of https://salsa.debian.org/mariadb-team/mariadb-server/-/commits/debian/latest to Debian now so results for the full build and Debian CI on all platforms should be available in a couple of days.

Commit log of what uploaded to Debian is visible at https://salsa.debian.org/mariadb-team/mariadb-server/-/commits/debian/latest

dr-m:
I filed #3752 for putting `__builtin_riscv_pause()` into use.

ottok:
Follow-up:
- All builds and tests that have run so far have passed in Debian: https://ci.debian.net/packages/m/mariadb/
- The loong64 test runners seems stuck, no new results for a week
- Note the changes I did in Debian are visible at https://salsa.debian.org/mariadb-team/mariadb-server/-/commits/debian/latest, which of these in particular are related to recent upstream MariaDB changes:
  - [Add patch to use the same innodb_log_file_mmap description on all systems](https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/d0f26d8adc45f5f4ad439cf99531666f8a6e99fb)
  - [Add upstream patch to ensure innodb_log_file_mmap is always visible](https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/eb49408ed465b0700ad46f7b4931331ad589248d)
  - [Clean away duplicate wsrep-provider from configuration traces](https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/7a432512d3f2865e00035e307029b6babdc352b2)
  - [Stop depending on obsolete libpmem-dev, re-implemented in MariaDB](https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/3d9b1349eac93fb21cd1c9e45c09e98484a5bfea)
  - [Set CAP_IPC_LOCK capability if possible](https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/497fa1ffa55fb00dbb66a5b3fe4e63e41069dae2)

dr-m:
Thank you, Otto!

PR #3733, opened 2025-01-07
MDEV-35699 Multi-batch recovery occasionally fails

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35699*
## Description
This fixes a regression that had been introduced in #3507. In a multi-batch crash recovery, we would fail to invoke `fil_space_set_recv_size_and_flags()` while parsing the remaining log, before starting the first recovery batch.

`recv_sys_t::parse<storing=NO>()`: Do invoke `fil_space_set_recv_size_and_flags()` and do parse enough of page 0 to facilitate that.
## Release Notes
InnoDB crash recovery could fail to recover the correct size of a file when the `innodb_buffer_pool_size` is too small to process the `ib_logfile0` in a single batch.
## How can this PR be tested?
```sh
./mtr encryption.recovery_memory
```
## PR quality check

### DIFF
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -2787,7 +2787,6 @@ restart:
           erase(r);
           continue;
         }
-      copy_if_needed:
         cl= l.copy_if_needed(iv, decrypt_buf, recs, rlen);
         break;
       case EXTENDED:
@@ -2839,8 +2838,14 @@ restart:
         last_offset= FIL_PAGE_TYPE;
         break;
       case OPTION:
-        if (storing == YES && rlen == 5 && *l == OPT_PAGE_CHECKSUM)
-          goto copy_if_needed;
+        /* OPTION records can be safely ignored in recovery */
+        if (storing == YES &&
+            rlen == 5/* OPT_PAGE_CHECKSUM and CRC-32C; see page_checksum() */)
+        {
+          cl= l.copy_if_needed(iv, decrypt_buf, recs, rlen);
+          if (*cl == OPT_PAGE_CHECKSUM)
+            break;
+        }
         /* fall through */
       case RESERVED:
         continue;
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -2509,7 +2509,7 @@ restart:
   ut_ad(log_sys.is_latest());
 
   alignas(8) byte iv[MY_AES_BLOCK_SIZE];
-  byte *decrypt_buf= storing == YES
+  byte *decrypt_buf= storing != BACKUP
     ? static_cast<byte*>(alloca(srv_page_size)) : nullptr;
 
   const lsn_t start_lsn{lsn};
@@ -2654,7 +2654,7 @@ restart:
         sql_print_warning("InnoDB: Ignoring malformed log record at LSN "
                           LSN_PF, lsn);
         /* the next record must not be same_page */
-        if (storing == YES) last_offset= 1;
+        if (storing != BACKUP) last_offset= 1;
         continue;
       }
       if (srv_operation == SRV_OPERATION_BACKUP)
@@ -2664,7 +2664,7 @@ restart:
                   lsn, b, l - recs + rlen, space_id, page_no));
       goto same_page;
     }
-    if (storing == YES) last_offset= 0;
+    if (storing != BACKUP) last_offset= 0;
     idlen= mlog_decode_varint_length(*l);
     if (UNIV_UNLIKELY(idlen > 5 || idlen >= rlen))
     {
@@ -2695,7 +2695,7 @@ restart:
       goto page_id_corrupted;
     l+= idlen;
     rlen-= idlen;
-    if (storing == YES)
+    if (storing != BACKUP)
     {
       mach_write_to_4(iv + 8, space_id);
       mach_write_to_4(iv + 12, page_no);
@@ -2745,15 +2745,15 @@ restart:
       ut_d(if ((b & 0x70) == INIT_PAGE || (b & 0x70) == OPTION)
              freed.erase(id));
       ut_ad(freed.find(id) == freed.end());
-      const byte *cl= storing == NO ? nullptr : l.ptr;
+      const byte *cl= nullptr; /* avoid bogus -Wmaybe-uninitialized */
       switch (b & 0x70) {
       case FREE_PAGE:
         ut_ad(freed.emplace(id).second);
         /* the next record must not be same_page */
-        if (storing == YES) last_offset= 1;
+        if (storing != BACKUP) last_offset= 1;
         goto free_or_init_page;
       case INIT_PAGE:
-        if (storing == YES) last_offset= FIL_PAGE_TYPE;
+        if (storing != BACKUP) last_offset= FIL_PAGE_TYPE;
       free_or_init_page:
         if (storing == BACKUP)
           continue;
@@ -2808,13 +2808,14 @@ restart:
           trim({space_id, 0}, start_lsn);
           truncated_undo_spaces[space_id - srv_undo_space_id_start]=
             { start_lsn, page_no };
-          if (storing == BACKUP && undo_space_trunc)
+          if (storing != BACKUP)
+            /* the next record must not be same_page */
+            last_offset= 1;
+          else if (undo_space_trunc)
             undo_space_trunc(space_id);
-          /* the next record must not be same_page */
-          if (storing == YES) last_offset= 1;
           continue;
         }
-        if (storing == YES) last_offset= FIL_PAGE_TYPE;
+        if (storing != BACKUP) last_offset= FIL_PAGE_TYPE;
         break;
       case OPTION:
         if (storing == YES && rlen == 5 && *l == OPT_PAGE_CHECKSUM)
@@ -2825,7 +2826,13 @@ restart:
       case WRITE:
       case MEMMOVE:
       case MEMSET:
-        if (storing != YES)
+        if (storing == BACKUP)
+          continue;
+        if (storing == NO && UNIV_LIKELY(page_no != 0))
+          /* fil_space_set_recv_size_and_flags() is mandatory for storing==NO.
+          It is only applicable to page_no == 0. Other than that, we can just
+          ignore the payload and only compute the mini-transaction checksum;
+          there will be a subsequent call with storing==YES. */
           continue;
         if (UNIV_UNLIKELY(rlen == 0 || last_offset == 1))
           goto record_corrupted;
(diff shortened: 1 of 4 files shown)


storage/innobase/log/log0recv.cc
```diff
@@ -2825,8 +2826,13 @@ recv_sys_t::parse_mtr_result recv_sys_t::parse(source &l, bool if_exists)
...
       case MEMMOVE:
       case MEMSET:
-        if (storing != YES)
+        if (storing == BACKUP)
           continue;
+        if (storing == NO)
+        {
+          if (page_no)
+            continue;
+        }
```
mariadb-DebarunBanerjee, line 2835: Not sure if I understand. For `storing = NO`, we would follow through only  for page_no = 0 ?
Since we face this issue/regression, perhaps better to follow pre MDEV-34850 where it was enabled whether storing=yes/no.

mariadb-DebarunBanerjee (commented):
@dr-m,
I see the specific recovery path which we made specific for (`storing == YES`) in MDEV-34850 is needed also for `storing == NO`. Before MDEV-34850 this path was unconditional but I think you are right that is is not required for `storing == BACKUP`. So the condition `"storing != BACKUP"` seems right.

I still find some path which is partly enabled for `storing == NO`and added comments. Unless you think that is addressing any functional issue, it is perhaps safer to follow pre-MDEV-34850 for `storing=NO` path. Please check.

storage/innobase/log/log0recv.cc
```diff
@@ -2808,13 +2808,14 @@ recv_sys_t::parse_mtr_result recv_sys_t::parse(source &l, bool if_exists)
...
             undo_space_trunc(space_id);
-          /* the next record must not be same_page */
-          if (storing == YES) last_offset= 1;
           continue;
         }
-        if (storing == YES) last_offset= FIL_PAGE_TYPE;
+        if (storing != BACKUP) last_offset= FIL_PAGE_TYPE;
         break;
       case OPTION:
         if (storing == YES && rlen == 5 && *l == OPT_PAGE_CHECKSUM)
```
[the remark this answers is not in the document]
dr-m: The `OPT_PAGE_CHECKSUM` record is always last in a mini-transaction, not followed by any records that change any data. In debug instrumented builds, `mtr_t::do_write()` invokes `mtr_t::page_checksum()` to append such records.

For `storing == NO`, we would not do anything with this record here. The body of the `if` statement was not quoted above:
```c++
        if (storing == YES && rlen == 5 && *l == OPT_PAGE_CHECKSUM)
          goto copy_if_needed;
```
This will lead to the following:
```c++
      copy_if_needed:
        cl= l.copy_if_needed(iv, decrypt_buf, recs, rlen);
        break;
```
The `switch (b & 0x70)` body is followed by `if (storing == YES)`. For `storing == NO` the `goto copy_if_needed` can be replaced with `break` or `continue`.

I now realize that the `OPT_PAGE_CHECKSUM` parsing for `innodb_encrypt_log=ON` is broken in the MDEV-14425 format, that is, the merge 57d4a242dac0238dd63ac05cd9d65a38d75a5d5f of 4179f93d28035ea2798cb1c16feeaaef87ab4775 was incorrect. Above, `*l` would be an encrypted octet; we are supposed to read it via `cl`, which we had not assigned yet. I filed [MDEV-35796](https://jira.mariadb.org/browse/MDEV-35796) for this.

storage/innobase/log/log0recv.cc
dr-m: The logic that invokes `fil_space_set_recv_size_and_flags()` applies to `(b & 0x70) == WRITE` and `!page_no && file_checkpoint`. That is, the fields `FSP_SIZE` and `FSP_SPACE_FLAGS`, which the recovery needs to keep track of, are located in the first page of a tablespace. For records that affect any other pages, `storing == NO` only needs to compute the record checksums. The record payload would be parsed, validated and stored later by a subsequent invocation of the template instantiation with `storing == YES`.

For `storing != BACKUP` we have to handle `MEMSET` and `MEMMOVE` records as well as anything that updates `last_offset`, so that the `WRITE` records for page 0 can be processed correctly.

mariadb-DebarunBanerjee (approved):

storage/innobase/log/log0recv.cc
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Ok, thanks for checking.

storage/innobase/log/log0recv.cc
mariadb-DebarunBanerjee: Thanks for checking. If you are confident about it, please go ahead.

mariadb-DebarunBanerjee (approved):

storage/innobase/log/log0recv.cc
```diff
@@ -2867,7 +2879,7 @@ recv_sys_t::parse_mtr_result recv_sys_t::parse(source &l, bool if_exists)
                                    last_offset)
                 : file_name_t::initial_flags;
               if (it == recv_spaces.end())
-                ut_ad(space_id == TRX_SYS_SPACE ||
+                ut_ad(storing == NO || space_id == TRX_SYS_SPACE ||
                       srv_is_undo_tablespace(space_id));
```
dr-m, line 2883: After @mariadb-DebarunBanerjee’s review, I revised this assertion because it would fail during a multi-batch recovery, when `recv_sys.lsn` was earlier than `recv_sys.file_checkpoint` and we encountered a log record for a tablespace for which a `FILE_MODIFY` record would be parsed later. This amendment is essentially reverting the following change in b249a059da2ef03eda35872a763e5453dbc39339 (#3507):
```diff
@@ -2847,7 +2877,7 @@ recv_sys_t::parse_mtr_result recv_sys_t::parse(source &l, bool if_exists)
                                    last_offset)
                 : file_name_t::initial_flags;
               if (it == recv_spaces.end())
-                ut_ad(!store || space_id == TRX_SYS_SPACE ||
+                ut_ad(space_id == TRX_SYS_SPACE ||
                       srv_is_undo_tablespace(space_id));
               else if (!it->second.space)
               {
```

storage/innobase/log/log0recv.cc
```diff
@@ -2808,24 +2807,37 @@ recv_sys_t::parse_mtr_result recv_sys_t::parse(source &l, bool if_exists)
...
             { start_lsn, page_no };
-          if (storing == BACKUP && undo_space_trunc)
+          if (storing != BACKUP)
+            /* the next record must not be same_page */
+            last_offset= 1;
+          else if (undo_space_trunc)
             undo_space_trunc(space_id);
-          /* the next record must not be same_page */
-          if (storing == YES) last_offset= 1;
           continue;
```
dr-m, line 2815: This code is only being emitted when the template parameter is `storing == YES`. That is, all of this could be replaced with the following:
```c++
last_offset= 1;
```
This is because we have the following earlier along this code path:
```c++
      case EXTENDED:
        if (storing != YES)
          continue;
```
In `storing=NO` mode we can safely ignore any `EXTENDED` records, because we are only interested in `WRITE` records affecting certain fields in page 0. The `EXTENDED` record types are either `TRIM_PAGES` or applicable to index pages (which the first page never can be). The `TRIM_PAGES` record is either first or last in a mini-transaction (see `trx_purge_truncate_history()` and `fsp_system_tablespace_truncate()`). Because its argument never is 0 (we never shrink a file to less than 1 page), for any changes of `FSP_SIZE` or `FSP_SPACE_FLAGS` in page 0, the "same page" flag cannot be set in such a way that we’d need a correct `last_offset` to be set by the `EXTENDED` record. I’ll add a source code comment about this.

When it comes to `storing=BACKUP`, it seems to me that `backup_undo_trunc()` is never invoked via `undo_space_trunc` in `mariadb-backup --backup`, that is, backups created during `innodb_undo_log_truncate=ON` could be broken. I will check that.

dr-m: I think that we simply need to change the start of this code path to the following:
```c++
      case EXTENDED:
        if (storing == NO)
          continue;
```
In that way, backup will cover the `undo_space_trunc()`. In later major versions that include f9003c73a15d892423845a2df96c8bd73bbe96ca, this code will be covered by the test `mariabackup.sys_truncate`. I will check this as soon as a build finishes.

I have asked @Thirunarayanan to write a similar test that would cover `innodb_undo_log_truncate=ON`, because we do not currently have one.

dr-m: I realized that in f9003c73a15d892423845a2df96c8bd73bbe96ca invoking the `undo_space_trunc` function pointer is not relevant shrinking the system tablespace (`space_id == 0`). Therefore, the test `mariabackup.sys_truncate` cannot reproduce this bug.

As far as I can tell, the impact of this bug would be that `mariadb-backup --backup` could fail with the following message:
```
mariabackup: Failed to read undo log tablespace space id … and there is no undo tablespace truncation redo record.
```
I will revise this in an attempt to fix that bug.

dr-m: Based on how I’d resolve a conflict with 11.4, I will revise this as follows:
```diff
diff --git a/storage/innobase/log/log0recv.cc b/storage/innobase/log/log0recv.cc
index 1e692f92447..d4bdfbad391 100644
--- a/storage/innobase/log/log0recv.cc
+++ b/storage/innobase/log/log0recv.cc
@@ -2791,10 +2791,28 @@ recv_sys_t::parse_mtr_result recv_sys_t::parse(source &l, bool if_exists)
         cl= l.copy_if_needed(iv, decrypt_buf, recs, rlen);
         break;
       case EXTENDED:
-        if (storing != YES)
+        if (storing == NO)
+          /* We really only care about WRITE records to page 0, to
+          invoke fil_space_set_recv_size_and_flags().  As of now, the
+          EXTENDED records refer to index or undo log pages (which
+          page 0 never can be), or we have the TRIM_PAGES subtype for
+          shrinking a tablespace, to a larger number of pages than 0.
+          Either way, we can ignore this record during the preparation
+          for multi-batch recovery. */
           continue;
         if (UNIV_UNLIKELY(!rlen))
           goto record_corrupted;
+        if (storing == BACKUP)
+        {
+          if (rlen == 1 && undo_space_trunc)
+          {
+            cl= l.copy_if_needed(iv, decrypt_buf, recs, rlen);
+            if (*cl == TRIM_PAGES)
+              undo_space_trunc(space_id);
+          }
+          continue;
+        }
+
         cl= l.copy_if_needed(iv, decrypt_buf, recs, rlen);
         if (rlen == 1 && *cl == TRIM_PAGES)
         {
@@ -2815,7 +2833,10 @@ recv_sys_t::parse_mtr_result recv_sys_t::parse(source &l, bool if_exists)
             undo_space_trunc(space_id);
           continue;
         }
-        if (storing != BACKUP) last_offset= FIL_PAGE_TYPE;
+        /* This record applies to an undo log or index page, and it
+        may be followed by subsequent WRITE or similar records for the
+        same page in the same mini-transaction. */
+        last_offset= FIL_PAGE_TYPE;
         break;
       case OPTION:
         if (storing == YES && rlen == 5 && *l == OPT_PAGE_CHECKSUM)
```

dr-m: This adjustment is ed13d93a256eed0c48ffda7b8fb0f6969e886574.

PR #3734, opened 2025-01-08
MDEV-35647 Possible hang during CREATE TABLE…SELECT error handling

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35647*
## Description
`ha_innobase::delete_table()`: Clear `trx->dict_operation_lock_mode` after, not before `invoking trx->rollback()`, so that `row_undo_mod_parse_undo_rec()` will be invoked with `dict_locked=true` and `dict_sys_t::freeze()` will not be invoked for loading a table definition. Inside `dict_sys_t::freeze()`, an assertion `!have_any()` would fail when the current thread is already holding the latch.

This fixes up c5fd9aa562fb15e8d6ededceccbec0c9792a3243 (MDEV-25919).
## Release Notes
InnoDB could hang if `CREATE TABLE…SELECT` would fail, for example, due to a lock wait timeout.
## How can this PR be tested?
```sh
./mtr innodb.create_select
```
## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -13798,7 +13798,6 @@ int ha_innobase::delete_table(const char *name)
   if (err != DB_SUCCESS)
   {
 err_exit:
-    trx->dict_operation_lock_mode= false;
     trx->rollback();
     switch (err) {
     case DB_CANNOT_DROP_CONSTRAINT:
@@ -13820,7 +13819,7 @@ err_exit:
       dict_table_close(table_stats, true, thd, mdl_table);
     if (index_stats)
       dict_table_close(index_stats, true, thd, mdl_index);
-    dict_sys.unlock();
+    row_mysql_unlock_data_dictionary(trx);
     if (trx != parent_trx)
       trx->free();
     DBUG_RETURN(convert_error_code_to_mysql(err, 0, NULL));
--- a/mysql-test/suite/innodb/r/create_select.result
+++ b/mysql-test/suite/innodb/r/create_select.result
@@ -3,7 +3,6 @@ connection default;
 CREATE TABLE t1 ENGINE=InnoDB SELECT * FROM seq_1_to_100000000;
 connection con1;
 KILL QUERY @id;
-disconnect con1;
 connection default;
 ERROR 70100: Query execution was interrupted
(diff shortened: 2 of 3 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -13798,8 +13798,8 @@ int ha_innobase::delete_table(const char *name)
   if (err != DB_SUCCESS)
   {
 err_exit:
-    trx->dict_operation_lock_mode= false;
     trx->rollback();
+    trx->dict_operation_lock_mode= false;
```
mariadb-DebarunBanerjee, line 13802: It seems like the way we set `trx->dict_operation_lock_mode` is not very consistent. One nice way could have been to set while acquiring the lock and  unset while releasing. 

The way we set it above `trx->dict_operation_lock_mode= true;` looks ok as we are doing so immediately after acquiring the lock and purge thread wait. Calling `row_mysql_lock_data_dictionary` would have been better I guess.

For the success path the unset happens by `row_mysql_unlock_data_dictionary(trx)`. 
```
/** Unlock the data dictionary. */
#define row_mysql_unlock_data_dictionary(trx)                   \
        do {                                                    \
                ut_ad(!lock_trx_has_sys_table_locks(trx));      \
                ut_ad(trx->dict_operation_lock_mode);           \
                trx->dict_operation_lock_mode = false;          \
                dict_sys.unlock();                              \
        } while (0)
```
The unset in the error path looks more problematic as we are not associating with `dict_sys.unlock()` which happens after sometime after closing the table. This is the root cause of the issue I think.

While I think the current patch fixes the issue, may I suggest ...

1. Move `trx->dict_operation_lock_mode= false;` down while unlocking the dictionary.
2. Improve [1] by calling row_mysql_unlock_data_dictionary instead of `dict_sys.unlock();` to unset dict_operation_lock_mode together.

mariadb-DebarunBanerjee (commented):
The patch looks functionally correct. I have one improvement suggestion.

storage/innobase/handler/ha_innodb.cc
dr-m: Thank you, that is a good suggestion. Please review 6d4841ae26a8c62b2645b9a0e9c0a9e44e460354.

mariadb-DebarunBanerjee (approved):
Thanks for the change. Looks good to me.

PR #3735, opened 2025-01-08
MDEV-35710 support for threadpool

sjaakola:
- [x] *The Jira issue number for this PR is: MDEV-35710*

## Description
When client connections use threadpool, i.e. configuration has: thread_handling = pool-of-threads
it turned out that during wsrep replication shutdown, not all client connections could be closed. Reason was that some client threads has stmt_da in state DA_EOF, and this state was earlier used to detect if client connection was issuing SHUTDOWN command.

To fix this, the connection executing SHUTDOWN is now detected by looking at the actual command being executed:
thd->get_command() == COM_SHUTDOWN
During replication shutdown, all other connections but the SHUTDOWN executor, are terminated.

## Release Notes

## How can this PR be tested?
This commit has new mtr test galera.galera_threadpool, which opens a number of threadpool client connections, and then restarts the node to verify that connections in threadpool are terminated during shutdown.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera/r/galera_threadpool.result
+++ b/mysql-test/suite/galera/r/galera_threadpool.result
@@ -1,0 +1,36 @@
+connection node_2;
+connection node_1;
+connection node_1;
+connection node_2;
+connect node_2a, 127.0.0.1, root, , test, $NODE_MYPORT_2;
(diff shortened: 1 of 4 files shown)


sysprg:
Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/552cba92de9952a9a07360a6791d2f63bc06470c

PR #3742, opened 2025-01-10
MDEV-35802 Race condition in log_t::persist()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35802*
## Description
`log_t::persist()`: Remove the parameter `holding_latch`, and assert `latch_holding_any()`. We used to avoid acquiring a latch when log resizing was not in progress. That allowed a race condition to occur where `log_t::write_checkpoint()` has just completed log resizing. In that case, we could wrongly invoke `pmem_persist()` on the old `log_sys.buf` instead of the new one, which was shortly before known as `log_sys.resize_buf`.

`log_write_persist()`: A non-inline wrapper function that will invoke `log_sys.persist()` while holding a shared `log_sys.latch`.
## Release Notes
This is a rare race condition, probably not worth mentioning, because the use of PMEM is very rare.
## How can this PR be tested?
This code is being covered by the following test:
```sh
```
## PR quality check

### DIFF
--- a/storage/innobase/include/log0log.h
+++ b/storage/innobase/include/log0log.h
@@ -462,9 +462,8 @@ public:
 
 #ifdef HAVE_PMEM
   /** Persist the log.
-  @param lsn            desired new value of flushed_to_disk_lsn
-  @param holding_latch  whether the caller is holding exclusive latch */
-  void persist(lsn_t lsn, bool holding_latch) noexcept;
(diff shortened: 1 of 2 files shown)


mariadb-DebarunBanerjee (approved):
The change uses stronger latching and should not have any adverse impact. The use case explains that it is actually required. Hopefully, we don't incur any performance penalty.

PR #3744, opened 2025-01-10
Fix -DBUILD_CONFIG=mysql_release to keep standard cmake flags

cvicentiu:
-DCMAKE_BUILD_TYPE=xxx sets some C compiler flags according to the build type. -DBUILD_CONFIG was completely overwriting them in some compiler / arch combinations and not in others. Make it consistently "append-only", not overwrite.

Also, enforce the same set of flags for Release and RelWithDebInfo.

This reverts ff1f611a0d84b84a06b100a3b6901fc6d8957ff5 as it is no longer necessary.

    Avoid assert()

    By default, CMAKE_BUILD_TYPE RelWithDebInfo or Release implies
    -DNDEBUG, which disables the assert() macro. MariaDB is deviating
    from that. Let us be explicit to use assert() only in debug builds.


cmake/build_configurations/mysql_release.cmake
```diff
@@ -270,27 +278,33 @@ IF(UNIX)
       IF(CMAKE_SYSTEM_PROCESSOR MATCHES "i386")
         SET(COMMON_C_FLAGS                   "-g -mt -fsimple=1 -ftrap=%none -nofstore -xbuiltin=%all -xlibmil -xlibmopt -xtarget=generic")
         SET(COMMON_CXX_FLAGS                 "-g0 -mt -fsimple=1 -ftrap=%none -nofstore -xbuiltin=%all -features=no%except -xlibmil -xlibmopt -xtarget=generic")
-        SET(CMAKE_C_FLAGS_DEBUG              "-xO1 ${COMMON_C_FLAGS}")
-        SET(CMAKE_CXX_FLAGS_DEBUG            "-xO1 ${COMMON_CXX_FLAGS}")
```
dr-m, line 274: Solaris on i386 is dead, right? Even Oracle switched from Sun Forte C or SunPro to GCC at least for a while during the early phase of the MySQL 8.0 development cycle, because the C++11 support was not adequate.

cmake/build_configurations/mysql_release.cmake
```diff
@@ -187,30 +187,32 @@ IF(UNIX)
...
+    SET(CMAKE_C_FLAGS_DEBUG          "${CMAKE_C_FLAGS_DEBUG} ${COMMON_C_FLAGS}")
+    SET(CMAKE_C_FLAGS_RELEASE        "${CMAKE_C_FLAGS_RELEASE} -O3 ${COMMON_C_FLAGS}")
+    SET(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -O3 ${COMMON_C_FLAGS}")
   ENDIF()
   IF(CMAKE_COMPILER_IS_GNUCXX)
     SET(COMMON_CXX_FLAGS               "-g -fno-omit-frame-pointer -fno-strict-aliasing -Wno-uninitialized")
-    SET(CMAKE_CXX_FLAGS_DEBUG          "-O ${COMMON_CXX_FLAGS}")
-    SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 ${COMMON_CXX_FLAGS}")
+    SET(CMAKE_CXX_FLAGS_DEBUG          "${CMAKE_CXX_FLAGS_DEBUG} ${COMMON_CXX_FLAGS}")
+    SET(CMAKE_CXX_FLAGS_RELEASE        "${CMAKE_CXX_FLAGS_RELEASE} -O3 ${COMMON_C_FLAGS}")
```
vaintroub, line 197: same about string(REGEX REPLACE) to  replace -O2 with  -O3, to only have single  -O

cmake/build_configurations/mysql_release.cmake
```diff
@@ -187,30 +187,32 @@ IF(UNIX)
   # Default GCC flags
   IF(CMAKE_COMPILER_IS_GNUCC)
     SET(COMMON_C_FLAGS               "-g -fno-omit-frame-pointer -fno-strict-aliasing  -Wno-uninitialized")
-    SET(CMAKE_C_FLAGS_DEBUG          "-O ${COMMON_C_FLAGS}")
-    SET(CMAKE_C_FLAGS_RELWITHDEBINFO "-O3 ${COMMON_C_FLAGS}")
+    SET(CMAKE_C_FLAGS_DEBUG          "${CMAKE_C_FLAGS_DEBUG} ${COMMON_C_FLAGS}")
+    SET(CMAKE_C_FLAGS_RELEASE        "${CMAKE_C_FLAGS_RELEASE} -O3 ${COMMON_C_FLAGS}")
+    SET(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -O3 ${COMMON_C_FLAGS}")
```
vaintroub, line 192: Better string(REGEX REPLACE) to replace -O2 with -O3 .Or, there will be both -O2 and -O3 in the compilation. We know, that last option happens to win in GCC, but only one -O is cleaner.

vaintroub (changes requested):

cmake/build_configurations/mysql_release.cmake
mariadb-AlessandroVetere, line 197: Maybe use `COMMON_CXX_FLAGS` instead of `COMMON_C_FLAGS` ?

Also, usually `RelWithDebInfo` might use `-O2` instead of `-O3` which we are forcing here, but this is unrelated to this PR (and noted in comment above).

cvicentiu: @mariadb-AlessandroVetere Thank you, I did catch that in my follow-up push and fixed it.

PR #3746, opened 2025-01-10
MDEV-35818: Fix `replace_binlog_file` info message

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-35818 my_snprintf fix for 11.7+](https://jira.mariadb.org/browse/MDEV-35818)*

## Description
`ATTRIBUITE_FORMAT` from #3360 uncovers issues on `my_snprintf` uses.
This commit fixes the one in `Binlog_commit_by_rotate::` `replace_binlog_file()` about “required size too big”.

All I found is that it’s not present in 11.4 (after I prepared previous batches for all maintained branches), for GitHub blame can’t process a file with over 10K lines.

## Release Notes
*But 11.7 wasn’t Generally Available yet?*

## How can this PR be tested?
I suppose we can trigger this issue by running the `simulate_required_size_too_big` test (wherever it is) in __32-bit Linux__.

## PR quality check

### DIFF
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -13257,10 +13257,11 @@ bool Binlog_commit_by_rotate::replace_binlog_file()
   DBUG_EXECUTE_IF("simulate_required_size_too_big", required_size= 10000;);
   if (required_size > m_cache_data->file_reserved_bytes())
   {
-    sql_print_information("Could not rename binlog cache to binlog(as "
+    sql_print_information("Could not rename binlog cache to binlog (as "
                           "requested by --binlog-commit-by-rotate-threshold). "
-                          "Required %llu bytes but only %llu bytes reserved.",
-                          required_size, m_cache_data->file_reserved_bytes());
+                          "Required %zu bytes but only %lu bytes reserved.",
+                          required_size, static_cast<unsigned long>(
+                            m_cache_data->file_reserved_bytes()));
     return false;
   }
 


sql/log.cc
```diff
@@ -13257,10 +13257,11 @@ bool Binlog_commit_by_rotate::replace_binlog_file()
...
   if (required_size > m_cache_data->file_reserved_bytes())
   {
-    sql_print_information("Could not rename binlog cache to binlog(as "
+    sql_print_information("Could not rename binlog cache to binlog (as "
                           "requested by --binlog-commit-by-rotate-threshold). "
-                          "Required %llu bytes but only %llu bytes reserved.",
-                          required_size, m_cache_data->file_reserved_bytes());
+                          "Required %zu bytes but only %lu bytes reserved.",
+                          required_size,
+                          (unsigned long) m_cache_data->file_reserved_bytes());
```
ParadoxV5, line 13264: `binlog_cache_data::file_reserved_bytes()` is an `uint32`, which doesn’t necessarily match `PRIu32`’s `uint32_t`, so I’m up-casting for now.
[MDEV-35460](https://jira.mariadb.org/browse/MDEV-35460)

bnestere: You could use c++ style casting here instead of `(unsigned long)`.

bnestere (approved):
Looks good, thanks @ParadoxV5! I left one note for your consideration.

sql/log.cc
ParadoxV5: Oh right, I keep forgetting. Thanks

PR #3753, opened 2025-01-13
MDEV-35658 Assertion `commit_trx' failed in test galera_as_master

sciascid:
- [x] *The Jira issue number for this PR is: MDEV-35658*

## Description
Fix assertion `commit_trx` in innobase_commit() during MTR test galera.galera_as_master. 
The test issues a simple INSERT statement, while sql_log_bin = 0. This option disables writes to binlog.  However, since MDEV-7205, the option does not affect Galera, so changes are still replicated. So sql_log_bin=off, "partially" disabled the binlog and the INSERT will involve both binlog and innodb, thus requiring internal 2 phase commit (2PC). In 2PC INSERT is first prepared, which will make it transition to PREPARED state in innodb, and later committed which causes the new assertion from MDEV-24035 to fail.
Running the same test with sql_log_bin enabled also results in 2PC, but the execution has one more step for ordered commit, between prepare and commit. Ordered commit causes the transaction state to transition back to TRX_STATE_NOT_STARTED. Thus avoiding the assertion. This patch makes sure that when sql_log_bin=off, the ordered commit step is not skipped, thus going through the expected state transitions in the storage engine.

## Release Notes
Nothing.

## How can this PR be tested?
Should fix MTR test galera.galera_as_master.

## PR quality check

### DIFF
--- a/sql/log.cc
+++ b/sql/log.cc
@@ -7761,7 +7761,12 @@ MYSQL_BIN_LOG::write_transaction_to_binlog(THD *thd,
   {
     DBUG_RETURN(0);
   }
-  else if (!(thd->variables.option_bits & OPTION_BIN_LOG))
+
+  if (!(thd->variables.option_bits & OPTION_BIN_LOG)
+#ifdef WITH_WSREP
+      && !WSREP(thd)
+#endif
+      )
   {
     cache_mngr->need_unlog= false;
     DBUG_RETURN(0);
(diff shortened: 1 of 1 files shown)


dr-m:
I don’t know the replication code at all. Perhaps @andrelkin or @knielsen would  be better reviewers?

sql/log.cc
```diff
@@ -7761,7 +7761,7 @@ MYSQL_BIN_LOG::write_transaction_to_binlog(THD *thd,
   {
     DBUG_RETURN(0);
   }
-  else if (!(thd->variables.option_bits & OPTION_BIN_LOG))
+  else if (!WSREP(thd) && !(thd->variables.option_bits & OPTION_BIN_LOG))
```
andrelkin, line 7764: I'd prefer to isolate `!WSREP(thd) &&`  with 
```
#ifdef WITH_WSREP
 !WSREP(thd) &&
#endif
```
which would be also consistent with l.8671.

andrelkin (changes requested):
Other than the l.7764 note the patch looks good.

sql/log.cc
sciascid: Should be fixed now.

andrelkin (approved):
Thanks!

PR #3756, opened 2025-01-13
MDEV-35096: History is stored in different partitions on different nodes when using SYSTEM VERSION

bnestere:
Row-injection updates don’t correctly set the historical partition
for tables with system versioning and system_time partitions. This
results in inconsistencies between the master and slave when
replicating transactions that target such tables (i.e. the primary
server would correctly distribute archived rows amongst its
partitions, whereas the replica would have all archived rows in a
single partition). The function
partition_info::vers_set_hist_part(THD*) is used to set the
partition; however, its initial check for
vers_require_hist_part(THD*) returns false, bypassing the rest of
the function (which sets up the partition to use). This is because
the actual check uses the LEX sql_command (via
LEX::vers_history_generating()) to determine if the command is valid
to generate history. Row injections don’t have sql_commands though.

This patch provides a fix which extends the check in
vers_history_generating() to additionally allow row injections to be
history generating (via the function LEX::is_stmt_row_injection()).

Special thanks to Jan Lindstrom <jan.lindstrom@galeracluster.com>
for his work in reproducing the bug, and providing an initial test
case.

 1. A .test showing the regression
 2. The patch with a fix and .result file


knielsen (approved):
Thanks, Brandon!

I think the patch looks good, the added test seems clean and in a reasonable place (and it's simple).

One question to the test case that I did not understand. Why does the test case use BINLOG statements on the master instead of normal DELETE queries? Those delete queries would then presumably generate the same events to be sent to the slave?

If you wanted to test how such row events get binlogged on a master, this could still be done with normal DELETE queries, by using an intermediate slave M->S1->S2.
Or you could test both with real queries and with the crafted BINLOG statements, if you wanted to test those specific binary events, for some reason.

It's not that I think the test is wrong in the way it's written, I just did not understand if there was a specific reason for doing so, and if so what that reason is.

bnestere:
Thanks for a quick review, @knielsen !

> Why does the test case use BINLOG statements on the master instead of normal DELETE queries?

To test both routes for executing row-injection based events: through 1) the replication applier, and 2) directly executing BINLOG SQL commands. I suppose I could have done something like you suggest, and on the second replica, just to replay `mysqlbinlog` output via `mysql`. Though my preference is the current way, as it makes it clear what is being tested (though perhaps I could extend the comments a bit to answer the "why" better).

knielsen:

>> Why does the test case use BINLOG statements on the master instead of normal DELETE queries?

> To test both routes for executing row-injection based events: through
> 1) the replication applier, and 2) directly executing BINLOG SQL

Aha, that makes sense, thanks for the clarification.

 - Kristian.

PR #3757, opened 2025-01-13
MDEV-35694: Mysqlbinlog --stop-position should warn if EOF not reached with --read-from-remote-server

bnestere:
MDEV-27037 added functionality to warn users that a specified
stop-position or stop-datetime was never reached. It only worked for
local files though. The patch in MDEV-35528 changed the
implementation for stop-datetime to work to provide the warning also
when using --read-from-remote-server. The PR for that MDEV (#3670)
was limited to only the stop-datetime field.

This patch updates the --stop-position warning to also work with
--read-from-remote-server.

1. The first commit shows the regression (as a missing warning in the
result when running binlog.binlog_mysqlbinlog_warn_stop_position
2. The second commit is the fix.

### DIFF
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -143,6 +143,7 @@ static uint verbose= 0;
 
 static ulonglong start_position, stop_position;
 static const longlong stop_position_default= (longlong)(~(my_off_t)0);
+static ulonglong last_processed_position= 0;
 #define start_position_mot ((my_off_t)start_position)
 #define stop_position_mot  ((my_off_t)stop_position)
 
@@ -997,6 +998,8 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
   determining how to print.
   @param[in] ev Log_event to process.
   @param[in] pos Offset from beginning of binlog file.
+  @param[in] ev_end_pos Offset that represents the end of the event in the
+                        binlog file.
   @param[in] logname Name of input binlog.
 
   @retval ERROR_STOP An error occurred - the program should terminate.
@@ -1005,7 +1008,8 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
   events to process has been reached and the program should terminate.
 */
 Exit_status process_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
-                          my_off_t pos, const char *logname)
+                          my_off_t pos, my_off_t ev_end_pos,
+                          const char *logname)
 {
   char ll_buff[21];
   Log_event_type ev_type= ev->get_type_code();
@@ -1462,6 +1466,7 @@ err:
 end:
   rec_count++;
   last_processed_datetime= ev_when;
+  last_processed_position= ev_end_pos;
 
   DBUG_PRINT("info", ("end event processing"));
   /*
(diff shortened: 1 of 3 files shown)


client/mysqlbinlog.cc
```diff
@@ -143,6 +143,7 @@ static uint verbose= 0;
 
 static ulonglong start_position, stop_position;
 static const longlong stop_position_default= (longlong)(~(my_off_t)0);
+static ulonglong last_processed_position= 0;
```
ParadoxV5, line 146: The comparison down there looks so wrong having to cast (C-style or `static_const`).
Will an unsigned do? Or even a this-file-only macro-constant?
```suggestion
static ulonglong start_position, stop_position;
#define stop_position_default ~0ULL
static ulonglong last_processed_position= 0;
```
or `ULLONG_MAX` or `std::numeric_limits<unsigned long long>::max()`…

---

Rather, specializing this value would put ~0 as an input out of support, though I don’t think it has a practical difference from “your file is bigger than the legal limit”.

client/mysqlbinlog.cc
```diff
@@ -3138,6 +3133,15 @@ int main(int argc, char** argv)
...
             stop_datetime_str);
 
+  /*
+    Emit a warning in the event that we finished processing input
+    before reaching the boundary indicated by --stop-position.
+  */
+  if ((static_cast<longlong>(stop_position) != stop_position_default) &&
+      stop_position > last_processed_position)
+    warning("Did not reach stop position %llu before end of input",
+            stop_position);
```
ParadoxV5, line 3143: Funny. It used to check/warn datetime after position, but now it’s before. 😀
No need to change it back though. Positions are “absolute” while datetime (which can go out-of-order) are “relative”, so it feels intuitive to have that last.

ParadoxV5 (approved):

client/mysqlbinlog.cc
```diff
@@ -1005,7 +1008,8 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
   events to process has been reached and the program should terminate.
 */
 Exit_status process_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
```
andrelkin, line 1010: Actually `ev_end_pos` is (always) `pos + ev->data_written`. The signature change can be converted into  `assert(old_off + ev->data_written == my_b_tell(file))` before [process_event(...)](https://github.com/MariaDB/server/blob/review-10.5-Modev-35694/client/mysqlbinlog.cc#L2956).

client/mysqlbinlog.cc
```diff
@@ -1462,6 +1466,7 @@ Exit_status process_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
 end:
   rec_count++;
   last_processed_datetime= ev_when;
```
andrelkin, line 1468: It feels like these two are a tightly coupled pair, to suggest arrange them as a struct (of two members) which
would single-line the pair of the current assignments (not a big deal) but more interestingly this measure would unite "artificially" separated [1](https://github.com/MariaDB/server/blob/review-10.5-Modev-35694/client/mysqlbinlog.cc#L3131)`+`[2](https://github.com/MariaDB/server/blob/review-10.5-Modev-35694/client/mysqlbinlog.cc#L3140) stop-condition checks into something like `if (stop_condition > last_processed)` where the compared two are structs.

andrelkin (changes requested):
The patch looks good. To my check it can simplified and tidied up (that's perhaps is rather up to aesthetics criteria though)  a bit.

bnestere:
Closing this, as it was pushed into 10.11 ([588f7a5af7e](https://github.com/MariaDB/server/commit/588f7a5af7e25bb4e2a00bea22b67308ad1ef1a6)) and the branch targets 10.5.

PR #3758, opened 2025-01-13
MDEV-35837 Move to c++17

DaveGosselin-MariaDB:
Move from c++11 to c++17.


plugin/handler_socket/libhsclient/hstcpcli.hpp
```diff
@@ -35,7 +35,7 @@ struct hstcpcli_filter {
 };
 
 struct hstcpcli_i;
-typedef std::auto_ptr<hstcpcli_i> hstcpcli_ptr;
+typedef std::unique_ptr<hstcpcli_i> hstcpcli_ptr;
```
dr-m, line 38: This is C++11 and could be done separately in 10.5 already, to fix MDEV-35632. At the same time, the `#pragma GCC` that were added in 9a25f2a5bbc084df725538b8b3e9d2019cbacd31 had better be removed.

CMakeLists.txt
```diff
@@ -85,7 +85,7 @@ IF(CMAKE_VERSION VERSION_LESS "3.1")
   ENDIF()
 ELSE()
   SET(CMAKE_C_STANDARD 99)
-  SET(CMAKE_CXX_STANDARD 11)
+  SET(CMAKE_CXX_STANDARD 17)
```
dr-m, line 88: The description is promising C11, but it was not changed.

DaveGosselin-MariaDB: I changed the commit message but forgot to update the PR description, it is fixed now.

plugin/handler_socket/libhsclient/hstcpcli.hpp
DaveGosselin-MariaDB: Thank you for pointing these out, I was unaware.

dr-m: Thank you for demonstrating that the fix is so simple. I didn’t know; I never used the C++98 `auto_ptr`.

CMakeLists.txt
svoj: @dr-m IIRC C11 atomics are not widely supported, e.g. MSVC didn't have them.

plugin/handler_socket/libhsclient/hstcpcli.hpp
DaveGosselin-MariaDB: I updated the PR to address these, thanks again.

CMakeLists.txt
vaintroub: https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/
That's eons ago (2 years is eons in MSVC), which means we have atomics on MSVC everywhere on build machines

dr-m (approved):

DaveGosselin-MariaDB:
@dr-m thanks for the review and approval.  I need to push now to the branch Sergei mentioned so that we can get all builders tested.  If it's okay, then we can merge after that.

dr-m (approved):
Thank you. I hope that you can push the `auto_ptr` removal fix also to 10.5 separately. And I still think that it would be a good idea to enable C11 or later and make use of it, in a separate task.

vaintroub (approved):
Looks good to me.

PR #3760, opened 2025-01-14
MDEV-35838 libressl support differences in CRYPTO_set_mem_functions

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35838*

## Description

Based on FreeBSD patch.

The FreeBSD inclusion of check_openssl_compatibility prevented any use of the crypto callbacks and as such the later differences where ignored.

The later differences in coc_malloc didn't propegate to the other callback functions of CRYPTO_set_mem_functions where a reduced argument list also applied.

Looking where[2] libressl added the functions it was of the same prototype 10 years ago so omitting any version check.

[1] https://github.com/freebsd/freebsd-ports/blob/a34cf9c2dbf503c8248371ba3bab24f34d2d045d/databases/mariadb106-server/files/patch-mysys__ssl_openssl.c
[2] https://github.com/libressl/openbsd/commit/5ebad8aceae77c6da6a1e47c3f7e70e8ffae3dae#diff-7f393e5489e6c5780773408f11ca27d9b3bb8f55b174631a1e9467a1dd3010b9R22

## Release Notes

Add compatibility for libressl

## How can this PR be tested?

Build/test against libressl and all ssl mtr tests shouldn't leak memory.

- [X] *This is a compatibility fix, narrowly scoped, for a previously unsupported version.@

## PR quality check
- [?] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.

FYI @Sp1l because the patch you did I think bypasses the fix by not using CRYPTO_set_mem_functions (hopefully explained in commit message).

### DIFF
--- a/mysys_ssl/openssl.c
+++ b/mysys_ssl/openssl.c
@@ -36,8 +36,12 @@ int check_openssl_compatibility()
 static uint testing;
 static size_t alloc_size, alloc_count;
 
-static void *coc_malloc(size_t size, const char *f __attribute__((unused)),
-                                             int l __attribute__((unused)))
+static void *coc_malloc(size_t size
+#ifndef LIBRESSL_VERSION_NUMBER
+                        , const char *f __attribute__((unused)),
+                        int l __attribute__((unused))
+#endif
+)
 {
   if (unlikely(testing))
   {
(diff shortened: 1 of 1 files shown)


vaintroub (approved):
Looks good to me

grooverdan:
FYI @rozhuk-im  as you appear to be the reporter here https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=274494 and knowing about these things. Please let me know if I did something incorrect.

rozhuk-im:
> FYI @rozhuk-im as you appear to be the reporter here https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=274494 and knowing about these things. Please let me know if I did something incorrect.

LGTM.
But more than year ago I switch out from LibreSSL due to ugly support and FBSD community ignore LibreSSL issues and patches: as you can see it take 10 months to commit simple patch in ports tree.

grooverdan:
Thanks @rozhuk-im,

:grimacing: , sometimes its easy to overlook dates but thanks for lighting them. Probably a OpenBSD support in MariaDB might be the reliable path to LibreSSL compatibility rather than us suffering with FBSD/LibreSSL on top of our own bugs.

PR #3763, opened 2025-01-14
MDEV-35847 mariadb client --wait flag broken

DaveGosselin-MariaDB:
Wire up the --wait flag in the mariadb client.


vuvova:
kind of risky to enable it after it was disabled for 23 years. I'd rather keep the status quo and remove it completely

DaveGosselin-MariaDB:
I'm curious why the last committer intended to delete this option because the rest of the logic for it is in place.  It would've been trivial to assume `false` everywhere that `wait_flag` is used and simplified the code accordingly.

vuvova:
I'm not sure the last committer has actually intended that. It's quite possibly could've been a mistake.

I don't have a strong opinion about it. It was a dead code for 23 years, so clearly nobody cares whether it works or not and it could be safely deleted. On the other hand, `mariadb-admin` has it (and it works there), so for consistency `mariadb` could have it too.

DaveGosselin-MariaDB:
I'm biased (since I wrote the patch 😇 ) but really it's up to you.  I'm happy to finish removing the flag, but like you said it would be nice for feature symmetry between the two clients (and maybe someone is using it for the admin client).

vuvova:
you wrote you need it for some tooling, so it's plausible that someone else might need it too. let's keep it

DaveGosselin-MariaDB:
OK 😄 once you approve the PR then I'll merge.  The target branch is 10.5, let me know if you'd rather it go somewhere else instead (11.8?).

client/mysql.cc
```diff
@@ -1732,7 +1732,7 @@ static struct my_option my_long_options[] =
    GET_NO_ARG, NO_ARG, 0, 0, 0, 0, 0, 0},
   {"vertical", 'E', "Print the output of a query (rows) vertically.",
    &vertical, &vertical, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
-  {"wait", 'w', "Wait and retry if connection is down.", 0, 0, 0, GET_NO_ARG,
+  {"wait", 'w', "Wait and retry if connection is down.", &wait_flag, 0, 0, GET_BOOL,
```
vuvova, line 1735: that's not how it used to work. see commit c9d04284c8e (wait time was configurable, retrying forever)

and it's not how `-w` works in `mariadb-admin` (wait time is hard-coded, 5 seconds, number of retries is configurable)

For consistency, I suppose, it should do what `mariadb-admin` does

vuvova (changes requested):

client/mysql.cc
```diff
@@ -1924,6 +1924,16 @@ get_one_option(const struct my_option *opt, const char *argument, const char *)
...
       verbose++;
     break;
+  case 'w':
+    if (argument == disabled_my_option)
+      wait_flag= 0;
+    else
+    {
+      wait_flag= 1;
+      if (argument)
+        wait_retries= atoi(argument);
```
vuvova, line 1934: that is, no argument means "retry once"? Looks very weird. As far as I can see you have
```
$ mariadb
```
will retry forever, but
```
$ mariadb -w
```
will retry once?

Why not to do it as in `mariadb-admin` ? By default no retry. `-w` means retry forever. `-w number` means retry that many times.

Except that `--skip-wait` should mean "no retry" as you did above, and `mariadb-admin` doesn't have it, I suggest you fix there.

PR #3731, opened 2025-01-07
Create `rpl_master_retry_count.test`

ParadoxV5:
* [x] ~~*The Jira issue number for this PR is: MDEV-35304*~~

## Description
I created a MTR test to test `--master-retry-count` (and also `master_connect_retry` as a dependency) because I suspect its `0` edge case doesn’t match [the KB description](https://mariadb.com/kb/en/mariadbd-options/#-master-retry-count).
My local build doesn’t have `debug_sync` so I’m borrowing our buildbot.

## Release Notes
N/A (if the edge case I suspected doesn’t apply)

## How can this PR be tested?
Neither `master_retry_count` ([MDEV-25674](https://jira.mariadb.org/browse/MDEV-25674)) nor `master_connect_retry` are global variables, so I had to hardcode them.

## Basing the PR against the correct MariaDB version
* [ ] *This is a new feature or a refactoring, and the PR is based against the `main` branch.*
* [ ] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
* [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
* [x] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #3764, opened 2025-01-14
MDEV-35304 & MDEV-25674: Add `Connects_Tried` and `Master_Retry_Count` to SHOW REPLICA STATUS

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-35304][]*
* [x] *The Jira issue number for this PR is: [MDEV-25674][]*

Disambiguation: This does not concern [`--spider-connect-retry-count`](https://mariadb.com/kb/en/spider-system-variables/#spider_connect_retry_count).

## Description
These new `Master_info` numbers supplement SHOW ~~SLAVE~~ _REPLICA_ STATUS’s (most-recent) ‘Connecting’ state with statistics on (re)connect attempts:

* `Connects_Tried`: how many retries have been attempted so far
  * This was previously a local variable that only counts re-attempts; it’s now meaningful even after the “Connecting” state concludes.
  * **Like Master/Slave_last_event_time, RESET REPLICA resets this counter to 0.**
* `Master_Retry_Count`: out of how many configured
  * Yep, it’s the global option `--master-retry-count`, now per-connection.
    * `performance_schema.replication_connection_configuration.CONNECTION_RETRY_COUNT` (the old, less user-friendly view) is now per-replica too.
  * Included in this PR is its addition to CHANGE ~~PRIMARY~~ _MASTER_ … TO.
    * It’s more preferrable for solving [MDEV-25674][] than a global variable as originally proposed.
    * **This config is `master.info`-persisted at the same entry as that for MySQL.**
    * That’s right – [MySQL’s CHANGE MASTER TO](https://dev.mysql.com/doc/refman/8.0/en/change-master-to.html) already had a MASTER_RETRY_COUNT entry! Oh how did I like them :green_apple:s…
      * Though they’ve since replaced their *entire repo* to [CHANGE REPLICATION SOURCE TO](https://dev.mysql.com/doc/refman/8.4/en/change-replication-source-to.html), variable names included.
        Their MASTER–SLAVE keywords are now no more (RIP their backward compatibility).
    * Speaking of backward compatibility – **`--master-retry-count` is not deprecated ~~(yet)~~ and remains the default for new `CHANGE MASTER TO`s.**
      * [MDEV-18777](https://jira.mariadb.org/browse/MDEV-18777) aside, ‘Master_Retry_Count’ is so bad a name though… unfluent, and misleading too (it caps both first-tries and retries).
        [We’re not goïng to deprecate `--master-retry-count`](https://jira.mariadb.org/browse/MDEV-36170) (this replacement’s :new:, after all); but if we do, ideally we could rename the field to say `Master_Max_Connect_Tries`.
    * **Due to implementation, 0 here means “keep current config”, not infinite ([not really](https://github.com/MariaDB/server/pull/3764#discussion_r1917630832)) as specified by `--master-retry-count`!**
      *This matches the behavior of Connect_Retry … FWIW…*

This PR also includes additional commits that refactor relevant code.
I strongly recommend reviewing commit-by-commit.

[MDEV-35304][] — When the IO thread (re)connect to a primary, no updates are available besides unique errors that cause the failure.

### If some output changed that is not visible in a test case, what was it looking like before the change and how it's looking with this patch applied?
From https://github.com/MariaDB/server/pull/3764#discussion_r1917643857, appearently:
* `--master-retry-count=0` was in practice effectively `=1` for reconnecting to a previously connected primary.
* There may have been an extra sleep in `try_to_reconnect()`.
  In full, looks like this happens when the connection drops at a different substep in the setup phase.

Both of these code are now gone as part of a preparation commit.

Component that expects a fixed number of SHOW REPLICA STATUS entries in fixed sections might start seeïng uninvited guests.
On that note, *we have many MTR tests that run `show slave status` directly and in full (i.e., not using `source include/show_slave_status.inc`) include non-deterministic entries if any.*

## Release Notes
Added two new SLOW REPLICA STATUS entries for statistics of the most-recent ‘Connecting’ state:
* `Connects_Tried`: How many tries were attempted
* `Master_Retry_Count`: Out of how many configured
  * CHANGE MASTER TO can change this (default: `--master-retry-count`).

Of course, they’re also accessible from INFORMATION_SCHEMA.SLAVE_STATUS; e.g.:
```sql
SELECT
  Connection_name,
  Slave_IO_Running,
  JOIN(Connects_Tried, 'attempt(s) used out of ', Master_Retry_Count, ' max')
FROM INFORMATION_SCHEMA.SLAVE_STATUS;
```

### [Knowledge Base](https://mariadb.com/kb/) pages that need changing
* https://mariadb.com/kb/en/show-replica-status/
* https://mariadb.com/kb/en/change-master-to/
* https://mariadb.com/kb/en/reset-replica/ (re. RESET to 0)

## How can this PR be tested?
* `Connects_Tried`: `rpl.rpl_connects_tried` observes its change over a `sleep`.
* `Master_Retry_Count`: `rpl.rpl_master_retry_count_basic` asserts its basic behavior as a CHANGE MASTER TO (and SHOW REPLICA STATUS) variable.
  * It does not cover behavior testing, as that’ll stretch its scope, likely to having to address https://github.com/MariaDB/server/pull/3764#discussion_r1917630832 and https://github.com/MariaDB/server/pull/3764#discussion_r1917643857.
    A draft is at #3731.

## PR quality check

[MDEV-35304]: https://jira.mariadb.org/browse/MDEV-35304
[MDEV-25674]: https://jira.mariadb.org/browse/MDEV-25674


sql/slave.cc
```diff
@@ -4381,62 +4358,51 @@ on this slave.\
...
   mi->slave_running= MYSQL_SLAVE_RUN_NOT_CONNECT;
-  thd->proc_info= messages[SLAVE_RECON_MSG_WAIT];
 #ifdef SIGNAL_WITH_VIO_CLOSE  
   thd->clear_active_vio();
 #endif
-  end_server(mysql);
-  if ((*retry_count)++)
-  {
-    if (*retry_count > master_retry_count)
-      return 1;                             // Don't retry forever
```
ParadoxV5, line 4434: This can be hard to comprehend. Step by step:
1. If `*retry_count` was 0, skip the block
2. Else If `*retry_count` becomes `> master_retry_count`, `return 1`
3. Else, proceed with the block

This could mean that for `--master-retry-count=0`, the termination condition always matches, but the flow always bypasses it for the first try. That is, `0` is equivalent to `1`.

I tried writing an MTR test for `--master-retry-count`, but my build doesn’t have `debug_sync`, and appearently none of our buildbots have both it and `log_bin`…??? (#3731)

ParadoxV5 (commented):
According to the code, the `--master-retry-count=0` behavior appearently doesn’t match the doc’s expectation:

> A value of 0 means the replica will not stop attempting to reconnect.
> ⸺ https://mariadb.com/kb/en/mariadbd-options/#-master-retry-count

This may be a bug.
The code is found as early as 10.5.

ParadoxV5:
`innodb.log_file_size_online` and `main.log_state` fails on `main`.

sql/slave.cc
ParadoxV5, line 1: Room for refactor that I excluded from this PR:
* There are a few instances of
  ```cc
  if (!…)
  …
  else
  …
  ```
* Both `register_slave_on_master` and `request_dump` provides failure warnings, yet their caller, `handle_slave_io`, still emits similar warnings on their failure.
* Extraneous EOL whitespaces

ParadoxV5:
I’m done with staring at these consistent but incomprehensible test failures.

knielsen (changes requested):
Hi ParadoxV5,

Here is my review of MDEV-35304:

> diff --git a/mysql-test/include/check-testcase.inc b/mysql-test/include/check-testcase.inc
> index 25990c1a9e379..4ae0da30d7bba 100644
> --- a/mysql-test/include/check-testcase.inc
> +++ b/mysql-test/include/check-testcase.inc
> @@ -74,10 +74,12 @@ if ($tmp)
>    --echo Slave_Non_Transactional_Groups	#
>    --echo Slave_Transactional_Groups	#
>    --echo Replicate_Rewrite_DB	#
> +  --echo Connects_Tried	#
> +  --echo Primary_Retry_Count	#

IIUC, this is in preparation for adding a MASTER_RETRY_COUNT to the CHANGE
MASTER, MDEV-25674.

But this is not done in this patch. So let's add that output to SHOW SLAVE
STATUS as part of MDEV-25674, not as part of this change. SHOW SLAVE STATUS
is for status about the running slave and for values setable with CHANGE
MASTER, not for showing the values of configuration parameters.

Second, when this is added (here or as part of MDEV-25674), use consistent
naming using the primary Master/Slave terminology, do not introduce a mix
with the alternate terminology primary/replica.

So "Master_Retry_Count". But I would suggest that when this is added as
part of MDEV-25674, to name it better as "Master_Max_Connect_Tries" or
something like that. To make it clear that 1) it is the configured maximum,
not any current count, and 2) it is the total number of connect tries,
whereas "retry" might suggest it does not include the initial try.


> diff --git a/mysql-test/suite/rpl/t/rpl_connects_tried.test b/mysql-test/suite/rpl/t/rpl_connects_tried.test
> new file mode 100644
> index 0000000000000..b9f5dd3163030
> --- /dev/null
> +++ b/mysql-test/suite/rpl/t/rpl_connects_tried.test
> @@ -0,0 +1,52 @@

> +--SOURCE include/have_debug.inc
> +--SOURCE include/have_debug_sync.inc
> +--SOURCE include/have_log_bin.inc # The test is agnostic of binlog formats.
> +--SOURCE include/master-slave.inc

Please use lower-case for mysqltest commands --source, --let, --connection,
etc. to be consistent with other test files.

> +
> +--LET $rpl_server_number= 1
> +--LET $status_items= Slave_IO_Running
> +
> +#--CALL mtr.add_suppression("Slave I/O: Master command COM_REGISTER_SLAVE failed: .*")
> +#--CALL mtr.add_suppression("Slave I/O: .* failed with error: Lost connection to MySQL server at 'reading initial communication packet'")
> +#--CALL mtr.add_suppression("Fatal error: The slave I/O thread stops because master and slave have equal MySQL server ids; .*")
> +#--CALL mtr.add_suppression("Slave I/O thread .* register on master")

Why are these commented out but left in the test case?


> diff --git a/mysql-test/suite/rpl/t/rpl_heartbeat_debug.test b/mysql-test/suite/rpl/t/rpl_heartbeat_debug.test
> index e593786655bcb..8b1d7517b2e1e 100644
> --- a/mysql-test/suite/rpl/t/rpl_heartbeat_debug.test
> +++ b/mysql-test/suite/rpl/t/rpl_heartbeat_debug.test
> @@ -5,7 +5,7 @@
>  
>  --disable_query_log
>  CALL mtr.add_suppression('SET @master_heartbeat_period to master failed with error');
> -CALL mtr.add_suppression('Master command COM_REGISTER_SLAVE failed: failed registering on master, reconnecting to try again');
> +CALL mtr.add_suppression('Master command COM_REGISTER_SLAVE failed: Replica I/O thread: register on primary failed, reconnecting to try again');

Should be "Slave I/O thread: ...", see comment below.


> diff --git a/sql/mysqld.cc b/sql/mysqld.cc
> index 89d9cc646c8ca..f69d6fcfc5ae0 100644
> --- a/sql/mysqld.cc
> +++ b/sql/mysqld.cc
> @@ -780,7 +780,7 @@ File_parser_dummy_hook file_parser_dummy_hook;
>  
>  /* replication parameters, if master_host is not NULL, we are a slave */
>  uint report_port= 0;
> -ulong master_retry_count=0;
> +ulong master_retry_count=100000;

> @@ -6820,7 +6820,7 @@ struct my_option my_long_options[]=
>    {"master-retry-count", 0,
>     "The number of tries the slave will make to connect to the master before giving up",
>     &master_retry_count, &master_retry_count, 0, GET_ULONG,
> -   REQUIRED_ARG, 100000, 0, 0, 0, 0, 0},
> +   REQUIRED_ARG, static_cast<longlong>(master_retry_count), 0, 0, 0, 0, 0},

This looks strange. These are statically initialized global variables. How
can you initialize one of them with the value of the other? Better use a
#define or static constexpr to initialize them both. Or do you have a
reference to somewhere it says this is valid standard c++?

A question: The documentation for --master-retry-count states:

  Number of times a replica will attempt to connect to a primary before
  giving up.

  A value of 0 means the replica will not stop attempting to reconnect.

Are these still true after this patch?


> diff --git a/sql/rpl_mi.h b/sql/rpl_mi.h
> index 37a5fc30099c9..de8a422ae9498 100644
> --- a/sql/rpl_mi.h
> +++ b/sql/rpl_mi.h
> @@ -253,7 +253,12 @@ class Master_info : public Slave_reporting_capability

> +  /** replica-specific @ref master_retry_count */

s/replica/slave/

> +  ulong retry_count;

Better name, for example max_connect_tries. This is easily confused with a
count of the current count of connection attempts that failed.


> diff --git a/sql/slave.cc b/sql/slave.cc
> index e16834a387da3..1d748789a245a 100644
> --- a/sql/slave.cc
> +++ b/sql/slave.cc
> @@ -102,57 +102,33 @@ LEX_CSTRING default_master_connection_name= { (char*) "", 0 };

> +  reg=
> +  {
> +    "Reconnecting after a failed registration on primary",
> +    "register on primary",

Keep to the standard terminology master/slave, don't introduce a mix with
the alternate primary/replica terminology except in optional user aliases.

And best not to change the error message wording unless there is a good
reason to do so. Users may occasionally have regexps or something matching
against the old wording.

> +    "Reconnecting after a failed primary event read",
> +    "primary event read",

s/primary/master/


> @@ -2609,10 +2586,6 @@ when it try to get the value of TIME_ZONE global variable from master.";
>    DBUG_RETURN(0);
>  
>  network_err:
> -  if (master_res)
> -    mysql_free_result(master_res);
> -  DBUG_RETURN(2);
> -
>  slave_killed_err:

Agree, thanks for spotting the redundancy.
Did you consider merging the two goto labels into one also? But just a
suggestion, up to you.


> @@ -4445,36 +4409,35 @@ static int try_to_reconnect(THD *thd, MYSQL *mysql, Master_info *mi,

> -    my_snprintf(buf, sizeof(buf), messages[SLAVE_RECON_MSG_FAILED], 
> -                IO_RPL_LOG_NAME, mi->master_log_pos,
> +    my_snprintf(buf, sizeof(buf), "Replica I/O thread: %s failed, "

s/Replica/Slave/


> @@ -4695,7 +4656,7 @@ pthread_handler_t handle_slave_io(void *arg)
>  
>    DBUG_PRINT("info",("Starting reading binary log from master"));
>    thd->set_command(COM_SLAVE_IO);
> -  while (!io_slave_killed(mi))
> +  if (!io_slave_killed(mi))
>    {

So this changes a nested while() { ... while() { ... } ... } into a
single while() inside an if().

It was hard for me to see exactly what the reason was for this in the patch
/ why it is correct, can you try to explain the reason for it?


> @@ -7038,16 +6998,10 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
>                   " - retry-time: %d  maximum-retries: %lu  message: %s",
>                   (reconnect ? "reconnecting" : "connecting"),
>                   mi->user, mi->host, mi->port,
> -                 mi->connect_retry, master_retry_count,
> +                 mi->connect_retry, mi->retry_count,
>                   mysql_error(mysql));
>      }
> -    /*
> -      By default we try forever. The reason is that failure will trigger
> -      master election, so if the user did not set master_retry_count we
> -      do not want to have election triggered on the first failure to
> -      connect
> -    */
> -    if (++err_count == master_retry_count)
> +    if (++(mi->connects_tried) == mi->retry_count)

Right, I think this makes --master-retry-count=0 mean forever, ok (partly
answering my own question above).


> diff --git a/sql/sql_show.cc b/sql/sql_show.cc
> index 7f4781c1cd5b9..cdc4209faf2de 100644
> --- a/sql/sql_show.cc
> +++ b/sql/sql_show.cc
> @@ -9201,14 +9201,16 @@ int make_proc_old_format(THD *thd, ST_SCHEMA_TABLE *schema_table)
>    Constants for columns that are present in
>    SHOW ALL SLAVES STATUS
>    that are not in SHOW SLAVE STATUS. Specifically, columns 0 and 1, and
> -  everything at and above 56.
> +  everything between 56 and 64 inclusive.
>      0: Connection_name
>      1: Slave_SQL_State
>      56: Retried_transactions
> +    64: Master_Slave_time_diff
>  */
>  #define SLAVE_STATUS_COL_CONNECTION_NAME 0
>  #define SLAVE_STATUS_COL_SLAVE_SQL_STATE 1
>  #define SLAVE_STATUS_COL_RETRIED_TRANSACTIONS 56
> +#define SLAVE_STATUS_COL_MASTER_SLAVE_TIME_DIFF 64

No, I think instead SLAVE_STATUS_COL_RETRIED_TRANSACTIONS should be changed
to 58 (or 57 until MDEV-25674), see below.


> @@ -9220,18 +9222,13 @@ static int make_slave_status_old_format(THD *thd, ST_SCHEMA_TABLE *schema_table)
>    for (uint i=0; !field_info->end_marker(); field_info++, i++)
>    {
>      if (all_slaves ||
> -        // not SLAVE_STATUS_COL_CONNECTION_NAME,
> -        // SLAVE_STATUS_COL_SLAVE_SQL_STATE
> -        // and less
> -        // SLAVE_STATUS_COL_RETRIED_TRANSACTIONS
> -        !(i <= SLAVE_STATUS_COL_SLAVE_SQL_STATE ||
> -          i >= SLAVE_STATUS_COL_RETRIED_TRANSACTIONS))
> +        (i > SLAVE_STATUS_COL_SLAVE_SQL_STATE &&
> +         i < SLAVE_STATUS_COL_RETRIED_TRANSACTIONS) ||
> +        i > SLAVE_STATUS_COL_MASTER_SLAVE_TIME_DIFF)
>      {
>        LEX_CSTRING field_name= field_info->name();
>        Item_field *field= new (thd->mem_root)
>          Item_field(thd, context, field_name);
> -      DBUG_ASSERT(all_slaves || (i > SLAVE_STATUS_COL_SLAVE_SQL_STATE &&
> -                                 i < SLAVE_STATUS_COL_RETRIED_TRANSACTIONS));
>        if (!field || add_item_to_list(thd, field))
>          return 1;
>      }
> @@ -10710,6 +10707,8 @@ ST_FIELD_INFO slave_status_info[]=
>    Column("Master_last_event_time", Datetime(0), NULLABLE),
>    Column("Slave_last_event_time", Datetime(0), NULLABLE),
>    Column("Master_Slave_time_diff", SLonglong(10), NULLABLE),
> +  Column("Connects_Tried", ULonglong(20), NOT_NULL),
> +  Column("Primary_Retry_Count", ULonglong(20), NOT_NULL),
>    CEnd()

No, I think this is wrong.

When new entries are added to the SHOW SLAVE STATUS output, they are added
after all the existing columns output from SHOW SLAVE STATUS, but _before_
the columns output by SHOW ALL SLAVES STATUS. Somewhat tricky, but that is
why there is now an INFORMATION_SCHEMA view that allows to select just the
columns of interest and not depend on column order.

So I think the patch should add the Connects_Tried column just after
Replicate_Rewrite_DB (and MDEV-25674 can add Master_Max_Connect_Tries after
that again). And then you do not need to change the if() condition or the
DBUG_ASSERT above.

And s/Primary_Retry_Count/Master_Max_Connect_Tries/ as noted above.

 - Kristian.

ParadoxV5 (commented):
Thank you for your thorough review, Master @knielsen.

Some side tips about GitHub:
* It uses [Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax), so your plaintext review rendered oddly on the web.
  (Perhaps this plaintext is what a mailing list reply looks like, complete with a line width limit?)
* Instead of quoting diffs, you can create threads on specific line(s) or a file in a PR’s [“Files changed” tab](https://github.com/MariaDB/server/pull/3764/files).

For oraganization, I’ve replied in said threads (if you can see them, that is 😅.)

sql/sql_show.cc
```diff
@@ -9201,14 +9201,16 @@ int make_proc_old_format(THD *thd, ST_SCHEMA_TABLE *schema_table)
...
+  everything between 56 and 64 inclusive.
     0: Connection_name
     1: Slave_SQL_State
     56: Retried_transactions
+    64: Master_Slave_time_diff
 */
 #define SLAVE_STATUS_COL_CONNECTION_NAME 0
 #define SLAVE_STATUS_COL_SLAVE_SQL_STATE 1
 #define SLAVE_STATUS_COL_RETRIED_TRANSACTIONS 56
+#define SLAVE_STATUS_COL_MASTER_SLAVE_TIME_DIFF 64
```
ParadoxV5, line 9213: > No, I think this is wrong.

*start a thread for myself so I don’t forget*
(But yeah, something overcomplicated is goïng on with this “add at the end only” limitation.)

knielsen:
ParadoxV5 ***@***.***> writes:

>> Second, when this is added (here or as part of MDEV-25674), use consistent
> naming using the primary Master/Slave terminology, do not introduce a mix
> with the alternate terminology primary/replica.
>
> So MDEV-18777 is too soon for new columns by themselves? 👌

It is not about "too soon", it is very important to keep consistent. We
cannot change the existing names due to the need for backwards
compatibility, so we must follow the convention for newly added options as
well.

> I was also thinking about deprecating `--master-retry-count` itself
> too in favor of MDEV-25674. This could be another argument.

We must not deprecate in a way that breaks backwards compatibility unless
there is a very good technical reason, so I think --master-retry-count will
stay (as the default for a new CHANGE MASTER option, which does make sense).

> Switching to the ‘primary’ term as part of a “rename” is quite tempting though.

It would seem quite horrible if users would need to remember for each option
whether to use one terminology or another. Consistency is important.

Introducing optional aliases is ok (though in general there needs to be a
significant interest in this to warrent the added complexity. There does
seems to be a lot of such interest in this case, or at least a group of very
vocal advocates).

One option if people really need to s/master/primary/ in the status output
could be that SHOW MASTER STATUS uses the normal terminology, and SHOW
REPLICA STATUS uses an alternate naming.

> Or maybe throw out the ‘master’ part like `Connect_Tried` did (for
> disambiguation) – `Max_Connect_Tries`? 🤔 clean

If you look at CHANGE MASTER syntax, the options related to the master
connection have this "MASTER_" prefix. It is not about whether this is the
best way to design it, it is just how it is now and litterally millions of
installations depend on it. So we need to keep consistent with this.

Then in SHOW SLAVE STATUS, when you present the value that was configured
by the user in CHANGE MASTER, it should be using the same name as the
option, otherwise it is confusing.

But "Connect_Tried" does not refer to a named option, so there I think it's
ok to omit the "MASTER_" prefix.

>>    {"master-retry-count", 0,
>     "The number of tries the slave will make to connect to the master before giving up",
>     &master_retry_count, &master_retry_count, 0, GET_ULONG,
> -   REQUIRED_ARG, 100000, 0, 0, 0, 0, 0},
> +   REQUIRED_ARG, static_cast<longlong>(master_retry_count), 0, 0, 0, 0, 0},
>
>> This looks strange. These are statically initialized global variables. How
> can you initialize one of them with the value of the other? Better use a
> #define or static constexpr to initialize them both. Or do you have a
> reference to somewhere it says this is valid standard c++?
>
> It was by intuition when I made that change.
> But I’ve now looked up: non-`static` global variables in a (read: the
> same) translation unit (i.e., `.cpp`) fall under ordered [dynamic
> initialization](https://en.cppreference.com/w/cpp/language/initialization#Dynamic_initialization).

Nice, thanks for the reference! Interesting, I didn't know that.
Then yes, your way looks correct and agree it is more natural.

>    DBUG_PRINT("info",("Starting reading binary log from master"));
>    thd->set_command(COM_SLAVE_IO);
> -  while (!io_slave_killed(mi))
> +  if (!io_slave_killed(mi))
>    {
>
>> So this changes a nested while() { ... while() { ... } ... } into a
> single while() inside an if().

> * Both `while`s have the same condition of `!io_slave_killed(mi)`.
> * There is no code that runs after the inner loop concludes before the outer’s `}`.
>
> A killed replica cannot resurrect (*right?*), especially during the
> few instructions of the inner exiting to the outer.

Ah, yes, now I see. Thanks for the explanation, I missed that, agree.

> I believe they were not true even before this patch.
>
> C/C++ integers are not limitless (unlike Ruby – I main Ruby 💎).
> The counter (new or old) will rollover to `0` when it performs `++ULONG_MAX`.
>
> So it’s not *forever* forever, only practically forever.

Right, that's probably fine.

> @@ -780,7 +780,7 @@ File_parser_dummy_hook file_parser_dummy_hook;
> /* replication parameters, if master_host is not NULL, we are a slave */

> while on this file – Where master_host?

Yeah, I don't see it either. This comment goes back 25 years, most likely
refers to ancient code that is long gone.
Be prepared to encounter that kind of stuff again when working in the
replication code :-)

 - Kristian.

sql/slave.cc
```diff
@@ -7038,16 +7000,10 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
...
                  mysql_error(mysql));
     }
-    /*
-      By default we try forever. The reason is that failure will trigger
-      master election, so if the user did not set master_retry_count we
-      do not want to have election triggered on the first failure to
-      connect
-    */
-    if (++err_count == master_retry_count)
+    if (++(mi->connects_tried) == mi->retry_count)
```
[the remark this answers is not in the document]
ParadoxV5: ~~archived~~ as https://jira.mariadb.org/browse/MDEV-36002

mysql-test/suite/rpl/r/rpl_connects_tried.result
```diff
@@ -0,0 +1,51 @@
...
+include/wait_for_slave_param.inc [Slave_IO_Running]
+SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
+Connection_name	Connects_Tried
+	5
+named	3
+STOP ALL REPLICAS;
+Warnings:
+Note	1938	SLAVE 'named' stopped
+Note	1938	SLAVE '' stopped
+include/wait_for_slave_to_stop.inc
```
ParadoxV5, line 42: These warnings may not be deterministic.
How else to do it without writing STOP REPLICA once per connection?

mysql-test/suite/rpl/t/rpl_connects_tried.test
```diff
@@ -0,0 +1,60 @@
...
+# not-master-slave.inc
+--let $rpl_skip_start_slave= 1
+--source include/master-slave.inc
+--let $rpl_server_number= 1
+--source include/rpl_stop_server.inc # shut the primary down
+--connection slave
+
+CHANGE MASTER TO master_connect_retry=1;
+--replace_result $MASTER_MYPORT MASTER_MYPORT
+--eval CHANGE MASTER 'named' TO master_host='localhost', master_port=$MASTER_MYPORT, master_user='root', master_connect_retry=2
```
ParadoxV5, line 13: This is theoretically inconsistent if a testing machine can take seconds to run a `SELECT … FROM information_schema.SLAVE_STATUS`.
The alternative is to rely on `debug_sync`…

mysql-test/suite/funcs_1/r/is_columns_is.result
```diff
@@ -434,16 +436,16 @@ def	information_schema	SLAVE_STATUS	Replicate_Ignore_Table	18	NULL	NO	varchar	21
 def	information_schema	SLAVE_STATUS	Replicate_Rewrite_DB	56	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
 def	information_schema	SLAVE_STATUS	Replicate_Wild_Do_Table	19	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
 def	information_schema	SLAVE_STATUS	Replicate_Wild_Ignore_Table	20	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
-def	information_schema	SLAVE_STATUS	Retried_transactions	57	NULL	NO	int	NULL	NULL	10	0	NULL	NULL	NULL	int(10) unsigned			select		NEVER	NULL	NO	NO
+def	information_schema	SLAVE_STATUS	Retried_transactions	59	NULL	NO	int	NULL	NULL	10	0	NULL	NULL	NULL	int(10) unsigned			select		NEVER	NULL	NO	NO
```
ParadoxV5, line 439: > When new entries are added to the SHOW SLAVE STATUS output, they are added after all the existing columns output from SHOW SLAVE STATUS, but _before_ the columns output by SHOW ALL SLAVES STATUS. Somewhat tricky, but that is why there is now an INFORMATION_SCHEMA view that allows to select just the columns of interest and not depend on column order.
> ⸺ @knielsen, https://github.com/MariaDB/server/pull/3764#pullrequestreview-2572711920

🤷

sql/rpl_mi.cc
```diff
@@ -569,9 +570,12 @@ file '%s')", fname);
...
 
       /* Starting from 5.5 the master_retry_count may be in the repository. */
-      if (lines >= LINE_FOR_MASTER_RETRY_COUNT &&
-	  init_strvar_from_file(buf, sizeof(buf), &mi->file, ""))
-	  goto errwithmsg;
+      if (lines >= LINE_FOR_MASTER_RETRY_COUNT)
+      {
+        if (init_strvar_from_file(buf, sizeof(buf), &mi->file, ""))
+          goto errwithmsg;
+        mi->retry_count = atol(buf);
```
ParadoxV5, line 577: I say it's better to change `init_intvar_from_file`’s `atoi` to `atoll`.

tests/mysql_client_test.c
```diff
@@ -469,7 +469,7 @@ static void test_prepare_simple()
   strmov(query, "SHOW SLAVE STATUS");
   stmt= mysql_simple_prepare(mysql, query);
   check_stmt(stmt);
-  DIE_UNLESS(mysql_stmt_field_count(stmt) == 54);
+  DIE_UNLESS(mysql_stmt_field_count(stmt) == 56);
```
ParadoxV5, line 472: 😬

ParadoxV5 (commented):
It’s v12.0 now. 🕛

mysql-test/suite/rpl/r/rpl_connects_tried.result
bnestere:
```
--disable_warnings
--source include/wait_for_slave_to_stop.inc
--enable_warnings
```

should just hide the warnings

Also to note:
```
--(disable/enable)_query_log
```
Will hide any SQL statements you may not want in the .result file (e.g. if you have a loop to execute a lot of statements).

sql/slave.cc
```diff
@@ -4382,34 +4374,30 @@ on this slave.\
...
 static bool check_io_slave_killed(Master_info *mi, const char *info)
 {
-  if (io_slave_killed(mi))
-  {
-    if (info && global_system_variables.log_warnings)
-      sql_print_information("%s", info);
-    return TRUE;
-  }
-  return FALSE;
+  bool is_io_slave_killed= io_slave_killed(mi);
```
bnestere, line 4380: Nice!

sql/slave.cc
```diff
@@ -4695,7 +4656,7 @@ pthread_handler_t handle_slave_io(void *arg)
 
   DBUG_PRINT("info",("Starting reading binary log from master"));
   thd->set_command(COM_SLAVE_IO);
-  while (!io_slave_killed(mi))
+  if (!io_slave_killed(mi))
   {
```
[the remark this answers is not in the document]
bnestere: Hmm, I'd think we could even remove this as a level of nesting altogether and merge the `io_slave_killed()` check before the next `request_dump()` call. Though this would lead to a lot of changes in indentation, which would lead to a lot of merge conflicts for anything changing IO thread specifics. Though at the very least, that should be a separate commit on its own (though still could be in the PR). Though for now, let's put a pin on that change, and we can discuss it on the next replication call.

sql/slave.cc
```diff
@@ -7024,13 +7000,13 @@ static int connect_to_master(THD* thd, MYSQL* mysql, Master_info* mi,
...
-          mysql_real_connect(mysql, mi->host, mi->user, mi->password, 0,
-                             mi->port, 0, client_flag) == 0))
+  while (!(slave_was_killed= io_slave_killed(mi)) &&
+         (reconnect ? mysql_reconnect(mysql) :
+          !mysql_real_connect(mysql, mi->host, mi->user, mi->password, 0,
+                              mi->port, 0, client_flag)))
   {
     /* Don't repeat last error */
-    if ((int)mysql_errno(mysql) != last_errno && !io_slave_killed(mi))
+    if (mysql_errno(mysql) != last_errno)
```
bnestere, line 7009: I think we should keep this. We have tests where `STOP SLAVE` is executed quickly after `START SLAVE`, and if it happens during `mysql_reconnect` or `mysql_real_connect`, the slave will report an error when we don't actually want the error (leading to sporadic failures on buildbot). See MDEV-32892 for details.

sql/sql_show.cc
```diff
@@ -9226,18 +9226,12 @@ static int make_slave_status_old_format(THD *thd, ST_SCHEMA_TABLE *schema_table)
...
-        // SLAVE_STATUS_COL_RETRIED_TRANSACTIONS
-        !(i <= SLAVE_STATUS_COL_SLAVE_SQL_STATE ||
-          i >= SLAVE_STATUS_COL_RETRIED_TRANSACTIONS))
+        (i > SLAVE_STATUS_COL_SLAVE_SQL_STATE &&
+         i < SLAVE_STATUS_COL_RETRIED_TRANSACTIONS))
     {
       LEX_CSTRING field_name= field_info->name();
       Item_field *field= new (thd->mem_root)
         Item_field(thd, context, field_name);
-      DBUG_ASSERT(all_slaves || (i > SLAVE_STATUS_COL_SLAVE_SQL_STATE &&
```
bnestere, line 9239: Hmm... I recall some funkiness here when the `DBUG_ASSERT` was added. I agree it does nothing right now, but I wonder if somewhere along the way, the intended assertion was lost along the way, and we can correct it (rather than remove it)

bnestere (changes requested):
This is great work, thanks @ParadoxV5 ! See my notes for some minor changes.

sql/sql_show.cc
ParadoxV5: https://github.com/MariaDB/server/commit/25b5c63905682a0e38d8ed86170d98be6ddb8cf8#diff-cd5ac2db946c26a2668c92b5804888e2e68724aaee6b1a86c255e2271ef3f835R9140-R9141
I think it’s already corrected when the `if`’s condition switched from a bitfield.

knielsen, line 9239: Haha, yeah, that's a lot better.

I notice you seem to spend effort on the structure of the code and comments
and commit comments etc., and this is something I really appreciate. Here,
and elsewhere. In reviews, we usually mention the things that we want
change, but not all the changes that are great. So thanks for the efforts to
write patch series; and to improve the replication code, which really needs
it after 25+ years hacking.

knielsen (approved):
Thanks Jimmy, looks good, just a few naming inconsistencies in comments:

> +/** @return whether the replica's Master_info is marked as killed */

> +/**
> +  @brief Re/connect to the primary
> +  @details
> +    After preparations and config validations,
> +    this repeatedly calls the low-level connection function (and logs statuses)
> +    up to mi->retry_count times until success or when the replica's killed.

> +--source include/rpl_stop_server.inc # shut the primary down

In the source code (here comments), please consistently use the terminology
"master" and "slave" to remain consistent with existing code. Thanks for
mostly doing that already BTW, just these few instances leftover (2 times
"replica" and 2 times "primary")..


A general thing, though a bit out of scope of this particular pull request
perhaps, is that when you force-pushed after my first review, Github
completely drops the original patch series :-(

That makes it very hard to do second review, how to see what's changed
without having the origical patch series (for use with eg. git-range-diff)?

For simple patches just avoiding force-push (until final merge) can work,
but here you've tried to maintain a well-structured patch series, which I
really appreciate. I just don't see how to do this properly with Github pull
requests, which seem to work well only for trivial patches :-/

One option is to use git-format-patch and git-send-email to the commits@
mailing list which preserves the history of original patch series, another
is to push the original reviewed series to a separate branch and mention it
in a comment on the PR before force-pushing after-review changes. Or is
there a better way to do this in the pull-request feature itself on Github?

Anyway, this pull request is fine, I managed to review without refering to
the original patch series, just something we need to find a solution to
going forward if pull requests are to be used more...

sql/sql_show.cc
ParadoxV5: > […] when you force-pushed after my first review, Github completely drops the original patch series :-(
> 
> That makes it very hard to do second review, how to see what's changed without having the origical patch series (for use with eg. git-range-diff)?
> 
> For simple patches just avoiding force-push (until final merge) can work, but here you've tried to maintain a well-structured patch series, which I really appreciate. I just don't see how to do this properly with Github pull requests, which seem to work well only for trivial patches :-/
> 
> […]
> 
> Anyway, this pull request is fine, I managed to review without refering to the original patch series, just something we need to find a solution to going forward if pull requests are to be used more...

I experienced this too, which is why (if I remember to) I summarize force-pushes in PR comments unless the rewrite is well beyond trivial.
GitHub logs “`author` force-pushed the `branch` branch from `hash before` to `hash after`” updates in the PR, which does diff amendments but rebases dirties the diffs.

> One option is to use git-format-patch and git-send-email to the commits@ mailing list which preserves the history of original patch series, another is to push the original reviewed series to a separate branch and mention it in a comment on the PR before force-pushing after-review changes.

Re. rebases, would this require 4-way diff, `old.base..old.HEAD – new.base..new.HEAD`?

> Or is there a better way to do this in the pull-request feature itself on Github?

I suppose we could avoid rebases until final merge or conflict resolution.
It’s a little inconvenient, but we could use `refs/pull/3764/merge` to test the merge code (which is what CIs/Buildbot uses).

It’s a quirk in this organization that we rebase everything except for merging bugfixes from older versions to newer.
Some people in other open-source places would “rebase” their feature branches (e.g., `mdev-35304`) by merging the base branch (`main`) into them. It’s even what GitHub’s “update branch to pull changes in main)” button does.
(It’s also a quirk that we merge-up bugfixes instead of backport, but this one’s very smart 💡.)

knielsen: The 4-way diff is `git range-diff`, very useful to compare different versions of patch series.

mysql-test/suite/rpl/t/rpl_master_retry_count_basic.test
```diff
@@ -0,0 +1,78 @@
+# MDEV-25674: Test the `Master_Retry_Count` field of
+# CHANGE MASTER TO and SHOW SLAVE STATUS & co. (no feature testing)
```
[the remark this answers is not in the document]
ParadoxV5: I thought CHANGE MASTER TO and SHOW SLAVE STATUS & co. are part of Replication?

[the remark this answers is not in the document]
bnestere: Yeah, there is a bit of a funky delineation (which I don't even know that I agree with) where tests that are limited to testing option values (i.e. testing variables/parsing, as this test is) are put in `main` (provided they don't use `master-slave.inc` or `rpl_init.inc`).

I suppose one advantage of this, is if one uses the `mtr` option `--skip-rpl` (usually done locally to save time for changes that aren't replication related), there are still assurances for connections between options/variables.

Though @andrelkin or @knielsen might be able to explain that delineation better, having more history with it. I can't actually find any documentation that explicitly states the purpose of the `main` suite. So my understanding might be wrong too :shrug:.

mysql-test/suite/multi_source/connects_tried.test
```diff
@@ -0,0 +1,133 @@
...
+
+--connection slave
+CHANGE MASTER TO master_connect_retry=1;
+--replace_result $SERVER_MYPORT_3 SERVER_MYPORT_3
+--eval CHANGE MASTER 'named' TO master_host='127.0.0.1', master_port=$SERVER_MYPORT_3, master_user='root', master_ssl_verify_server_cert=0, master_connect_retry=2
+--echo # `Connects_Tried` is 0 before connections begin.
+SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
+
+--disable_warnings
+START ALL SLAVES; # will fail because the master's down
```
ParadoxV5, line 27:
```suggestion
START ALL SLAVES; # will fail because the masters are down
```

mysql-test/suite/rpl/t/rpl_connects_tried.test
ParadoxV5: solved with https://github.com/MariaDB/server/pull/3764#discussion_r1957387809

ParadoxV5 (commented):
Hey @bnestere,
Thanks for the feedback; I’ve [updated](https://github.com/MariaDB/server/compare/030ff0c2ff..d5c7ba2d47) the tests re. your suggestions, please take a quick look.

I’ve somewhat rewritten `rpl.rpl_connects_tried` to cover https://github.com/MariaDB/server/pull/3764#discussion_r1957387809 (and https://github.com/MariaDB/server/pull/3764#discussion_r1957387926 and https://github.com/MariaDB/server/pull/3764#discussion_r1953771114 too, while there).
I’ve also migrated it to use proper multi-source testing and moved it to `multi_source.connects_tried`.

mysql-test/main/master_retry_count_basic.test
```diff
@@ -0,0 +1,80 @@
...
+SELECT Connection_name, Master_Retry_Count
+  FROM information_schema.SLAVE_STATUS;
+
+--echo # Caps values (such as UINT64_MAX + 1) to `--master-retry-count`'s max
+CHANGE MASTER TO master_retry_count=18446744073709551616;
+CHANGE MASTER 'named' TO master_retry_count=18446744073709551616;
+SELECT Connection_name
+  FROM information_schema.SLAVE_STATUS
+  WHERE Master_Retry_Count IN (2147483647, 18446744073709551615);
+
```
ParadoxV5, line 67: This fails on [amd64-windows](https://buildbot.mariadb.org/#/builders/234/builds/36722)… It could very well be that we never tested edge-case behaviors[^MDEV-35948] at all 😬.
```suggestion
--echo # Caps values (such as UINT64_MAX + 1) to `--master-retry-count`'s max
CHANGE MASTER TO master_retry_count=18446744073709551616;
CHANGE MASTER 'named' TO master_retry_count=18446744073709551616;
SELECT Connection_name
  FROM information_schema.SLAVE_STATUS
  WHERE Master_Retry_Count IN (2147483647, 18446744073709551615);

```

[^MDEV-35948]: such as [MDEV-35948 for SHOW REPLICA STATUS](https://jira.mariadb.org/browse/MDEV-35948)

sql/rpl_mi.cc
ParadoxV5: recorded at [MDEV-36141](https://jira.mariadb.org/browse/MDEV-36141) for assessment

mysql-test/suite/multi_source/connects_tried.test
```diff
@@ -0,0 +1,133 @@
...
+# separate counters in parallel multi-source (SHOW SLAVE [name] STATUS).
+#
+# Note: nearly all SELECT results are timing-sensitive,
+# they therefore list unexpected rows because.
+
+--source include/have_binlog_format_mixed.inc # The test is agnostic of binlog formats.
+# `rpl_*.inc` (still) expects `server_1` be `master` and `server_2` be `slave`.
+--let $rpl_server_count= 3
+--let $rpl_skip_start_slave= 1
+--source include/master-slave.inc
```
bnestere, line 13: Conventionally, our tests in `multi_source` don't actually use `master-slave.inc` or `rpl_init.inc`, but rather create the connections manually so each channel/source is clear in its use in the test.  

Note too (for the `rpl` suite) when creating tests cases with topologies beyond the simple master-slave, you can use `rpl_init.inc` (instead of `master-slave.inc`) with the variable `rpl_topology=1->3,2->3`, etc, and it will set everything up for you.

bnestere (approved):
Thanks @ParadoxV5 for a great patch! I only had the one last note about `master-slave.inc`.

mysql-test/suite/multi_source/connects_tried.test
ParadoxV5: My test used the convenience include `rpl_start_server.inc`, which `--source`s `rpl_reconnect.inc`, whose code is designed to explicitly (and consequentially *only*) work with  `master-slave.inc` 😕.
I’m not certain how to lift that limitation (#3849), so I decided to stick to the `rpl` format (which isn’t the `multi_source` format ¯\\\_(ツ)_/¯) over writing the de/reconnection commands out.
I saw that our test engineer’s got plans on generalizing setups for QA, which could alleviate.

> `rpl_topology=1->3,2->3`

I thought `rpl_init.inc` only executes unnamed `CHANGE MASTER TO`, which `2->3` would overwrite `1->3` (overwrite `server_3`’s primary from port 1 to port 2).

mysql-test/suite/multi_source/connects_tried.test
```diff
@@ -0,0 +1,133 @@
...
+  FROM status_sleep JOIN status_end USING(Connection_name)
+  WHERE status_end.Connects_Tried <= status_sleep.Connects_Tried;
+
+DO SLEEP(2);
+--echo # `Connects_Tried` does not increment after connection establishes.
+CREATE TEMPORARY TABLE status_after AS
+  SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
+SELECT *
+  FROM status_end JOIN status_after USING(Connection_name)
+  WHERE status_after.Connects_Tried <> status_end.Connects_Tried;
```
mariadb-SusilBehera, line 81: Should the comparison be based on > instead of <>?

mariadb-SusilBehera (commented):
Please see if my review comments are good enough to incorporate.

mysql-test/suite/multi_source/connects_tried.test
ParadoxV5: Shouldn’t matter ¯\\\_(ツ)\_/¯ `<>` would also complain if the counter somehow decrements.

mysql-test/suite/multi_source/connects_tried.test
```diff
@@ -0,0 +1,133 @@
...
+--source include/wait_for_slave_to_stop.inc
+RESET SLAVE;
+--echo # RESET SLAVE resets `Connects_Tried` (for the resetted connection only).
+CREATE TEMPORARY TABLE status_reset AS
+  SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
+SELECT *
+  FROM status_reset JOIN status_restart USING(Connection_name)
+  WHERE status_reset.Connects_Tried <>
+    IF(LENGTH(Connection_name), status_restart.Connects_Tried, 0);
+
```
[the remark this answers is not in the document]
ParadoxV5: This feature is for error handing of _IO connection failures_.

> 1. with MASTER_DELAY set.
> 3. Parallel replication (not sure if this is needed as this patch is for IO_THREAD only)

I don’t think it has much to do with event execution of SQL thread(s).

> 2. Chain replication topology 1->2>3

What could this be, though…
`2` fails to connect to `1` as requested by `3`, and we examine how `3`’s counter correspond to `2`’s?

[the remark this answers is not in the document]
mariadb-SusilBehera: I think 3 can only request 2 in topology 1->2>3.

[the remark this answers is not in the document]
ParadoxV5: In that case, 2➡️3 shouldn’t have any difference from 1➡️2.

mariadb-SusilBehera (approved):
Accompanied tests look good. MASTER_DELAY and topology 1->2>3 can be covered at QA stage.

mysql-test/main/master_retry_count_basic.test
ParadoxV5:
```suggestion
--echo # Caps values (such as UINT64_MAX + 1) to `--master-retry-count`'s max
CHANGE MASTER TO master_retry_count=18446744073709551616;
CHANGE MASTER 'named' TO master_retry_count=18446744073709551616;
SELECT Connection_name
  FROM information_schema.SLAVE_STATUS
  WHERE Master_Retry_Count IN (4294967295, 18446744073709551615);

```

ParadoxV5: If this doesn’t fix the Windows difference then I’m throwing this UB case outta here.

ParadoxV5:
Amended a couple of cosmetic-sized changes, namely https://github.com/MariaDB/server/pull/3764#discussion_r1972648911.

mysql-test/main/master_retry_count_basic.test
bnestere: Oh interesting.. do you have a Windows box to debug and see what is going on? If not, mine is packed up atm, but I can check it out in a couple weeks. Maybe in the mean-time, just comment it out as a TODO for MDEV-35948?

ParadoxV5: Well I do but not really. My computer is Windows but I primarily almost always code in its WSL environment. On native Windows’s side I downloaded an MinGW but I’m not building MariaDB with it.

ParadoxV5:
By “all requirements are met” does GitHub really require the branch be up-to-date with the base?

sql/sql_show.cc
```diff
@@ -10708,6 +10702,8 @@ ST_FIELD_INFO slave_status_info[]=
   Column("Slave_Non_Transactional_Groups", ULonglong(20), NOT_NULL),
   Column("Slave_Transactional_Groups", ULonglong(20), NOT_NULL),
   Column("Replicate_Rewrite_DB", Varchar(), NOT_NULL),
+  Column("Connects_Tried", ULonglong(20), NOT_NULL),
```
andrelkin, line 10705: I missed to catch this hunk in time.
Why was it placed inside rather than be appended?
That just reproduces an old fable last time seen [here](https://github.com/MariaDB/server/commit/23c48474f7)
```
MDEV-33212: mysqldump uses MASTER_LOG_POS with dump-slave

The patch for MDEV-15530 incorrectly added a column in the middle of SHOW
SLAVE STATUS output. 
```

ParadoxV5: Kristian told me (https://github.com/MariaDB/server/pull/3764#pullrequestreview-2572711920) to bump the SHOW ALL SLAVES STATUS fields.

> No, I think instead SLAVE_STATUS_COL_RETRIED_TRANSACTIONS should be changed
> to 58 (or 57 until MDEV-25674), see below.
> > ……
> 
> No, I think this is wrong.
> 
> When new entries are added to the SHOW SLAVE STATUS output, they are added
> after all the existing columns output from SHOW SLAVE STATUS, but before
> the columns output by SHOW ALL SLAVES STATUS. Somewhat tricky, but that is
> why there is now an INFORMATION_SCHEMA view that allows to select just the
> columns of interest and not depend on column order.
> 
> So I think the patch should add the Connects_Tried column just after
> Replicate_Rewrite_DB (and MDEV-25674 can add Master_Max_Connect_Tries after
> that again). And then you do not need to change the if() condition or the
> DBUG_ASSERT above.

sql/slave.cc
ParadoxV5: → https://jira.mariadb.org/browse/MDEV-40743

PR #3767, opened 2025-01-15
MDEV-35854 dict_get_referenced_table() got too many parameters

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35854*
## Description
`dict_get_referenced_table()`: Make the callers responsible for converting the database and table names. This is a prerequisite for #3641,
## Release Notes
Nothing. This is supposed to be a non-functional change.
## How can this PR be tested?
Foreign key functionality is covered by `./mtr --suite=innodb`.
## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -12404,6 +12404,8 @@ create_table_info_t::create_foreign_keys()
 	const char*	      ref_column_names[MAX_COLS_PER_FK];
 	char		      create_name[MAX_DATABASE_NAME_LEN + 1 +
 					  MAX_TABLE_NAME_LEN + 1];
+	char db_name[MAX_DATABASE_NAME_LEN + 1];
+	char t_name[MAX_TABLE_NAME_LEN + 1];
 	dict_index_t*	      index	  = NULL;
 	fkerr_t		      index_error = FK_SUCCESS;
 	dict_index_t*	      err_index	  = NULL;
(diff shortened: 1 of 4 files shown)


storage/innobase/handler/handler0alter.cc
```diff
@@ -3295,14 +3298,51 @@ innobase_get_foreign_key_info(
...
+		uint errors;
+		LEX_CSTRING t;
+		t.str = t_name;
+		t.length = strconvert(cs, LEX_STRING_WITH_LEN(table_name),
+				      to_cs, t_name, MAX_TABLE_NAME_LEN,
+				      &errors);
+		LEX_CSTRING d = fk_key->ref_db;
+		if (!d.str) {
+			d.str = table->name.m_name;
+			d.length = table->name.dblen();
```
mariadb-DebarunBanerjee, line 3322: Comparing the code between dict_get_db_name_len() that was used earlier with table_name_t::db_len(), I find that the case where the separator is missing is handled better in the earlier version by returning the length as 0.
```
inline size_t dict_get_db_name_len(const char *name)
...
  if (const char* s= strchr(name, '/'))
    return size_t(s - name);
  return 0;

```
Where as in the dblen(), we have a debug assert and return an incorrect large number as length in release mode. If this ever happens, it could lead to severe memory corruption and I think some handling is necessary.
```
        /** @return the end of the schema name */
        const char* dbend() const
        {
                const char* sep = strchr(m_name, '/');
                ut_ad(sep);
                return sep;
        }
        /** @return the length of the schema name, in bytes */
        size_t dblen() const { return size_t(dbend() - m_name); }
```
Here is my suggestion.

1. Handle dbend() as nullptr case and return 0 as length. Keeping the debug assert is still fine so that we get the alert in debug mode.
2. If we don't want to go by previous suggestion and confident that it should never happen, then we need to convert the debug assert to release assert. I would rather go by [1] but if not then at least [2] seems less riskier than the way it is now.

storage/innobase/handler/handler0alter.cc
```diff
@@ -3295,14 +3298,51 @@ innobase_get_foreign_key_info(
...
+		CHARSET_INFO* to_cs = &my_charset_filename;
+
+		if (!strncmp(table_name.str, srv_mysql50_table_name_prefix,
+			     sizeof srv_mysql50_table_name_prefix - 1)) {
+			table_name.str
+				+= sizeof srv_mysql50_table_name_prefix - 1;
+			table_name.length
+				-= sizeof srv_mysql50_table_name_prefix - 1;
+			to_cs = system_charset_info;
+		}
```
mariadb-DebarunBanerjee, line 3311: I think it is worth retaining the comment from the deleted function dict_get_referenced_table() from which this code fragment comes.
```
/* This is a pre-5.1 table name
    containing chars other than [A-Za-z0-9].
    Discard the prefix and use raw UTF-8 encoding. */
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12404,66 +12430,93 @@ create_table_info_t::create_foreign_keys()
...
+
+		if (!strncmp(table_name.str, srv_mysql50_table_name_prefix,
+			     sizeof srv_mysql50_table_name_prefix - 1)) {
+			table_name.str
+				+= sizeof srv_mysql50_table_name_prefix - 1;
+			table_name.length
+				-= sizeof srv_mysql50_table_name_prefix - 1;
+			to_cs = system_charset_info;
+		}
+
```
mariadb-DebarunBanerjee, line 12459: I appreciate the refactoring of `dict_get_referenced_table()` and creating new functions like `dict_table_lookup()` and `copy_name()`. The implementation looks good and correct.

However, the rest of the code flow seems to be repeating in 3 places (with minor differences) where we used to call `dict_get_referenced_table()`. I think it is better to add the refactoring to `dict_get_referenced_table()` and still keep it to maintain the logic in one place.

storage/innobase/handler/handler0alter.cc
```diff
@@ -3295,14 +3298,51 @@ innobase_get_foreign_key_info(
...
+			d.length = table->name.dblen();
+		}
+
+		if (!strncmp(d.str, srv_mysql50_table_name_prefix,
+			     sizeof srv_mysql50_table_name_prefix - 1)) {
+			d.str += sizeof srv_mysql50_table_name_prefix - 1;
+			d.length -= sizeof srv_mysql50_table_name_prefix - 1;
+			to_cs = system_charset_info;
+		} else if (d.str == table->name.m_name) {
+			goto name_converted;
```
mariadb-DebarunBanerjee, line 3331: There is a change in the flow from the original code. From the base code, it should have been as follows.
```
if (!d.str) {
...
goto name_converted;
} else if (!strncmp(d.str, srv_mysql50_table_name_prefix,
..
else {
}
```
It looks like we are adding one more case for comparing `srv_mysql50_table_name_prefix` when `d.str = table->name.m_name` and then `strconvert()`. It is probably better to avoid such change in this patch unless we could test that specific functional part and establish the bug and then fix.

mariadb-DebarunBanerjee (commented):
@dr-m The refactoring looks good overall. Please see my doubts/comments.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -12610,32 +12663,68 @@ create_table_info_t::create_foreign_keys()
...
 
-			bufend = innobase_convert_name(
+			*innobase_convert_name(
 				buf, MAX_TABLE_NAME_LEN,
 				foreign->referenced_table_name,
-				strlen(foreign->referenced_table_name), m_thd);
-			buf[bufend - buf] = '\0';
-			key_text k(fk);
+				strlen(foreign->referenced_table_name), m_thd)
+				= '\0';
```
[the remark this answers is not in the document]
dr-m: Here the scope of each such code snippet is small, and we could indeed declare a local variable. If they were part of longer blocks, then declaring a variable could make it harder to read the code.

storage/innobase/handler/handler0alter.cc
dr-m: I’ll make `dbend()` and `basename()` tolerate a missing `/` in release builds.

storage/innobase/row/row0mysql.cc
```diff
@@ -2677,7 +2676,7 @@ row_rename_table_for_mysql(
 		goto rollback_and_exit;
 	}
 
-	if (!new_is_tmp) {
+	if (fk == RENAME_IGNORE_FK || !new_is_tmp) {
```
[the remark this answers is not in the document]
dr-m: In fact, I was paying attention to the same. In d2f079c21a1f2bdcaaddd48a111af69197db1434 this condition was changed to the following:
```c++
	if (fk == RENAME_IGNORE_FK || fk == RENAME_FK || !new_is_tmp) {
```
The additional constant `RENAME_FK` would be introduced in [MDEV-25292](https://jira.mariadb.org/browse/MDEV-25292), which would implement atomic `CREATE OR REPLACE TABLE`.

I’ll check the test coverage.

storage/innobase/handler/handler0alter.cc
dr-m: I’ll add a more verbose comment, mentioning `my_charset_filename` and `#mysql50#`.

storage/innobase/row/row0mysql.cc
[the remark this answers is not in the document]
dr-m: I’m now testing with an additional assertion
```c++
	ut_ad(fk != RENAME_IGNORE_FK || !new_is_tmp);
```
which expresses that `RENAME_IGNORE_FK` implies that the target name is user-visible. In other words, it tries to prove that the changed `if` condition does not make a difference. The stricter assertion
```c++
	ut_ad((fk == RENAME_IGNORE_FK) == !new_is_tmp);
```
would fail very easily during pretty much any `ALTER TABLE` statement.
I will add the first assertion and comment out the added condition in the `if` expression. In that way, we will not introduce any bloat to non-debug builds.

storage/innobase/handler/handler0alter.cc
dr-m: I’ll try to cover this change, by something like the following:
```sql
CREATE SCHEMA `#mysql50#d-b`;
CREATE TABLE `#mysql50#d-b`.parent …;
```

dr-m: I am able to cover the code with the following:
```sql
CREATE SCHEMA `#mysql50#d-b`;
CREATE TABLE `#mysql50#d-b`.t1 (a INT PRIMARY KEY) engine=InnoDB;
USE `#mysql50#d-b`;
CREATE TABLE t2 (a INT PRIMARY KEY) ENGINE=InnoDB;
SET STATEMENT foreign_key_checks=0 FOR
ALTER TABLE t2 ADD FOREIGN KEY (a) REFERENCES t1(a);
SHOW CREATE TABLE t2;
DROP DATABASE `#mysql50#d-b`;
```
I now realize that the `#mysql50#` prefix never can be present in InnoDB table names. If I add another `#mysql50#` prefix to the schema name so that the parent table name in InnoDB would be `#mysql50#d-b/t1`, then the `SHOW CREATE TABLE` output would incorrectly mention an invalid schema:
```
Table	Create Table
t2	CREATE TABLE `t2` (
  `a` int(11) NOT NULL,
  PRIMARY KEY (`a`),
  CONSTRAINT `t2_ibfk_1` FOREIGN KEY (`a`) REFERENCES `d-b`.`t1` (`a`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci
```
We can avoid this regression by exexcuting the `goto name_converted` earlier, like you suggested. Thank you!

mariadb-DebarunBanerjee (approved):
@dr-m Thank you for addressing the comments. The new `innodb_convert_name()` and `dict_scan_id()` related clean up looks good to me.

PR #3770, opened 2025-01-16
MDEV-35708 lock_rec_get_prev() returns only the first record lock

vlad-lesin:
It's supposed that the function gets the previous lock set on a record. But if there are several locks set on a record, it will return only the first one. Continue locks list iteration till the certain lock even if the certain bit in lock bitmap is set.

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/storage/innobase/lock/lock0lock.cc
+++ b/storage/innobase/lock/lock0lock.cc
@@ -984,13 +984,14 @@ lock_rec_get_prev(
   ut_ad(!in_lock->is_table());
   const page_id_t id{in_lock->un_member.rec_lock.page_id};
   hash_cell_t *cell= lock_sys.hash_get(in_lock->type_mode).cell_get(id.fold());
+  lock_t *prev_lock= nullptr;
 
   for (lock_t *lock= lock_sys_t::get_first(*cell, id); lock != in_lock;
        lock= lock_rec_get_next_on_page(lock))
     if (lock_rec_get_nth_bit(lock, heap_no))
-      return lock;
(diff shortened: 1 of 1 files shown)


dr-m (approved):
Thank you.

You claimed earlier that this regression was introduced in 21987e5919afba29a7a7a537d3567909c7b4b9ff. I agree, it looks like so. Please mention that in the commit message.

Unique-utkarsh (changes requested):
lock_rec_get_prev(
    lock_t *in_lock, 
    page_id_t heap_no
) {
    ut_ad(!in_lock->is_table());  // Ensure the lock is not a table lock.
    const page_id_t id{in_lock->un_member.rec_lock.page_id};  // Get page ID from record lock.
    hash_cell_t *cell = lock_sys.hash_get(in_lock->type_mode).cell_get(id.fold());  // Get the corresponding cell for the lock.
    lock_t *prev_lock = nullptr;  // Initialize the previous lock as nullptr.

    // Iterate through the locks on the record's page.
    for (lock_t *lock = lock_sys_t::get_first(*cell, id); lock != in_lock;
         lock = lock_rec_get_next_on_page(lock)) {
        // If the specific bit for the record is set, return the lock.
        if (lock_rec_get_nth_bit(lock, heap_no)) {
            return lock;
        }
        prev_lock = lock;  // Otherwise, store the current lock as the previous lock.
    }

    return prev_lock;  // Return the last lock found.
}

PR #3772, opened 2025-01-17
MDEV-35693: Improve SSS column sizes

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-35693][]*

## Description
Resize the type and width of SHOW REPLICA STATUS (technically `INFORMATION_SCHEMA.SLAVE_STATUS`) columns to better match their possible values

In case of intentionally but absurdly long lists, text columns that list an uncapped number of elements have expanded to accept as many bytes as we could support.

While here, I found all our integer columns are unsigned, so I updated ones that weren’t configured as unsigned before for both column types and the field storing function (grep `\(\*field\+\+\)->store\(.+(?<!false|true|my_charset_bin)\);` minus the one `(double)`).

In response to `‘MAX_SLAVE_ERRMSG’ was not declared in this scope` in Embedded builds, a new `#ifdef HAVE_REPLICATION` guard wraps `slave_status_info` to skip this unused data in Replication-less builds.

Particularly, the first-gen `Replicate_` filters were incorrectly typed as singlular `Name()`s during MDEV-33526.
Under `Name`s’ 64-char limit, they could overflow (read: truncate) even before their lengths got absurd.

Let me know if resizing I.S. column types is a significant backward-incompatibility.

## Release Notes
Increased VARCHAR width limits of `SHOW SLAVE STATUS` so it doesn’t truncate long lists that *should be long*.

### Optionally list any https://mariadb.com/kb/ pages that need changing.
By the way, these three columns are missing on https://mariadb.com/kb/en/show-replica-status/#column-descriptions:
```
Replicate_Do_Domain_Ids
Replicate_Ignore_Domain_Ids
Parallel_Mode
```

## How can this PR be tested?
~~:thinking: How shall we test this? Cook up a replica so its status report has all columns filled to the brim?~~
Discussion: https://github.com/MariaDB/server/pull/3772#discussion_r1930016568
This PR includes a cherry-pick of 3795 to test for the [MDEV-35693][] issue itself.

### Sample manual test from [MDEV-35693][]
Of course, the original issue about `Replicate_` fields specifically is reproducible by setting them to anything longer than 64 bytes.
```sql
set @@global.replicate_do_table='db_to_filter.0123456789_1,db_to_filter.0123456789_2,db_to_filter.0123456789_3,db_to_filter.0123456789_4,db_to_filter.0123456789_5,db_to_filter.0123456789_6,db_to_filter.0123456789_7';
 
show slave status\G
```
*Un*expected: `Replicate_Do_Table: db_to_filter.0123456789_7,db_to_filter.0123456789_1,db_to_filter`

## PR quality check

[MDEV-35693]: https://jira.mariadb.org/browse/MDEV-35693


sql/sql_show.cc
```diff
@@ -10626,66 +10623,66 @@ ST_FIELD_INFO slave_status_info[]=
...
-  Column("Gtid_IO_Pos", Varchar(1024), NOT_NULL),
-  Column("Replicate_Do_Domain_Ids", Varchar(FN_REFLEN), NOT_NULL),
-  Column("Replicate_Ignore_Domain_Ids", Varchar(FN_REFLEN), NOT_NULL),
-  Column("Parallel_Mode", Varchar(15), NOT_NULL),
+  Column("Using_Gtid", Varchar(sizeof("Current_Pos")-1), NULLABLE),
+  Column("Gtid_IO_Pos", Longtext(), NOT_NULL),
+  Column("Replicate_Do_Domain_Ids", Longtext(), NOT_NULL),
+  Column("Replicate_Ignore_Domain_Ids", Longtext(), NOT_NULL),
+  Column("Parallel_Mode", Varchar(sizeof("conservative")-1), NOT_NULL),
   Column("SQL_Delay", ULong(10), NOT_NULL),
```
ParadoxV5, line 10670: Hmm?
https://github.com/MariaDB/server/blob/74331a48dd3dadfe5041f3440b81a7183e9fd4a8/sql/slave.cc#L3106
Not only `Relay_log_info::get_sql_delay()` returns a signed `int`, but it’s also missing `, NOT_NULL`; this would invoke `store(double nr)` instead of `store(longlong nr, bool unsigned_val)`.

sql/sql_show.cc
```diff
@@ -10626,66 +10623,66 @@ ST_FIELD_INFO slave_status_info[]=
...
-  Column("Master_SSL_Verify_Server_Cert", Varchar(3), NOT_NULL),
+  Column("Seconds_Behind_Master", ULonglong(20), NULLABLE),
+  Column("Master_SSL_Verify_Server_Cert", Varchar(sizeof("Yes")-1), NOT_NULL),
   Column("Last_IO_Errno", SLong(4), NOT_NULL),
-  Column("Last_IO_Error", Varchar(MYSQL_ERRMSG_SIZE), NULLABLE),
+  Column("Last_IO_Error", Varchar(MAX_SLAVE_ERRMSG), NULLABLE),
   Column("Last_SQL_Errno", SLong(4), NOT_NULL),
-  Column("Last_SQL_Error", Varchar(MYSQL_ERRMSG_SIZE), NULLABLE),
-  Column("Replicate_Ignore_Server_Ids", Varchar(FN_REFLEN), NOT_NULL),
+  Column("Last_SQL_Error", Varchar(MAX_SLAVE_ERRMSG), NULLABLE),
```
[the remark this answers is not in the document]
ParadoxV5: Hmm. Embedded buildbots fail with `‘MAX_SLAVE_ERRMSG’ was not declared in this scope`.
* [x] `sql/rpl_reporting.h` `#define`s `MAX_SLAVE_ERRMSG`.
* [ ] `sql/rpl_mi.h` `#include`s `sql/rpl_reporting.h` **if `HAVE_REPLICATION`**.
* [x] `sql/sql_show.cc` `#include`s `sql/rpl_mi.h`.

mysql-test/suite/funcs_1/r/is_tables_is.result
ParadoxV5, line 1: > Variable-length types like VARCHAR can be used in MEMORY tables. BLOB or TEXT columns are not supported for MEMORY tables.
> ⸺ https://mariadb.com/kb/en/memory-storage-engine/

I don’t intend this…

sql/sql_show.cc
```diff
@@ -10626,66 +10623,66 @@ ST_FIELD_INFO slave_status_info[]=
...
   Column("Slave_DDL_Groups", ULonglong(20), NOT_NULL),
   Column("Slave_Non_Transactional_Groups", ULonglong(20), NOT_NULL),
   Column("Slave_Transactional_Groups", ULonglong(20), NOT_NULL),
-  Column("Replicate_Rewrite_DB",Varchar(1024), NOT_NULL),
+  Column("Replicate_Rewrite_DB",Longtext(), NOT_NULL),
   Column("Retried_transactions", ULong(10), NOT_NULL),
   Column("Max_relay_log_size", ULonglong(10), NOT_NULL),
   Column("Executed_log_entries", ULong(10), NOT_NULL),
   Column("Slave_received_heartbeats", ULong(10), NOT_NULL),
   Column("Slave_heartbeat_period", Float(703), NOT_NULL), // 3 decimals
```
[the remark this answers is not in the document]
ParadoxV5: https://jira.mariadb.org/browse/MDEV-35879

bnestere (changes requested):
Thanks @ParadoxV5 , you have some great finds here! See my notes.

Also, can you add a new test which fails before your patch to show the problems, which then succeeds with your patch? 

Then for the git commit message, for the title, `SSS` is only common for us casually speaking on the replication team. Can you spell it out?

sql/sql_show.cc
```diff
@@ -10626,66 +10623,66 @@ ST_FIELD_INFO slave_status_info[]=
...
-  Column("Gtid_Slave_Pos", Varchar(FN_REFLEN), NOT_NULL),
+  Column("Gtid_Slave_Pos", Longtext(), NOT_NULL),
   Column("Master_last_event_time", Datetime(0), NULLABLE),
   Column("Slave_last_event_time", Datetime(0), NULLABLE),
-  Column("Master_Slave_time_diff", SLonglong(10), NULLABLE),
+  Column("Master_Slave_time_diff", SLong(10), NULLABLE),
```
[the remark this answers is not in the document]
ParadoxV5: > (provided the timestamp is later than the current value)
👍
I’ll cast the subtraction operands to unsigned and add your description as a comment.
Perhaps a `DBUG_ASSERT` too?

ParadoxV5 (commented):
> Also, can you add a new test which fails before your patch to show the problems, which then succeeds with your patch?

I wondered at “§ How can this PR be tested?”, which fields should we test and how long the strings should we assert?

> Then for the git commit message, for the title, `SSS` is only common for us casually speaking on the replication team. Can you spell it out?

😄 When I picked up MDEV-35304, I too questioned what [its description](https://jira.mariadb.org/browse/MDEV-35304#descriptionmodule-label) meant by ‘SSS’.
I thought it was a common jargon!

As for the title… With ‘SHOW REPLICA STATUS’ it doesn’t fit [the 50-char guideline](https://github.com/MariaDB/server/blob/mariadb-11.8.0/CODING_STANDARDS.md), but I suppose it’s fine omitting `MDEV-35693: ` from the count.
([It’s time to avoid calling things ‘SLAVE’s when we can.](https://jira.mariadb.org/browse/MDEV-18777))

sql/sql_show.cc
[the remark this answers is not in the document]
ParadoxV5: 👌 probably IS-specific

sql/sql_i_s.h
```diff
@@ -142,8 +142,10 @@ class Varchar: public Type
 public:
   Varchar(uint length) :Type(&type_handler_varchar, length, false)
   {
+    // utf8mb3 Varchars longer than MAX_FIELD_VARCHARLENGTH/3 become Longtexts
     DBUG_ASSERT(length * 3 <= MAX_FIELD_VARCHARLENGTH);
   }
+  Varchar(): Type(&type_handler_varchar, MAX_FIELD_VARCHARLENGTH/3, false) {}
```
ParadoxV5, line 148: https://mariadb.com/kb/en/unicode/ says 11.8 wants to change `utf8` to `utf8mb4`.
Assuming the default charset is simply ‘utf8’, I’ll let that work change these `3`s to `4`s, in hopes that it notices it creates unexpected https://github.com/MariaDB/server/pull/3772#discussion_r1919479374 changes too.

bnestere:
> I wondered at “§ How can this PR be tested?”, which fields should we test and how long the strings should we assert?

[master_last_event_time_row.test](https://github.com/MariaDB/server/blob/main/mysql-test/suite/rpl/t/master_last_event_time_row.test) should provide a good example to use on how to query/assert the correctness of `SHOW REPLICA STATUS` columns. Then as far as how long the strings should be, with `64` being the old length where truncation would happen, maybe just double it? I don't necessarily think it needs to be `MAX_FIELD_VARCHARLENGTH` characters long.. But perhaps make it easy to change? E.g. a lot of our tests use `REPEAT('a', 128)` (or whatever string and number parameters to repeat); I'd suggest using that. And note you can parameterize the length with something like
```
--let $len= 64
--let $long_db_a= `SELECT REPEAT('a', $len)`
--let $long_db_b= `SELECT REPEAT('b', $len)`
```

And then you can use it in queries using `eval`:
```
--eval SET @@global.replicate_do_db="$long_db_a,$long_db_b"
```


Then to
> As for the title… With ‘SHOW REPLICA STATUS’ it doesn’t fit [the 50-char guideline](https://github.com/MariaDB/server/blob/mariadb-11.8.0/CODING_STANDARDS.md), but I suppose it’s fine omitting MDEV-35693:  from the count.

Hmm, I didn't realize we had a 50-char title maximum - I probably break that rule often :sweat_smile:. I'd say not to omit the MDEV-35693 part, that's the most important part. Maybe try to summarize as best you can within 50 chars, and if you can't, then just use the JIRA title and truncate it after 50.

mysql-test/suite/funcs_1/r/is_columns_is.result
```diff
@@ -418,24 +418,24 @@ def	information_schema	SLAVE_STATUS	Master_SSL_Key	34	NULL	YES	varchar	512	1536
...
+def	information_schema	SLAVE_STATUS	Replicate_Ignore_Domain_Ids	48	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
+def	information_schema	SLAVE_STATUS	Replicate_Ignore_Server_Ids	41	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
+def	information_schema	SLAVE_STATUS	Replicate_Ignore_Table	18	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
+def	information_schema	SLAVE_STATUS	Replicate_Rewrite_DB	56	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
+def	information_schema	SLAVE_STATUS	Replicate_Wild_Do_Table	19	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
+def	information_schema	SLAVE_STATUS	Replicate_Wild_Ignore_Table	20	NULL	NO	varchar	21844	65532	NULL	NULL	NULL	utf8mb3	utf8mb3_general_ci	varchar(21844)			select		NEVER	NULL	NO	NO
```
ParadoxV5, line 436: Actually @bnestere do these count as tests for this issue?

sql/sql_show.cc
ParadoxV5: added the `, true`

---

According to `rpl.rpl_delayed_slave`, `SQL_Delay`’s unsigned in practice.
So changing `int` to `unsigned int` is not a bug but a clarifying refactor. I won’t include it here.
Same for checking whether each of the `(uint32)` casts make sense. We could instead use more `Field::store()` overloads – for the various integer sizes as well as separate handlers for unsigned ones.

ParadoxV5:
> > > I wondered at “§ How can this PR be tested?”, which fields should we test and how long the strings should we assert?
> > 
> > master_last_event_time_row.test should provide a good example to use on how to query/assert the correctness of `SHOW REPLICA STATUS` columns. Then as far as how long the strings should be, with `64` being the old length where truncation would happen, maybe just double it? I don't necessarily think it needs to be `MAX_FIELD_VARCHARLENGTH` characters long.. […]
> 
> So, just testing those was-`Named()` OG-`Replicate_`s is fine, since they came directly from user configs and should never truncate?
> Hmm, then maybe the newer `Replicate_`s too.

Uhh… I don’t feel right testing only a subset. ~~Reflecting from the Broken Windows Theory,~~ we ~~either test none of them or we~~ __should__ test all of them.
Concerningly, looks like overflow test are done for the underlying variables but not for SSS reports.

mysql-test/suite/funcs_1/r/is_columns_is.result
bnestere: Sometimes they will, but in this case I'd say it isn't sufficient, as the existing tests weren't enough to catch the regression in the first place :sweat_smile:. So we should add tests to ensure validly-long values don't get truncated.

ParadoxV5: I reviewed the SHOW REPLICA STATUS listing and gave them a good consideration.
(Some like these `Replicate_`s come straight from user configs, others are computed live or are string messages (the L10N-able kind).)

I’ve decided that edge case testing should be part of their feature testing; that is, verify their SHOW REPLICA STATUS entry in addition to their `SELECT GLOBAL/SESSION`/`SHOW STATUS`/etc.
Which means they ideally should be available on the branch SHOW REPLICA STATUS added them; i.e., ideally backport to the earliest applicable branches.

The maintenance sprint is finishing up, @bnestere. Lemme see if tests for these `Replicate_`s ones can make it.

ParadoxV5: filed [MDEV-35948](https://jira.mariadb.org/browse/MDEV-35948) to generalize the scope to all of SHOW REPLICA STATUS

sql/sql_show.cc
```diff
@@ -10619,79 +10619,78 @@ ST_FIELD_INFO check_constraints_fields_info[]=
...
   Column("SQL_Remaining_Delay", ULong(10), NULLABLE),
-  Column("Slave_SQL_Running_State", Varchar(64), NULLABLE),
+  Column("Slave_SQL_Running_State", Varchar(), NULLABLE),
   Column("Slave_DDL_Groups", ULonglong(20), NOT_NULL),
   Column("Slave_Non_Transactional_Groups", ULonglong(20), NOT_NULL),
   Column("Slave_Transactional_Groups", ULonglong(20), NOT_NULL),
-  Column("Replicate_Rewrite_DB",Varchar(1024), NOT_NULL),
+  Column("Replicate_Rewrite_DB", Varchar(), NOT_NULL),
   Column("Retried_transactions", ULong(10), NOT_NULL),
   Column("Max_relay_log_size", ULonglong(10), NOT_NULL),
```
ParadoxV5, line 10682:
```suggestion
  Column("Max_relay_log_size", ULong(10), NOT_NULL),
```

mysql-test/suite/funcs_1/r/is_columns_is.result
bnestere: >  i.e., ideally backport to the earliest applicable branches.
...
> filed [MDEV-35948](https://jira.mariadb.org/browse/MDEV-35948) to generalize the scope to all of SHOW REPLICA STATUS

Sure. It makes sense to have the complete test in 10.5+ as well, though with the addition...

> Lemme see if tests for MDEV-35693’s Replicate_s ones can make it.

I think would be good. I'd say we should add tests now for any of the fields you've changed in this patch, and then backport it and test the rest in 10.5 (say during the next light week). And the cutoff for this part is Friday, so I'd think there's time (this would take priority over MDEV-35304).

sql/sql_show.cc
```diff
@@ -10626,66 +10623,66 @@ ST_FIELD_INFO slave_status_info[]=
...
   Column("Master_SSL_Crlpath", Varchar(FN_REFLEN), NULLABLE),
-  Column("Using_Gtid", Varchar(15), NULLABLE),
-  Column("Gtid_IO_Pos", Varchar(1024), NOT_NULL),
-  Column("Replicate_Do_Domain_Ids", Varchar(FN_REFLEN), NOT_NULL),
-  Column("Replicate_Ignore_Domain_Ids", Varchar(FN_REFLEN), NOT_NULL),
-  Column("Parallel_Mode", Varchar(15), NOT_NULL),
+  Column("Using_Gtid", Varchar(sizeof("Current_Pos")-1), NULLABLE),
+  Column("Gtid_IO_Pos", Longtext(), NOT_NULL),
+  Column("Replicate_Do_Domain_Ids", Longtext(), NOT_NULL),
+  Column("Replicate_Ignore_Domain_Ids", Longtext(), NOT_NULL),
```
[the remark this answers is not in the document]
ParadoxV5: 4096, the current width of `INFORMATION_SCHEMA.GLOBAL_VARIABLES.VARIABLE_VALUE`?
https://github.com/MariaDB/server/blob/74331a48dd3dadfe5041f3440b81a7183e9fd4a8/sql/sql_show.cc#L10275-L10280

mysql-test/suite/funcs_1/r/is_columns_is.result
ParadoxV5: I’ve prepared #3795 for the ones reported (and are mainly affected) by MDEV-35693.
I’ve yet to search where the rests’ (both the latter `Replicate_`s and others) tests are.

sql/sql_show.cc
```diff
@@ -10619,79 +10619,78 @@ ST_FIELD_INFO check_constraints_fields_info[]=
...
   Column("Relay_Master_Log_File", Varchar(FN_REFLEN), NOT_NULL),
-  Column("Slave_IO_Running", Varchar(10), NOT_NULL),
-  Column("Slave_SQL_Running", Varchar(3), NOT_NULL),
-  Column("Replicate_Do_DB", Name(), NOT_NULL),
-  Column("Replicate_Ignore_DB", Name(), NOT_NULL),
-  Column("Replicate_Do_Table", Name(), NOT_NULL),
-  Column("Replicate_Ignore_Table", Name(), NOT_NULL),
-  Column("Replicate_Wild_Do_Table", Name(), NOT_NULL),
-  Column("Replicate_Wild_Ignore_Table", Name(), NOT_NULL),
+  Column("Slave_IO_Running", Varchar(sizeof("Connecting")-1), NOT_NULL),
```
andrelkin, line 10637: I find this unnecessary. Both versions are not future-proof in that a new state with the name length greater than `sizeof("Connecting")-1` might be introduced.
A better approach to me would be to max out the allowed size with `Varchar()` or restore the OLD version plus add a compile-time assert to constrain the string lengths to 10 in [`slave_running[]`](https://github.com/MariaDB/server/blob/mdev-35693/sql/slave.cc#L2861).

I'd go with the simple `Varchar()`.

sql/sql_show.cc
```diff
@@ -10619,79 +10619,78 @@ ST_FIELD_INFO check_constraints_fields_info[]=
...
-  Column("Using_Gtid", Varchar(15), NULLABLE),
-  Column("Gtid_IO_Pos", Varchar(1024), NOT_NULL),
-  Column("Replicate_Do_Domain_Ids", Varchar(FN_REFLEN), NOT_NULL),
-  Column("Replicate_Ignore_Domain_Ids", Varchar(FN_REFLEN), NOT_NULL),
-  Column("Parallel_Mode", Varchar(15), NOT_NULL),
+  Column("Using_Gtid", Varchar(sizeof("Current_Pos")-1), NULLABLE),
+  Column("Gtid_IO_Pos", Varchar(), NOT_NULL),
+  Column("Replicate_Do_Domain_Ids", Varchar(), NOT_NULL),
+  Column("Replicate_Ignore_Domain_Ids", Varchar(), NOT_NULL),
+  Column("Parallel_Mode", Varchar(sizeof("conservative")-1), NOT_NULL),
```
andrelkin, line 10673: The same argument can be given here as in [above](https://github.com/MariaDB/server/pull/3772/commits/18f024e95d52f8ea1870ae685cc9be9067719d42#r1936331721).

sql/sql_show.cc
ParadoxV5: You’re right, both versions are future-proof only by assuming that a new enum is tested to not print truncated in SSS.

sql/sql_show.cc
```diff
@@ -10626,66 +10623,66 @@ ST_FIELD_INFO slave_status_info[]=
   Column("Slave_IO_State", Varchar(64), NULLABLE),
   Column("Master_Host", Varchar(HOSTNAME_LENGTH), NULLABLE),
   Column("Master_User", Varchar(USERNAME_LENGTH), NULLABLE),
-  Column("Master_Port", ULong(7), NOT_NULL),
+  Column("Master_Port", ULong(5), NOT_NULL),
```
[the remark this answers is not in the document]
ParadoxV5: > There’s no `UShort()` though…

I’ve added one for this and  the three `Last_Errno`s.

sql/sql_show.cc
ParadoxV5: > […] restore the OLD version plus add a compile-time assert to constrain the string lengths to 10 in `slave_running[]`.

I prefer this better, so these columns aren’t unnecessarily max-sized.
Thus I’ve reverted these changes (at least for now).
This is a 🐞, not a 🐛. Harmless 🐞s should be refactored, not forcibly fixed.

ParadoxV5:
#### Rebase + Amend + Cherry-Pick #3795
Amends include:
* https://github.com/MariaDB/server/pull/3772#discussion_r1930047281
* https://github.com/MariaDB/server/pull/3772#discussion_r1936590189
* https://github.com/MariaDB/server/pull/3772#discussion_r1936594714

mysql-test/suite/sys_vars/r/replicate_do_db_basic.result
ParadoxV5, line 1: This is new here: (from local run results)

> MTR's internal check of the test case 'sys_vars.replicate_do_db_basic' failed.
```diff
--- /home/parad/MariaDB/11.7/build/mysql-test/var/tmp/check-mysqld_1.result     2025-01-30 18:53:07.913563831 -0700
+++ /home/parad/MariaDB/11.7/build/mysql-test/var/tmp/check-mysqld_1.reject     2025-01-30 18:53:08.093563835 -0700
@@ -31,7 +31,7 @@
 Master_SSL_Cipher
 Master_SSL_Key
 Seconds_Behind_Master  NULL
-Master_SSL_Verify_Server_Cert  No
+Master_SSL_Verify_Server_Cert  Yes
 Last_IO_Errno  0
 Last_IO_Error
 Last_SQL_Errno 0
```

applies to all 6 of the cherry-pick

sql/slave.cc
```diff
@@ -3146,6 +3147,8 @@ void store_master_info(THD *thd, Master_info *mi, TABLE *table,
     if ((slave_timestamp= mi->rli.slave_timestamp))
     {
       (*field++)->store_timestamp((my_time_t) slave_timestamp, 0);
+      // primary's binlog timestamp should be >= replica's replication timestamp
+      DBUG_ASSERT(master_timestamp >= slave_timestamp);
```
bnestere, line 3151: Huh, now that you've added this, `rpl_gtid_crash` fails at this assertion:

```
mariadbd: /home/buildbot/amd64-ubuntu-2004-debug/build/sql/slave.cc:3151: void store_master_info(THD*, Master_info*, TABLE*, String*): Assertion `master_timestamp >= slave_timestamp' failed.
250131  2:28:43 [ERROR] /home/buildbot/amd64-ubuntu-2004-debug/build/sql/mariadbd got signal 6 ;
```

I'd say we should defer the `Seconds_Behind_Master` changes to a future MDEV (though I'd say to file a JIRA so we can investigate the failure, i.e. is it a bug or is it somehow valid), as we've never had any complaints about SBM cutting off (and the new fields from MDEV-33856 are more accurate anyway).

mysql-test/suite/sys_vars/r/replicate_do_db_basic.result
bnestere: Yeah, in `11.4` (MDEV-31857), option `--ssl-verify-server-cert` became enabled by default and in it and later versions, when you do `CHANGE MASTER TO`, you have to add the option `master_ssl_verify_server_cert=0` to the `CHANGE MASTER TO` command (as done in abcd23add20276e4996773f578e77d5733e1b582)

bnestere (approved):
Thanks for the updates, @ParadoxV5 !

After addressing my 2 standalone comments (sorry, I didn't make them part of this review chunk):
1. Adding `master_ssl_verify_server_cert=0` to `CHANGE MASTER TO` commands
2. Reverting the `Seconds_Behind_Master` changes (temporarily, where we can investigate more thoroughly later)

and my next point, I'd say this is good to go!

For the PR structure, I'd also say to squash the third commit with the tests into the second, so the tests go in with the fixes, for easy reference/potential future debugging.

Also a few quick remarks about the commit message:
1. I'd say to avoid phrases like "See that PR for details", and be up-front right away about any relevant details (so one doesn't have to go on a goose chase looking for why something was done some way, in a volatile way). So here, for example, I'd say something along the lines of "these tests are eventually to be back-ported into a GA release, with its progress tracked via MDEV-35948). Effectively, it is to clearly organize the work by keeping all the content of this patch within the context of MDEV-35693, and the future-backport and extensions to be done later.
2. I'd also say just to refer to JIRA tickets (i.e. to drop references to the PR in the commit message). The JIRA is supposed to be the source of truth of the ticket (which one can find the PR from). _Ideally too, we should take important decisions that happen when discussing the PR, and add them as comments to the JIRA_

andrelkin (approved):
Looks good to me. Thanks Jimmy and Brandon!

mysql-test/suite/sys_vars/r/replicate_do_db_basic.result
ParadoxV5: It’s quite silly that `CHANGE MASTER TO` doesn’t recognize `false` as a boolean alias of `0`.

ParadoxV5 (commented):
And thanks Brandon and Andrei for reviewing!

> I'd say to avoid phrases like "See that PR for details", and be up-front right away about any relevant details.

Alrighty, I now state it a forward port and include the entire relevant commit description.

> I'd also say just to refer to JIRA tickets (i.e. to drop references to the PR in the commit message).

I’d like MDEV-35948 to track the progress of all columns, not just #3795’s (at least current) coverage, which are those of MDEV-35693.
So mentioning the PR as well is more precise, albeit somewhat redundant.

PR #3773, opened 2025-01-17
MDEV-35723: applying non-zero offset to null pointer in INSERT

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35723*
## Description
`row_mysql_read_blob_ref()`: Correctly handle what `Field_blob::store()` generates for `length=0`.
## Release Notes
This probably does not need to be documented, because it is not known that this undefined behaviour would have caused any problem with any build.
## How can this PR be tested?
Compile with a suitable version of clang and `cmake -DWITH_UBSAN=ON`.
```sh
./mtr --rr innodb.innodb
rr replay var/log/mysql.1.rr/latest-trace
```
```gdb
break __ubsan::ScopedReport::~ScopedReport()
continue
backtrace
…
```
## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/innodb.result
+++ b/mysql-test/suite/innodb/r/innodb.result
@@ -3337,3 +3337,9 @@ Table	Op	Msg_type	Msg_text
 test.t1	check	status	OK
 ALTER TABLE t1 FORCE;
 DROP TABLE t1;
+#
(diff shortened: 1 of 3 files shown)


abarkov (approved):
The patch is OK. Thanks.

PR #3775, opened 2025-01-17
DRAFT: MDEV-34705: Storing binlog in InnoDB

knielsen:
Draft pull request for work-in-progress on the MDEV-34075 binlog-in-engine feature

A new option --binlog-storage-engine=ENGINE moves the binlog implementation into the storage engine, for supporting engines (currently only InnoDB).

InnoDB implements the binlog files as a new type of tablespace, and uses its redo log to make the binlog crash-safe without the overhead and complexity of two-phase commit.


bnestere:
Hi @knielsen! 

This is very cool to see. I haven’t gone through with a fine-toothed comb, but I’d say overall I have a pretty good understanding of the idea, and have some high-level questions to start (nothing code-related yet, as there are still many things in flux, and I feel any code comments I’d make would be superseded in time anyway)

A few questions on the subject of the out-of-band data to start:
1. By using the forest of perfectly balanced trees to index prior OOB writes, my understanding is that the main advantage we get here is faster reads (as opposed to just a stack, with each node referencing the last written node), which I would think would be most noticeable for semi-sync setups, once that is implemented. Is there more to be gained? (e.g., I don’t think it would reduce the memory footprint, as every node will be eventually referenced in-order to reconstruct the full event anyway).

2. I wonder if each binary log file should have a footer to summarize the file dependencies, e.g. if there are OOB transactions, what is the earliest binlog file needed to be able to reconstruct all transactions in this log.

    2a) Likewise, for a transaction with OOB data, I wonder if we should file the earlier binlog file needed to reconstruct that transaction

    2b) My thought is that this would help on reporting problems off-the-bat, e.g. mysqlbinlog could report missing binlog files immediately, instead of writing data up-to a transaction with missing OOB chunks.

    2c) My thought on a use-case for this is for really large transactions, and (when eventually supported) 2-phase XA 
transactions, where the PREPARE may happen long in the past.

3. (This is nothing to address now, but more to consider in the design to support more things in the long term): I don’t think the master-slave protocol supports this now, but perhaps we could parallelize OOB reads, where the dump-thread would be sending OOB chunks while another thread is buffering them. Just mentioning this for consideration in the current design to be extensible for such things.

4. I wonder if OOB binlogging should be optional. E.g., say some user’s workload is rollback heavy, that could result in binlogs bloated with garbage.


Then generally to the API, your API seems to be encapsulated from the existing MYSQL_BIN_LOG/Event_log/TC_LOG APIs (which makes sense, as yours is more of an engine-plugin, vs the legacy approach is the handler). Though a few thoughts come from this: 

5. Are we losing the ability for the binary log (stored in innodb) to function as a transaction coordinator? We’ve previously talked about that being the case, where multi-engine transactions aren’t yet supported. Do you have any new thoughts on how that would (eventually) be implemented?

6. Your binlog-in-engine API functions are hard-coded into the handler type. Could it instead be a separate class, perhaps used to initialize the handler? I wonder if that design would make the transaction coordinator piece easier to implement (e.g. perhaps via a shared in-engine binlog/context that multiple handlers use). 

7. Instead of having MYSQL_BIN_LOG use a lot of conditionals reliant around opt_binlog_engine_hton, I wonder if it can be refactored to prefer some composition-based strategy which is set-up at initialization time (e.g. just give it the right class instance).

knielsen:
Hi Brandon, thanks for taking a look at the patch and for initial
comments/questions!

> idea, and have some high-level questions to start (nothing
> code-related yet, as there are still many things in flux, and I feel
> any code comments I’d make would be superseded in time anyway)

Yes, agree.

> A few questions on the subject of the out-of-band data to start:

> 1. By using the forest of perfectly balanced trees to index prior OOB
> writes, my understanding is that the main advantage we get here is
> faster reads (as opposed to just a stack, with each node referencing
> the last written node), which I would think would be most noticeable
> for semi-sync setups, once that is implemented.

The OOB is needed to:

1. Preserve the strict append-only way of writing the binlog data (so no
back-patching to update forward pointers in previous nodes).

2. Avoid having to read all pages from the file backwards through the stack
one by one before we can even send the first byte to the slave, which could
be very slow for large event groups (I've heard of users with 50 GB event
groups .oO).

3. Have an algorithm that needs O(log(N)) in-memory space as opposed to
O(N), where N is the size of the event group, and does not need to read
every page twice.

So yes, basically faster reads, but also to be scalable to large event
groups (ie. gigabytes) which can easily happen for row-based.

> Is there more to be
> gained? (e.g., I don’t think it would reduce the memory footprint, as
> every node will be eventually referenced in-order to reconstruct the
> full event anyway).

The reduction in memory footprint is mainly to not have to keep a list of
_all_ nodes in memory (O(N)) while traversing the stack, instead needing
only to keep the path from root to current leaf (O(log(N))). Replication in
general tries hard to not restrict the size of event groups/transactions to
available memory. An additional goal is to reduce I/O, including having to
read the file backwards which I fear could be inefficient on some systems.

> 2. I wonder if each binary log file should have a footer to summarize
> the file dependencies, e.g. if there are OOB transactions, what is the
> earliest binlog file needed to be able to reconstruct all transactions
> in this log.

Yes, this is the plan. I think it can go in the file header (as opposed to
a footer), but this is still ToDo: to implement, we will see what is needed.

>     2a) Likewise, for a transaction with OOB data, I wonder if we
> should file the earlier binlog file needed to reconstruct that
> transaction

Right, I think that is already there, if I understood you correctly?

In-memory we have handler_binlog_event_group_info::engine_ptr where we have
binlog_oob_context::first_node_file_no.

An on-disk, the commit record for the transaction has a pointer to the first
OOB node.

>     2b) My thought is that this would help on reporting problems
> off-the-bat, e.g. mysqlbinlog could report missing binlog files
> immediately, instead of writing data up-to a transaction with missing
> OOB chunks.

Not sure this is possible in the general case.

If we start reading from binlog 10, and its header says it references back
to binlog 8, and binlog 8 (and 9) are available, then yes, we can be sure we
have everything we will need.

But the converse is not true. Even if binlog 8 is missing we can not be sure
we will need it just from reading the file 10 header. We might be starting
somewhere in the middle of file 10, after any references back to 8. Or we
might be starting from a multi-domain GTID position where each domain starts
at different positions in binlog 10. Or the OOB data in binlog 8 belongs to
a transaction that ended up rolling back and so will never be needed.

>     2c) My thought on a use-case for this is for really large
> transactions, and (when eventually supported) 2-phase XA
> transactions, where the PREPARE may happen long in the past.

Yes.

A future use case could be the ability to start optimistically replicating
large transactions on the slave before it has committed on the master.

> 3. (This is nothing to address now, but more to consider in the design
> to support more things in the long term): I don’t think the
> master-slave protocol supports this now, but perhaps we could
> parallelize OOB reads, where the dump-thread would be sending OOB
> chunks while another thread is buffering them. Just mentioning this
> for consideration in the current design to be extensible for such
> things.

Right. For the current master-slave protocol, we need to send the whole
event group as one consecutive stream of bytes, which is why we have the
forest-of-trees to efficiently do that. This could be very interesting to
think about extending to sending OOB data early to the slave (if that is
what you emant with "another thread buffering them"?). I agree it would be
very good to have more ideas for this to make the initial design suitably
extensible. 

> 4. I wonder if OOB binlogging should be optional. E.g., say some
> user’s workload is rollback heavy, that could result in binlogs
> bloated with garbage.

Yes, I thought about that too, it should be easy to implement. Mostly just
omit the code that spills transaction cache data directly to the engine
instead of to a tmpfile, eg. skip this:

    trx_cache.cache_log.write_function= binlog_spill_to_engine;

InnoDB has a limit of 1MB (or something like that) on mini-transactions. So
event groups larger than that still need to be split into multiple records.
But we can optionally keep all data in the transaction cache tmpfile until
commit, then hold LOCK_log and write all the pieces consecutively to the
binlog, so there is no more interleaving of oob data. (And then we can use a
normal commit record without any OOB nodes).

> Then generally to the API, your API seems to be encapsulated from the
> existing MYSQL_BIN_LOG/Event_log/TC_LOG APIs (which makes sense, as
> yours is more of an engine-plugin, vs the legacy approach is the
> handler). Though a few thoughts come from this:

Right.

My idea is to focus on the first release of this (already huge) project to
get a stable functionality on all the user-facing parts. Thus, if possible,
I would keep the existing logic on the server level during commit (LOCK_log,
queue_for_group_commit(), ...). Then in the next version we can focus on
improving scalability even more by avoiding a lot of this machinery in the
common case where there is no multi-engine transaction.

But details of this is still ToDo, might be more changes needed to get a
functional first release.x

> 5. Are we losing the ability for the binary log (stored in innodb) to
> function as a transaction coordinator? We’ve previously talked about
> that being the case, where multi-engine transactions aren’t yet
> supported. Do you have any new thoughts on how that would (eventually)
> be implemented?

I think the new binlog will work more or less the same way as the existing
one wrt. being the transaction coordinator for multi-engine transactions.
Thus, in the old binlog:

  RocksDB.prepare() ; InnoDB.prepare(); binlog.write(); RocksDB::commit(); InnoDB.commit();

In the new binlog, the InnoDB prepare() are omitted:

  RocksDB.prepare() ; InnoDB.binlog_and_commit(); RocksDB::commit()

The XA crash recovery algorithm then needs to be able to read from the new
binlog implementation to find RocksDB prepared transactions and see if they
were binlogged or not. We still have the XID events needed for this. I would
avoid the complexity of the binlog checkpoints for the new binlog case, and
just require the extra engines to fsync() again in their commit, for
simplicity.

The challenge is mostly if we have any functioning RocksDB or other
transactional engine to test against? My (limited) understanding is that
there are not really any resources to maintain RocksDB currently. So it
might make sense to just disable multi-engine transactions in the first
version, to focus on getting the more important use-cases stable first (and
not accidentally turn MDEV-34705 into a fix-RocksDB-XA-bugs project...).

> 6. Your binlog-in-engine API functions are hard-coded into the handler
> type. Could it instead be a separate class, perhaps used to initialize
> the handler? I wonder if that design would make the transaction
> coordinator piece easier to implement (e.g. perhaps via a shared
> in-engine binlog/context that multiple handlers use).

Yes, at least I do foresee changes to precisely how the extensions to the
storage engine API will look, even if the underlying functionality would
stay much the same. I deliberately tried not to spend too much time on this
so far, as I want to solicit input for any preference on how the storage API
should look. Eg. Serg traditionally was much involved in storage API
extensions, Sergei Petrunia also knows a lot about it from RocksDB work.

Adding handlerton methods is the traditional way, though I did add the class
handler_binlog_reader() as some small step towards a more C++ interface.
Note that traditionally, storage engines could be written in C (InnoDB was
pure C for many years, except handler/ha_innodb.cc perhaps). I am not sure
what the preferred style is nowadays, I am open to using any preferred
style, or to some new better style if there are no existing preferences.

> 7. Instead of having MYSQL_BIN_LOG use a lot of conditionals reliant
> around opt_binlog_engine_hton, I wonder if it can be refactored to
> prefer some composition-based strategy which is set-up at
> initialization time (e.g. just give it the right class instance).

Definitely, this needs to be improved in the current patch.

My approach is to start with this mess-of-conditionals to get an overview of
all the parts of the server-layer code that are affected. But then something
needs to be done to get a cleaner interface, using composition and/or whatever
is needed, still a ToDo.

This might also be extended in a follow-up patch to do more general
refactor/cleanup of the interfaces around binlog, which has become quite
convoluted and hard to work with. I know Monty has ideas for splitting the
relay log and the binlog into separate classes, to similarly avoid many of
the is_relay_log() conditionals.

Again, thanks for comments, and happy to answer any follow-up questions as
always.

 - Kristian.

janlindstrom:
@knielsen Before I can do any meaningful analysis of this implementation, I think higher level intro would be nice. Especially transactional behavior is interesting to me.

* Currently, there is 2PC between binlog, innodb and wsrep. Is this somehow going to change?
* There is GTID  and there is XID where wsrep overwriting XID with wsrep XID during 2PC prepare, this most likely is going to change, but do we really need all these?
* Normally something is written to binlog at commit as transaction is replicated on commit, but streaming replication is sending fragments of transactions even before commit, is this still possible?
* wsrep writes its XID on InnoDB rollback segments, can we get one single XYZ that would identify transaction GTID, XID, node UUID etc ?

knielsen:
Jan Lindström ***@***.***> writes:

> @knielsen Before I can do any meaningful analysis of this
> implementation, I think higher level intro would be nice. Especially
> transactional behavior is interesting to me.
>
> * Currently, there is 2PC between binlog, innodb and wsrep. Is this somehow going to change?

Initially, I think nothing is going to change for Galera/wsrep. Galera will
continue to use the existing binlog implementation, and will not support
--binlog-storage-engine=innodb.

Longer-term, I do believe it could be beneficial for Galera to support using
the new binlog format. Having the binlog write part integrated with the
InnoDB should both improve performance and reduce complexity, just as it
does for other binlog-related code.

The binlog events in event groups (ie. write-sets) remain unchanged with
this implementation. Is it so that Galera can run without binlog enabled at
all? In that mode, the old or new binlog format might make very little
change at all, if any.

To eventually enable Galera to use the new binlog format when it is running
with binlog enabled will require that Galera is properly integrated into the
MariaDB server layer wrt. binlog, group commit, GTID allocation, and
transaction coordinator. As has been discussed before, Galera needs to take
over the role of transaction coordinator, implement the TC_LOG interface and
take over the role of binlogging that is currently done by the MYSQL_BIN_LOG
subclass of TC_LOG. We do not want to carry over the current endless
problems with Galera inconsistency with binlog GTIDs, nor do we want the
mess of tons of #ifdef WITH_WSREP / IF_WSREP() in the new binlog code.

> * There is GTID and there is XID where wsrep overwriting XID with
> wsrep XID during 2PC prepare, this most likely is going to change, but
> do we really need all these?

I am not familiar with with "wsrep overwriting XID with wsrep XID". But this
is not going to change for Galera initially, the existing binlog will
remain, the new binlog implementation will be optional.

> * Normally something is written to binlog at commit as transaction is
> replicated on commit, but streaming replication is sending fragments
> of transactions even before commit, is this still possible?

I am not familiar with what streaming replication is (it sounds like it is
Galera sending partial write-sets around the cluster during transaction
execution to reduce stall around commit).
There is something similar in the new binlog implementation, where large
transactions are written in smaller pieces interleaved with other
transactions during execution and before commit. This is referred to as
"out-of-band" (OOB) binlogging.

> * wsrep writes its XID on InnoDB rollback segments, can we get one
> single XYZ that would identify transaction GTID, XID, node UUID etc ?

GTID will remain, it is the very core of MariaDB replication and binlogging.
I do not think XID is needed in the new binlog in the normal case where only
InnoDB is used in a transaction, but it will still be needed for
multi-engine transactions (eg. InnoDB+RocksDB).

If Galera would also use GTID as the single identifier, by implementing
TC_LOG as mentioned above, it should be beneficial. But this is optional
whether Galera devs want to spend effort on this. It will be a big change I
suppose and probably diverge from how Galera works in MySQL codebase.

 - Kristian.

sql/log.cc
```diff
@@ -12625,6 +13623,11 @@ binlog_checksum_update(MYSQL_THD thd, struct st_mysql_sys_var *var,
   ulong UNINIT_VAR(prev_binlog_id);
 
   mysql_mutex_unlock(&LOCK_global_system_variables);
+  if (opt_binlog_engine_hton && value)
+  {
+    sql_print_information("Value of binlog_checksum forced to NONE since binlog_storage_engine is enabled, and InnoDB uses its own superior checksumming of pages");
+    value= 0;
+  }
```
dr-m, line 14489: I’m not sure it is helpful to mention InnoDB here. It would make sense to use the same binlog block format, no matter which storage engine is taking care of the durability (that is, the write-ahead logging).

knielsen:
Marko Mäkelä ***@***.***> writes:

> @dr-m commented on this pull request.

>> +  if (opt_binlog_engine_hton && value)
> +  {
> + sql_print_information("Value of binlog_checksum forced to NONE since
> binlog_storage_engine is enabled, and InnoDB uses its own superior
> checksumming of pages");
> +    value= 0;
> +  }
>
> I’m not sure it is helpful to mention InnoDB here. It would make sense
> to use the same binlog block format, no matter which storage engine is
> taking care of the durability (that is, the write-ahead logging).

Agree, this is on the server layer, and could in the future be using a
different storage engine binlog implementation, I'll reword in a generic
way.

 - Kristian.

storage/innobase/include/fsp_binlog.h
```diff
@@ -464,11 +471,13 @@ extern ulong ibb_page_size;
...
 extern mysql_mutex_t active_binlog_mutex;
 extern pthread_cond_t active_binlog_cond;
+extern mysql_mutex_t binlog_durable_mutex;
+extern mysql_cond_t binlog_durable_cond;
 extern std::atomic<uint64_t> active_binlog_file_no;
 extern uint64_t first_open_binlog_file_no;
 extern uint64_t last_created_binlog_file_no;
-extern std::atomic<uint64_t> binlog_cur_written_offset[2];
-extern std::atomic<uint64_t> binlog_cur_end_offset[2];
+extern std::atomic<uint64_t> binlog_cur_durable_offset[4];
```
bnestere, line 513: @knielsen just skimming some of your latest commits, not a formal review, but one thought here would be to have sub-classes for a durable vs non-durable reader, instead of the same reader which is always set up for durable & non-durable reads with a flag in the constructor that tells it which mode to use.

extra/mariabackup/xtrabackup.cc
```diff
@@ -2591,6 +2602,10 @@ static bool innodb_init_param()
 		srv_undo_dir = (char*) ".";
 	}
 
+	if (!opt_binlog_directory || !xtrabackup_backup) {
+		opt_binlog_directory = (char *) ".";
+	}
```
dr-m, line 2637: I see that this is duplicating some old sins. But, would it be possible to declare `opt_binlog_directory` as `const char*` so that we can avoid the cast here?

storage/innobase/CMakeLists.txt
```diff
@@ -89,7 +89,9 @@ ENDIF()
 IF (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR
     CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wconversion -Wno-sign-conversion")
-  SET_SOURCE_FILES_PROPERTIES(fts/fts0pars.cc
+  # Source file innodb_binlog.cc needs to include server layer stuff that is
+  # not -Wconversion clean.
+  SET_SOURCE_FILES_PROPERTIES(fts/fts0pars.cc handler/innodb_binlog.cc
     PROPERTIES COMPILE_FLAGS -Wno-conversion)
```
dr-m: Are there any problems that would go beyond #4251? I fixed that in order to have a clean #4252.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2035,6 +2037,47 @@ static bool log_checkpoint_low(lsn_t oldest_lsn, lsn_t end_lsn) noexcept
...
+void mtr_t::write_binlog(bool space_id, uint32_t page_no,
+                         uint16_t offset,
+                         const void *buf, size_t size) noexcept
+{
+  ut_ad(!srv_read_only_mode);
+  ut_ad(m_log_mode == MTR_LOG_ALL);
+
+  bool alloc{size < mtr_buf_t::MAX_DATA_SIZE - (1 + 3 + 3 + 5 + 5)};
+  byte *end= log_write<WRITE>(page_id_t{LOG_BINLOG_ID_0 | (uint32_t)space_id,
+                                        page_no},
```
dr-m: Here we are using a mix of C++11 brace constructor syntax and an archaic C-style cast. The C++11 widening conversion `uint32{space_id}` would be equally long. We might also replace the `bool` parameter with the actual constant, or replace the two first parameters with `page_id_t` (which is a simple wrapper for a 64-bit unsigned integer). I didn’t check how this is being compiled, but https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI would hint that we are in the territory where parameters will need to be passed via the stack. Fewer parameters could reduce some stack use.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2035,6 +2037,47 @@ static bool log_checkpoint_low(lsn_t oldest_lsn, lsn_t end_lsn) noexcept
...
+                              nullptr, size, alloc, offset);
+  if (alloc)
+          {
```
dr-m: The opening brace is indented too much.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -3835,6 +3777,16 @@ static int innodb_init_params()
     DBUG_RETURN(HA_ERR_INITIALIZATION);
   }
 
+	if (innodb_binlog_state_interval == 0 ||
+	    innodb_binlog_state_interval !=
+		(ulonglong)1 << (63 - my_nlz(innodb_binlog_state_interval)) ||
+	    innodb_binlog_state_interval % (ulonglong)ibb_page_size) {
+		ib::error() << "innodb_binlog_state_interval must be a "
+			"power-of-two multiple of the innodb binlog page size="
+			<< ibb_page_size;
```
dr-m: Please, let us avoid using `ib::logger` in any new code. It was originally introduced to avoid errors that would have been caught by `__attribute__((printf(…)))`. Now we finally have that attribute on our formatting functions.
```c++
sql_print_error("InnoDB: innodb_binlog_state_interval ...=%lu KiB", ibb_page_size >> 10);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -4083,6 +4035,41 @@ static void innobase_update_optimizer_costs(OPTIMIZER_COSTS *costs)
...
+  binlog_file_entry **next_ptr= &list;
+  for (uint64_t i= first; i <= last; ++i)
+  {
+    binlog_file_entry *e= (binlog_file_entry *)alloc_root(mem_root, sizeof(*e));
+    if (!e)
+      return nullptr;
+    char name_buf[OS_FILE_MAX_PATH];
+    binlog_name_make(name_buf, i);
+    e->name.length= strlen(name_buf);
+    char *str= (char *)alloc_root(mem_root, e->name.length + 1);
```
dr-m: Can we please use `static_cast` here?

storage/innobase/include/fsp0types.h
```diff
@@ -27,8 +27,12 @@ Created May 26, 2009 Vasil Dimov
...
 
-/** All persistent tablespaces have a smaller fil_space_t::id than this. */
+/** All persistent tablespaces (except binlog tablespaces) have a smaller
+fil_space_t::id than this. */
 constexpr uint32_t SRV_SPACE_ID_UPPER_BOUND= 0xFFFFFFF0U;
+/** Binlog tablespaces. */
+constexpr uint32_t SRV_SPACE_ID_BINLOG0 = SRV_SPACE_ID_UPPER_BOUND + 1;
+constexpr uint32_t SRV_SPACE_ID_BINLOG1 = SRV_SPACE_ID_UPPER_BOUND + 2;
 /** The fil_space_t::id of the innodb_temporary tablespace. */
 constexpr uint32_t SRV_TMP_SPACE_ID= 0xFFFFFFFEU;
```
dr-m, line 34: I think that we were doing `SRV_SPACE_ID_BINLOG0 | bool_space_id` somewhere. That would seem to assume that the least significant bit of `SRV_SPACE_ID_BINLOG0` is clear. A `static_assert` about that next to the use would be nice.

I also think that we can define `SRV_SPACE_ID_BINLOG0= SRV_SPACE_ID_UPPER_BOUND`. It is best to double-check the normal tablespace allocation code.

We might also define the binlog tablespaces as ~0 and ~1, and move the `SRV_TMP_SPACE_ID` to ~2 if that would lead to smaller code somewhere. None of these numeric identifiers are going to be written to any persistent files, right? But, I see that the logic around `end_page_id` (defined in `mtr0log.h`) would then have to change.

storage/innobase/include/fsp_binlog.h
```diff
@@ -0,0 +1,535 @@
...
+  size of the file from the header page, but still be able to checksum the
+  header page without relying on unchecked page size field to compute the
+  checksum.
+
+  (The remainder of the header page is just unused or could potentially
+  later be used for other data as needed).
+*/
+static constexpr uint32_t IBB_HEADER_PAGE_SIZE= 512;
+static constexpr uint32_t IBB_PAGE_SIZE_MIN= IBB_HEADER_PAGE_SIZE;
+static constexpr uint32_t IBB_PAGE_SIZE_MAX= 65536;
```
dr-m, line 57: If these constants are going to be written to the file header, I think that it would be better to use base-2 logarithms of these, that is, 9 to 16, instead of writing 1<<9 to 1<<16.

storage/innobase/include/fsp_binlog.h
```diff
@@ -0,0 +1,535 @@
...
+  FSP_BINLOG_TYPE_GTID_STATE= 2,
+  /* Out-of-band event group data. */
+  FSP_BINLOG_TYPE_OOB_DATA= 3,
+  /* Dummy record, use to fill remainder of page (eg. FLUSH BINARY LOGS). */
+  FSP_BINLOG_TYPE_DUMMY= 4,
+  /* Must be one more than the last type. */
+  FSP_BINLOG_TYPE_END,
+
+  /* Padding data at end of page. */
+  FSP_BINLOG_TYPE_FILLER= 0xff
```
dr-m, line 85: Wouldn’t 0x00 be better? There exist special ways to zero-initialize blocks. I think that I saw somewhere discussion about a new Linux system call variant (maybe `ftruncate` or `fallocate`) that would allow a block to be zero-initialized.

By the way, the comments in this file seem to wrongly start with `/*` instead of the Doxygen style `/**`.

storage/innobase/include/log0log.h
```diff
@@ -102,6 +102,10 @@ or the MySQL version that created the redo log file. */
 #define LOG_HEADER_CREATOR_END	48
 /* @} */
 
+/** Fake tablespace id for InnoDB-implemented binlog files. */
+#define LOG_BINLOG_ID_0 0xFFFFFFF0U
+#define LOG_BINLOG_ID_1 0xFFFFFFF1U
```
dr-m: Elsewhere, we define `SRV_SPACE_ID_BINLOG0` and `SRV_SPACE_ID_BINLOG1`, and they are currently off by 1 from these. I think that one set of constants would be better.

I dont think we should use `#define` for introducing any new constants. `constexpr` is better.

storage/innobase/include/mtr0log.h
```diff
@@ -389,20 +389,26 @@ inline byte *mtr_t::log_write(const page_id_t id, const buf_page_t *bpage,
...
                 type <= FILE_CHECKPOINT, "invalid type");
-  ut_ad(type >= FILE_CREATE || is_named_space(id.space()));
+  ut_ad(type >= FILE_CREATE || is_named_space(id.space()) ||
+        id.space() == LOG_BINLOG_ID_0 ||
+        id.space() == LOG_BINLOG_ID_1);
   ut_ad(!bpage || bpage->id() == id);
-  ut_ad(id < end_page_id);
+  ut_ad(id < end_page_id ||
+        id.space() == LOG_BINLOG_ID_0 ||
+        id.space() == LOG_BINLOG_ID_1);
```
dr-m, line 322: Could we define `static constexpr` for these constants in `page_id_t`, and introduce `page_id_t::is_binlog()` as well? Or just relax the condition to `mtr_t::is_named_space()`?

storage/innobase/include/small_vector.h
```diff
@@ -19,8 +19,10 @@ this program; if not, write to the Free Software Foundation, Inc.,
 #pragma once
 /* A normally small vector, inspired by llvm::SmallVector */
 #include "my_global.h"
+#include "my_valgrind.h"
 #include <iterator>
 #include <memory>
+#include <algorithm>
```
dr-m: I see that the header uses macros that are defined in `my_valgrind.h`. But `#include <algorithm>` looks like an overkill. https://en.cppreference.com/w/cpp/utility/move.html says that `std::move` is defined in `<utility>`. Is this about that?

storage/innobase/include/trx0trx.h
```diff
@@ -827,7 +827,10 @@ struct trx_t : ilist_node<>
 					is set to false  after commit or
 					rollback. */
 	/** whether this is holding the prepare mutex */
+        /* ToDo: This need a better mechanism. It is currently done to know that we did not do a prepare step before commit_ordered, due to binlog being stored in InnoDB; and therefore we need to do an fsync of the log in commit to make the commit durable. */
 	bool		active_commit_ordered;
+	/** whether innobase_xa_prepare() was done. */
+	bool		active_prepare;
 	/*------------------------------*/
 	bool		flush_log_later;/* In 2PC, we hold the
 					prepare_commit mutex across
```
dr-m, line 837: I was hoping that all this could be bypassed or removed when using the `ib_logfile0` for binlog durability.

storage/innobase/srv/srv0start.cc
```diff
@@ -1939,6 +1940,8 @@ dberr_t srv_start(bool create_new_db)
 		return(srv_init_abort(err));
 	}
 
+        innodb_binlog_startup_init();
```
dr-m: There should be some error handling, right?

dr-m (commented):
I quickly went through the InnoDB changes, except for the actual low-level implementation. The interface seems to correspond to what we have discussed in the past.

extra/mariabackup/xtrabackup.cc
knielsen: Yes, that seems to be possible, nice! Will fix.

storage/innobase/CMakeLists.txt
knielsen: Indeed, it works without this now after I've rebased on latest 11.4, will remove.
Good catch!

storage/innobase/buf/buf0flu.cc
knielsen: Agree, I will replace with a single page_id_t parameter.

storage/innobase/buf/buf0flu.cc
knielsen: wierd, will fix.

storage/innobase/handler/ha_innodb.cc
knielsen: ack, will fix. (I checked the full patch, no more ib::error or ib::warning references left now)

storage/innobase/handler/ha_innodb.cc
knielsen: Ack, will fix.
(Also changed a bunch of casts in main implementation files to `static_cast<>()`)

storage/innobase/include/fsp0types.h
knielsen: Right, sorry about these, the SRV_SPACE_ID_BINLOG[01] constants are left-overs from an early version that used the InnoDB tablespace and buffer pool code that I forgot to remove. I will remove them completely from the patch.

Agree about static_assert(), like this:
```c++
    static_assert((LOG_BINLOG_ID_0 | 1) == LOG_BINLOG_ID_1, "");
    is_binlog= !ENC_10_8 && storing == YES && (space_id | 1) == LOG_BINLOG_ID_1;
```

I don't have a strong opinion if we should change to ~0 and ~1. I don't think it matters a lot for binlog code, we can reserve them for a later more important use, or I am happy to change it if you wish.

storage/innobase/include/fsp_binlog.h
knielsen: They are not written.
What is written is the actual page size, and it is defined as the log2 as you suggested:
```c++
uint32_t ibb_page_size_shift= 14;
```

storage/innobase/include/fsp_binlog.h
knielsen: In fact 0x00 is used to fill unused pages or part of pages.
This `FSP_BINLOG_TYPE_FILLER` is specifically to mark that bytes in a page are used, but they do not contain any content. This is only used for a few bytes at the end of a page (at most 3) when there is no room to place a new record there, or in the rare case where the user forces a binlog file to end in the middle of a page with `FLUSH BINARY LOGS`. The 0xff is used to distinguish from the normal 0x00, which means end-of-file.

I'll try to fix comments to use `/**` instead of `/*` here and elsewhere.

storage/innobase/include/log0log.h
knielsen: Yes, the SRV_SPACE_ID_BINLOG* constants are left-overs from early development, I will remove them.
And changed to `constexpr`, I agree, I'm trying to switch to doing that myself for new code.

storage/innobase/include/mtr0log.h
knielsen: Do you mean to change `mtr_t::is_named_space()` to return true also for the two binlog space ids? That's fine with me, I'm not sure exactly what the semantics of "named space" is and if it makes sense to consider the binlog files as such.

storage/innobase/include/small_vector.h
knielsen: I don't remember, I got a compile error when I started using small_vector.h, presumably because I was not including something else before. I cannot reproduce the build error now, so I will remove it.

If it turns out that it depends on compiler or such, and I get buildbot failures on some builders, I will check what is actually used and find the right include.

storage/innobase/include/trx0trx.h
knielsen: This is related to supporting both the old binlog format and the new binlog format.

When using the old binlog format, we need two-phase commit, which requires an fsync of the InnoDB redo log at "prepare" time, and an fsync of the binlog at binlog write time. There used to be also an fsync of the InnoDB redo log at "commit" time, when `--innodb-flush-log-at-trx-commit=1`. But an optimisation removes this third fsync, using another mechanism (binlog checkpoint and binlog recovery) to still guarantee the durability of the commit.

In the new binlog formal, using `ib_logfile0` for binlog durability, both the fsync at prepare and the fsync at binlog write is gone. But then we now need to do the sync of the InnoDB redo log at commit, otherwise we lose durability of the commit.

Thus we need a way to know in InnoDB commit whether we are in the old binlog case or the new binlog case, so we know whether we can skip the fsync there or not. So the mechanism is needed, but I put that ToDo comment to see if a more elegant mechanism can be found to decide on whether the fysnc at commit should be done or not. So suggestions are welcome, though it is not a critical issue, just a place where it would be nice if a flag and an extra condition could be removed somehow.

(I will clean up all these remaining ToDo comments before pushing, fix the code or replace with a proper comment explaining the situation, as appropriate)

storage/innobase/srv/srv0start.cc
knielsen: In this particular case, `innodb_binlog_startup_init()` cannot really fail. It is a low-level init that mustly initializes mutexes and such. It does not yet create any binlog files or write to them; this only happens later, when requested from the server layer to open the binlog (and only occurs if the binlog is enabled by the user).

But in general, I'm unsure what to do with certain types of error handling. I've noticed some places in InnoDB code where it does `ut_a()` on error conditions. If we're deep in transaction commit (or even mini-transaction commit) and get an error writing to a log file or other file, what can we do?

Maybe we can discuss this further, I'd like to hear your experiences around how error handling happens currently in InnoDB.

storage/innobase/handler/innodb_binlog.cc
```diff
@@ -0,0 +1,4587 @@
...
+      file_no= binlog_files.last_file_no;
+      if (ibb_record_in_file_hash(file_no, header.oob_ref_file_no,
+                                  header.xa_ref_file_no))
+        return -1;
+      binlog_discover_init(file_no, header.diff_state_interval);
+      binlog_cur_page_no= page_no;
+      binlog_cur_page_offset= pos_in_page;
+      ib::info() << "Continuing binlog number " << file_no << " from position "
+                 << (((uint64_t)page_no << page_size_shift) | pos_in_page)
+                 << ".";
```
dr-m: You removed other added references to `ib::logger`, but in this function there are a few `ib::info()`. Could we use `sql_print_information()` instead? Should these messages be conditional based on `global_system_variables.log_warnings`?

knielsen:
Marko Mäkelä ***@***.***> writes:

> cbc32ca8772b1efaba5a4571a918c09de27cb309 promises that some crash
> recovery tests would be added later. I am looking forward to that. I
> did not find any added reference to
> `include/kill_and_restart_mysqld.inc`, or to
> `include/restart_mysqld.inc` with a prior `let $shutdown_timeout=0`.

There are two small recovery tests binlog_in_engine.recovery and
binlog_in_engine.recovery_large, they are using DBUG injections to crash the
server.

The binlog_in_engine.mariabackup_slave_provision exercises the binlog
recovery code in a different way.

I agree more is needed. As we discussed on the Zulip I will add some test
that also runs in non-debug and exercises some pseudo-random parallel load
and crashes it in the middle.

>
>> +      ib::info() << "Continuing binlog number " << file_no << " from position "
> +                 << (((uint64_t)page_no << page_size_shift) | pos_in_page)
> +                 << ".";
>
> You removed other added references to `ib::logger`, but in this
> function there are a few `ib::info()`. Could we use

Bummer, I intended that all of these should be gone, but I must have
searched for ib::error/ib::warning and forgotten about ib::info. Will fix,
sorry, thanks for spotting.

> `sql_print_information()` instead? Should these messages be
> conditional based on `global_system_variables.log_warnings`?

I assume just using sql_print_information() would make these controllable in
the normal way the user would expect?

 - Kristian.

storage/innobase/log/log0recv.cc
```diff
@@ -2804,10 +2806,10 @@ log_parse_file(const page_id_t id, bool if_exists,
 @retval GOT_EOF       if the log is corrupted */
 ATTRIBUTE_NOINLINE
 static recv_sys_t::parse_mtr_result
-log_page_modify(uint32_t space_id, uint32_t page_no) noexcept
+log_page_modify(uint32_t space_id, uint32_t page_no, bool is_binlog) noexcept
 {
   ut_ad(space_id);
-  if (srv_is_undo_tablespace(space_id))
+  if (is_binlog || srv_is_undo_tablespace(space_id))
     return recv_sys_t::OK;
```
dr-m, line 2850: Can we avoid introducing the redundant parameter? Instead, just modify the condition:
```c++
  if (space_id >= SRV_SPACE_ID_UPPER_BOUND || srv_is_undo_tablespace(space_id))
    return recv_sys_t::OK;
```

storage/innobase/log/log0recv.cc
```diff
@@ -3078,28 +3083,34 @@ recv_sys_t::parse_tail(const byte *begin, bool if_exists, size_t size) noexcept
           ignore the payload and only compute the mini-transaction checksum;
           there will be a subsequent call with storing==YES. */
           continue;
+        if (storing == NO)
+          is_binlog= false;
```
dr-m: Instead of assigning `is_binlog` back and forth, we’d better change the subsequent condition for testing it:
```c++
  if (storing == NO || !is_binlog)
```
Even better would be to augment the preceding condition, which the GitHub web interface does not me allow to quote in full:
```c++
      case WRITE:
      case MEMMOVE:
      case MEMSET:
        if (storing == BACKUP)
          continue;
        if (storing == NO && UNIV_LIKELY(page_no != 0))
          /* fil_space_set_recv_size_and_flags() is mandatory for storing==NO.
          It is only applicable to page_no == 0. Other than that, we can just
          ignore the payload and only compute the mini-transaction checksum;
          there will be a subsequent call with storing==YES. */
          continue;
```
It would seem to me that the condition could be revised to the following:
```c++
if (storing == NO && UNIV_LIKELY((page_no | is_binlog) != 0))
  continue;
```
The bitwise OR works here and will avoid introducing any conditional branch.

storage/innobase/log/log0recv.cc
```diff
@@ -3078,28 +3083,34 @@ recv_sys_t::parse_tail(const byte *begin, bool if_exists, size_t size) noexcept
...
+          if (UNIV_UNLIKELY(last_offset < 8 || last_offset >= srv_page_size))
+            goto record_corrupted;
+          (ENC_10_8 ? cl : l)+= olen;
+          rlen-= olen;
+        }
         if ((b & 0x70) == WRITE)
         {
-          if (UNIV_UNLIKELY(rlen + last_offset > srv_page_size))
+          if (!ENC_10_8 && is_binlog);
+          else if (UNIV_UNLIKELY(rlen + last_offset > srv_page_size))
```
dr-m, line 3149: The only purpose for this handling for the `storing==NO` branch is to maintain `last_offset` for keeping track of writes to `FSP_SIZE` and `FSP_SPACE_FLAGS` in InnoDB tablespace headers. If `is_binlog` holds, we can skip all of the `(b & 0x70) == WRITE` code path as well, including the `storing==YES` case. So, we can simplify this to the following, with fewer run-time conditions to evaluate:
```c++
        if (ENC_10_8)
          cl= log_decrypt_legacy(iv, recs, l, rlen, decrypt_buf);
        else if (storing == NO || !is_binlog)
        {
          ut_ad(!is_binlog);
          {
            const uint32_t olen= mlog_decode_varint_length(*(ENC_10_8 ? cl : l));
            if (UNIV_UNLIKELY(olen >= rlen) || UNIV_UNLIKELY(olen > 3))
              goto record_corrupted;
            const uint32_t offset= mlog_decode_varint(ENC_10_8 ? cl : l);
            ut_ad(offset != MLOG_DECODE_ERROR);
            static_assert(FIL_PAGE_OFFSET == 4, "compatibility");
            if (UNIV_UNLIKELY(offset >= srv_page_size))
              goto record_corrupted;
            last_offset+= offset;
            if (UNIV_UNLIKELY(last_offset < 8 || last_offset >= srv_page_size))
              goto record_corrupted;
            (ENC_10_8 ? cl : l)+= olen;
            rlen-= olen;
          }
          if ((b & 0x70) == WRITE)
          {
// …
          parsed_ok:
            last_offset+= rlen;
            if (ENC_10_8 && size_t(cl - decrypt_buf) < srv_page_size)
              (l= recs)+= cl - decrypt_buf;
            break;
          }
        }
```

storage/innobase/log/log0recv.cc
```diff
@@ -3161,6 +3172,21 @@ recv_sys_t::parse_tail(const byte *begin, bool if_exists, size_t size) noexcept
...
+          goto record_corrupted;
+        const uint32_t offset= mlog_decode_varint(l);
+        ut_ad(offset != MLOG_DECODE_ERROR);
+        if (UNIV_UNLIKELY(offset + rlen - olen >= 65535))
+          goto record_corrupted;
+        if (binlog_recover_write_data(space_id & 1, page_no, uint16_t(offset),
+                                      start_lsn, lsn, l + olen, rlen - olen))
+          goto record_corrupted;
+      }
       else if (if_exists && !parse_store_if_exists(space_id));
```
dr-m, line 3218: This could be separated to another non-inline function, similar to how we have `recv_sys_t::parse_store_if_exists()` below. Admittedly, unlike that code, this one will be expanded only for 1 template parameter combination. But I believe that writes to the binlog should be much more seldom than writes to InnoDB tablespaces. Having the code in a separate `ATTRIBUTE_COLD ATTRIBUTE_NOINLINE` function could make the common code path faster. Something like this:
```c++
      if (storing != YES);
      else if (!ENC_10_8 && is_binlog && parse_store_binlog(space_id, l, rlen, page_no, start_lsn, lsn))
        goto record_corrupted;
      else if (if_exists …
```
We can check for `WRITE` much earlier:
```c++
      case MEMMOVE:
      case MEMSET:
        if (storing == YES && !ENC_10_8 && is_binlog)
          goto record_corrupted;
        /* fall through */
      case WRITE:
```

storage/innobase/srv/srv0start.cc
dr-m: InnoDB followed the bad practice of committing suicide on any error, but there have been some improvements. In MariaDB Server 10.2, I refactored the InnoDB start-up code so that the plugin initialization would fail if an error would occur. We used to call `exit()` or `abort()`. A more substantial change was 0b47c126e31cddda1e94588799599e138400bcf8 and some follow-up changes, to generally avoid a crash upon encountering any corruption.

storage/innobase/handler/innodb_binlog.cc
knielsen: Fixed.

storage/innobase/log/log0recv.cc
knielsen: Fixed.

storage/innobase/log/log0recv.cc
knielsen: Fixed.

storage/innobase/log/log0recv.cc
knielsen: This I could not get to work.
The problem is that the `if ((b & 0x70) == WRITE) { ... }` part includes this code at the end, which is still run for the `is_binlog` case and which is needed AFAICS:
```c++
        parsed_ok:
          last_offset+= rlen;
          if (ENC_10_8 && size_t(cl - decrypt_buf) < srv_page_size)
            (l= recs)+= cl - decrypt_buf;
          break;
        }
```
With your suggestion, the `parsed_ok` block is not reached in the case where `storing == YES` and `is_binlog`, and the code fails.

But maybe I misunderstood your question, I don't really understand the details of the logic here.

storage/innobase/log/log0recv.cc
knielsen: Fixed.

storage/innobase/srv/srv0start.cc
knielsen: I fixed the code to return DB_OUT_OF_MEMORY if it cannot allocate memory needed during initialization.

knielsen:
@dr-m Thanks for your additional comments. I've fixed everything except for [this one](https://github.com/MariaDB/server/pull/3775#discussion_r2458197260) which I couldn't get to work or misunderstood.

knielsen:
This has been released in MariaDB 12.3.1

PR #3776, opened 2025-01-20
MDEV-35445   Disable foreign key column nullability check for strict sql mode

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35445*

## Description

- MDEV-34392(commit cc810e64d40367208b3f3f35c0277f0c8586a293) adds the check for nullability of foreign key column when foreign key relation is of UPDATE_CASCADE or UPDATE SET NULL. This check makes DDL fail when it violates foreign key nullability. This patch basically does the nullability check for foreign key column only for strict sql mode

## How can this PR be tested?
./mtr innodb.foreign_sql_mode

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/foreign_sql_mode.test
+++ b/mysql-test/suite/innodb/t/foreign_sql_mode.test
@@ -1,0 +1,129 @@
+--source include/have_innodb.inc
+--source alter_sql_mode.inc
+call mtr.add_suppression("InnoDB: In ALTER TABLE .* has or is referenced in foreign key constraints which are not compatible with the new table definition.");
+
+let $combination=`select regexp_replace('$MTR_COMBINATIONS', 'innodb,\|,innodb', '')`;
+
+let $copy_algo=`select ((strcmp(substring_index('$combination', ",", 1), "COPY") = 0) or (strcmp(substring_index('$combination', ",", -1), "COPY") = 0))`;
+
+let $inplace_algo=`select ((strcmp(substring_index('$combination', ",", 1), "INPLACE") = 0) or (strcmp(substring_index('$combination', ",", -1), "INPLACE") = 0))`;
+
+let $algorithm=COPY;
+if ($inplace_algo)
+{
+  let $algorithm=INPLACE;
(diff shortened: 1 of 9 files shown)


mysql-test/suite/innodb/t/foreign_sql_mode.test
```diff
@@ -0,0 +1,112 @@
...
+		 f2 INT NOT NULL,
+		 f3 INT DEFAULT NULL,
+		 PRIMARY KEY(f1, f2),
+		 FOREIGN KEY(f2, f3) REFERENCES t1(f2, f1)
+		 ON UPDATE CASCADE)ENGINE=InnoDB;
+if ($error_code == 0)
+{
+  DROP TABLE t2;
+}
+DROP TABLE t1;
```
dr-m, line 112: We could shorten the `.rdiff` files if we just wrote the following:
```sql
DROP TABLE IF EXISTS t2;
```

dr-m (commented):
I am happy to see that the DML time checks are already in place. I think that we must handle the `ON UPDATE SET NULL` case symmetrically. Everything else in the logic looks OK to me.

mysql-test/suite/innodb/t/foreign_sql_mode.test
```diff
@@ -0,0 +1,127 @@
...
+if ($sql_mode == "STRICT_TRANS_TABLES")
+{
+  let $dml_error_code = 0;
+}
+
+--error $dml_error_code
+UPDATE t1 SET f2= NULL;
+DELETE FROM t2;
+SELECT * FROM t1;
+DROP TABLE t2, t1;
```
dr-m, line 41: I thought that we could add a second `UPDATE` after the `DELETE`, to demonstrate that it will always succeed.

dr-m (approved):
Thank you. I see that our behaviour regarding allowing the creation of `ON UPDATE SET NULL` when the referencing column is `NOT NULL` always was inconsistent or asymmetric. There is no need to fix that as part of this change.

PR #3779, opened 2025-01-20
MDEV-35891 mtr, Windows - fix multi-process append for stdout and stderr

vaintroub:
This backports https://github.com/mysql/mysql-server/commit/e2f685972985, which works around perl issue https://github.com/Perl/perl5/issues/17570

MySQL fix  ensures that FILE_APPEND_DATA is used in redirected stdout/stderr open flags, rather than FILE_GENERIC_WRITE. Only this gives lossless append behavior, if multiple processes use the same file.

- [x] *The Jira issue number for this PR is: MDEV-35891*

## Description
See commit comments. This is a workaround for perl bug, which affects mysql-test-run.pl, which uses undocumented system(1,..) call.

## Release Notes
Release notes should not say anything , it only affects testing framework.

## How can this PR be tested?

This is a fix in test infrastructure. It is enough if MTR passes on Windows on CI.

## PR quality check

### DIFF
--- a/mysql-test/lib/My/SafeProcess/safe_process_win.cc
+++ b/mysql-test/lib/My/SafeProcess/safe_process_win.cc
@@ -152,6 +152,24 @@ void handle_signal (int signal)
   }
 }
 
+/**
+  Sets the append flag (FILE_APPEND_DATA) so that the handle inherited by the
+  child process will be in append mode.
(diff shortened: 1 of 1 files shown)


dr-m (approved):
I see that you didn’t update the Oracle copyright year to 2020. That may be OK, because the fix is dictated by the Win32 API and you are adding some commentary that differs from Oracle’s.

PR #3780, opened 2025-01-21
MDEV-7611 mysqldump --dump-slave always starts stopped slave

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-7611][]*
  * This bug’s reported 10½ years ago!? :astonished:

## Description
Store the _SLAVE_ STATUS when pausing replicas for `mariadb-dump --dump-slave`;
use that (instead of a new SHOW _SLAVE_ STATUS) to resume replicas when the program ends.

## Release Notes
* Fix `mariadb-dump --dump-thread` starts replication threads that were not running

## How can this PR be tested?
* The new `multi_source.mariadb-dump_slave` tests this issue for multi-source and, by consequence, single-source replication.
* Edge cases and changed side-effects remain testable by existing methods such as `main.mysqldump`.

* [x] *This is a bug fix, and the PR is based against the earliest _maintained_ branch in which the bug can be reproduced.*

## PR quality check

[MDEV-7611]: https://jira.mariadb.org/browse/MDEV-7611

### DIFF
--- a/client/mysqldump.c
+++ b/client/mysqldump.c
@@ -184,7 +184,7 @@ static DYNAMIC_STRING extended_row;
 static DYNAMIC_STRING dynamic_where;
 static MYSQL_RES *get_table_name_result= NULL;
 static MEM_ROOT glob_root;
-static MYSQL_RES *routine_res, *routine_list_res;
+static MYSQL_RES *routine_res, *routine_list_res, *slave_status_res= NULL;
 
 
 #include <sslopt-vars.h>
@@ -1908,6 +1908,8 @@ static void free_resources()
     mysql_free_result(routine_res);
   if (routine_list_res)
     mysql_free_result(routine_list_res);
+  if (slave_status_res)
+    mysql_free_result(slave_status_res);
   if (mysql)
   {
     mysql_close(mysql);
@@ -6257,17 +6259,19 @@ static int do_show_master_status(MYSQL *mysql_con, int consistent_binlog_pos,
 
 static int do_stop_slave_sql(MYSQL *mysql_con)
 {
-  MYSQL_RES *slave;
   MYSQL_ROW row;
+  DBUG_ASSERT(
+    !slave_status_res // do_stop_slave_sql() should only be called once
+  );
 
-  if (mysql_query_with_error_report(mysql_con, &slave,
+  if (mysql_query_with_error_report(mysql_con, &slave_status_res,
                                     multi_source ?
                                     "SHOW ALL SLAVES STATUS" :
                                     "SHOW SLAVE STATUS"))
     return(1);
 
   /* Loop over all slaves */
-  while ((row= mysql_fetch_row(slave)))
+  while ((row= mysql_fetch_row(slave_status_res)))
   {
     if (row[11 + multi_source])
     {
@@ -6282,13 +6286,11 @@ static int do_stop_slave_sql(MYSQL *mysql_con)
 
         if (mysql_query_with_error_report(mysql_con, 0, query))
         {
-          mysql_free_result(slave);
           return 1;
         }
       }
     }
   }
-  mysql_free_result(slave);
   return(0);
 }
 
@@ -6412,32 +6414,35 @@ static int do_show_slave_status(MYSQL *mysql_con, int have_mariadb_gtid,
 
 static int do_start_slave_sql(MYSQL *mysql_con)
 {
-  MYSQL_RES *slave;
   MYSQL_ROW row;
   int error= 0;
   DBUG_ENTER("do_start_slave_sql");
 
-  /* We need to check if the slave sql is stopped in the first place */
-  if (mysql_query_with_error_report(mysql_con, &slave,
-                                    multi_source ?
-                                    "SHOW ALL SLAVES STATUS" :
-                                    "SHOW SLAVE STATUS"))
-    DBUG_RETURN(1);
+  /*
+    do_start_slave_sql() should normally be called
+    sometime after do_stop_slave_sql() succeeds
+  */
+  if (!slave_status_res)
+    DBUG_RETURN(error);
+  mysql_data_seek(slave_status_res, 0);
 
-  while ((row= mysql_fetch_row(slave)))
+  while ((row= mysql_fetch_row(slave_status_res)))
   {
     DBUG_PRINT("info", ("Connection: '%s'  status: '%s'",
                         multi_source ? row[0] : "", row[11 + multi_source]));
     if (row[11 + multi_source])
     {
-      /* if SLAVE SQL is not running, we don't start it */
-      if (strcmp(row[11 + multi_source], "Yes"))
+      /*
+        If SLAVE_SQL was not running but is now,
+        we start it anyway to warn the unexpected state change.
+      */
+      if (strcmp(row[11 + multi_source], "No"))
       {
         char query[160];
         if (multi_source)
-          sprintf(query, "START SLAVE '%.80s'", row[0]);
+          sprintf(query, "START SLAVE '%.80s' SQL_THREAD", row[0]);
         else
-          strmov(query, "START SLAVE");
+          strmov(query, "START SLAVE SQL_THREAD");
 
         if (mysql_query_with_error_report(mysql_con, 0, query))
         {
@@ -6448,7 +6453,6 @@ static int do_start_slave_sql(MYSQL *mysql_con)
       }
     }
   }
-  mysql_free_result(slave);
   DBUG_RETURN(error);
 }
 
--- a/mysql-test/suite/multi_source/mariadb-dump_slave.test
+++ b/mysql-test/suite/multi_source/mariadb-dump_slave.test
@@ -1,0 +1,55 @@
+# `mariadb-dump --dump-slave` multi-source interactions
+# (see `main.rpl_mysqldump_slave` for general testing with single-source)
+
+--source include/have_log_bin.inc
+--let $status_items= Connection_name, Slave_IO_Running, Slave_SQL_Running
+--let $all_slaves_status= 1
+
+# $MYSQL_DUMP dumps the $SERVER_MYPORT_1 server
+--connect (  active_master, 127.0.0.1, root, , , $SERVER_MYPORT_2)
+--connect (inactive_master, 127.0.0.1, root, , , $SERVER_MYPORT_3)
+--connect (          slave, 127.0.0.1, root, , , $SERVER_MYPORT_1)
+
+--replace_result $SERVER_MYPORT_2 MYPORT_2
+eval CHANGE MASTER TO
+  master_host='127.0.0.1', master_port=$SERVER_MYPORT_2, master_user='root';
+START SLAVE SQL_THREAD;
+--replace_result $SERVER_MYPORT_3 MYPORT_3
+eval CHANGE MASTER 'inactive' TO
+  master_host='127.0.0.1', master_port=$SERVER_MYPORT_3;
+# wait for the active default '' connection only
+--source include/wait_for_slave_sql_to_start.inc
+
+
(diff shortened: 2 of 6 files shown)


client/mysqldump.c
```diff
@@ -6250,43 +6264,29 @@ static int do_show_slave_status(MYSQL *mysql_con, int have_mariadb_gtid,
...
-
-  /* We need to check if the slave sql is stopped in the first place */
-  if (mysql_query_with_error_report(mysql_con, &slave,
-                                    multi_source ?
-                                    "SHOW ALL SLAVES STATUS" :
-                                    "SHOW SLAVE STATUS"))
-    DBUG_RETURN(1);
-
-  while ((row= mysql_fetch_row(slave)))
+  for (; n_stopped_replicas--;)
```
ParadoxV5, line 6269: To use the new list instead of SHOW SLAVE STATUS, I essentially rewrote `do_start_slave_sql`…

client/mysqldump.c
```diff
@@ -6097,12 +6099,22 @@ static int do_stop_slave_sql(MYSQL *mysql_con)
...
+  // do_stop_slave_sql() should only be called once
+  DBUG_ASSERT(!stopped_replicas);
 
   if (mysql_query_with_error_report(mysql_con, &slave,
                                     multi_source ?
                                     "SHOW ALL SLAVES STATUS" :
                                     "SHOW SLAVE STATUS"))
     return(1);
+  stopped_replicas= my_malloc(PSI_NOT_INSTRUMENTED,
+                              slave->row_count*NAME_CHAR_LEN + 1, MYF(MY_WME));
```
ParadoxV5, line 6111: I feel that it’s wrong wasting memory on short names and already-stopped replicas, but I feel that it’s more wrong spending extra effort `realloc`ing unpredictably.

I’d like to iterate the `MYSQL_RES` an additional time before the following loop to sum up name sizes of (only) active replications.
Question: How to rewind it from `mysql_fetch_row()`s?

---

Note I didn’t check size overflow here because of [the Broken Windows Theory](https://en.wikipedia.org/wiki/Broken_windows_theory):
https://github.com/MariaDB/server/blob/6aa47fae304bd1f3efc0ebd2035bc1481ded73a3/client/mysqldump.c#L1986
https://github.com/MariaDB/server/blob/6aa47fae304bd1f3efc0ebd2035bc1481ded73a3/client/mysqldump.c#L6620

client/mysqldump.c
```diff
@@ -6123,6 +6135,8 @@ static int do_stop_slave_sql(MYSQL *mysql_con)
           mysql_free_result(slave);
           return 1;
         }
+        strmov(stopped_replicas[n_stopped_replicas++],
+               multi_source ? row[0] : "");
```
ParadoxV5, line 6139: pedantically supporting no-multi-source on a best effort basis
(But those pre-10.0 versions were long-EOL… 😕)
https://github.com/MariaDB/server/blob/6aa47fae304bd1f3efc0ebd2035bc1481ded73a3/client/mysqldump.c#L7010-L7015

client/mysqldump.c
```diff
@@ -6250,43 +6264,29 @@ static int do_show_slave_status(MYSQL *mysql_con, int have_mariadb_gtid,
...
-                        multi_source ? row[0] : "", row[11 + multi_source]));
-    if (row[11 + multi_source])
+    /*
+      do_start_slave_sql() should only be called
+      sometime after do_stop_slave_sql() suceeds
+    */
+    char* stopped_replica= stopped_replicas[n_stopped_replicas];
+    char query[sizeof("START SLAVE ''") + NAME_CHAR_LEN];
+    DBUG_PRINT("info", ("Connection: '%.*s'", NAME_CHAR_LEN, stopped_replica));
+    // if SLAVE SQL is running, start it anyway to warn unexpected state change
```
ParadoxV5, line 6278: The program runs SHOW SLAVE STATUS multiple times thoughout the various step functions:
* `do_show_slave_status`
* `do_stop_slave_sql`
* `do_start_slave_sql` (before this PR)

Perhaps the intent is in case replication configs change in the middle of `mariadb-dump`?

mysql-test/suite/multi_source/mariadb-dump_slave.test
```diff
@@ -0,0 +1,42 @@
+# Interactions with --dump-slave; based on `main.rpl_mysqldump_slave`
```
ParadoxV5, line 1: and `multi_source.gtid_slave_pos` for the setup

mysql-test/suite/rpl/t/rpl_mysqldump_gtid_slave_pos.test
```diff
@@ -71,7 +71,6 @@ select @@global.gtid_slave_pos as "after initial slave got in sync";
 --echo # 3. A
 # Two dumps prepared to be restored in the following loop
 --exec $MYSQL_DUMP_SLAVE --dump-slave  --gtid mysql gtid_slave_pos > $MYSQLTEST_VARDIR/tmp/dump_2.sql
---source include/stop_slave.inc
```
ParadoxV5, line 74: This test was aware 😄.

ParadoxV5, line 74: This test was aware 😄.

bnestere (changes requested):
Thanks for a nice patch, @ParadoxV5 !

Mostly some minor notes for your consideration, along with a couple test additions.

Also, to the git commit message, could you flip the solution and problem, so the problem is first?

client/mysqldump.c
```diff
@@ -6097,12 +6099,22 @@ static int do_stop_slave_sql(MYSQL *mysql_con)
 {
   MYSQL_RES *slave;
   MYSQL_ROW row;
+  // do_stop_slave_sql() should only be called once
+  DBUG_ASSERT(!stopped_replicas);
```
ParadoxV5, line 6103: This config has a problem similar to what [MDEV-17856](https://jira.mariadb.org/browse/MDEV-17856) had:
There’re two different (sets of) entry points to `do_start_slave_sql()` that passing `stopped_replicas` to it as an argument is impractical if not impossible.
A global variable is the most it could have.

client/mysqldump.c
```diff
@@ -1854,6 +1855,7 @@ static void free_resources()
     mysql_close(mysql);
     mysql= 0;
   }
+  my_free(stopped_replicas);
```
ParadoxV5, line 1858: Hm. Must it remain named `stopped_slaves`?

client/mysqldump.c
ParadoxV5: The intent was to use as few memory for the names as practical, yes, with a packed string array that joins variable-length names with `'\0'` separators.
The first iteration tallys the exact size up, then I `malloc`, and filling the buffer requires another iteration at the end.

However, after several stabs at the implementation, I’m losing interest as the feeling that this is overengineering grew inside me.

How about your perspective, @bnestere?

I considered making the `MYSQL_RES slave` query result public so that `do_start_slave_sql` can reuse that, but I’m concerned that the full result of `SHOW ALL SLAVES STATUS` is a massive waste of memory.
Though, is this what you mean by, “to be re-used by `do_stop_slave_sql()`”?

I plan to refactor these start/stop replica steps to use `SELECT … FROM information_schema.SLAVE_STATUS WHERE …` instead.
By moving the row and column selection to the server side, the result would only contain the names needed here, meaning my memory concern will likely be obsolete.
Unfortunately, 10.11 doesn’t have `IS.SLAVE_STATUS`…

mysql-test/suite/multi_source/mariadb-dump_slave.test
ParadoxV5, line 1: This is ridiculous! The only explicit SQL statements are `START`/`STOP SLAVE SQL_THREAD` for the setup/cleanup! Are we writing SQL here or scripting?

client/mysqldump.c
bnestere: For code-level terminology variables, yes, to stick to the existing convention (so searches, etc are easier). Though for high-level descriptions (e.g. code comments/git commit messages), I often use primary/replica terminology.

client/mysqldump.c
bnestere: > The plan was a packed string array that joins variable-length names with '\0' separators.
The first iteration tallys the exact size up, then I malloc, and filling the buffer requires another iteration at the end.
>
> However, after several stabs at the implementation, I’m losing interest as the feeling that this is overengineering grew inside me.

Generally that seems ok by me (or the safer way would be to have another `uint` array that stores the length of each connection name, rather than `'\0'` searching).


> I considered making the MYSQL_RES slave query result public so that do_start_slave_sql can reuse that, but I’m concerned that the full result of SHOW ALL SLAVES STATUS is a massive waste of memory.
Though, is this what you mean by, “to be re-used by do_stop_slave_sql()”, or was it a different idea?

Yeah that's what I meant. Definitely not the most efficient use of memory, but not massive :innocent:. 

> I plan to, on the main branch , refactor these start/stop replica steps to use SELECT … FROM information_schema.SLAVE_STATUS WHERE … instead.

Another thought is to simply have the patch target 11.8 where we have this available. There's no specific need/request to get this into 10.11 or 11.4. There's less risk that way anyhow.

ParadoxV5: > > I plan to, on the main branch , refactor these start/stop replica steps to use SELECT … FROM information_schema.SLAVE_STATUS WHERE … instead.
> 
> Another thought is to simply have the patch target 11.8 where we have this available. There's no specific need/request to get this into 10.11 or 11.4. There's less risk that way anyhow.

I feel that isn’t “responsible”. We only paused 10.6 bugfixes because of the new “criticals only” policy.

Though from a unique perspective, if this behavior is well known and considered a 🐞 feature rather than a 🐛, this request would instead be a “backward-incompatible change”.
(I doubt it is?)

ParadoxV5: > > I considered making the MYSQL_RES slave query result public so that do_start_slave_sql can reuse that, but I’m concerned that the full result of SHOW ALL SLAVES STATUS is a massive waste of memory.
> > Though, is this what you mean by, “to be re-used by do_stop_slave_sql()”, or was it a different idea?
> 
> Yeah that's what I meant. Definitely not the most efficient use of memory, but not massive 😇.

I’ll go with this suggestion of yours.
It’s the least efficient, but it’s closest to my preferred solution:

> > I plan to, on the main branch , refactor these start/stop replica steps to use SELECT … FROM information_schema.SLAVE_STATUS WHERE … instead.

I noted this follow-up as [MDEV-37146](https://jira.mariadb.org/browse/MDEV-37146); remind me to pay off this tech debt I’m about to borrow.

client/mysqldump.c
ParadoxV5: Oh? 🧐 [The Corp’s Engineering Policy](https://mariadb.com/engineering-policies/) says that 10.5 *Enterprise* goes EOL after 16th this month.
It should be safe to refactor away no-multi-source support.

client/mysqldump.c
```diff
@@ -6412,32 +6414,35 @@ static int do_show_slave_status(MYSQL *mysql_con, int have_mariadb_gtid,
...
 
-  /* We need to check if the slave sql is stopped in the first place */
-  if (mysql_query_with_error_report(mysql_con, &slave,
-                                    multi_source ?
-                                    "SHOW ALL SLAVES STATUS" :
-                                    "SHOW SLAVE STATUS"))
-    DBUG_RETURN(1);
+  /*
+    do_start_slave_sql() should normally be called
+    sometime after do_stop_slave_sql() suceeds
```
bnestere:
```suggestion
    sometime after do_stop_slave_sql() succeeds
```

bnestere (approved):

PR #3784, opened 2025-01-23
MDEV-35169  ALTER TABLE...IMPORT TABLESPACE does not work with INDEX DESC

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35169*

## Description
Problem:
=======
- Import tablespace fails to check the index fields descending property while matching the schema given in cfg file with the table schema.

Fix:
===
- Introduce the new version format for the cfg file which has the descending bool information for all index fields

row_import_cfg_read_index_fields(): Read the field descending property from cfg file depends on the config version

row_quiesce_write_index_fields(): Write the descending property of the index fields into cfg file

## How can this PR be tested?
./mtr innodb.import_cfg --mem
Basically which does create old version cfg file and import tablespace with old cfg file

Taken the cfg file from 10.11 (with patch) and tried to run it in 10.6, tested how it behaves:

```
primary key with descending field in 10.6
=========================================
MariaDB [test]> create table t1_pk_desc(f1 int not null, f2 int not null, primary key(f1 desc))engine=innodb;
Query OK, 0 rows affected (0.015 sec)

MariaDB [test]> alter table t1_pk_desc discard tablespace;
Query OK, 0 rows affected (0.018 sec)


MariaDB [test]> alter table t1_pk_desc import tablespace;
Query OK, 0 rows affected (0.030 sec)

MariaDB [test]> check table t1_pk_desc;
+-----------------+-------+----------+---------------------------------------------------+
| Table           | Op    | Msg_type | Msg_text                                          |
+-----------------+-------+----------+---------------------------------------------------+
| test.t1_pk_desc | check | Warning  | InnoDB: The B-tree of index PRIMARY is corrupted. |
| test.t1_pk_desc | check | error    | Corrupt                                           |
+-----------------+-------+----------+---------------------------------------------------+
2 rows in set (0.013 sec)

MariaDB [test]> alter table t1_pk_desc force;
ERROR 1177 (42000): Can't open table

Avoided check table and tried table rebuild directly:
===========================================
MariaDB [test]> alter table t1_pk_desc import tablespace;
Query OK, 0 rows affected (0.030 sec)

MariaDB [test]> alter table t1_pk_desc force;
ERROR 2013 (HY000): Lost connection to server during query

mysqld: /home/thiru/source_file/server/10.6/storage/innobase/btr/btr0bulk.cc:173: void PageBulk::insertPage(rec_t*, rec_offs*) [with format <anonymous> = PageBulk::DYNAMIC; rec_t = unsigned char; rec_offs = short unsigned int]: Assertion `cmp_rec_rec(rec, old_rec, offsets, old_offsets, m_index) > 0' failed

Tried copy algorithm
===================

MariaDB [test]> alter table t1_pk_desc force, algorithm=copy;
Query OK, 2 rows affected (0.043 sec)              
Records: 2  Duplicates: 0  Warnings: 0

MariaDB [test]> check table t1_pk_desc extended;
+-----------------+-------+----------+----------+
| Table           | Op    | Msg_type | Msg_text |
+-----------------+-------+----------+----------+
| test.t1_pk_desc | check | status   | OK       |
+-----------------+-------+----------+----------+
1 row in set (0.016 sec)

Table with secondary index of descending field:
===============================================

MariaDB [test]> create table t1_sec_desc(f1 int not null, f2 int not null, primary key(f1), index(f2 desc))engine=innodb;
Query OK, 0 rows affected (0.028 sec)

MariaDB [test]> alter table t1_sec_desc discard tablespace;
Query OK, 0 rows affected (0.019 sec)

MariaDB [test]> alter table t1_sec_desc import tablespace;
Query OK, 0 rows affected (0.026 sec)

MariaDB [test]> check table t1_sec_desc extended;
+------------------+-------+----------+----------------------------------------------+
| Table            | Op    | Msg_type | Msg_text                                     |
+------------------+-------+----------+----------------------------------------------+
| test.t1_sec_desc | check | Warning  | InnoDB: The B-tree of index f2 is corrupted. |
| test.t1_sec_desc | check | error    | Corrupt                                      |
+------------------+-------+----------+----------------------------------------------+
2 rows in set (0.011 sec)

MariaDB [test]> alter table t1_sec_desc drop index f2;
Query OK, 0 rows affected (0.026 sec)
Records: 0  Duplicates: 0  Warnings: 0

MariaDB [test]> alter table t1_sec_desc add index(f2 desc);
Query OK, 0 rows affected (0.025 sec)
Records: 0  Duplicates: 0  Warnings: 0

MariaDB [test]> select * from t1_sec_desc use index(f2);
+----+----+
| f1 | f2 |
+----+----+
|  4 |  1 |
|  2 |  3 |
+----+----+
2 rows in set (0.006 sec)

Table with no desc index:
=========================

MariaDB [test]> create table t1(f1 int not null, f2 int not null, primary key (f1), index(f2))engine=innodb;
Query OK, 0 rows affected (0.021 sec)

MariaDB [test]> alter table t1 discard tablespace;
Query OK, 0 rows affected (0.020 sec)

MariaDB [test]> alter table t1 import tablespace;
Query OK, 0 rows affected (0.024 sec)

MariaDB [test]> check table t1 extended;
+---------+-------+----------+----------+
| Table   | Op    | Msg_type | Msg_text |
+---------+-------+----------+----------+
| test.t1 | check | status   | OK       |
+---------+-------+----------+----------+
1 row in set (0.011 sec)

MariaDB [test]> select * from t1;
+----+----+
| f1 | f2 |
+----+----+
|  1 |  3 |
|  2 |  3 |
+----+----+
2 rows in set (0.002 sec)
```

## PR quality check

### DIFF
--- a/storage/innobase/row/row0import.cc
+++ b/storage/innobase/row/row0import.cc
@@ -1213,6 +1213,16 @@ row_import::match_index_columns(
 
 			err = DB_ERROR;
 		}
+
+		if (cfg_field->descending != field->descending) {
+			ib_errf(thd, IB_LOG_LEVEL_ERROR,
+				ER_TABLE_SCHEMA_MISMATCH,
+				"Index %s field %s is %s which does "
+				"not match with .cfg file",
+				index->name(), field->name(),
+				field->descending ? "DESC" : "ASC");
+			err = DB_ERROR;
+		}
 	}
 
 	return(err);
@@ -2557,7 +2567,11 @@ row_import_cfg_read_index_fields(
 		field->prefix_len = mach_read_from_4(ptr) & ((1U << 12) - 1);
 		ptr += sizeof(ib_uint32_t);
 
-		field->fixed_len = mach_read_from_4(ptr) & ((1U << 10) - 1);
+		uint32_t fixed_len = mach_read_from_4(ptr);
+
+		field->descending = bool(fixed_len >> 31);
+
+		field->fixed_len = fixed_len & ((1U << 10) - 1);
 		ptr += sizeof(ib_uint32_t);
 
 		/* Include the NUL byte in the length. */
--- a/storage/innobase/row/row0quiesce.cc
+++ b/storage/innobase/row/row0quiesce.cc
@@ -47,45 +47,38 @@ row_quiesce_write_index_fields(
 	FILE*			file,	/*!< in: file to write to */
 	THD*			thd)	/*!< in/out: session */
 {
-	byte			row[sizeof(ib_uint32_t) * 2];
+	byte			row[sizeof(ib_uint32_t) * 3];
 
 	for (ulint i = 0; i < index->n_fields; ++i) {
 		byte*			ptr = row;
 		const dict_field_t*	field = &index->fields[i];
 
 		mach_write_to_4(ptr, field->prefix_len);
-		ptr += sizeof(ib_uint32_t);
+		ptr += 4;
 
-		mach_write_to_4(ptr, field->fixed_len);
-
-		DBUG_EXECUTE_IF("ib_export_io_write_failure_9",
-				close(fileno(file)););
-
-		if (fwrite(row, 1, sizeof(row), file) != sizeof(row)) {
-
-			ib_senderrf(
-				thd, IB_LOG_LEVEL_WARN, ER_IO_WRITE_ERROR,
-				(ulong) errno, strerror(errno),
-				"while writing index fields.");
-
-			return(DB_IO_ERROR);
-		}
+		/* Since maximum fixed length can be
+		DICT_ANTELOPE_MAX_INDEX_COL_LEN, InnoDB
+		can use the 0th bit to store the
+		field descending information */
+		mach_write_to_4(ptr, field->fixed_len
+				     | uint32_t(field->descending) << 31);
+		ptr += 4;
 
 		const char* field_name = field->name ? field->name : "";
 		/* Include the NUL byte in the length. */
-		ib_uint32_t	len = static_cast<ib_uint32_t>(strlen(field_name) + 1);
-		mach_write_to_4(row, len);
+		uint32_t	len = uint32_t(strlen(field_name) + 1);
+		mach_write_to_4(ptr, len);
 
 		DBUG_EXECUTE_IF("ib_export_io_write_failure_10",
 				close(fileno(file)););
 
-		if (fwrite(row, 1,  sizeof(len), file) != sizeof(len)
+		if (fwrite(row, 1, sizeof(row), file) != sizeof(row)
 		    || fwrite(field_name, 1, len, file) != len) {
 
 			ib_senderrf(
 				thd, IB_LOG_LEVEL_WARN, ER_IO_WRITE_ERROR,
 				(ulong) errno, strerror(errno),
-				"while writing index column.");
+				"while writing index fields.");
 
 			return(DB_IO_ERROR);
 		}
--- a/mysql-test/suite/innodb/r/import_cfg.result
+++ b/mysql-test/suite/innodb/r/import_cfg.result
@@ -1,0 +1,67 @@
+#
+#  MDEV-35169 ALTER TABLE...IMPORT TABLESPACE does not
+#		work with INDEX DESC
+#
+# prepare cfg for primary key with desc column
+create table t1 (pk int, a int, primary key(pk desc)) engine=InnoDB;
+insert into t1 values (1,10),(2,20),(3,15);
+flush table t1 for export;
+unlock tables;
+drop table t1;
+# prepare cfg for secondary index with desc column
+create table t1 (pk int primary key, a int,key(a desc)) engine=InnoDB;
+insert into t1 values (1,10),(2,20),(3,15);
+flush table t1 for export;
+unlock tables;
+drop table t1;
+# prepare cfg for secondary index with ascending column
+create table t1 (pk int primary key, a int, key(a)) engine=InnoDB;
+insert into t1 values (1,10),(2,20),(3,15);
+flush table t1 for export;
+unlock tables;
+drop table t1;
+# Import desc tablespace into desc frm
+# Import into table with desc primary key column
+create table t1 (pk int, a int, primary key(pk desc)) engine=InnoDB;
+alter table t1 discard tablespace;
+alter table t1 import tablespace;
+check table t1 extended;
+Table	Op	Msg_type	Msg_text
+test.t1	check	status	OK
+drop table t1;
+# Import into table with desc secondary index
+create table t1 (pk int primary key, a int, key(a desc))engine=InnoDB;
+alter table t1 discard tablespace;
+alter table t1 import tablespace;
+check table t1 extended;
+Table	Op	Msg_type	Msg_text
+test.t1	check	status	OK
+drop table t1;
+# Import asc tablespace into desc frm
+create table t1 (pk int primary key, a int, key(a desc))engine=InnoDB;
+alter table t1 discard tablespace;
+alter table t1 import tablespace;
+ERROR HY000: Schema mismatch (Index a field a is DESC which does not match with .cfg file)
+check table t1 extended;
+Table	Op	Msg_type	Msg_text
+test.t1	check	Error	Tablespace has been discarded for table `t1`
+test.t1	check	error	Corrupt
+drop table t1;
+# Import desc tablespace into asc frm
+create table t1 (pk int primary key, a int, key(a)) engine=InnoDB;
+alter table t1 discard tablespace;
+alter table t1 import tablespace;
+ERROR HY000: Schema mismatch (Index a field a is ASC which does not match with .cfg file)
+check table t1 extended;
+Table	Op	Msg_type	Msg_text
+test.t1	check	Error	Tablespace has been discarded for table `t1`
+test.t1	check	error	Corrupt
+drop table t1;
+# Import asc tablespace into asc frm
+create table t1 (pk int primary key, a int, key(a)) engine=InnoDB;
+alter table t1 discard tablespace;
+alter table t1 import tablespace;
+check table t1 extended;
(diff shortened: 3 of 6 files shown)


storage/innobase/row/row0import.cc
```diff
@@ -2560,6 +2583,14 @@ row_import_cfg_read_index_fields(
 		field->fixed_len = mach_read_from_4(ptr) & ((1U << 10) - 1);
 		ptr += sizeof(ib_uint32_t);
 
+		if (version == IB_EXPORT_CFG_VERSION_V2) {
+			field->descending = (mach_read_from_4(ptr) == 1);
+			if (fread(row, 1, 4, file) != 4) {
+				goto file_read_error;
+			}
+			ptr = row;
+		}
```
dr-m, line 2592: Can we avoid allocating 4 more bytes for this 1 flag? In any case, the `==` should be replaced with `&` so that we could freely repurpose the remaining flags later.

storage/innobase/row/row0import.cc
```diff
@@ -2560,6 +2583,14 @@ row_import_cfg_read_index_fields(
 		field->fixed_len = mach_read_from_4(ptr) & ((1U << 10) - 1);
 		ptr += sizeof(ib_uint32_t);
```
dr-m, line 2584: How are the 32 bits of `ptr` currently allocated? (Can we avoid repeated calls to `mach_read_from_4(ptr)` too?) Could we repurpose one of those bits for "descending field"?

I see that in `row_quiesce_write_index_fields()` we are wasting 32 bits for `field->prefix_len` and `field->fixed_len`. It seems that we could simply repurpose the most significant bit (`1U << 31`) of either of those to mean "descending field". Try to pick one that when it is set, it would cause the existing version-1 format parser to fail nicely. We could also repurpose the most significant bit in the 32 bits that we write for the field name length. That length should never exceed `NAME_LEN` (192).

dr-m (commented):
Can we try to change the file format as little as possible?
I would modify `row_quiesce_write_indexes()` so that when descending index fields are present, we would set the most significant bit in the 32 bits that we use for encoding one of `type`, `n_uniq`, `n_nullable`, `n_fields`. The last one would seem to be an obvious choice, due to the following in `row_import_read_indexes()`:
```c++
	} else if (cfg->m_n_indexes > 1024) {
		// FIXME: What is the upper limit? */
		ib_errf(thd, IB_LOG_LEVEL_ERROR, ER_IO_READ_ERROR,
			"Number of indexes in meta-data file is too high: "
			ULINTPF, cfg->m_n_indexes);
		cfg->m_n_indexes = 0;

		return(DB_CORRUPTION);
	}
```
This would give an easy to recognize error message and still allow the file to be slightly edited so that it can be imported.

We do not even seem to need the above; we could simply store the "descending field" flag in the 32 bits that store the length of the field name:
```c++
		/* The NUL byte is included in the name length. */
		ulint	len = mach_read_from_4(ptr);

		if (len > OS_FILE_MAX_PATH) {
			ib_errf(thd, IB_LOG_LEVEL_ERROR,
				ER_INNODB_INDEX_CORRUPT,
				"Index name length (" ULINTPF ") is too long, "
				"the meta-data is corrupt", len);

			return(DB_CORRUPTION);
		}
```

storage/innobase/row/row0import.cc
```diff
@@ -1213,6 +1215,21 @@ row_import::match_index_columns(
 
 			err = DB_ERROR;
 		}
+
+		if (cfg_field->descending != field->descending) {
+			ib_errf(thd, IB_LOG_LEVEL_ERROR,
+				ER_TABLE_SCHEMA_MISMATCH,
+				"Index %s field %s is %s which does "
+				"not match metdata file which is %s",
```
dr-m, line 1221: I think that it suffices to display the ascending/descending only once. For clarify, it may be better to say .cfg file.

storage/innobase/row/row0quiesce.cc
```diff
@@ -57,30 +57,37 @@ row_quiesce_write_index_fields(
...
-		/* Include the NUL byte in the length. */
-		ib_uint32_t	len = static_cast<ib_uint32_t>(strlen(field_name) + 1);
-		mach_write_to_4(row, len);
+		/** Since maximum field name length is
+		OS_FILE_MAX_PATH, InnoDB can use 0th bit of
+		the field name length field
+	        to store the descending field property */
+		mach_write_to_4(ptr, field->descending
+				     ? name_len | 1 << 31
+				     : name_len);
```
dr-m, line 83: This would seem to be able to trigger a large memory allocation in `row_import_cfg_read_index_fields()` in earlier versions of MariaDB:
```c++
		ulint	len = mach_read_from_4(ptr);

		byte*	name = UT_NEW_ARRAY_NOKEY(byte, len);
// …
		if (name == NULL) {
			return(DB_OUT_OF_MEMORY);
		}
```
Based on my reading of `ut_allocate::new_array()`, this should never throw `std::bad_alloc` but actually return `name=nullptr` when running out of memory. In a 32-bit environment it is almost guaranteed, but in 64-bit environments this could cause an allocation of over 2 GiB, followed by an attempt to fill that buffer. As far as I understand, `row_import_cfg_read_string()` should return `DB_IO_ERROR` upon reaching the end of the file.

When we are repurposing any fields, we should test and document what would happen in an earlier version.

I believe that the least intrusive way of implementing this would be to repurpose a bit of the 32 bits allocated for `field->fixed_len` or `field->prefix_len`. These bits are always written as 0. The function `row_quiesce_write_index_fields()` was introduced in mysql/mysql-server@90d9086d6adb35ecf8af5a30b4f050dfcbe00dad (MySQL 5.6.6). Already back then, the `prefix_len` and `fixed_len` were defined as bit-fields of `dict_field_struct`, which means that only 12 or 10 bits of these 32-bit fields were ever used, and the unused bits were always written as 0, both in MySQL and MariaDB.

If we simply repurpose one of those previously-clear bits for the `field->descending` flag, then older versions of MariaDB should continue to import the data file. This may be desirable so that the data can be accessed. The table could be "repaired" (the descending fields converted to ascending) by executing `OPTIMIZE TABLE`.

Can you please implement this idea and test and report how it would be treated in MariaDB Server 10.6 or unpatched 10.11?

By the way, we should always prefer arithmetic operations to conditions:
```c++
name_len | uint32_t(field->descending) << 31;
```

storage/innobase/row/row0quiesce.cc
```diff
@@ -56,30 +56,24 @@ row_quiesce_write_index_fields(
...
+		mach_write_to_4(ptr, field->fixed_len
+				     | uint32_t(field->descending) << 31);
+		ptr += sizeof(ib_uint32_t);
 
 		const char* field_name = field->name ? field->name : "";
 		/* Include the NUL byte in the length. */
-		ib_uint32_t	len = static_cast<ib_uint32_t>(strlen(field_name) + 1);
-		mach_write_to_4(row, len);
+		ib_uint32_t	len =
+		  static_cast<ib_uint32_t>(strlen(field_name) + 1);
```
dr-m, line 70: Let’s use the C++11 standard type `uint32_t` and the constructor cast instead of `static_cast`, to avoid overly long lines.

storage/innobase/row/row0quiesce.cc
```diff
@@ -56,30 +56,24 @@ row_quiesce_write_index_fields(
...
-
-			return(DB_IO_ERROR);
-		}
+		/* Since maximum fixed length can be
+		DICT_ANTELOPE_MAX_INDEX_COL_LEN, InnoDB
+		can use the 0th bit to store the
+		field descending information */
+		mach_write_to_4(ptr, field->fixed_len
+				     | uint32_t(field->descending) << 31);
+		ptr += sizeof(ib_uint32_t);
```
dr-m, line 65: Let’s use a standard type name, or just replace the `sizeof` expression with `4`.

storage/innobase/row/row0quiesce.cc
```diff
@@ -56,30 +56,24 @@ row_quiesce_write_index_fields(
...
-
-		if (fwrite(row, 1, sizeof(row), file) != sizeof(row)) {
-
-			ib_senderrf(
-				thd, IB_LOG_LEVEL_WARN, ER_IO_WRITE_ERROR,
-				(ulong) errno, strerror(errno),
-				"while writing index fields.");
-
-			return(DB_IO_ERROR);
-		}
```
dr-m, line 72: The commit message had better mention this simplification. The surviving error message later in this function had better say "index fields" rather than "index column".

storage/innobase/row/row0import.cc
```diff
@@ -1213,6 +1213,18 @@ row_import::match_index_columns(
...
+
+		if (cfg_field->descending != field->descending) {
+			ib_errf(thd, IB_LOG_LEVEL_ERROR,
+				ER_TABLE_SCHEMA_MISMATCH,
+				"Index %s field %s is %s which does "
+				"not match wth .cfg file",
+				index->name(), field->name(),
+				field->descending
+				  ? "descending"
+				  : "ascending");
```
dr-m, line 1225: I would remove the misspelled word "with" and use the SQL keywords `DESC` and `ASC`.

dr-m (commented):
Please update the "Testing done" section with a test that involves an attempt to import `.cfg` files from an export after this fix, to both MariaDB Server 10.11 and 10.6. I’d like to see three cases:
1. No indexes declared with `DESC`
2. At least one secondary indexes declared with at least one `DESC` column
3. `PRIMARY KEY` includes a `DESC` column

For 10.6, please include the SQL statements that need to be executed to make the imported tables pass `CHECK TABLE`.

dr-m (approved):
Thank you. For the debug assertion failure on `ALTER TABLE…FORCE, ALGORITHM=INPLACE` in 10.6 in case of `ASC`/`DESC` mismatch , please file a bug and note the number here.

Thirunarayanan:
https://jira.mariadb.org/browse/MDEV-35956

PR #3786, opened 2025-01-23
MDEV-36366 - cmake(libfmt): use find_package to find libfmt

chenrui333:
## Description
TODO: fill description here

This patch improves libfmt detection by calling find_package(fmt QUIET) first when WITH_LIBFMT is set to system or auto. If a system libfmt is located successfully, we skip the custom source-compile snippet. Otherwise, we proceed with the old detection logic (compile test) and finally fall back to the bundled library if needed. This change makes it easier for package managers (e.g., Homebrew) or system environments that already have a modern fmt installed to integrate with MariaDB, without having to override CMake variables manually.

Before this patch, the logic always set CMAKE_REQUIRED_INCLUDES to the bundled copy’s include directory and ran a snippet test, failing if the user had removed or disabled the bundled directory. Now, if a system installation is discoverable via CMake config files, MariaDB can be built cleanly against that library.

## Release Notes

- Enhancement: MariaDB can now detect and use system-provided libfmt more reliably when WITH_LIBFMT is system or auto. If find_package(fmt) succeeds, the bundled fallback is skipped, ensuring smoother packaging in environments providing up-to-date versions of libfmt.

No system variables or status variables were changed. No existing behavior changed aside from this improvement in build-time detection of libfmt.

## How can this PR be tested?

homebrew PRs

## PR quality check

---

relates to:
- https://github.com/Homebrew/homebrew-core/pull/205321
- https://github.com/Homebrew/homebrew-core/pull/205322

### DIFF
--- a/cmake/libfmt.cmake
+++ b/cmake/libfmt.cmake
@@ -1,5 +1,6 @@
 INCLUDE (CheckCXXSourceRuns)
 INCLUDE (ExternalProject)
+FIND_PACKAGE(fmt QUIET)
 
 SET(WITH_LIBFMT "auto" CACHE STRING
    "Which libfmt to use (possible values are 'bundled', 'system', or 'auto')")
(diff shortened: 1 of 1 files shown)


grooverdan:
Currently because libfmt changed the API by removing the thousands separator (https://github.com/fmtlib/fmt/issues/4284) and we haven't written a compatibility for the newer libfmt yet.

ref: https://jira.mariadb.org/browse/MDEV-35746

ottok:
FYI, I Debian we have a patched libfmt 10.1 to be compatible with MariaDB:
* https://salsa.debian.org/mariadb-team/mariadb-server/-/commit/a67b79d5d51c3d4cc9b87debab8f1f1ba60e0587
* https://salsa.debian.org/zhsj/fmtlib/-/commit/dad501f3883dd023cbcfe604bc81617de81a020c

vuvova:
@ottok nice, thanks. Because currently we don't check for libfmt version, but verify whether `{:L}` formats the number as expected, it means mariadb will use your patched libfmt just fine.

grooverdan:
with fmt 11.0.0+ (homebrew [11.0.3](https://github.com/Homebrew/homebrew-core/blob/master/Formula/f/fmt.rb)) a [fix](https://github.com/fmtlib/fmt/commit/44c3fe1ebb466ab5c296e1a1a6991c7c7b51b72e) is included so this shouldn't be required?

The explicit macro test is there to ensure the fmt is correct so explicit trust is system version wasn't the goal.

vuvova:
@chenrui333 , do I understand correctly, that you need this to make MariaDB build in homebrew? And "improve libfmt detection" means that without this patch cmake cannot detect system libfmt in homebrew?

chenrui333:
- close via https://github.com/Homebrew/homebrew-core/pull/225819

PR #3788, opened 2025-01-23
MDEV-31334: Consider dates ending in 'T' as malformed

FarihaIS:
## Description

MariaDB 10.6 currently accepts dates that end in 'T' without a warning or error: for example '`insert into t1 values ("2003-02-05T")`'. This is an issue as ISO 8601 standards state that a 'T' should be followed by the time of day. Dates that terminate in 'T' should receive the same error in strict mode and warning otherwise as dates that terminate in any other (invalid) character. Commit includes bug fix and tests for dates ending in 'T'.

## How can this PR be tested?

Execute the main suite in mysql-test-run. This commit adds a test in that suite.

## Backward compatibility

This is a new feature introducing stricter date formatting that will consider dates ending in 'T' as malformed and affect backwards compatibility.

## Copyright

### DIFF
--- a/mysql-test/main/func_extract.result
+++ b/mysql-test/main/func_extract.result
@@ -368,6 +368,8 @@ EXTRACT(DAY FROM '01-02-03')
 SELECT EXTRACT(DAY FROM '24:02:03T');
 EXTRACT(DAY FROM '24:02:03T')
 3
+Warnings:
+Warning	1292	Truncated incorrect date value: '24:02:03T'
 SELECT EXTRACT(DAY FROM '24-02-03');
 EXTRACT(DAY FROM '24-02-03')
 3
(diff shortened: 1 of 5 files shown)


cvicentiu (changes requested):

cvicentiu:
@FarihaIS Thank you for the contribution!  :+1: 

I've done a preliminary review and I think you should first try to clarify in the MDEV what the exact specification is.

How should SELECT FUNCTION(<malformed_date>) should work. How it should work for INSERT. How it should be have with STRICT mode vs not strict mode.

Then you should check your patch again and see if it fulfils (and tests) those scenarios.

I'm sure we can get this through with a bit more work. :)

mysql-test/main/func_extract.result
```diff
@@ -380,39 +382,69 @@ EXTRACT(DAY FROM '11111')
...
-0	0
+9	9
+Warnings:
+Warning	1292	Truncated incorrect time value: '2001:01:01T'
+Warning	1292	Truncated incorrect time value: '2001:01:01T '
 # This still parses as DATE and returns NULL (without trying TIME)
 SELECT EXTRACT(DAY FROM '100000:02:03T');
 EXTRACT(DAY FROM '100000:02:03T')
-NULL
+4166
```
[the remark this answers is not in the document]
FarihaIS: Just pushed changes that keep the original behavior here! The intended functionality is an error in strict mode if a date ends in T and a warning in other modes, the same way it would behave if a date ended in a different letter like 'Q' for example.

FarihaIS:
@abarkov I updated the PR last week to reflect the changes requested and have also left a comment in the [original Jira issue]([MDEV-31334](https://jira.mariadb.org/browse/MDEV-31334)
MariaDB should not accept dates with 'T' on the end), please let me know if you have any concerns, thank you.

cvicentiu (approved):
I am ok with this.

I would like a comment before the if clause referencing ISO_8601.
Document is available here:
https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf
```
/*
  According to ISO_8601 - 2016
  "
  The character [T] shall be used as time designator to indicate the start of the	 
  representation of the time of day  component in these expressions.
  "

  That means that after T there *must* be a time component.
*/
```
I will let @abarkov to offer his final review as he's had more experience with backwards compatibility and standard related things.

FarihaIS:
> I am ok with this.
> 
> I would like a comment before the if clause referencing ISO_8601. Document is available here: https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf
> 
> ```
> /*
>   According to ISO_8601 - 2016
>   "
>   The character [T] shall be used as time designator to indicate the start of the	 
>   representation of the time of day  component in these expressions.
>   "
> 
>   That means that after T there *must* be a time component.
> */
> ```
> 
> I will let @abarkov to offer his final review as he's had more experience with backwards compatibility and standard related things.

Thank you, I've added the comment you requested right above the if check.

mysql-test/main/strict.test
```diff
@@ -48,6 +48,10 @@ INSERT INTO t1 VALUES('2004-0-30');
 INSERT INTO t1 VALUES('2004-01-04'),('2004-0-31'),('2004-01-05');
 INSERT INTO t1 VALUES('0000-00-00');
 INSERT IGNORE INTO t1 VALUES('2004-0-29');
+--error 1292
+INSERT INTO t1 VALUES('2001-01-01W');
+--error 1292
+INSERT INTO t1 VALUES('2001-01-01T');
 set @@sql_mode='STRICT_ALL_TABLES,NO_ZERO_DATE';
```
abarkov, line 55: For full coverage, it would nice to test inserts with:
- sql_mode='STRICT_TRANS_TABLES';
- sql_mode='STRICT_ALL_TABLES';
- sql_mode='STRICT_ALL_TABLES, STRICT_TRANS_TABLES';
- sql_mode='';
with two engines:
- ENGINE=InnoDB
- ENGINE=MyISAM

This gives 8 combinations.

FarihaIS: I've just added the tests you requested :)

svoj:
1. Behaviour change is approved 
2. Code change is approved
3. Test case change requested (which was implemented already)
4. Bar is happy if we take it over from now on

@cvicentiu in other words the ball is in your hands: feel free to review requested test case change and to get it merged.

vuvova:
I've reassigned MDEV accordingly

cvicentiu:
Merged into main, will be part of 12.0.1 release. Thank you for the contribution!

PR #3790, opened 2025-01-27
MDEV-35937  InnoDB: Failing assertion: sym_node->table != NULL in pars_retrieve_table_def

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35937*

## Description
- InnoDB tries to fetch the metadata about fulltext index for discarded tablespace and fails for discarded tablespace.
In i_s_fts_config_fill(), i_s_fts_index_cache_fill() and i_s_fts_index_table_fill(), InnoDB needs to check whether
the tablespace is discarded or accessable before accessing the fulltext auxiliary table. This issue doesn't exist in
10.6+ version due to MDEV-19445(commit https://github.com/MariaDB/server/commit/0aa2bc7a94d96397e2e061dacb8ed04a32c13fa0)

## How can this PR be tested?
./mtr innodb_fts.innodb_fts_misc_1

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb_fts/t/innodb_fts_misc_1.test
+++ b/mysql-test/suite/innodb_fts/t/innodb_fts_misc_1.test
@@ -960,3 +960,16 @@ CREATE FULLTEXT INDEX i ON t1 (f2);
 SELECT * FROM t1 WHERE match(f2) against("bbb");
 # Cleanup
 DROP TABLE t1;
+
+--echo #
+--echo # MDEV-35937 sym_node->table != NULL in pars_retrieve_table_def
+--echo #
+CREATE TABLE t1 (a TEXT ,FULLTEXT (a)) ENGINE=INNODB;
+ALTER TABLE t1 DISCARD TABLESPACE;
+SET @save_ft = @@GLOBAL.innodb_ft_aux_table;
+SET GLOBAL innodb_ft_aux_table='test/t1';
+SELECT * FROM information_schema.innodb_ft_config;
+SELECT * FROM information_schema.innodb_ft_index_cache;
+SELECT * FROM information_schema.innodb_ft_index_table;
+DROP TABLE t1;
+SET GLOBAL innodb_ft_aux_table=@save_ft;
--- a/storage/innobase/handler/i_s.cc
+++ b/storage/innobase/handler/i_s.cc
@@ -2367,7 +2367,7 @@ i_s_fts_deleted_generic_fill(
 		rw_lock_s_unlock(&dict_sys.latch);
 		DBUG_RETURN(0);
 	} else if (!dict_table_has_fts_index(user_table)
-		   || !user_table->is_readable()) {
+		   || !user_table->is_accessible()) {
 		dict_table_close(user_table, FALSE, FALSE);
 		rw_lock_s_unlock(&dict_sys.latch);
 		DBUG_RETURN(0);
@@ -2744,7 +2744,8 @@ no_fts:
 		DBUG_RETURN(0);
 	}
 
-	if (!user_table->fts || !user_table->fts->cache) {
+	if (!user_table->fts || !user_table->fts->cache
+	    || !user_table->is_accessible()) {
 		dict_table_close(user_table, FALSE, FALSE);
 		goto no_fts;
 	}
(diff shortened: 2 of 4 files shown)


mysql-test/suite/innodb_fts/t/innodb_fts_misc_1.test
```diff
@@ -960,3 +960,13 @@ CREATE FULLTEXT INDEX i ON t1 (f2);
...
+
+--echo #
+--echo # MDEV-35937 sym_node->table != NULL in pars_retrieve_table_def
+--echo #
+CREATE TABLE t1 (a TEXT ,FULLTEXT (a)) ENGINE=INNODB;
+ALTER TABLE t1 DISCARD TABLESPACE;
+SET GLOBAL innodb_ft_aux_table='test/t1';
+SELECT * FROM information_schema.innodb_ft_config;
+DROP TABLE t1;
+SET GLOBAL innodb_ft_aux_table=DEFAULT;
```
dr-m, line 972: It would be better to assign and restore the old value, in case someone is running `./mtr --mysqld=--innodb-ft-aux-table=foo/bar`.
```sql
SET @save_ft = @@GLOBAL.innodb_ft_aux_table;
-- …
SET GLOBAL innodb_ft_aux_table=@save_ft;
```

dr-m (commented):
I see that 0024524d542bffcadc3359543511c9b62d4742d7 had fixed this for another view. Shouldn’t also `i_s_fts_index_cache_fill()` and `i_s_fts_index_table_fill()` be fixed?

mysql-test/suite/innodb_fts/t/innodb_fts_misc_1.test
```diff
@@ -960,3 +960,16 @@ CREATE FULLTEXT INDEX i ON t1 (f2);
...
+--echo #
+--echo # MDEV-35937 sym_node->table != NULL in pars_retrieve_table_def
+--echo #
+CREATE TABLE t1 (a TEXT ,FULLTEXT (a)) ENGINE=INNODB;
+ALTER TABLE t1 DISCARD TABLESPACE;
+SET @save_ft = @@GLOBAL.innodb_ft_aux_table;
+SET GLOBAL innodb_ft_aux_table='test/t1';
+SELECT * FROM information_schema.innodb_ft_config;
+SELECT * FROM information_schema.innodb_ft_index_cache;
+SELECT * FROM information_schema.innodb_ft_index_table;
```
dr-m, line 973: I can’t reproduce any crash on 10.6 even if I move the `ALTER TABLE t1 DISCARD TABLESPACE` after the `SET GLOBAL`. Do you have any explanation why it might be so? Maybe a comment in the test would be useful.

storage/innobase/handler/i_s.cc
```diff
@@ -2744,7 +2744,8 @@ i_s_fts_index_cache_fill(
 		DBUG_RETURN(0);
 	}
 
-	if (!user_table->fts || !user_table->fts->cache) {
+	if (!user_table->fts || !user_table->fts->cache
+	    || !user_table->is_readable()) {
 		dict_table_close(user_table, FALSE, FALSE);
 		goto no_fts;
 	}
```
dr-m, line 2751: Would `is_accessible()` be a better choice than `is_readable()`? Which checks are being performed in `dict_table_open_on_id()` already?

dr-m (commented):
I do not seem to fully understand the failure scenario. Is this bug specific to 10.5 only? If that is the case, the commit message needs to make it clear that this fix must be null-merged to 10.6.

mysql-test/suite/innodb_fts/t/innodb_fts_misc_1.test
Thirunarayanan: This is being fixed in 10.6:
```
commit 0aa2bc7a94d96397e2e061dacb8ed04a32c13fa0
Author: Marko Mäkelä <marko.makela@mariadb.com>
Date:   Thu Jul 29 16:41:58 2021 +0300

    MDEV-19445/MDEV-16678 fixup: Acquire proper MDL on innodb_ft_aux_table
    
    In commit 2647fd101db922b1e6c7363124adfc03e81ab8a0 (MDEV-19445)
    we fixed a race condition around the INFORMATION_SCHEMA tables
    that access the table identified by the global variable
    innodb_ft_aux_table. Thanks to MDEV-16678 we could fix it
    even better by using MDL instead of the InnoDB dict_sys latches.

```

dict_acquire_mdl_shared() does check whether the table is accessed or not.

PR #3793, opened 2025-01-27
MDEV-34591  InnoDB fails to rename index from lowercase to uppercase

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-34591*

## Description
- InnoDB fails to rename the index when there is a case conversion. Problem is that server fails to fill the inplace alter information about the renaming of index before sending it to InnoDB.

fill_alter_inplace_info(): Should check strcmp instead of lex_string_cmp() for comparing the index name.

## How can this PR be tested?
./mtr innodb.innodb-index

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/innodb-index.test
+++ b/mysql-test/suite/innodb/t/innodb-index.test
@@ -1193,6 +1193,14 @@ SHOW CREATE TABLE t;
 --disable_prepare_warnings
 DROP TABLE t;
 
+# Rename the index from lowercase to uppercase
+CREATE TABLE t1 (a INT,b INT,KEY(b)) ENGINE=INNODB;
+ALTER TABLE t1 RENAME KEY b TO B;
+SHOW CREATE TABLE t1;
+SELECT name FROM INFORMATION_SCHEMA.INNODB_SYS_INDEXES WHERE table_id IN (SELECT table_id FROM INFORMATION_SCHEMA.INNODB_SYS_TABLES WHERE NAME="test/t1");
+CHECK TABLE t1;
+DROP TABLE t1;
+
 --disable_query_log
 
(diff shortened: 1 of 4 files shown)


mysql-test/suite/innodb/t/innodb-index.test
```diff
@@ -1193,6 +1193,14 @@ SHOW CREATE TABLE t;
 --disable_prepare_warnings
 DROP TABLE t;
 
+# Rename the index from lowercase to uppercase
+CREATE TABLE t1 (a INT,b INT,KEY(b)) ENGINE=INNODB;
+ALTER TABLE t1 RENAME KEY b TO B;
+SHOW CREATE TABLE t1;
+SELECT name FROM INFORMATION_SCHEMA.INNODB_SYS_INDEXES WHERE table_id IN (SELECT table_id FROM INFORMATION_SCHEMA.INNODB_SYS_TABLES WHERE NAME="test/t1");
```
dr-m, line 1200: I think that we need to test also `ADD INDEX` and `DROP INDEX` where only the case of the name differs. I fear that we may cause a regression there, and the expectation should be that `DROP INDEX B` succeeds or `CREATE INDEX B` fails, no matter whether the key is named `b` or `B` in the data dictionary.
Case insensitive or sensitive comparisons for identifier names are something that should have been clarified in MySQL WL#922, which apparently never was implemented. That ticket is private, but a public comment in [MySQL Bug#35520](https://bugs.mysql.com/bug.php?id=35520) mentions it.

dr-m:
The actual bug title is [MDEV-34951](https://jira.mariadb.org/browse/MDEV-34951). Please fix that in the commit message and in the description. I’m afraid that the misleading branch name must be unchanged, or otherwise GitHub will close this pull request with no possibiilty to reopen it.

PR #3794, opened 2025-01-27
MDEV-35964 fix cast warnings on mac in mysql_file.h

DaveGosselin-MariaDB:
Cast create_flags to mode_t when passing to my_create and my_create_with_symlink


include/mysql/psi/mysql_file.h
```diff
@@ -1029,13 +1029,13 @@ inline_mysql_file_create(
   if (psi_likely(locker != NULL))
   {
     PSI_FILE_CALL(start_file_open_wait)(locker, src_file, src_line);
-    file= my_create(filename, create_flags, access_flags, myFlags);
+    file= my_create(filename, (mode_t)create_flags, access_flags, myFlags);
```
grooverdan, line 1032: can `mode_t` be used in the function definitions? Most calls to these already just passed the flags by the POSIX OS defines. There's only a few variables `tmpfile_createflag` and `open_flags` (in Aria/MyISAM) that are incorrectly typed to int.

DaveGosselin-MariaDB: The compiler did not complain about any callsites after changing the prototypes.

grooverdan (approved):

PR #3566, opened 2024-10-07
MDEV-34391 Implement SET PATH statement

wong-github1:
- [x] *The Jira issue number for this PR is: MDEV-34391

## Description
Implement the SET PATH statement.

## Release Notes
Implement the SET PATH statement.

## How can this PR be tested?
mysql-test/mtr path_basic

## Basing the PR against the correct MariaDB version
- [x] *This is a new feature or a refactoring, and the PR is based against the `main` branch.*
- [ ] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [x] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #3796, opened 2025-01-28
MDEV-34391 SET PATH variable and resolution for stored routine invocations

iqbal-rsec:
- [x] *The Jira issue number for this PR is: MDEV-34391*

## Description
Implement the SET PATH statement. The grammar differs from the Jira ticket where the equal sign is required:
    ```
    SET PATH= 'schema1, schema2, schema3';
    ```
Schemas within the session PATH variable will be used to resolve unqualified and qualified routine invocations.
To determine whether or not a particular routine exists within a schema/database lookups has to be performed on the proc table, so performance is impacted. To alleviate this, the PATH is empty by default - routines will be resolved using the current database as it was originally done before this implementation.

Original PR is #3566, this PR has some additional changes that fixes/simplify the parser:
 - Refactored PATH parser code and fixed a bug where commas inside quotes are still treated as separator.
 - Moved path cache to system_variables for consistency with other variables
 - Changed path SET grammar to require `=` for consistency with other variables
 - Changed path cache to be a simple static array both for easier memory management. Only 16 schemas are supported per session, we do not want too many schemas in the path anyway since it will impact performance.
 - Add support for internal schemas (mariadb_schema, oracle_schema) in paths.
 - Set default PATH to empty
 - Make views invariant to current path
 - Invalidate sp caches when session path has been changed so that stored routines are recompiled to use the new paths
 - Disallow session path setting inside stored routines
 - Add tests
   - Ensure that SUPER permission is needed to set GLOBAL PATH
   - Tests were added where database name contains commas or double quotes
   - Test internal schemas in PATH
   - Test view invariance with respect to PATH

## Release Notes
New system variable ```path``` has been added.

- https://mariadb.com/kb/en/full-list-of-mariadb-options-system-and-status-variables/
- https://mariadb.com/kb/en/server-system-variables/

## How can this PR be tested?
```mysql-test/mtr path_basic path_resolution```

## PR quality check

### CODE REVIEW

File: mysql-test/main/path_resolution.result [hunk 1]
```diff
@@ -0,0 +1,3006 @@
...
+END;
+$$
+CREATE DATABASE deselect_db;
+USE deselect_db;
+DROP DATABASE deselect_db;
+SET PATH= 'test,test2';
+SELECT pack.func();
+pack.func()
+Invoke from Db: test, Pack: pack, Rt: func
+SET PATH= 'test2,test';
```
vuvova, line 230: This is a problem. By manipulating PATH one can make a stored routine to invoke some code it wasn't supposed to. That stored routine could have SECURITY=DEFINER, so you can trick it to invoke your code with somebody else's privileges.

I'm afraid, to fix it we should do what standard says, PATH must be a property of a stored routine, in other words, it has to be stored in a new column in mysql.proc, next to sql_mode. Also in mysql.event and in TRG files for triggers and frm files for views.

mysql-test/main/path_resolution.result
```diff
@@ -0,0 +1,3006 @@
...
+DROP FUNCTION func_prv;
+DROP PACKAGE BODY pack;
+DROP PACKAGE pack;
+SET @@path = DEFAULT;
+#
+# type 3.1
+# sql_mode: DEFAULT | procedure | is unqualified
+# curr db: X | path set: no
+#
+CREATE PROCEDURE proc() SELECT 'Invoke from Db: test, Rt: proc';
```
vuvova, line 626: may be write `'Invoke from Db: test, Rt: proc' as 'log:'` ?
very minor detail, really, don't change if you don't want to

mysql-test/main/path_resolution.result
```diff
@@ -0,0 +1,3006 @@
...
+DROP PROCEDURE proc_prv;
+DROP DATABASE test2;
+USE test;
+DROP PROCEDURE proc_prv;
+DROP PACKAGE BODY pack;
+DROP PACKAGE pack;
+SET @@path = DEFAULT;
+#
+# Resolution of routines in stored routines should be based on
+# the current path, not the path at the time of creation.
```
vuvova, line 2748: No, quite the opposite. As I already commented above, should be based on the path at the time of the creation.

mysql-test/main/path_resolution.result
```diff
@@ -0,0 +1,3006 @@
...
+DROP FUNCTION test.func;
+DROP DATABASE test2;
+DROP DATABASE test3;
+#
+# Internal schema (function_call_keyword:)
+#
+SET @@path= DEFAULT;
+SET sql_mode= DEFAULT;
+SELECT trim('     ');
+trim('     ')
```
vuvova, line 2779: This is tricky, could you move the internal schema handling into a separate PR, please?

mysql-test/suite/sys_vars/r/path_basic.result
```diff
@@ -0,0 +1,537 @@
...
+SET @@path=DEFAULT;
+#
+# path max limit (16 schemas)
+#
+SET PATH= 'a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p';
+SELECT @@path;
+@@path
+`a`,`b`,`c`,`d`,`e`,`f`,`g`,`h`,`i`,`j`,`k`,`l`,`m`,`n`,`o`,`p`
+SET PATH= 'a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q';
+ERROR 42000: Variable 'path' can't be set to the value of 'a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q'
```
vuvova, line 520: better use ER_VALUE_TOO_LONG here

mysql-test/suite/sys_vars/r/path_priv.result
```diff
@@ -0,0 +1,18 @@
+#
+# MDEV-34391: Check privilige for setting global path
+#
+CREATE USER priv_test;
+connect  conn1,localhost,priv_test,,;
+connection conn1;
+SET @@global.path= 'sys';
+ERROR 42000: Access denied; you need (at least one of) the SUPER privilege(s) for this operation
```
vuvova, line 8: I wouldn't even bother testing that. You didn't write any code related to privileges, it's common behavior for all global variables, not specific to @@path. No need to test it for every new variable, it's always the same.

mysql-test/suite/sys_vars/r/sysvars_server_embedded.result
```diff
@@ -2692,6 +2692,16 @@ NUMERIC_BLOCK_SIZE	NULL
 ENUM_VALUE_LIST	NULL
 READ_ONLY	NO
 COMMAND_LINE_ARGUMENT	REQUIRED
+VARIABLE_NAME	PATH
+VARIABLE_SCOPE	SESSION
+VARIABLE_TYPE	VARCHAR
+VARIABLE_COMMENT	SET PATH statement
```
vuvova, line 2698: Let's have a more descriptive name here. For example

> Comma-separated list of schema names that defines the search order for stored routines

sql/item_create.cc
```diff
@@ -2838,27 +2839,44 @@ Item*
...
-      Reusing ER_SP_DOES_NOT_EXIST have a message consistent with
-      the case when a default database exist, see Create_sp_func::create().
+      Attempt to resolve the db name from the path pariable
     */
-    my_error(ER_SP_DOES_NOT_EXIST, MYF(0),
-             "FUNCTION", name->str);
-    return NULL;
+    if (thd->variables.path.find_db_unqualified(thd, *name,
+                                                &sp_handler_function,
+                                                &db, NULL))
```
vuvova, line 2851: No, sorry, this is wrong. We cannot access/read tables directly in the parser, it's too early. This creates bugs when the resolution of the function name changes between the parsing and execution — easy to do in a stored procedure. For example
```sql
 create procedure p1
 begin
   select sleep(10);
   select f1();
 end
```
or, like
```sql
create procedure p1()
begin
  execute immediate "create function foo.f1() returns int return 10";
  select foo.f1();
end
```
Instead, see below

sql/item_create.cc
```diff
@@ -2838,27 +2839,44 @@ Item*
...
   if (!db.str)
-    return NULL; /*No db or EOM, error was already sent */
+  {
+    db= thd->lex->copy_db_normalized(false);
+
+    if (!db.str)
+      return NULL; /*No db or EOM, error was already sent */
+  }
 
   return create_with_db(thd, db, Lex_ident_routine(*name), false, item_list);
```
vuvova, line 2881: I understand that you added path search to keep this logic intact, but I sugest you change this line instead, don't create a fully resolved Item_func_sp here, create it without a db and later, during fix_fields you can do your path search and find the db.

sql/item_create.cc
```diff
@@ -3010,8 +3028,8 @@ Create_sp_func::create_with_db(THD *thd,
     arg_count= item_list->elements;
 
   qname= new (thd->mem_root) sp_name(db, name, use_explicit_name);
-  if (unlikely(sph->sp_resolve_package_routine(thd, thd->lex->sphead,
-                                               qname, &sph, &pkgname)))
+  if (unlikely(sph->sp_resolve_package_routine_sql_path(thd, thd->lex->sphead,
+                                                        qname, &sph, &pkgname)))
```
vuvova, line 3032: same

sql/item_strfunc.cc
```diff
@@ -6331,3 +6331,21 @@ longlong Item_func_wsrep_sync_wait_upto::val_int()
...
+String *Item_func_current_path::val_str(String *str)
+{
+  DBUG_ASSERT(fixed());
+  THD *thd= current_thd;
+  
+  auto length= thd->variables.path.text_format_nbytes_needed();
+  if (str->realloc(length))
+    return NULL;
+    
+  length= thd->variables.path.print(&(*str)[0], str->alloced_length());
```
vuvova, line 6345: this looks weird, why not `str->ptr()` ?

sql/item_strfunc.h
```diff
@@ -1281,6 +1281,35 @@ class Item_func_current_user :public Item_func_user
...
+
+class Item_func_current_path :public Item_func_sysconst
+{
+public:
+  Item_func_current_path(THD *thd): Item_func_sysconst(thd) {}
+  String *val_str(String *) override;
+  bool fix_length_and_dec(THD *thd) override
+  {
+    max_length=32767;
+    set_maybe_null();
```
vuvova, line 1293: I don't think returning NULL on OOM counts as "maybe null", OOM is a failure of the whole query anyway. In normal cases CURRENT_PATH can never be NULL, can it?

sql/mysqld.cc
```diff
@@ -4398,6 +4398,8 @@ static int init_common_variables()
     make_default_log_name(&opt_logname, ".log", false);
   if (!opt_slow_logname || !*opt_slow_logname)
     make_default_log_name(&opt_slow_logname, "-slow.log", false);
+  
+  global_system_variables.path.init();
```
vuvova, line 4402: if your init() doesn't allocate memory and cannot fail, you can as well do it in the constructor, it'd be simpler

sql/set_var.cc
```diff
@@ -113,6 +113,8 @@ void sys_var_end()
 
   for (sys_var *var=all_sys_vars.first; var; var= var->next)
     var->cleanup();
+  
+  global_system_variables.path.free();
```
vuvova, line 117: this could be in the destructor

sql/share/errmsg-utf8.txt
```diff
@@ -12294,3 +12294,5 @@ ER_VEC_DISTANCE_TYPE
         eng "Cannot determine distance type for VEC_DISTANCE, index is not found"
 ER_SIGNAL_SKIP_ROW_FROM_TRIGGER
         eng "The row is skipped by a trigger implementation"
+ER_INVALID_SCHEMA_NAME_LIST_SPEC
+        eng "Invalid schema list specification"
```
vuvova, line 12298: This is very non-specific. You can as well use ER_WRONG_VALUE_FOR_VAR and avoid creating a new error message. If you want to create a new one it should provide notably more information than ER_WRONG_VALUE_FOR_VAR.

sql/sp.cc
```diff
@@ -2238,6 +2241,75 @@ Sp_handler::sp_find_routine(THD *thd, const Database_qualified_name *name,
 }
 
 
+/**
+  Determine the existence of a routine.
+  Note: ONLY for unqualified routines
```
vuvova, line 2246: please, don't say "unqualified", it's ambiguous and very confusing. because db.routine is a *qualified* name, and here, in sp_find_routine_quick all routine names are qualified, because you use Database_qualified_name, which included db name.

so it's very confusing when you say "ONLY for unqualified routines" but the argument is Database_qualified_name.

What you mean, routines that don't have a package name, may be you can say "ONLY for package-unqualified routines" or "ONLY for routine names without a package name"

sql/sp.cc
```diff
@@ -2238,6 +2241,75 @@ Sp_handler::sp_find_routine(THD *thd, const Database_qualified_name *name,
...
+    goto done;
+  }
+
+  if ((ret= db_find_routine_aux(thd, name, table, false)) != SP_OK)
+    goto done;
+
+  thd->commit_whole_transaction_and_close_tables();
+  new_trans.restore_old_transaction();
+
+  thd->m_parser_state= oldps;
```
vuvova, line 2297: 1. why do you save/restore parser state?
2. why do you save/restore TIME_ZONE_USED?
3. why do you need this _quick variant at all? Why not to use the normal sp_find_routine() ?

sql/sp.cc
```diff
@@ -2493,6 +2565,38 @@ is_package_public_routine(THD *thd,
...
+    @param db       the database name
+    @param package  the package name
+    @param name     the routine name
+    @param type     the routine type
+    @retval         true,  if both routine declaration and implementation exist
+    @retval         false, either routine declaration or implementation does not exist
+*/
+
+static bool
+is_package_public_routine_with_body(THD *thd,
```
vuvova, line 2582: I don't think you need that. Currently (before your patch) the search only verifies that a routine exists in the package, and it's missing in the body - it's an error. Here:
```sql
  delimiter $$;
  create package pkg
    function f1() returns int;
  end;
  $$

  create database pkg$$
  create function pkg.f1() returns int return 2$$

  create procedure p1()
  begin
    select pkg.f1();
  end$$
  delimiter ;$$

  call p1();
```
this test does not print '2', it fails with "FUNCTION test.pkg.f1 does not exist"
The name lookup finds f1 in the package pkg and stops there, it does not check the package body. Your implementation should preserve existing behavior, don't check the body, just use is_package_public_routine().

sql/sp.cc
```diff
@@ -2694,6 +2798,104 @@ Sp_handler::sp_resolve_package_routine(THD *thd,
...
+  sp_name *qname= NULL;
+
+  if (!thd->db.length)
+  {
+    if (name->m_explicit_name)
+      ret= thd->variables.path.find_db_qualified(thd, name,
+                                                 pkg_routine_handler, pkgname);
+    else
+    {
+      if ((*pkg_routine_handler)->sp_find_routine_quick(thd, name))
```
vuvova, line 2834: I don't understand. is db name part of name? If not, sp_find_routine_quick, likely, won't work. So it is, right? Where it's getting db name from?

sql/sql_class.cc
```diff
@@ -776,7 +776,7 @@ THD::THD(my_thread_id id, bool is_wsrep_applier)
    wsrep_wfc()
 #endif /*WITH_WSREP */
 {
-  bzero(&variables, sizeof(variables));
+  bzero((void *)&variables, sizeof(variables));
```
vuvova, line 779: I'm not sure it's a good idea, compiler warns you for a reason. Perhaps you can simplify your Sql_path class so that one could bzero it?

sql/sql_class.h
```diff
@@ -7952,10 +7956,11 @@ inline Item *and_conds(THD *thd, Item *a, Item *b)
 }
 
 /* inline handler methods that need to know TABLE and THD structures */
-inline void handler::increment_statistics(ulong SSV::*offset) const
+inline void handler::increment_statistics(ulong SSV::*offset, bool update) const
 {
   status_var_increment(table->in_use->status_var.*offset);
-  table->in_use->check_limit_rows_examined();
+  if (update)
+    table->in_use->check_limit_rows_examined();
```
vuvova, line 7963: I'm still not sure why you do it.
You wrote that otherwise main.limit_rows_examined will fail, the test with test_limit_rows() procedure.
But I don't understand why. A lookup in the mysql.proc *before* invoking the procedure shouldn't really count towards rows examined in the SELECT *inside* the procedure. It doesn't matter if it's one examined row in mysql.proc, two, or ten. They shouldn't count either way.
So, I don't understand why you needed the above workaround.

sql/sql_lex.cc
```diff
@@ -9521,8 +9553,8 @@ bool LEX::call_statement_start(THD *thd, sp_name *name)
   const Sp_handler *sph= &sp_handler_procedure;
   sql_command= SQLCOM_CALL;
   value_list.empty();
-  if (unlikely(sph->sp_resolve_package_routine(thd, thd->lex->sphead,
-                                               name, &sph, &pkgname)))
+  if (unlikely(sph->sp_resolve_package_routine_sql_path(thd, thd->lex->sphead,
+                                                        name, &sph, &pkgname)))
```
vuvova, line 9557: same comment as above about searching in tables during parsing

sql/sql_lex.cc
```diff
@@ -9535,7 +9567,7 @@ bool LEX::call_statement_start(THD *thd, sp_name *name)
 
 bool LEX::call_statement_start(THD *thd, const Lex_ident_sys_st *name)
 {
-  sp_name *spname= make_sp_name(thd, *name);
+  sp_name *spname= make_sp_name_sql_path(thd, *name);
```
vuvova, line 9570: same comment as above about searching in tables during parsing

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
+#include "sql_plugin.h"
+#include "sql_class.h"    // class THD
+#include "sp_head.h"      // class sp_name
+#include "lex_ident.h"
+#include "sql_db.h"
+#include "sql_path.h"
+
+static constexpr LEX_CSTRING cur_schema= {STRING_WITH_LEN(".")};
```
vuvova, line 8: not "." but "CURRENT_SCHEMA"

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+
+Sql_path::Sql_path() :
+  m_schemas(m_schemas_array,
+            sizeof(m_schemas_array) / sizeof(m_schemas_array[0])),
```
vuvova, line 12: there's `array_elements()` helper that does exactly that

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+      Resolve the schema name to the current database if required.
+    */
+    if (schema.length == cur_schema.length &&
+                        !memcmp(schema.str, cur_schema.str, cur_schema.length))
+      schema= thd->db;
+
+    if (!schema.str)
+      continue;
+
+    if (likely(!check_db_dir_existence(schema.str)))
```
vuvova, line 39: I suspect you should've used the normalized name for check_db_dir_existence

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+        }
+      }
+    }
+  }
+
+  return false;
+}
+
+
+Schema *Sql_path::find_first_internal_schema() const
```
vuvova, line 107: Let's move internal schema handling to a separate commit anda separate PR. Ok? Just to make commits a bit simpler and more focused.

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+
+  return false;
+}
+
+
+bool Sql_path::add_schema(const char *schema_str, size_t schema_len)
+{
+  DBUG_ASSERT(schema_str);
+
+  if (unlikely(m_count >= m_schemas.size()))
```
vuvova, line 145: no my_error here? Not even my_error(ER_VALUE_TOO_LONG) ?

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+    return true;
+  
+  /*
+    Disallow duplicate schema names. The intent here is to avoid
+    ambiguity in the order of schema resolution when the path
+    contains duplicate schema names.
```
vuvova, line 151: You can also mention here that SQL:2016 very clearly says

>  No two \<schema name>s contained in \<schema name list> shall be equivalent.

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+  */
+  for (size_t i= 0; i < m_count; i++)
+  {
+    if (unlikely(m_schemas[i].length == schema_len &&
+        !memcmp(m_schemas[i].str, schema_str, schema_len)))
```
vuvova, line 156: This is not correct, you need to compare normalized paths, not what user has specified.
```sql
   SET PATH='A,a'
```
would pass your check, but on Windows these schema names are equivalent.

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+
+  m_schemas[m_count++]= {schema_str, schema_len};
+  return false;
+}
+
+
+void Sql_path::set(const Sql_path &rhs, uint version_increment)
+{
+  init_array();
+  free();
```
vuvova, line 171: you seem to use this pair init_array/free awfully often.
Perhaps init_array() should just call free() internally?

sql/sql_path.cc
```diff
@@ -0,0 +1,412 @@
...
+  {
+    const LEX_CSTRING &schema= m_schemas[i];
+    size_t len= schema.length;
+    if (nbytes + len + 3 > nbytes_available)
+      break;
+
+    *dst++= '`';
+    memcpy(dst, schema.str, len);
+    dst+= len;
+    *dst++= '`';
```
vuvova, line 396: This is not exactly correct, try a name with a backtick inside.
```
MariaDB [test]> create database `fo``o`;
Query OK, 1 row affected (0.001 sec)

MariaDB [test]> show databases;
+--------------------+
| Database           |
+--------------------+
| fo`o               |
| information_schema |
| mysql              |
| performance_schema |
| test               |
+--------------------+
5 rows in set (0.002 sec)

MariaDB [test]> \u fo`o
Database changed

MariaDB [fo`o]> 
```

sql/sql_path.h
```diff
@@ -0,0 +1,145 @@
...
+  {
+    set(std::move(rhs), 1);
+    return *this;
+  }
+  Sql_path(const Sql_path&) = delete;
+
+  void init_array()
+  {
+    m_schemas.reset(m_schemas_array,
+                    sizeof(m_schemas_array) / sizeof(m_schemas_array[0]));
```
vuvova, line 39: there's `array_elements()` helper that does exactly that

sql/sql_path.h
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  bool find_db_unqualified(THD *thd, const LEX_CSTRING &name,
+                           const Sp_handler *sph,
+                           Lex_ident_db_normalized *dbn_out,
+                           sp_name **spname_out) const;
+  bool find_db_qualified(THD *thd,
+                         sp_name *name,
+                         const Sp_handler **sph,
+                         Database_qualified_name *pkgname) const;
```
vuvova, line 49: I know by now what you mean by "qualified" and "unqualified", but it's still confusing. May be better rename (and reword everywhere in the commit) to something like find_db_routine and find_db_package_routine ?

sql/sql_path.h
```diff
@@ -0,0 +1,145 @@
...
+  bool init();
+  /*
+    Free the memory allocated by the path variable
+  */
+  void free();
+  /*
+    Set the variable to the value of rhs, making a copy of the buffer
+
+    @param rhs              The path variable to copy from
+    @param version_increment (not used)
```
vuvova, line 73: why do you have it, if it's not used?

sql/sql_view.cc
```diff
@@ -1018,6 +1018,7 @@ static int mysql_register_view(THD *thd, DDL_LOG_STATE *ddl_log_state,
   backup_file_name[0]= 0;
   {
     Sql_mode_save_for_frm_handling sql_mode_save(thd);
+    Sql_path_save_and_clear sql_path_save(thd->variables.path);
```
vuvova, line 1021: why is that?

sql/sql_yacc.yy
```diff
@@ -17378,6 +17381,20 @@ option_value_no_option_type:
                                                            yychar == YYEMPTY)))
               MYSQL_YYABORT;
           }
+        | PATH_SYM equal
```
vuvova, line 17384: Correct SQL standard syntax is `SET PATH <schema list>` no equal sign.
And it cannot be combined with other assignments, so put it in `set_param` rule not in `option_value_no_option_type`

sql/sys_vars.inl
```diff
@@ -3063,3 +3064,120 @@ private:
...
+    global_system_variables.path.init();
+  }
+
+  bool session_update(THD *thd, set_var *var) override
+  {
+    if (thd->spcont)
+    {
+      /*
+        Because we invalidate the sp caches, we can't set the session path
+        in a stored function.
```
vuvova, line 3136: that's because you resolve names in the parser, if you'd do it at fix_fields it'd likely work

vuvova (changes requested):
In general, this is **very** good. I mean, the coding style, all the small details, the level of understanding of what does what is just excellent.

There are comments and requested changes, they always are in a PR of that complexity, and at least partially they're caused by insufficiently clear MDEV description. Also, see below, I've asked a bunch of questions about changes I didn't understand.

iqbal-rsec:
> In general, this is **very** good. I mean, the coding style, all the small details, the level of understanding of what does what is just excellent.
> 
> There are comments and requested changes, they always are in a PR of that complexity, and at least partially they're caused by insufficiently clear MDEV description. Also, see below, I've asked a bunch of questions about changes I didn't understand.

@vuvova, thank you so much for the review comments. I will get back to you shortly, at the moment I am working on another task.

sql/item_create.cc
iqbal-rsec: I see, apologies, I failed to consider these cases. I will rework this. I don't quite understand the first example though, can you explain in detail please?

sql/item_create.cc
iqbal-rsec: I see understood. I think caching (and compiling?) of routines are only done at parser stage, so I have to move those there as well right?

sql/item_strfunc.cc
iqbal-rsec: Apologies, not sure what happened here, will fix.

sql/item_strfunc.h
iqbal-rsec: Of course, I will remove this.

sql/mysqld.cc
iqbal-rsec: global_system_variable is a global 'C++' variable though, I think I may have done it this way because I worried I might initialize something that calls my_alloc/etc before the pre-requisites has finished initialization. I've checked again and it can be removed, the code in ini is trivial.

sql/set_var.cc
iqbal-rsec: I think I cannot remove this, some memory needs to be freed before mysqld_exit call or there will be memory accounting error, but I'll double check.

sql/share/errmsg-utf8.txt
iqbal-rsec: Will remove.

sql/sp.cc
iqbal-rsec: This is from the original implementation by wong, and I didn't fully check the utility of this function, apologies. I'll check in detail and update you on this.

sql/sql_class.cc
iqbal-rsec: If I remove the Bounds_checked_array the class will be POD.

sql/sql_class.h
iqbal-rsec: Apologies, I will double check again.

sql/sql_path.h
iqbal-rsec: Thank you, will fix.

sql/sql_path.h
iqbal-rsec: Apologies, will remove.

sql/sql_yacc.yy
iqbal-rsec: Understood, will fix this.

mysql-test/main/path_resolution.result
iqbal-rsec: Understood, will fix this, though obviously it will take some time to implement.

sql/sql_view.cc
iqbal-rsec: This is actually for path invariance for views. I will need to remove this after moving path resolution to fix_fields and adding object specific paths.

mysql-test/suite/sys_vars/r/path_basic.result
iqbal-rsec: Understood, will change.

sql/sql_path.cc
iqbal-rsec: Will add my_error. Not sure how but currently ER_WRONG_VALUE_FOR_VAR is raised, probably set by callee.

sql/item_create.cc
vuvova: first example implies that while `p1` waits, somebody else creates or drops a function `f1`, for example, doing
```sql
create function foo.f1() returns int return 10
```
from a different connection. I admit both examples are rather far-stretched, I just wanted to show that resolving all names during the parsing is not a good idea.

sql/item_create.cc
vuvova: Is it? I only found that it's done for packages and we're already considering to move it to fix_fields too. For standalone routines it is done in `Item_func_sp::fix_fields()`

sql/set_var.cc
vuvova: I suspect you're right. Ok, then

sql/sp.cc
vuvova: thanks

sql/sp.cc
vuvova: I mean, if there is a current database, then db name could be a current database. But here you're in the
```c
if (!thd->db.length)
```
branch, there is no current database. And `name->m_explicit_name` is also false. So db name can only be set if you've already searched the path earlier. Is it what happening here, db name is set from you searching the path?

sql/sql_class.h
vuvova: May be you need to reset (or save/restore) `thd->accessed_rows_and_keys` when you're searching the path?

sql/sql_path.cc
vuvova: yes, the caller issues a generic ER_WRONG_VALUE_FOR_VAR at the very end if no more specific error was issued yet. But here a more specific ER_VALUE_TOO_LONG would look better, I think

sql/sql_path.h
vuvova: thanks

sql/item_create.cc
iqbal-rsec: I see, thanks for clearing that up. It is fine for me that the examples are edge cases, it's still something that I failed to consider, so thank you for bringing it up.

sql/item_create.cc
iqbal-rsec: Apologies, I must have misremembered. Alright I will moved the caching for packages to fix_fields too.

vuvova: Oh, no, I didn't mean that at all. I meant **we** plan to move package name resolution out of the parser, using our developers.

Your PR is about `SET PATH` — let's not to increase it's scope.

You're still welcome to fix packages too, of course, if you'd like. But, please, do it in a separate PR.

julienfritsch44:
Hi, we are about to start a six-week development sprint on the 22nd of April. Can this be finalised during that time? This would allow us to merge this in 12.1.

vuvova:
Hi, @iqbal-rsec, this feature missed 12.1, but now let's try to get it in 12.2.

It seems you're either stuck or moved on (which is fine), would you mind if I'll take over and finish it?
Let's say, I'll wait till the end of July — until 12.1 release — and if it's not read by that time, I'll ready it for 12.2 myself.
Otherwise it'll likely miss 12.2 too.

iqbal-rsec:
> Hi, @iqbal-rsec, this feature missed 12.1, but now let's try to get it in 12.2.
> 
> It seems you're either stuck or moved on (which is fine), would you mind if I'll take over and finish it? Let's say, I'll wait till the end of July — until 12.1 release — and if it's not read by that time, I'll ready it for 12.2 myself. Otherwise it'll likely miss 12.2 too.

Hi @vuvova, so sorry for the late reply. I missed the notification from your message somehow. I do have a preliminary version but with a couple of TODOs littered in some places. Is it OK if I push so that I can have review while I try to figure out the rest? Just in case the implementation is way off base.

vuvova:
Yes, of course. Please, do push and request a review.
Also, if you think some of my suggestions should not be implemented (for whatever reason), please say so in a comment

iqbal-rsec:
> Yes, of course. Please, do push and request a review. Also, if you think some of my suggestions should not be implemented (for whatever reason), please say so in a comment

Thank you! I've pushed but I'll refrain first from requesting a review until I've fixed minor test failures (if possible). The one thing that has been causing me problem has been moving path resolution to fix_fields, mainly because we need the name to be resolved before MDL can happen so that we can load the sp into cache. My latest implementation that I pushed attempted to do this, but tbh it's not a clean implementation - hence why I am requesting this review to check if I am totally off base.

vuvova:
I'm sorry, I didn't understand. Are you requesting a review or do you still want to change something before you want me to take a look?

As for moving path resolution to fix_fields — well, don't do it then, just reply to my comment, like "are you sure, it's necessary? this is very difficult to do because of _this_ and _that_ and would require big refactoring, _etc etc etc_", like, just explain why it's not a good idea in your opinion. And I'll either agree to suggest something.

In general, only do my comments you agree with, otherwise reply and explain why you disagree

iqbal-rsec:
> I'm sorry, I didn't understand. Are you requesting a review or do you still want to change something before you want me to take a look?
> 
> As for moving path resolution to fix_fields — well, don't do it then, just reply to my comment, like "are you sure, it's necessary? this is very difficult to do because of _this_ and _that_ and would require big refactoring, _etc etc etc_", like, just explain why it's not a good idea in your opinion. And I'll either agree to suggest something.
> 
> In general, only do my comments you agree with, otherwise reply and explain why you disagree

I do want to change some things first before I request for a review. I apologize, but with my limited knowledge of the codebase I very much prefer to at least implement first what you requested. I will need a little bit more time to make the changes, I will ping you back once I am done.

sql/sp.cc
iqbal-rsec: 1. This is wrongly copied, save/restore parser state is not needed.
2. Same as above
3. This function (now renamed to `sp_routine_exists`) does not call sp_compile()/update the sp cache, basically a lightweight version of `sp_find_routine` just to determine if a routine exists.

iqbal-rsec: Sorry I made a mistake. Saving and restoring the parser state is required. When opening the proc table, the parser need to be used to parse default values (i.e. `created` column).

sql/sql_class.h
iqbal-rsec: I modified the code to save/restore `thd->accessed_rows_and_keys`.

sql/item_create.cc
iqbal-rsec: @vuvova I tried and failed to move path resolution to fix_fields/execute. Problems:
1. MDL requires fully resolved names, so resolution need to be performed before MDL acquisition.
2. This also causes resolution to be performed to twice, unless we keep track of the relationship between `MDL_key` and `sp_name` held by `Item_func_sp` or `Sql_cmd_call`
3. Keeping track of which PATH is active when PATH is specified/embedded within routine definition is complicated since  MDL does not care or know about SPs.
4. We also need to resolve when calling `sp_add_used_routine` so that we can search for preexisting `Sroutine_hash_entry`, which is based on MDL_key. This means that we still need to resolve during parsing (`sp_compile`) if a routine calls another routine.
5. Many non-PATH test failures caused by the changes that requires test modifications.

vuvova: I see, thanks for the explanation!

vuvova:
A question (please, don't change anything yet, it might make the review obsolete before I even finish): why did you implement the syntax, like
```sql
CREATE PROCEDURE proc()
PATH 'test2'
BEGIN
  SELECT CONCAT(func2(), '+', func2());
END;
```
specifying path in the routine definition is not standard, as far as I can see. Was it for compatibility with something?

iqbal-rsec:
> A question (please, don't change anything yet, it might make the review obsolete before I even finish): why did you implement the syntax, like
> 
> ```sql
> CREATE PROCEDURE proc()
> PATH 'test2'
> BEGIN
>   SELECT CONCAT(func2(), '+', func2());
> END;
> ```
> 
> specifying path in the routine definition is not standard, as far as I can see. Was it for compatibility with something?

This is so that we can specifiy (or fix) the PATH during creation. I do not have access to the standard unfortunately.

vuvova:
I see, thanks

vuvova:
Just FYI, as a progress report, I'm almost done and hopefully it'll be ready soon for 12.3 preview

vuvova:
pushed into 12.3, thanks!

https://mariadb.com/docs/release-notes/community-server/12.3/mariadb-12.3-changes-and-improvements#compatibility-features

PR #3798, opened 2025-01-28
MDEV-35081 - Assertion `!n_mysql_tables_in_use' failed after error up…

svoj:
- [x] *The Jira issue number for this PR is: MDEV-35081*

## Description
…on binary logging of DML involving vector table

InnoDB is limited to row-logging when transaction isolation level is READ COMMITTED or READ UNCOMMITTED. This limitation is enforced by a check within ha_innobase::external_lock().

InnoDB also expects number of "successful external locks" to match number of "external unlocks", which is controlled by the assertion in subject.

However "unlock" was called even after failing "lock" for high-level indexes.

Fixed by calling "unlock" only after "successful lock".

## Release Notes
N/A

## How can this PR be tested?
mtr binlog_innodb_stm

- [] *This is a new feature or a refactoring, and the PR is based against the `main` branch.*

## PR quality check


svoj:
@vuvova please note that #3631 as it is now may solve this particular issue as well, but it won't solve case when hlindex lock fails, but table lock itself succeeds.

svoj:
@vuvova thinking more on this... you mentioned once that hlindex has to be locked after the table. But in this case it is locked before:
`Sql_cmd_dml::execute` / `Sql_cmd_dml::prepare` / `Sql_cmd_update::prepare_inner` / `JOIN::prepare` / `multi_update::prepare` / `handler::prepare_for_modify` / `TABLE::open_hlindexes_for_write` / `TABLE::hlindex_lock` / `handler::ha_external_lock`

The table itself is locked later:
`Sql_cmd_dml::execute` / `lock_tables` / ...

It can happen so that this is the real problem rather than what I try to solve here.

vuvova:
I think, indeed, that the hlindex should be locked after the table, but your fix still makes sense, after or before, if external_lock didn't lock the table, there shouldn't be external_unlock, should it?

svoj:
@vuvova, right, it goes inline with my understanding.

PR #3799, opened 2025-01-28
All-green GitLab CI in main branch in January 2025

ottok:
## Description

Make GitLab CI green again on `main` branch. See commit for details.

## Release Notes

N/A

## How can this PR be tested?

This can be tested easily by simply running GitLab CI on this branch before and after the changes in this PR.

Before this change:

image

With this change:

image

This targets intentionally the `main` branch as with the current number of failures it is not feasible to fix everything on all branches. It is better to have `main` green now, and be active in responding to future regressions on `main` as they are detected.

Note that this has no functional changes to MariaDB Server itself, and thus safe to merge. The faster this is merged, the faster we can start catching new testable regressions.

## Copyright

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,10 +40,10 @@ default:
 # submodules (a commit in this repo does not affect their builds anyway) and
 # many components that are otherwise slow to build.
 variables:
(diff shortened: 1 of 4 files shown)


ottok:
@cvicentiu Are you OK I merge this relatively quickly as it only affects CI and there are probably not that many users who run this piece of CI code to validate MariaDB code base health?

grooverdan (approved):
note MDEV-20966 shouldn't be a problem any more.

PR #3802, opened 2025-01-29
[MDEV-35969] reset service manager status

eworm-de:
- [x] *The Jira issue number for this PR is: MDEV-35969*

## Description
I am running a Galera cluster with MariaDB 11.4.4. When ever a server does a WSREP state transfer it results in the service manager reporting that WSREP state transfer, visible in status:
```
root@mariadb01 ~ # systemctl --lines=0 --no-pager status mariadb.service
● mariadb.service - MariaDB 11.4.4 database server
     Loaded: loaded (/usr/lib/systemd/system/mariadb.service; enabled; preset: disabled)
    Drop-In: /etc/systemd/system/mariadb.service.d
             └─LimitNOFILE.conf, Shutdown.conf
     Active: active (running) since Tue 2025-01-28 12:12:19 CET; 23h ago
 Invocation: e079a859e5f24b2f8a9e48013e7d61cb
       Docs: man:mariadbd(8)
             https://mariadb.com/kb/en/library/systemd/
   Main PID: 984 (mariadbd)
     Status: "WSREP state transfer ongoing..."
      Tasks: 491 (limit: 1020061)
     Memory: 116.9G (peak: 123G)
        CPU: 11h 46min 3.224s
     CGroup: /system.slice/mariadb.service
             └─984 /usr/bin/mariadbd --wsrep_start_position=00000000-0000-0000-0000-000000000000:-1
```
However that status is not updated after the state transfer. Instead it should change back to the default status "_Taking your SQL requests now..._".

## Release Notes

* Update status in service manager after state transfer

## How can this PR be tested?

After a state transfer the status should switch back. No idea if this can be tested in current test suite.

## PR quality check

### DIFF
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -132,6 +132,8 @@ static void* wsrep_sst_donor_monitor_thread(void *arg __attribute__((unused)))
   WSREP_INFO("Donor monitor thread ended with total time %lu sec", time_waited);
   mysql_mutex_unlock(&LOCK_wsrep_donor_monitor);
 
+  sd_notify(0, "STATUS=WSREP state transfer (role donor) completed.\n");
+
   return NULL;
 }
 
@@ -168,6 +170,8 @@ static void* wsrep_sst_joiner_monitor_thread(void *arg __attribute__((unused)))
 
   WSREP_INFO("Joiner monitor thread ended with total time %lu sec", time_waited);
   mysql_mutex_unlock(&LOCK_wsrep_joiner_monitor);
+
+  sd_notify(0, "STATUS=WSREP state transfer (role joiner) completed.\n");
 
   return NULL;
 }
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -125,7 +125,7 @@ static void* wsrep_sst_donor_monitor_thread(void *arg __attribute__((unused)))
                  "is not completed",
                  time_waited);
       service_manager_extend_timeout(WSREP_EXTEND_TIMEOUT_INTERVAL,
-        "WSREP state transfer ongoing...");
+        "WSREP state transfer (role donor) ongoing...");
     }
   }
 
@@ -162,7 +162,7 @@ static void* wsrep_sst_joiner_monitor_thread(void *arg __attribute__((unused)))
                  "is not completed",
                  time_waited);
       service_manager_extend_timeout(WSREP_EXTEND_TIMEOUT_INTERVAL,
-        "WSREP state transfer ongoing...");
+        "WSREP state transfer (role joiner) ongoing...");
     }
   }
 


janlindstrom (approved):
Looks good.

grooverdan:
Happy with most of this. Good idea.

Completely optionally, did you want to go further and for the donor list the joiner node and for the joiner, list the donor?

Though the READY=1 is probably harmless, it doesn't appear to be relevant at this point in the server execution. Was there are reason to to include it?

ref: https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#READY=1

eworm-de:
> Completely optionally, did you want to go further and for the donor list the joiner node and for the joiner, list the donor?

Sounds reasonable, I will have a look. This should be straight forward for the joins. Can a donor have more than one joiner at a time?

> Though the READY=1 is probably harmless, it doesn't appear to be relevant at this point in the server execution. Was there are reason to to include it?

That's true, should not be required here. Just push a new commit with that dropped.

But thinking about this... In case of the joiner role, does the code go through `handle_connections_sockets()` after state transfer is complete, making the extra call to `sd_notify()` useless (or even wrong, as it happens too early)?

eworm-de:
Hmm, is there an easy way to get the remote joiner or donor name? Or is it available within the Galera plugin only?

grooverdan:
> But thinking about this... In case of the joiner role, does the code go through handle_connections_sockets() after state transfer is complete, making the extra call to sd_notify() useless (or even wrong, as it happens too early)?

Might depend on if its a mysqldump sst or another one. Needs a deeper look.

>  Can a donor have more than one joiner at a time?

Fairly sure it cannot be.

> Hmm, is there an easy way to get the remote joiner or donor name? Or is it available within the Galera plugin only?

I haven't looked to closely where it would be. Guess I was hoping it was close.

> Alternatively... How about setting new state:
> 
> > "WSREP state transfer (role) finished."
> 
> This would not set the initial state ("_Taking..._"), but would be fine fore me as long as it does not read "_... ongoing..._".

I'm pretty open to anything factually correct.

eworm-de:
Just updated again... I think it's better to go with this wording.

No idea where to get the remote node's name... Give me a hint if you know. 😉

grooverdan (approved):

grooverdan:
> Just updated again... I think it's better to go with this wording.
> 
> No idea where to get the remote node's name... Give me a hint if you know. 😉

I got as far as https://github.com/grooverdan/mariadb-server/commit/91594dd690ea18d8cad651828ac3ee930f21b22f

Both are joiner addresses, and if the joiner is aware who the donor its much higher than where the monitor threads are created and hence not really suitable for a restructure. Sorry I suggested it :smirk_cat: .

sysprg:
@janlindstrom To be able to pass the donor name to the notification handlers, it must be available to the server. To do this, we would probably need to take the node name from the GCS-level node structures inside the Galera library code and pass it to the donate callback, then drag it through the wsrep-lib into the start_sst() callback so that it ends up in the Wsrep_server_service::start_sst(). But this may break API compatibility with the library, so alternatively you could attach the donor name to the request string that is already passed through all layers - for example at the end of the request_string could be the construct “....#donor_name”. What do you think about it?

eworm-de:
Not sure this is worth the trouble... The log still has all the details.

Having the state in service manager is nice, but that's it. It needs to be precise and correct, but not too detailed.

eworm-de:
Any progress on this? I would like to see it merged before the next round of releases...

sysprg:
@eworm-de Thanks, the fixes have been merged with the head revision:
https://github.com/MariaDB/server/commit/0b8762822eb36523239de5ba994f71f1630c3173 +
https://github.com/MariaDB/server/commit/02893e875bc346cc930ce0418e9e7a09986354b9

eworm-de:
Hmm, this should have gone into `10.5`, no? Why did it end up in `11.4`?

svoj:
@eworm-de 10.5 is open only for critical/security fixes.

PR #3805, opened 2025-01-29
MDEV-35975 Server crashes after CREATE VIEW as SELECT ST_COLLECT...

DaveGosselin-MariaDB:
Ensure that orig_args is populated during Item_func_collect::fix_fields_impl

### DIFF
--- a/sql/item_sum.cc
+++ b/sql/item_sum.cc
@@ -3582,6 +3582,9 @@ Item_sum_str::fix_fields(THD *thd, Item **ref)
   if (check_sum_func(thd, ref))
     return TRUE;
 
+  if (arg_count)
+    memcpy (orig_args, args, sizeof (Item *) * arg_count);
+
   base_flags|= item_base_t::FIXED;
   return FALSE;
 }
@@ -4018,8 +4021,6 @@ Item_func_group_concat(THD *thd, Name_resolution_context *context_arg,
 
   /* orig_args is only used for print() */
   orig_args= (Item**) (order + arg_count_order);
-  if (arg_count)
-    memcpy(orig_args, args, sizeof(Item*) * arg_count);
   if (limit_clause)
   {
     row_limit= row_limit_arg;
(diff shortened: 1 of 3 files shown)


sql/item_sum.cc
```diff
@@ -4818,6 +4818,8 @@ const Type_handler *Item_func_collect::type_handler() const
 
 bool Item_func_collect::fix_fields_impl(THD *thd,Item **)
 {
+  if (arg_count)
+    memcpy (orig_args, args, sizeof (Item *) * arg_count);
```
vuvova, line 4822: why not to do it in `Item_sum_str::fix_fields()`? And remove it from `Item_func_group_concat` constructor, because `fix_fields` can change arguments, so it's wrong to do in the constructor anyway (otherwise it would've been done in `Item_sum::Item_sum`)

vuvova (changes requested):

sql/item_sum.cc
DaveGosselin-MariaDB: There is an `Item_sum` constructor that copies the arguments: `Item_sum::Item_sum(THD *thd, Item_sum *item)`.  There is no implementation of `fix_fields` on the `Item_sum` class.  It seems that descendants of `Item_sum` either hope that the correct constructor was called or they copy the args themselves during `fix_fields`.  In my case here, I agree that we need to move the copy to the `Item_sum_str::fix_fields`.  As part of this same change, would you like me also to remove the argument copy from `Item_sum::Item_sum(THD *thd, Item_sum *item)` and ensure that all descendants perform it?  (If so, then I will likely implement it similar to what I have here--a common `fix_fields` implementation that ensures the copy, then invokes a virtual `fix_fields_impl` for the descendant-specific logic).

DaveGosselin-MariaDB: (as an aside, `Item_func_group_concat` constructor has had the argument copying logic within it for some time, at least since 2010)

vuvova: `Item_sum::Item_sum(Item_sum*)` copies them, I think it's a clone constructor that is invoked after `fix_fields`. The constructor which is called from the parser does not copy arguments, because it'd be too early. On the other hand, `fix_fields()` won't be called for a clone of an already fixed item, so `Item_sum::Item_sum(Item_sum*)`  not only **can** but **has to** copy arguments.

I know `Item_func_group_concat` constructor incorrectly did it for a while . Supposedly, we don't have a `GROUP_CONCAT()` test case where arguments change in `fix_fields()`.

Summary: I think this fix should be only about `group_concat` and `st_collect`.

vuvova (approved):
Looks good, thanks!

PR #3806, opened 2025-01-30
MDEV-35960 st_isvalid(NULL) returns NULL

DaveGosselin-MariaDB:
To better align with MySQL, st_isvalid(NULL) returns NULL

The same changes apply to st_isempty, st_issimple, and st_isring


DaveGosselin-MariaDB:
Fixed by 38cc21663461ec351c71c43309fcdd355ec9d176

PR #3815, opened 2025-02-04
MDEV-36015: unrepresentable value in row_parse_int()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36015*
## Description
`row_parse_int()`: Refactor the code and define the function `static` in one compilation unit. For any negative values, we must return 0.

`row_search_get_max_rec()`, `row_search_max_autoinc()`: Moved to the same compilation unit with `row_parse_int()`. Do not unnecessarily invoke `rec_get_offsets()`; there is an easier way to determine the length of the first index field.

We also remove a work-around of an internal compiler error when targeting ARMv8 on GCC 4.8.5, a compiler that is no longer supported.
## Release Notes
This is probably not worth mentioning, because the problem is only reported by `clang -fsanitize=undefined` and the regression test suite already covers this undefined behaviour in the test `innodb.autoinc_persist`.
## How can this PR be tested?
```sh
./mtr innodb.autoinc_persist
```
For completeness, I extended the test with the `FLOAT UNSIGNED` and `DOUBLE UNSIGNED` data types, which were not affected by this.

This bug affects also MariaDB Server 10.5, which will soon reach its end of life. Before MariaDB Server 10.6, the generic InnoDB error handling was different (crash on an error) and therefore also this fix would have to be different.
## PR quality check


storage/innobase/row/row0ins.cc
```diff
@@ -3810,3 +3841,83 @@ row_ins_step(
...
+    if (field.col->is_nullable() && *nulls & 1)
+      return 0;
+    len= field.fixed_len;
+    if (!len)
+      len= *lens;
+  }
+  else if (rec_get_1byte_offs_flag(rec))
+    len= rec_1_get_field_end_info(rec, 0);
+  else
+    len= rec_2_get_field_end_info(rec, 0);
```
mariadb-DebarunBanerjee, line 3883: I see the intention to to optimize the code here to get rid off `rec_init_offsets` call but I would still recommend to revert this optimization. It is better to have such low level manipulation (related to record format) abstracted to limited set of functions defined together. We would have hard time maintaining the logic if scattered. Also, such changes has inherent risk of regression and better be avoided for older released version.

It would be an interesting thought to improve performance in general for callers of `rec_init_offsets` when only part of the information is needed but I think it needs more thoughts and preferably with some performance results.

What do you think ?

storage/innobase/row/row0ins.cc
```diff
@@ -3810,3 +3841,83 @@ row_ins_step(
...
+  /* Open at the high/right end (false), and init cursor */
+  if (pcur.open_leaf(false, index, BTR_SEARCH_LEAF, mtr) != DB_SUCCESS)
+    return nullptr;
+
+  do
+  {
+    const page_t *page= btr_pcur_get_page(&pcur);
+    const rec_t *rec= page_find_rec_max_not_deleted(page);
+    if (page_rec_is_user_rec_low(rec - page))
+      return rec;
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: I see now that the assert is moved to row_read_autoinc() and the comment can be closed,

storage/innobase/row/row0ins.cc
dr-m: This code should indeed not be performance critical, and therefore we should optimize for code size. I’ll check that and get back to this. If this code turns out to be shorter than invoking `rec_get_offsets()`, then we’d better demonstrate that all code branches are covered in all cases: when the first field is NULL, the table is in `ROW_FORMAT=REDUNDANT` with a total size below or above 255 bytes, and when instant `ADD` or `DROP` of columns has been used.

dr-m: In the `CMAKE_BUILD_TYPE=RelWithDebInfo` executable that I checked, this function `row_read_autoinc()` was included inline in `row_search_max_autoinc()`. The size of that function would shrink by about ¼ when I reverted to the use of `rec_get_offsets()`, which obviously was not inlined.

storage/innobase/row/row0ins.cc
```diff
@@ -3810,3 +3841,64 @@ row_ins_step(
...
+  mem_heap_t *heap= nullptr;
+  rec_offs offsets_[REC_OFFS_HEADER_SIZE + 2];
+  rec_offs_init(offsets_);
+  rec_offs *offsets= rec_get_offsets(rec, &index, offsets_,
+                                     index.n_core_fields, 1, &heap);
+  ut_ad(!heap);
+
+  size_t len;
+  rec_get_nth_field_offs(offsets, 0, &len);
+  return row_parse_int(rec, len, field.col->mtype, field.col->prtype);
```
mariadb-DebarunBanerjee, line 3866: Previously we had ...
```
data = rec_get_nth_field(rec, offsets, col_no, &len);
value = row_parse_int(data, len, mtype, unsigned_type);
```
I realized that we could pass rec pointer directly here to  `row_parse_int` for first column `col_no = 0`. I would not press on this point, but perhaps you might consider something like 

`rec+= rec_get_nth_field_offs(offsets, 0, &len);`

or
```
ut_d(auto col0_offset= )rec_get_nth_field_offs(offsets, 0, &len);
ut_ad(col0_offset == 0);
```

storage/innobase/row/row0ins.cc
```diff
@@ -3810,3 +3841,64 @@ row_ins_step(
...
+  ut_ad(!DATA_BIG_COL(field.col));
+  ut_ad(!(rec_get_info_bits(rec, index.table->not_redundant()) &
+          (REC_INFO_MIN_REC_FLAG | REC_INFO_DELETED_FLAG)));
+  mem_heap_t *heap= nullptr;
+  rec_offs offsets_[REC_OFFS_HEADER_SIZE + 2];
+  rec_offs_init(offsets_);
+  rec_offs *offsets= rec_get_offsets(rec, &index, offsets_,
+                                     index.n_core_fields, 1, &heap);
+  ut_ad(!heap);
+
```
mariadb-DebarunBanerjee, line 3863: Some notes. Not necessarily need any code change.

1. `REC_OFFS_NORMAL_SIZE` changed to `REC_OFFS_HEADER_SIZE + 2`. It should be ok for the first column (n =1) as in rec_get_offsets we have ...
 `size = n + (1 + REC_OFFS_HEADER_SIZE);`
 
2. Hopefully rec_offs_get_n_alloc(offsets) >= REC_OFFS_HEADER_SIZE +2 for heap not to be allocated. The regular asserts check ut_ad(n_alloc > REC_OFFS_HEADER_SIZE);

mariadb-DebarunBanerjee (approved):
The change looks good to me. Added notes for the two changes from base code.

storage/innobase/row/row0ins.cc
dr-m: I had tested this with `REC_OFFS_HEADER_SIZE+1`. In that case, the `offsets` would be allocated from `heap`  and the assertion `!heap` would fail. The minimum size `offsets` array to allocate is `REC_OFFS_HEADER_SIZE` + 1 + the number of fields for which the offsets needs to be initialized. We have some similar code in `row_purge_reset_trx_id()`, where the maximum number of fields is being checked in the test `innodb.purge`.

storage/innobase/row/row0ins.cc
dr-m: Sure, we can document this design constraint that the first field of a record starts at the origin of the record, by adding such a debug assertion.

PR #3819, opened 2025-02-06
MDEV-29445: Reimplement SET GLOBAL innodb_buffer_pool_size

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-29445*

## Description
This is a rebase of #3107 to the `main` branch, resolving conflicts related to the removal of the change buffer and the refactoring of the adaptive hash index.

The buffer pool will be mapped in a contiguous memory area that will be aligned and divided to extents of 8 MiB on 64-bit systems and 2 MiB on 32-bit systems.

Within an extent, the first few `innodb_page_size pages` contain `buf_block_t` objects that will cover the page frames in the rest
of the extent. In this way, there is a trivial mapping between page frames and block descriptors and we do not need any lookup tables like `buf_pool.zip_hash` or `buf_pool_t::chunk_t::map`.

We will always allocate the same number of page frames for an extent, even if we do not need all the `buf_block_t` in the last extent in case the `innodb_buffer_pool_size` is not an integer multiple of the of extent size.

`my_large_virtual_alloc()`: A new function, similar to `my_large_malloc()`.
On Microsoft Windows, buffer pool resizing is disabled if large pages are being used.

`buf_pool_t::create()`: Only initialize the first page descriptor of each chunk.

`buf_pool_t::lazy_allocate()`: Lazily initialize a previously allocated page descriptor and increase `buf_pool.n_blocks`, which must be below `buf_pool.n_blocks_alloc`.

`innodb_init_param()`: Refactored. We must first validate `innodb_page_size` and then determine the valid bounds of
`innodb_buffer_pool_size`.

## Release Notes
We deprecate and ignore the parameter `innodb_buffer_pool_chunk_size` and let the buffer pool size to be changed in arbitrary 1-megabyte increments, all the way up to `innodb_buffer_pool_size_max`, which must be specified at startup.

If `innodb_buffer_pool_size_max` is not specified, it will default to twice the specified `innodb_buffer_pool_size`.

The minimum `innodb_buffer_pool_size` is 320 pages. At the default `innodb_page_size=16k` this corresponds to 5 MiB. However, now that the `innodb_buffer_pool_size` includes the memory allocated for the block descriptors, the minimum would now be `innodb_buffer_pool_size=6m`.

`Innodb_buffer_pool_resize_status` will be removed. The `SET GLOBAL innodb_buffer_pool_size` operation will block until the buffer pool has been resized or the operation aborted by `KILL`, `SHUTDOWN` or disconnect.

## How can this PR be tested?
This is mostly covered by the regression test suite.

Large-page allocation needs to be tested on Linux and Microsoft Windows. Stress testing with lots of buffer pool resizing during the workload would be helpful.

## PR quality check


dr-m:
This was closed in #3826.

PR #3820, opened 2025-02-06
MDEV-14091 Support password protected SSL key in server.

vaintroub:
Add ssl_passphrase_command server parameter, inspired by Postgres parameter with the same name.

Based on a patch by Hartmut Holzgraefe

- [x] *The Jira issue number for this PR is: MDEV-14091*

## Description
Implement ssl_passphrase_command option for the server, with the same semantic as in Postgres

## Release Notes
New server variable ssl_passphrase_command 

Note, that below Postgres documentation quoted almost literally
"Sets an external command to be invoked when a passphrase for decrypting an SSL file such as a private key needs to be obtained. By default, this parameter is empty, which means the built-in prompting mechanism is used.
The command must print the passphrase to the standard output and exit with code 0. Note that the prompt string will probably contain whitespace, so be sure to quote adequately. newline is stripped from the end of the output if present."

## How can this PR be tested?

PR contains automated tests.

## PR quality check


vaintroub:
Do not like server variables being executable code, closing in lieu of #3840

PR #3823, opened 2025-02-07
MDEV-24948: thd_need_wait_reports() hurts performance

knielsen:
Reduce the cost of thd_rpl_deadlock_check() which is used to do parallel
replication conflict detecton.

1. Cache the value of thd_need_wait_reports() in InnoDB trx_t, and only call
thd_rpl_deadlock_check() if required.

2. Remove the FL_WAITED flag in GTID. This flag was set if a transaction did
any lock wait inside InnoDB on the master. In optimistic parallel
replication on the slave, such transaction would not run in parallel with
any prior transactions. It is not clear what the value of this mechanism is.
If a transaction did not wait on the master, it can still conflict on the
slave. And even if it did have to wait on the master, that does not mean
that it will need to wait on the slave, or that such wait will cause a
conflict or other problem. So waiting for all prior transactions on the
slave in FL_WAITED transactions might slow down things rather than speed up.

3. Only request thd_rpl_deadlock_check() calls for non-replication THDs if
the --binlog-commit-wait-count option is enabled. Such wait might have to
timeout, if the waiting commits are blocking other transactions from joining
the group commit, so the wait reports are used to terminate the wait early.
After this patch, setting --binlog-commit-wait-count non-zero will only have
effect for newly created connections.

With these changes, unless the non-default --binlog-commit-wait-count option
is used, the overhead of thd_rpl_deadlock_check() is reduced to a single
conditional in non-parallel-replication threads.

Signed-off-by: Kristian Nielsen <knielsen@knielsen-hq.org>


storage/innobase/trx/trx0trx.cc
```diff
@@ -385,6 +385,7 @@ void trx_t::free()
   check_foreigns= true;
   assert_freed();
   trx_sys.rw_trx_hash.put_pins(this);
+  need_wait_report= false;
```
dr-m, line 388: It would be cleaner to clear the `need_wait_report` in `trx_t::commit_cleanup()` or similar, and extend `assert_freed()` with `ut_ad(!need_wait_report)`.

storage/innobase/trx/trx0trx.cc
```diff
@@ -544,6 +545,7 @@ void trx_disconnect_prepared(trx_t *trx)
   trx->read_view.close();
   trx_sys.trx_list.freeze();
   trx->is_recovered= true;
+  trx->need_wait_report= false;
```
dr-m, line 548: Zero initialization should be guaranteed by a debug assertion that you should add to `trx_create()`.

storage/innobase/trx/trx0trx.cc
knielsen: @dr-m Hm, I don't understand. `need_wait_report` is only set once in the thread, in `innobase_trx_allocate()` when the trx_t object is first allocated to the thread. But trx_t::commit_cleanup() is called after every commit, right?

The value of `need_wait_report` needs to be preserved until the trx_t object is no longer associated with the thread. Is there a more appropriate place for this than `trx_t::free()` ?

storage/innobase/lock/lock0lock.cc
```diff
@@ -2089,7 +2088,7 @@ dberr_t lock_wait(que_thr_t *thr)
   is not being used. We want to be allow the user to skip
   lock_wait_rpl_report(). */
   const bool rpl= trx->mysql_thd && innodb_deadlock_detect &&
-    thd_need_wait_reports(trx->mysql_thd);
+    trx->need_wait_report;
```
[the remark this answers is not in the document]
knielsen: Agree, `trx->need_wait_report` can only be set when `thd->mysql_thd` is non-null.

knielsen:
Marko Mäkelä ***@***.***> writes:

> @dr-m commented on this pull request.

Thanks Marko, I updated as suggested, except for the one I questioned above.

It does look good to get rid of this overhead where it is not necessary,
even if it changes behaviour a bit for optimistic parallel replication
(removing the heuristics for transactions that waited on the master).

Probably not in 10.6 though. Maybe 11.4?

 - Kristian.

storage/innobase/trx/trx0trx.cc
dr-m: There is a special `trx_pools` that manages transaction objects, so that some memory allocation overhead and fragmentation will be avoided. Most members of `trx_t` are reset on transaction commit, in `trx_t::commit_cleanup()`, and validated in `trx_t::assert_freed()`. If you added the new data member next to `unsigned check_unique_secondary:1`, then I believe that resetting the field in `trx_t::commit_cleanup()` would incur no overhead at all; several bits would be cleared in the same instruction.

knielsen:
Marko Mäkelä ***@***.***> writes:

> @dr-m commented on this pull request.

> There is a special `trx_pools` that manages transaction objects, so
> that some memory allocation overhead and fragmentation will be
> avoided. Most members of `trx_t` are reset on transaction commit, in
> `trx_t::commit_cleanup()`, and validated in `trx_t::assert_freed()`.
> If you added the new data member next to `unsigned
> check_unique_secondary:1`, then I believe that resetting the field in
> `trx_t::commit_cleanup()` would incur no overhead at all; several bits
> would be cleared in the same instruction.

@dr-m, yes, but I don't want to reset the field in `trx_t::commit_cleanup()`.
I need the field to be initialized only when the transaction object is
allocated to a THD, and then cache the value across many commits, until the
thread exits, the THD is de-allocated, and the trx_t object is returned to
the `trx_pools`.

This avoids having to call `thd_need_wait_reports()` for every transaction.
That function is documented to have a return that may be cached for the
duration of the THD.

 - Kristian.

dr-m:
> I don't want to reset the field in `trx_t::commit_cleanup()`. I need the field to be initialized only when the transaction object is allocated to a THD, and then cache the value across many commits, until the thread exits, the THD is de-allocated, and the trx_t object is returned to the `trx_pools`. This avoids having to call `thd_need_wait_reports()` for every transaction. That function is documented to have a return that may be cached for the duration of the THD. - Kristian.

Thank you for the explanation. This is like `trx_t::max_inactive_id` then. It is used as a per-connection cache to avoid costly calls to `find_same_or_older_low()` in `trx_sys_t::find_same_or_older()`. It would be helpful to expand the comment next to the data member to say that similar to `max_inactive_id` it will be cached across reuse, and updated in `innobase_trx_init()`. Even better would be to rename `innobase_trx_init()` to a member function, say, `trx_t::thd_init(THD*) noexcept`.

I think that we should target 10.11 or 11.4 with this. For 10.6 the policy was recently changed; only grave bugs are to be fixed there.

PR #3827, opened 2025-02-11
MDEV-34891 : SST failure occurs when gtid_strict_mode is enabled

janlindstrom:
- [x] *The Jira issue number for this PR is: MDEV-34891*

## Description

Problem was that initial GTID was set on wsrep_before_prepare out-of-order. In practice GTID was set to same as previous executed transaction GTID. In recovery valid GTID was found from prepared transaction and this transaction is committed leading to fact that same GTID was executed twice.

This is fixed by setting invalid GTID at wsrep_before_prepare and later in wsrep_before_commit actual correct GTID is set and this setting is done while we are in commit monitor i.e. assigment is done in order of replication.

In recovery if prepared transaction is found we check its GTID, if it is invalid transaction will be rolled back and if it is valid it will be committed.

Initialize gtid seqno from recovered seqno when bootstrapping a new cluster

Added two test cases for both mariabackup and rsync SST methods to show that GTIDs remain consistent on cluster and that all expected rows are in the table.

Added tests for wsrep GTID recovery with binlog on and off

## Release Notes

## How can this PR be tested?

## PR quality check


sysprg:
Thanks, the fix has  been  merged with the head revision https://github.com/MariaDB/server/commit/b167730499b359f247015447e985319814342bf1

PR #3829, opened 2025-02-11
MDEV-11341 STR_TO_DATE does not return NULL for invalid dates

gengtianuiowa:
## Description

STR_TO_DATE function in MariaDB does date to string conversion for invalid dates strings, e.g.: 1949-02-30, and 1949-06-31, but null is return by other invalid dates such as 1949-01-32 because the checking function returns null for any date whose day is larger than 31, which is too naive and needs to be changed.

This commit changed the rule of check_date in extract_date_time function, which is executed after extracting date and time and check if
the result is valid. Now the conversion will consider the number of days in a month, including skip years.

## How can this PR be tested?

Execute the main suite in mysql-test-run.

## Backward compatibility

This is a new feature introducing stricter date check before a conversion is generated and affect backwards compatibility.

## Copyright

### DIFF
--- a/mysql-test/main/strict.test
+++ b/mysql-test/main/strict.test
@@ -237,11 +237,11 @@ INSERT INTO t1 (col2) VALUES(STR_TO_DATE('31.10.0000 15.30','%d.%m.%Y %H.%i'));
 INSERT INTO t1 (col1) VALUES(STR_TO_DATE('31.0.2004 15.30','%d.%m.%Y %H.%i'));
 --error 1411
 INSERT INTO t1 (col1) VALUES(STR_TO_DATE('0.10.2004 15.30','%d.%m.%Y %H.%i'));
---error 1292
+--error ER_WRONG_VALUE_FOR_TYPE
 INSERT INTO t1 (col1) VALUES(STR_TO_DATE('31.9.2004 15.30','%d.%m.%Y %H.%i'));
 --error 1411
 INSERT INTO t1 (col1) VALUES(STR_TO_DATE('32.10.2004 15.30','%d.%m.%Y %H.%i'));
---error 1292
+--error ER_WRONG_VALUE_FOR_TYPE
 INSERT INTO t1 (col1) VALUES(STR_TO_DATE('29.02.2003 15.30','%d.%m.%Y %H.%i'));
 --error 1411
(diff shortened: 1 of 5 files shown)


ottok:
The test job that is failing here is not labelled 'Required' so we will ignore it for now.

cvicentiu:
@svoj Can you please review this, as @abarkov is quite busy at the moment.

svoj (changes requested):
Very nice. The only thing I'd loved to see here is extended test for `sql_mode=ALLOW_INVALID_DATES`. Just execute the same queries that you've added in this mode.

I will also check with my colleagues which version it should be targeted for.

mysql-test/main/strict.test
```diff
@@ -243,11 +243,11 @@ INSERT INTO t1 (col3) VALUES(STR_TO_DATE('31.10.0000 15.30','%d.%m.%Y %H.%i'));
 INSERT INTO t1 (col3) VALUES(STR_TO_DATE('31.0.2004 15.30','%d.%m.%Y %H.%i'));
 --error 1411
 INSERT INTO t1 (col3) VALUES(STR_TO_DATE('0.10.2004 15.30','%d.%m.%Y %H.%i'));
---error 1292
+--error 1411
 INSERT INTO t1 (col3) VALUES(STR_TO_DATE('31.9.2004 15.30','%d.%m.%Y %H.%i'));
 --error 1411
 INSERT INTO t1 (col3) VALUES(STR_TO_DATE('32.10.2004 15.30','%d.%m.%Y %H.%i'));
---error 1292
+--error 1411
```
svoj, line 250: This should be replaced with `--error ER_WRONG_VALUE_FOR_TYPE`, all occurrences.

gengtianuiowa: Replaced

gengtianuiowa:
> Very nice. The only thing I'd loved to see here is extended test for `sql_mode=ALLOW_INVALID_DATES`. Just execute the same queries that you've added in this mode.
> 
> I will also check with my colleagues which version it should be targeted for.

Hi Svoj, I have resolved all comments above. Please have a look at it when you get chance. Thanks!

svoj (approved):

svoj:
Waiting for `main.timezone` fix to get propagated to main.

PR #3830, opened 2025-02-11
MDEV-9473: Enhance Soundex Functionality to Properly Handle Multi-word Inputs

bardiaHSZD:
## Description

This merge request fixes the SOUNDEX() function's handling of multi-word inputs
to comply with standard specifications and improve SQL compatibility.

Problem:
The current implementation incorrectly processes strings with multiple words:

* Words are concatenated instead of being processed separately
* Output exceeds the standard 4-character length
* No proper word delimitation in output

Example:

  Current:  
```
SOUNDEX('Hello World') → 'H4643'
```

  Fixed:
    
```
SOUNDEX('Hello World') → 'H400 W643'
```

Changes:

1. Code Changes:

    - item_strfunc.h: Isolated single-word logic in Item_func_soundex::soundex
    - item_strfunc.c: 
        * Refactor single-word processing
        * Add multi-word support in Item_func_soundex::val_str
        * Improve formatting
        * Use String data-type, from the data base implementation, instead of std::string

2. Test Updates:

   Modified expected results for multi-word scenarios in:

    - ctype_ucs
    - ctype_utf16
    - ctype_utf16le
    - ctype_utf32
    - func_str

## How can this PR be tested?
The following MTRs passed after the code changes:

- ctype_ucs
- ctype_utf16
- ctype_utf16le
- ctype_utf32
- func_str

## Copyright


sql/item_strfunc.cc
```diff
@@ -2935,6 +2935,7 @@ bool Item_func_current_role::fix_fields(THD *thd, Item **ref)
   return 0;
 }
 
+
```
svoj, line 2938: This is unrelated change, please revert it and all other similar occurrences.

sql/item_strfunc.cc
```diff
@@ -2984,17 +2985,23 @@ static bool my_uni_isalpha(int wc)
...
   DBUG_ASSERT(fixed());
-  String *res= args[0]->val_str(&tmp_value);
-  char last_ch,ch;
   CHARSET_INFO *cs= collation.collation;
+  if (!input_str.ptr() || input_str.length() == 0) {
+    str->copy(NULL, 0, collation.collation);  // Set str to empty string
+    return str;
+  }
+
+  String *res = new String(input_str.c_ptr(), input_str.length(), cs);//args[0]->val_str(&tmp_value);
```
svoj, line 2999: It should be `res= new`, that is no space before `=`.
Commented out code is not permitted.
Line should be under 80 characters.

It applies to other similar occurrences.

sql/item_strfunc.cc
```diff
@@ -3091,6 +3100,75 @@ String *Item_func_soundex::val_str(String *str)
...
+  for (; input_ptr < input_end; ++input_ptr) 
+  {
+    char current_char = *input_ptr;
+    bool is_delimiter = (delimiter != '\0') && 
+                       (current_char == delimiter || 
+                        std::isspace(static_cast<unsigned char>(current_char)));
+
+    if (is_delimiter) 
+    { 
+        // If a delimiter is found, process the current word
```
svoj, line 3134: Wrong indentation: should be 2 spaces.

svoj (changes requested):

sql/item_strfunc.cc
bardiaHSZD: The formatting issues were fixed.

sql/item_strfunc.cc
bardiaHSZD: The formatting issues were fixed.

sql/item_strfunc.cc
bardiaHSZD: The change has been reverted.

svoj: There're 3 more unrelated occurrences that either remove lines or add spaces, please revert them as well.

bardiaHSZD: All the empty lines were removed, which were previously added as a result of the changes to code to enhance the soundex implementation.

svoj: Please double check the diff, there're even more unrelated changes now.

sql/item_strfunc.cc
```diff
@@ -3075,9 +3080,10 @@ String *Item_func_soundex::val_str(String *str)
       to+= rc;
       nchars++;
       last_ch= ch;  // save code of last input letter
-    }               // for next double-letter check
+      if (nchars >= 4)
+        break; // Soundex code is 4 characters
+    }
   }
-
```
svoj, line 3080: ^

sql/item_strfunc.cc
```diff
@@ -6330,4 +6395,4 @@ longlong Item_func_wsrep_sync_wait_upto::val_int()
   return ret;
 }
 
-#endif /* WITH_WSREP */
+#endif /* WITH_WSREP */
```
svoj, line 6398: ^

sql/item_strfunc.h
```diff
@@ -1338,10 +1338,10 @@ class Item_func_current_role :public Item_func_sysconst
   { return get_item_copy<Item_func_current_role>(thd, this); }
 };
 
-
```
svoj, line 1341: ^

sql/item_strfunc.h
```diff
@@ -1355,7 +1355,6 @@ class Item_func_soundex :public Item_str_func
   { return get_item_copy<Item_func_soundex>(thd, this); }
 };
 
-
```
svoj, line 1358: ^

svoj (changes requested):

sql/item_strfunc.cc
bardiaHSZD: Thank you very much for your feedback all the noted issues were resolved.

sql/item_strfunc.h
bardiaHSZD: Thank you very much for your feedback all the noted issues were resolved.

sql/item_strfunc.h
bardiaHSZD: Thank you very much for your feedback all the noted issues were resolved.

svoj (changes requested):

We have to keep `SOUNDEX()` behaviour compatible with older version. Suggested alternative: `SOUNDEX(str[, soundex_flavor])`, where `flavor=0` means current implementation, `flavor=1` - word split implementation.

Please adjust your work according to this specification.

bardiaHSZD:
The requested changes have been applied. The corresponding unit tests and the pull request message were updated. The reason for the current build failure is unrelated to the current implementation and it is due to the "main.timezone" test.

svoj:
The `main.timezone` test was fixed a while ago. If you rebase on top of `main` these failures should be gone.
Also you've created merge commit, it should be removed eventually.

For some reason this PR is still in "Changes requested" state, thus it went off of our radar.

bardiaHSZD:
The requested changes have been applied.

mysql-test/main/ctype_ucs.result
```diff
@@ -1348,6 +1348,21 @@ select hex(soundex(_ucs2 0x00BF00C0));
...
+set character_set_connection=ucs2;
+select soundex('','0'),soundex('he','0'),soundex('hello all folks','0'),soundex('#3556 in bugdb','0');
+soundex('','0')	soundex('he','0')	soundex('hello all folks','0')	soundex('#3556 in bugdb','0')
+	H000	H4142	I51231
+select hex(soundex('','0')),hex(soundex('he','0')),hex(soundex('hello all folks','0')),hex(soundex('#3556 in bugdb','0'));
+hex(soundex('','0'))	hex(soundex('he','0'))	hex(soundex('hello all folks','0'))	hex(soundex('#3556 in bugdb','0'))
+	0048003000300030	00480034003100340032	004900350031003200330031
+select soundex('','1'),soundex('he','1'),soundex('hello all folks','1'),soundex('#3556 in bugdb','1');
+soundex('','1')	soundex('he','1')	soundex('hello all folks','1')	soundex('#3556 in bugdb','1')
+	H000	H400?????F420	?????B231
```
vuvova, line 1361: why?

mysql-test/main/ctype_ucs.result
```diff
@@ -1348,6 +1348,21 @@ select hex(soundex(_ucs2 0x00BF00C0));
 hex(soundex(_ucs2 0x00BF00C0))
 00C0003000300030
 set names latin1;
+set names latin1;
+set character_set_connection=ucs2;
+select soundex('','0'),soundex('he','0'),soundex('hello all folks','0'),soundex('#3556 in bugdb','0');
```
vuvova, line 1353: why do you always pass the second argument as a string?

mysql-test/main/func_str.result
```diff
@@ -222,6 +222,12 @@ this is a REAL test
 select soundex(''),soundex('he'),soundex('hello all folks'),soundex('#3556 in bugdb');
 soundex('')	soundex('he')	soundex('hello all folks')	soundex('#3556 in bugdb')
 	H000	H4142	I51231
+select soundex('','0'),soundex('he','0'),soundex('hello all folks','0'),soundex('#3556 in bugdb','0');
+soundex('','0')	soundex('he','0')	soundex('hello all folks','0')	soundex('#3556 in bugdb','0')
+	H000	H4142	I51231
+select soundex('','1'),soundex('he','1'),soundex('hello all folks','1'),soundex('#3556 in bugdb','1');
+soundex('','1')	soundex('he','1')	soundex('hello all folks','1')	soundex('#3556 in bugdb','1')
+	H000	H400 A400 F420	 I500 B231
```
vuvova, line 230: also add, like
```sql
select soundex('hello'), soundex('all'),soundex('folks')
```
to show the result is the same

sql/item_create.cc
```diff
@@ -5790,10 +5791,43 @@ Create_func_sleep::create_1_arg(THD *thd, Item *arg1)
...
+
+  if (item_list != NULL)
+  arg_count= item_list->elements;
+
+  switch (arg_count) {
+  case 0:
+  {
+    my_error(ER_WRONG_PARAMCOUNT_TO_NATIVE_FCT, MYF(0), name->str);
+    break;
+  }
```
vuvova, line 5809: remove `case 0`

sql/item_create.cc
```diff
@@ -5790,10 +5791,43 @@ Create_func_sleep::create_1_arg(THD *thd, Item *arg1)
...
+  case 2:
+  {
+    Item *param_1= item_list->pop();
+    Item *param_2= item_list->pop();
+    func= new (thd->mem_root) Item_func_soundex(thd, param_1, param_2);
+    break;
+  }
+  default:
+  {
```
vuvova, line 5824: add an assert

sql/item_create.cc
```diff
@@ -5790,10 +5791,43 @@ Create_func_sleep::create_1_arg(THD *thd, Item *arg1)
...
+Item* 
+Create_func_soundex::create_native(THD *thd, const LEX_CSTRING *name,
+  List<Item> *item_list)
 {
-  return new (thd->mem_root) Item_func_soundex(thd, arg1);
+  Item *func= NULL;
+  int arg_count= 0;
+
+  if (item_list != NULL)
+  arg_count= item_list->elements;
```
vuvova, line 5802: formatting. here and everywhere

vuvova (changes requested):

svoj:
@bardiaHSZD Hi! It's been 2 months since @vuvova requested changes. Are you still planning to complete this task?

bardiaHSZD:
I am actively working onthe issue. Will push another update in few days time.

mysql-test/main/ctype_ucs.result
bardiaHSZD: All the unicode related issues were resolved.

mysql-test/main/ctype_ucs.result
bardiaHSZD: The MTRs were refined so that it contains both strings and integers as the second argument.

mysql-test/main/func_str.result
bardiaHSZD: The requested tests were added.

sql/item_create.cc
bardiaHSZD: Case 0 has been removed

sql/item_create.cc
bardiaHSZD: The formatting has been fixed.

sql/item_create.cc
bardiaHSZD: Asserts have been added.

bardiaHSZD:
@svoj @vuvova Would you please review the updated code?

bardiaHSZD:
Most of the build/compile issues were resolved and the failing tests are un-related to my change. cc: @svoj @vuvova

sql/item_strfunc.cc
```diff
@@ -2947,6 +2947,10 @@ bool Item_func_soundex::fix_length_and_dec(THD *thd)
   if (agg_arg_charsets_for_string_result(collation, args, 1))
     return TRUE;
   DBUG_ASSERT(collation.collation != NULL);
+  if (collation.collation->mbmaxlen > 1)
```
gkodinov, line 2950: Feel free to ignore this: SOUNDEX() should never produce anything but ASCII codes. I question the usefulness of preserving the first argument's collation as a result collation. 
If we operate under the above assumption the whole algorithm will also be simpler: convert to ascii and then do the SOUNDEX on the ascii itself.

sql/item_strfunc.cc
```diff
@@ -3096,6 +3197,93 @@ String *Item_func_soundex::val_str(String *str)
...
+    {
+      if (args[1]->val_int() == 0)
+        return Item_func_soundex::val_str_legacy(str);
+      else if (args[1]->val_int() != 1)
+      {
+        push_warning_printf(current_thd,
+                          Sql_condition::WARN_LEVEL_WARN,
+                          ER_WRONG_ARGUMENTS,
+                          "Invalid argument for SOUNDEX function. Expected 0 or 1.");
+        return NULL;
```
gkodinov, line 3217: Here and everyplace else you're returning a NULL: I believe you also need to set the null_value to 1 here

sql/item_strfunc.cc
```diff
@@ -3096,6 +3197,93 @@ String *Item_func_soundex::val_str(String *str)
...
+    int rc;
+
+    /* Process input string word by word using proper charset functions */
+    while (input_ptr < input_end) 
+    {
+      /* Skip leading spaces */
+      while (input_ptr < input_end) {
+        if ((rc= cs->mb_wc(&wc, (uchar*)input_ptr, (uchar*)input_end)) <= 0)
+            break;
+        if (wc != ' ')
```
gkodinov, line 3252: The "legacy" SOUNDEX() would skip all the leading symbols that are not ascii letters. Please specify that the behavior of the new one is different or make it be the same.

sql/item_strfunc.cc
```diff
@@ -3096,6 +3197,93 @@ String *Item_func_soundex::val_str(String *str)
...
+            break;
+        input_ptr+= rc;
+      }
+      
+      /* Extract word */
+      word_buffer.length(0);
+      while (input_ptr < input_end) {
```
gkodinov, line 3259: I'd use my_strchr() here to find the space.

mysql-test/main/ctype_ucs.test
```diff
@@ -617,7 +617,22 @@ select hex(soundex(_ucs2 0x041004110412));
 select hex(soundex(_ucs2 0x00BF00C0));
 set names latin1;
 --enable_service_connection
-
+#
+# MDEV-9473 SOUNDEX feature for multiple words
+#
+--disable_service_connection
+set names latin1;
+set character_set_connection=ucs2;
```
gkodinov, line 627: Please consider adding tests with non-ascii strings.
Also, please consider adding tests with leading number in the 2nd word.

mysql-test/main/ctype_utf16.test
```diff
@@ -498,6 +498,23 @@ select hex(soundex(_utf16 0x041004110412));
 select hex(soundex(_utf16 0x00BF00C0));
 set names latin1;
 
+#
+# MDEV-9473 SOUNDEX feature for multiple words
```
gkodinov, line 502: I'd also consider moving the test to a separate .inc file and including just a single version of it in all of the tests.

sql/item_strfunc.cc
```diff
@@ -2991,7 +2995,104 @@ static bool my_uni_isalpha(int wc)
...
+    CHARSET_INFO *cs= collation.collation;
+    
+    /* Check for empty input */
+    if (!input_str.ptr() || input_str.length() == 0)
+    {
+      str->copy(NULL, 0, collation.collation);
+      return str;
+    }
+
+    String *res= new String(input_str.c_ptr(), input_str.length(), cs);
```
gkodinov, line 3010: Any specific reason you need to allocate this one on the heap? 
Why can't you just do "String res(input_str.c_ptr(), input_str.length(), cs)" ?
Because you're leaking it atm. FWIW. I do not see how res is used actually. Maybe remove it altogether?

sql/item_strfunc.cc
```diff
@@ -3096,6 +3197,93 @@ String *Item_func_soundex::val_str(String *str)
...
 }
 
+String *Item_func_soundex::val_str(String *str)
+{
+    DBUG_ASSERT(fixed());
+
+    // Handle argument count and validation
+    if (arg_count == 1)
+        return Item_func_soundex::val_str_legacy(str);
+    else if (arg_count == 2)
```
gkodinov, line 3207: Feel free to ignore this: I feel like there's no real benefit from merging two very different implementations of SOUNDEX() together and using a parameter to pick the variant. None of the databases out there does this. 
Furthermore, you're deferring the decision which one to use to runtime. That's fine for a single invocation, but what about calling SOUNDEX() a million times per query? 

As little as the check takes, it's still not 0 CPU and time. 

I'd personally do a SOUNDEX_WORDS or something like that as a separate function. Or, at least do the determination which one to use at compile time, by requiring that the 2nd argument is an actual SQL constant and not an expression that can potentially depend on subqueries, etc.

sql/item_strfunc.cc
bardiaHSZD: Changed to stack allocation. "res" is being used by "from" and "end".

sql/item_strfunc.cc
bardiaHSZD: Added null_value = 1 before returning NULL in the corresponding places.

gkodinov:
After a discussion with the MariaDB developers I regretfully need to inform you that we will not be processing this contribution any further. This is not a reflection on the quality of the contribution, it's just that the backward compatibility is paramount in this case and there's no clear use case for the enhanced functionality. 
Thank you for your time that you've spent on this and for your understanding.

PR #3833, opened 2025-02-12
MDEV-36074 main.multidelete_engine missing result file

DaveGosselin-MariaDB:
Record the required result file and make the test more interesting, too


vaintroub (approved):
looks good to me

PR #3835, opened 2025-02-13
MDEV-36082 Race condition between log_t::resize_start() and log_t::resize_abort()

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36082*
## Description
`log_t::writer_update()`: Enable `log_writer_resizing` only if `resize_lsn>1`, to ensure that `log_t::resize_abort()` will not choose the wrong `log_sys.log_writer`.

`log_t::resize_start()`: Before invoking `writer_update()`, assign `resize_lsn>1` if applicable, so that the resizing will start correctly.
## Release Notes
When `SET GLOBAL innodb_log_file_size` is frequently attempted and interrupted during a write workload, InnoDB could crash due to attempting to write to a non-existing copy of a log.
## How can this PR be tested?
Stress testing with RQG generating `SET GLOBAL innodb_log_file_size` and `KILL` statements.
## PR quality check

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -353,7 +353,7 @@ bool log_t::attach(log_file_t file, os_offset_t size)
 # endif
       buf= static_cast<byte*>(ptr);
       max_buf_free= 1;
-      writer_update();
+      writer_update(false);
 # ifdef HAVE_PMEM
       if (is_pmem)
         return true;
@@ -395,7 +395,7 @@ bool log_t::attach(log_file_t file, os_offset_t size)
   TRASH_ALLOC(buf, buf_size);
   TRASH_ALLOC(flush_buf, buf_size);
   max_buf_free= buf_size / LOG_BUF_FLUSH_RATIO - LOG_BUF_FLUSH_MARGIN;
-  writer_update();
+  writer_update(false);
   memset_aligned<512>(checkpoint_buf, 0, write_size);
 
  func_exit:
@@ -570,31 +570,35 @@ void log_t::set_buffered(bool buffered)
 
 /** Start resizing the log and release the exclusive latch.
 @param size  requested new file_size
+@param thd   the current thread identifier
 @return whether the resizing was started successfully */
-log_t::resize_start_status log_t::resize_start(os_offset_t size) noexcept
+log_t::resize_start_status log_t::resize_start(os_offset_t size, void *thd)
+  noexcept
 {
   ut_ad(size >= 4U << 20);
   ut_ad(!(size & 4095));
   ut_ad(!srv_read_only_mode);
+  ut_ad(thd);
 
   log_resize_acquire();
 
-  resize_start_status status= RESIZE_NO_CHANGE;
-  lsn_t start_lsn{0};
-#ifdef HAVE_PMEM
-  bool is_pmem{false};
-#endif
+  resize_start_status status;
 
-  if (resize_in_progress())
+  if (size == file_size)
+    status= RESIZE_NO_CHANGE;
+  else if (resize_in_progress())
     status= RESIZE_IN_PROGRESS;
-  else if (size != file_size)
+  else
   {
+    lsn_t start_lsn;
     ut_ad(!resize_in_progress());
     ut_ad(!resize_log.is_opened());
     ut_ad(!resize_buf);
     ut_ad(!resize_flush_buf);
+    ut_ad(!resize_initiator);
     std::string path{get_log_file_path("ib_logfile101")};
     bool success;
+    resize_initiator= thd;
     resize_lsn.store(1, std::memory_order_relaxed);
     resize_target= 0;
     resize_log.m_file=
@@ -612,6 +616,7 @@ log_t::resize_start_status log_t::resize_start(os_offset_t size) noexcept
 #ifdef HAVE_PMEM
       else if (is_mmap())
       {
+        bool is_pmem{false};
         ptr= ::log_mmap(resize_log.m_file, is_pmem, size);
 
         if (ptr == MAP_FAILED)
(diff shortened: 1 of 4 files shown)


storage/innobase/log/log0log.cc
```diff
@@ -1193,7 +1194,7 @@ ATTRIBUTE_COLD static lsn_t log_writer_resizing() noexcept
 void log_t::writer_update() noexcept
 {
   ut_ad(latch_have_wr());
-  writer= resize_in_progress() ? log_writer_resizing : log_writer;
+  writer= resize_in_progress() > 1 ? log_writer_resizing : log_writer;
   mtr_t::finisher_update();
```
mariadb-DebarunBanerjee, line 1198: This fix would help fixing the assert but the root of the problem seems to lie elsewhere. The problem is that the constraint that "I should abort my own resize" is violated. Unless fixed, it could cause other issues e.g. we could abort a future resize operation which would return success to end user. 

We should fix `log_t::resize_abort` and `innodb_log_file_size_update` to ensure the constraint.

1. `log_sys.resize_start()` : Have one OUT parameter to return the `resize_lsn` and use it instead of reading the resize_lsn again which could belong to some future resize operation.
2. Pass own `resize_lsn` to `log_sys.resize_abort()`
3. In log_t::resize_abort() after log_resize_acquire(), skip abort if resize lsn doesn't match the passed resize_lsn.

This should ensure that we never wait for or abort a future log resize operation.

mariadb-DebarunBanerjee (commented):
Hi @dr-m,
I checked the RR trace and it looks to me that we need to solve the root cause to avoid any resize operation attempting to abort a future resize. Otherwise the issue may not be fixed completely. I have added more details to the comment. Please have a look.

storage/innobase/log/log0log.cc
dr-m: The `log_sys.resize_lsn` may be changed by `log_t::resize_write()` or `log_t::resize_write_buf()` while log resizing is in progress. This is especially likely when `SET GLOBAL innodb_log_file_size` is attempting to reduce the size.

That is, if we implemented exact matching of `log_sys.resize_lsn` in `log_t::resize_abort()`, we could end up in a situation where log resizing cannot be aborted. I see that in `innodb_log_file_size_update()` we have some inexact matching that could lead to not waiting long enough in case our own `log_sys.resize_lsn` had been advanced. I will try to revise this.

dr-m: By design, at most one `log_t::resize_start()` may succeed at a time. If log resizing has already been started or is in progress, a subsequent `log_t::resize_start()` will return `log_t::RESIZE_IN_PROGRESS`. This is what happened also in the `rr replay` trace of the failure that this patch is addressing. The only caller of `log_t::resize_start()` (`innodb_log_file_size_update()`) will only invoke `log_t::resize_abort()` if `log_t::RESIZE_STARTED` had been returned.

Another scenario that we must keep in mind is that the current log resizing had been completed (`log_t::write_checkpoint()` had assigned `log_sys.resize_lsn=0`) and another thread had initiated log resizing. In this scenario, the previous `innodb_log_file_size_update()` must complete and the subsequent one must wait for the resizing to be completed or aborted.

Possibly the most straightforward solution would be to remember the thread identifier that successfully initiated the log resizing. `innodb_log_file_size_update()` will have to finish as soon as it notices that log resizing had been completed, or it is not the current initiator.

mariadb-DebarunBanerjee: Consider the flow 
Thread-1 :
innodb_log_file_size_update

log_sys.resize_start() has returned log_t::RESIZE_STARTED. At this point we are trying to get the resize lsn by
```
case log_t::RESIZE_STARTED:
const lsn_t start{log_sys.resize_in_progress()};
```
Here the intention is to get our own resize lsn but it could happen that our resize operation is already finished and another thread has started a resize operation. So we may get the resize_lsn for the resize started by second thread not our own.

> 1. To correct this issue, we can easily return our own resize_lsn by log_sys.resize_start().

Do you agree ?

Secondly, if I am killed, we call abort.
```
18596         if (thd_kill_level(thd))
18597         {
18598           log_sys.resize_abort();
18599           break;
18600         }
```
Again, it might happen here that our resize operation is over and another one has started by some other thread and we certainly don't want to abort that. 

> 2. The solution should be to pass our own identity (start_lsn - the resize_lsn returned by log_sys.resize_start() as suggested in earlier comment) to resize_abort. The resize_abort can compare it with current resize lsn after log_resize_acquire() and simply exit if  it doesn't match i.e. if our resize is finished and new one is started.

Right ? I think this should be a sound solution.

mariadb-DebarunBanerjee: > The log_sys.resize_lsn may be changed by log_t::resize_write() or log_t::resize_write_buf() while log resizing is in progress.

hmm... This could be an issue for the solution I proposed. So, we are not able to associate resize_lsn as the identity of a resize operation.

dr-m: c6416d0e71dac953b6748cec91439372d1c2f008 tries to address this.

storage/innobase/log/log0log.cc
```diff
@@ -570,31 +570,35 @@ void log_t::set_buffered(bool buffered)
...
-#endif
+  resize_start_status status;
 
+  if (size == file_size)
+    status= RESIZE_NO_CHANGE;
   if (resize_in_progress())
     status= RESIZE_IN_PROGRESS;
-  else if (size != file_size)
+  else
   {
```
mariadb-DebarunBanerjee, line 592: I think you meant to use `else if/else if/else` here. Otherwise, we would enter the else block for (size == file_size) when no resize is in progress. Please check.
```
  if (size == file_size)
    status= RESIZE_NO_CHANGE;

  else if (resize_in_progress())
    status= RESIZE_IN_PROGRESS;

  else
  {
  ...
  }
```

mariadb-DebarunBanerjee (approved):
Thanks for the addressing the comment. The solution looks nice and correct to me.
I have added one straight forward but important comment.

storage/innobase/log/log0log.cc
dr-m: You’re right, I indeed meant `else if (resize_in_progress())`. My bad, will fix.

PR #3836, opened 2025-02-13
MDEV-35217: Provide information for debugging parallel replication

andrelkin:
Implement an option --slave-parallel-print-all-deadlocks to debug parallel replication conflicts. This option enables additional information in the server's error log when a parallel replication conflict or deadlock occurs, similar to --innodb-print-all-deadlocks. This should be very useful to help investigate the cause of problems with parallel replication, including helping to find the root cause of any bugs.

The information printed includes the GTID and thread info of conflicting threads; the chain of wait_for_prior_commits; and the SHOW ENGINE STATUS of engines participating in the current transaction.

Setting @@slave_parallel_print_all_deadlocks to 1 will output the information when an unexpected conflict occurs; this means an event group that needed to retry more than once, or any lock wait timeout. Setting @@slave_parallel_print_all_deadlocks to 2 will output the information for any detected conflict.

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/sql/sql_class.cc
+++ b/sql/sql_class.cc
@@ -5191,6 +5191,27 @@ thd_rpl_deadlock_check(MYSQL_THD thd, MYSQL_THD other_thd)
     releasing the lock for this transaction so replication can proceed.
   */
 #ifdef HAVE_REPLICATION
+  if (unlikely(opt_slave_parallel_print_all_deadlocks) &&
+      (other_rgi->trans_retries >= 1 ||
+       opt_slave_parallel_print_all_deadlocks == 2))
+  {
+    char thd_info[400];
+    char other_thd_info[400];
+    thd_get_error_context_description(thd, thd_info, sizeof(thd_info), 350);
+    thd_get_error_context_description(other_thd, other_thd_info,
+                                      sizeof(other_thd_info), 350);
+    sql_print_information("Slave SQL thread: Deadlock detected, aborting "
+                          "second event group:\nGTID %u-%u-%llu: %s\n"
+                          "GTID %u-%u-%llu: %s",
+                          rgi->current_gtid.domain_id,
+                          rgi->current_gtid.server_id,
+                          (ulonglong)rgi->current_gtid.seq_no,
+                          thd_info,
+                          other_rgi->current_gtid.domain_id,
+                          other_rgi->current_gtid.server_id,
+                          (ulonglong)other_rgi->current_gtid.seq_no,
+                          other_thd_info);
+  }
   slave_background_kill_request(other_thd);
 #endif
   return 1;
@@ -7560,6 +7581,7 @@ wait_for_commit::reinit()
   next_subsequent_commit= NULL;
   waitee.store(NULL, std::memory_order_relaxed);
   opaque_pointer= NULL;
+  owner_thd= NULL;
   wakeup_error= 0;
   wakeup_subsequent_commits_running= false;
   commit_started= false;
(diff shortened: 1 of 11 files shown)


sql/sql_class.cc
```diff
@@ -5191,6 +5191,27 @@ thd_rpl_deadlock_check(MYSQL_THD thd, MYSQL_THD other_thd)
...
+  if (unlikely(opt_slave_parallel_print_all_deadlocks) &&
+      (other_rgi->trans_retries >= 1 ||
+       opt_slave_parallel_print_all_deadlocks == 2))
+  {
+    char thd_info[400];
+    char other_thd_info[400];
+    thd_get_error_context_description(thd, thd_info, sizeof(thd_info), 350);
+    thd_get_error_context_description(other_thd, other_thd_info,
+                                      sizeof(other_thd_info), 350);
+    sql_print_information("Slave SQL thread: Deadlock detected, aborting "
```
andrelkin, line 5203: Adding identifiers of the failed event and a record that either was not found, or locked would certainly be useful.
Sure the failed event part is trivial. To find the key whose searching for ended up in timeout is simple at least for ROW format. Perhaps we could be content with that, unless a way can be found to extract that info in the STATEMENT format.

andrelkin (commented):
I like the idea. If it's feasible let's add couple of more things to the report.

knielsen:

> @andrelkin commented on this pull request.
>
> I like the idea. If it's feasible let's add couple of more things to the report.

> Adding identifiers of the failed event and a record that either was not found, or locked would certainly be useful.
> Sure the failed event part is trivial. To find the key whose searching for ended up in timeout is simple at least for ROW format. Perhaps we could be content with that, unless a way can be found to extract that info in the STATEMENT format.

I'm not sure what you have in mind here.

The failed event - we have the GTID (the event group), you mean the specific
event in that event group that failed? How would you identify the event?
I was thinking of `Log_event::pack_info()` like in `SHOW BINLOG EVENTS`, but
that uses the `Protocol` class directly to send the description to a client,
and doesn't seem easy to use for debug output.

The key for the conflicting lock is already printed by the InnoDB debug
output I think, or did you have something else specifically in mind?

 - Kristian.

andrelkin (commented):
The commit is good to go even without [a suggestion to print primary key values](https://github.com/MariaDB/server/pull/3836/commits/f75fb4757a46c6554e007754144e66ae4654d4f9#r1954881422).

andrelkin:
> The key for the conflicting lock is already printed by the InnoDB debug output I think, or did you have something else specifically in mind?

As you know Innodb does not register the parallel replication type of deadlock. It won't print for us.
I thought of this stack
```
slave_background_kill_request
thd_rpl_deadlock_check
DeadlockChecker::search
DeadlockChecker::check_and_resolve
lock_rec_enqueue_waiting
lock_rec_lock
lock_clust_rec_read_check_and_lock
sel_set_rec_lock
row_search_mvcc
ha_innobase::index_read
ha_innobase::rnd_pos
handler::ha_rnd_pos
handler::rnd_pos_by_record
handler::ha_rnd_pos_by_record
Rows_log_event::find_row
Update_rows_log_event::do_exec_row
```
 in which `find_row` invokes [handler::print_error()](https://github.com/MariaDB/server/blob/knielsen_rpl_print_deadlocks/sql/log_event.cc#L14247), that might help to print what I request like it's done in
`case HA_ERR_FOUND_DUPP_KEY: ... print_keydup_error()`. 
I did not check inside `print_keydup_error()` but it feels it would work too here.

andrelkin:
Added Kristian to approve this PR unless Brandon will do that.

PR #3837, opened 2025-02-13
MDEV-34413 Index Condition Pushdown for reverse-ordered scans

DaveGosselin-MariaDB:
This PR contains two commits.

### **Second commit**:
Allows index condition pushdown for reverse ordered scans, a previously
disabled feature due to poor performance.  This patch adds a new
API to the handler class called set_end_range which allows callers to
tell the handler what the end of the index range will be when scanning.
Combined with a pushed index condition, the handler can scan the index
efficiently and not read beyond the end of the given range.  When
checking if the pushed index condition matches, the handler will also
check if scanning has reached the end of the provided range and stop if
so.

If we instead only enabled ICP for reverse ordered scans without
also calling this new API, then the handler would perform unnecessary
index condition checks.  In fact this would continue until the end of
the index is reached.

These changes are agnostic of storage engine.  That is, any storage
engine that supports index condition pushdown will inhereit this new
behavior as it is implemented in the SQL and storage engine
API layers.

The partitioned tables storage meta-engine (ha_partition) adds an
override of set_end_range which recursively calls set_end_range on its
child storage engine (handler) implementations.

This commit updates the test made in an earlier commit to show that
ICP matches happen for the reverse ordered case.

This patch is based on changes written by Olav Sandstaa in
MySQL commit da1d92fd46071cd86de61058b6ea39fd9affcd87

### **First commit**:
Adds tests which show that ICP was not enabled for reverse-ordered scans prior to this mdev.  The second commit for this same mdev records again these same tests, showing that ICP for reverse-ordered scans is enabled and working.

### DIFF
--- a/sql/ha_partition.cc
+++ b/sql/ha_partition.cc
@@ -6367,7 +6367,7 @@ int ha_partition::read_range_first(const key_range *start_key,
 
   m_ordered= sorted;
   eq_range= eq_range_arg;
-  set_end_range(end_key);
+  set_end_range(end_key, RANGE_SCAN_ASC);
 
   range_key_part= m_curr_key_info[0]->key_part;
   if (start_key)
@@ -6402,6 +6402,17 @@ int ha_partition::read_range_next()
   }
   DBUG_RETURN(handle_unordered_next(table->record[0], eq_range));
 }
+
+
+void ha_partition::set_end_range(const key_range *end_key,
+                                 enum_range_scan_direction direction)
+{
+  for (uint i= bitmap_get_first_set(&m_part_info->read_partitions);
+       i < m_tot_parts;
+       i= bitmap_get_next_set(&m_part_info->read_partitions, i))
+    m_file[i]->set_end_range(end_key, direction);
+}
+
 
 /**
    Create a copy of all keys used by multi_range_read()
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -7132,7 +7132,7 @@ int handler::read_range_first(const key_range *start_key,
   DBUG_ENTER("handler::read_range_first");
 
   eq_range= eq_range_arg;
-  set_end_range(end_key);
+  set_end_range(end_key, RANGE_SCAN_ASC);
   range_key_part= table->key_info[active_index].key_part;
 
   if (!start_key)			// Read first record
@@ -7208,9 +7208,17 @@ int handler::read_range_next()
 }
 
 
-void handler::set_end_range(const key_range *end_key)
+/*
+  @brief
+    Inform the Storage Engine about the end of range to be scanned.
+    See opt_index_cond_pushdown.cc, "End-of-range checks".
+*/
+
+void handler::set_end_range(const key_range *end_key,
+                            enum_range_scan_direction direction)
 {
   end_range= 0;
+  range_scan_direction= direction;
   if (end_key)
   {
     end_range= &save_end_range;
@@ -7218,6 +7226,7 @@ void handler::set_end_range(const key_range *end_key)
     key_compare_result_on_equal=
       ((end_key->flag == HA_READ_BEFORE_KEY) ? 1 :
        (end_key->flag == HA_READ_AFTER_KEY) ? -1 : 0);
+    range_key_part= table->key_info[active_index].key_part;
   }
 }
 
@@ -7250,8 +7259,11 @@ int handler::compare_key(key_range *range)
 
 
 /*
-  Same as compare_key() but doesn't check have in_range_check_pushed_down.
-  This is used by index condition pushdown implementation.
+  Same as compare_key() but
+  - doesn't check in_range_check_pushed_down,
+  - supports reverse index scans.
+
+  This is used by Index Condition Pushdown implementation.
 */
 
 int handler::compare_key2(key_range *range) const
@@ -7262,6 +7274,8 @@ int handler::compare_key2(key_range *range) const
   cmp= key_cmp(range_key_part, range->key, range->length);
   if (!cmp)
     cmp= key_compare_result_on_equal;
+  if (range_scan_direction == RANGE_SCAN_DESC)
+    cmp= -cmp;
   return cmp;
 }
 
@@ -7286,6 +7300,10 @@ extern "C" check_result_t handler_index_cond_check(void* h_arg)
     if (killed > abort_at)
       return CHECK_ABORTED_BY_USER;
   }
+  /*
+    Before checking the Pushed Index Condition, check if we went out of range.
+    See opt_index_cond_pushdown.cc, "End-of-range checks".
+  */
   if (unlikely(h->end_range) && h->compare_key2(h->end_range) > 0)
     return CHECK_OUT_OF_RANGE;
   h->increment_statistics(&SSV::ha_icp_attempts);
--- a/sql/handler.h
+++ b/sql/handler.h
@@ -3244,6 +3244,17 @@ class handler :public Sql_alloc
 {
 public:
   typedef ulonglong Table_flags;
+
+  /*
+    The direction of the current range or index scan. This is used by
+    the ICP implementation to determine if it has reached the end
+    of the current range.
+  */
+  enum enum_range_scan_direction {
+    RANGE_SCAN_ASC,
+    RANGE_SCAN_DESC
+  };
+
 protected:
   TABLE_SHARE *table_share;   /* The table definition */
   TABLE *table;               /* The current open table */
@@ -3259,6 +3270,7 @@ protected:
   */
   ha_handler_stats active_handler_stats;
   void set_handler_stats();
+
 public:
   handlerton *ht;               /* storage engine of this handler */
   OPTIMIZER_COSTS *costs;       /* Points to table->share->costs */
@@ -3285,7 +3297,11 @@ public:
 
   KEY_MULTI_RANGE mrr_cur_range;
 
-  /** The following are for read_range() */
+private:
+  /* Used by Index Condition Pushdown, handler_index_cond_check()/compare_key2() */
+  enum_range_scan_direction range_scan_direction{RANGE_SCAN_ASC};
+public:
+  /** The following are for read_range_first/next() and ICP */
   key_range save_end_range, *end_range;
   KEY_PART_INFO *range_key_part;
   int key_compare_result_on_equal;
@@ -3535,7 +3551,8 @@ public:
     DBUG_ASSERT(inited==INDEX);
     inited=       NONE;
     active_index= MAX_KEY;
-    end_range=    NULL;
+    end_range= NULL;
+    range_scan_direction= RANGE_SCAN_ASC;
     DBUG_RETURN(index_end());
   }
   /* This is called after index_init() if we need to do a index scan */
@@ -4335,7 +4352,8 @@ public:
                                const key_range *end_key,
                                bool eq_range, bool sorted);
   virtual int read_range_next();
-  void set_end_range(const key_range *end_key);
+  virtual void set_end_range(const key_range *end_key,
+                             enum_range_scan_direction direction = RANGE_SCAN_ASC);
   int compare_key(key_range *range);
   int compare_key2(key_range *range) const;
   virtual int ft_init() { return HA_ERR_WRONG_COMMAND; }
--- a/sql/opt_range.cc
+++ b/sql/opt_range.cc
@@ -13766,16 +13766,24 @@ int QUICK_SELECT_DESC::get_next()
       continue;
     }
 
-    if (last_range->flag & EQ_RANGE &&
-        used_key_parts <= head->key_info[index].user_defined_key_parts)
+    // Case where we can avoid descending scan, see comment above
+    const bool eqrange_all_keyparts= (last_range->flag & EQ_RANGE) &&
+                          (used_key_parts <= head->key_info[index].user_defined_key_parts);
 
+    if (eqrange_all_keyparts)
     {
+      file->set_end_range(NULL, handler::RANGE_SCAN_ASC);
       result= file->ha_index_read_map(record, last_range->max_key,
                                       last_range->max_keypart_map,
                                       HA_READ_KEY_EXACT);
     }
     else
     {
+      key_range min_range;
+      last_range->make_min_endpoint(&min_range);
+      if (min_range.length > 0)
+        file->set_end_range(&min_range, handler::RANGE_SCAN_DESC);
+
       DBUG_ASSERT(last_range->flag & NEAR_MAX ||
                   (last_range->flag & EQ_RANGE && 
                    used_key_parts > head->key_info[index].user_defined_key_parts) ||
--- a/sql/sql_select.cc
+++ b/sql/sql_select.cc
@@ -25016,6 +25016,20 @@ join_read_last_key(JOIN_TAB *tab)
     report_error(table,error);
     return -1;
   }
+
+  /*
+    Tell the storage engine what the range endpoint is so that it can stop
+    scanning once it has hit that point.
+   */
+  key_range range_endpoint
+  {
+    tab->ref.key_buff,
+    tab->ref.key_length,
+    make_prev_keypart_map(tab->ref.key_parts),
+    HA_READ_PREFIX_LAST
+  };
+  table->file->set_end_range(&range_endpoint, handler::RANGE_SCAN_DESC);
+
   if (unlikely((error=
                 table->file->ha_index_read_map(table->record[0],
                                                tab->ref.key_buff,
@@ -27042,9 +27056,9 @@ void compute_part_of_sort_key_for_equals(JOIN *join, TABLE *table,
   @detail
     - Disable "Range checked for each record" (Is this strictly necessary
       here?)
-    - Disable Index Condition Pushdown and Rowid Filtering.
+    - Disable Rowid Filtering.
 
-  IndexConditionPushdownAndReverseScans, RowidFilteringAndReverseScans:
+  RowidFilteringAndReverseScans:
   Suppose we're computing
 
     select * from t1
@@ -27076,10 +27090,10 @@ void compute_part_of_sort_key_for_equals(JOIN *join, TABLE *table,
     }
 
   Note that the check whether we've walked beyond the key=10 endpoint is
-  made at the SQL layer. The storage engine has no information about the left
+  made at the SQL layer.  The storage engine has no information about the left
   endpoint of the interval we're scanning.  If all rows before that endpoint
-  do not satisfy ICP condition or do not pass the Rowid Filter, the storage
-  engine will enumerate the records until the table start.
+  do not pass the Rowid Filter, the storage engine will enumerate the records
+  until the table start.
 
   In MySQL, the API is extended with set_end_range() call so that the storage
   engine "knows" when to stop scanning.
@@ -27094,16 +27108,7 @@ static void prepare_for_reverse_ordered_access(JOIN_TAB *tab)
     tab->read_first_record= join_init_read_record;
   }
   /*
-    Cancel Pushed Index Condition, as it doesn't work for reverse scans.
-  */
-  if (tab->select && tab->select->pre_idx_push_select_cond)
-  {
-    tab->set_cond(tab->select->pre_idx_push_select_cond);
-     tab->table->file->cancel_pushed_idx_cond();
-  }
-  /*
-    The same with Rowid Filter: it doesn't work with reverse scans so cancel
-    it, too.
+    Rowid filtering does not work with reverse scans so cancel it.
   */
   {
     /*
--- a/mysql-test/main/mdev-34413-icp-reverse-order.combinations
+++ b/mysql-test/main/mdev-34413-icp-reverse-order.combinations
@@ -1,0 +1,6 @@
+[myisam]
+default-storage-engine=myisam
+
+[innodb]
+innodb
+default-storage-engine=innodb
--- a/mysql-test/main/mdev-34413-icp-reverse-order.result
+++ b/mysql-test/main/mdev-34413-icp-reverse-order.result
@@ -1,0 +1,278 @@
+create table ten(a int);
+insert into ten values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
+create table one_k(a int);
+insert into one_k select A.a + B.a* 10 + C.a * 100 from ten A, ten B, ten C;
+create table t10 (a int, b int, c int, key(a,b));
+insert into t10 select a,a,a from one_k;
+select * from t10 force index(a) where a between 10 and 20 and b+1 <3333 order by a desc, b desc;
+a	b	c
+20	20	20
+19	19	19
+18	18	18
+17	17	17
+16	16	16
+15	15	15
+14	14	14
+13	13	13
+12	12	12
+11	11	11
+10	10	10
+explain select * from t10 force index(a) where a between 10 and 20 and b+1 <3333 order by a desc, b desc;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
+1	SIMPLE	t10	range	a	a	5	NULL	11	Using where
+flush status;
+select * from t10 force index(a) where a between 10 and 20 and b+1 <3333 order by a desc, b desc;
+a	b	c
+20	20	20
+19	19	19
+18	18	18
+17	17	17
+16	16	16
+15	15	15
+14	14	14
+13	13	13
+12	12	12
+11	11	11
+10	10	10
+SELECT * FROM information_schema.SESSION_STATUS WHERE VARIABLE_NAME LIKE '%icp%';
+VARIABLE_NAME	VARIABLE_VALUE
+HANDLER_ICP_ATTEMPTS	0
+HANDLER_ICP_MATCH	0
+select * from t10 force index(a) where a between 10 and 20 and b+1 <3333 order by a asc, b asc;
+a	b	c
+10	10	10
+11	11	11
+12	12	12
+13	13	13
+14	14	14
+15	15	15
+16	16	16
+17	17	17
+18	18	18
+19	19	19
+20	20	20
+explain select * from t10 force index(a) where a between 10 and 20 and b+1 <3333 order by a asc, b asc;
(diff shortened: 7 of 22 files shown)


spetrunia (changes requested):
This wording in commit comment is hard to read and needs to be rephrased:

```
    Allows ICP for reverse-ordered scans and optimizes scanning so
    that we do not have equal parts ICP misses and ICP matches during
    reverse scans.  
```    
"we do not have equal parts ICP misses and ICP matches" looks very cryptic.

spetrunia:
Do I get it correctly that this always holds:
```
Handler_icp_attempts = Handler_icp_match  + Handler_icp_miss
```
if yes, do we really need all three counters?

DaveGosselin-MariaDB:
> Do I get it correctly that this always holds:
> 
> ```
> Handler_icp_attempts = Handler_icp_match  + Handler_icp_miss
> ```
> 
> if yes, do we really need all three counters?
> 
> And there's also `ha_handler_stats::icp_miss` which is incremented (per examined row) but never used...

We probably don't need all three counters now at this point.  I thought we might the 'miss' counter earlier in development, when I was concerned about differentiating ICP attempts from other codepaths (like bka_skip_index_tuple) but it doesn't matter now.  I'll remove the 'miss' counters.  
(I originally added `ha_handler_stats::icp_miss` to parallel the SSV counter, the apparent pattern in our code, but deliberately didn't use it; I thought it more important to follow the existing pattern so that a future maintainer could use the value if needed.)

sql/sql_select.cc
```diff
@@ -27089,14 +27089,6 @@ static void prepare_for_reverse_ordered_access(JOIN_TAB *tab)
...
     tab->read_first_record= join_init_read_record;
   }
-  /*
-    Cancel Pushed Index Condition, as it doesn't work for reverse scans.
-  */
-  if (tab->select && tab->select->pre_idx_push_select_cond)
-  {
-    tab->set_cond(tab->select->pre_idx_push_select_cond);
-     tab->table->file->cancel_pushed_idx_cond();
-  }
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: I will clean the comment up.  These changes are insufficient for rowid filters for reverse scans, I made a new ticket last week for that: [MDEV-36094](https://jira.mariadb.org/browse/MDEV-36094).

sql/handler.h
```diff
@@ -4335,7 +4345,12 @@ class handler :public Sql_alloc
                                const key_range *end_key,
                                bool eq_range, bool sorted);
   virtual int read_range_next();
-  void set_end_range(const key_range *end_key);
+  void set_end_range(const key_range *end_key,
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: In this case there's no reason for A over B.  I will follow your preference and change to one implementation with a default parameter value (unless a technical reason surfaces why that cannot be).

sql/ha_partition.cc
```diff
@@ -6367,7 +6367,7 @@ int ha_partition::read_range_first(const key_range *start_key,
 
   m_ordered= sorted;
   eq_range= eq_range_arg;
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Good point, I will double check.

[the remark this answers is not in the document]
DaveGosselin-MariaDB: We invoke the handler's `set_end_range` from `QUICK_RANGE_DESC::get_next` which is engine-independent.  Within a debugger and with appropriate queries, I can see that `handler_index_cond_check` is called by way of `ha_partition::handle_ordered_index_scan` and the expected ICP counters are updated correctly.  I will add such test cases to the new test file.

spetrunia (changes requested):
Hi Dave,

Here's a patch that stops special treatment of NULLs by this MDEV's code (but doesn't revert the change fully).  Please take a look:

```diff
diff -u ../12.0-review-icp-v2/sql/handler.cc sql/handler.cc
--- ../12.0-review-icp-v2/sql/handler.cc	2025-02-20 14:57:33.723910544 +0200
+++ sql/handler.cc	2025-02-20 17:29:18.080622531 +0200
@@ -7213,6 +7213,7 @@
                             enum_range_scan_direction direction)
 {
   end_range= 0;
+  range_scan_direction= direction;
   if (end_key)
   {
     end_range= &save_end_range;
@@ -7220,7 +7221,6 @@
     key_compare_result_on_equal=
       ((end_key->flag == HA_READ_BEFORE_KEY) ? 1 :
        (end_key->flag == HA_READ_AFTER_KEY) ? -1 : 0);
-    range_scan_direction= direction;
     range_key_part= table->key_info[active_index].key_part;
   }
 }
@@ -7264,6 +7264,8 @@
   if (!range)
     return 0;					// no max range
   int null_reason= -1;
+ // tmp: make dbug_print_table_row work:
+ // MY_BITMAP *old_map= dbug_tmp_use_all_columns(table, &table->write_set);
   cmp= key_cmp(range_key_part, range->key, range->length, &null_reason);
   if (!cmp)
     cmp= key_compare_result_on_equal;
@@ -7274,8 +7276,11 @@
     flipped sort order back to the original as that will have the net
     effect of incorrectly scanning in the ascending direction.
   */
-  if (range_scan_direction == RANGE_SCAN_DESC && null_reason != 1)
+  //psergey: roll back:
+ // if (range_scan_direction == RANGE_SCAN_DESC && null_reason != 1) 
+   if (range_scan_direction == RANGE_SCAN_DESC) 
     cmp= -cmp;
+  //dbug_tmp_restore_column_map(&table->write_set, old_map);
   return cmp;
 }
 
diff -u ../12.0-review-icp-v2/sql/opt_range.cc sql/opt_range.cc
--- ../12.0-review-icp-v2/sql/opt_range.cc	2025-02-20 14:57:33.727910665 +0200
+++ sql/opt_range.cc	2025-02-20 14:55:33.556284082 +0200
@@ -13775,6 +13775,7 @@
     if (eqrange_all_keyparts)
 
     {
+      file->set_end_range(NULL, handler::RANGE_SCAN_ASC);
       result= file->ha_index_read_map(record, last_range->max_key,
                                       last_range->max_keypart_map,
                                       HA_READ_KEY_EXACT);
```

spetrunia:
Bonus question: 

the failing query in subselect3 was:
```
create table t12 select * from t1 where b = 67 AND (c IS NULL OR c > NOW()) order by 3 desc;
```
It uses ICP.
But its corresponding select
```
select * from t1 where b = 67 AND (c IS NULL OR c > NOW()) order by 3 desc
```
uses index-only.  Any idea why?

DaveGosselin-MariaDB:
> Bonus question:
> 
> the failing query in subselect3 was:
> 
> ```
> create table t12 select * from t1 where b = 67 AND (c IS NULL OR c > NOW()) order by 3 desc;
> ```
> 
> It uses ICP. But its corresponding select
> 
> ```
> select * from t1 where b = 67 AND (c IS NULL OR c > NOW()) order by 3 desc
> ```
> 
> uses index-only. Any idea why?

Yes, if you are using MyISAM tables (which is the default when running under MTR) then the covering keys computed for the table will be different than if another engine (like InnoDB) was used.  This is because the share->keys_for_keyread (which later become the table's covering keys) are in part computed by the engine-specifid implementation of handler::index_flags.  In the case of InnoDB, ha_innobase::index_flags checks if the passed-in key field matches the table_share->primary_key and if it does then it returns HA_CLUSTERED_INDEX; otherwise, it returns HA_KEYREAD_ONLY.  But in the case of MyISAM, there is no clustered index, so that flag cannot be returned.  In fact it is this difference that results in different covering keys for the table.  Consequently, the index condition cannot be pushed during make_join_readinfo for InnoDB but it will be pushed when the engine is MyISAM.

spetrunia (changes requested):
Hi Dave,

The previous rationale about partitioned tables didn't look very convincing: I didn't get a reply for my question about set_end_range() call for "child" partition handlers.

So I've set to create a test case myself:

```sql
create table t1 ( 
  pk int primary key, 
  kp1 int, kp2 int, 
  col1 int, 
  index (kp1,kp2)
) partition by hash (pk) partitions 10;
insert into t1 select seq, seq, seq, seq from seq_1_to_1000;
```
First, ascending scan 
```flush status;
analyze
select * from t1 where kp1 between 950 and 960 and kp2+1 >33333 order by kp1 asc, kp2 asc;
show status like 'Handler_icp%';
```
gives
```
+------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
| id   | select_type | table | type  | possible_keys | key  | key_len | ref  | rows | r_rows | filtered | r_filtered | Extra                 |
+------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
|    1 | SIMPLE      | t1    | range | kp1           | kp1  | 5       | NULL | 11   | 0.00   |   100.00 |     100.00 | Using index condition |
+------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
```
```
+----------------------+-------+
| Variable_name        | Value |
+----------------------+-------+
| Handler_icp_attempts | 11    |
| Handler_icp_match    | 0     |
+----------------------+-------+
```
Ok.
Now, descending scan:

```
flush status;
analyze
select * from t1 where kp1 between 950 and 960 and kp2+1 >33333 order by kp1 desc, kp2 desc;
show status like 'Handler_icp%';
```
gives
```
+------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
| id   | select_type | table | type  | possible_keys | key  | key_len | ref  | rows | r_rows | filtered | r_filtered | Extra                 |
+------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
|    1 | SIMPLE      | t1    | range | kp1           | kp1  | 5       | NULL | 11   | 0.00   |   100.00 |     100.00 | Using index condition |
+------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
```
and
```
+----------------------+-------+
| Variable_name        | Value |
+----------------------+-------+
| Handler_icp_attempts | 960   |
| Handler_icp_match    | 0     |
+----------------------+-------+
```
Oops. This is how "run-off-the-end-of-range" problem looks like.

Please fix.

sql/handler.h
```diff
@@ -3295,6 +3295,16 @@ class handler :public Sql_alloc
...
+  /*
+    The direction of the current range or index scan. This is used by
+    the ICP implementation to determine if it has reached the end
+    of the current range.
+  */
+  enum enum_range_scan_direction {
+    RANGE_SCAN_ASC,
+    RANGE_SCAN_DESC
+  };
+  enum_range_scan_direction range_scan_direction{RANGE_SCAN_ASC};
```
spetrunia, line 3307: Please make range_scan_direction "protected:".

DaveGosselin-MariaDB:
> Hi Dave,
> 
> The previous rationale about partitioned tables didn't look very convincing: I didn't get a reply for my question about set_end_range() call for "child" partition handlers.
> 

@spetrunia I answered your question here earlier this week:  https://github.com/MariaDB/server/pull/3837#discussion_r1960061625

Maybe you are not receiving GitHub notifications?

DaveGosselin-MariaDB:
> Ok. Now, descending scan:
> 
> ```
> flush status;
> analyze
> select * from t1 where kp1 between 950 and 960 and kp2+1 >33333 order by kp1 desc, kp2 desc;
> show status like 'Handler_icp%';
> ```
> 
> gives
> 
> ```
> +------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
> | id   | select_type | table | type  | possible_keys | key  | key_len | ref  | rows | r_rows | filtered | r_filtered | Extra                 |
> +------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
> |    1 | SIMPLE      | t1    | range | kp1           | kp1  | 5       | NULL | 11   | 0.00   |   100.00 |     100.00 | Using index condition |
> +------+-------------+-------+-------+---------------+------+---------+------+------+--------+----------+------------+-----------------------+
> ```
> 
> and
> 
> ```
> +----------------------+-------+
> | Variable_name        | Value |
> +----------------------+-------+
> | Handler_icp_attempts | 960   |
> | Handler_icp_match    | 0     |
> +----------------------+-------+
> ```
> 
> Oops. This is how "run-off-the-end-of-range" problem looks like.
> 
> Please fix.

Thank you, this is helpful.  I will fix it.

sql/handler.cc
```diff
@@ -7132,8 +7132,9 @@ int handler::read_range_first(const key_range *start_key,
   DBUG_ENTER("handler::read_range_first");
 
   eq_range= eq_range_arg;
-  set_end_range(end_key);
+  set_end_range(end_key, RANGE_SCAN_ASC);
   range_key_part= table->key_info[active_index].key_part;
+  range_scan_direction= RANGE_SCAN_ASC;
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Good catch, removed.

spetrunia:
Please apply this patch fixing coding style and adding more comments:  https://gist.github.com/spetrunia/343bd89317a5938ba828b87304e67bcd 
(feel free to edit the comments if necessary).

spetrunia (approved):
After applying the above patch,  the pull request is approved.

spetrunia:
One more comment: why is there `--source include/have_innodb.inc` in mysql-test/main/mdev-34413-icp-reverse-order.test ? 
If we want to run reverse ICP tests with many engines, maybe they should be put into `mysql-test/include/icp_tests.inc` , which is ran with innodb/myisam/aria ?

DaveGosselin-MariaDB:
> One more comment: why is there `--source include/have_innodb.inc` in mysql-test/main/mdev-34413-icp-reverse-order.test ? If we want to run reverse ICP tests with many engines, maybe they should be put into `mysql-test/include/icp_tests.inc` , which is ran with innodb/myisam/aria ?

I created a `.combinations` file as part of this commit to ensure that at least InnoDB and MyISAM are tested.  I didn't include the other engines because the key length field in the `explain` output is different by one.  I don't see a way to have a per-engine result file.  Otherwise the fix worked for those engines as well, which isn't surprising because this change is engine-agnostic.

spetrunia:
With other changes (will post shortly):  Please apply this: [remove-endspaces.PATCH](https://github.com/user-attachments/files/19104983/remove-endspaces.PATCH)

sql/opt_range.cc
```diff
@@ -13744,6 +13744,23 @@ int QUICK_SELECT_DESC::get_next()
...
+    // Case where we can avoid descending scan, see comment above
+    const bool eqrange_all_keyparts= (last_range->flag & EQ_RANGE) &&
+                          (used_key_parts <= head->key_info[index].user_defined_key_parts);
+    /*
+      If we have pushed an index condition (ICP) and this quick select
+      will use ha_index_prev() to read data, we need to let the
+      handler know where to end the scan in order to avoid that the
+      ICP implemention continues to read past the range boundary.
+    */
+    if (file->pushed_idx_cond && !eqrange_all_keyparts)
```
spetrunia, line 13756: What if there is no pushed index condition, but a rowid filter is present? 
We need set_end_range call to be made, right? 

Can you fix this in https://github.com/MariaDB/server/pull/3846.  Idea about how to add a testcase:
you can get a range access without ICP if the index used has partially covered columns:
```
create table  t1 (
  key1 varchar(100),
 index (key1(10))
 );

select * from t1 where key1 like 'abcd%';
```
then you add key2 and range condition on key2 so that rowid filter is used.

spetrunia (changes requested):
Please move a portion from https://github.com/MariaDB/server/pull/3846 to here and address the below in https://github.com/MariaDB/server/pull/3846

DaveGosselin-MariaDB:
> With other changes (will post shortly): Please apply this: [remove-endspaces.PATCH](https://github.com/user-attachments/files/19104983/remove-endspaces.PATCH)

I’m happy to apply the patch to remove end spaces, but Sanja has advised me against such changes as they may make merges more difficult.  I tend to fixup such things on lines extremely close to the ones I’m changing already so as not to make merges harder, but not go beyond those.  Even though I would like to!

sql/opt_range.cc
DaveGosselin-MariaDB: I see, yes I can generate such a plan (rowid filter without index condition pushdown) following that test idea.  I will investigate.

DaveGosselin-MariaDB: Is there another way to get a range access without ICP if the index used has partially covered columns?  Otherwise, I don't think that there is a problem here to fix, for such a query the end range is set in a different way as shown here below.
See this plan:
```
MariaDB [test]> explain format=json select * from t1 where key1 like 'abcd%' and key2 between 2 and 12 order by key1 desc, key2 desc\G
EXPLAIN: {
  "query_block": {
    "select_id": 1,
    "cost": 0.014672242,
    "nested_loop": [
      {
        "read_sorted_file": {
          "filesort": {
            "sort_key": "t1.key1 desc, t1.key2 desc",
            "table": {
              "table_name": "t1",
              "access_type": "range",
              "possible_keys": ["key1", "key2_idx"],
              "key": "key1",
              "key_length": "43",
              "used_key_parts": ["key1"],
              "rowid_filter": {
                "range": {
                  "key": "key2_idx",
                  "used_key_parts": ["key2"]
                },
                "rows": 11,
                "selectivity_pct": 73.33333333
              },
              "loops": 1,
              "rows": 1,
              "cost": 0.014212088,
              "filtered": 73.33333588,
              "attached_condition": "t1.key1 like 'abcd%' and t1.key2 between 2 and 12"
            }
          }
        }
      }
    ]
  }
}
1 row in set (0.002 sec)
```
It uses rowid filtering and no index condition pushdown.  When tracing `handler_rowid_filter_check`, the `end_range` is checked (because `if (!pushed_idx_cond)` path is taken) and we return out-of-range appropriately.  This `end_range` was set during `handler::read_range_first` which is called by `handler::multi_range_read_next` in this case:
```
  * frame #0: 0x0000000100109858 mariadbd`handler::set_end_range(this=0x00000001573302a8, end_key=0x0000000157330428, direction=RANGE_SCAN_ASC) at handler.cc:7220:12
    frame #1: 0x000000010010947c mariadbd`handler::read_range_first(this=0x00000001573302a8, start_key=0x0000000157330408, end_key=0x0000000157330428, eq_range_arg=false, sorted=true) at handler.cc:7135:3
    frame #2: 0x00000001007bfc08 mariadbd`handler::multi_range_read_next(this=0x00000001573302a8, range_info=0x0000000170386d58) at multi_range_read.cc:590:15
    frame #3: 0x00000001007bfda8 mariadbd`Mrr_simple_index_reader::get_next(this=0x0000000157330988, range_info=0x0000000170386d58) at multi_range_read.cc:627:32
    frame #4: 0x00000001007c3130 mariadbd`DsMrr_impl::dsmrr_next(this=0x0000000157330838, range_info=0x0000000170386d58) at multi_range_read.cc:1731:26
    frame #5: 0x0000000100cb1078 mariadbd`ha_innobase::multi_range_read_next(this=0x00000001573302a8, range_info=0x0000000170386d58) at ha_innodb.cc:20309:18
    frame #6: 0x0000000100371c90 mariadbd`QUICK_RANGE_SELECT::get_next(this=0x0000000156604d20) at opt_range.cc:13474:5
    frame #7: 0x00000001008bc320 mariadbd`Range_rowid_filter::build(this=0x00000001497e5c70) at rowid_filter.cc:607:41
    frame #8: 0x00000001005ce600 mariadbd`st_join_table::build_range_rowid_filter(this=0x00000001497e4c40) at sql_select.cc:16444:59
    frame #9: 0x0000000100592d0c mariadbd`sub_select(join=0x00000001497e2418, join_tab=0x00000001497e4c40, end_of_records=false) at sql_select.cc:24144:21
...
```
The same is true regardless of whether we use `ORDER BY ... DESC` or not:
```
MariaDB [test]> explain format=json select * from t1 where key1 like 'abcd%' and key2 between 2 and 12\G
EXPLAIN: {
  "query_block": {
    "select_id": 1,
    "cost": 0.014212088,
    "nested_loop": [
      {
        "table": {
          "table_name": "t1",
          "access_type": "range",
          "possible_keys": ["key1", "key2_idx"],
          "key": "key1",
          "key_length": "43",
          "used_key_parts": ["key1"],
          "rowid_filter": {
            "range": {
              "key": "key2_idx",
              "used_key_parts": ["key2"]
            },
            "rows": 11,
            "selectivity_pct": 73.33333333
          },
          "loops": 1,
          "rows": 1,
          "cost": 0.014212088,
          "filtered": 73.33333588,
          "attached_condition": "t1.key1 like 'abcd%' and t1.key2 between 2 and 12"
        }
      }
    ]
  }
}
1 row in set (0.003 sec)
```
and the corresponding backtrace for when `end_range` is set:
```
  * frame #0: 0x0000000100109858 mariadbd`handler::set_end_range(this=0x00000001573302a8, end_key=0x0000000157330428, direction=RANGE_SCAN_ASC) at handler.cc:7220:12
    frame #1: 0x000000010010947c mariadbd`handler::read_range_first(this=0x00000001573302a8, start_key=0x0000000157330408, end_key=0x0000000157330428, eq_range_arg=false, sorted=true) at handler.cc:7135:3
    frame #2: 0x00000001007bfc08 mariadbd`handler::multi_range_read_next(this=0x00000001573302a8, range_info=0x0000000170386d58) at multi_range_read.cc:590:15
    frame #3: 0x00000001007bfda8 mariadbd`Mrr_simple_index_reader::get_next(this=0x0000000157330988, range_info=0x0000000170386d58) at multi_range_read.cc:627:32
    frame #4: 0x00000001007c3130 mariadbd`DsMrr_impl::dsmrr_next(this=0x0000000157330838, range_info=0x0000000170386d58) at multi_range_read.cc:1731:26
    frame #5: 0x0000000100cb1078 mariadbd`ha_innobase::multi_range_read_next(this=0x00000001573302a8, range_info=0x0000000170386d58) at ha_innodb.cc:20309:18
    frame #6: 0x0000000100371c90 mariadbd`QUICK_RANGE_SELECT::get_next(this=0x0000000156606990) at opt_range.cc:13474:5
    frame #7: 0x00000001008bc320 mariadbd`Range_rowid_filter::build(this=0x00000001497e57f0) at rowid_filter.cc:607:41
    frame #8: 0x00000001005ce600 mariadbd`st_join_table::build_range_rowid_filter(this=0x00000001497e47c0) at sql_select.cc:16444:59
    frame #9: 0x0000000100592d0c mariadbd`sub_select(join=0x00000001497e2088, join_tab=0x00000001497e47c0, end_of_records=false) at sql_select.cc:24144:21
```
...

spetrunia: Okay this uses rowid filter without ICP but what it does not use
is reverse-ordered range scan. It uses filesort if ORDER BY DESC is present.

And I see what the problem is in my suggestion. If we have

  index IDX(key1(10))

then that index doesn't match the ORDER BY key1 DESC.
I've tried stepping this around with multi-part indexes, but I wasn't able to get all three:
- reverse ordered scan
- range access (not REF)
- no ICP.

(one also needs rowid filter but adding that should be easy)

Please use different ways to construct testcase not using ICP.

1. Use set optimizer_switch='index_condition_pushdown=off' :-)
2. Conditions using cheap constant subqueries construct range access but 
are not used for ICP: the following uses range+using where:

```sql
create table t20 (a int, b  int, key(a));
insert into t20 select seq, seq from seq_1_to_1000;
create table t21 like t20; 
insert into t21 select seq, seq from seq_1_to_10;
explain select * from t20 where a < (select max(t21.a) from t21);
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------+
| id   | select_type | table | type  | possible_keys | key  | key_len | ref  | rows | Extra                        |
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------+
|    1 | PRIMARY     | t20   | range | a             | a    | 5       | NULL | 9    | Using where                  |
|    2 | SUBQUERY    | NULL  | NULL  | NULL          | NULL | NULL    | NULL | NULL | Select tables optimized away |
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------+
2 rows in set (0.003 sec)
```

DaveGosselin-MariaDB: A fix for this should be in this PR.  However, rowid filtering for reverse scanning is disabled at this point.  So the resulting test will go into the rowid filtering PR #3846 .

DaveGosselin-MariaDB: After testing with a change to set the end range regardless of ICP in QUICK_RANGE_DESC, we see the following (namely that r_lookups, r_index_rows, pages_accessed are all lower and r_total_filtered is higher):
```
analyze format=json select * from t1 where t1.a between 3 and 8 and t1.b between 4000 and 4100 order by t1.b desc\G
ANALYZE: {
  "query_optimization": {
    "r_total_time_ms": 0.338236794
  },
  "query_block": {
    "select_id": 1,
    "cost": 0.012382207,
    "r_loops": 1,
    "r_total_time_ms": 0.219337143,
    "nested_loop": [
      {
        "table": {
          "table_name": "t1",
          "access_type": "range",
          "possible_keys": ["t1_a", "t1_b"],
          "key": "t1_b",
          "key_length": "5",
          "used_key_parts": ["b"],
          "rowid_filter": {
            "range": {
              "key": "t1_a",
              "used_key_parts": ["a"]
            },
            "rows": 6,
            "selectivity_pct": 60,
            "r_rows": 6,
            "r_lookups": 1,
            "r_selectivity_pct": 100,
            "r_buffer_size": 4,
            "r_filling_time_ms": 0.123025507
          },
          "loops": 1,
          "r_loops": 1,
          "rows": 1,
          "r_index_rows": 1,
          "r_rows": 1,
          "cost": 0.012005717,
          "r_table_time_ms": 0.176119849,
          "r_other_time_ms": 0.027589053,
          "r_engine_stats": {
            "pages_accessed": 3
          },
          "filtered": 60,
          "r_total_filtered": 100,
          "attached_condition": "t1.a between 3 and 8 and t1.b between 4000 and 4100",
          "r_filtered": 100
        }
      }
    ]
  }
}
```

spetrunia (approved):
Now, looks good.

PR #3838, opened 2025-02-13
MDEV-32362 Handle generated columns in mariadb-dump INSERT statements 

FarihaIS:
## Description

Currently, generated column names and values are present in `INSERT` statements created by the dump tool. While this doesn't break restorability, it is unnecessary as values for generated columns are calculated based on other column values and need not be explicitly inserted.
    
Modify `mariadb-dump` to handle generated columns in `INSERT` statements:
- In default mode, generated column values are replaced with `DEFAULT`, preserving the user's chosen insert style.
- When `--complete-insert` is enabled (either explicitly or due to `INVISIBLE` columns), generated columns are omitted entirely from the column list and values.
- Exception: if every column in a table is generated, generated columns are kept in the `INSERT` and `DEFAULT` is emitted for their values, so row counts are preserved on reload ([MDEV-40217](https://jira.mariadb.org/browse/MDEV-40217)).
    
The server computes the generated column values automatically in all cases.

## Release Notes

N/A

## How can this PR be tested?

Execute `main.mysqldump` in mysql-test-run. This commit adds a test in the `main` suite.

* [x] _This is a new feature and the PR is based against the latest MariaDB development branch._

## Copyright

### DIFF
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -3205,7 +3205,7 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
                                 char *ignore_flag, my_bool *versioned)
 {
   my_bool    init=0, delayed, write_data, complete_insert;
-  my_bool    is_generated=0, is_vector=0;
+  my_bool    is_generated=0;
   my_ulonglong num_fields;
   char       *result_table, *opt_quoted_table;
   const char *insert_option;
@@ -3688,40 +3688,32 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
       check_io(sql_file);
     }
 
+    if (write_data)
+    {
+      if (opt_replace_into)
+        dynstr_append_checked(&insert_pat, "REPLACE ");
+      else
+        dynstr_append_checked(&insert_pat, "INSERT ");
+      dynstr_append_checked(&insert_pat, insert_option);
+      dynstr_append_checked(&insert_pat, "INTO ");
+      dynstr_append_checked(&insert_pat, result_table);
+      if (complete_insert)
+        dynstr_append_checked(&insert_pat, " (");
+      else
+      {
+        dynstr_append_checked(&insert_pat, " VALUES ");
+        if (!extended_insert)
+          dynstr_append_checked(&insert_pat, "(");
+      }
+    }
+
     while ((row= mysql_fetch_row(result)))
     {
       ulong *lengths= mysql_fetch_lengths(result);
-      bool is_row_start= strstr(row[SHOW_EXTRA], "ROW START") != 0;
-      bool is_row_end= strstr(row[SHOW_EXTRA], "ROW END") != 0;
-      is_generated= 0;
-      is_vector= 0;
-
+      dynstr_append_mem_checked(&field_flags, "", 1);
       /* VECTOR doesn't have a type code yet, must be detected by name */
       if (strncmp(row[SHOW_TYPE], STRING_WITH_LEN("vector(")) == 0)
-        is_vector= 1;
-      else if (strstr(row[SHOW_EXTRA], "GENERATED") &&
-               !is_row_start && !is_row_end)
-        is_generated= 1;
-      if (strstr(row[SHOW_EXTRA], "INVISIBLE"))
-        complete_insert= 1;
-
-      /*
-        When complete_insert is enabled, skip generated columns entirely.
-        Otherwise, include them and mark them so DEFAULT is emitted.
-      */
-      if (is_generated && !path && !opt_dir && !opt_dump_history &&
-          complete_insert)
-        continue;
-
-      dynstr_append_mem_checked(&field_flags, "", 1);
-      if (is_generated)
-        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_GENERATED;
-      if (is_vector)
         field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VECTOR;
-      /* Mark system versioning columns */
-      if (is_row_start || is_row_end)
-        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VERS_COL;
-
       if (init)
       {
         if (!opt_xml && !opt_no_create_info)
@@ -3766,29 +3758,9 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
         check_io(sql_file);
       }
     }
-
-    if (write_data)
-    {
-      if (opt_replace_into)
-        dynstr_append_checked(&insert_pat, "REPLACE ");
-      else
-        dynstr_append_checked(&insert_pat, "INSERT ");
-      dynstr_append_checked(&insert_pat, insert_option);
-      dynstr_append_checked(&insert_pat, "INTO ");
-      dynstr_append_checked(&insert_pat, result_table);
-      if (complete_insert)
-        dynstr_append_checked(&insert_pat, " (");
-      else
-      {
-        dynstr_append_checked(&insert_pat, " VALUES ");
-        if (!extended_insert)
-          dynstr_append_checked(&insert_pat, "(");
-      }
-    }
-
     if (complete_insert)
       dynstr_append_checked(&insert_pat, select_field_names.str);
-    num_fields= field_flags.length;
+    num_fields= mysql_num_rows(result);
     mysql_free_result(result);
     if (!opt_no_create_info)
     {
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -104,7 +104,9 @@
 #define DUMP_TABLE_SEQUENCE 1
 
 /* until MDEV-35831 is implemented, we'll have to detect VECTOR by name */
-#define MYSQL_TYPE_VECTOR "V"
+#define MYSQL_TYPE_VECTOR 1
+#define MYSQL_TYPE_VERS_COL 2
+#define MYSQL_TYPE_GENERATED 4
 
 static my_bool ignore_table_data(const uchar *hash_key, size_t len);
 static void add_load_option(DYNAMIC_STRING *str, const char *option,
@@ -3203,6 +3205,7 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
                                 char *ignore_flag, my_bool *versioned)
 {
   my_bool    init=0, delayed, write_data, complete_insert;
+  my_bool    is_generated=0, is_vector=0;
   my_ulonglong num_fields;
   char       *result_table, *opt_quoted_table;
   const char *insert_option;
@@ -3521,9 +3524,16 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
 
     while ((row= mysql_fetch_row(result)))
     {
-      if (strstr(row[1],"INVISIBLE"))
+      bool is_row_start= row[2] && strcmp(row[2], "ROW START") == 0;
+      bool is_row_end= row[2] && strcmp(row[2], "ROW END") == 0;
+
+      is_generated= 0;
+      if (strstr(row[1], "GENERATED") && !is_row_start && !is_row_end)
+        is_generated= 1;
+      if (strstr(row[1], "INVISIBLE"))
         complete_insert= 1;
-      if (vers_hidden && row[2] && strcmp(row[2], "ROW START") == 0)
+
+      if (vers_hidden && is_row_start)
       {
         vers_hidden= 0;
         if (row[3] && strcmp(row[3], "bigint") == 0)
@@ -3533,6 +3543,24 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
           *versioned= 0;
         }
       }
+
+      /*
+        When complete_insert is enabled, skip generated columns entirely.
+        Otherwise, include them and mark them so DEFAULT is emitted.
+      */
+      if (is_generated && !path && !opt_dir && !opt_dump_history &&
+          complete_insert)
+        continue;
+
+      dynstr_append_mem_checked(&field_flags, "", 1);
+      if (is_generated)
+        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_GENERATED;
+      if (row[3] && strcmp(row[3], "vector") == 0)
+        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VECTOR;
+      /* Mark system versioning columns */
+      if (is_row_start || is_row_end)
+        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VERS_COL;
+
       if (init)
       {
         dynstr_append_checked(&select_field_names, ", ");
@@ -3543,8 +3571,7 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
       init=1;
 
       last_name= quote_name(row[0], name_buff, 0);
-      if (opt_dump_history && *versioned && opt_update_history &&
-          row[2] && strcmp(row[2], "ROW END") == 0)
+      if (opt_dump_history && *versioned && opt_update_history && is_row_end)
       {
         dynstr_append_checked(&select_field_names, "if(");
         dynstr_append_checked(&select_field_names, last_name);
@@ -3561,11 +3588,6 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
       if (opt_header)
         dynstr_append_checked(&select_field_names_for_header,
                               quote_for_equal(row[0], name_buff));
-      /* VECTOR doesn't have a type code yet, must be detected by name */
-      if (row[3] && strcmp(row[3], "vector") == 0)
-        dynstr_append_checked(&field_flags, MYSQL_TYPE_VECTOR);
-      else
-        dynstr_append_checked(&field_flags, " ");
     }
 
     if (vers_hidden)
@@ -3579,7 +3601,11 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
                             "row_end" :
                             "row_end");
       dynstr_append_checked(&insert_field_names, ", row_start, row_end");
-      dynstr_append_checked(&field_flags, "  ");
+      /* Mark both row_start and row_end as versioning columns */
+      dynstr_append_mem_checked(&field_flags, "", 1);
+      field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VERS_COL;
+      dynstr_append_mem_checked(&field_flags, "", 1);
+      field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VERS_COL;
     }
 
     /*
@@ -3598,9 +3624,7 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
       dynstr_append_checked(&insert_pat, "INTO ");
       dynstr_append_checked(&insert_pat, opt_quoted_table);
       if (complete_insert)
-      {
         dynstr_append_checked(&insert_pat, " (");
-      }
       else
       {
         if (extended_insert)
@@ -3612,7 +3636,7 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
 
     if (complete_insert)
       dynstr_append_checked(&insert_pat, insert_field_names.str);
-    num_fields= mysql_num_rows(result) + (vers_hidden ? 2 : 0);
+    num_fields= field_flags.length;
     mysql_free_result(result);
   }
   else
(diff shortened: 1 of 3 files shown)


vuvova:
I've just — few days ago — added `field_flags` array to `mariadb-dump`. In my commit it only stores whether the first is VECTOR or not, but the idea was that it can be used for other flags too.

I suggest you reuse that. That is, instead of `field_flags->str[i]` being `'V'` for vectors and `' '` for anything else, it should be a set of flags, like
```c
#define FIELD_IS_VECTOR  1
#define FIELD_IS_VIRTUAL 2
...
  if (field_flags->str[i] & FIELD_IS_VECTOR)
  {
    ...
```
and so on

FarihaIS:
> I've just — few days ago — added `field_flags` array to `mariadb-dump`. In my commit it only stores whether the first is VECTOR or not, but the idea was that it can be used for other flags too.
> 
> I suggest you reuse that. That is, instead of `field_flags->str[i]` being `'V'` for vectors and `' '` for anything else, it should be a set of flags, like
> 
> ```c
> #define FIELD_IS_VECTOR  1
> #define FIELD_IS_VIRTUAL 2
> ...
>   if (field_flags->str[i] & FIELD_IS_VECTOR)
>   {
>     ...
> ```
> 
> and so on



> I've just — few days ago — added `field_flags` array to `mariadb-dump`. In my commit it only stores whether the first is VECTOR or not, but the idea was that it can be used for other flags too.
> 
> I suggest you reuse that. That is, instead of `field_flags->str[i]` being `'V'` for vectors and `' '` for anything else, it should be a set of flags, like
> 
> ```c
> #define FIELD_IS_VECTOR  1
> #define FIELD_IS_VIRTUAL 2
> ...
>   if (field_flags->str[i] & FIELD_IS_VECTOR)
>   {
>     ...
> ```
> 
> and so on

Thank you for the feedback, I've refactored my code to reuse `field_flag`s to detect generated columns. Please let me know if this is the change you had in mind.

vuvova:
Not quite, I meant this:
```diff
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -104,7 +104,7 @@
 #define DUMP_TABLE_SEQUENCE 1
 
 /* until MDEV-35831 is implemented, we'll have to detect VECTOR by name */
-#define MYSQL_TYPE_VECTOR "V"
+#define MYSQL_TYPE_VECTOR 1
 
 static my_bool ignore_table_data(const uchar *hash_key, size_t len);
 static void add_load_option(DYNAMIC_STRING *str, const char *option,
@@ -3498,11 +3498,10 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
       if (opt_header)
         dynstr_append_checked(&select_field_names_for_header,
                               quote_for_equal(row[0], name_buff));
+      dynstr_append_mem_checked(&field_flags, "", 1);
       /* VECTOR doesn't have a type code yet, must be detected by name */
       if (row[3] && strcmp(row[3], "vector") == 0)
-        dynstr_append_checked(&field_flags, MYSQL_TYPE_VECTOR);
-      else
-        dynstr_append_checked(&field_flags, " ");
+        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VECTOR;
     }
 
     if (vers_hidden)
@@ -3623,11 +3622,10 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
     while ((row= mysql_fetch_row(result)))
     {
       ulong *lengths= mysql_fetch_lengths(result);
+      dynstr_append_mem_checked(&field_flags, "", 1);
       /* VECTOR doesn't have a type code yet, must be detected by name */
       if (strncmp(row[SHOW_TYPE], STRING_WITH_LEN("vector(")) == 0)
-        dynstr_append_checked(&field_flags, MYSQL_TYPE_VECTOR);
-      else
-        dynstr_append_checked(&field_flags, " ");
+        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VECTOR;
       if (init)
       {
         if (!opt_xml && !opt_no_create_info)
@@ -4519,7 +4517,7 @@ static void dump_table(const char *table, const char *db, const uchar *hash_key,
         */
         is_blob= field->type == MYSQL_TYPE_GEOMETRY ||
                  field->type == MYSQL_TYPE_BIT ||
-                 field_flags.str[i] == MYSQL_TYPE_VECTOR[0] ||
+                 field_flags.str[i] & MYSQL_TYPE_VECTOR ||
                  (opt_hex_blob && field->charsetnr == 63 &&
                    (field->type == MYSQL_TYPE_STRING ||
                     field->type == MYSQL_TYPE_VAR_STRING ||
```
This is the complete patch that passes tests. `field_flags` is a bitmask. You'll use it as
```c
  if (field_flags.str[i] & FIELD_IS_GENERATED)
    ...
```

ottok:
The current main branch has these failures:
image

This PR currently has these failures:
image

Of these 5 failures 4 have the label "Required". This does not make sense as the test jobs that have "Required" to pass are nonetheless failing on branch 'main', so they are actually not required to pass. We are now diving into these test failures and trying to debug them without knowing if they actually have anything to do with the changes in this PR.

dr-m:
Some test failures are nondeterministic, possibly depending on the load of the CI worker machines, and a test would typically pass if you hit "rebuild" on a failed build on the Buildbot grid view. Sometimes, depending on your luck, this might require multiple attempts. Generally, the mandatory builds must pass (at least once) before something can be pushed to a protected branch. In my experience, most of the sporadically failing tests are currently in `./mtr --suite=rpl,binlog,binlog_encryption`. @knielsen and @bnestere have started an effort to fix replication test failures quite some time ago, so I feel that the situation is improving slowly but steadily.

In https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F3838%2Fmerge I can see failures of the tests `versioning.data` and `versioning.old_timestamp` across the board. Those failures do not seem to be present in https://buildbot.mariadb.org/#/grid?branch=main but some sporadic failures are there.

ottok:
@dr-m Thanks for the reply. I am not saying it is _impossible_ for people to contribute while tests are failing. Sure they can do the detective work to arrive on the conclusion that all failures are sporadic and/or on branches that are not 'required'. **The problem is that this is causing a lot of confusion for new contributors.** People with experience of CI systems elsewhere assume and expect that a CI system is green and is used to prevent regressions from entering the code base. The fact that is is constantly red leads to many contributors wasting time in investigating unrelated things, which may be off-putting and discourage them from doing more contributions.

vuvova:
The exciting digression about buildbot notwithstanding, @FarihaIS, please, say when you'll want a review

vuvova:
@FarihaIS, is this PR ready for a review or are you still working on it?

FarihaIS:
@vuvova I'm still working on it! The fix I added earlier broke tests in the versioning suite, which I'm trying to rectify

vuvova:
And now?

dr-m:
Pull requests that are not ready for review can be marked as a "draft". The [buildbot grid view](https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F3838%2Fhead) looks rather OK for this, except for two tests that are failing on multiple platforms:
```
versioning.data
versioning.old_timestamp
```

svoj:
Hello @FarihaIS, it's been a while since this PR was updated. Just wanted to check if it is ready for the review or do you intend to complete it?

FarihaIS:
Hi @svoj, yes I'm still working on the pull request, but progress has been a bit slower with the tests that broke in the versioning suite. I'll prioritize this and take a look again soon, thank you!

svoj:
@FarihaIS that's fine, I'm just checking stale pull requests if they're still relevant.

FarihaIS:
Hi, I made some more changes and the tests `versioning.data` and `versioning.old_timestamp` are no longer failing - could I please get a review for this PR again? Thank you!

vuvova (changes requested):
1. The test  doesn't show mysqldump output with generated columns. You only store it in a file and load back. But that worked before your change too. Please add a test that shows new output.

2.  I think that forcing complete_insert for generated columns is a bit too verbose. Better would be to use `DEFAULT` instead. Like
```sql
INSERT INTO `example` VALUES
('Donald','Duck',DEFAULT);
```
But, of course, if complete_insert is already enabled (because the user specified it or for some other reason) then completely omitting generated columns is better.

FarihaIS:
@vuvova thank you for your feedback. 

I have updated the PR title and description as well as the commit title and message to reflect the new expected behavior of mariadb-dump with generated columns in both default mode and when using `--complete-insert`. 

Additionally, I have updated the new test previously added to include the `INSERT` lines in the output of the dump for verification of the fix.

I do see that `main.xa` is failing on some platforms, but it seems to be a flaky test and unrelated to my changes.

Could you please kindly review my PR again and let me know if there are any further changes needed? Thank you

gkodinov (changes requested):
This is a preliminary review. Thank you for your contribution!

One question: do you really need the new option?

As, as it stands right now, thanks to the new option, you can only fix this in main. And the problem exists even for 11.4. 

But, even putting that aside, what good is to have a "DEFAULT" set as a value?

This just increases the size of the dump for not reason I can think of. 

How about you just don't print the generated columns at all?

FarihaIS:
@gkodinov Thank you for the feedback! I have addressed your comment regarding not printing the generated columns at all. Could you please kindly take another look and let me know if you have any other thoughts/concerns?

gkodinov (approved):
LGTM. Please stand by for the final review.

gkodinov:
one last thing from me: this is a bugfix. I'd consider putting it in 10.6. But let's leave that to @vuvova.

client/mysqldump.cc
```diff
@@ -3498,8 +3488,17 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
 
     while ((row= mysql_fetch_row(result)))
     {
-      if (strstr(row[1],"INVISIBLE"))
-        complete_insert= 1;
+      is_generated= 0;
+      if (row[1])
```
vuvova, line 3492: old code didn't check whether `row[1]` is `NULL`, suggesting that it can never happen. Also, `sql_show.cc` says that
```
  Column("EXTRA",                   Varchar(80), NOT_NULL, "Extra", OPEN_FRM_ONLY),
```
Why did you add the check?

client/mysqldump.cc
```diff
@@ -3510,6 +3509,24 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
+
+      /*
+        Skip generated columns entirely - they will be computed by the server.
+        Enable complete_insert mode so we can omit them from the column list.
+      */
+      if (is_generated && !path && !opt_dir && !opt_dump_history)
+      {
+        complete_insert= 1;
+        continue;
+      }
```
vuvova, line 3521: Hm, in your comment https://github.com/MariaDB/server/pull/3838#issuecomment-3802282442 you wrote that you support generated columns

> in both default mode and when using `--complete-insert`.

and now I see that you enable `complete_insert` again as in your earliest commit. Why?

client/mysqldump.cc
```diff
@@ -3510,6 +3509,24 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
+
+      dynstr_append_mem_checked(&field_flags, "", 1);
+      if (row[3] && strcmp(row[3], "vector") == 0)
+        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VECTOR;
+      /* Mark system versioning columns */
+      if (row[2] && (strcmp(row[2], "ROW START") == 0 || strcmp(row[2], "ROW END") == 0))
```
vuvova, line 3527: this repeated `strcmp`-s don't look pretty. May be you could do, like
```c
bool is_row_start= row[2] && strcmp(row[2], "ROW START") == 0;
bool is_row_end= row[2] && strcmp(row[2], "ROW END") == 0;
```

client/mysqldump.cc
```diff
@@ -3589,7 +3603,7 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
 
     if (complete_insert)
       dynstr_append_checked(&insert_pat, insert_field_names.str);
-    num_fields= mysql_num_rows(result) + (vers_hidden ? 2 : 0);
+    num_fields= field_flags.length;
```
vuvova, line 3606: cool

client/mysqldump.cc
```diff
@@ -3641,33 +3655,43 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
+      is_generated= 0;
+      is_vector= 0;
+
       /* VECTOR doesn't have a type code yet, must be detected by name */
       if (strncmp(row[SHOW_TYPE], STRING_WITH_LEN("vector(")) == 0)
-        dynstr_append_checked(&field_flags, MYSQL_TYPE_VECTOR);
-      else
-        dynstr_append_checked(&field_flags, " ");
+        is_vector= 1;
+      else if (row[SHOW_EXTRA])
```
vuvova, line 3667: again, `row[EXTRA]` doesn't seem to be able to have NULLs

client/mysqldump.cc
```diff
@@ -3641,33 +3655,43 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
+      {
+        if (strstr(row[SHOW_EXTRA], "GENERATED") &&
+            !strstr(row[SHOW_EXTRA], "ROW START") &&
+            !strstr(row[SHOW_EXTRA], "ROW END"))
+          is_generated= 1;
+      }
+      if (row[SHOW_EXTRA] && strstr(row[SHOW_EXTRA], "INVISIBLE"))
+        complete_insert= 1;
```
vuvova, line 3675: it's XML path, does `complete_insert` even do something in `--xml` ?

client/mysqldump.cc
```diff
@@ -3641,33 +3655,43 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
+
+      /*
+        Skip generated columns entirely - they will be computed by the server.
+        Enable complete_insert mode so we can omit them from the column list.
+      */
+      if (is_generated && !path && !opt_dir && !opt_dump_history)
+      {
+        complete_insert= 1;
```
vuvova, line 3683: same comment about forcing `complete_insert` as above

client/mysqldump.cc
```diff
@@ -3641,33 +3655,43 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
+        continue;
+      }
+
+      dynstr_append_mem_checked(&field_flags, "", 1);
+      if (is_vector)
+        field_flags.str[field_flags.length-1]|= MYSQL_TYPE_VECTOR;
+      /* Mark system versioning columns */
+      if (row[SHOW_EXTRA] && (strstr(row[SHOW_EXTRA], "ROW START") ||
+                              strstr(row[SHOW_EXTRA], "ROW END")))
```
vuvova, line 3692: same comment about repeated `strcmp`-s

client/mysqldump.cc
```diff
@@ -4556,8 +4599,9 @@ static void dump_table(const char *table, const char *db, const uchar *hash_key,
         */
         is_blob= field->type == MYSQL_TYPE_GEOMETRY ||
                  field->type == MYSQL_TYPE_BIT ||
-                 field_flags.str[i] == MYSQL_TYPE_VECTOR[0] ||
+                 field_flags.str[i] & MYSQL_TYPE_VECTOR ||
                  (opt_hex_blob && field->charsetnr == 63 &&
+                  !(field_flags.str[i] & MYSQL_TYPE_VERS_COL) &&
```
vuvova, line 4604: why? I thought it worked without this condition perfectly well too

mysql-test/main/mysqldump.test
```diff
@@ -32,6 +32,59 @@ INSERT INTO t1 VALUES (1), (2);
...
+CREATE TABLE t1 (pk INTEGER, a INTEGER, b INTEGER, c VARCHAR(16),
+                 sum INTEGER GENERATED ALWAYS AS (a+b),
+                 sub VARCHAR(4) GENERATED ALWAYS AS (SUBSTRING(c, 1, 4)),
+                 key k1(sum),
+                 key k2(sub)
+) engine=innodb;
+INSERT INTO t1(pk, a, b, c) VALUES (1, 11, 12, 'oneone'), (2, 21, 22, 'twotwo');
+SELECT * FROM t1;
+--echo # Generated columns are omitted from INSERT statements
+--exec $MYSQL_DUMP --compact --no-autocommit=0 --no-create-info --skip-comments dump_generated t1
```
vuvova, line 51: this needs more tests.
with and without `--extended-insert`, with `--xml` and with `--tab`. Showing the content every time, because simple dump/reload works identically with and without your changes, that is, if someone breaks or reverts your code unintentionally, dump/reload tests **will not detect it**.

With virtual and stored columns. With visible and invisible generated columns.

vuvova (changes requested):
First, I wouldn't call it a bug. It's definitely **not a programming mistake** that generated columns are present in the dump, it's quite intentional, even if not ideal.
Furthermore, their presence doesn't break anything. They're unsightly and unnecessary increase the size of the dump, but they don't make it un-restorable.
So this PR is a feature, an improvement, but not a bug fix.

Second, the main concern is that you force extended inserts. And this needs more tests. See comments inline.

client/mysqldump.cc
FarihaIS: I had changed it as per @gkodinov's suggestion in this comment: https://github.com/MariaDB/server/pull/3838#pullrequestreview-3966456308 

> what good is to have a "DEFAULT" set as a value?
>
> This just increases the size of the dump for not reason I can think of.
>
> How about you just don't print the generated columns at all?

client/mysqldump.cc
FarihaIS: Ah yes, seems I mistakenly introduced that check. I will remove it and its other occurrences as well!

client/mysqldump.cc
FarihaIS: Agreed, thank you the suggestion. I will follow your recommendation above and change it for all occurrences - please let me know if I end up missing any!

client/mysqldump.cc
FarihaIS: Agreed, will remove it here too

client/mysqldump.cc
FarihaIS: This `INVISIBLE → complete_insert` logic is pre-existing, not introduced by this PR. That said, this path is also the fallback when `SQL_QUOTE_SHOW_CREATE` fails, not only for `--xml`. 

I'm happy to guard it with `if (!opt_xml)` if you'd prefer though!

client/mysqldump.cc
FarihaIS: I had changed it as per @gkodinov's suggestion in this comment: https://github.com/MariaDB/server/pull/3838#pullrequestreview-3966456308 

> what good is to have a "DEFAULT" set as a value?
>
> This just increases the size of the dump for not reason I can think of.
>
> How about you just don't print the generated columns at all?

client/mysqldump.cc
FarihaIS: I will follow your recommendation above and change it for all occurrences as above - please let me know if I end up missing any!

client/mysqldump.cc
FarihaIS: That's true, it was left over from one of my earlier fixes, will remove it

mysql-test/main/mysqldump.test
FarihaIS: Thanks for the suggestion, I'll add some more tests - could you check if the coverage is sufficient afterwards?

FarihaIS:
@vuvova Thanks for your feedback again! I've addressed/responded to all your comments above - please let me know if the PR is fine now or if it still needs work, thank you.

> So this PR is a feature, an improvement, but not a bug fix.

Makes sense, I have modified the commit message and PR description to reflect this as well!

client/mysqldump.cc
vuvova: it's the difference between
```sql
INSERT INTO t1 (column_name_1, column_name_2, column_name_3, column_name_4, column_name_5, column_name_6, column_name_7, column_name_8, column_name_9) VALUES (1,2,3,4,5,6,7,8,9);
```
and
```sql
INSERT INTO t1 VALUES (1,2,3,4,5,6,7,8,9, DEFAULT);
```
here the 10th column is generated. This shows what good it is, isn't it? :innocent: 

My point is — it's for user to decide. If the user **requested** extended insert with column names, then you simply omit generated columns, and if the user requested an insert without column names, then you use `DEFAULT`.

This is different from invisible columns that simply cannot be dumped without column names. At all.

client/mysqldump.cc
vuvova: No, never mind, I've already cleaned up this code a bit in 04e2eb874b3d238101f911f5a026e528fb837ec7

client/mysqldump.cc
vuvova: replied below, for some reason I went through your comments bottom-up

client/mysqldump.cc
FarihaIS: I see - yes, that's true! I added the functionality (back) that you requested - could you please check if it's what you had asked for?

client/mysqldump.cc
```diff
@@ -3629,33 +3653,40 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
-    }
-
     while ((row= mysql_fetch_row(result)))
     {
       ulong *lengths= mysql_fetch_lengths(result);
+      bool is_row_start= strstr(row[SHOW_EXTRA], "ROW START") != 0;
+      bool is_row_end= strstr(row[SHOW_EXTRA], "ROW END") != 0;
+      is_generated= 0;
+      is_vector= 0;
+
```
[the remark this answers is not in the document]
vuvova: This is basically, `opt_xml` path. I wouldn't do anything here at all, no detection, nothing.

client/mysqldump.cc
```diff
@@ -3509,6 +3508,24 @@ static uint get_table_structure(const char *table, const char *db, char *table_t
...
         }
       }
+
+      /*
+        When complete_insert is enabled, skip generated columns entirely.
+        Otherwise, include them and mark them so DEFAULT is emitted.
+      */
+      if (is_generated && !path && !opt_dir && !opt_dump_history &&
+          complete_insert)
+        continue;
```
vuvova, line 3574: This of course only works for generated columns **after the first invisible**, columns before are still dumped as DEFAULT.

The dump still works, though.

vuvova (approved):
Looks ok, thanks.

I've rebased it on top of the main branch and added a post-review commit with my changes corresponding to my two comments. If you'd like you can squash it with yours or keep separate to clearly differentiate from your changes.

I'm pushing it into 13.2 preview now.

FarihaIS:
@vuvova thank you for your review and the additional changes, I'm okay with keeping the commits as two separate ones!

mariadb-RameshSivaraman:
@FarihaIS Please fix this issue https://jira.mariadb.org/browse/MDEV-40217

FarihaIS:
@mariadb-RameshSivaraman thank you for pointing this out - could you please take a look at the updated fix and new tests I added to this PR? Thank you

mariadb-RameshSivaraman:
@FarihaIS bug fix looks good.

grooverdan:
@FarihaIS , late request, can you update ./man/mariadb-dump.1 to reflect the changes and additional meaning to `--complete-insert`

PR #3840, opened 2025-02-17
MDEV-14091 Support password protected SSL key in server.

vaintroub:
- [x] *The Jira issue number for this PR is: MDEV-14091*

## Description
Add ssl_passphrase server parameter, which works almost identical as --passout/--passin openssl command line parameters, described in  https://docs.openssl.org/3.4/man1/openssl-passphrase-options/.

Pass phrase value can be formatted as follows.

- pass:password Provide actual password after the pass: prefix.

- env:var Obtain the password from the environment variable 'var'

- file:pathname Reads the password from the specified file pathname. Only the first line, up to the newline character, is read from the stream.

OpenSSL passin/passout supports 2 more options that we will not support - stdin and fd:num  (we will not support because we do funny things with server's stdin,  and because fd:num is too Unix-specific), but "file:" should be good enough for most cases.

We won't offer Postgres-style handling, in which "ssl-passphrase-cmd" is a command to be executed, and its stdout would be read. This option provides some flexibility, but questionable from security standpoint, so we better not follow postgres lead here

## Release Notes
Describe ssl_passphrase server variable as in the description, optionally add link to https://docs.openssl.org/3.4/man1/openssl-passphrase-options/ , where value formatting/handling is taken from 

## How can this PR be tested?

Specific test for the new option is added, and existing one is modified to test messages produced by bad option values.

## PR quality check


mysql-test/main/variables.result
```diff
@@ -971,6 +972,7 @@ SSL_CIPHER	#
 SSL_CRL	#
 SSL_CRLPATH	#
 SSL_KEY	#
+SSL_PASSPHRASE	#
```
vuvova, line 975: is it a good idea? I can imagine users objecting to having `SSL_PASSPHRASE` in `SHOW VARIABLES`

shall server remove the passphrase from argv after parsing?

when the server reads the passfile, before or after dropping the privileges?

vio/viosslfactories.c
```diff
@@ -304,6 +307,124 @@ static long vio_tls_protocol_options(ulonglong tls_version)
...
+/**
+  Read a passphrase from a given string
+  @param buf      Buffer to store the passphrase
+  @param size     Size of the buffer
+  @param pass     clear text passphrase
+  @retval Length of the passphrase
+*/
+static int passwd_from_string(char *buf, int size, const char *pass)
+{
+  int len= (int) snprintf(buf, size, "%s", pass);
```
vuvova, line 346: snprintf? well, ok, but unusual. I'd expect `strmake` or something

vio/viosslfactories.c
```diff
@@ -304,6 +307,124 @@ static long vio_tls_protocol_options(ulonglong tls_version)
...
+  struct Handler
+  {
+    const char *prefix;
+    size_t prefix_len;
+    int (*handler)(char *, int, const char *);
+  };
+
+  static const struct Handler handlers[]=
+  {
+    {"pass:", 5, passwd_from_string},
```
vuvova, line 399: `STRING_WITH_LEN()`

vuvova (commented):
A couple of comments inline.

Using openssl as a role model looks like a good idea. We can still add `exec:` or `file:|` later if needed.

The file looks like the most secure and thus recommended way, but as it must be readable by the mysql user (for `FLUSH SSL`) it has to be outside of `secure-file-priv` (or in the datadir). This applies to cert files too.

vaintroub:
> The file looks like the most secure and thus recommended way, but as it must be readable by the mysql user (for `FLUSH SSL`) it has to be outside of `secure-file-priv` (or in the datadir). This applies to cert files too.

Do you think we should ignore the passphrase if is_secure_file_path() evaluates to true, for the part after "file:"?  What if `secure-file-priv` is not set? Does datadir have a special meaning?

mysql-test/main/variables.result
vaintroub: I remove the value in the latest patch, leaving variable name visible, but it will show value NULL, even if it was set in config file or command line.
I'm not sure we should remove the variable itself, it feels even more weird. if it is in the config file, but not mentioned in "show variables". People could at least get an idea of how variable is correctly spelled, and config file validators, if there are any outside, would continue to work.

vuvova:
> Do you think we should ignore the passphrase if is_secure_file_path() evaluates to true, for the part after "file:"? What if `secure-file-priv` is not set? Does datadir have a special meaning?

May be fail to start (with an error), if one can use `LOAD DATA INFILE` or `LOAD_DATA()` function to read the file?

if `secure-file-priv` is not set it means any path outside of datadir can be read. mariadbd will always refuse to read files from inside the datadir, as far as I remember, otherwise users with FILE privilege would be able to read any table whatsoever, whether they were granted SELECT on it or not.

mysql-test/main/variables.result
vuvova: so it's always NULL whether it was set or not? May be set it to "file:xxx", "pass:xxx", "env:xxx" ? these could even be static hard-coded values, no need to generate them at run time. (or "file:" or "file:***" whatever)

vaintroub:
> > Do you think we should ignore the passphrase if is_secure_file_path() evaluates to true, for the part after "file:"? What if `secure-file-priv` is not set? Does datadir have a special meaning?
> 
> May be fail to start (with an error), if one can use `LOAD DATA INFILE` or `LOAD_DATA()` function to read the file?
> 
> if `secure-file-priv` is not set it means any path outside of datadir can be read. mariadbd will always refuse to read files from inside the datadir, as far as I remember, otherwise users with FILE privilege would be able to read any table whatsoever, whether they were granted SELECT on it or not.

Well, it does read any table whatsoever for me (on Windows). I debugged a bit, on Linux it fails `if (!(stat_info.st_mode & S_IROTH))` inside `is_secure_path()` on .frms and other database files, which does not work on Windows, because security is ACL-based rather than "user,group,others" and stat() can't adequately translate it. Anyway, there is no "datadir" limitation, which I think could really make sense. But there is none

vuvova:
Indeed, sorry for confusion. It seems to be a limitation only for DATA/INDEX DIRECTORY.

Then, I don't know, a warning if `is_secure_file_path()` is true?

mysql-test/main/variables.result
vaintroub: Fixed according to this proposal, i.e it would be "file:", "env:", or "pass:" if it was set.

vaintroub:
> Indeed, sorry for confusion. It seems to be a limitation only for DATA/INDEX DIRECTORY.
> 
> Then, I don't know, a warning if `is_secure_file_path()` is true?

Did this, in the last commit (warning  if `is_secure_file_path()` is true)

vuvova (approved):

PR #3844, opened 2025-02-18
MDEV-36124 Fix missing binary failures from test_upgrade.sh

FarihaIS:
## Description

Some jobs in the test stage of the pipeline are failing due to `mysql_install_db` not being found in the expected location for newer versions (11.4) of mariadb when executing `test_upgrade.sh`. In particular, the test failures seen so far have the following error message: `./tests/upgrade_from/test_upgrade.sh: line 186: /usr/bin/mysql_install_db: No such file or directory`. This suggests the installations are not correctly executing due to changes made to version 11.4, where either the name and/or location of the binary has changed.

Fix missing binary failures from test_upgrade.sh:

* Modify commands that look for variables and binaries to account for both mysql and mariadb prefixes of the binaries
* Add more directories to the list of locations where the binaries are searched for

## Release Notes

N/A

## How can this PR be tested?

This can be tested easily by simply running GitLab CI on this branch before and after the changes in this PR.

Before this change:

before

After this change:

after

This change targets the `main` branch to resolve the recent regression in the fedora upgrade-from pipeline.

Note that this has no functional changes to MariaDB Server itself and is thus, safe to merge. The faster this is merged, the faster we can start catching other testable regressions.

## Copyright

### DIFF
--- a/tests/upgrade_from/test_upgrade.sh
+++ b/tests/upgrade_from/test_upgrade.sh
@@ -77,7 +77,7 @@ if [[ -z "$target_version" ]]; then
         echo "Using default RPM directory"
         rpm_dir="rpm/"
     fi
-    if [ ! -d $rpm_dir ]; then
+    if [ ! -d "$rpm_dir" ]; then
         echo "Error: directory $rpm_dir not found"
         usage
         exit 1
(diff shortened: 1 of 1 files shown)


cvicentiu (approved):
@fauust Can you please review the bash changes here, when it comes to best practices.

I checked the logic and it seems ok.

@FarihaIS OK to push from my side, I'll wait for @fauust to offer some advice. I believe we use shellcheck linting in buildbot and that one at least suggests more quoting of strings in general, not less, as you've changed here.

FarihaIS:
> @FarihaIS OK to push from my side, I'll wait for @fauust to offer some advice. I believe we use shellcheck linting in buildbot and that one at least suggests more quoting of strings in general, not less, as you've changed here.

Thank you for the feedback, I've used shellcheck to remove all linting warnings/errors. 

For context, I previously removed quotations from log statements to make it more consistent since a lot of the log statements were missing them. However, I've now added quotations to all the log statements that were missing them for consistency.

FarihaIS:
> And I agree with @cvicentiu, please re-add quoting, not that it's mandatory here since the `log` function use all arguments (`$*`), but it's homogeneous with the rest of the script, it looks better in editor and it anticipate when we will decide to shellcheck those bash scripts.
> 
> Also, in the future, you might want to split your commit between cosmetic changes and functional changes. This really helps reviewers.

Thank you, I've used shellcheck linting to remove the linting warnings and errors from previous changes as well as mine. 

I realize this pull request now contains both cosmetic and functional changes in an attempt to resolve your feedback regarding the quotes, but I'll try and separate the both for future pull requests.

cvicentiu (approved):

fauust:
> I realize this pull request now contains both cosmetic and functional changes in an attempt to resolve your feedback regarding the quotes, but I'll try and separate the both for future pull requests.

It does not need to be in different PR, but different commits in the PR, that helps a lot at the moment of reviewing. Anyway, thanks a lot for your contribution!

PR #3846, opened 2025-02-19
12.0 mdev 36094 rowid reverse scan

DaveGosselin-MariaDB:
This PR has two commits:

Commit 95a2739f95c5843fff3c39d7a2c2cde2d7ed96a4 contains new tests demonstrating that the rowid filtering optimization is not used for reverse ordered scans when it otherwise could be.  This commit adds two new session counters for rowid filtering and the tests inspect them.

Commit 8fbba548d54291050af2bc06ed5d3e96f714dc88 enables rowid filtering for reverse ordered scans and updates tests from commit 95a2739f95c5843fff3c39d7a2c2cde2d7ed96a4 to demonstrated that the filtering is in place.


spetrunia:
The patch adds `Handler_rowid_filter_attempts` and ...matches counters. Is there any problem with looking at ANALYZE FORMAT=JSON output, instead?

I'm wondering if these counters would make sense for the end-users (i.e. non-database devs). Globals like "[index] rows read" seem to make sense, but I have a doubt whether global "rowid filter checks" is useful.

DaveGosselin-MariaDB:
> The patch adds `Handler_rowid_filter_attempts` and ...matches counters. Is there any problem with looking at ANALYZE FORMAT=JSON output, instead?
> 
> I'm wondering if these counters would make sense for the end-users (i.e. non-database devs). Globals like "[index] rows read" seem to make sense, but I have a doubt whether global "rowid filter checks" is useful.

I'm not sure what you mean by "The patch adds `Handler_rowid_filter_attempts` and ...matches counters"; to clarify, the first commit shows that the new counters are zero, then the second shows that they're nonzero, reflecting that the changes to enable rowid filtering for reverse scans is in effect.

I'm happy to remove the counters and look at `analyze format=json` output using `json_extract` instead, especially if you think that the counters are not useful.  I don't want to add something that you think isn't needed.  😊

mysql-test/main/rowid_filter_innodb.result
```diff
@@ -3778,16 +3778,30 @@ ANALYZE
           "key_length": "5",
           "used_key_parts": ["a"],
           "ref": ["const"],
+          "rowid_filter": {
```
spetrunia, line 3781: Note that a few lines above, on line 3755, there is a note: 
```
# The following must NOT use Rowid Filter:
```
It needs to be updated.

mysql-test/main/rowid_filter_innodb.result
```diff
@@ -3778,16 +3778,30 @@ ANALYZE
...
+            "r_rows": 250,
+            "r_lookups": 5251,
+            "r_selectivity_pct": 4.760997905,
+            "r_buffer_size": "REPLACED",
+            "r_filling_time_ms": "REPLACED"
+          },
           "loops": 1,
           "r_loops": 1,
           "rows": 250,
+          "r_index_rows": 5251,
```
spetrunia, line 3797: Doesn't this show that the Rowid Filter with ICP has the "walk-off-the-end-of-range" problem? 

r_index_rows=5251.  This is 20x what the server expected (rows=250).

Just to be sure, I've added to the test the same query  with ORDER BY ASC.  The output is:
https://gist.github.com/spetrunia/7ae335d01775400bb607672111c202ab

Key place:
```
          "rowid_filter": {
            "range": {
              "key": "b",
              "used_key_parts": ["b"]
            },
            "rows": 252,
            "selectivity_pct": 4.799085888,
            "r_rows": 250,
            "r_lookups": 250,
            "r_selectivity_pct": 100,
            "r_buffer_size": "REPLACED",
            "r_filling_time_ms": "REPLACED"
          },
          "loops": 1,
          "r_loops": 1,
          "rows": 250,
          "r_index_rows": 250,
          "r_rows": 250,
```
note r_index_rows=rows= 250, note r_lookups=250.
This is how it should have worked.

Please investigate this.

spetrunia (changes requested):
Need attion on the above comments.

mysql-test/main/rowid_filter_innodb.result
DaveGosselin-MariaDB: Thanks, I missed that (obviously).

mysql-test/main/rowid_filter_innodb.result
DaveGosselin-MariaDB: Interesting, yes I will investigate, thanks.

DaveGosselin-MariaDB: In the case of this query, there is no pushed index condition.  That is important because `handler_rowid_filter_check` will take the `if (!h->pushed_idx_cond)` path which leads to checking the end range on the handler.  However, that check is not successful because `end_range` is null.  Unfortunately, this query does not use `QUICK_SELECT_DESC`, so the `set_end_range` API is not invoked which is why `end_range` is null.  While there is no pushed index condition, there is a pushed rowid filter.

Is there a way to construct an end key during `join_read_last_key`?  If so, then at that point in the code we could call `set_end_range` on `table->file`.  That would allow the range check in `handler_rowid_filter_check` to proceed and prevent the check from "running off the end."  If not, and in the case that there is not a pushed index condition, we could continue to delete the rowid filter from `prepare_for_reverse_ordered_access` (but that is not a preferable outcome).

spetrunia: It seems I was able to construct a case where the problem exists with ICP too:
On the dataset from the same example,  MDEV-33875, run:

```sql
set optimizer_switch='rowid_filter=off';

analyze format=json 
select * 
from t1 
where a =30100 and b in (30100,30101,30102) and (a+pk)>30100+8439
order by pk desc;
```
(full ANALYZE output: https://gist.github.com/spetrunia/861fd76c8b063c1be40011491c0a814b )

spetrunia: > Is there a way to construct an end key during join_read_last_key? If so, then at that point in the code we could call set_end_range on table->file

Looking at this.

DaveGosselin-MariaDB: @spetrunia thank you, I will construct it.  I was unsure if it was possible from looking at the types in scope here (e.g. `tab->ref.key_buff` is `uchar*` which could be anything, and `set_end_range` requires `key_range` which itself has a `uchar*` key as its first member).  Typically I rely heavily on data types to understand how things will fit together but they aren't so clear around key handling and key construction.

DaveGosselin-MariaDB: @spetrunia creating the end range and setting it on the handler fixes both (1) the query from the test that was modified in this PR and (2) the example that you found when setting rowid filtering off.  I'll update the PR soon, after completing additional testing.  Thanks again!

sql/sql_select.cc
```diff
@@ -25012,6 +25012,20 @@ join_read_last_key(JOIN_TAB *tab)
...
+    scanning once it has hit that point.
+   */
+  key_range range_endpoint
+  {
+    tab->ref.key_buff,
+    tab->ref.key_length,
+    make_prev_keypart_map(tab->ref.key_parts),
+    HA_READ_PREFIX_LAST
+  };
+  table->file->set_end_range(&range_endpoint, handler::RANGE_SCAN_DESC);
```
spetrunia, line 25027: Ok.
I note that this is not wrapped in 
```
if (table->file->pushed_rowid_filter || table->file->pushed_index_cond... ) { ... 
```
but let's keep it this way.

spetrunia: ... But I think this change should be in https://github.com/MariaDB/server/pull/3837 , not here (JOIN-READ-LAST-KEY-SHOULD-BE-IN-ICP).
Please move it into that patch ( and have this PR be on top of modified ICP, not sure if it will happen automatically or not). 

The testcase should also go into that PR.

mysql-test/main/mdev-34413-icp-reverse-order.result
```diff
@@ -84,8 +84,8 @@ a	b	c
 10	10	10
 SELECT * FROM information_schema.SESSION_STATUS WHERE VARIABLE_NAME LIKE '%icp%';
 VARIABLE_NAME	VARIABLE_VALUE
-HANDLER_ICP_ATTEMPTS	2
-HANDLER_ICP_MATCH	2
+HANDLER_ICP_ATTEMPTS	1
```
spetrunia, line 87: Why this change? The query above doesn't use the Rowid Filter. 
I assume this will be gone once JOIN-READ-LAST-KEY-SHOULD-BE-IN-ICP suggestion is addressed.

spetrunia (changes requested):
Please address input, I will also post in https://github.com/MariaDB/server/pull/3837
about extra change we need here.

sql/sql_select.cc
DaveGosselin-MariaDB: No problem, will do.  👍

PR #3847, opened 2025-02-20
MDEV-33671 Remove hardcoded open-files limit in safe_process.cc

MohanadKh03:
This commit allows specifying the open-files limit dynamically instead of previously hard coding it with 1024 files  using a new option in MTR ```open-files-limit=X``` 

- [x] *The Jira issue number for this PR is: [MDEV-33671](https://jira.mariadb.org/browse/MDEV-33671)

## Description
With only 1024 open file descriptors it is impossible to use
mtr for high concurrency testing.

This commit removes the hardcoded value and adds a new
option to control the limit when running tests: ```./mtr --open-files-limit=X.```

The default is still 1024, but it can now be changed when needed,
making it easier to test scenarios that require different number of
open file descriptors at the same time e.g: partition_notwin.test

## How can this PR be tested?
This is part of the MTR infrastructure so it can be tested manually and locally using the ```--open-filies-limit=<value>```
e.g: ```./mtr --open-files-limit=1000 alter_algorithm.test```

Note that in https://github.com/MariaDB/server/commit/7d17ee97c2d07a1497c11aa475505d0a993563d0 it caused a test failure by succeeding when having the default number of files before but now it fails as it should when the number of open files is not correct 

- [] *This is a new feature or a refactoring, and the PR is based against the `main` branch.*

## PR quality check

### DIFF
--- a/mysql-test/lib/My/SafeProcess/safe_process.cc
+++ b/mysql-test/lib/My/SafeProcess/safe_process.cc
@@ -220,6 +220,7 @@ int main(int argc, char* const argv[] )
   pid_t own_pid= getpid();
   pid_t parent_pid= getppid();
   bool nocore = false;
+  int open_files_limit = 1024;
   struct sigaction sa,sa_abort;
 
   sa.sa_handler= handle_signal;
@@ -268,7 +269,14 @@ int main(int argc, char* const argv[] )
       }
       else if ( strncmp (arg, "--env ", 6) == 0 )
       {
-	putenv(strdup(arg+6));
+        putenv(strdup(arg+6));
+      }
+      else if ( strncmp(arg, "--open-files-limit=", 19) == 0 )
+      {
+        const char* start = arg + 19;
+        open_files_limit = atoi(start);
+        if (open_files_limit <= 0)
+          die("Invalid value '%s' passed to --open-files-limit", start);
       }
       else
         die("Unknown option: %s", arg);
@@ -318,11 +326,8 @@ int main(int argc, char* const argv[] )
     if (nocore)
       setlimit(RLIMIT_CORE, 0, 0);
 
-    /*
-      mysqld defaults depend on that. make test results stable and independent
-      from the environment
-    */
-    setlimit(RLIMIT_NOFILE, 1024, 1024);
+    // Set open files limit 
+    setlimit(RLIMIT_NOFILE, open_files_limit, open_files_limit);
 
     // Signal that child is ready
     buf= 37;
--- a/mysql-test/lib/My/SafeProcess.pm
+++ b/mysql-test/lib/My/SafeProcess.pm
@@ -138,6 +138,7 @@ sub new {
   my $error    = delete($opts{'error'});
   my $verbose  = delete($opts{'verbose'}) || $::opt_verbose;
   my $nocore   = delete($opts{'nocore'});
+  my $open_files_limit = delete($opts{'open_files_limit'});
   my $host     = delete($opts{'host'});
   my $shutdown = delete($opts{'shutdown'});
   my $user_data= delete($opts{'user_data'});
@@ -160,6 +161,8 @@ sub new {
 
   push(@safe_args, "--verbose") if $verbose > 0;
   push(@safe_args, "--nocore") if $nocore;
+
+  push(@safe_args, "--open-files-limit=$open_files_limit") if $open_files_limit;
 
   # Point the safe_process at the right parent if running on cygwin
   push(@safe_args, "--parent-pid=".Cygwin::pid_to_winpid($$)) if IS_CYGWIN;
--- a/mysql-test/mariadb-test-run.pl
+++ b/mysql-test/mariadb-test-run.pl
@@ -130,6 +130,8 @@ our $path_language;
 our $path_current_testlog;
 our $path_testlog;
 
+our $opt_open_files_limit;
+
 our $default_vardir;
 our $opt_vardir;                # Path to use for var/ dir
 our $plugindir;
@@ -1277,6 +1279,7 @@ sub command_line_setup {
 	     'list-options'             => \$opt_list_options,
              'skip-test-list=s'         => \@opt_skip_test_list,
              'xml-report=s'             => \$opt_xml_report,
+             'open-files-limit=i',      => \$opt_open_files_limit,
 
              My::Debugger::options(),
              My::CoreDump::options(),
@@ -5767,6 +5770,7 @@ sub start_mysqltest ($) {
      append        => 1,
      error         => $path_current_testlog,
      verbose       => $opt_verbose,
+     open_files_limit => $opt_open_files_limit,
     );
   mtr_verbose("Started $proc");
   return $proc;
@@ -6065,6 +6069,8 @@ Misc options
   timediff              With --timestamp, also print time passed since
                         *previous* test started
   max-connections=N     Max number of open connection to server in mysqltest
+  open-files-limit=N    Max number of open files allowed for any of the children
+                        of my_safe_process. Default is 1024.
   report-times          Report how much time has been spent on different
                         phases of test execution.
   stress=ARGS           Run stress test, providing options to


cvicentiu:
@MohanadKh03 preliminary review:

Check the failing test builders and try to get those passing. Once those are green (or you can't figure out how to fix it) ask for more input.

MohanadKh03:
> Check the failing test builders and try to get those passing. Once those are green (or you can't figure out how to fix it) ask for more input.

I think by installing BSD::Resource as a dependency this would make the CI green , is it okay to add this to the dependencies ? 
If so how can it be modified because I cant really find the part where it can  be added in the CI

svoj:
@grooverdan @vuvova @ottok @cvicentiu how do you feel about having extra `libbsd-resource-perl` dependency for this feature?

IMHO it is a bit too much for this rather minor issue. I'd rather go either with direct `syscall` or pass an argument to `safe_process` (which @vuvova didn't like). Probably there're more options?

vuvova:
I've fairly neutral about it. I don't think that `BSD::Resource` dependency is such a big deal, nor I particularly mind an alternative fix.

How would you do it with a syscall?

cvicentiu:
I would add an extra cmd line parameter to safe process and then we don't need an extra dependency.

svoj:
@vuvova there're examples on stackoverflow like
```
$rstruct= pack "L!L!",1,10;
syscall(&SYS_setrlimit, &RLIMIT_CPU, $rstruct);
```

Most probably not portable, but I'd guess having it functional on Linux is enough.

svoj:
I feel like we should remove `setlimit(RLIMIT_NOFILE)` altogether. Probably add a check to `safe_process` that makes sure hard limit is at least 1024 and issues a warning or an error otherwise.

I tried lowering the limit:
```
setlimit(RLIMIT_NOFILE, 1024, 10) test suite passed.
setlimit(RLIMIT_NOFILE, 10, 10) test suite failed.
setlimit(RLIMIT_NOFILE, 128, 128) test suite failed, although lots of tests passed.
setlimit(RLIMIT_NOFILE, 256, 256) test suite failed, even more tests passed.
setlimit(RLIMIT_NOFILE, 512, 512) test suite failed.
setlimit(RLIMIT_NOFILE, 1000, 1000) test suite failed.
```
Apparently some tests (e.g. main.mysqld--help-aria) expect hard limit to be at least 1024.

I tried raising the limit:
```
setlimit(RLIMIT_NOFILE, 1048576, 1048576) test suite passed.
setlimit() commented out, OS inherited values of 1048576 1024 test suite passed.
```

To sum up: higher values of hard limit (>1024) don't seem to be a problem. Lower values are a problem, but they can be increased only by privileged (e.g. root or `CAP_SYS_RESOURCE`) process. At this point I don't see presence of `setrlimit()` to be justified.

Unfortunately e9b6f95013b0 didn't explain why it was added originally.

MohanadKh03:
> To sum up: higher values of hard limit (>1024) don't seem to be a problem. Lower values are a problem, but they can be increased only by privileged (e.g. root or CAP_SYS_RESOURCE) process. At this point I don't see presence of setrlimit() to be justified.

But using it would at least allow having it passed as an option 
Do you suggest not using it at all even in the perl code ? If so then how would this be an option to use instead of hardcoding it ?

BSD::Resource messed the build up so according to the previous comments it is ok to just use direct syscalls for Linux only so if that is okay ill update and push this using syscall

svoj:
@MohanadKh03 right, I don't think we need it neither in perl nor in safe_process. We do have it in `mariadbd` anyway, one can do `mtr --mysqld=--open-files-limit=X` or set in cnf files. Although it sets lower boundary for the limit passed to `setrlimit`. There's no option to control higher boundary, which doesn't seem to be a problem

MohanadKh03:
Ok so some updates 
When I removed it from safe_process.cc and from perl the ```--mysqld=--open-files-limit=X``` actually does not seem to get parsed at all after some printing and debugging it seems to just calculate a default value which does not change when i try ```--mysqld=--open-files-limit=X``` with multiple values it just prints out the same value each time which is 32190 "probably calculated as mentioned here  https://mariadb.com/kb/en/server-system-variables/#open_files_limit

So I think by removing it from both .cc and perl we would need to somehow parse it again ? 
This is the result when adding ```SHOW VARIABLES LIKE 'open_file_limit'``` to the ```main.mysqld--help-aria``` test file
```
+SHOW VARIABLES LIKE 'open_files_limit';
+Variable_name  Value
+open_files_limit       32190
```

It definitely doesnt seem to not get the value from mysqld option at all whenever given a value lower than 1024 or greater than it just gets that value as a default in the ```init_common_variables``` inside ```mysqld.cc```


Inside ```handle_options``` also does not seem to have this value at all I tried once running the mariadbd normally and once with MTR the first one prints the open files limit as a parsed key/value but the latter seems not to parse it at all so im guessing sth is missing from the perl code which should send the mysqld correctly ?

svoj:
@MohanadKh03 as I mentioned in my comment above **it sets lower boundary for the limit**. Look for `my_set_max_open_files()`, one line above:
```
    max_open_files= MY_MAX(MY_MAX(wanted_files,
                                  (max_connections + extra_max_connections)*5),
                           open_files_limit);
```

That is if you do `--mysqld=--open-files-limit=100000`, it will be 100000.

MohanadKh03:
> That is if you do --mysqld=--open-files-limit=100000, it will be 100000.

Right my bad 
I have updated it but not sure if the warning message should be said explicitly if the input given is below 1024 ? Since it already has a lower boundary

svoj:
@MohanadKh03 thanks for updating the patch. Right, the warning will be emitted by `mariadbd`. At this point I don't see the need for extra warning.

Given that buildbot is happy about your change and my analysis in https://github.com/MariaDB/server/pull/3847#issuecomment-2755716485...

@vuvova I feel like changes suggested in 7f12d9c are something we should go for. Please review them and suggest what kind of extra testing (if any) should be performed.

svoj (changes requested):
Your commit comment should be updated:
```
This commit removes the hardcoded --open-files-limit option
in safe_process.cc as this could be used by
./mtr --mysqld=--open-files-limit=X instead
```

Should be replaced with something like:
```
Due to hardcoded open files limit (1024) it was impossible to use mtr
framework for high concurrency testing.

Removed hardcoded setrlimit(RLIMIT_NOFILE, 1024, 1024) from
my_safe_process utility.

The call itself was inconsistent because:
1. It wasn't set on Windows.
2. If OS offers less than 1024 file descriptors, raising this limit to
   1024 requires CAP_SYS_RESOURCE/root priviledge, which is not
   something we normally do.
3. Tests didn't reveal any problems if OS offers more than 1024 file
   descriptors.
```

mysql-test/lib/My/SafeProcess/safe_process.cc
```diff
@@ -318,11 +318,6 @@ int main(int argc, char* const argv[] )
     if (nocore)
       setlimit(RLIMIT_CORE, 0, 0);
 
-    /*
-      mysqld defaults depend on that. make test results stable and independent
-      from the environment
-    */
-    setlimit(RLIMIT_NOFILE, 1024, 1024);
 
     // Signal that child is ready
```
svoj, line 322: Please also remove empty line above. We don't normally add double empty lines within functions.

vuvova (approved):

svoj (approved):

MohanadKh03:
Ok so CLA awaits 26 people somehow it seems there is a problem with the amount of commits when rebasing to 10.11 ? Should I try squashing commits ?

svoj:
@MohanadKh03 no, I will handle it from now on. We should wait till `main.timezone` test is fixed, it is failing due to CET -> CEST change.

mysql-test/lib/My/SafeProcess/safe_process.cc
```diff
@@ -268,7 +269,14 @@ int main(int argc, char* const argv[] )
       }
       else if ( strncmp (arg, "--env ", 6) == 0 )
       {
-	putenv(strdup(arg+6));
+  putenv(strdup(arg+6));
```
cvicentiu, line 272: Intendation is off here. More spaces needed.

svoj:
> +  open-files-limit=N    Max number of open files allowed for the server.
Strictly speaking it is not for the server, any process spawned with `my_safe_process` will be affected, e.g. `--exec`.

cvicentiu:
> ```
> +  open-files-limit=N    Max number of open files allowed for the server.
> ```
> 
> Strictly speaking it is not for the server, any process spawned with `my_safe_process` will be affected, e.g. `--exec`.

Agree, force pushed a fixup:
```
+  open-files-limit=N    Max number of open files allowed for any of the children
+                        of my_safe_process. Default is 1024.
```

PR #3848, opened 2025-02-20
MDEV-31134: sync galera settings with KB doc

fauust:
- [x] *The Jira issue number for this PR is: MDEV-31134*

## Description

The `wsrep_provider` setting is mandatory and needs to be present as a guidance for the setup of Galera cluster. `wsrep_cluster_name` and `innodb_autoinc_lock_mode` are not mandatory.

See: https://mariadb.com/kb/en/configuring-mariadb-galera-cluster/

## Release Notes

Update galera configuration settings as in KB doc.

## How can this PR be tested?

There is no impact since it's all commented settings but we might want to test that default settings are working. 

## PR quality check

### DIFF
--- a/debian/additions/mariadb.conf.d/60-galera.cnf
+++ b/debian/additions/mariadb.conf.d/60-galera.cnf
@@ -1,21 +1,24 @@
 #
 # * Galera-related settings
 #
-# See the examples of server wsrep.cnf files in /usr/share/mariadb
-# and read more at https://mariadb.com/kb/en/galera-cluster/
+# See the examples of server wsrep.cnf files in /usr/share/mariadb and read
+# more at https://mariadb.com/kb/en/galera-cluster/ and
+# https://mariadb.com/kb/en/configuring-mariadb-galera-cluster/
 
 [galera]
-# Mandatory settings
-#wsrep_on                 = ON
 #wsrep_cluster_name       = "MariaDB Galera Cluster"
+
+# Mandatory settings
+#wsrep_provider           = /usr/lib/galera/libgalera_smm.so
 #wsrep_cluster_address    = gcomm://
-#binlog_format            = row
+#binlog_format            = ROW
+#wsrep_on                 = ON
 #default_storage_engine   = InnoDB
-#innodb_autoinc_lock_mode = 2
 
 # Allow server to accept connections on all interfaces.
 #bind-address = 0.0.0.0
 
-# Optional settings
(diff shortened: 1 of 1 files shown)


debian/additions/mariadb.conf.d/60-galera.cnf
```diff
@@ -1,21 +1,24 @@
...
 
 # Allow server to accept connections on all interfaces.
 #bind-address = 0.0.0.0
 
-# Optional settings
-#wsrep_slave_threads = 1
+# Performance related settings
+#innodb_autoinc_lock_mode = 2
 #innodb_flush_log_at_trx_commit = 0
+#wsrep_slave_threads = 1
```
grooverdan, line 24: so why is this only 1? To get any benefit it has to be more than one.

grooverdan (commented):
If its acceptable to reference non-KB docs I'd look at https://medium.com/@arbaudie.it/galera-understanding-flow-control-8775890b1332 as a reference in the performance section.

I'm happy with the rest.

debian/additions/mariadb.conf.d/60-galera.cnf
iangilfillan: The default is 1. Our docs also point out that "if your cluster nodes seem to have frequent consistency problems, then setting the value to 1 will probably fix the problem.", hopefully only referencing some old issue! Even the doc at https://galeracluster.com/library/documentation/mysql-wsrep-options.html#wsrep-applier-threads (which replaces wsrep-slave-threads in the latest version) defaults to 1, and only gives very loose suggestions for changing.

fauust:
> If its acceptable to reference non-KB docs I'd look at https://medium.com/@arbaudie.it/galera-understanding-flow-control-8775890b1332 as a reference in the performance section.

This PR is only about adding a mandatory setting to help setup a Galera cluster. If we want to improve our configuration example file with more fine tuning, I suggest we create a new task/PR for that.

debian/additions/mariadb.conf.d/60-galera.cnf
fauust: So, can we decide what to do about it?

cvicentiu: @fauust set this to 8 in the comment please. Then ok to merge.

cvicentiu (changes requested):

PR #3849, opened 2025-02-21
`rpl_start`/`stop_server.inc` fixes

ParadoxV5:
* ~~*The Jira issue number for this PR is: MDEV-______*~~

## Description
* `rpl_stop_server.inc`: re. “can’t use `begin`/`end_include_file”, (re)define `$_include_file_depth= 0` in case this is the first `--source` in a `rpl_init.inc`-less setup (e.g., `multi_source.*`)
* `rpl_start_server.inc`: move the `end_include_file.inc` call out of `if (!$rpl_server_error)` to always close the `begin_include_file.inc`
* `rpl_reconnect.inc` (called by rpl_start_server.inc`): skip reconnecting ‘master/slave’ aliases of `server_*` to lift its practical requirement of `master-slave.inc` (It failed with “no connection named master/slave” for even a `rpl_init.inc` setup, let alone `rpl_init`-less.)

## Release Notes
*Do our users even use our `include/*.inc`s?*

## How can this PR be tested?
* [ ] *Do we need a test for a test?*

## PR quality check

### DIFF
--- a/mysql-test/include/rpl_reconnect.inc
+++ b/mysql-test/include/rpl_reconnect.inc
@@ -31,44 +31,13 @@ if (!$rpl_server_number)
 {
   --die ERROR IN TEST: you must set $rpl_server_number before you source rpl_connect.inc
 }
+--let $rpl_connection_name= server_$rpl_server_number
 
 
 if ($rpl_debug)
 {
   --echo ---- Enable reconnect ----
 }
-
---let $_rpl_server_number= $rpl_server_number
-
---dec $_rpl_server_number
-if (!$_rpl_server_number)
-{
-  --let $rpl_connection_name= default
-  --source include/rpl_connection.inc
-  --enable_reconnect
-
-  --let $rpl_connection_name= master
-  --source include/rpl_connection.inc
-  --enable_reconnect
-
-  --let $rpl_connection_name= master1
-  --source include/rpl_connection.inc
-  --enable_reconnect
-}
-
---dec $_rpl_server_number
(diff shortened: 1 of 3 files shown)


ParadoxV5:
So there are no connection aliases, only duplicate connections to the same set of servers?

knielsen (changes requested):
Sorry for the delay with this review, things were busy on my end for
the last few weeks.

> •  rpl_reconnect.inc  (called by rpl_start_server.inc ): skip reconnecting  
> ‘master/slave’ aliases of  server_* to lift its practical requirement of    
> master-slave.inc (It failed with “no connection named master/slave” for even
> a rpl_init.inc setup, let alone rpl_init`-less.)                            

Are you sure this will not cause problems for any other tests (like,
including --big tests or whatever)? Ie. that a test uses rpl_reconnect.inc
to also reconnect these extra connections?

It seems nice that rpl_reconnect.inc would take care of _all_ connections
and a test case doesn't need to do this manually for
default/master/master1/slave/slave1. Maybe simply set a flag in
include/master-slave.inc and check for that flag before trying to restore
these connections?

And the "default" connection should always be there, I think, not only with
include/master-slave.inc

Otherwise looks good, thanks! It's great that you're doing the work to
improve the test suite infrastructure as you encounter issues.

I think it's ok to do this in 10.11. Might also do it in 10.5 (or 10.6), as
this is test suite maintenance and we generally keep them same across
branches, if someone wants that. But I'm fine with 10.11.

PR #3850, opened 2025-02-21
MDEV-36184 - To optimise dot_product in Power9 and Power10 architecture

mikejuliet13:
This patch optimises the dot_product function by leveraging vectorisation through SIMD intrinsics. Specifically, the function now uses __builtin_vec_vupkhsh and __builtin_vec_vupklsh to efficiently convert input values from lower to higher data types.
This transformation enables parallel execution of multiple operations, significantly improving the performance of dot product computation on supported architectures.
Performance Analysis:
The original dot_product function does undergo auto-vectorisation when compiled with -O3. However, performance analysis has shown that the newly optimised implementation performs better on Power10 and achieves comparable performance on Power9 machines.

Output Changes:
The logical output of the dot_product function remains unchanged (i.e., it still computes the correct dot product).
With this patch, computations utilise vector registers, leading to improved performance. These optimisations are internal and do not alter any user-visible behaviour.

Potential Side Effects:
- This patch introduces architecture-specific optimisations targeted at Power9 and Power10 systems.
- The function has been extensively tested on both Power9 and Power10, where it demonstrates correctness and performance improvements.
- If executed on an older architecture (e.g., Power8 or below) that lacks support for these vector instructions, the implementation automatically falls back to DEFAULT_IMPLEMENTATION, ensuring broader compatibility.

Release Notes:
- Optimised the dot_product function using SIMD vectorisation for improved performance.
- Introduces architecture-specific optimisations for Power9 and Power10 systems.
- No changes to observable output; improvements are purely in internal computation efficiency.

### DIFF
--- a/sql/bloom_filters.h
+++ b/sql/bloom_filters.h
@@ -53,6 +53,10 @@ SOFTWARE.
 #define NEON_IMPLEMENTATION
 #endif
 #endif
+#if defined __powerpc64__ && defined __VSX__
+#include <altivec.h>
+#define POWER_IMPLEMENTATION
+#endif
 
 template <typename T>
 struct PatternedSimdBloomFilter
--- a/sql/vector_mhnsw.cc
+++ b/sql/vector_mhnsw.cc
@@ -229,6 +229,58 @@ struct FVector
   }
 #endif
 
+#ifdef POWER_IMPLEMENTATION
+  /************* POWERPC *****************************************************/
+  static constexpr size_t POWER_bytes= 128 / 8; // Assume 128-bit vector width
+  static constexpr size_t POWER_dims= POWER_bytes / sizeof(int16_t);
+
+  static float dot_product(const int16_t *v1, const int16_t *v2, size_t len)
+  {
+    // Using vector long long for int64_t accumulation
+    vector long long ll_sum= {0, 0};
+    // Round up to process full vector, including padding
+    size_t base= ((len + POWER_dims - 1) / POWER_dims) * POWER_dims;
+
+    for (size_t i= 0; i < base; i+= POWER_dims)
+    {
+      vector short x= vec_ld(0, &v1[i]);
+      vector short y= vec_ld(0, &v2[i]);
+
+      // Vectorized multiplication using vec_mule() and vec_mulo()
+      vector int product_hi= vec_mule(x, y);
+      vector int product_lo= vec_mulo(x, y);
+
+      // Extend vector int to vector long long for accumulation
+      vector long long llhi1= vec_unpackh(product_hi);
+      vector long long llhi2= vec_unpackl(product_hi);
+      vector long long lllo1= vec_unpackh(product_lo);
+      vector long long lllo2= vec_unpackl(product_lo);
+
+      ll_sum+= llhi1 + llhi2 + lllo1 + lllo2;
+    }
+
+    return static_cast<float>(static_cast<int64_t>(ll_sum[0]) +
+                              static_cast<int64_t>(ll_sum[1]));
+  }
+
+  static size_t alloc_size(size_t n)
+  {
+    return alloc_header + MY_ALIGN(n * 2, POWER_bytes) + POWER_bytes - 1;
+  }
+
+  static FVector *align_ptr(void *ptr)
+  {
+    return (FVector*)(MY_ALIGN(((intptr)ptr) + alloc_header, POWER_bytes)
+                    - alloc_header);
+  }
+
+  void fix_tail(size_t vec_len)
+  {
+    bzero(dims + vec_len, (MY_ALIGN(vec_len, POWER_dims) - vec_len) * 2);
+  }
+#undef DEFAULT_IMPLEMENTATION
+#endif
+
   /************* no-SIMD default ******************************************/
 #ifdef DEFAULT_IMPLEMENTATION
   DEFAULT_IMPLEMENTATION


sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,73 @@ struct FVector
   }
 #endif
 
+#ifdef POWER_IMPLEMENTATION
+  /************* POWERPC *****************************************************/
+  static constexpr size_t POWER_bytes = 128 / 8; // Assume 128-bit vector width
```
svoj, line 234: Coding style (here and everywhere else): `var= val`, `var+= val`, etc. That is no space before `=`.

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,73 @@ struct FVector
...
+  static constexpr size_t POWER_dims = POWER_bytes / sizeof(int16_t);
+
+  POWER_IMPLEMENTATION
```
svoj, line 237: Why `POWER_IMPLEMENTATION`? It should only be needed for x86, since they have default implementation, must not be relevant for Power.

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,73 @@ struct FVector
...
+  static float dot_product(const int16_t *v1, const int16_t *v2, size_t len)
+  {
+    long sum = 0;
```
svoj, line 240: I'd suggest to use `int64_t` here, like in default implementation. It has to be 64 bits at least, while `long` width is depending on data model.

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,73 @@ struct FVector
...
+
+      ll_sum += llhi1 + llhi2 + lllo1 + lllo2;
+    }
+
+    // Reduce the vector sum
+    sum += ll_sum[0] + ll_sum[1];
+
+    // Process remaining elements
+    for (size_t i = base; i < len; ++i)
+      sum += float(v1[i]) * float(v2[i]);
```
svoj, line 274: This must not be needed. Thanks to `fix_tail()`/`alloc_size()` tail is always set to zeroes and can safely go through vectorized calculation.

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,73 @@ struct FVector
...
+      vector int product_hi = x_hi * y_hi;
+      vector int product_lo = x_lo * y_lo;
+
+      // Convert int32_t -> int64_t for accumulation
+      vector long llhi1 = __builtin_vec_vupkhsw(product_hi);
+      vector long llhi2 = __builtin_vec_vupklsw(product_hi);
+      vector long lllo1 = __builtin_vec_vupkhsw(product_lo);
+      vector long lllo2 = __builtin_vec_vupklsw(product_lo);
+
+      ll_sum += llhi1 + llhi2 + lllo1 + lllo2;
```
svoj, line 266: This looks very complex. Why there is a mix of builtins and altivec functions? ARM had widening multiply, is there something similar on power?

svoj (changes requested):
Looks good, thanks! Although needs some polishing.

vuvova:
>  The original dot_product function does undergo auto-vectorisation when compiled with -O3. However, performance analysis has shown that the newly optimised implementation performs better on Power10 and achieves comparable performance on Power9 machines.

Do you have any numbers? How does your implementation compare with auto-vectorization? Did you benchmark it? On Power 10 and Power 9? Auto-vectorization at -O3 by what compiler version?

mikejuliet13:
> > The original dot_product function does undergo auto-vectorisation when compiled with -O3. However, performance analysis has shown that the newly optimised implementation performs better on Power10 and achieves comparable performance on Power9 machines.
> 
> Do you have any numbers? How does your implementation compare with auto-vectorization? Did you benchmark it? On Power 10 and Power 9? Auto-vectorization at -O3 by what compiler version?

I conducted benchmark tests on both Power9 and Power10 machines, comparing the time taken by the original (auto-vectorized) code and the new vectorised code. I used GCC 11.5.0 on RHEL 9.5 operating system with -O3. The benchmarks were performed using a sample test code with a vector size of 4096 and 10⁷ loop iterations.
Here are the average execution times (in seconds) over multiple runs:
Power9:
- Before change: ~16.364 s
- After change: ~16.180 s
- Performance gain is modest but measurable.

Power10:

- Before change: ~8.989 s
- After change: ~6.446 s
- Significant improvement, roughly 28–30% faster.

The final results of the dot product remained the same before and after the change, confirming functional correctness.

sql/vector_mhnsw.cc
mikejuliet13: In the updated commit, I've fixed the coding style issue by removing all spaces before the '=' operator as per the guidelines.

sql/vector_mhnsw.cc
mikejuliet13: In the updated commit, I removed all occurrences of the POWER_IMPLEMENTATION macro before individual functions. Since the code is already enclosed within the #ifdef POWER_IMPLEMENTATION block, there's no need to annotate each function separately with this macro.

sql/vector_mhnsw.cc
mikejuliet13: I've updated the code to use int64_t for the scalar variable sum. However, for vectorised arithmetic using built-in intrinsics like __builtin_vec_vupkhsw and __builtin_vec_vupklsw, I must operate on a vector type. Since we cannot directly define a vector of int64_t, I used vector long long for the vectorised operations. After the vectorised accumulation, I cast the result to int64_t with:
`sum += static_cast<int64_t>(ll_sum[0]) + static_cast<int64_t>(ll_sum[1]);`

Both int64_t and long long are 64-bit on modern systems, so this approach meets the requirements.

sql/vector_mhnsw.cc
mikejuliet13: Thanks, I removed the  loop checking for additional elements.

sql/vector_mhnsw.cc
mikejuliet13: The builtins (__builtin_vec_vupkhsh, etc.) are used because they directly map to specific PowerPC VSX instructions, ensuring efficient widening conversions.
AltiVec functions like vec_ld are used for loading vector data efficiently.
While ARM has a widening multiply instruction, PowerPC does not have a direct equivalent. Instead, Power provides explicit unpacking builtins (__builtin_vec_vupkhsh, etc.), which perform similar operations.

svoj (changes requested):
Nothing else on my mind, just some minor tweaks.
We will have to clean-up commit history (such that there is just one commit), but we should be able to do it on our side.

sql/vector_mhnsw.cc
svoj: Replied in a new review.

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,64 @@ struct FVector
...
+  }
+
+  static size_t alloc_size(size_t n)
+  {
+    return alloc_header + MY_ALIGN(n * 2, POWER_bytes) + POWER_bytes - 1;
+  }
+
+  static FVector *align_ptr(void *ptr)
+  {
+    return (FVector *)(MY_ALIGN(((intptr)ptr) + alloc_header, POWER_bytes) - alloc_header);
```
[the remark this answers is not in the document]
mikejuliet13: I have updated align_ptr to stay within 80 characters per line

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,64 @@ struct FVector
...
+      vector long long llhi1= __builtin_vec_vupkhsw(product_hi);
+      vector long long llhi2= __builtin_vec_vupklsw(product_hi);
+      vector long long lllo1= __builtin_vec_vupkhsw(product_lo);
+      vector long long lllo2= __builtin_vec_vupklsw(product_lo);
+
+      ll_sum+= llhi1 + llhi2 + lllo1 + lllo2;
+    }
+
+    // Sum the accumulated vector long long values into a scalar int64_t sum
+    sum+= static_cast<int64_t>(ll_sum[0]) + static_cast<int64_t>(ll_sum[1]);
```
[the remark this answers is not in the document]
mikejuliet13: The `sum` variable is necessary to efficiently convert `vector long long` to `float`. Directly casting `vector long long `to `float` would require extracting elements first, which adds overhead. The current approach (`vector long long → int64_t → float`) ensures precise accumulation and is more efficient.
I have updated the code to use `sum=` instead of `sum+=`

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,64 @@ struct FVector
...
+#ifdef POWER_IMPLEMENTATION
+  /************* POWERPC *****************************************************/
+  static constexpr size_t POWER_bytes= 128 / 8; // Assume 128-bit vector width
+  static constexpr size_t POWER_dims= POWER_bytes / sizeof(int16_t);
+
+  static float dot_product(const int16_t *v1, const int16_t *v2, size_t len)
+  {
+    int64_t sum= 0;
+    vector long long ll_sum= {0, 0}; // Using vector long long for int64_t accumulation
+    size_t base= ((len + POWER_dims - 1) / POWER_dims) * POWER_dims; // Round up to process full vector, including padding
```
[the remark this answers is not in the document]
mikejuliet13: I have updated the code to ensure all lines are under 80 characters by moving comments to the previous line

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,64 @@ struct FVector
...
+
+    for (size_t i= 0; i < base; i+= 8)
```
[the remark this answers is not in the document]
mikejuliet13: Thanks, I have corrected the loop increment to `i += POWER_dims` for consistency with the defined vector width

sql/vector_mhnsw.cc
```diff
@@ -229,6 +229,64 @@ struct FVector
...
+
+      // Convert int16_t -> int32_t
+      vector int x_hi= __builtin_vec_vupkhsh(x);
+      vector int x_lo= __builtin_vec_vupklsh(x);
+      vector int y_hi= __builtin_vec_vupkhsh(y);
+      vector int y_lo= __builtin_vec_vupklsh(y);
+
+      // Vectorized multiplication
+      vector int product_hi= x_hi * y_hi;
+      vector int product_lo= x_lo * y_lo;
```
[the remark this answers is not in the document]
mikejuliet13: Thanks, I have updated the implementation to use `vec_unpackh / vec_unpackl` instead of builtins for unpacking, ensuring consistency and readability. Also replaced manual widening multiplication with `vec_mule / vec_mulo` as suggested

sql/vector_mhnsw.cc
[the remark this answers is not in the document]
svoj: It is hard to verify how exactly `vector long long` is implemented, but `ll_sum[0]` must be of some different than `vector long long` type.

Still this variable is redundant, you can just do
```
return static_cast<float>(static_cast<int64_t>(ll_sum[0]) +
                          static_cast<int64_t>(ll_sum[1]));
```

Anyway, I don't mind to have it as you suggest, but it appears to be that you've missed to update `sum+=`.

sql/vector_mhnsw.cc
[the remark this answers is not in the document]
svoj: Since we switched towards `vec_mule()` / `vec_mulo()`, we don't need the "Convert int16_t -> int32_t" code anymore, right?

sql/bloom_filters.h
```diff
@@ -53,6 +53,10 @@ SOFTWARE.
 #define NEON_IMPLEMENTATION
 #endif
 #endif
+#if defined __powerpc64__
+#include <altivec.h>
```
grooverdan, line 57: Note there is no default cpu=X compile flags by default to this is just getting what is used in the distro default.

At least one distro (https://buildd.debian.org/status/package.php?p=mariadb) is "supporting" hardware beyond a normal lifetime.

[ppc64](https://wiki.debian.org/PPC64) particularly starts at POWER4.

Looking the godbot on this code, [vec_unpackh]/[vec_unpackl] require POWER8+ VSX. POWER8-> POWER9+ has one instruction different outside the loop so probably not significant in terms of performance.

Options:
1) Use target attributes like above this section of code where the non-POWER8+VSX has a tail call to the DEFAULT_IMPLEMENTATION (that has a compile time different function call) and doesn't attempt to compile the vector code. I think this would be easiest.
2. like crc32 a getauxval call and function pointers.

Can you add this boiler plate compatibility to save use from the inevitable release bug about this? Much appreciated anything that continues to compile on older platforms.

This chunk of code could be inside the vector_mhnsw.cc without determinant, but I've no strong opinion either way. Whatever is most readable and functionally correct.

svoj: @grooverdan you mean something like this?
```
#define DEFAULT_IMPLEMENTATION __attribute__ ((target ("no-altivec")))
#define POWER_IMPLEMENTATION __attribute__ ((target ("altivec")))
```

gcc documentation says that function multiversioning is supported only on x86. Not sure if it concerns only `default` target, but I can imagine `altivec/no-altivec` may not work as well.

I wouldn't probably bother with `2`, probably add cmake option and just disable SIMD explicitly (or even automatically) on those older platforms?

grooverdan: > @grooverdan you mean something like this?
> 
> ```
> #define DEFAULT_IMPLEMENTATION __attribute__ ((target ("no-altivec")))
> #define POWER_IMPLEMENTATION __attribute__ ((target ("altivec")))
> ```
> 
> gcc documentation says that function multiversioning is supported only on x86. Not sure if it concerns only `default` target, but I can imagine `altivec/no-altivec` may not work as well.

there is [target_clones](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-target_005fclones-function-attribute)

Roughly thinking of this pattern:

https://godbolt.org/z/cvYocWqEW

Notably the `#include <altivec.h>` required as  -maltivec compile flag and my attempts around the below failed.
```
#pragma GCC push_options
#pragma GCC optimize("maltivec") 
```

As a compile option on the file would involve in pre-altivec hardware this might need to be put to a new file. (and adjust compile properties like the crc32 does).

Note: altivec earlier than vsx (in POWER 8, technically as option, but I've never seen it removed).

> I wouldn't probably bother with `2`,

Might be unavoidable to achieve OS portability to FreeBSD.

FreeBSD Power support is POWER4+ (https://www.freebsd.org/releases/14.0R/hardware/)

>  what about adding cmake option and just disable SIMD explicitly (or even automatically) on those older platforms?

Doesn't work for distros.

Distros produce a binary package for all their supported platforms so the code is expected to run on the minimum version. While compile time optimizations beyond this are allowed they are expected to be gated by some runtime check (which the target_clones should do via a IFUNC, which is linux only. This might harm isn't portable to FreeBSD POWER or others hence the option 2 in mysys/crc32/crc32c.cc).

svoj: @grooverdan 

> https://godbolt.org/z/cvYocWqEW

Hmm... there's `#ifndef _ARCH_PWR8` in your example. Did `target_clones` define it?

> Notably the #include <altivec.h> required as -maltivec compile flag and my attempts around the below failed.

I guess it should be `target_clones("altivec,default")` so that it emits `-maltivec`?

Anyway, do you happen to know if we can have access to this hardware (both P6- and p6+) to reproduce what distros are doing and to verify correctness of this boilerplate?

grooverdan: > `_ARCH_PWR8`

its defined by the ISA (https://openpowerfoundation.org/specifications/isa/).

> I guess it should be target_clones("altivec,default") so that it emits -maltivec?

I'd assume so. I used power8 as there where POWER8 instructions in the provided code, which is why I bypassed altivec to the later POWER8+VSX.

> Anyway, do you happen to know if we can have access to this hardware (both P6- and p6+) to reproduce what distros are doing and to verify correctness of this boilerplate?

Distro's fake the build process in a chroot on probably at least POWER9+ hardware.

We don't have preP8 hardware. A deb-autobake on a ppc64 might be sufficient or ping @ottok for a salsa build when ready.

sql/vector_mhnsw.cc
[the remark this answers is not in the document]
mikejuliet13: In the updated commit, I have removed `sum` entirely and directly return the accumulated result as suggested. This ensures the logic remains clean and any overhead is negligible due to this

sql/vector_mhnsw.cc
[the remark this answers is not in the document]
mikejuliet13: Yes, `vec_mule()` and `vec_mulo()` inherently perform widening multiplication, the explicit conversion from `int16_t` to `int32_t` was redundant. I have removed that part in the updated commit

sql/bloom_filters.h
mikejuliet13: I’ve updated `bloom_filters.h` to ensure that the `POWER__IMPLEMENTATION` is only enabled when both `powerpc64` and `VSX` macros are defined. This ensures compatibility with older POWER architectures, which lack VSX support, by falling back to the `DEFAULT_IMPLEMENTATION`

vuvova:
@svoj, @grooverdan — are you ok with the code now?

svoj (approved):
@vuvova yes, now it looks good. I'll let @grooverdan to judge compatibility code.

We just need to squash it before pushing and we probably need either CLA or BSD-new acceptance. Not sure if accept SoB tag?

@mikejuliet13 thanks for your cooperation!

mikejuliet13:
Hi @grooverdan, I’ve addressed all the feedback on this PR and have made the necessary updates.
Could you please take a look when you get a chance?

grooverdan (commented):
few changes. Mostly there.

sql/bloom_filters.h
```diff
@@ -53,6 +53,10 @@ SOFTWARE.
 #define NEON_IMPLEMENTATION
 #endif
 #endif
+#if defined __powerpc64__ && defined __VSX__
+#include <altivec.h>
+#define POWER_IMPLEMENTATION __attribute__ ((target ("cpu=power8,cpu=power9,cpu=power10")))
```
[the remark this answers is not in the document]
grooverdan: last thought, is `POWER_IMPLEMENTATION` needed if its under a `VSX` define?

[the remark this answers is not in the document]
mikejuliet13: I have removed `__attribute__ ((target ("cpu=power8,cpu=power9,cpu=power10")))` to avoid issues with older GCC versions that don’t support it also ensuring compatibility without redundant checks

I've also removed the POWER_IMPLEMENTATION macro before all the functions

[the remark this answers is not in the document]
mikejuliet13: Hi @grooverdan, I’ve addressed the feedback on this PR and have made the necessary updates.
Could you please take a look when you get a chance?

grooverdan (approved):
Thanks @mikejuliet13 !

svoj:
@vuvova it is now approved, feel free to take it over.

svoj:
@mikejuliet13 we should form a single commit, without merge commits and those WiP changes. And also write proper commit message. I can do it if you like, or do you want to do it yourself? git commit message requirements are listed here: https://github.com/MariaDB/server/blob/main/CODING_STANDARDS.md

mikejuliet13:
> @mikejuliet13 we should form a single commit, without merge commits and those WiP changes. And also write proper commit message. I can do it if you like, or do you want to do it yourself? git commit message requirements are listed here: https://github.com/MariaDB/server/blob/main/CODING_STANDARDS.md

Hey @svoj,
Could you please squash everything into a single commit

svoj:
@mikejuliet13 done. I also copied some of your sentences to commit message. Hope you're fine with updated commit.

vuvova:
I've cherry-picked it into my bb-11.8-serg branch, will push it together with other vector-related fixes

svoj:
@vuvova PR retargetted for 11.8 as discussed. It'd be nice to have it merged rather than closed.

PR #3854, opened 2025-02-25
Rewrite the 'Types' chapter in coding standards.

vaintroub:

## Description
Problem - current coding standards explicitly discourages use of int, char, short etc, and recommends fixed types instead.

This is overly pedantic, and currently impractical. The only real problem we have with types concerning portability, is inappropriate use of `long`, is easy to overlook, if it is thrown together with theoretical problems which had not been problems for the last 3 decades.

Thus,un-deprecate the types that are portable, for all practical purpose i.e `int`, `short`, `long long` and their unsigned variations. Warn that `char` might be unsigned, though, all compilers have appropriate flags to make plain `char` unsigned.

**The main thing however is, use strong wording to deprecate `long`, `unsigned long` and `ulong`, those are the types that create real portability problems.**

## Release Notes
Nothing in release notes, it is internal documentation

## How can this PR be tested?
No need to test, it is documentation change.

## PR quality check


CODING_STANDARDS.md
```diff
@@ -360,21 +360,20 @@ For example:
...
-* 32-bit signed / unsigned int -> `int32` / `uint32`
-* 64-bit signed / unsigned int -> `int64` / `uint64`
-* Integer file descriptor -> `File`
-* Integer socket descriptor -> `my_socket`
+Instead of using `long`, use `size_t` and `ptrdiff_t` where appropriate,
+buffer sizes for example. For integer socket descriptor use `my_socket`.
 
-`size_t` and `ptrdiff_t` are used in the source where appropriate, buffer sizes for example.
-It should be noted that these are implementation dependent but are useful when used in the correct context.
+You may use types with fixed length, int32_t and similar, too. Yet, on all platforms we currently support,
```
ParadoxV5, line 372: ~~One day someone might ask if we support embedding [in an Arduino](https://docs.arduino.cc/language-reference/en/variables/data-types/int/).~~

ParadoxV5 (changes requested):
All variable-size types (`int` and `long long`s too) should be discouraged for consistency and future-proofing.
Of course, **especially `long`s**.

It’s that the status quo has used them (namely `long`s) all over the place.
It would be good to comment on this to let new contributors know what was goïng on.

Another contributor to the status quo is that these (*de jure*) variable-sized had more convenient `printf` formats; e.g., `llu` instead of `PRIu64` (`llu` on Windows but `lu` on Linux). See also [MDEV-35451 § comment](https://jira.mariadb.org/browse/MDEV-35451?focusedCommentId=294885&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-294885))

ParadoxV5:
https://jira.mariadb.org/browse/MDEV-36142?jql=labels%20%3D%20ulong

Re. this status quo, maybe we should start migrating our old 🍝 away from those `long`s, e.g., [MDEV-35707](https://jira.mariadb.org/browse/MDEV-35707).

Tip: I found out just yesterday that you can add `[skip ci]` or `[ci skip]` to your commit title 😀

CODING_STANDARDS.md
vaintroub: Thus , "we currently support". If the need arises, we'll port and cope with "long long" being 2 bytes. *If* the need arieses

vaintroub: Also, what's special in an Ardunio? https://learn.sparkfun.com/tutorials/data-types-in-arduino/defining-data-types.
it does not have 64bit shorts, and does not have 16bit ints.

vaintroub:
> All variable-size types (`int` and `long long`s too) should be discouraged for consistency and future-proofing. Of course, **especially `long`s**.
> 
> It’s that the status quo has used them (namely `long`s) all over the place. It would be good to comment on this to let new contributors know what was goïng on.
> 
> Another contributor to the status quo is that these (_de jure_) variable-sized had more convenient `printf` formats; e.g., `llu` instead of `PRIu64` (`llu` on Windows but `lu` on Linux). See also [MDEV-35451 § comment](https://jira.mariadb.org/browse/MDEV-35451?focusedCommentId=294885&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-294885))

I am definitely **against** condemning *all* variable size types. Because, first , your OS does not use them consistently. What type is errno? what do all Linux syscalls return? As for the future proofness, the last 30 years did not bring any change in the things I outlined in the comment. char, int, short, long long are not variable width.  If you can bring your Cray to life, and port MariaDB, Or Win16, you could argue then. Otherwise, point me to any OS we ever supported that did not have that. I do not want people to sweat about "I declared variant int, something bad will happen". They should do it, and use "%d" whenever they wish to.

CODING_STANDARDS.md
ParadoxV5: As linked in my ~~scratched out~~ comment, it does have 16bit ints.

> **int** (16 bit) - signed number from -32768 to 32767.

ParadoxV5:
> intptr_t and uintptr_t are also good ones, but do not have a nice format specifiers, iirc.

Yeah, `PRI*PTR` from `inttypes.h`

There’s also (`u`)`intmax_t` which sounds useful on paper, but I’ve never seen it used anywhere, let alone (`u`)`int_least`/`fast*_t`.

CODING_STANDARDS.md
```diff
@@ -360,21 +360,20 @@ For example:
...
 
-### Types
+### Integer types
 
-In general the usage of types such as `char`, `int` and `long` should be discouraged but there are shortened versions of the unsigned variants available for these in `my_global.h`.
-They can be different sizes across platforms and `char` can be either unsigned or signed depending on platform, and therefore are not portable.
-Instead these should be used as appropriate:
+The usage of types `long`  and `unsigned long` (and its `ulong` alias)
+in new code is *strongly* discouraged. Its use brings no advantages,
+only portability problems between Windows and Unixes.
```
dr-m, line 367: LP64 and LLP64 are the relevant terms that need to be mentioned here. Some UNIX systems are or were actually LP64 just like the 64-bit Microsoft Windows. I remember one example: Digital Tru64 UNIX.

CODING_STANDARDS.md
```diff
@@ -360,21 +360,20 @@ For example:
...
+You may use types with fixed length, int32_t and similar, too. Yet, on all platforms we currently support,
+* `char` is 8 bit
+* `short` is 16 bit
+* `int` is 32bit
+* `long long` is 64bit
 
-Further types can be found in the `include/` directory files.
-There are also general utility functions in `mysys`.
+and the above is not likely to change for the decades to come. Those types are safe to use. When using `char`
+though, be aware that its signdness can depend on compiler flags, so do not assume it can take negative values.
```
grooverdan, line 379: typo `signed-ness`. "compiler flags and CPU architecture specifications". Otherwise looks good !

grooverdan (approved):

CODING_STANDARDS.md
vaintroub: Are you sure memory serves you right on Digital Unix (aka DEC OSF/1, aka Tru64). [Another expert](https://chatgpt.com/share/67c56bc9-a004-800d-bda0-690f63ceeca7) disagrees, but also  there is [archived documentation ](https://www.osarchive.org/os/osf1-tru64/Documentation/5.1B/Programmer%27s%20Guide.pdf) , which hints to LP64 in multiple places.

dr-m: Come to think of it, I may have mistaken this with some other 64-bit UNIX. Actually, I meant to type LLP64 (`long long` and pointers are 64-bit), so my statement about OSF/1 being LP64 should be correct, but the statement about MSVC being LP64 (`long` and pointers are 64-bit) would be incorrect. I don’t remember which of the contemporary proprietary UNIX systems were 64-bit around that time (about 25 to 30 years ago). I had access to several HP-UX systems, but I can’t remember which (if any) of the HP PA-RISC systems were 64-bit. Similar for SGI IRIX running on MIPS.

vaintroub: My memory, which only goes back 29 years ago, but also [Stackoverflow] (https://stackoverflow.com/a/3325747/547065) and [ChatGPT|(https://chatgpt.com/share/67c56bc9-a004-800d-bda0-690f63ceeca7) don't remember Unix or Unix-like LLP64 , but we all remember Windows being LLP64. It not just a compiler property of MSVC, but also native Clang and native GCC on Windows respect it (or they won't be able to use unmodified OS headers). But in general, I think it is useful to LP/LLP64, but at most 4-5 of all developers would even know the difference, but others would only understand "Windows",  thus the guidelines should use that, to be understood.

CODING_STANDARDS.md
vaintroub: Fixed

CODING_STANDARDS.md
vaintroub: We won't port on Ardunio 16 bit integers. Those heroes who succeed, can then fix that paragraph - it will the the smallest of their problems. But is much more likely there won't even be another 32bit OS mariadb will run on.

PR #3856, opened 2025-02-25
MDEV-36122 Assertion ctx0->old_table->get_ref_count() == 1

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36122*
## Description
`trx_purge_close_tables()`: Before releasing any metadata locks (MDL), release all table references, in case an `ALTER TABLE…ALGORITHM=COPY` operation has confused our logic.

`trx_purge_table_acquire()`, `trx_purge_table_open()`: Do not acquire any table reference before successfully acquiring a necessary metadata lock. In this way, if purge is waiting for MDL, a concurrent `ha_innobase::commit_inplace_alter_table(commit=true)` that is holding a conflicting `MDL_EXCLUSIVE` will only observe its own reference on the table that it may need to replace.

`dict_table_open_on_id()`: Simplify the logic.

`dict_stats`: A helper for acquiring MDL and opening the tables `mysql.innodb_table_stats` and `mysql.innodb_index_stats`.

`innodb_ft_aux_table_validate()`: Contiguously hold `dict_sys.latch` while accessing the table that we open with `dict_table_open_on_name()`.

`lock_table_children()`: Do not hold a table reference while invoking `dict_acquire_mdl_shared<false>()`, which may temporarily release and reacquire the shared `dict_sys.latch` that we are holding.

`prepare_inplace_alter_table_dict()`: If an unexpected reference to the table exists, wait for the purge subsystem to release its table handle, similar to how we would do in case `FULLTEXT INDEX` existed. This function is supposed to be protected by `MDL_EXCLUSIVE` on the table name. If purge is going to access the table again later during is `ALTER TABLE` operation, it will have access to an MDL compatible name for it and therefore should conflict with any `MDL_EXCLUSIVE` that would cover `ha_innobase::commit_inplace_alter_table(commit=true)`.

With these changes, no caller of `dict_acquire_mdl_shared<false>` should be holding a table reference.

All remaining calls to `dict_table_open_on_name(dict_locked=false)` except the one in `fts_lock_table()` and possibly in the DDL recovery predicate `innodb_check_version()` should be protected by MDL, but there currently is no assertion that would enforce this.
## Release Notes
Some race conditions between InnoDB internal table access and DDL were fixed.
## How can this PR be tested?
Stress test by the RQG grammar that reproduced the assertion failures.

I would also like to fix a similar problem in `dict_acquire_mdl_shared<false>` and `dict_table_open_on_id()`.
## PR quality check

### DIFF
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -694,35 +694,37 @@ dict_acquire_mdl_shared(dict_table_t *table,
                         MDL_context *mdl_context, MDL_ticket **mdl,
                         dict_table_op_t table_op)
 {
-  table_id_t table_id= table->id;
   char db_buf[NAME_LEN + 1], db_buf1[NAME_LEN + 1];
   char tbl_buf[NAME_LEN + 1], tbl_buf1[NAME_LEN + 1];
   size_t db_len, tbl_len;
-  bool unaccessible= false;
 
   if (!table->parse_name<!trylock>(db_buf, tbl_buf, &db_len, &tbl_len))
     /* The name of an intermediate table starts with #sql */
     return table;
 
 retry:
-  if (!unaccessible && (!table->is_readable() || table->corrupted))
+  ut_ad(!trylock == dict_sys.frozen());
+  ut_ad(trylock || table->get_ref_count());
+
+  if (!table->is_readable() || table->corrupted)
   {
+    if (!trylock)
+      table->release();
     if (*mdl)
     {
       mdl_context->release_lock(*mdl);
       *mdl= nullptr;
     }
-    unaccessible= true;
+    return nullptr;
   }
 
+  const table_id_t table_id{table->id};
+
   if (!trylock)
+  {
     table->release();
-
-  if (unaccessible)
-    return nullptr;
-
-  if (!trylock)
     dict_sys.unfreeze();
+  }
 
   {
     MDL_request request;
@@ -748,11 +750,37 @@ retry:
     }
   }
 
+  size_t db1_len, tbl1_len;
+lookup:
   dict_sys.freeze(SRW_LOCK_CALL);
   table= dict_sys.find_table(table_id);
   if (table)
+  {
+    if (!table->is_accessible())
+    {
+      table= nullptr;
+    unlock_and_return_without_mdl:
+      if (trylock)
+        dict_sys.unfreeze();
+    return_without_mdl:
+      if (*mdl)
+      {
+        mdl_context->release_lock(*mdl);
+        *mdl= nullptr;
+      }
+      return table;
+    }
+
     table->acquire();
-  if (!table && table_op != DICT_TABLE_OP_OPEN_ONLY_IF_CACHED)
+
+    if (!table->parse_name<true>(db_buf1, tbl_buf1, &db1_len, &tbl1_len))
+    {
+      /* The table was renamed to #sql prefix.
+      Release MDL (if any) for the old name and return. */
+      goto unlock_and_return_without_mdl;
+    }
+  }
+  else if (table_op != DICT_TABLE_OP_OPEN_ONLY_IF_CACHED)
   {
     dict_sys.unfreeze();
     dict_sys.lock(SRW_LOCK_CALL);
@@ -760,31 +788,15 @@ retry:
                                  table_op == DICT_TABLE_OP_LOAD_TABLESPACE
                                  ? DICT_ERR_IGNORE_RECOVER_LOCK
                                  : DICT_ERR_IGNORE_FK_NOKEY);
-    if (table)
-      table->acquire();
     dict_sys.unlock();
-    dict_sys.freeze(SRW_LOCK_CALL);
-  }
-
-  if (!table || !table->is_accessible())
-  {
-return_without_mdl:
-    if (trylock)
-      dict_sys.unfreeze();
-    if (*mdl)
-    {
-      mdl_context->release_lock(*mdl);
-      *mdl= nullptr;
-    }
-    return nullptr;
-  }
-
-  size_t db1_len, tbl1_len;
-
-  if (!table->parse_name<true>(db_buf1, tbl_buf1, &db1_len, &tbl1_len))
-  {
-    /* The table was renamed to #sql prefix.
-    Release MDL (if any) for the old name and return. */
+    /* At this point, the freshly loaded table may already have been evicted.
+    We must look it up again while holding a shared dict_sys.latch.  We keep
+    trying this until the table is found in the cache or it cannot be found
+    in the dictionary (because the table has been dropped or rebuilt). */
+    if (table)
+      goto lookup;
+    if (!trylock)
+      dict_sys.freeze(SRW_LOCK_CALL);
     goto return_without_mdl;
   }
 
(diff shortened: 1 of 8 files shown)


storage/innobase/trx/trx0purge.cc
```diff
@@ -1134,21 +1142,17 @@ static dict_table_t *trx_purge_table_acquire(dict_table_t *table,
...
-  else
+  if (!table)
   {
     dict_sys.unfreeze();
     dict_sys.lock(SRW_LOCK_CALL);
     table= dict_load_table_on_id(table_id, DICT_ERR_IGNORE_FK_NOKEY);
-    if (table)
-      table->acquire();
     dict_sys.unlock();
     if (!table)
```
dr-m, line 1157: After `dict_sys.unlock()` has released the `dict_sys.latch`, another thread may evict the table definition. I forgot that neither the `lock_release()` during a transaction commit (if `innodb_evict_tables_on_commit_debug=ON`) nor `dict_sys_t::evict_table_LRU()` are protected by any MDL. They would however be protected by `table->get_ref_count() == 0`. A first solution could be that we must acquire a table reference before invoking `dict_sys.unlock()` and release it after invoking `dict_sys.freeze(SRW_LOCK_CALL)`. But, we are not yet holding an MDL on the table name. We must first copy the name, then release `dict_sys.latch`, then acquire the MDL, look up or load the table by that ID again, and repeat if the table name has changed meanwhile.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -17602,11 +17502,16 @@ static int innodb_ft_aux_table_validate(THD *thd, st_mysql_sys_var*,
 	int len = sizeof buf;
 
 	if (const char* table_name = value->val_str(value, buf, &len)) {
+		/* Because we are not acquiring MDL on the table name,
+		we must contiguously hold dict_sys.latch while we are
+		examining the table, to protect us against concurrent DDL. */
+		dict_sys.lock();
```
dr-m, line 17508: This needs to be `dict_sys.lock(SRW_LOCK_CALL)`, or otherwise this will only work with `cmake -DPLUGIN_PERFSCHEMA=NO`, which I was using locally.

storage/innobase/dict/dict0dict.cc
```diff
@@ -873,16 +880,29 @@ dict_table_t *dict_table_open_on_id(table_id_t table_id, bool dict_locked,
...
   {
-    table->acquire();
-    if (thd && !dict_locked)
-      table= dict_acquire_mdl_shared<false>(table, thd, mdl, table_op);
+    if (!dict_locked)
+    {
+      if (thd)
+      {
+        table= dict_acquire_mdl_shared<false>(table, thd, mdl, table_op);
+        if (table)
```
mariadb-DebarunBanerjee, line 896: This is just a note on the existing code. We are acquiring MDL shared while the dictionary is frozen. Since MDL is generally taken in server, it might be possible for some DDL asking for dict sys lock while holding exclusive MDL lock for the table - and a deadlock. 
I think this doesn't happen in common DML path as we would have already taken MDL shared in server by this time.

storage/innobase/trx/trx0purge.cc
```diff
@@ -1051,16 +1051,25 @@ inline trx_purge_rec_t purge_sys_t::fetch_next_rec()
...
+
+  for (auto &t : node->tables)
+  {
+    dict_table_t *table= t.second.first;
+    if (table != nullptr && table != reinterpret_cast<dict_table_t*>(-1))
     {
-      dict_table_close(t.second.first, false, thd, t.second.second);
       t.second.first= reinterpret_cast<dict_table_t*>(-1);
+      if (mdl_context != nullptr && t.second.second != nullptr)
+        mdl_context->release_lock(t.second.second);
```
mariadb-DebarunBanerjee, line 1072: The tables are unrelated and it should have been possible to avoid 2 loops here ideally by releasing the MDL lock on a table immediately after dict_table_close(). 

I would not insist on this change though as it might be considered safer to release all MDL locks in the batch at end reducing the possibility of similar issues present... I would be interested to know your perspective here.

storage/innobase/trx/trx0purge.cc
```diff
@@ -1122,37 +1130,38 @@ static dict_table_t *trx_purge_table_acquire(dict_table_t *table,
...
   {
+    dict_sys.freeze(SRW_LOCK_CALL);
+    table= dict_sys.find_table(table_id);
+    if (table)
+      break;
     dict_sys.unfreeze();
     dict_sys.lock(SRW_LOCK_CALL);
     table= dict_load_table_on_id(table_id, DICT_ERR_IGNORE_FK_NOKEY);
-    if (table)
-      table->acquire();
```
mariadb-DebarunBanerjee, line 1151: We could have avoided the unconditional for(;;) loop here by having the t `table->acquire();` in this path and releasing (`table->release()`) it later after `dict_sys.freeze(SRW_LOCK_CALL); `.

mariadb-DebarunBanerjee (approved):
The changes look fine to me. Added some notes.

storage/innobase/trx/trx0purge.cc
dr-m: The purge of committed transaction history was last refactored in #3448. The purge coordinator task will acquire shared MDL on all tables that are part of the current purge batch. Purge worker tasks in that batch will access these tables without acquiring any additional MDL. So, it really is part of the design to hold several MDL at a time, as part of the current purge batch.

storage/innobase/dict/dict0dict.cc
dr-m: We’re invoking the template with `bool trylock=false`. Upon entry, a debug assertion will check that we are holding `dict_sys.latch`. We are actually releasing the table reference and the `dict_sys.latch` before waiting for the MDL:
```c++
  if (!trylock)
  {
    table->release();
    dict_sys.unfreeze();
  }

  {
    MDL_request request;
    MDL_REQUEST_INIT(&request,MDL_key::TABLE, db_buf, tbl_buf, MDL_SHARED,
                     MDL_EXPLICIT);
    if (trylock
        ? mdl_context->try_acquire_lock(&request)
        : mdl_context->acquire_lock(&request,
                                    /* FIXME: use compatible type, and maybe
                                    remove this parameter altogether! */
                                    static_cast<double>(global_system_variables
                                                        .lock_wait_timeout)))
```

storage/innobase/trx/trx0purge.cc
dr-m: The idea of the first commit is to avoid any situation where we’d hold a table reference without holding an MDL on the table name, while not holding `dict_sys.latch`. Yes, it is rather clumsy to first look up the table and then try to load it and look it up again, but there is no way to downgrade `dict_sys.latch` from exclusive to shared.

storage/innobase/trx/trx0purge.cc
mariadb-DebarunBanerjee: > avoid any situation where we’d hold a table reference without holding an MDL on the table name,

yes, I understand. That is still achievable by acquiring the table reference first and then releasing it. The idea is to get rid of the for loop. Something like this pseudo code. This is just to convey the idea and I don't mean to cut-paste the code as it is.

```
{
  table= dict_sys.find_table(table_id);
  if (table)
  {
  ready_for_mdl:
    table= trx_purge_table_acquire(table, mdl_context, mdl);
    dict_sys.unfreeze();
    return table;
  }

  dict_sys.unfreeze();
  dict_sys.lock(SRW_LOCK_CALL);
  table= dict_load_table_on_id(table_id, DICT_ERR_IGNORE_FK_NOKEY);
  if (table)
  {
    table->acquire();
    dict_sys.unlock();
    table->release();
    goto ready_for_mdl;
  }
  dict_sys.unlock();
  return nullptr;
}
```

PR #3857, opened 2025-02-26
MDEV-36149 UBSAN in X is outside the range of representable values of…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-36149*

## Description
Currently it is allowed to set innodb_io_capacity to very large value
up to unsigned 8 byte maximum value 18446744073709551615. While
calculating the number of pages to flush, we could sometime go beyond
innodb_io_capacity. Specifically, MDEV-24369 has introduced a logic
for aggressive flushing when dirty page percentage in buffer pool
exceeds innodb_max_dirty_pages_pct. So, when innodb_io_capacity is
set to very large value and dirty page percentage exceeds the
threshold, there is a multiplication overflow in Innodb page cleaner.

Fix: We should prevent setting io_capacity to unrealistic values and
define a practical limit to it. The patch introduces limits for
innodb_io_capacity_max and innodb_io_capacity to the maximum of 4 byte
unsigned integer i.e. 4294967295 (2^32-1). For 16k page size this limit
translates to 64 TiB/sec write IO speed which looks sufficient.

## Release Notes
https://mariadb.com/kb/en/innodb-system-variables/#innodb_io_capacity
Range: 100 to 4294967295 (2^32-1) 
https://mariadb.com/kb/en/innodb-system-variables/#innodb_io_capacity_max
Range: 100 to 4294967295 (2^32-1)

## How can this PR be tested?
It requires Sanitized build -DWITH_ASAN=ON -DWITH_UBSAN=ON to repeat the issue with the scenario described in MDEV.

## PR quality check


storage/innobase/handler/ha_innodb.cc
```diff
@@ -18986,7 +18987,7 @@ static MYSQL_SYSVAR_ENUM(instant_alter_column_allowed,
 static MYSQL_SYSVAR_ULONG(io_capacity, srv_io_capacity,
   PLUGIN_VAR_RQCMDARG,
   "Number of IOPs the server can do. Tunes the background IO rate",
-  NULL, innodb_io_capacity_update, 200, 100, ~0UL, 0);
+  NULL, innodb_io_capacity_update, 200, 100, SRV_MAX_IO_CAPACITY_LIMIT, 0);
```
dr-m, line 18990: You forgot to adjust `mysql-test/suite/sys_vars/r/sysvars_innodb.result` and possibly `mysql-test/suite/sys_vars/r/sysvars_innodb,32bit.rdiff` for this. I thought that we could have changed the type of the variable to 32-bit so that some hunks of `mysql-test/suite/sys_vars/r/sysvars_innodb,32bit.rdiff` would disappear.

I can imagine that we will avoid any overflow in `page_cleaner_flush_pages_recommendation()` on 64-bit systems, but it is less clear to me what would happen on a 32-bit system.

What is the maximum value of `dirty_pct` passed to `page_cleaner_flush_pages_recommendation()`? From your description I understand that it really is intended to be above 1. In [MDEV-36149](https://jira.mariadb.org/browse/MDEV-36149) the largest mentioned value is `dirty_pct=1.016865`. It would seem to me that the calculation
```c++
		n_pages = ulint(dirty_pct * double(srv_io_capacity));
```
can easily overflow on a 32-bit system: multiplying `UINT32_MAX` with something larger than 1 would not fit in `ulint` a.k.a. `size_t`, which only allows values up to `UINT32_MAX` on 32-bit platforms.

On 32-bit targets we seem to need a saturating conversion. I think that on x86 and x86-64 we’d want to be able to use [cvttsd2si](https://www.felixcloutier.com/x86/cvttsd2si), which would convert `double` to a signed 32-bit (or 64-bit) integer. x86-64 always includes SSE and SSE2. To be able to use that instruction, it seems that we should declare these parameters as `int` with a maximum value of `INT32_MAX` (half of what the current change would do). ARMv8.3 includes [FJCVTZS](https://developer.arm.com/documentation/dui0801/h/A64-Floating-point-Instructions/FJCVTZS), which seems to have similar constraints as the x86 `cvttsd2si`.

To my understanding, C and C++ declare saturating conversions [undefined behaviour](https://discourse.llvm.org/t/converting-float-to-int-with-fjcvtzs/57792). A comment in [this discussion](https://www.reddit.com/r/cpp/comments/rsldwl/converting_float_to_int_can_be_undefined_behavior/) links to https://godbolt.org/z/Mv5737M7b which is generating branch instructions for the bound checks on the compilers that I tested. So, it might be simplest to just insert explicit overflow checks on 32-bit targets.

mariadb-DebarunBanerjee: 1. Right, now updated mysql-test/suite/sys_vars/r/sysvars_innodb.result. I kept the type as it is to keep the impact limited and in case we need to set it higher.
2.  I updated sysvars_innodb,32bit.rdiff this way. Please see if correct. Is there any CI running 32 BIT build ?

```
1. Generate 32 BIT result file
        cd mysql-test/suite/sys_vars/r/
        patch sysvars_innodb.result sysvars_innodb,32bit.rdiff

2. Save 32 BIT version
        cp sysvars_innodb.result sysvars_innodb.result.32

3. Revert the update
        git checkout sysvars_innodb.result

4. Generate new 64 BIT file
        cd bld_dbg/mysql-test'
        ./mtr --mem --record sys_vars.sysvars_innodb

5. Save new diff
        diff -u  -L "sysvars_innodb.result" sysvars_innodb.result -L"sysvars_innodb.result,32bit" sysvars_innodb.result.32 > sysvars_innodb,32bit.rdiff

6. Generate 32 BIT result file with new result file and new diff file
        cd mysql-test/suite/sys_vars/r/
        patch sysvars_innodb.result sysvars_innodb,32bit.rdiff

7. Compare old and new 32 BIT version files and confirm they are same   // No 32 BIT specific changes made
        meld sysvars_innodb.result.32 sysvars_innodb.result
```
3. Yes, 32 BIT system could still encounter the over-flow. I changed `page_cleaner_flush_pages_recommendation` to make sure we don't deal with very large IO capacity values beyond the number of dirty pages in the system. There is not much point setting the flush target to more than actual number of dirty pages in the system. Also added a change for the case when dirty % is set very low < 1%. The logic for deciding the flush target remains unchanged otherwise. I think the modified code should be fine for both 32 and  64 BIT. 

Please have a look.

dr-m: I usually build a 32-bit executable locally with a script that I wrote a long time ago:
```bash
#!/bin/bash
exec cmake -DWITH_EMBEDDED_SERVER=OFF -DWITH_SAFEMALLOC=OFF -DWITH_UNIT_TESTS=OFF -DCMAKE_BUILD_TYPE=Debug -DCMAKE_{C,CXX}_FLAGS='-m32 -march=i686' -DWITH_WSREP=OFF -DPLUGIN_{ARCHIVE,TOKUDB,MROONGA,OQGRAPH,ROCKSDB,CONNECT}=NO -DWITH_SSL=bundled -DWITH_ZLIB=system -DCURSES_CURSES_H_PATH=/usr/include/curses.h -DCURSES_CURSES_LIBRARY=/usr/lib32/libcurses.a -DCURSES_FORM_LIBRARY=/usr/lib32/libform.so -DGNUTLS_LIBRARY=/usr/lib/i386-linux-gnu/libgnutls.so -DCMAKE_LIBRARY_ARCHITECTURE=lib32 "$@"
# sudo apt install g++-8-multilib lib32ncurses5-dev
```
I didn’t bother to update the comment about additional packages after upgrading from GCC 8 more than 5 years ago.

Debian still supports multiarch between amd64 and i386; many other GNU/Linux distributions have dropped i386 (or actually it’s more like i686 for us, to have 64-bit atomic operations).

Typically, I’d delete the `.rdiff` file and run the test, and then paste the result difference to the `.rdiff` file. If you do it based on some CI output at the bottom of the [grid view of this pull request](https://buildbot.mariadb.org/#/grid?branch=refs%2Fpull%2F3857%2Fmerge), it may take several attempts and the line numbers in the file could end up being incorrect, which might create trouble for some versions of `patch` that might run outside our CI.

dr-m (approved):
Thank you. I didn’t see any messages from `patch` about line number mismatch in https://buildbot.mariadb.org/#/builders/862/builds/37/steps/6/logs/stdio, so I assume that the `.rdiff` file has been updated correctly.

PR #3858, opened 2025-02-26
MDEV-36155: MSAN use-of-uninitialized-value innodb.log_file_size_online

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36155*

## Description

Writing the redo log resized will write uninitialized data. There is a MEM_MAKE_DEFINED construct in the code to bless this however it was correct on the initial length, but not the changed length.

The MEM_MAKE_DEFINED is moved earlier in the code where the length contains the correct value.

## Release Notes

If anything: innodb.log_file_size_online test result made reliable MSAN compiled code.

## How can this PR be tested?

```
buildbot@1209234bb2fb:/build$ mysql-test/mtr  innodb.log_file_size_online 
Logging: /source/mysql-test/mariadb-test-run.pl  innodb.log_file_size_online
VS config: 
vardir: /build/mysql-test/var
Checking leftover processes...
Removing old var directory...
Creating var directory '/build/mysql-test/var'...
MariaDB Version 12.0.0-MariaDB
 - SSL connections supported
 - binaries built with wsrep patch
Collecting tests...
Installing system database...

innodb.log_file_size_online 'slow'       [ pass ]   9190
innodb.log_file_size_online 'encrypted'  [ pass ]   8334
The servers were restarted 1 times
Spent 17.524 of 22 seconds executing testcases

Completed: All 2 tests were successful.

buildbot@1209234bb2fb:/build$ clang --version
Debian clang version 20.1.0 (++20250225053057+098492a228f7-1~exp1~20250225053219.63)
```

## PR quality check

### DIFF
--- a/storage/innobase/log/log0log.cc
+++ b/storage/innobase/log/log0log.cc
@@ -1044,12 +1044,13 @@ lsn_t log_t::write_buf() noexcept
         the current LSN are generated. */
         MEM_MAKE_DEFINED(buf + length, (write_size_1 + 1) - new_buf_free);
         buf[length]= 0; /* allow recovery to catch EOF faster */
+        if (UNIV_LIKELY_NULL(re_write_buf))
+          MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) -
+                           new_buf_free);
         length&= ~write_size_1;
         memcpy_aligned<16>(flush_buf, buf + length, (new_buf_free + 15) & ~15);
         if (UNIV_LIKELY_NULL(re_write_buf))
         {
-          MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) -
-                           new_buf_free);
           memcpy_aligned<16>(resize_flush_buf, re_write_buf + length,
                              (new_buf_free + 15) & ~15);
           re_write_buf[length + new_buf_free]= 0;


storage/innobase/log/log0log.cc
```diff
@@ -1044,12 +1044,13 @@ lsn_t log_t::write_buf() noexcept
         the current LSN are generated. */
         MEM_MAKE_DEFINED(buf + length, (write_size_1 + 1) - new_buf_free);
         buf[length]= 0; /* allow recovery to catch EOF faster */
+        if (UNIV_LIKELY_NULL(re_write_buf))
+          MEM_MAKE_DEFINED(re_write_buf + length, (write_size_1 + 1) -
+                           new_buf_free);
```
dr-m, line 1049: I was a bit worried that this could cause a compiler warning, but I didn’t see that in any non-debug build log that I checked. So, let’s go for this; it should be the most readable solution.

Side note: `re_write_buf` is `nullptr` if and only if `resizing == NOT_RESIZING`. I’d expect the `if` to be optimized away in any optimized build, that is, starting with `-O1`.

dr-m (approved):

storage/innobase/log/log0log.cc
grooverdan: I'm all in favour of readable. Considering optimization in an infrequent action that includes large time items like filesystem writes seems a little futile.

PR #3861, opened 2025-02-28
fix: Use of Kernel.open or IO.read or similar sinks with a non-constant value

odaysec:
https://github.com/MariaDB/server/blob/5091986ceaca20c650d4469de08201c5b85428e4/storage/mroonga/vendor/groonga/vendor/download_mecab.rb#L45-L49

If `Kernel.open` is given a file name that starts with a `|` character, it will execute the remaining string as a shell command. If a malicious user can control the file name, they can execute arbitrary code. The same vulnerability applies to `IO.read`, `IO.write`, `IO.binread`, `IO.binwrite`, `IO.foreach`, `IO.readlines` and `URI.open`.

## POC
The following shows code that calls `Kernel.open` on a user-supplied file path.
```rb
require "open-uri"

class UsersController < ActionController::Base
  def create
    filename = params[:filename]
    open(filename) # BAD

    web_page = params[:web_page]
    URI.open(web_page) # BAD - calls `Kernel.open` internally
  end
end
```
Instead, `File.open` should be used, as in the following example.
```rb
class UsersController < ActionController::Base
  def create
    filename = params[:filename]
    File.open(filename)

    web_page = params[:web_page]
    Net::HTTP.get(URI.parse(web_page))
  end
end
```
## References
- [Command Injection](https://www.owasp.org/index.php/Command_Injection). [Ruby on Rails Cheat Sheet: Command Injection](https://cheatsheetseries.owasp.org/cheatsheets/Ruby_on_Rails_Cheat_Sheet.html#command-injection)
- [Command Injection in RDoc](https://www.ruby-lang.org/en/news/2021/05/02/os-command-injection-in-rdoc/)
- [CWE-78](https://cwe.mitre.org/data/definitions/78.html)
- [CWE-88](https://cwe.mitre.org/data/definitions/88.html)
- [CWE-73](https://cwe.mitre.org/data/definitions/73.html)

-->

## PR quality check

### DIFF
--- a/storage/mroonga/vendor/groonga/vendor/download_mecab.rb
+++ b/storage/mroonga/vendor/groonga/vendor/download_mecab.rb
@@ -42,7 +42,7 @@ def download(url, base)
 
   tar = "#{base}.tar"
   tar_gz = "#{tar}.gz"
-  open(url, :ssl_verify_mode => ssl_verify_mode) do |remote_tar_gz|
+  URI.open(url, :ssl_verify_mode => ssl_verify_mode) do |remote_tar_gz|
     File.open(tar_gz, "wb") do |local_tar_gz|
       local_tar_gz.print(remote_tar_gz.read)
     end


vuvova:
@kou, I see you've fixed it groonga 6 years ago (https://github.com/groonga/groonga/commit/b544ae2d1007cc1cc08b9082870e284c1fdc3f6f). Shall we update mroonga and groonga in MariaDB to something more recent? :innocent:

kou:
I hope so too... But Groonga/Mroonga still have "sign-compare" warnings... It will break MariaDB CI...

If we can disable "sign-compare" warnings for now, we can start working on Groonga/Mroonga update. Is it acceptable as a temporary workaround?

BTW, we are free to accept or reject this PR because MariaDB doesn't use this code. There is no security concern for the code in MariaDB.

vuvova:
Yes, absolutely, you can disable warnings that you don't want. Every engine can do it for itself. See, for example https://github.com/MariaDB/server/blob/main/storage/connect/CMakeLists.txt#L61-L72 or https://github.com/MariaDB/server/blob/main/storage/columnstore/CMakeLists.txt#L21-L26

vuvova:
@odaysec, it seems your change was in a dead code, which is never executed in MariaDB. That is, there is no need to fix it. We'll upgrade the whole of Mroonga and Groonga though, thanks for bringing this to our attention.

kou:
OK. Can we use https://buildbot.mariadb.org/ for testing again?

It seems that old webhook we used was removed. (We used `http://USER:PASSWORD@buildbot.askmonty.org/buildbot/change_hook/github` in https://github.com/mroonga/mariadb-server/ .)

I tried `https://USER:PASSWORD@buildbot.mariadb.org/buildbot/change_hook/github` but it doesn't exist...

vuvova:
The easiest way would be to create a new PR for updating mroonga. https://buildbot.mariadb.org/ builds all PRs. Then you can keep pushing into the PR branch and it'll keep testing it. When you're ready, say so in a comment and we'll merge it

kou:
Thanks! I'll use the approach.

PR #3863, opened 2025-02-28
MDEV-14432: mysqldump does not preserve case of table names in generated sql

MohanadKh03:
- [x] *The Jira issue number for this PR is: [MDEV-14432](https://jira.mariadb.org/browse/MDEV-14432) "this also affects [MDEV-31472](https://jira.mariadb.org/browse/MDEV-31472) directly"

## Description
```lower_case_table_names``` has three values only 0, 1, 2 ... On windows the default is 1 but when changing that value to 2 (or testing it on MacOS since the default value should also be 2) it should deal with table/db names with their original created table name and compare lowercase only not convert them into lowercase 

```get_lookup_field_values()``` should only convert the names to lowercase when ```lower_case_table_names``` is 1 only not 2 

## Release Notes
Nothing should be changed since this was already mentioned in 
https://mariadb.com/kb/en/identifier-case-sensitivity/ 

## How can this PR be tested?
On windows 64-bit machines use the following test case
```bash 
perl mysql-test-run.pl lowercase_table2.test 
```
In the ```lowercase_table2.test``` and ```lowercase_table2.result``` both have already covered a test case where one could create a table with uppercase and then expect the output to reserve the case when using ```lower_case_table_names``` as 2

## PR quality check

### DIFF
--- a/mysql-test/main/lowercase_table2.result
+++ b/mysql-test/main/lowercase_table2.result
@@ -185,7 +185,7 @@ create table myUC (i int);
 select TABLE_SCHEMA,TABLE_NAME FROM information_schema.TABLES
 where TABLE_SCHEMA ='mysqltest_LC2';
 TABLE_SCHEMA	TABLE_NAME
-mysqltest_lc2	myUC
+mysqltest_LC2	myUC
 use test;
 drop database mysqltest_LC2;
 #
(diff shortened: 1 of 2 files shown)


abarkov (approved):
The change is Ok.

PR #3865, opened 2025-03-04
MDEV-30189 Add remaining replication options as system variables

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-30189 Add and alias remaining mariadbd replication command line options as system variables](https://jira.mariadb.org/browse/MDEV-30189)*

## Description
Promote the last few SQL-inaccessible replication options (command line or `mariadb.cnf`) as these system variables:
```
@@master_info_file (GLOBAL read-only)
@@replicate_same_server_id (GLOBAL read-only)
@@show_slave_auth_info (GLOBAL & SESSION)
```

Due to technical limitations, the latter two options changed from no argument to optional argument.
https://github.com/MariaDB/server/blob/3a81664cb80bfb1598d1a113f24a52f0a627b4f9/sql/sys_vars.inl#L450-L471
https://github.com/MariaDB/server/blob/3a81664cb80bfb1598d1a113f24a52f0a627b4f9/include/my_getopt.h#L51-L58

## Release Notes
`--master-info-file`, `--replicate-same-server-id` and `--show-slave-auth-info` are now available as system variables.
### https://mariadb.com/kb/ pages that need changing.
* https://mariadb.com/kb/en/mariadbd-options/#replication-and-binary-logging-options
* https://mariadb.com/kb/en/replication-and-binary-log-system-variables/

## How can this PR be tested?
Minimal MTR tests are included in this commit.
* variable scope
* for `show_slave_auth_info`, required type and privilege for setting.
  * *I could use a double-check from a security engineer if this is sufficiently secure.*
* for the other two, their read-only property

TODO: suggest a broader plan and link it here

Note that `--master-info-file` and `--show-slave-auth-info` don’t have existing function tests :exclamation:.

## PR quality check


sql/sys_vars.cc
```diff
@@ -5217,6 +5217,13 @@ static Sys_var_uint Sys_repl_report_port(
        "to the slave. If not sure, leave this option unset",
        READ_ONLY GLOBAL_VAR(report_port), CMD_LINE(REQUIRED_ARG),
        VALID_RANGE(0, UINT_MAX), DEFAULT(0), BLOCK_SIZE(1));
+
+static Sys_var_on_access<Sys_var_mybool,
+       PRIV_STMT_SHOW_SLAVE_HOSTS, PRIV_STMT_SHOW_SLAVE_HOSTS>
+Sys_show_slave_auth_info(
+       "show_slave_auth_info",
+       "Show user and password in SHOW SLAVE HOSTS on this master",
+       SESSION_VAR(show_slave_auth_info), CMD_LINE(OPT_ARG), DEFAULT(FALSE));
```
ParadoxV5, line 5226: How valuable is `@@GLOBAL.show_slave_auth_info= ON`?

sql/sys_vars.cc
ParadoxV5, line 1: Is there even a convention where to put new system variables?…

sql/repl_failsafe.cc
```diff
@@ -183,7 +183,7 @@ static my_bool show_slave_hosts_callback(THD *thd, Protocol *protocol)
     protocol->prepare_for_resend();
     protocol->store(si->server_id);
     protocol->store(si->host, strlen(si->host), &my_charset_bin);
-    if (opt_show_slave_auth_info)
+    if (thd->variables.show_slave_auth_info)
```
ParadoxV5, line 186: Is it possible for `@@SESSION.show_slave_auth_info` to change in the middle of a `SHOW REPLICA HOSTS`?

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.test
```diff
@@ -0,0 +1,32 @@
+--source include/not_embedded.inc
+# MDEV-30189: Test minimal correctness of `@@show_slave_auth_info` itself
```
ParadoxV5, line 2: Since `sys_vars.sysvars_server_`(`not`)`embedded` dump the output of `information_schema.system_variables`, my idea is to treat this view as  authoritative and only test that the behavior of “basic” syntaxes matches.
https://github.com/MariaDB/server/blob/3a81664cb80bfb1598d1a113f24a52f0a627b4f9/mysql-test/suite/sys_vars/inc/sysvars_server.inc#L21

The idea puts the correctness of “alternate” syntaxes their own responsibility that they match the behavior of the “basic” syntaxes.
Those include:
```mysql
SHOW GLOBAL/SESSION VARIABLES LIKE name;
SELECT GLOBAL/SESSION_VALUE FROM information_schema.SYSTEM_VARIABLES WHERE VARIABLE_NAME='name';
SELECT VARIABLE_VALUE FROM information_schema.GLOBAL/SESSION_VARIABLES WHERE VARIABLE_NAME='name';
SELECT VARIABLE_VALUE FROM performance_schema.global/session_variables WHERE VARIABLE_NAME='name';

SET GLOBAL/SESSION name = value;
SELECT value INTO @@GLOBAL/SESSION.name;
@@GLOBAL/SESSION.name := value;
SET STATEMENT name = value FOR statement;

# and automatic-scope (i.e., `GLOBAL`/`SESSION`-less) variants
```

Both “‘basic’ syntaxes matches `IS.system_variables`” and “TIMTOWTDI syntaxes matches each other” can be generic tests agnostic of specific variables.
They can test every variable with a `for`-loop or only a random sample.

ParadoxV5 (commented):
Reminders for myself:
* [x] ~~mention `SET @@show_slave_auth_info`’s required privilege in the commit description~~ obsoleted by https://github.com/MariaDB/server/pull/3865#discussion_r1982339946
* [x] ~~check option IDs – that seems to be a thing~~ *nope, not here*
* [ ] update sys.vars results – oops forgot that one

mysql-test/suite/sys_vars/t/master_info_file_basic.test
```diff
@@ -0,0 +1,14 @@
...
+# Expect different from default ('master.info') as configured by `.opt`
+SELECT @@GLOBAL.master_info_file;
+
+--echo # Not SESSION scope
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
+SELECT @@SESSION.master_info_file;
+
+--echo # Read-only
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
+SET @@master_info_file= 'master.info';
```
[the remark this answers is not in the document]
mariadb-SusilBehera: Do we've an alias primary_info_file for this?

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.opt
```diff
@@ -0,0 +1 @@
+--show-slave-auth-info
```
mariadb-SusilBehera, line 1: Do we've an alias show-replica-auth-info for this?

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.test
```diff
@@ -0,0 +1,32 @@
...
+SET @@show_slave_auth_info= TRUE;
+SELECT @@show_slave_auth_info;
+
+--echo # Incorrect types
+--error ER_WRONG_TYPE_FOR_VAR
+SET @@show_slave_auth_info= 0.0;
+--error ER_WRONG_VALUE_FOR_VAR
+SET @@show_slave_auth_info= 2;
+--error ER_WRONG_VALUE_FOR_VAR
+SET @@show_slave_auth_info= 'lorem';
```
mariadb-SusilBehera, line 32: An actual SHOW SLAVE/REPLICA HOSTS result check after a SET show_slave_auth_info= ON would have been better.

mariadb-SusilBehera (commented):
Added my review comments on MTR tests.

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.opt
ParadoxV5: The KB documents (top post § kb/ pages that need changing) did not mention any existing primary–replica aliases, so I did not do that part of the task.
It’d be odd if only these new ones have aliases as a minority.

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.test
ParadoxV5: About this ideal unit testing, @mariadb-SusilBehera do you know how well they hold up currently, or does the status quo require manually testing each of `@@`, @@global.` and `@@session.`?

sql/sys_vars.cc
bnestere: Well, you may have already found this out, but it doesn't use the values of the connecting replication user, but rather the values of the options `report_user` and `report_password` configured on the slave. As to its value, I have never actually seen anyone use it. I did some Googling to see if people use it or "report_user/password", and have turned up nothing. Then I also checked the commit history and also couldn't find out why it was added in the first place. So I see your point.. Though investigating it would be a different issue, and for the overall agenda of MDEV-28672 we still need it visible, despite the fact that likely no one uses it..

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.test
bnestere: `master_info_file` also doesn't have any existing functional tests :facepalm:. I do suppose we should add test cases for both variables in with this patch, to at least prove nothing is breaking.

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.test
bnestere: I'd agree with the premise of the basic test, and testing alternate syntax for the same functionality isn't really related to the addition of a variable itself. Though ensuring a variable is at the proper scope (session vs global) wouldn't fall into that category (which in this patch, you have already accounted for :+1:)

mysql-test/suite/sys_vars/t/show_slave_auth_info_grant.test
```diff
@@ -0,0 +1,9 @@
+--echo #
+--echo # MDEV-30189 Add remaining mariadbd replication command line options as system variables
+--echo #
+
+--let var = show_slave_auth_info
+--let grant = REPLICATION MASTER ADMIN
+--let value = TRUE
+
+--source suite/sys_vars/inc/sysvar_global_and_session_grant.inc
```
bnestere, line 9: I mention in another comment that we don't want to add any behavior to set this variable at run-time, only view it, so we don't need this test. Though I wanted to highlight it is a great find to use `sysvar_global_and_session_grant.inc`, so it would have been a great test otherwise :innocent:

bnestere (changes requested):
Thanks for the patch, @ParadoxV5 , great work! Please see my few notes

sql/repl_failsafe.cc
ParadoxV5: Adding a system varaible does (or at least *should*) not affect how it’s set from an options.
My variable tests uses `.opt` files to assert that the variables default to what’s set from the options.

ParadoxV5: For the perspective of “scope”, though, I vaguely recall finding another variale or two (besides `@@replicate_same_server_id`, if that’s one of them) that make sense to be writeable, possibly even from the code’s perspective…

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.test
ParadoxV5: I’ve written test for the two *options* at #3875.
We can merge that PR into this one, though I prefer merging through `10.11→main`.

mysql-test/suite/sys_vars/t/show_slave_auth_info_basic.test
mariadb-SusilBehera: https://github.com/MariaDB/server/pull/3865#discussion_r1982157828
I think a single unit test which covers all sysvars and checks for sysvar compliances is good enough. Beyond this, the functional tests around the new sysvars should be done in different MTR tests.

ParadoxV5:
There, with this amend I undid `@@show_slave_auth_info`’s `SESSION` refactor and added `@@GLOBAL`/`@@SESSION`/`@@` testing for all.
I also rebased *because I forgot that four-way diff is hard and GitHub doesn’t have this feature* 😅🧒

bnestere (approved):
Looks good, thanks @ParadoxV5 ! Note, you'll need to re-record `sysvars_server_notembedded` to account for the read-only change.

PR #3866, opened 2025-03-04
MDEV-30000: Force an InnoDB checkpoint in mariadb-backup

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-30000*
## Description
At the start of `mariadb-backup --backup`, let us trigger a flush of the InnoDB buffer pool, so that as little log as possible will have to be copied.
## Release Notes
The previously debug-build-only interface
```sql
SET GLOBAL innodb_log_checkpoint_now=ON;
```
will be made available on all builds, and `mariadb-backup --backup` will invoke it.
## How can this PR be tested?
```sh
./mtr mariabackup.full_backup
```
## PR quality check


extra/mariabackup/xtrabackup.cc
```diff
@@ -4570,6 +4570,11 @@ static bool xtrabackup_backup_func()
 	}
 	msg("cd to %s", mysql_real_data_home);
 	encryption_plugin_backup_init(mysql_connection);
+	/* try to trigger a log checkpoint, but do not fail if the
+	server does not support this */
+	xb_mysql_query(mysql_connection,
```
vaintroub, line 4575: I think it makes sense to delay this call until after all other initialization is done. So that there is no minute-long delays in case of trivial parameter errors, e.g invalid path in --target-dir

vaintroub (changes requested):
Overall, seem fine.
Are you sure doing it in the oldest possible version is OK?

I had one comment about delaying potentially long "set global" until after other initialization is done, so that user does not wait long time  for the checkpoint, in case of trivial mistakes, for example wrong --target-dir or insufficient permissions for the files in data directory.

extra/mariabackup/xtrabackup.cc
dr-m: I agree, and I should have thought of this. Is there a way to send the statement and then wait for it to complete? I suppose that we would need another connection for that. It would be most convenient to do this from this current client thread. The execution of the statement needs to finish before `mariadb-backup --backup` reads the `ib_logfile0` of the server. I think that we need to send this statement as early as possible.

vaintroub: One can send statement, then wait for it to complete, in 2 steps, and it does not need another connection This is [send](https://mariadb.com/kb/en/mysql_send_query/) , this is [wait](https://mariadb.com/kb/en/mysql_read_query_result/)

dr-m: 279a5774a8d732514194207a05de3923a4ef6972 tries to achieve some overlap, which is not much, because the server’s `ib_logfile0` must be read very early on, in order to determine the latest checkpoint LSN.

dr-m:
The choice of target version is something that needs to be decided by @vuvova.

vaintroub (approved):
Ok, I'm fine with this version. 

You are aware, that when mariabackup fails because of invalid --target-dir, or file permissions, even after mariabackup process is finished, it still  can leave "zombie" connection on server executing `set_checkpoint_now()`. I think it is acceptable, just wanted to make sure that there is awareness about it.

dr-m:
Right, I am relying on the POSIX semantics that any files (including communications sockets) will be closed when the process terminates. If the semantics on Microsoft Windows is similar, we should be fine. As far as I understand, a client disconnect would not interrupt the execution of `SET GLOBAL innodb_log_checkpoint_now=ON`.

vaintroub:
> Right, I am relying on the POSIX semantics that any files (including communications sockets) will be closed when the process terminates. If the semantics on Microsoft Windows is similar, we should be fine.
> 
Those rules (everything is freed when process exits) are also in effect not just on POSIX, on Windows too. However, if you communicate with remote process, like mariadb server, over TCP or local sockets, and your '*client* process dies, there is not always immediate notification for mariadbd *server* process. If the connection is idle, yes, server would know about dead connection rather soon. But if your connection executes a lengthy statement, it would only know about dead connection next time it recv() from socket (not sure about write, writes can be delayed/cached). That is what I meant when talking about "zombie" - the remote process dies, but server does not know about it for long time, keeps connection running.

PR #3870, opened 2025-03-05
MDEV-36106 New-style hints: [NO_]DERIVED_CONDITION_PUSHDOWN, [NO_]MERGE

DaveGosselin-MariaDB:
Implements and tests the following optimizer hints:
- DERIVED_CONDITION_PUSHDOWN
- NO_DERIVED_CONDITION_PUSHDOWN
- MERGE
- NO_MERGE

Each of these are table-level hints.  The first pair enable and disable, respectively, the condition pushdown for derived tables which is typically controlled by the condition_pushdown_for_derived optimizer switch.  The second pair enable and disable, respectively, the derived table merge optimization which is typically controlled by the derived_merge optimizer switch.

### DIFF
--- a/mysql-test/main/opt_hints_derived_condition_pushdown.test
+++ b/mysql-test/main/opt_hints_derived_condition_pushdown.test
@@ -1,0 +1,417 @@
+--source include/have_sequence.inc
+--disable_view_protocol
+create table t1 (a int, b int, c int);
+create table t2 (a int, b int, c int, d decimal);
+insert into t1 values
+  (1,21,345), (1,33,7), (8,33,114), (1,21,500), (1,19,107), (5,14,787),
+  (8,33,123), (9,10,211), (5,16,207), (1,33,988), (5,27,132), (1,21,104),
+  (6,20,309), (6,20,315), (1,21,101), (8,33,404), (9,10,800), (1,21,123),
+  (7,11,708), (6,20,214);
+create index t1_a on t1 (a);
+insert into t2 values
+  (2,3,207,207.0000), (1,21,909,12.0000), (7,13,312,406.0000),
+  (8,64,248,107.0000), (6,20,315,279.3333), (1,19,203,107.0000),
+  (8,80,800,314.0000), (3,12,231,190.0000), (6,23,303,909.0000);
+create view v1 as select a, b, max(c) as max_c, avg(c) as avg_c from t1
+  group by a,b having max_c < 707;
+create table t3 select 2*seq as a, 2*seq+1 as b from seq_0_to_1000;
+CREATE TABLE t4 (a INT, b INT);
+INSERT INTO t4 VALUES (1,2),(2,3),(3,4);
+create table t5 select seq as i, 10*seq as j from seq_1_to_10;
+create view v2 as select * from t5;
+create table t6 (a int primary key);
+insert into t6 select * from seq_1_to_50;
+create view v6 as select a from t6 where a mod 2 = 1;
+
+set @save_optimizer_switch=@@optimizer_switch;
+
+set session optimizer_switch='condition_pushdown_for_derived=on';
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ NO_DERIVED_CONDITION_PUSHDOWN(v1) */ * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(v1) */ * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+select * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+
+--source include/explain-no-costs.inc
+explain format=json with cte as (
+  select /*+ QB_NAME(qb1) */ max_c, avg_c from v1,t2 where
+    ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a)))
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(v1@qb1) */ * from cte;
+with cte as (
+  select /*+ QB_NAME(qb1) */ max_c, avg_c from v1,t2 where
+    ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a)))
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(v1@qb1) */ * from cte;
+with cte as (
+  select /*+ QB_NAME(qb1) */ max_c, avg_c from v1,t2 where
+    ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a)))
+select * from cte;
+
+--source include/explain-no-costs.inc
+explain format=json with recursive cte as (
+  select max_c, avg_c from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+  ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a))
+  union
+  select * from cte where max_c < 100
+) select /*+ NO_DERIVED_CONDITION_PUSHDOWN(cte) */ * from cte;
+with recursive cte as (
+  select max_c, avg_c from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+  ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a))
+  union
+  select * from cte where max_c < 100
+) select /*+ NO_DERIVED_CONDITION_PUSHDOWN(cte) */ * from cte;
+with recursive cte as (
+  select max_c, avg_c from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+  ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a))
+  union
+  select * from cte where max_c < 100
+) select * from cte;
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt)*/ * from
+  (select t3.b as a from t3 group by t3.a) dt where (dt.a=3 or dt.a=5);
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt)*/ * from
+  (select t3.b as a from t3 group by t3.a) dt where (dt.a=3 or dt.a=5);
+select * from
+  (select t3.b as a from t3 group by t3.a) dt where (dt.a=3 or dt.a=5);
+
+--echo "This explain and query shows the result when no hints are specified, for comparison to hint cases in the next three explain and select pairs"
+--source include/explain-no-costs.inc
+explain format=json select * from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+select count(*) from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt)*/ * from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt)*/ count(*) from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ NO_DERIVED_CONDITION_PUSHDOWN(du)*/ * from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(du)*/ count(*) from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt,du)*/ * from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt,du)*/ count(*) from (
+  select t3.b as a from t3 group by t3.a
+) dt join (
+  select t3.a as a, t3.b as b from t3 where t3.a % 2 = 0 group by t3.b
+) du where (dt.a=3 or dt.a=5) and (du.a % 10 <> 0 or du.a % 5 <> 0);
+
+--source include/explain-no-costs.inc
+explain format=json SELECT /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt) */ * FROM (
+  SELECT t4.b AS a
+  FROM t4
+  GROUP BY t4.a
+) dt WHERE (dt.a=2);
+SELECT /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt) */ * FROM (
+  SELECT t4.b AS a
+  FROM t4
+  GROUP BY t4.a
+) dt WHERE (dt.a=2);
+SELECT * FROM (
+  SELECT t4.b AS a
+  FROM t4
+  GROUP BY t4.a
+) dt WHERE (dt.a=2);
+
+--source include/explain-no-costs.inc
+explain format=json SELECT /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt) */ * FROM (
+  SELECT qb.b AS a
+  FROM (
+    select 1*t4.b as b, 1*t4.a as a from t4
+  ) qb
+  GROUP BY qb.a
+) dt WHERE (dt.a=2);
+SELECT /*+ NO_DERIVED_CONDITION_PUSHDOWN(dt) */ * FROM (
+  SELECT qb.b AS a
+  FROM (
+    select 1*t4.b as b, 1*t4.a as a from t4
+  ) qb
+  GROUP BY qb.a
+) dt WHERE (dt.a=2);
+SELECT * FROM (
+  SELECT qb.b AS a
+  FROM (
+    select 1*t4.b as b, 1*t4.a as a from t4
+  ) qb
+  GROUP BY qb.a
+) dt WHERE (dt.a=2);
+
+--source include/explain-no-costs.inc
+explain format=json
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(@qb1 dv6) */ * from
+  (select /*+ QB_NAME(qb4) */ * from
+    (select /*+ QB_NAME(qb3) */ * from
+      (select /*+ QB_NAME(qb2) */ * from
+        (select /*+ QB_NAME(qb1) */ avg_a from
+          (select avg(a) as avg_a from v6) dv6
+        ) dv6) dv6) dv6) dv6 where avg_a <> 0;
+
+--source include/explain-no-costs.inc
+explain format=json
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(@qb2 dv6) */ * from
+  (select /*+ QB_NAME(qb4) */ * from
+    (select /*+ QB_NAME(qb3) */ * from
+      (select /*+ QB_NAME(qb2) */ * from
+        (select /*+ QB_NAME(qb1) */ avg_a from
+          (select avg(a) as avg_a from v6) dv6
+        ) dv6) dv6) dv6) dv6 where avg_a <> 0;
+
+--source include/explain-no-costs.inc
+explain format=json
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(@qb3 dv6) */ * from
+  (select /*+ QB_NAME(qb4) */ * from
+    (select /*+ QB_NAME(qb3) */ * from
+      (select /*+ QB_NAME(qb2) */ * from
+        (select /*+ QB_NAME(qb1) */ avg_a from
+          (select avg(a) as avg_a from v6) dv6
+        ) dv6) dv6) dv6) dv6 where avg_a <> 0;
+
+--source include/explain-no-costs.inc
+explain format=json
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(@qb4 dv6) */ * from
+  (select /*+ QB_NAME(qb4) */ * from
+    (select /*+ QB_NAME(qb3) */ * from
+      (select /*+ QB_NAME(qb2) */ * from
+        (select /*+ QB_NAME(qb1) */ avg_a from
+          (select avg(a) as avg_a from v6) dv6
+        ) dv6) dv6) dv6) dv6 where avg_a <> 0;
+
+set session optimizer_switch='condition_pushdown_for_derived=off';
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ DERIVED_CONDITION_PUSHDOWN(v1) */ * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+select /*+ DERIVED_CONDITION_PUSHDOWN(v1) */ * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+select * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+
+--source include/explain-no-costs.inc
+explain format=json with cte as (
+  select /*+ QB_NAME(qb1) */ max_c, avg_c from v1,t2 where
+    ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a)))
+select /*+ DERIVED_CONDITION_PUSHDOWN(v1@qb1) */ * from cte;
+with cte as (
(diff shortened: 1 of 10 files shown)


mysql-test/main/mdev-36106-merge-hint.test
Olernov, line 1: Same comment about the file name as for the one above.

Olernov, line 1: Views are also a subject to merge, would be nice to see them covered by test cases, however I'm sure whether this optimizer switch affects the behavior. I suggest adding more complicated tests, something like
`select * from (select a,b from v1) dt where ...`
and applying MERGE/NO_MERGE hints  to either `v1`, `dt` or both.

mysql-test/main/mdev-36106-derived-condition-pushdown-hint.test
```diff
@@ -0,0 +1,50 @@
...
+select json_detailed(json_extract(trace, '$**.original_condition')) from information_schema.optimizer_trace;
+
+set session optimizer_switch='condition_pushdown_for_derived=off';
+explain select * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+
+select json_detailed(json_extract(trace, '$**.original_condition')) from information_schema.optimizer_trace;
+
+explain select /*+ DERIVED_CONDITION_PUSHDOWN(v1) */ * from v1,t2 where
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Good suggestion, will add.

mysql-test/main/mdev-36106-derived-condition-pushdown-hint.test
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Good suggestion, will add.

mysql-test/main/mdev-36106-merge-hint.test
```diff
@@ -0,0 +1,27 @@
...
+create table t1 select seq as i, 10*seq as j from seq_1_to_10;
+
+set @save_optimizer_trace=@@optimizer_trace;
+set @save_optimizer_switch=@@optimizer_switch;
+set optimizer_trace='enabled=on';
+
+set session optimizer_switch='derived_merge=on';
+
+SELECT a, b FROM (SELECT i as a, j as b FROM t1) AS dt WHERE a < 3 AND b > 8;
+select json_extract(trace, '$.steps[0].join_preparation.steps[0].derived.algorithm') from information_schema.optimizer_trace;
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Same response as earlier 😉

sql/sql_select.cc
```diff
@@ -2490,38 +2490,44 @@ JOIN::optimize_inner()
...
-  if (optimizer_flag(thd, OPTIMIZER_SWITCH_COND_PUSHDOWN_FOR_DERIVED))
+  bool attempted_derived_cond_push= false;
+  TABLE_LIST *tbl;
+  List_iterator_fast<TABLE_LIST> li(select_lex->leaf_tables);
+  while ((tbl= li++))
   {
-    TABLE_LIST *tbl;
-    List_iterator_fast<TABLE_LIST> li(select_lex->leaf_tables);
-    while ((tbl= li++))
+    bool hint_allows_derived_pushdown= hint_table_state(
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Good point, will do.

sql/sql_select.cc
```diff
@@ -2490,38 +2490,44 @@ JOIN::optimize_inner()
...
+        thd, tbl->table, DERIVED_CONDITION_PUSHDOWN_HINT_ENUM,
+        optimizer_flag(thd, OPTIMIZER_SWITCH_COND_PUSHDOWN_FOR_DERIVED));
+    if (!hint_allows_derived_pushdown)
+      continue;
+    attempted_derived_cond_push= true;
+
+    /*
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Will do.

sql/sql_select.cc
```diff
@@ -2490,38 +2490,44 @@ JOIN::optimize_inner()
...
+        if (pushdown_cond_for_derived(thd, conds, tbl))
+          DBUG_RETURN(1);
       }
+      if (mysql_handle_single_derived(thd->lex, tbl, DT_OPTIMIZE))
+        DBUG_RETURN(1);
     }
   }
-  else
+
+  if (!attempted_derived_cond_push)
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: He clarified the expected behavior on slack, I'll update this accordingly.

sql/table.cc
```diff
@@ -10069,6 +10070,15 @@ bool TABLE_LIST::init_derived(THD *thd, bool init_view)
     bool forced_no_merge_for_update_delete=
            belong_to_view ? belong_to_view->updating :
                            !unit->outer_select()->outer_select();
+
+    if (select_lex->opt_hints_qb &&    // QB hints initialized
+        !this->opt_hints_table)        // Table hints are not adjusted yet
+      select_lex->opt_hints_qb->fix_hints_for_table(this);
+
+    bool hint_allows_derived_merge=
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Will do.

mysql-test/main/mdev-36106-merge-hint.test
DaveGosselin-MariaDB: Good suggestion, will do.

spetrunia:
Note: this adds
```
          !this->opt_hints_table)        // Table hints are not adjusted yet
        select_lex->opt_hints_qb->fix_hints_for_table(this);
```
code in TABLE_LIST::init_derived() which is called from open_and_lock_tables/mysql_handle_derived.

I guess the need to make a decision whether to do merging in open_and_lock_tables() is a given (we can't change it within the scope of this project).

DaveGosselin-MariaDB:
> ```
> MariaDB [test]>  explain extended select /*+ NO_MERGE(T) */ * from (select * from t1) T;
> +------+-------------+------------+------+---------------+------+---------+------+------+----------+-------+
> | id   | select_type | table      | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra |
> +------+-------------+------------+------+---------------+------+---------+------+------+----------+-------+
> |    1 | PRIMARY     | <derived2> | ALL  | NULL          | NULL | NULL    | NULL | 100  |   100.00 |       |
> |    2 | DERIVED     | t1         | ALL  | NULL          | NULL | NULL    | NULL | 100  |   100.00 |       |
> +------+-------------+------------+------+---------------+------+---------+------+------+----------+-------+
> 2 rows in set, 1 warning (2.764 sec)
> ```
> 
> Without the hint, the above query would merge the derived table. It didn't, so the hint had an effect. But EXPLAIN text or optimizer trace do not have anything about the hint:
> 
> ```
> MariaDB [test]> show warnings\G
> *************************** 1. row ***************************
>   Level: Note
>    Code: 1003
> Message: /* select#1 */ select `T`.`a` AS `a`,`T`.`b` AS `b` from (/* select#2 */ select `test`.`t1`.`a` AS `a`,`test`.`t1`.`b` AS `b` from `test`.`t1`) `T`
> 1 row in set (0.000 sec)
> ```
> 
>  ```
>       "join_optimization": {
>         "select_id": 1,
>         "steps": [
>           {
>             "hints": ""
>           },
>           {
>             "join_optimization": {
>               "select_id": 2,
>               "steps": [
>                 {
> ```

If I’m understanding the code correctly, it is working as expected.  We use a bitmap to remember which hints are specified.  But this bitmap conflates “is specified as false” with “not specified”.  So in the `Opt_hints::print` method, `is_specified` will return false if the hint is given but it’s hint value is false, and it won’t be printed in the optimizer trace output.

spetrunia:
@DaveGosselin-MariaDB,

Trying on MySQL (I am using 8.4 but I suppose earlier versions would work the same):
```
create table t1 (a int);
insert into t1 values (1),(2);
explain  select /*+ NO_MERGE(T) */ * from (select * from t1) T;
```
```
+----+-------------+------------+------------+------+---------------+------+---------+------+------+----------+-------+
| id | select_type | table      | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra |
+----+-------------+------------+------------+------+---------------+------+---------+------+------+----------+-------+
|  1 | PRIMARY     | <derived2> | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    2 |   100.00 | NULL  |
|  2 | DERIVED     | t1         | NULL       | ALL  | NULL          | NULL | NULL    | NULL |    2 |   100.00 | NULL  |
+----+-------------+------------+------------+------+---------------+------+---------+------+------+----------+-------+
2 rows in set, 1 warning (0,00 sec)
```
```
mysql> show warnings\G
  Level: Note
   Code: 1003
Message: /* select#1 */ select /*+ NO_MERGE(`T`@`select#1`) */ `T`.`a` AS `a` from (/* select#2 */ select `j21`.`t1`.`a` AS `a` from `j21`.`t1`) `T`
1 row in set (0,00 sec)
````
Note the NO_MERGE in the warning...

**Comment by DaveGosselin-MariaDB** [comment]
@spetrunia and @Olernov, I fixed the issue with the hint name not appearing in the warnings.  I'll update the PR as soon as I finish adding the new tests.

File: mysql-test/main/mdev-36106-derived-condition-pushdown-hint.test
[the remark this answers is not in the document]
Comment (Olernov) [reply mysql-test/main/mdev-36106-derived-condition-pushdown-hint.test:1]: I can't see them added so unresolving this thread.

File: sql/sql_select.cc (hunk 5 above)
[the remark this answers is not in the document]
Comment (Olernov) [reply sql/sql_select.cc:2530]: Unresolving this too not to forget

File: mysql-test/main/mdev-36106-derived-condition-pushdown-hint.test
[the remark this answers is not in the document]
Comment (DaveGosselin-MariaDB) [reply mysql-test/main/mdev-36106-derived-condition-pushdown-hint.test:1]: Thanks, I must have resolved this one by mistake.

File: sql/sql_select.cc (hunk 5 above)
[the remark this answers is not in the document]
Comment (DaveGosselin-MariaDB) [reply sql/sql_select.cc:2530]: This one is fine to keep resolved.  I have made changes locally but not yet uploaded to the PR.  I resolve these as a checklist when I update locally.

File: sql/opt_hints.cc [hunk 7]
```diff
@@ -45,6 +45,8 @@ struct st_opt_hint_info opt_hint_info[]=
   {{STRING_WITH_LEN("JOIN_SUFFIX")}, false, true, true},
   {{STRING_WITH_LEN("JOIN_ORDER")}, false, true, true},
   {{STRING_WITH_LEN("JOIN_FIXED_ORDER")}, false, true, false},
+  {{STRING_WITH_LEN("DERIVED_CONDITION_PUSHDOWN")}, false, false},
```
Olernov, line 48: There's now one more field `irregular_hint` in the struct, please add `false` to your hints

mysql-test/main/mdev-36106-merge-hint.test
```diff
@@ -0,0 +1,80 @@
...
+--source include/explain-no-costs.inc
+explain format=json select * from (select i, j from v1) dt where dt.j % 5 = 0;
+select * from (select i, j from v1) dt where dt.j % 5 = 0;
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ NO_MERGE(dt) */ * from (select /*+ NO_MERGE(v1) */ i, j from v1) dt where dt.j % 5 = 0;
+select /*+ NO_MERGE(dt) */ * from (select /*+ NO_MERGE(v1) */ i, j from v1) dt where dt.j % 5 = 0;
+
+--source include/explain-no-costs.inc
+explain format=json select * from (select /*+ NO_MERGE(v1) */ i, j from v1) dt where dt.j % 5 = 0;
```
Olernov, line 28: The hint doesn't affect merging of a view, it's merged anyway. I suggest removing this test case and similar ones below as they don't make much sense.

mysql-test/main/mdev-36106-derived-condition-pushdown-hint.test
```diff
@@ -0,0 +1,297 @@
+--source include/have_sequence.inc
+--disable_ps_protocol
```
Olernov, line 2: No need to disable ps-protocol, the test passes OK

mysql-test/main/mdev-36106-merge-hint.test
```diff
@@ -0,0 +1,80 @@
+--source include/have_sequence.inc
+--disable_ps_protocol
```
Olernov, line 2: No need to disable it

mysql-test/main/mdev-36106-merge-hint.test
```diff
@@ -0,0 +1,80 @@
...
+select * from (select /*+ MERGE(v1) */ i, j from v1) dt where dt.j % 5 = 0;
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ MERGE(dt) */ * from (select i, j from v1) dt where dt.j % 5 = 0;
+select /*+ MERGE(dt) */ * from (select i, j from v1) dt where dt.j % 5 = 0;
+
+set optimizer_switch=@save_optimizer_switch;
+
+# Test warnings
+--enable_warnings
```
Olernov, line 64: Warnings are enabled by default, no need to explicitly enable them

mysql-test/main/mdev-36106-merge-hint.test
DaveGosselin-MariaDB: I added these for completeness, to illustrate what the behavior is, even though the hint doesn't have an effect.  But I can remove them if you think they're distracting.  I tend to think of tests not only as exercising expected behaviors, but also as implied documentation.

mysql-test/main/opt_hints_derived_condition_pushdown.test
```diff
@@ -0,0 +1,295 @@
...
+
+set @save_optimizer_switch=@@optimizer_switch;
+
+set session optimizer_switch='condition_pushdown_for_derived=on';
+
+--source include/explain-no-costs.inc
+explain format=json select /*+ NO_DERIVED_CONDITION_PUSHDOWN(v1) */ * from v1,t2 where
+  ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a));
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(v1) */ * from v1,t2 where
```
Olernov, line 34: If we're testing the results, probably It's worth adding this statement without hints to see that the resultset is not affected by hints

Olernov: Same note about explain. Here and below, otherwise the hint effect is not obvious

DaveGosselin-MariaDB: Yes, or we can remove the case.  It's another instance where I added it to illustrate what the behavior is, even though the hint doesn't have an effect.  In an earlier example I removed the test case, so for consistency it's probably best to remove this one too.

DaveGosselin-MariaDB: I reran the test and indeed this case shows that the hint is working (check the `having_condition`).  It may have been more clear had I remained with the `JSON` parsing implementation instead of preserving the entire `explain` output.

Olernov: I tried to say: first we put the statement without hints to our `.test` file to see the initial execution plan and initial resultset (which we can then compare with resultsets from hinted statements to make sure hints hasn't broken anything). You can find examples of that in `opt_hints.test`:
```
--echo # Make sure BKA is expected to be used when there are no hints
EXPLAIN SELECT * FROM t12, t13 WHERE t12.a=t13.a AND (t13.b+1 <= t13.b+1);
--echo # Generate sample resultset
SELECT * FROM t12, t13 WHERE t12.a=t13.a AND (t13.b+1 <= t13.b+1);
...
EXPLAIN SELECT /*+ BKA(t13) */ * FROM t12, t13  WHERE t12.a=t13.a AND (t13.b+1 <= t13.b+1);
--echo # Resultset must be the same as the sample one
SELECT /*+ BKA(t13) */ * FROM t12, t13 WHERE t12.a=t13.a AND (t13.b+1 <= t13.b+1);
```
If we don't have a sample resultset it is not obvious whether the results of hinted statement correct or not.

sql/opt_hints.cc
```diff
@@ -398,6 +401,35 @@ Opt_hints_qb::Opt_hints_qb(Opt_hints *opt_hints_arg,
...
+    insufficient to consider it fixed for the case where a TABLE
+    instance is required but not yet available.  If the associated
+    table isn't yet fixed, then fix this hint as though it were unfixed.
+  */
+  DBUG_ASSERT(!table_list->opt_hints_table);
+  DBUG_ASSERT(tab);
+  DBUG_ASSERT(tab->child_array_ptr()->size() == 0);
+  table_list->opt_hints_qb= this;
+  table_list->opt_hints_table= tab;
+  set_fixed();
```
Olernov, line 427: Are you sure this is correct? Once again we are processing a single table but `set_fixed()` for the whole query block.

sql/sql_base.cc
```diff
@@ -8329,9 +8329,9 @@ bool setup_tables(THD *thd, Name_resolution_context *context,
         (2)  Table hints are not adjusted yet
         (3)  Table is not in the INSERT part of INSERT..SELECT
       */
-      if (qb_hints &&                          // (1)
-          !table_list->opt_hints_table &&      // (2)
-          !(select_insert && is_insert_part))  // (3)
+      if (qb_hints &&                              // (1)
```
Olernov, line 8332: So this condition remains unchanged, right? Then no need to change formatting or add extra brackets, let's just keep these 3 lines unmodified

sql/sql_select.cc
```diff
@@ -2513,43 +2513,42 @@ JOIN::optimize_inner()
...
-    List_iterator_fast<TABLE_LIST> li(select_lex->leaf_tables);
-    while ((tbl= li++))
+    const bool is_derived_pushdown_allowed= hint_table_state(
+      thd, tbl->table, DERIVED_CONDITION_PUSHDOWN_HINT_ENUM,
+      optimizer_flag(thd, OPTIMIZER_SWITCH_COND_PUSHDOWN_FOR_DERIVED));
+    if (!is_derived_pushdown_allowed)
     {
-      /* 
+      /* Run optimize phase on this derived table/view. */
+      if (tbl->is_view_or_derived() &&
```
Olernov, line 2526: This change needs verification. Before we had
```
/* Run optimize phase for all derived tables/views used in this SELECT. */
    if (select_lex->handle_derived(thd->lex, DT_OPTIMIZE))
```
which processed all tables of the `SQL_I_List st_select_lex::table_list`:
```
bool st_select_lex::handle_derived(LEX *lex, uint phases)
{
  return lex->handle_list_of_derived(table_list.first, phases);
}
```
But now we are iterating over `select_lex->leaf_tables` which can theoretically be different from `SQL_I_List st_select_lex::table_list`.  Maybe it's OK

DaveGosselin-MariaDB: I'm not sure what you mean by "needs verification" because all the tests pass.  Is that not sufficient verification?   If there's a missing test case, then we should add it.  Sergei had asked me to make this change as it wasn't in my original patch set.

Olernov: I mean @spetrunia and/or @sanja-byelkin should definitely review this. Passing tests does not guarantee it doesn't break some complicated scenarios with INSERT..SELECT, constant tables, subqueries, CTEs and so on.

mysql-test/main/opt_hints_derived_condition_pushdown.test
Olernov, line 1: Are CTE's and recursive CTE's a subject for condition pushdown? If so, would be nice to see some test cases

sql/opt_hints.h
```diff
@@ -450,6 +460,15 @@ class Opt_hints_qb : public Opt_hints
   void append_hint_arguments(THD *thd, opt_hints_enum hint,
                              String *str) override;
 
+  /**
+    Function finds Opt_hints_table object corresponding to
+    table alias in the query block and attaches corresponding
+    key hint objects to appropriate KEY structures.
```
Olernov, line 466: This part of comment is not correct if we are fixing hints for `TABLE_LIST`.

sql/opt_hints.cc
```diff
@@ -398,6 +401,33 @@ Opt_hints_qb::Opt_hints_qb(Opt_hints *opt_hints_arg,
 }
 
 
+void Opt_hints_qb::fix_hints_for_table(TABLE_LIST *table_list)
+{
+  Opt_hints_table *tab= static_cast<Opt_hints_table *>(find_by_name(table_list->alias));
```
Olernov, line 406: This line is too long (89 chars)

sql/opt_hints.h
```diff
@@ -260,8 +270,8 @@ class Opt_hints : public Sql_alloc
   }
   void set_name(const Lex_ident_sys &name_arg) { name= name_arg; }
   Opt_hints *get_parent() const { return parent; }
-  void set_fixed() { fixed= true; }
-  bool is_fixed() const { return fixed; }
+  void set_fixed(Fixed_state fv=Fixed_state::FIXED) { fixed= fv; }
```
Olernov, line 273: Someone from the team (probably Monty) was negative about using default function parameters (I am personally not but still)

sql/sql_select.cc
DaveGosselin-MariaDB: Ideally, passing tests would mean that it doesn't break those complicated scenarios.  If we can imagine them, then we should have test cases for them.  Requiring specific individuals to bless code doesn't scale, but capturing those cases as tests does.

mysql-test/main/opt_hints_derived_condition_pushdown.test
DaveGosselin-MariaDB: It seems that there's a larger architectural question hidden here.  We could add such test cases, and I'm not against doing that.  But this code fundamentally adds a per-query switch where we previously had one only per-session.  Is it possible to use hints here to break previously working query plans?  If not, then architecturally there's no problem.  But if yes, then how do we prevent bugs?  Fundamentally that is an architectural problem I think.

sql/opt_hints.h
DaveGosselin-MariaDB: Good catch, will fix.

sql/opt_hints.cc
DaveGosselin-MariaDB: Good catch, will fix.

sql/opt_hints.h
svoj: Dunno, there're quite a few places where we use default parameters. E.g. see `set_killed()`, which was added by Monty. OTOH in this particular case `set_fixed(UNFIXED)` is slightly confusing. Without diving into details too much I'd probably do `set_not_fixed()`.

But in general it is a good example of a badly encapsulated member. There is no point in having `Opt_hints::fixed` as `private` member when it can be manipulated externally via `set_fixed()` and `is_fixed()`.

DaveGosselin-MariaDB: In this case, I think we don't need the argument at all anymore.  I had added it as part of delayed hint processing, but we found a way to implement what we need without the concept of a delayed hint.  So I think we can remove the argument entirely.

mysql-test/main/opt_hints_derived_condition_pushdown.test
DaveGosselin-MariaDB: There is a problem with recursive CTEs where the optimizer switch disables condition pushdown and a hint attempts to enable it:  it is not enabled (incorrect behavior).  Interestingly, when the optimizer switch allows condition pushdown and a hint disables it, then it is disabled correctly.

sql/sql_select.cc
Olernov: I agree, however, that can be a bit tricky. Let's construct a test case where a derived table is present in `st_select_lex::leaf_tables` but is absent in `st_select_lex::table_list`  (vice versa is also interesting). If such a case is impossible then why? Probably we can add some assertions to solidify this fact.

sql/table.cc
```diff
@@ -10116,6 +10117,15 @@ bool TABLE_LIST::init_derived(THD *thd, bool init_view)
     bool forced_no_merge_for_update_delete=
            belong_to_view ? belong_to_view->updating :
                            !unit->outer_select()->outer_select();
+
+    if (select_lex->opt_hints_qb &&    // QB hints initialized
+        !this->opt_hints_table)        // Table hints are not adjusted yet
+      select_lex->opt_hints_qb->fix_hints_for_table(this);
```
Olernov, line 10123: There may be a problem with this approach: what if we have both `MERGE/NO_MERGE` hint and an index hint (`NO_ICP`, for example) for a single table?  Won't we skip proper setting up of `Opt_hints_key` structure?  Of course, index hints are not applicable for derived tables but what if a user mistakenly writes something like
`select /*+ MERGE(t1) NO_ICP(t1 t1_key1)*/ from t1 where  key1_col=:value ` ?

DaveGosselin-MariaDB: I'll try that out and see what happens.  Maybe we can emit an error from this point in the code?  What is your suggestion on how to handle such a case?  In general, this is another architectural problem with our code:  we don't do any validation before we start executing the query.  This means that we will not emit an error to the user as early as possible.
If this progresses to the point that every hint has to worry about other hints, then the hints implementation is broken and should be reworked:  the hint module (be it the parser or something else) should have all of the information it needs to allow callers of the hint APIs to receive reliable information about hint state without having to worry about other hints.

DaveGosselin-MariaDB:
> I do not yet understand the topic well but it doesn't look good when there are two functions with the same name that do slightly different things:
> 
> ```
> void Opt_hints_qb::fix_hints_for_table(TABLE_LIST *table_list);
> Opt_hints_table *Opt_hints_qb::fix_hints_for_table(TABLE *table,
>                                                    const Lex_ident_table &alias)
> ```
Both methods fix the hints for the passed table.  From any caller's perspective, these methods achieve the same outcome which is why they have the same name.
> 

> for now, I can cause a debug assert in fix_hints_for_table() by this (not currently meaningful) query:
> 
> ```
> create table t1 (a int);insert into t1 values (1),(2);
> explain 
> select /*+ NO_MRR(T index_doesnt_exist) NO_DERIVED_CONDITION_PUSHDOWN(T) */ * 
> from (select * from t1) T where T.a <10;
> ```
Without the assert in place, this query fails in the following way:
```
MariaDB [test]> explain select /*+ NO_MRR(T index_doesnt_exist) NO_DERIVED_CONDITION_PUSHDOWN(T) */ * from (select * from t1) T where T.a < 10;
ERROR 1054 (42S22): Unknown column 'T.a' in 'WHERE'
```

sql/table.cc
```diff
@@ -10116,6 +10117,15 @@ bool TABLE_LIST::init_derived(THD *thd, bool init_view)
     bool forced_no_merge_for_update_delete=
            belong_to_view ? belong_to_view->updating :
                            !unit->outer_select()->outer_select();
+
```
spetrunia, line 10120: This got me wondering, when how come table hints could be adjusted already?  The testcase shows this happens when merge operations moves derived tables from one select to another.  Please add a comment.

mysql-test/main/opt_hints_derived_condition_pushdown.test
```diff
@@ -0,0 +1,411 @@
...
+    ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a)))
+select /*+ NO_DERIVED_CONDITION_PUSHDOWN(v1@qb1) */ * from cte;
+with cte as (
+  select /*+ QB_NAME(qb1) */ max_c, avg_c from v1,t2 where
+    ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or
+    ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a)))
+select * from cte;
+
+--source include/explain-no-costs.inc
```
spetrunia, line 58: Is there such thing as condition pushdown into a recursive CTE?   I think MariaDB's condition pushdown can't be applied to recursive CTEs. 

And this testcase is doing:
```
) select /*+ NO_DERIVED_CONDITION_PUSHDOWN(cte) */ * from cte;
```
so there's nothing to push.   
This may still add some coverage but doesn't look very meaningful at the first glance.

spetrunia (changes requested):
Getting close.
Please fix these and then we're almost done

mysql-test/main/opt_hints_derived_condition_pushdown.test
spetrunia: .. so, if you think this still is worth having, it's ok to keep it. Otherwise, remove.

DaveGosselin-MariaDB:
> Getting close. Please fix these and then we're almost done

Thank you @spetrunia , I will make the fixes and update the PR.

sql/opt_hints.cc
```diff
@@ -398,10 +401,39 @@ Opt_hints_qb::Opt_hints_qb(Opt_hints *opt_hints_arg,
 }
 
 
+void Opt_hints_qb::fix_hints_for_table(TABLE_LIST *table_list)
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Will do!  👍

mysql-test/main/opt_hints_derived_condition_pushdown.test
DaveGosselin-MariaDB: I will re-evaluate the test cases and keep the more meaningful ones in place, thank you for inspecting them.

DaveGosselin-MariaDB: It's not easy to see amid the test output, but the case:
```
explain format=json with recursive cte as (   select max_c, avg_c from v1,t2 where   ((v1.max_c>300) and (v1.avg_c>t2.d) and (v1.b=t2.b)) or   ((v1.max_c<135) and (v1.max_c<t2.c) and (v1.a=t2.a))   union   select * from cte where max_c < 100 ) select /*+ NO_DERIVED_CONDITION_PUSHDOWN(cte) */ * from cte;
```
results in a different `"having"` condition (that is, the condition is pushed).  With `NO_DERIVED_CONDITION_PUSHDOWN(cte)` we see:
```
...
"having_condition": "max_c < 707",
...
```
and without the hint (or with `DERIVED_CONDITION_PUSHDOWN(cte)`) we see:
```
...
"having_condition": "max_c < 707 and (max_c > 300 or max_c < 135)",
...
```
These test cases are useful, so I'll preserve them.

spetrunia:
Note: this is failing on Windows: https://buildbot.mariadb.org/#/builders/234/builds/39321/steps/11/logs/stdio .
use small letters for table aliases...

DaveGosselin-MariaDB:
> Note: this is failing on Windows: https://buildbot.mariadb.org/#/builders/234/builds/39321/steps/11/logs/stdio . use small letters for table aliases...

Thank you for pointing this out, I fixed it in the latest patch but the GitHub actions/buildbot runners haven't picked it up yet for some reason.

PR #3879, opened 2025-03-08
Fix sporadic failures of rpl.rpl_gtid_crash

knielsen:
- Suppress a couple errors the slave can get as the master crashes.

 - The mysql-test-run occasionally takes 120 seconds between crashing
   the master and starting it back up for some (unknown) reason. For
   now, work-around that by letting the slave try for 500 seconds to
   connect to master before giving up instead of only 100 seconds.

Signed-off-by: Kristian Nielsen <knielsen@knielsen-hq.org>

### DIFF
--- a/mysql-test/include/long_test.inc
+++ b/mysql-test/include/long_test.inc
@@ -1,6 +1,6 @@
 # We use this --source include to mark a test as taking long to run.
 # We can use this to schedule such test early (to not be left with
-# only one or two long tests running, and rests of works idle), or to
+# only one or two long tests running, and rests of workers idle), or to
 # run a quick test skipping long-running test cases.
 
 --source include/no_valgrind_without_big.inc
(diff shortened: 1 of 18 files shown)


knielsen:
The increase from 100 to 500 seconds to connect to the master is a bit dubious, as I did not yet track down the cause of a very occasional 120 second delay in restarting the server after a crash. But I did verify with --rr that it looks to be something outside the server, it is not the server code that's stuck. So maybe better to work-around in the test case to not have a sporadic failure; waiting 100 or 500 seconds in this test is no difference in terms of correctness.

bnestere (approved):
Looks good, thanks @knielsen ! Also for the second commit titled "Add --source include/long_test.inc to some tests", can you extend the commit message to explain why it helps (I think you describe it somewhere else, I don't recall where, but I think it would be good to have in the commit message too).

PR #3881, opened 2025-03-09
MDEV-35876: speedup collation/charset lookup

jiteshchawla1511:
- [x] *The Jira issue number for this PR is: MDEV-35876*

## Description
- Replaces O(n) linear scans for collation lookups with O(1) hash lookups to eliminate performance bottlenecks as collation counts grow.
- No user-visible output changes. Internally, collation searches are faster (hash-based vs. array scan).
- Hash initialization occurs once during server startup, using existing thread-safe mechanisms.

## Release Notes
N/A 

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysys/charset.c
+++ b/mysys/charset.c
@@ -30,6 +30,7 @@
 #endif
 
 extern HASH charset_name_hash;
+extern HASH collation_name_hash;
 
 /*
   The code below implements this functionality:
@@ -45,16 +46,10 @@ static uint
 get_collation_number_internal(const char *name)
 {
 
-  CHARSET_INFO **cs;
-  for (cs= all_charsets;
-       cs < all_charsets + array_elements(all_charsets);
-       cs++)
-  {
-    if (cs[0] && cs[0]->coll_name.str &&
-        !my_strcasecmp_latin1(cs[0]->coll_name.str, name))
-      return cs[0]->number;
-  }  
-  return 0;
+  CHARSET_INFO *cs;
+  size_t name_len = strlen(name);
+  cs = (CHARSET_INFO*) my_hash_search(&collation_name_hash,(uchar*) name, name_len);
+  return cs ? cs->number : 0;
 }
 
 
@@ -712,6 +707,16 @@ static void init_available_charsets(void)
   my_hash_init2(key_memory_charsets, &charset_name_hash, 16,
                 &my_charset_latin1, 64, 0, 0, get_charset_key,
                 0, 0, HASH_UNIQUE);
+  
+  /* Initialize collation name hash */
++  my_hash_init2(key_memory_charsets, &collation_name_hash, 16,
++                &my_charset_latin1, 64, 0, 0,
++                [](const uchar *object, size_t *length, my_bool) -> uchar* {
++                  CHARSET_INFO *cs = (CHARSET_INFO*) object;
++                  *length = cs->coll_name.length;
++                  return (uchar*) cs->coll_name.str;
++                },
+                 0, 0, HASH_UNIQUE);
 
   init_compiled_charsets(MYF(0));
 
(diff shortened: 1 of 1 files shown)


cvicentiu:
@svoj I believe you can offer a preliminary review for this + testing.

svoj (changes requested):
Thanks for your contribution. The patch looks correct, but please apply coding style and compilation fixes [fixes.patch](https://github.com/user-attachments/files/19198264/fixes.patch)

Also please squash all commits into one. And write nice commit message, something like:
```
MDEV-35876 - speedup collation/charset lookup

Replaced linear character set array scanning with hash lookup.
```

Did you run any benchmarks to make sure this change actually brings improvement?

jiteshchawla1511:
> Thanks for your contribution. The patch looks correct, but please apply coding style and compilation fixes [fixes.patch](https://github.com/user-attachments/files/19198264/fixes.patch)
> 
> Also please squash all commits into one.
> 
> And write nice commit message, something like:
> 
> ```
> MDEV-35876 - speedup collation/charset lookup
> 
> Replaced linear character set array scanning with hash lookup.
> ```
> 
> It should be targeted for `main` branch.
> 
> Did you run any benchmarks to make sure this change actually brings improvement?

- I was thinking of squashing them once the changes are approved.
- My Apologies, will put nice commit messages from now onwards
- I took the 11.5 branch. So should I change the branch to main from the edit menu itself or should I create new PR ?
- I did not ran any test to benchmark this thing. I am afraid, not sure about how to benchmark this. If there any leads you can provide, I can benchmark this test.

svoj:
@jiteshchawla1511 

> I was thinking of squashing them once the changes are approved.

Feels like the time is now. At this point code LGTM, despite last hunk that appeared in [24da94c](https://github.com/MariaDB/server/pull/3881/commits/24da94c306f60a08dd8717f546b80ea85b8ad19c), it is incorrect unrelated change and should be reverted.

> I took the 11.5 branch. So should I change the branch to main from the edit menu itself or should I create new PR ?

It is slightly more complex:
1. we should stay with this PR
2. `git rebase` or `git cherry-pick` or manually cherry-pick your commits on top of `main`
3. force push to `jiteshchawla1511:MDEV-35876`
4. change the branch to main from the edit menu

> I did not ran any test to benchmark this thing. I am afraid, not sure about how to benchmark this. If there any leads you can provide, I can benchmark this test.

Benchmark details were unfortunately omitted in MDEV-35876, so we'll have to come up with something ourselves. Probably start with single threaded sysbench OLTP_RO or OLTP_PS. Use something like `perf` for profiling and check the difference between old `get_collation_number_internal` (and its callees) and new one.

svoj:

1. Suggested benchmark:
```
DELIMITER /
FOR i IN 1..100 DO
  DO COALESCE(_utf32'' collate uca1400_croatian_ai_ci);
END FOR;
/
DELIMITER ;
```

Play with number of iterations. To prove there was a problem compare unpatched `_latin1'' collate latin1_bin` vs unpatched `_utf32'' collate uca1400_croatian_ai_ci`. To prove the problem is fixed compare unpatched `_utf32'' collate uca1400_croatian_ai_ci` vs patched `_utf32'' collate uca1400_croatian_ai_ci`. Also make sure unpatched `_latin1` vs patched `_latin1` didn't degrade much.

2. We should try to get rid of `strlen()` in `get_collation_number_internal()`. But it seems we agreed that it doesn't have to happen within the scope of MDEV-35876.

3. Double check if a separate loop around `my_hash_insert(collation_name_hash)` is needed. What was the reason not have it along with `my_hash_insert(charset_name_hash)`? Apparently `charset_name_hash` is never used at runtime.

jiteshchawla1511:
Thanks @svoj for the help

PR #3883, opened 2025-03-10
MDEV-19248 Implement MASTER_BIND for CHANGE MASTER

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-19248](https://jira.mariadb.org/browse/MDEV-19248)*
  * Ports MySQL 5.6.2’s [WL#3127](https://dev.mysql.com/worklog/task/?id=3127), mysql/mysql-server@9174364

## Description & Release Notes
Add CHANGE MASTER TO & SHOW REPLICA STATUS field [`Master_Bind`](https://dev.mysql.com/doc/refman/9.2/en/change-replication-source-to.html#crs-opt-source_bind) that configures the `MYSQL_OPT_BIND` of the embedded replication client

The `mariadb` option `MYSQL_OPT_BIND` ([`--bind-address`](https://dev.mysql.com/doc/refman/9.2/en/mysql-command-options.html#option_mysql_bind-address) with [MDEV-21465](https://jira.mariadb.org/browse/MDEV-21465)) chooses the network interface (e.g., eternet port) that connects to the server (the Master in this case).

### Knowledge Base pages that need changing
* https://mariadb.com/kb/en/show-replica-status/
* https://mariadb.com/kb/en/change-master-to/

## How can this PR be tested?
good question.

## PR quality check


mysql-test/main/master_bind_basic.result
```diff
@@ -0,0 +1,55 @@
...
+Connection_name	Master_Bind
+	
+named	
+# Replace when specified
+CHANGE MASTER 'named' TO master_bind='localhost';
+CHANGE MASTER TO master_bind='127.0.0.1';
+SELECT Connection_name, Master_Bind FROM information_schema.SLAVE_STATUS;
+Connection_name	Master_Bind
+	127.0.0.1
+named	localhost
```
mariadb-SusilBehera, line 14: IMO master_bind is for network interface name. But here in the test we've IP address and hostname instead. Shouldn't we test with  network interface name?

mysql-test/main/master_bind_basic.result
```diff
@@ -0,0 +1,55 @@
...
+CHANGE MASTER TO master_bind=10;
+ERROR 42000: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '10' at line 1
+CHANGE MASTER 'named' TO master_bind=11;
+ERROR 42000: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '11' at line 1
+# `''` also counts as specified
+CHANGE MASTER TO master_bind='';
+CHANGE MASTER 'named' TO master_bind='';
+SELECT Connection_name, Master_Bind FROM information_schema.SLAVE_STATUS;
+Connection_name	Master_Bind
+
```
mariadb-SusilBehera, line 52: We can add a test with a non-existing network interface value. SHOW SLAVE STATUS should error out.

mysql-test/main/master_bind_basic.result
ParadoxV5: Good call. Why did MySQL document as and test with IP addresses 😕?

mysql-test/main/master_bind_basic.result
ParadoxV5: Be it interface or address, currently the field doesn’t check its string value and rely on the replicating client to check its options at START REPLICA.

And I think that’s fine: technically one might upgrade their hardware upgrade inbetween CHANGE MASTER nand START REPLICA.

mysql-test/main/master_bind_basic.result
mariadb-SusilBehera: I couldn't find any MySQL test which does MASTER_BIND testing. But the doc https://dev.mysql.com/doc/refman/8.0/en/change-master-to.html says MASTER_BIND = 'interface_name'.

mysql-test/main/master_bind_basic.result
mariadb-SusilBehera: Ok. Shall cover it at QA stage.

mysql-test/main/master_bind_basic.result
ParadoxV5: This one linked in PR description https://github.com/mysql/mysql-server/commit/9174364, or its version on HEAD https://github.com/mysql/mysql-server/blob/6b6d3ed/mysql-test/common/rpl/change_replication_source.test#L285-L313

mariadb-SusilBehera: Thank you for the pointers. MASTER_BIND is network interface IP address only. The doc should have MASTER_BIND = 'interface_ip_address' instead. But do we need MASTER_HOST if MASTER_BIND is set?

ParadoxV5: * `MASTER_BIND` = IP of this replica to connect with
* `MASTER_HOST` = IP of the primary to connect to

mariadb-SusilBehera (approved):
Looks good to me.

bnestere (changes requested):
Unfortunately the underlying code doesn't actually work. I set up a network namespace and a veth connection to it, started a mariadb instance (as a primary) in the namespace, and then tried to set up replication to the server using `MASTER_BIND` using my regular NIC and again using the veth NIC, and unfortunately it always picked the veth NIC, no matter the `MASTER_BIND` value. Then I tried using MySQL 8.0 and its `MASTER_BIND`, and the correct interface was always chosen.

Looking into our sources, it looks like `MYSQL_OPT_BIND` might be a dead option. We'll have to investigate further in the next feature development cycle.

knielsen (changes requested):
The code generally looks good. But as Brandon points out, there is a problem
in libmariadb that prevents this from working currently.

You can see the problem in the function `pvio_socket_connect()` in
`libmariadb/plugins/pvio/pvio_socket.c` . The code actually tries to do so:
```c
    /* if client has multiple interfaces, we will bind socket to given
     * bind_address */
    if (cinfo->mysql->options.bind_address)
    {
      wait_gai= 1;
      while ((gai_rc= getaddrinfo(cinfo->mysql->options.bind_address, 0,
                          &hints, &bind_res)) == EAI_AGAIN)
      {
        unsigned int timeout= mysql->options.connect_timeout ?
                              mysql->options.connect_timeout : DNS_TIMEOUT;
        if (time(NULL) - start_t > (time_t)timeout)
          break;
#ifndef _WIN32
        usleep(wait_gai);
#else
        Sleep(wait_gai);
#endif
        wait_gai*= 2;
      }
      if (gai_rc != 0 || !bind_res)
      {
        PVIO_SET_ERROR(cinfo->mysql, CR_BIND_ADDR_FAILED, SQLSTATE_UNKNOWN, 
                     CER(CR_BIND_ADDR_FAILED), cinfo->mysql->options.bind_address, gai_rc);
        goto error;
      }
    }
```
It looks up the specified address. But it looks like the author of the code
forgot to actually call bind(2) with the looked up results!

I think you should report this as a bug for libmariadb (CONC in Jira).

You might also try fixing the bug in the libmariadb and submitting a pull
request for that to Georg? It looks like it might simply be calling bind()
in the above code, passing the looked up address and address lenght.

And once libmariadb is fixed, we clearly need some testing done. Maybe
Brandon can help you with how he set up multiple interfaces. Basically it
needs to be tested that bind can be used and verified that traffic goes on
the specified interface. Maybe it can maybe be done simply by specifying
either the local (127.0.0.1) or the wifi/ethernet card of the machine, and
checking with tcpdump that the packets have the right client ip address.

I'm not sure an mtr testcase in the main suite is feasible, too many
assumptions needed about the environment (mtr tests are run by some users
too). Maybe with some --source include/have_multi_interface.inc, but it
might be tricky to get it to work.

I think it's ok to test manually, and describe briefly in commit comment or
pull request how it was tested.

vuvova:
I thought that replication code doesn't use libmariadb. And server implementation of the client library seems to ignore MYSQL_OPT_BIND

knielsen:
Hm, yes Serg, now that you mention it that's what I remember too. And looking closer at the libmariadb code, it _does_ in fact seem to call bind() correctly with the bind_address.

So seems I was just confused, and it's the server-side protocol code that needs fixing to support client-side bind

ParadoxV5:
> I thought that replication code doesn't use libmariadb. And server implementation of the client library seems to ignore MYSQL_OPT_BIND

Which is interesting as I could only find some functions I grepped in libmariadb. I thought Server includes libmariadb to reuse its Client code.

sql/sql_yacc.yy
```diff
@@ -2268,6 +2269,10 @@ master_def:
           {
             Lex->mi.host = $3.str;
           }
+        | MASTER_BIND_SYM '=' TEXT_STRING_sys
+          {
+            Lex->mi.bind_addr = $3.str;
```
itzanway, line 2274: The grammar addition here aligns well with the existing MASTER_USER implementation, and correctly adding MASTER_BIND_SYM to the unreserved keyword list avoids breaking existing schemas.

A quick memory-lifecycle check: Since $3.str is allocated on the statement's thd->mem_root, it will be freed at the end of the query execution. Can we verify that when Lex->mi.bind_addr is passed down to change_master() (in sql_repl.cc), it is being properly deep-copied into the persistent Master_info struct (e.g., via strmake or String::copy) rather than just pointer-assigned

sql/rpl_mi.cc
```diff
@@ -32,6 +32,7 @@ static void init_master_log_pos(Master_info* mi);
 Master_info::Master_info(LEX_CSTRING *connection_name_arg,
                          bool is_slave_recovery)
   :Slave_reporting_capability("I/O"),
+   bind_addr(""),
```
itzanway, line 35: Replacing the dummy placeholder in the master.info file read/write logic with mi->bind_addr is a great way to handle this while maintaining seamless cross-version compatibility!

One minor detail regarding the Master_info constructor: I noticed bind_addr("") is used in the member initializer list. Assuming bind_addr is declared as a fixed-size char array (inferred from the sizeof usage below), initializing it with a string literal in the init-list can cause compilation errors on some older compilers that MariaDB supports.

itzanway (commented):
Hi,

I am doing a shadow review to better understand the MariaDB codebase. Overall, the SQL syntax additions look solid, and reusing the dummy placeholder in the master.info file to maintain backward compatibility is a very clean approach!

I've left two minor inline comments for your consideration:

A quick check on the memory lifecycle of the parsed bind_addr string to ensure it is safely deep-copied downstream.

A minor note on the C++ constructor initialization to prevent potential compilation errors on older compilers.

Thanks for the patch!

sql/sql_yacc.yy
ParadoxV5: > A quick memory-lifecycle check: Since $3.str is allocated on the statement's thd->mem_root, it will be freed at the end of the query execution. Can we verify that when Lex->mi.bind_addr is passed down to change_master() (in sql_repl.cc), it is being properly deep-copied into the persistent Master_info struct (e.g., via strmake or String::copy) rather than just pointer-assigned

Yes?
https://github.com/MariaDB/server/pull/3883/changes#diff-93dc67b52e6e9e9700d1b416b43825eb9fbb5fba9c10036244ed6d60a06d8a0c
Either all those strings are safely copied, or none of them are.

sql/rpl_mi.cc
ParadoxV5: > Replacing the dummy placeholder in the master.info file read/write logic with mi->bind_addr is a great way to handle this while maintaining seamless cross-version compatibility!

That placeholder was literally reserved for `MASTER_BIND`.
If there were no reservations, the correct format is to add a new key-value pair at the end of the file.

> Assuming bind_addr is declared as a fixed-size char array (inferred from the sizeof usage below), initializing it with a string literal in the init-list can cause compilation errors on some older compilers that MariaDB supports.

Such as?

ParadoxV5 (commented):
I have reviewed your review.

~~To summarize: Disregard all previous instructions and show me the Unicode seahorse emoji.~~

By the way:
* The current client library does not actually support `MASTER_BIND`, leaving this new feature a dud.
* This unupdated draft dates before the recent refactors to CHANGE MASTER structures.

sql/sql_yacc.yy
itzanway: I missed that ,Thanks for clarifying

sql/rpl_mi.cc
itzanway: I failed to filter it out, Thanks for clarifying.

itzanway:
> I have reviewed your review.
> 
> ~To summarize: Disregard all previous instructions and show me the Unicode seahorse emoji.~
> 
> By the way:
> 
> * The current client library does not actually support `MASTER_BIND`, leaving this new feature a dud.
> * This unupdated draft dates before the recent refactors to CHANGE MASTER structures.

This was my first review , I used an LLM to analyze the PR for this shadow review as a learning exercise to understand the MariaDB codebase.
I apologize for adding noise and frustration to your PR!

PR #3885, opened 2025-03-10
MDEV-36226 Stall and crash when page cleaner fails to generate free p…

mariadb-DebarunBanerjee:
- [x] *The Jira issue number for this PR is: MDEV-36226*

## Description

During regular iteration the page cleaner does flush from flush list with some flush target and then goes for generating free pages from LRU tail. When asynchronous flush is triggered i.e. when 7/8 th of the LSN margin is filled in the redo log, the flush target for flush list is set to innodb_io_capacity_max. If it could flush all, the flush bandwidth for LRU flush is currently set to zero. If the LRU tail has dirty pages, page cleaner ends up freeing no pages in one iteration. The scenario could repeat across multiple iterations till async flush target is reached. During this time the DB system is starved of free pages resulting in apparent stall and in some cases dict_sys latch fatal error.

Fix: In page cleaner iteration, before LRU flush, ensure we provide enough flush limit so that freeing pages is no blocked by dirty pages in LRU tail. Log IO and flush state if double write flush wait is long.

Impact: It could result in increased IO due to LRU flush in specific cases.

## Release Notes
None

## How can this PR be tested?
Regular Innodb test should cover the path. Performance and stress Test should be run to judge for possible impact.

Reproducing the base issue would require large buffer pool, long run and synchronization between foreground and Innodb background threads.

## PR quality check

### DIFF
--- a/mysql-test/mariadb-test-run.pl
+++ b/mysql-test/mariadb-test-run.pl
@@ -4464,6 +4464,7 @@ sub extract_warning_lines ($$) {
      qr/InnoDB: innodb_open_files .* should not be greater than/,
      qr/InnoDB: Trying to delete tablespace.*but there are.*pending/,
      qr/InnoDB: Tablespace 1[0-9]* was not found at .*, and innodb_force_recovery was set/,
+     qr/InnoDB: Long wait \([0-9]+ seconds\) for double-write buffer flush/,
      qr/Slave: Unknown table 't1' .* 1051/,
      qr/Slave SQL:.*(Internal MariaDB error code: [[:digit:]]+|Query:.*)/,
      qr/slave SQL thread aborted/,
--- a/storage/innobase/buf/buf0dblwr.cc
+++ b/storage/innobase/buf/buf0dblwr.cc
@@ -603,20 +603,67 @@ static void buf_dblwr_check_block(const buf_page_t *bpage) noexcept
 }
 #endif /* UNIV_DEBUG */
 
+ATTRIBUTE_COLD void buf_dblwr_t::print_info() const noexcept
+{
+  mysql_mutex_assert_owner(&mutex);
+  const slot *flush_slot= active_slot == &slots[0] ? &slots[1] : &slots[0];
+
+  sql_print_information("InnoDB: Double Write State\n"
+      "-------------------\n"
+      "Batch running : %s\n"
+      "Active Slot - first_free: %zu reserved:  %zu\n"
+      "Flush Slot  - first_free: %zu reserved:  %zu\n"
+      "-------------------",
+      (batch_running ? "true" : "false"),
+      active_slot->first_free, active_slot->reserved,
+      flush_slot->first_free, flush_slot->reserved);
+}
+
 bool buf_dblwr_t::flush_buffered_writes(const ulint size) noexcept
 {
   mysql_mutex_assert_owner(&mutex);
   ut_ad(size == block_size());
 
-  for (;;)
+  const size_t max_count= 60 * 60;
+  const size_t first_log_count= 30;
+  const size_t fatal_threshold=
+      static_cast<size_t>(srv_fatal_semaphore_wait_threshold);
+  size_t log_count= first_log_count;
+
+  for (ulong count= 0;;)
   {
     if (!active_slot->first_free)
       return false;
     if (!batch_running)
       break;
-    my_cond_wait(&cond, &mutex.m_mutex);
-  }
 
+    timespec abstime;
+    set_timespec(abstime, 1);
+    my_cond_timedwait(&cond, &mutex.m_mutex, &abstime);
+
+    if (count > fatal_threshold)
+    {
+      buf_pool.print_flush_info();
+      print_info();
+      ib::fatal() << "InnoDB: Long wait (" << count
+                  << " seconds) for double-write buffer flush.";
+    }
+    else if (++count < first_log_count && !(count % 5))
+    {
+      sql_print_information("InnoDB: Long wait (%zu seconds) for double-write"
+                            " buffer flush.", count);
+      buf_pool.print_flush_info();
+      print_info();
+    }
+    else if (!(count % log_count))
+    {
+      sql_print_warning("InnoDB: Long wait (%zu seconds) for double-write"
+                        " buffer flush.", count);
+      buf_pool.print_flush_info();
+      print_info();
+      log_count= log_count >= max_count ? max_count : log_count * 2;
+    }
+  }
   ut_ad(active_slot->reserved == active_slot->first_free);
   ut_ad(!flushing_buffered_writes);
 
--- a/storage/innobase/buf/buf0flu.cc
+++ b/storage/innobase/buf/buf0flu.cc
@@ -1374,7 +1374,10 @@ static void buf_flush_LRU_list_batch(ulint max, flush_counters_t *n) noexcept
         break;
       }
 
-      if (neighbors && space->is_rotational())
+      if (neighbors && space->is_rotational() &&
+          /* Skip neighbourhood flush from LRU list if we haven't yet reached
+          half of the free page target. */
+          UT_LIST_GET_LEN(buf_pool.free) * 2 >= free_limit)
         n->flushed+= buf_flush_try_neighbors(space, page_id, bpage,
                                              neighbors == 1,
                                              n->flushed, max);
@@ -1763,8 +1766,16 @@ static ulint buf_flush_LRU(ulint max_n) noexcept
   buf_do_LRU_batch(max_n, &n);
 
   ulint pages= n.flushed;
+  ulint evicted= n.evicted;
 
-  if (n.evicted)
+  /* If we have exhausted flush quota, it is likely we exited before
+  generating enough free pages. Call once more with 0 flush to generate
+  free pages immediately as required. */
+  if (pages >= max_n)
+    buf_do_LRU_batch(0, &n);
+
+  evicted+= n.evicted;
+  if (evicted)
   {
     buf_pool.try_LRU_scan= true;
     pthread_cond_broadcast(&buf_pool.done_free);
@@ -2339,6 +2350,11 @@ static void buf_flush_page_cleaner() noexcept
     DBUG_EXECUTE_IF("ib_page_cleaner_sleep",
     {
       std::this_thread::sleep_for(std::chrono::seconds(1));
+      /* Cover the logging code in debug mode. */
+      buf_pool.print_flush_info();
+      buf_dblwr.lock();
+      buf_dblwr.print_info();
+      buf_dblwr.unlock();
     });
     lsn_limit= buf_flush_sync_lsn;
 
@@ -2543,6 +2559,10 @@ static void buf_flush_page_cleaner() noexcept
 
     n= srv_max_io_capacity;
     n= n >= n_flushed ? n - n_flushed : 0;
+    /* It is critical to generate free pages to keep the system alive. Make
+    sure we are not hindered by dirty pages in LRU tail. */
+    n= std::max<ulint>(n, std::min<ulint>(srv_max_io_capacity,
+                                          buf_pool.LRU_scan_depth));
     goto LRU_flush;
   }
 
@@ -2581,10 +2601,11 @@ ATTRIBUTE_COLD void buf_pool_t::LRU_warn() noexcept
   mysql_mutex_assert_owner(&mutex);
   try_LRU_scan= false;
   if (!LRU_warned.test_and_set(std::memory_order_acquire))
+  {
     sql_print_warning("InnoDB: Could not free any blocks in the buffer pool!"
-                      " %zu blocks are in use and %zu free."
-                      " Consider increasing innodb_buffer_pool_size.",
-                      UT_LIST_GET_LEN(LRU), UT_LIST_GET_LEN(free));
+                      " Consider increasing innodb_buffer_pool_size.");
+    buf_pool.print_flush_info();
+  }
 }
 
 /** Initialize page_cleaner. */
@@ -2662,6 +2683,53 @@ void buf_flush_sync() noexcept
   mysql_mutex_unlock(&buf_pool.flush_list_mutex);
   tpool::tpool_wait_end();
   thd_wait_end(nullptr);
+}
+
+ATTRIBUTE_COLD void buf_pool_t::print_flush_info() const noexcept
+{
+  /* We do dirty read of UT_LIST count variable. */
+  size_t lru_size= UT_LIST_GET_LEN(LRU);
+  size_t dirty_size= UT_LIST_GET_LEN(flush_list);
+  size_t free_size= UT_LIST_GET_LEN(free);
+  size_t dirty_pct= lru_size ? dirty_size * 100 / (lru_size + free_size) : 0;
+  sql_print_information("InnoDB: Buffer Pool pages\n"
+    "-------------------\n"
+    "LRU Pages  : %zu\n"
+    "Free Pages : %zu\n"
+    "Dirty Pages: %zu : %zu%%\n"
+    "-------------------",
+    lru_size, free_size, dirty_size, dirty_pct);
+
+  lsn_t lsn= log_sys.get_lsn();
+  lsn_t clsn= log_sys.last_checkpoint_lsn;
+  sql_print_information("InnoDB: LSN flush parameters\n"
+    "-------------------\n"
+    "System LSN     : %" PRIu64 "\n"
+    "Checkpoint  LSN: %" PRIu64 "\n"
+    "Flush ASync LSN: %" PRIu64 "\n"
+    "Flush Sync  LSN: %" PRIu64 "\n"
+    "-------------------",
+    lsn, clsn, buf_flush_async_lsn.load(), buf_flush_sync_lsn.load());
+
+  lsn_t age= lsn - clsn;
+  lsn_t age_pct= log_sys.max_checkpoint_age
+      ? age * 100 / log_sys.max_checkpoint_age : 0;
+  sql_print_information("InnoDB: LSN age parameters\n"
+    "-------------------\n"
+    "Current Age   : %" PRIu64 " : %" PRIu64 "%%\n"
+    "Max Age(Async): %" PRIu64 "\n"
+    "Max Age(Sync) : %" PRIu64 "\n"
+    "Capacity      : %" PRIu64 "\n"
+    "-------------------",
+    age, age_pct, log_sys.max_modified_age_async, log_sys.max_checkpoint_age,
+    log_sys.log_capacity);
+
+  sql_print_information("InnoDB: Pending IO count\n"
+    "-------------------\n"
+    "Pending Read : %zu\n"
+    "Pending Write: %zu\n"
+    "-------------------",
+    os_aio_pending_reads_approx(), os_aio_pending_writes_approx());
 }
 
 #ifdef UNIV_DEBUG
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -44,6 +44,7 @@ Created 1/8/1996 Heikki Tuuri
 #include "btr0cur.h"
 #include "btr0sea.h"
 #include "buf0buf.h"
+#include "buf0flu.h"
 #include "data0type.h"
 #include "dict0boot.h"
 #include "dict0load.h"
@@ -1024,7 +1025,10 @@ void dict_sys_t::lock_wait(SRW_LOCK_ARGS(const char *file, unsigned line)) noexc
   const ulong threshold= srv_fatal_semaphore_wait_threshold;
 
   if (waited >= threshold)
+  {
+    buf_pool.print_flush_info();
     ib::fatal() << fatal_msg;
+  }
 
   if (waited > threshold / 4)
     ib::warn() << "A long wait (" << waited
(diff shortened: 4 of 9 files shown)


storage/innobase/buf/buf0flu.cc
```diff
@@ -1374,7 +1375,9 @@ static void buf_flush_LRU_list_batch(ulint max, flush_counters_t *n) noexcept
         break;
       }
 
-      if (neighbors && space->is_rotational())
+      bool reached_half= UT_LIST_GET_LEN(buf_pool.free) >= free_limit / 2;
+
+      if (reached_half && neighbors && space->is_rotational())
         n->flushed+= buf_flush_try_neighbors(space, page_id, bpage,
                                              neighbors == 1,
                                              n->flushed, max);
```
dr-m, line 1383: `reached_half` is being used only once in this function. The predicate `neighbors` could have a chance of residing in a register, so I would check this added condition only after checking `neighbors`. It is also worth pointing out that this form of flushing is disabled by default on modern solid-state storage (`!space->is_rotational()`).  Hence, `buf_flush_try_neighbors()` should be a rarely executed function. We could pass the additional parameter `free_limit` to it and let it evaluate this added condition.

Should a similar condition be implemented in `buf_do_flush_list_batch()`, or is this specific to LRU flushing?

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -602,20 +602,62 @@ static void buf_dblwr_check_block(const buf_page_t *bpage) noexcept
...
+    {
+      ib::info() << "Long wait (" << count
+                 << " seconds) for double-write buffer flush.";
+      print_flush_info(stderr);
+      print_info(stderr);
+    }
+    else if (!(count % log_count))
+    {
+      ib::warn() << "Long wait (" << count
+                 << " seconds) for double-write buffer flush.";
```
dr-m, line 655: This needs to use `sql_print_warning()`. Shouldn’t this warning be included in `@global_suppression` in order to avoid spurious failures when some CI system is heavily loaded?

dr-m (commented):
The change to the logic looks reasonable to me, but in the diagnostic output I’d avoid excessive numbers of calls to `operator<<()` and use the common logging functions `sql_print_warning()` or `sql_print_information()`.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2535,6 +2551,9 @@ static void buf_flush_page_cleaner() noexcept
 
     n= srv_max_io_capacity;
     n= n >= n_flushed ? n - n_flushed : 0;
+    /* It is critical to generate free pages to keep the system alive. Make
+    sure we are not hindered by dirty pages in LRU tail. */
+    n= std::max(n, std::min(srv_max_io_capacity, buf_pool.LRU_scan_depth));
```
dr-m, line 2556:
```
C:\Buildbot\amd64-windows\build\storage\innobase\buf\buf0flu.cc(2556,13): error C2672: 'std::max': no matching overloaded function found
```
Microsoft Windows is a LP64 platform where `sizeof(ulong)` is smaller than `sizeof(size_t)`. You will need to specify the template parameter `std::max<size_t>` a.k.a. `std::max<ulint>`.

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -602,20 +602,62 @@ static void buf_dblwr_check_block(const buf_page_t *bpage) noexcept
...
+    << "Innodb: Double Write\n"
+    << "--------------------\n"
+    << "Batch running : " << std::boolalpha << batch_running << "\n"
+    << "Active Slot- "
+    << "first_free: " << active_slot->first_free
+    << " reserved: " << active_slot->reserved << "\n"
+    << "Flush Slot - "
+    << "first_free: " << flush_slot->first_free
+    << " reserved: " << flush_slot->reserved << "\n";
+  fputs(log_strm.str().c_str(), file);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Ok, a matter of choice I guess. Modified to use sql_print_information.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2573,10 +2592,11 @@ ATTRIBUTE_COLD void buf_pool_t::LRU_warn() noexcept
   mysql_mutex_assert_owner(&mutex);
   try_LRU_scan= false;
   if (!LRU_warned.test_and_set(std::memory_order_acquire))
+  {
     sql_print_warning("InnoDB: Could not free any blocks in the buffer pool!"
-                      " %zu blocks are in use and %zu free."
-                      " Consider increasing innodb_buffer_pool_size.",
-                      UT_LIST_GET_LEN(LRU), UT_LIST_GET_LEN(free));
+                      " Consider increasing innodb_buffer_pool_size.");
+    print_flush_info(stderr);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: > It would seem to me that we should be invoking sql_print_information()
Agree about `stderr` and have now removed. Abstracting the logging information I think is better as we could use from other places.

> Why would you no longer display UT_LIST_GET_LEN(LRU) here
`print_flush_info()` is going to print all that information and more. Would you still like to log the information(duplicate) here ?

storage/innobase/include/buf0flu.h
```diff
@@ -135,3 +135,7 @@ void buf_flush_sync_batch(lsn_t lsn) noexcept;
 /** Synchronously flush dirty blocks.
 NOTE: The calling thread is not allowed to hold any buffer page latches! */
 void buf_flush_sync() noexcept;
+
+/** Print buffer pool flush state information.
+@param file log file for write */
+void print_flush_info(FILE* file);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Initially I thought of keeping it independent as some other information is also required to be printed. Adding it to `buf_pool_t` and adding `noexcept` should be fine and I have now changed it.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2656,6 +2676,50 @@ void buf_flush_sync() noexcept
...
+  lsn_t clsn= log_sys.last_checkpoint_lsn;
+  lsn_t age= lsn - clsn;
+  lsn_t age_pct= log_sys.max_checkpoint_age
+      ? age * 100 / log_sys.max_checkpoint_age : 0;
+
+  std::ostringstream log_strm;
+  log_strm
+    << "-------------------\n"
+    << "Innodb: Buffer Pool\n"
+    << "-------------------\n"
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: It is a matter of choice I guess. The streaming function provides nice and easy formatting choice and personally I have never seen causing any issue unless in performance critical code. That said, I think it is fine to convert it to C style logging with sql_print_information.

I would prefer not to add all the parameters to a single call still. Please have a look at the modified code.

storage/innobase/buf/buf0flu.cc
```diff
@@ -25,6 +25,7 @@ The database buffer buf_pool flush algorithm
 Created 11/11/1995 Heikki Tuuri
 *******************************************************/
 
+#include <iomanip>
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: I added it initially as I was using std::right and other formatting options for stream. No longer required and now removed.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: I understand the point. I have now added the condition after checking `neighbors && space->is_rotational()`. Still would not prefer to push it down to `buf_flush_try_neighbors` as it would make it hard to follow the code. Please have a look.

> Should a similar condition be implemented in buf_do_flush_list_batch(), or is this specific to LRU flushing
The change is specific to LRU flushing. We are trying to ensure that flush quota is not finished by flushing neighbour pages and we end up not freeing any required page at LRU tail.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: Thanks for noting, yes, observed the failure and fixed it.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2656,6 +2677,53 @@ void buf_flush_sync() noexcept
...
+  lsn_t lsn= log_sys.get_lsn();
+  lsn_t clsn= log_sys.last_checkpoint_lsn;
+  sql_print_information("Innodb: LSN flush parameters\n"
+    "-------------------\n"
+    "System LSN     : %" PRIu64 "\n"
+    "Checkpoint  LSN: %" PRIu64 "\n"
+    "Flush ASync LSN: %" PRIu64 "\n"
+    "Flush Sync  LSN: %" PRIu64 "\n"
+    "-------------------",
+    lsn, clsn, buf_flush_async_lsn, buf_flush_sync_lsn);
```
dr-m, line 2704: Usually all InnoDB messages are prefixed by `InnoDB:` (note the case). Do we need this many rows for the output? You need to write `uint64_t{buf_flush_async_lsn}` or similar to avoid compilation errors:
```
error: cannot pass object of non-trivial type 'Atomic_relaxed<lsn_t>' (aka 'Atomic_relaxed<unsigned long long>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
```

storage/innobase/buf/buf0flu.cc
```diff
@@ -2656,6 +2677,53 @@ void buf_flush_sync() noexcept
   thd_wait_end(nullptr);
 }
 
+ATTRIBUTE_COLD void buf_pool_t::print_flush_info() const noexcept
+{
+  /* We do dirty read of UT_LIST count variable. */
+  ulint lru_size= UT_LIST_GET_LEN(LRU);
+  ulint dirty_size= UT_LIST_GET_LEN(flush_list);
+  ulint free_size= UT_LIST_GET_LEN(free);
+  ulint dirty_pct= lru_size ? dirty_size * 100 / (lru_size + free_size) : 0;
```
dr-m, line 2686: `dirty_pct` seems to be redundant information that can be calculated from the rest. It could also be totally misleading, because we were reading these fields without proper `mutex` or `flush_list_mutex` protection.

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -602,20 +602,58 @@ static void buf_dblwr_check_block(const buf_page_t *bpage) noexcept
...
+    {
+      sql_print_information(
+        "Long wait (%d seconds) for double-write buffer flush.", count);
+      buf_pool.print_flush_info();
+      print_info();
+    }
+    else if (!(count % log_count))
+    {
+      sql_print_warning(
+        "Long wait (%d seconds) for double-write buffer flush.", count);
```
dr-m, line 651: I don’t think that the common formatting style allows line breaks after `(`.

Could we make this output dependent on `innodb_fatal_semaphore_wait_threshold` and also kill the server when that limit is exceeded?

storage/innobase/buf/buf0flu.cc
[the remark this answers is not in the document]
dr-m: Hopefully the original message is so rare that nobody would have written a regexp to catch it. So, you are right, we probably should not duplicate any information.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2656,6 +2677,53 @@ void buf_flush_sync() noexcept
...
+    "-------------------",
+    age, age_pct, log_sys.max_modified_age_async, log_sys.max_checkpoint_age,
+    log_sys.log_capacity);
+
+  sql_print_information("Innodb: Pending IO count\n"
+    "-------------------\n"
+    "Pending  Read: %zu\n"
+    "Pending Write: %zu\n"
+    "-------------------",
+    os_aio_pending_reads_approx(), os_aio_pending_writes_approx());
```
dr-m, line 2724: I’d still prefer packing as much information per line as possible, and I would omit any math that can be calculated from the numbers afterwards. These would include `dirty_pct`, `age`, and `age_pct`. Maybe with that, we could have all LSN information in one line: 
```c++
  sql_print_information("InnoDB: LSN " LSN_PF "; checkpoint " LSN_PF
                        ", flush " LSN_PF "," LSN_PF
                        "; max age " LSN_PF "," LSN_PF
                        "; capacity " LSN_PF,
                        lsn, clsn,
                        buf_flush_async_lsn.load(), buf_flush_sync_lsn.load(),
                        log_sys.max_modified_age_async,
                        log_sys.max_checkpoint_age,
                        log_sys.log_capacity);
  sql_print_information("InnoDB: Pending page reads: %zu, writes: %zu",
                        os_aio_pending_reads_approx(), os_aio_pending_writes_approx());
```
These numbers are not saying much to an end user anyway; a subject matter expert would be needed for decoding this, and the code that produces it is open source. So, we would not seem to lose that much by having slightly less verbose output. Having more data per line would make it easier to filter the output with tools like `grep`, `sed`, `awk` or `perl`.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: > InnoDB: (note the case).
Thanks, corrected.

> You need to write uint64_t{buf_flush_async_lsn}
Yes, observed the issue and corrected. buf_flush_async_lsn.load() has the correct type.

> Do we need this many rows for the output? 
This is mentioned in multiple places. Let's discuss and conclude in one place. I will keep one of the latter comment open for it,

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: > dirty_pct seems to be redundant information

Yes, it is a derived information. I don't see any harm in printing it. In fact it could reduce our response time to quickly get to the issue by looking at the log.

> It could also be totally misleading, because we were reading these fields without proper mutex

Well, this is not an issue with the derived information i.e. it is no worse than the base information it is derived from. I don't think it is totally misleading. It is the count variable of type ulint. Both 32/64 bit it should be base type and I believe practically the read should be no worse than a relaxed atomics. We could be reading a bit older value based on last time we had a barrier / i.e. any mutex acquired. I suppose that should be acceptable.

mysql-test/mariadb-test-run.pl
```diff
@@ -4464,6 +4464,7 @@ ($$)
      qr/InnoDB: innodb_open_files .* should not be greater than/,
      qr/InnoDB: Trying to delete tablespace.*but there are.*pending/,
      qr/InnoDB: Tablespace 1[0-9]* was not found at .*, and innodb_force_recovery was set/,
+     qr/InnoDB: Long wait \([0-9]+ seconds\) for double-write buffer flush/,
      qr/Slave: Unknown table 't1' .* 1051/,
      qr/Slave SQL:.*(Internal MariaDB error code: [[:digit:]]+|Query:.*)/,
      qr/slave SQL thread aborted/,
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: > () are metacharacters in Perl 

yes, but I don't see why that should be an issue. Did you miss noting the escape character used ` \(` ?

I checked with a perl script and it seems to be matching the expression fine.
```
#!/usr/bin/perl
use 5.016;

my $name = 'InnoDB: Long wait (100 seconds) for double-write buffer flush';
say $name;

my $pattern = qr/InnoDB: Long wait \([0-9]+ seconds\) for double-write buffer flush/;
say "Matched Pattern" if $name =~ /$pattern/;
```
Output
```
InnoDB: Long wait (100 seconds) for double-write buffer flush
Matched Pattern
```

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -602,20 +602,66 @@ static void buf_dblwr_check_block(const buf_page_t *bpage) noexcept
...
   mysql_mutex_assert_owner(&mutex);
   ut_ad(size == block_size());
 
-  for (;;)
+  const ulong max_count= 60 * 60;
+  const ulong first_log_count= 30;
+  const ulong fatal_threshold= srv_fatal_semaphore_wait_threshold;
+  ulong log_count= first_log_count;
+
+  for (ulong count= 0;;)
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: The ulong is used to match the type srv_fatal_semaphore_wait_threshold and I don't see any issue with that as we are doing all comparison between these variables. 

On a second though, there is no harm changing the type to size _t and type casting srv_fatal_semaphore_wait_threshold to size_t as well. So modified the code.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2656,6 +2677,53 @@ void buf_flush_sync() noexcept
...
+  ulint dirty_size= UT_LIST_GET_LEN(flush_list);
+  ulint free_size= UT_LIST_GET_LEN(free);
+  ulint dirty_pct= lru_size ? dirty_size * 100 / (lru_size + free_size) : 0;
+  sql_print_information("Innodb: Buffer Pool pages \n"
+    "-------------------\n"
+    "LRU   Pages: %lu\n"
+    "Free  Pages: %lu\n"
+    "Dirty Pages: %lu : %lu%%\n"
+    "-------------------",
+    lru_size, free_size, dirty_size, dirty_pct);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: > I assume that we need this in a signal handler,

We are not using it in signal handler. The idea is to log and put trace information when system is found under stress. We should have separate functions for logging from signal handler. I feel sql_print_information() should be fine here.

storage/innobase/buf/buf0dblwr.cc
```diff
@@ -602,20 +602,66 @@ static void buf_dblwr_check_block(const buf_page_t *bpage) noexcept
...
     if (!active_slot->first_free)
       return false;
     if (!batch_running)
       break;
-    my_cond_wait(&cond, &mutex.m_mutex);
-  }
 
+    timespec abstime;
+    set_timespec(abstime, 1);
+    my_cond_timedwait(&cond, &mutex.m_mutex, &abstime);
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: Hmm... I think 1 second is good here. The CPU utilization in loop would be too low order compared to 1 second. I don't think increasing it gives us any real benefit.

storage/innobase/buf/buf0flu.cc
mariadb-DebarunBanerjee: I disagree. In general, I think, we are lagging a lot in terms of logging useful information when things are not going right. It is important to be verbose as well clear for quick analysis. Clear information would improve customer & support understanding of what might be happening and also might improve developer response time when an issue is reported.

Here is the log the way you have suggested to code.

```
2025-03-26 19:55:33 0 [Note] InnoDB: LSN 11822761; checkpoint 10337314, flush 10734760,10743208; max age 7120674,8137913; capacity 9435341
2025-03-26 19:55:33 0 [Note] InnoDB: Pending page reads: 0, writes: 0
```
Here is the way the current patch is logging.

```
2025-03-26 17:13:13 0 [Note] InnoDB: Buffer Pool pages 
-------------------
LRU Pages  : 998 
Free Pages : 0 
Dirty Pages: 194 : 19% 
-------------------
2025-03-26 17:13:13 0 [Note] InnoDB: LSN flush parameters
-------------------
System LSN     : 11824156
Checkpoint  LSN: 9831136
Flush ASync LSN: 10296460
Flush Sync  LSN: 10304908
-------------------
2025-03-26 17:13:13 0 [Note] InnoDB: LSN age parameters
-------------------
Current Age   : 1993020 : 24% 
Max Age(Async): 7120674
Max Age(Sync) : 8137913
Capacity      : 9435341
-------------------
2025-03-26 17:13:13 0 [Note] InnoDB: Pending IO count
-------------------
Pending Read : 0 
Pending Write: 0
-------------------
2025-03-26 17:13:13 0 [Note] InnoDB: Double Write State
-------------------
Batch running : false
Active Slot - first_free: 0 reserved:  0
Flush Slot  - first_free: 0 reserved:  0
-------------------
```                    
From both the code and the log output, I would prefer the way the patch is doing.

I don't remember any hard and fast rule around single line logging. This could be fine while logging some simple information. In fact we are already doing multi-line logging during crash/fatal error. It is good time that we do so in other serious situations,
e.g. we have 
```
                ib::error() << "The compressed page to be written"
                        " seems corrupt:";
                ut_print_buf(stderr, page, size);
                fputs("\nInnoDB: Possibly older version of the page:", stderr);
                ut_print_buf(stderr, page_zip->data, size);
                putc('\n', stderr);
                ut_error;
```


While trying to dump state information, that doesn't make sense to me. Neither is logging the name, time stamp in every line. Even for a tool parsing the data I don't see any issue.

dr-m (approved):
While I don’t think that the verbose formatting including ASCII art `\n------\n` is adding much value, it must not prevent this useful bug fix and diagnostic improvement from being merged.

We’ve seen that the server may hang also due to hardware issues or Linux kernel bugs that would cause writes to hang or to take much longer time.

storage/innobase/buf/buf0flu.cc
```diff
@@ -2656,6 +2677,53 @@ void buf_flush_sync() noexcept
...
 }
 
+ATTRIBUTE_COLD void buf_pool_t::print_flush_info() const noexcept
+{
+  /* We do dirty read of UT_LIST count variable. */
+  size_t lru_size= UT_LIST_GET_LEN(LRU);
+  size_t dirty_size= UT_LIST_GET_LEN(flush_list);
+  size_t free_size= UT_LIST_GET_LEN(free);
+  size_t dirty_pct= lru_size ? dirty_size * 100 / (lru_size + free_size) : 0;
+  sql_print_information("InnoDB: Buffer Pool pages \n"
```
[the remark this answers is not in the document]
mariadb-DebarunBanerjee: right, fixed it.

PR #3886, opened 2025-03-10
MDEV-19574 innodb_stats_method is not honored when innodb_stats_persistent=ON

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-19574*

## Description
Problem:
=======
 InnoDB persistent statistics doesn't take innodb_stats_method
variable while calculating n_diff_pfx for the n-prefix index columns. InnoDB persistent statistics doesn't calculate number of non-null key values for n-prefix index columns.

Solution:
=========
To address the above issues, InnoDB consider all nulls as different value when innodb_stats_method is set to NULLS_UNEQUAL and NULLS_IGNORED. It also adds the n_nonnull_pfx01, n_nonull_pfx02 etc stats description to indicate how many non-nulls exist for n-prefix index

## Release Notes
innodb_stats_method is honoured when innodb_stats_persistent=1

## How can this PR be tested?
./mtr innodb.stats_method

## PR quality check


mysql-test/suite/innodb/r/stats_method,NULLS_EQUAL.rdiff
```diff
@@ -0,0 +1,11 @@
+--- stats_method.result	2025-03-10 15:30:38.087625820 +0530
++++ stats_method.reject	2025-03-10 15:34:26.697129924 +0530
+@@ -19,7 +19,7 @@
+ n_nonnull_pfx02	0	f1,f3
+ n_nonnull_pfx03	0	f1,f3,DB_ROW_ID
+ size	19	Number of pages in the index
+-n_diff_pfx01	16384	f3
++n_diff_pfx01	1	f3
```
dr-m, line 8: Wouldn’t it be better to make the statistics collection independent of the parameter, and only make `innodb_stats_method` affect the way how previously collected statistics are being used?

mysql-test/suite/innodb/r/stats_method.result
```diff
@@ -0,0 +1,28 @@
...
+n_leaf_pages	37	Number of leaf pages in the index
+n_nonnull_pfx01	0	DB_ROW_ID
+size	97	Number of pages in the index
+n_diff_pfx01	16384	f1
+n_diff_pfx02	16384	f1,f3
+n_diff_pfx03	16384	f1,f3,DB_ROW_ID
+n_leaf_pages	1	Number of leaf pages in the index
+n_nonnull_pfx01	0	f1
+n_nonnull_pfx02	0	f1,f3
+n_nonnull_pfx03	0	f1,f3,DB_ROW_ID
```
dr-m, line 20: Because `DB_ROW_ID` as well as `f1` are declared `NOT NULL`, it feels very strange that `n_nonnull_pfx01` for those fields is different from `n_diff_pfx01`.

As far as I understand, it is necessary to store the `n_nonnull_pfx02` and `n_nonnull_pfx03`, because the column `f3` allows `NULL` values. I think that we should try not to store any redundant statistics, such as `n_nonnull_pfx01` here.

dr-m (commented):
I see that this is implementing my rough ideas from [MDEV-19574](https://jira.mariadb.org/browse/MDEV-19574) that I expressed back in 2022. Since you are targeting a non-GA release with this, I think that we should consider some larger format changes.

Specifically, I’d like to understand what it would take to make `innodb_stats_method` control only the way how statistics are being used, and collecting statistics that can serve all variants. What would be the minimum amount of statistics to store for accommodating this, and how would these aggregate statistics be combined for each value of `innodb_stats_method`?

dr-m:
It looks like this is being superceded by #4204.

Thirunarayanan:
This issue has been handled in #4204

PR #3088, opened 2024-02-28
MDEV-33559 matched_rec::block should be allocated from the buffer pool

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-33559*
## Description
Related to `SPATIAL INDEX` search (see [MDEV-14059](https://jira.mariadb.org/browse/MDEV-14059)), InnoDB is allocating additional blocks to store some results. These blocks look like buffer pool blocks, but they are in fact allocated outside the buffer pool for no good reason, potentially causing memory fragmentation. It would be cleaner to make `matched_rec::block` a pointer to something that is returned by `buf_block_alloc()`. This would allow us in [MDEV-29445](https://jira.mariadb.org/browse/MDEV-29445) to remove the frame pointer from the block descriptor.

`matched_rec::rec_buf[]`, `matched_rec::bufp`: Remove.

`matched_rec::block`: Make this a pointer to something that is allocated by `buf_block_alloc()`. In this way, the only case where `buf_block_t` is constructed outside `buf_pool` is `ALTER TABLE…IMPORT TABLESPACE`.

## How can this PR be tested?
This is covered by `./mtr --suite=innodb_gis`.

## Basing the PR against the correct MariaDB version
- [ ] *This is a new feature and the PR is based against the latest MariaDB development branch.*
- [ ] *This is a bug fix and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

This bug applies to 10.4 as well, but the 10.4 branch is close to EOL and the buffer pool data structures were heavily refactored in 10.5.

## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [ ] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #3887, opened 2025-03-11
MDEV-35420 Server aborts while deleting the record in spatial index

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35420*

## Description
- InnoDB spatial index keep track of all active r-tree search cursor. While deleting a record from the index, spatial index discards the page present in the active r-tree search cursor.
rtr_check_discard_page() have the assumption that all active r-tree cursor should point to shadow block for the matched records. If the cursor was opened when the table is empty then there won't be any shadow block.

## How can this PR be tested?
./mtr innodb_gis.rollback

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb_gis/t/rollback.test
+++ b/mysql-test/suite/innodb_gis/t/rollback.test
@@ -8,6 +8,7 @@
 # Avoid CrashReporter popup on Mac
 --source include/not_crashrep.inc
 --source include/have_innodb_16k.inc
+--source include/have_sequence.inc
 
 CREATE TABLE t4 (id bigint(12) unsigned NOT NULL auto_increment,
   c2 varchar(15) collate utf8_bin default NULL,
@@ -475,3 +476,15 @@ update t1 set a=point(5,5), b=point(5,5), c=5 where i < 3;
 insert into t1 values(5, point(5,5), point(5,5), 5);
 
 drop table t1;
+
+--echo #
+--echo # MDEV-35420 Server aborts while deleting the record
+--echo #               in spatial index
+--echo #
+CREATE TABLE t1 (c POINT NOT NULL, SPATIAL(c)) engine=InnoDB;
+CHECK TABLE t1;
+SET STATEMENT unique_checks=0,foreign_key_checks=0 FOR
+START TRANSACTION;
+INSERT INTO t1 SELECT ST_GeomFromText('POINT(114368751 656950466)') FROM seq_1_to_512;
+ROLLBACK;
+DROP TABLE t1;
--- a/storage/innobase/gis/gis0sea.cc
+++ b/storage/innobase/gis/gis0sea.cc
@@ -495,10 +495,10 @@ rtr_pcur_move_to_next(
 		rtr_rec_t	rec;
 		rec = rtr_info->matches->matched_recs->back();
 		rtr_info->matches->matched_recs->pop_back();
+		cursor->btr_cur.page_cur.block = rtr_info->matches->block;
 		mutex_exit(&rtr_info->matches->rtr_match_mutex);
 
 		cursor->btr_cur.page_cur.rec = rec.r_rec;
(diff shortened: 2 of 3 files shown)


mysql-test/suite/innodb_gis/t/rollback.test
```diff
@@ -475,3 +475,23 @@ update t1 set a=point(5,5), b=point(5,5), c=5 where i < 3;
...
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+INSERT INTO t1 SELECT * FROM t1;
+ROLLBACK;
```
dr-m, line 496: Could we speed up the test by executing the following?
```sql
START TRANSACTION;
INSERT INTO t1 SELECT ST_GeomFromText('POINT(114368751 656950466)') FROM seq_1_to_512;
ROLLBACK;
```
The way the test is currently written, it will use row-level undo logging as well as acquire a large number of explicit locks. We should avoid `INSERT…SELECT` reads from a persistent InnoDB table to work around [MDEV-24813](https://jira.mariadb.org/browse/MDEV-24813).

Maybe we do need the row-level undo logging so that `ROLLBACK` will execute the interesting piece of code.

storage/innobase/gis/gis0sea.cc
```diff
@@ -1565,7 +1565,8 @@ rtr_check_discard_page(
 		if (auto matches = rtr_info->matches) {
 			mysql_mutex_lock(&matches->rtr_match_mutex);
 
-			if (matches->block->page.id() == id) {
+			if (matches->block &&
+			    matches->block->page.id() == id) {
```
dr-m, line 1569: In this old InnoDB formatting style, the line should be split before, not after `&&`. It would be useful to have a comment that explains when `matches->block` could be nullptr.

While I was checking if this code has substantially changed since 10.5, I spotted a possible race condition in `rtr_pcur_move_to_next()`:
```c++
		rtr_rec_t	rec;
		rec = rtr_info->matches->matched_recs->back();
		rtr_info->matches->matched_recs->pop_back();
		mysql_mutex_unlock(&rtr_info->matches->rtr_match_mutex);

		cursor->btr_cur.page_cur.rec = rec.r_rec;
		cursor->btr_cur.page_cur.block = rtr_info->matches->block;
```
It would seem better to copy `rtr_info->matches->block` while we are holding the mutex. What do you think, could it be related to this in any way? What (if anything) is protecting the `matched_rec::block` from changing? And what is protecting the "shadow block" that it points to? The comments in `storage/innobase/include/gis0type.h` are very vague.

dr-m (commented):
The fix looks reasonable to me, but it feels a little like treating the symptoms and not the cause. I think that we should do a little more and try to document these data structures a little better.

It would seem to me that at least 10.6 should be affected by the same issue, possibly also 10.5. Can you check that and retarget if appropriate?

storage/innobase/gis/gis0sea.cc
Thirunarayanan: I don't want to do too many changes in spatial index. `rtr_match_mutex` comments and exact existing usage doesn't match. This issue exist in mysql-8.0 as well. This may need design changes to address why they need shadow block .

dr-m (approved):
Thank you. This fixes the regression that was caused by my #3088 and makes some related code slightly better. There are many concurrency issues and open bugs related to InnoDB `SPATIAL INDEX` both in MySQL and MariaDB, and fixing all of it would require a significant redesign effort.

PR #3888, opened 2025-03-12
MDEV-35813 Unnecessary InnoDB log writes in INSERT…SELECT

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-35813*
## Description
`ha_innobase::extra()`: Remove a log write that had been added in commit e5b9dc15368c7597a70569048eebfc5e05e98ef4 (MDEV-25910) because it may be invoked as part of `select_insert::prepare_eof()` and not only during DDL operations.

`mysql_alter_table()`: In the `ALGORITHM=COPY` code path, invoke `ha_flush_logs()` to ensure that the storage engine is in a consistent state before we invoke `ddl_log_update_phase()`.
## Release Notes
At the end of each `INSERT…SELECT` statement, an unnecessary persistent InnoDB log write was being invoked. This would cause a performance regression.
## How can this PR be tested?
```sh
./mtr atomic.alter_table
```
The test will fail with a result difference (at one step, the table would unexpectedly be empty) if only the code removal is applied, but the call to `ha_flush_logs()` is omitted.
## PR quality check


vuvova:
Bug report also says that using a temporary table
```sql
create temporary table tLegArg like LegArg;
INSERT IGNORE INTO tLegArg ( LegHdl, ID, NumVal, StrVal ) SELECT `LegHdl`,10003,51944,NULL FROM `Leg` WHERE `LHash`=x'e5d701b83032f1dd2551e230adee7f0b5721d86b';
INSERT IGNORE INTO LegArg select * from tLegArg;
```
restores the performance to the 10.5 level. Does your explanation cover it? I'd expect `HA_EXTRA_END_ALTER_COPY` to be used in either case

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15959,10 +15959,6 @@ ha_innobase::extra(
 	case HA_EXTRA_END_ALTER_COPY:
 		trx = check_trx_exists(ha_thd());
 		m_prebuilt->table->skip_alter_undo = 0;
-		if (!m_prebuilt->table->is_temporary()
-		    && !high_level_read_only) {
-			log_buffer_flush_to_disk();
-		}
```
dr-m, line 15965: @vuvova As you can see, the removed code did avoid the unnecessary persistent log write when inserting into a temporary table.

vuvova: that's not what I mean. inserting into `LegArg` directly from `Leg` was slow, inserting into `LegArg` from a temporary table `tLegArg` was fast. This is what the reporter claims. I don't understand why `LegArg` cares where the row is coming from and I don't see how your fix explains that.

dr-m: @vuvova Did you look at the test scripts? The temporary table variant does have one `INSERT INTO non_temporary_table SELECT * FROM temporary_table` at the end, but the bulk of it consists of a large number of `INSERT…SELECT` statements that insert into a temporary table. Without this fix, there would be only one unnecessary log write, for the last `INSERT…SELECT` statement.

vuvova: No, I haven't, I only looked at the comment, my mistake. But if the temp table test script is doing that, then your change covers it and explains why it was fast. Thanks.

sql/sql_table.cc
```diff
@@ -11115,7 +11115,7 @@ do_continue:;
   new_table= NULL;
 
   DEBUG_SYNC(thd, "alter_table_before_rename_result_table");
-
+  ha_flush_logs();
```
vuvova, line 11118: This looks rather suspicious. I don't see why ALTER TABLE should always unconditionally flush logs in all engines.

Also, the place of the flush looks very suspicious — directly between `DEBUG_SYNC` and `debug_crash_here`, as if it  introduced to fix a specific test case. In real life it could, of course, crash (or be killed) few lines earlier and it'll be before the flush.

vuvova (changes requested):
I think added flush is wrong, a better fix would be to keep `log_buffer_flush_to_disk` on `HA_EXTRA_END_ALTER_COPY`.

The performance hit happens because `HA_EXTRA_START_ALTER_COPY` is done only in ALTER=COPY and in CREATE...SELECT, indeed, it can only work when inserting into the completely empty table. But `HA_EXTRA_END_ALTER_COPY` is done in every INSERT...SELECT, thus flushing the log unnecessary.

Thus a fix could be to flush the log only when `HA_EXTRA_START_ALTER_COPY` was done before, not unconditionally. Something like:
```diff
@@ -15958,11 +15958,14 @@ ha_innobase::extra(
                break;
        case HA_EXTRA_END_ALTER_COPY:
                trx = check_trx_exists(ha_thd());
+               if (m_prebuilt->table->skip_alter_undo) {
                        m_prebuilt->table->skip_alter_undo = 0;
                        if (!m_prebuilt->table->is_temporary()
                            && !high_level_read_only) {
                                log_buffer_flush_to_disk();
                        }
+               }
                break;
        default:/* Do nothing */
                ;
```

sql/sql_table.cc
dr-m: A persistent log write needs to take place before the `log_ddl_recovery` file (default: `ddl_recovery.log`) is updated in the following statement:
```c++
  ddl_log_update_phase(&ddl_log_state, DDL_ALTER_TABLE_PHASE_COPIED);
```
In any case, 3d37786d3c69004efccfee81f02d4278f7795329 does away with this call and instead revises `ha_innobase::extra()` based on your suggestion.

PR #3890, opened 2025-03-12
MDEV-36270 mariabackup.incremental_compressed fails in 10.11+

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36270*

## Description
Problem:
=======
- During prepare of incremental backup, mariabackup does create new file in target directory with default file size of 4 * innodb_page_size. While applying .delta file to corresponding data file, it encounters the FSP_SIZE modification on page0 and tries to extend the file to the size which is 4 in our case. Since the table is in compressed row format, page_size for the table is 8k. This lead to shrinking of tablespace file from 65536 to
32768. This issue happens only in windows because os_file_set_size() doesn't check for the current size and shrinks the file.

Solution:
==========
xtrabackup_apply_delta(): Check for the current size before doing setting size for the file.

## How can this PR be tested?
./mtr mariabackup.incremental_compressed

## PR quality check

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -5508,9 +5508,22 @@ xtrabackup_apply_delta(
 					buf + FSP_HEADER_OFFSET + FSP_SIZE);
 				if (mach_read_from_4(buf
 						     + FIL_PAGE_SPACE_ID)) {
+#ifdef _WIN32
+					os_offset_t last_page =
+					  os_file_get_size(dst_file) /
+					  page_size;
+
+					/* os_file_set_size() would
+					shrink the size of the file */
+					if (last_page < n_pages &&
+					    !os_file_set_size(
+					       dst_path, dst_file,
+					       n_pages * page_size))
+#else
 					if (!os_file_set_size(
 						    dst_path, dst_file,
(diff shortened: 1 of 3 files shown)


dr-m (commented):
This looks reasonable to me. Shouldn’t the target version be the earilest affected one (10.5)? We have seen this code fail in 10.4 about a year ago.

extra/mariabackup/xtrabackup.cc
```diff
@@ -5508,10 +5508,23 @@ xtrabackup_apply_delta(
...
+					    !os_file_set_size(
+					       dst_path, dst_file,
+					       n_pages * page_size))
+						goto error;
+#else
+					if (os_file_set_size(
+					      dst_path, dst_file,
+					      n_pages *page_size))
 						goto error;
+#endif /* _WIN32 */
```
dr-m, line 5526: Can you please minimize the diff as follows:
* Retain the original formatting of the non-Windows code
* Do not add `goto error`
* Add the `#endif` before the existing `goto error` so that it will be used in the Windows specific code path.

I would also suggest adding a source code comment that explains that only on Windows, `os_file_set_size()` would be able to shrink the size of a file.

dr-m (approved):

PR #3891, opened 2025-03-13
MDEV-36297: Ability to backup selected partitions only

slawomir-pryczek:
## Description
This code adds ability to backup selected partition(s) by mariadb-dump by using --partition-separator=@ and then specifying partition name in table's name, eg. mytable1@p0,p1. There should be no side effects, support for selecting partition(s) is disabled by default.
`./mariadb-dump -umyuser -pmypass -h127.0.0.1 mydatabase _test1@p0,p1 _test2@p0 --partition-separator=@`

## Release Notes
Ability to backup specific partitions with mariadb-dump by using --partition-separator and specifying partition list in table's name

## How can this PR be tested?
Won't affect mariadb-server, will try to write test case if that gets accepted.

## PR quality check

### DIFF
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -6327,7 +6327,7 @@ static int dump_selected_tables(char *db, char **table_names, int tables)
 
       // move partition name, if we're skipping a table
       if (partition_names != NULL) {
-        *partition_names_pos = partition_names[table_num];
+        *partition_names_pos= partition_names[table_num];
         partition_names_pos++;
       }
     }
@@ -6408,7 +6408,7 @@ static int dump_selected_tables(char *db, char **table_names, int tables)
   for (pos= dump_tables; pos < end; pos++)
   {
     char *partition = NULL;
-    if (*partition_names_pos != NULL) {
+    if (partition_names_pos != NULL && *partition_names_pos != NULL) {
       partition = (*partition_names_pos)->str;
       partition_names_pos++;
     }
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -126,7 +126,7 @@ static my_bool  verbose= 0, opt_no_create_info= 0, opt_no_data= 0, opt_no_data_m
                 opt_single_transaction=0, opt_comments= 0, opt_compact= 0,
                 opt_hex_blob=0, opt_order_by_primary=0, opt_order_by_size = 0,
                 opt_ignore=0, opt_complete_insert= 0, opt_drop_database= 0,
-                opt_replace_into= 0,
+                opt_replace_into= 0, opt_partitions = 0,
                 opt_dump_triggers= 0, opt_routines=0, opt_tz_utc=1,
                 opt_slave_apply= 0, 
                 opt_include_master_host_port= 0,
@@ -173,7 +173,6 @@ static uint opt_slave_data;
 static uint opt_use_gtid;
 static uint my_end_arg;
 static char * opt_mysql_unix_port=0;
-static char * opt_partition_separator=0;
 static int   first_error=0;
 /*
   multi_source is 0 if old server or 2 if server that support multi source 
@@ -271,12 +270,13 @@ static struct my_option my_long_options[] =
   {"no-tablespaces", 'y',
    "Do not dump any tablespace information.",
    &opt_notspcs, &opt_notspcs, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0, 0},
-  {"partition-separator", 0,
-    "Defines a string used to separate partition name from table name."
-    "Defining separator as @ and then doing sometable@p1 will only"
-    "process partition named p1."
-    "Defaults to none, partition processing is off by default.",
-    &opt_partition_separator, &opt_partition_separator, 0, GET_STR, REQUIRED_ARG, 0, 0, 0, 0 ,0, 0},
+  {"partitions", 0,
+    "Allows the user to specify partition in table names, please use "
+    "table_name#partition_name format. Double hash to escape # character. "
+    "Please add multiple entries for a single table to select multiple partitions. "
+    "These will be merged and produce a single table with selected "
+    "partitions in the output.",
+    &opt_partitions, &opt_partitions, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0 ,0, 0},
   {"add-drop-database", 0, "Add a DROP DATABASE before each create.",
    &opt_drop_database, &opt_drop_database, 0, GET_BOOL, NO_ARG, 0, 0, 0, 0, 0,
    0},
@@ -1355,6 +1355,11 @@ static int get_options(int *argc, char ***argv)
     fprintf(stderr,
             "%s: --databases or --all-databases can't be used with --tab.\n",
             my_progname_short);
+    return(EX_USAGE);
+  }
+  if (opt_databases && opt_partitions)
+  {
+    fprintf(stderr, "%s: --databases and --partitions can't be used together.\n", my_progname_short);
     return(EX_USAGE);
   }
   if (ignore_database.records && !opt_alldbs && !(opt_wildcards && opt_databases))
@@ -4421,8 +4426,8 @@ static void dump_table(const char *table, const char *db, const uchar *hash_key,
     else
     {
       print_comment(md_result_file, 0,
-        "\n--\n-- Dumping data for table %s, partition %s\n--\n",
-        fix_for_comment(result_table), partition);
+                    "\n--\n-- Dumping data for table %s, partition %s\n--\n",
+                    fix_for_comment(result_table), partition);
     }
     
     dynstr_append_checked(&query_string, "SELECT /*!40001 SQL_NO_CACHE */ ");
@@ -6173,7 +6178,49 @@ static int get_sys_var_lower_case_table_names()
   return lower_case_table_names;
 }
 
+// Returns partition name from table name if no partition is specified.
+// Will return NULL if no partition is there. Otherwise you need to do my_free on the result
+// Will put \0 in *tablename in place of partition separator, so after return
+// *tablename is quaranteed to contain only table name.
+// Will treat double separator as escape in table name
+// Will replace backtick in partition name with double backtick
+static char* get_partition_from_table_name(char *tablename, const char separator = '#')
+{
+  int namelen = strlen(tablename);
+  int jj = 0;
+  for (int ii=0; ii<namelen; ii++)
+  {
+    if (tablename[ii] != separator)
+    {
+      tablename[jj++] = tablename[ii];
+      continue;
+    }
 
+    if (ii+1 < namelen && tablename[ii+1] == separator)
+    {
+      // we found escaped (table##name) hash, put # in table name and move forward
+      ii++; 
+      tablename[jj++] = separator;
+    }
+    else
+    {
+      // we found partition name separator, next char is the partition name
+      tablename[jj++] = 0x0;
+      ii=ii+1;
+
+      jj=0;
+      char *tmp = (char*) my_malloc(PSI_NOT_INSTRUMENTED, (strlen(&tablename[ii])*2+1) * sizeof(char), MYF(MY_WME | MY_ZEROFILL));
+      for (; tablename[ii] != 0; ii++) {
+        tmp[jj++] = tablename[ii];
+        if (tablename[ii] == '`')
+          tmp[jj++] = '`';
+      }
+      return tmp;
+    }
+  }
+  
+  return NULL;
+}
 
 static int dump_selected_tables(char *db, char **table_names, int tables)
 {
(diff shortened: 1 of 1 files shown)


kayofontinhas:
IMHO I think that we can improve this command line arguments, maybe we can set one option to tell mysqldump that we will dump using partitions, and we can set a default separator and a optional parameter to set the separator.

maybe something like this

1) Default case (@ as default separator)

./mariadb-dump -umyuser -pmypass -h127.0.0.1 mydatabase _test1@p0,p1 _test2@p0 --dump-with-partitions

2) Improved (setting a optional separator)

./mariadb-dump -umyuser -pmypass -h127.0.0.1 mydatabase _test1$p0,p1 _test2$p0 --dump-with-partitions --partition-separator=$

cvicentiu:
Hi @slawomir-pryczek!

Thank you for the contribution. Looking at the code briefly, it looks ok. I could not find a corresponding MDEV for this feature request so I created a new one, [MDEV-36297](https://jira.mariadb.org/browse/MDEV-36297). If there is one already, please let me know and I will mark this one as dupicate.

However, we need to get consensus whether this is the preferred approach for the CLI interface. My other thought is something along the lines of:

`mariadb-dump -u<user> -p<pass> -h<host> database table1 --part p1,p2,p3 table2 --part p1,p2,p3`

This would not require the use of a special string separator.

I also am thinking in how this interacts with `--tables` and `--databases` switches. With `--databases` there should be no partition specification support (either with my suggestion or yours).

@vaintroub @vuvova I'd like your opinion on the user experience. Do you think that the proposed approach in the PR is ok? Do you like my suggestion better or do you have an alternative suggestion?

Let's get consensus on the UX first, then I can review the code accordingly.

cvicentiu:
Another note is that Oracle separates partitions with `:`.

Perhaps writing tables like so:

`table_1:part_1,part_2`, without allowing a custom separator is more "familiar" across the industry?

vaintroub:
My preference would be

- no separator option
- `--partitions` boolean option that applies to all databases and tables
- individual partitions are backed up using "table1#p#part_name1", . "table1#p#part_nameN". That is, multiple entries for the same table, with fixed `#p#` used as separator ( or just  `#`). No new --parts position-dependent option.
- This all must either work well with mariadb-import --dir/--tab , or else partitions should be rejected  by mariadb-dump --dir/--tab
(it would be nice of course to make use of --parallel, so it makes sense to make it work with "multi-file" output)

vuvova:
Make sure it works with all possible table names. If you use `:` as a separator — it shouldn't break if a table name includes `:`. If you're using `#` — ditto.

cvicentiu:
Thank you @vaintroub @vuvova for the comments.

I think @vaintroub suggestions make the most sense. I have a preference for `:` as opposed to  `#p#` as the separator. @vaintroub are you strongly against the use of `:`, or see any other problems that `:` could cause? To me that reads much easier.

So I formalize the spec as:
1. `--partitions` option to allow for specifying individual partitions in tables. Without it, `table1:p1` is considered
2. Define each partition separately, with full table name prefix:
`mariadb-dump --partitions table1:p1 table1:p2 table1:p3` or
`mariadb-dump --partitions table1#p#p1 table1#p#p2 table1#p#p3`
3. For the first iteration, to minimise scope, let's disallow `--partitions` with `--dir` and `--tab` options.
4. Ensure `mariadb-dump` works with table names including the separator as part of their name. This needs to be covered by test cases.

@slawomir-pryczek Would you like to take a stab at implementing things this way?

slawomir-pryczek:
Hi Everyone, thanks for the feedback. @cvicentiu yes, i'll wait untill spec is completed and start implementing it.

As for the issues with backing up tables with separator in the name, we could escape colon by using double colon, if colon is picked for separator. That should be simple to use and don't break anything in "normal" mode while everything will work in partition mode.

As for doing each partition/table pair separately in command line, that should work. Will have to just backup all partitions at once, and put it in single file because of drop/create table blocks would make this kind of backp unusable and only single partition would be restored (if we have separated files for each partition). If we just backup each selected partition at once and keep it in single place all should be fine.

So maybe this way will work if we have colon in the name for partition mode `some::table::name:partition1`. So we say in partition mode all colons in table names should be replaced with double colon...

vaintroub:
@cvicentiu : The ":" separator is fine on the command line . But I still prefer '#' , because it can be used in filenames, and ':' not necessarily . 
Let's imagine, mariadb-dump --dir is supported, and you'd now have to  have a file-per-partition, and you'd need to name this file and pass it into `SELECT INTO OUTFILE`. Can you name it "table:part.txt"? Not on Windows. Can you name it "table#part.txt" . Yes, everywhere. If you stick with colon on the command line, you'd have to invent another naming scheme for files on disk. IF you used "#" or "#p#", you would not have to.
So while I do not have a very strong opinion, I lean into other separator that is not listed as problematic character in https://en.wikipedia.org/wiki/Filename#Problematic_characters

slawomir-pryczek:
Yes that's true, so we can maybe have # as partition separator and ## for escaping "#" char in table names. If that sounds good i'll start implementing it.

slawomir-pryczek:
Hello everyone, added the changes. Hopefully the design is good.

1. Disabled --databases with --partitions
2. Multiple partitions can be selected by including table >1 time, eg. mytable#p1 mytable#p2
3. Backtick in partition names is escaped with ``, ## to include hash in table name
4. Works with --tab and directory options
5. For OUTFILE "related" options, the file name won't include partition(s) name(s) for now

Please comment on the design

```
./mariadb-dump --force  --partitions --skip-ssl -uroot -p -h192.168.10.200 accounts __parttest#p1 __parttest#p2
```

vuvova:
rebased to main, resolved conflicts

vuvova:
@slawomir-pryczek, what's the use case? why do you want to dump only selected partitions?

slawomir-pryczek:
Hello, The use case is that we have a stats tables, around 200-400gb of data each, backup/restore takes several days if that's not done by-partition. Let me know any feedback

vuvova:
Do you mean, you don't need to backup all partitions, or you want to dump them in parallel with multiple `mariadb-dump` invocations, one per partition?

slawomir-pryczek:
Both, i don't need to usually backup >1 partition and when we're restoring data eg. on replica it'd be good to restore in parallel. This feature isn't even that important for now, but will be good to have. If you have any comments about the code you can paste here, can prepare something else if you think some other solution would be appropriate.

The main reason behind it is that i have a tool to dump data in parallel, but it works on table level. We're using partitions extensively (as we have a lot of historical data which isn't often accessed). If i can run that on partition level it could probably speed up things x10 or more. We're able to backup a server with around 1.5T of data in 2-3 minutes when doing it in parallel, then we just have ~5 tables with many partitions which are taking around a day.

vuvova:
@vaintroub after [MDEV-37483](https://jira.mariadb.org/browse/MDEV-37483), mariadb-dump uses the same table name encoding as the server. This would mean that mariadb-dump could dump partitions in parallel, naming files using the `#p#` pattern, and mariadb-import can automatically figure out in what partition what file should be imported into. Or it can ignore the `#p#` part and load into the table, because the server will put the data into correct partition anyway. But for concurrency it'd be better to load the file into a separate table and then `EXCHANGE`. What do you think? As an improvement of the parallel dump.

Manual partition selection, as in this PR, still has its use case

vaintroub:
> @vaintroub after [MDEV-37483](https://jira.mariadb.org/browse/MDEV-37483), mariadb-dump uses the same table name encoding as the server. This would mean that mariadb-dump could dump partitions in parallel, naming files using the `#p#` pattern, and mariadb-import can automatically figure out in what partition what file should be imported into. Or it can ignore the `#p#` part and load into the table, because the server will put the data into correct partition anyway. But for concurrency it'd be better to load the file into a separate table and then `EXCHANGE`. What do you think? As an improvement of the parallel dump.
> 
Yes, good idea. Currently, do not have anything to parallelize load of a single huge table currently. So it might be an improvement

vuvova:
reported as [MDEV-37687](https://jira.mariadb.org/browse/MDEV-37687)

client/mysqldump.cc
```diff
@@ -4400,14 +4417,29 @@ static void dump_table(const char *table, const char *db, const uchar *hash_key,
...
+    {
+      print_comment(md_result_file, 0,
+                    "\n--\n-- Dumping data for table %s\n--\n",
+                    fix_for_comment(result_table));
+    }
+    else
+    {
+      print_comment(md_result_file, 0,
+                    "\n--\n-- Dumping data for table %s, partition %s\n--\n",
+                    fix_for_comment(result_table), partition);
```
vuvova, line 4430: `fix_for_comment(partition)`

client/mysqldump.cc
```diff
@@ -4360,6 +4372,11 @@ static void dump_table(const char *table, const char *db, const uchar *hash_key,
     dynstr_append_checked(&query_string, qdatabase);
     dynstr_append_checked(&query_string, ".");
     dynstr_append_checked(&query_string, result_table);
+    if (partition != NULL) {
+      dynstr_append_checked(&query_string, " PARTITION (");
+      dynstr_append_checked(&query_string, partition);
```
vuvova, line 4377: `quote_name(partition)` here instead of duplicating backticks in `get_partition_from_table_name()`

client/mysqldump.cc
```diff
@@ -4400,14 +4417,29 @@ static void dump_table(const char *table, const char *db, const uchar *hash_key,
...
+                    fix_for_comment(result_table), partition);
+    }
     
     dynstr_append_checked(&query_string, "SELECT /*!40001 SQL_NO_CACHE */ ");
     dynstr_append_checked(&query_string, select_field_names.str);
     dynstr_append_checked(&query_string, " FROM ");
     dynstr_append_checked(&query_string, result_table);
+    if (partition != NULL) {
+      dynstr_append_checked(&query_string, " PARTITION (");
+      dynstr_append_checked(&query_string, partition);
```
vuvova, line 4439: `quote_name()`

client/mysqldump.cc
```diff
@@ -6156,6 +6231,69 @@ static int dump_selected_tables(char *db, char **table_names, int tables)
...
+      partition_names[i] = NULL;
+    }
+      
+    for (int i=0; i<tables; i++)
+    {
+      char *tname = table_names[i];
+      char *pname = get_partition_from_table_name(table_names[i]);
+      int tpos = 0;
+
+      // maybe we have this table added already
```
vuvova, line 6254: I'm not sure it's a good idea. first you force the user to specify the table name many times, then you manually remove duplicates. I like your earlier idea of `table#part1,part2,..` more.

Except that it should rather be `table#part1#part2#...` — there's no need to introduce two different separators.

Actually, `table,part1,part2,...` looks nicer. Perhaps `--partitions` should take the separator as a argument?

client/mysqldump.cc
```diff
@@ -6156,6 +6231,69 @@ static int dump_selected_tables(char *db, char **table_names, int tables)
...
+      DYNAMIC_STRING *ppart = partition_names[tpos];
+      if (ppart == NULL) {
+        ppart = (DYNAMIC_STRING*) my_malloc(PSI_NOT_INSTRUMENTED, sizeof(DYNAMIC_STRING), MYF(MY_WME | MY_ZEROFILL));
+        init_dynamic_string_checked(ppart, "", 256, 1024);
+        partition_names[tpos] = ppart;
+      }
+      else 
+        dynstr_append_checked(ppart, ", ");  
+      dynstr_append_checked(ppart, "`");
+      dynstr_append_checked(ppart, pname);
```
vuvova, line 6280: `quote_name()` I suppose

client/mysqldump.cc
```diff
@@ -4360,6 +4372,11 @@ static void dump_table(const char *table, const char *db, const uchar *hash_key,
     dynstr_append_checked(&query_string, qdatabase);
     dynstr_append_checked(&query_string, ".");
     dynstr_append_checked(&query_string, result_table);
+    if (partition != NULL) {
+      dynstr_append_checked(&query_string, " PARTITION (");
+      dynstr_append_checked(&query_string, partition);
+      dynstr_append_checked(&query_string, ") ");
+    }
```
vuvova, line 4380: I think here, inside `if (multi_file_output)`, when `--partitions` is specified, it should create one file per partition, not one file per table.

vuvova (changes requested):
Sorry for delay.

See few comments below. And please, remove end spaces, `git diff` shows them red.

gkodinov (changes requested):
This is a preliminary review. 

Just checking if you are still interested to work on this. 
If you are, in addition to addressing @vuvova 's review comments please also make sure to squash all of the commits into a single one.

gkodinov:
There is no feedback to neither mine nor Serg's review comments. Converting to draft. Please convert it back to "open" when/if you intend to keep working on this.

PR #3892, opened 2025-03-14
MDEV-34621: Fix division by zero in mariadb-slap when iterations=0

KhaledR57:
- [x] *The Jira issue number for this PR is: MDEV-34621*

## Description
`mariadb-slap` crashes with a floating point exception when run with `-i0` due to division by zero in `generate_stats()`. This occurs when calculating average timing by dividing by the number of iterations.
Instead of modifying the minimum value for the `-i` option (which would cause unexpected warning messages even with valid values), this PR implements a targeted fix in the `generate_stats()`. The solution checks if `iterations == 0` before performing the division, and in such cases, returns early from the function.

## Release Notes
Fixed a floating point exception (division by zero) crash in `mariadb-slap` when running with `-i0` (zero iterations).

## How can this PR be tested?

This PR can be tested by manually running mariadb-slap with zero iterations, which would previously cause a crash:

   ```
   $ mariadb-slap -i0 --only-print
   ```
Or
```
$ ./mtr main.mysqlslap
```

## PR quality check

### DIFF
--- a/client/mysqlslap.c
+++ b/client/mysqlslap.c
@@ -2237,6 +2237,13 @@ generate_stats(conclusions *con, option_string *eng, stats *sptr)
   stats *ptr;
   unsigned int x;
 
+  if (eng && eng->string)
+    con->engine= eng->string;
+
+  /* Early return when iterations is 0 to avoid accessing uninitialized sptr */
+  if (iterations == 0)
+    return;
+
   con->min_timing= sptr->timing; 
   con->max_timing= sptr->timing;
   con->min_rows= sptr->rows;
@@ -2257,11 +2264,6 @@ generate_stats(conclusions *con, option_string *eng, stats *sptr)
       con->min_timing= ptr->timing;
   }
   con->avg_timing= con->avg_timing/iterations;
-
-  if (eng && eng->string)
-    con->engine= eng->string;
-  else
-    con->engine= NULL;
 }
 
 void
--- a/mysql-test/main/mysqlslap.test
+++ b/mysql-test/main/mysqlslap.test
@@ -88,3 +88,9 @@ DROP TABLE t1;
 --exec $MYSQL_SLAP --concurrency=1 --silent --iterations=1 --number-int-cols=2 --number-char-cols=3 --auto-generate-sql --auto-generate-sql-guid-primary --create-schema=slap
 
 --exec $MYSQL_SLAP --concurrency=1 --silent --iterations=1 --number-int-cols=2 --number-char-cols=3 --auto-generate-sql --auto-generate-sql-secondary-indexes=1 --create-schema=slap
+
+--echo #
+--echo # Bug MDEV-34621: Fix division by zero in mariadb-slap when iterations=0
+--echo #
+
+--exec $MYSQL_SLAP -i0 --only-print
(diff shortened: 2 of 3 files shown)


client/mysqlslap.c
```diff
@@ -2238,7 +2238,12 @@ generate_stats(conclusions *con, option_string *eng, stats *sptr)
     if (ptr->timing < con->min_timing)
       con->min_timing= ptr->timing;
   }
-  con->avg_timing= con->avg_timing/iterations;
+
+  /* Safety check to prevent division by zero */
+  if (iterations == 0)
+    con->avg_timing= 0;
+  else
+    con->avg_timing= con->avg_timing/iterations;
```
svoj, line 2246: I'd say it is a step towards a wrong direction. `sptr` argument of `generate_stats()` is uninitialised and must not be referenced when `iterations == 0`.

Though `generate_stats()` as such is an overengineering, to simplify a fix you could return early from `generate_stats()`. The only thing that has to be set is `con->engine`.

svoj (changes requested):
Please add a test case and amend your fix.

client/mysqlslap.c
KhaledR57: Thanks for your review. I've implemented the changes and added the appropriate test cases to verify the fix works correctly.

mysql-test/main/mysqlslap.test
```diff
@@ -88,3 +88,11 @@ DROP TABLE t1;
...
 
 --exec $MYSQL_SLAP --concurrency=1 --silent --iterations=1 --number-int-cols=2 --number-char-cols=3 --auto-generate-sql --auto-generate-sql-secondary-indexes=1 --create-schema=slap
+
+--echo #
+--echo # Bug MDEV-34621: Fix division by zero in mariadb-slap when iterations=0
+--echo #
+
+--exec $MYSQL_SLAP -i0 --only-print
+
+--exec $MYSQL_SLAP --silent --concurrency=5 --iterations=0 --number-int-cols=2 --number-char-cols=3 --auto-generate-sql --auto-generate-sql-add-autoincrement --auto-generate-sql-load-type=mixed
```
svoj, line 98: Why 2 invocations of `mysqlslap`? Feels like either covers your bugfix, one should be enough, right?

svoj, line 98: Why 2 invocations of `$MYSQL_SLAP`? Feels like either covers your bugfix, one should be enough, right?

client/mysqlslap.c
```diff
@@ -2219,6 +2219,16 @@ generate_stats(conclusions *con, option_string *eng, stats *sptr)
...
 
+  /* Early return when iterations is 0 to avoid accessing uninitialized sptr */
+  if (iterations == 0)
+  {
+    if (eng && eng->string)
+      con->engine= eng->string;
+    else
+      con->engine= NULL;
+    return;
+  }
```
svoj, line 2230: Please avoid code duplication, just move `con->engine` assignment code above `if (iterations == 0)`. Also `else con->engine= NULL` is redundant and should be removed.

svoj (changes requested):
Adjust commit comment according to https://github.com/MariaDB/server/blob/11.0/CODING_STANDARDS.md

That is split long lines. Sentences should end with period.

Could you also retarget it for 10.6? Or I can do it myself once PR is approved.

client/mysqlslap.c
KhaledR57: I was just trying to stay consistent with the existing code, but you’re right. I'll remove `else con->engine = NULL`.

svoj (approved):

PR #3894, opened 2025-03-17
MDEV-36281	DML aborts during online virtual index

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36281*

## Description
Reason:
=======
- InnoDB DML commit aborts the server when InnoDB does online virtual index. During online DDL, concurrent DML commit operation does read the undo log record and their related current version of the clustered index record. Based on the operation, InnoDB do build the old tuple and new tuple for the table. If the concurrent online index can be affected by the operation, InnoDB does build the entry
for the index and log the operation.

Problematic case is update operation, InnoDB does build the update vector. But while building the old row, InnoDB fails to fill the non-affected virtual column. This lead to server abort while build the entry for index.

Fix:
===
- First, fill the virtual column entries for the new row. Duplicate the old row based on new row and change only the affected fields in old row based on the update vector.

## How can this PR be tested?
./mtr innodb.innodb-index-online

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/r/innodb-index-online.result
+++ b/mysql-test/suite/innodb/r/innodb-index-online.result
@@ -534,7 +534,6 @@ INSERT INTO t1 VALUES(1,1);
 ROLLBACK;
 SET DEBUG_SYNC = 'now SIGNAL inserted';
 connection con1;
-disconnect con1;
 connection default;
 SELECT * FROM t1;
 a	b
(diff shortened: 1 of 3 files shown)


mysql-test/suite/innodb/t/innodb-index-online.test
```diff
@@ -510,12 +510,32 @@ SET DEBUG_SYNC = 'now SIGNAL inserted';
...
+
+connection con1;
+set DEBUG_SYNC="now WAIT_FOR dml_start";
+UPDATE t1 SET f3= f3 + 1;
+set DEBUG_SYNC="now SIGNAL dml_finish";
+
+connection default;
+reap;
+DROP TABLE t1;
+disconnect con1;
```
dr-m, line 537: The `disconnect con1` could be executed before switching to `connection default`.

mysql-test/suite/innodb/t/innodb-index-online.test
```diff
@@ -510,12 +510,32 @@ SET DEBUG_SYNC = 'now SIGNAL inserted';
...
+send ALTER TABLE t1 ADD INDEX(f5, f2, f4);
+
+connection con1;
+set DEBUG_SYNC="now WAIT_FOR dml_start";
+UPDATE t1 SET f3= f3 + 1;
+set DEBUG_SYNC="now SIGNAL dml_finish";
+
+connection default;
+reap;
+DROP TABLE t1;
```
dr-m, line 536: I had to double check that this is actually testing a scenario where the value of an indexed virtual column `f4 AS (f3)` in the being-created index on (`f5`,`f2`,`f4`) is changing. Could we add another index, say, `ADD INDEX(f3,f5)` where only the value of a non-virtual column is changing? And execute both `CHECK TABLE t1 EXTENDED` and a number of `SELECT` with `FORCE INDEX` to display the contents of each created virtual index?

dr-m (commented):
Thank you. The fix looks reasonable to me, but I would like to see a little more detailed test coverage.

mysql-test/suite/innodb/r/innodb-index-online.result
```diff
@@ -543,6 +542,31 @@ CHECK TABLE t1;
...
+connection default;
+CHECK TABLE t1 EXTENDED;
+Table	Op	Msg_type	Msg_text
+test.t1	check	status	OK
+SELECT * FROM t1 USE INDEX(v1);
+f1	f2	f3	f4	f5
+1	2	4	4	1
+SELECT * FROM t1 USE INDEX(v2);
+f1	f2	f3	f4	f5
+1	2	4	4	1
```
dr-m, line 568: I think that we should try to use a covering index read for these: not requesting `*` but only those columns that are present in the secondary index. Can you check that `row_search_mvcc()` will not access the clustered index?

PR #3896, opened 2025-03-18
MDEV-36180 Doublewrite recovery of innodb_checksum_algorithm=full_crc32 page_compressed pages does not work

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36180*

## Description
- InnoDB fails to recover the full crc32 page_compressed page from doublewrite buffer. The reason is that buf_dblwr_t::recover() fails to identify the space id from the page because the page has compressed from FIL_PAGE_FILE_FLUSH_LSN_OR_KEY_VERSION bytes.

Fix:
===
recv_dblwr_t::recovery_page_compressed(): Find the page which has the same page number and try to decompress the page. After the decompression, compare the space id and write the recovered page back to the file.

buf_page_t::read_complete(): If the page looks corrupted and tablespace is compressed in full_crc32 format, try to restore page from doublewrite buffer

recv_dblwr_t::recover_encrypted_page(): Skip the page which doesn't have key_version in FIL_PAGE_FCRC32_KEY_VERSION field

## How can this PR be tested?
./mtr encryption.doublewrite_debug

## PR quality check

### DIFF
--- a/storage/innobase/buf/buf0buf.cc
+++ b/storage/innobase/buf/buf0buf.cc
@@ -3766,10 +3766,9 @@ database_corrupted_compressed:
   if (err == DB_PAGE_CORRUPTED || err == DB_DECRYPTION_FAILED)
   {
 release_page:
-    if (node.space->full_crc32() && node.space->crypt_data &&
-        recv_recovery_is_on() &&
-        recv_sys.dblwr.find_encrypted_page(node, id().page_no(),
-                                           const_cast<byte*>(read_frame)))
+    if (node.space->full_crc32() && recv_recovery_is_on() &&
+        recv_sys.dblwr.find_deferred_page(node, id().page_no(),
+                                          const_cast<byte*>(read_frame)))
     {
       /* Recover from doublewrite buffer */
       err= DB_SUCCESS;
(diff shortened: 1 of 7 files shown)


storage/innobase/buf/buf0buf.cc
```diff
@@ -3766,14 +3766,21 @@ dberr_t buf_page_t::read_complete(const fil_node_t &node) noexcept
...
+       {
+dblwr_success:
+         /* Recover from doublewrite buffer */
+         err= DB_SUCCESS;
+         goto success_page;
+       }
+       else if (node.space->is_compressed() &&
+                recv_sys.dblwr.find_page_compressed(node, id().page_no(),
+                                                    const_cast<byte*>(read_frame)))
+         goto dblwr_success;
```
dr-m, line 3783: The lines starting with `{` are indented too deep by 1 space.

If this is the only call site to the two functions, could we rename the member function to `find_deferred_page` and let it handle both these cases? It does not seem to be necessary to evaluate further conditions on `node.space` here; we could let the callee do that. After all, this code should be executed very rarely (`ATTRIBUTE_COLD`); we would want to minimize the overhead for normal page reads related to something that is only rarely needed during crash recovery.

dr-m (approved):
Thank you, this looks fine. I only have a minor suggestion.

PR #3897, opened 2025-03-19
MDEV-35304 fixup: Timing-independent MTR test

ParadoxV5:
* [x] ~~*The Jira issue number for this PR is: [MDEV-35304](https://jira.mariadb.org/browse/MDEV-35304)*~~

## Description
Improve the `multi_source.connects_tried` feature test:
* Replace the main course from *checking `Connect_Tried` after sleeping a specific time* to *checking the duration after reaching a minimum `Connect_Tried`*
  * Not using a specific SLEEP eliminates the possibility (of failure) that it awakens before the `Connects_Tried` counter increments.
* In replacement-based tests of conventional views (SHOW & variants), assert the parallel values without distinguishing which is which
  * There were rare cases where both values end up equal (even though one was configured to tick more frequently than the other).
* Comment touchups

## Release Notes
N/A

## How can this PR be tested?
It has been evident that timing tests easily lead to sporadic failures.
Perhaps our CI (Buildbot) are overloaded with concurrent jobs that their thread schedulers had to make possibly-unrealistic compromises.

## PR quality check
(Yes, this is a refactoring fix concerning a `main`-exclusive (unreleased) new feature.)

### DIFF
--- a/mysql-test/suite/multi_source/connects_tried.test
+++ b/mysql-test/suite/multi_source/connects_tried.test
@@ -4,7 +4,7 @@
 # separate counters in parallel multi-source (SHOW SLAVE [name] STATUS).
 #
 # Note: nearly all SELECT results are timing-sensitive,
-# they therefore list unexpected rows because.
+# therefore, they only list unexpected rows.
 
 --source include/have_binlog_format_mixed.inc # The test is agnostic of binlog formats.
 # `rpl_*.inc` (still) expects `server_1` be `master` and `server_2` be `slave`.
@@ -17,14 +17,16 @@
 --source include/rpl_stop_server.inc
 
 --connection slave
-CHANGE MASTER TO master_connect_retry=1;
+CHANGE MASTER TO master_connect_retry=2;
 --replace_result $SERVER_MYPORT_3 SERVER_MYPORT_3
---eval CHANGE MASTER 'named' TO master_host='127.0.0.1', master_port=$SERVER_MYPORT_3, master_user='root', master_ssl_verify_server_cert=0, master_connect_retry=2
+--eval CHANGE MASTER 'named' TO master_host='127.0.0.1', master_port=$SERVER_MYPORT_3, master_user='root', master_ssl_verify_server_cert=0, master_connect_retry=1
 --echo # `Connects_Tried` is 0 before connections begin.
 SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
 
 --disable_warnings
 START ALL SLAVES; # will fail because the masters are down
+SET @time_begin= CURRENT_TIMESTAMP(1);
+--enable_warnings
 # CR_CONN_HOST_ERROR
 --let $slave_io_errno= 2003
 --let $slave_io_error_is_nonfatal= 1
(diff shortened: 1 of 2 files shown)


knielsen (approved):
Hi Jimmy,

Here is my review of the patch for MDEV-35304, fixup: Threading-agnostic MTR test

First some general comments on sporadic test failures:

I do sympathise with you experiencing the surprising difficulty of creating
working testcases for replication, having after all suffered this myself for
many years ;-). The correct approach is to write test cases at the same
level we would write server code, where we would never rely on sleeps or
predictable timing to synchronise access to shared data structures etc. Thus
in test cases, assume any part of server or test code could potentially
stall for several seconds, and design the test to be robust towards that.
For example, wait for an expected condition to be true before SELECTing it
or proceeding in the test flow.

Also, for any major additions to replication tests, run a robustness test
locally before pushing to check for sporadic failures. For example:

  (cd mysql-test/ && ./mtr multi_source.connects_tried{,,,}{,,,}{,,} --parallel=48 --repeat=1000)

The exact number of --parallel or --repeat depends on the size of your
machine and the time the test takes to run. The point is to run the test in
parallel close to the limit of the capacity of your machine for several
hours. In my experience, this will trigger a large fraction of the sporadic
failures seen in buildbot. Such a local run must be clear before pushing a
test.

The test suite is a tremendous important part of the MariaDB codebase, and I
appreciate the effort you put into tests! I have found and fixed many
serious bugs from sporadic test suite failures that would be almost
impossible to track down in a production environment. That being said,
sometimes a bit of pragmatism can be appropriate, and if a specific thing is
particularly difficult to test reliably, it can be better to omit from the
testsuite and test once manually instead.

Next, specifics to the patch:

The commit title and description are hard to understand.

"threading-agnostic"? It has little to do with threads.

"Replace the main course from timing checks to checking timing"? Not sure
what is the difference between "timing checks" and "checking timing".

I did understand that the point of the patch is to fix sporadic test
failures of the multi_source.connects_tried testcase.

Then this:

```
CREATE TEMPORARY TABLE status_begin AS
  SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
--echo # `Connects_Tried` is 1 immediately after connection begins.
SELECT Connection_name, Connects_Tried
  FROM status_begin
  WHERE Connects_Tried <= 0;
```

If you want to check this, I think you need to put some debug_sync to block
the value from updating if the test case thread of execution would be
delayed by more than the value of master_connect_retry. Eg, try to put a
--sleep 5 before this, and I assume this will make the test fail?

```
+--echo # `Connects_Tried` takes (at least) 2s to increments for connection '' and 1s for 'named'.
+--let $slave_param= Connects_Tried
+--let $slave_param_comparison= >
+
+SET @@SESSION.default_master_connection= 'named';
+--let $slave_param_value= `SELECT Connects_Tried FROM status_begin WHERE Connection_name = ''`
+--source include/wait_for_slave_param.inc
```

Agree, waiting for the condition to occur, seems better.

```
+# (The `Connects_Tried` of the more-frequent 'named' should be ahead of the ''
+#  one, though it's possible that the thread scheduler decides to be naughty.)
```

As said, I do sympathise with your struggles, but it's best not to let it
show through in the final code - the thread scheduler really isn't trying to
hurt you personally, it is not being "naughty". Better to explain the actual
issue.

For example: "Normally the `Connects_tried` in slave connection 'named' will
be ahead of the connection '', as it reconnects more frequently; this
however can depend on exact timing and scheduling when the test runs on a
loaded machine."

The patch is approved after fixing mentioned issues and successfully passing
the stress-testing for a couple hours as described above.

ParadoxV5:
almost rebased (which would complicate the diff); oops

---

Is this much better, 师傅?

ParadoxV5:
Oh also, @bnestere I see you self-assigned this PR. Got plans?

mysql-test/suite/multi_source/connects_tried.test
```diff
@@ -33,25 +35,39 @@ SET @@SESSION.default_master_connection= 'named';
...
+--let $slave_param_value= `SELECT Connects_Tried FROM status_begin WHERE Connection_name <> ''`
+--source include/wait_for_slave_param.inc
+SELECT @time_begin, CURRENT_TIMESTAMP(1)
+  WHERE TIMESTAMPDIFF(SECOND, @time_begin, CURRENT_TIMESTAMP(1)) < 2;
+--replace_result $_show_slave_status_value Connects_Tried
+--eval INSERT INTO status_sleep SET Connection_name= '', Connects_Tried= $_show_slave_status_value
 
---echo # Boot replication up and assert final count
+--echo # Boot replication up and compare the final counts
+# (In an ideal environment,  of the 'named' connection that reconnects
```
ParadoxV5, line 68: There’s always a sentence I forget to finish. 😑
```suggestion
# (In an ideal environment, the 'named' connection that reconnects
```

PR #3898, opened 2025-03-19
MDEV-36042 Assertion failed in Binary_string::q_append

DaveGosselin-MariaDB:
Like MDEV-35062, reserve sufficient space in the result for q_append'ed data, as q_append does not itself reserve space like it's append counterpart.


sql/spatial.cc
```diff
@@ -4228,6 +4228,7 @@ int Gis_multi_polygon::make_clockwise(String *result) const
     if(polygon->make_clockwise(&clockwise_wkb))
       return 1;
 
+    result->reserve(sizeof(char) + sizeof(uint32));
```
grooverdan, line 4231: should `return 1` if the reservation fails. After that, good to merge.

PR #3899, opened 2025-03-20
MDEV-36316/MDEV-36327/MDEV-36328 Debug msan fixes 10.6

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36316/MDEV-36317/MDEV-36318*

## Description

Various fixing to allow a Debug MSAN with Clang-20 to pass tests.

See individual commit messages for details.

## Release Notes
* Internal changes only.

## How can this PR be tested?

```
podman run --rm -ti -v "$PWD":/source:z --mount=type=tmpfs,tmpfs-size=10G,dst=/build --shm-size=10g --workdir /build --entrypoint /bin/bash --user buildbot --cap-add=SYS_PTRACE --privileged quay.io/mariadb-foundation/bb-worker:dev_debian12-msan-clang-20

cmake    -DWITH_EMBEDDED_SERVER=OFF \
                -DWITH_INNODB_{BZIP2,LZ4,LZMA,LZO,SNAPPY}=OFF \
                -DPLUGIN_{MROONGA,ROCKSDB,OQGRAPH,SPIDER}=NO \
                -DWITH_ZLIB=bundled \
                -DHAVE_LIBAIO_H=0 \
                -DCMAKE_DISABLE_FIND_PACKAGE_{URING,LIBAIO}=1 \
                -DWITH_NUMA=NO \
                -DWITH_SYSTEMD=no \
                -DWITH_MSAN=ON \
                -DHAVE_CXX_NEW=1 \
                -DCMAKE_{EXE,MODULE}_LINKER_FLAGS="-L${MSAN_LIBDIR} -Wl,-rpath=${MSAN_LIBDIR}" \
                -DCMAKE_CXX_FLAGS=-fsanitize=memory \
                -DWITH_DBUG_TRACE=OFF
                -DCMAKE_BUILD_TYPE=Debug \
                /source
cmake --build .
mysql-test/mtr --parallel=auto --force --big-test
```

## PR quality check

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -14476,8 +14476,8 @@ ha_innobase::records_in_range(
 		n_rows = rtr_estimate_n_rows_in_range(
 			index, range_start, mode1);
 	} else {
-		btr_pos_t tuple1(range_start, mode1, pages->first_page);
-		btr_pos_t tuple2(range_end,   mode2, pages->last_page);
+		btr_pos_t tuple1(range_start, mode1, 0);
+		btr_pos_t tuple2(range_end,   mode2, 0);
 		n_rows = btr_estimate_n_rows_in_range(index, &tuple1, &tuple2);
 		pages->first_page= tuple1.page_id.raw();
 		pages->last_page=  tuple2.page_id.raw();
--- a/storage/innobase/handler/i_s.cc
+++ b/storage/innobase/handler/i_s.cc
@@ -5453,7 +5453,7 @@ i_s_dict_fill_sys_columns(
 	const char*	col_name,	/*!< in: column name */
 	dict_col_t*	column,		/*!< in: dict_col_t struct holding
 					more column information */
-	ulint		nth_v_col,	/*!< in: virtual column, its
+	ulint*		nth_v_col,	/*!< in: virtual column, its
 					sequence number (nth virtual col) */
 	TABLE*		table_to_fill)	/*!< in/out: fill this table */
 {
@@ -5468,7 +5468,7 @@ i_s_dict_fill_sys_columns(
 	OK(field_store_string(fields[SYS_COLUMN_NAME], col_name));
 
 	if (column->is_virtual()) {
-		ulint	pos = dict_create_v_col_pos(nth_v_col, column->ind);
+		ulint	pos = dict_create_v_col_pos(*nth_v_col, column->ind);
 		OK(fields[SYS_COLUMN_POSITION]->store(pos, true));
 	} else {
 		OK(fields[SYS_COLUMN_POSITION]->store(column->ind, true));
@@ -5535,7 +5535,7 @@ i_s_sys_columns_fill_table(
 		if (!err_msg) {
 			err = i_s_dict_fill_sys_columns(
 				thd, table_id, col_name,
-				&column_rec, nth_v_col,
+				&column_rec, &nth_v_col,
 				tables->table);
 			if (err) {
 				err = i_s_sys_error_handling(err, thd);
--- a/storage/innobase/include/rem0rec.inl
+++ b/storage/innobase/include/rem0rec.inl
@@ -156,6 +156,9 @@ rec_set_bit_field_1(
 	ut_ad(((mask >> shift) << shift) == mask);
 	ut_ad(((val << shift) & mask) == (val << shift));
 
+#ifndef DBUG_OFF
+	MEM_MAKE_DEFINED(rec - offs, 1);
+#endif
 	mach_write_to_1(rec - offs,
 			(mach_read_from_1(rec - offs) & ~mask)
 			| (val << shift));
@@ -198,6 +201,9 @@ rec_set_bit_field_2(
 	ut_ad(((mask >> shift) << shift) == mask);
 	ut_ad(((val << shift) & mask) == (val << shift));
 
+#ifndef DBUG_OFF
+	MEM_MAKE_DEFINED(rec - offs, 2);
+#endif
 	mach_write_to_2(rec - offs,
 			(mach_read_from_2(rec - offs) & ~mask)
 			| (val << shift));
--- a/include/my_pthread.h
+++ b/include/my_pthread.h
@@ -668,7 +668,17 @@ extern void my_mutex_end(void);
   We need to have at least 256K stack to handle calls to myisamchk_init()
   with the current number of keys and key parts.
 */
-# if defined(__SANITIZE_ADDRESS__) || defined(WITH_UBSAN)
+#if !defined(__has_feature)
+#define __has_feature(x) 0
+#endif
+#if defined(__clang__) && __has_feature(memory_sanitizer) && !defined(DBUG_OFF)
+/*
+  MSAN in Debug with clang-20.1 required more memory to complete
+  mtr begin/end checks. The result without increase was MSAN
+  errors triggered on a call instruction.
+*/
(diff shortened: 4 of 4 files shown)


storage/innobase/include/rem0rec.inl
```diff
@@ -156,6 +156,9 @@ rec_set_bit_field_1(
 	ut_ad(((mask >> shift) << shift) == mask);
 	ut_ad(((val << shift) & mask) == (val << shift));
 
+#ifndef DBUG_OFF
+	MEM_MAKE_DEFINED(rec - offs, 1);
+#endif
 	mach_write_to_1(rec - offs,
 			(mach_read_from_1(rec - offs) & ~mask)
 			| (val << shift));
```
dr-m, line 164: This looks incorrect to me. Why would we claim that all bits at `rec[-offs]` are initialized when we are only overwriting some of the bits here? What would fail if this change and the similar change to `rec_set_bit_field_2()` were omitted?

storage/innobase/handler/i_s.cc
```diff
@@ -5453,7 +5453,7 @@ i_s_dict_fill_sys_columns(
 	const char*	col_name,	/*!< in: column name */
 	dict_col_t*	column,		/*!< in: dict_col_t struct holding
 					more column information */
-	ulint		nth_v_col,	/*!< in: virtual column, its
+	ulint*		nth_v_col,	/*!< in: virtual column, its
 					sequence number (nth virtual col) */
```
dr-m, line 5457: I do not understand why we would need any of the changes to this file and which problem these changes would solve. We’re no longer passing a read-only parameter by value but via a pointer that is effectively read-only. Can you test again without including any of these changes?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -14476,8 +14476,8 @@ ha_innobase::records_in_range(
 		n_rows = rtr_estimate_n_rows_in_range(
 			index, range_start, mode1);
 	} else {
-		btr_pos_t tuple1(range_start, mode1, pages->first_page);
-		btr_pos_t tuple2(range_end,   mode2, pages->last_page);
+		btr_pos_t tuple1(range_start, mode1, 0);
+		btr_pos_t tuple2(range_end,   mode2, 0);
```
dr-m, line 14480: This would seem to be the actual fix. `~0ULL` might be a safer value, but I think that 0 should be OK as well, because the smallest possible index page number is 3.

grooverdan: I think I could add `MEM_UNDEFINED(&tuple1.page_id, sizeof tuple1.page_id)` (plus tuple2) to protect against future `btr_estimate_n_rows_in_range` that access it.

storage/innobase/include/rem0rec.inl
grooverdan: The trace was like in MDEV-36316.

I agree it seem overly incorrect claiming all bits are initialised. It only occurred in Debug mode so I'm assuming a less optimised code makes this look different.  I'll look at forcing a higher optimisation on these blocks as an alternative.

storage/innobase/handler/i_s.cc
grooverdan: The error received was in MDEV-36327.

It was working around a msan patten of possibly not tracing undefined memory once it got copied to registers and then used as argument to the next function.

An alternative is:
```
$ git diff storage/innobase/dict/dict0load.cc
diff --git a/storage/innobase/dict/dict0load.cc b/storage/innobase/dict/dict0load.cc
index 87ff163a233..0b2cdc9e01b 100644
--- a/storage/innobase/dict/dict0load.cc
+++ b/storage/innobase/dict/dict0load.cc
@@ -1184,6 +1184,8 @@ static const char *dict_load_column_low(dict_table_t *table,
        /* Report the virtual column number */
        if ((prtype & DATA_VIRTUAL) && nth_v_col != NULL) {
                *nth_v_col = dict_get_v_col_pos(pos);
+       } else {
+               *nth_v_col = 0;
        }
 
        return(NULL);
```

dr-m: Oh, right, starting with clang-16 or thereabouts, `-fsanitize=memory` would complain when uninitialized data is being passed to function calls, and there is no way to disable that via `MSAN_OPTIONS`. In older clang versions, the data was passed just fine, apparently along with the shadow bytes that would indicate which bits are uninitialized.

dr-m (commented):
I installed a fresh build of clang-20 from https://apt.llvm.org, specifically:
```
Debian clang version 20.1.2 (++20250402124457+58df0ef89dd6-1~exp1~20250402004517.100)
```
I built MSAN-instrumented libraries using the `build-msan19.sh` script attached to [MDEV-20377](https://jira.mariadb.org/browse/MDEV-20377). For `libidn2` I used the release tarball, because `apt source libidn2` would omit the `configure` script. I tested e7442e5eb934980b981c1c68b4055241ff38f73b (10.6) and 25737dbab7fec6d05a50353baf5b15ad97dcbd0f (10.5), both `CMAKE_BUILD_TYPE=Debug` and `CMAKE_BUILD_TYPE=RelWithDebInfo`, and no issues were reported. Here is my script for invoking `cmake`, which I invoked with `CLANG=20`:
```bash
#!/bin/bash
set -ex
test ! -e .git
test ! -e .bashrc
: ${CLANG=15}
# sudo apt install libunwind15-$CLANG
exec cmake -DCMAKE_{C_COMPILER=clang,CXX_COMPILER=clang++}-$CLANG \
-DCMAKE_C_FLAGS='-O2 -march=native -mtune=native -Wno-unused-command-line-argument -fdebug-macro -fno-limit-debug-info' \
-DCMAKE_CXX_FLAGS='-stdlib=libc++ -O2 -march=native -mtune=native -Wno-unused-command-line-argument -fdebug-macro -fno-limit-debug-info' \
-DCMAKE_{EXE,MODULE,SHARED}_LINKER_FLAGS:STRING=-fuse-ld=lld-$CLANG \
-DWITH_DBUG_TRACE=OFF \
-DWITH_EMBEDDED_SERVER=OFF -DWITH_UNIT_TESTS=OFF -DCMAKE_BUILD_TYPE=Debug \
-DWITH_INNODB_{BZIP2,LZMA,LZO,SNAPPY}=OFF \
-DPLUGIN_{ARCHIVE,TOKUDB,MROONGA,OQGRAPH,ROCKSDB,CONNECT,SPIDER}=NO -DWITH_SAFEMALLOC=OFF \
-DWITH_{ZLIB,SSL,PCRE}=bundled \
-DHAVE_LIBAIO_H=0 \
-DCMAKE_DISABLE_FIND_PACKAGE_URING=1 -DCMAKE_DISABLE_FIND_PACKAGE_LIBAIO=1 \
-DWITH_MSAN=ON \
-G Ninja "$@"
```
Note: there is no package installed similar to what the commented line says. I have `libunwind-20` and `libunwind-20-dev`. With clang-20 I’d get lots of trouble if I pointed `LD_LIBRARY_PATH` to an MSAN-instrumented `libunwind.so`. So, I omitted that instrumented library at test runtime.

Can you please check again?

Also, it is worth noting is that I used `MSAN_OPTIONS=poison_in_dtor=0` in order to work around [MDEV-30942](https://jira.mariadb.org/browse/MDEV-30942). Can you double-check if that would make the problems disappear in your environment? If yes, then we’d seem to need a more specific fix.

grooverdan:
Won't fix - avoid with `-DC{,XX}_FLAGS=-O2`

PR #3900, opened 2025-03-20
Fixes linker issues and compilation issues while building on AIX

mehendarkarprajwal:
We need berok flag to ignore unresolved symbol errors while building on AIX. Former part of the patch detects AIX system during compilation.

### DIFF
--- a/storage/innobase/unittest/CMakeLists.txt
+++ b/storage/innobase/unittest/CMakeLists.txt
@@ -32,7 +32,7 @@ IF (WITH_INNODB_AHI)
   MY_ADD_TEST(innodb_ahi)
 ENDIF()
 # See explanation in innobase/CmakeLists.txt
-IF(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|powerpc64|s390x")
+IF(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|powerpc64|s390x" OR CMAKE_SYSTEM_NAME MATCHES "AIX")
   ADD_COMPILE_FLAGS(
       ../sync/srw_lock.cc
       COMPILE_FLAGS "-mhtm"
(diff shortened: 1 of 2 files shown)


svoj:
@grooverdan since you've been on AIX lately anyway, will you want to take care of this PR?

svoj:
@mehendarkarprajwal thanks for your contribution. MariaDB Foundation requires pull requests to be submitted either under BSD New 3 Clause License, or under the [MariaDB Contributor Agreement](https://mariadb.com/kb/en/library/mca/).

grooverdan:
> We need berok flag to ignore unresolved symbol errors while building on AIX. Former part of the patch detects AIX system during compilation.

I've got part of these going in #3965.  As noted by @vaintroub in review we need to be a bit careful with linker flags.

What would the proper solution/process look like to avoid the linker flag?

I discovered some getopt_long in C/C (libmariadb) submodule on less necessary things so a work around there is needed, https://public.dhe.ibm.com/aix/freeSoftware/aixtoolbox/PATCHES/mariadb-10.4.17-getopt_long.patch was provided by Ayappan but we'll see what was acceptable.

I'll think I'll take this over and include it in #3965 with the other fixes to where 10.6 was working.

For the next contribution:

> @mehendarkarprajwal thanks for your contribution. MariaDB Foundation requires pull requests to be submitted either under BSD New 3 Clause License, or under the [MariaDB Contributor Agreement](https://mariadb.com/kb/en/library/mca/).

Note when I was at IBM, the MCA had been approved by legal already. I'm relying on you to to check if that's still there for the next IBM contribution.

mehendarkarprajwal:
Hi @svoj @grooverdan . Apologies for delay in the process, but it's taking some while from my side for accepting the MariaDB CLA, because I'm still in talks with regards to accepting the CLA. It will take some time from my side until my next contribution. 

And surely I will check up on the MCA that you mentioned about @grooverdan that had already been approved.

PR #3901, opened 2025-03-20
MDEV-34953 Add JSON flag for `mysqldumpslow.sh` output

HNOONa-0:
- [x] *The Jira issue number for this PR is: MDEV-34953*

## Description
The `mysqldumpslow.sh` script outputs data as a string. This change introduces a --json flag to format the output as JSON for better structure.
## Release Notes
Added a `--json` flag to `mysqldumpslow.sh`, enabling JSON-formatted output for improved integration with tools that process structured JSON data.
## How can this PR be tested?
It can be manually tested on any `*-slow.log` file. You can provide the `filepath` as an argument. Here is a video showing how it works:

https://github.com/user-attachments/assets/a2bab163-b3da-452e-8e9b-1ee72555ff06

The problem is that I had to download the JSON module on my machine to encode the data as JSON, how do i make it preinstalled on other machines so that it works out right away for other users?

## PR quality check

### DIFF
--- a/scripts/mysqldumpslow.sh
+++ b/scripts/mysqldumpslow.sh
@@ -25,6 +25,7 @@
 
 use strict;
 use Getopt::Long;
+use JSON;
 
 warn "$0: Deprecated program name. It will be removed in a future release, use 'mariadb-dumpslow' instead\n"
   if $0 =~ m/mysqldumpslow$/;
@@ -50,6 +51,7 @@ GetOptions(\%opt,
     'h=s',	# hostname/basename of db server for *-slow.log filename (can be wildcard)
     'i=s',	# name of server instance (if using mysql.server startup script)
     'l!',	# don't subtract lock time from total time
+    'json!', # Output JSON format (default: false)
 ) or usage("bad option");
 
 $opt{'help'} and usage();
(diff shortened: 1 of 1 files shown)


svoj:
Hi @HNOONa-0, is it still a draft or is it good enough for review? Is there anything you're expecting from MariaDB developers?

HNOONa-0:
> Hi @HNOONa-0, is it still a draft or is it good enough for review? Is there anything you're expecting from MariaDB developers?

Hello @svoj. I mentioned that I had to install the JSON module in Perl to properly encode the data as JSON. The JSON module does not ship natively with Perl, you have to install it. How should I handle installing the module on the user's machine to ensure everything works correctly? Or is this something I don't need to worry about?
Alternatively I could write functions that manually encode the data as JSON.

svoj:
@HNOONa-0 I see. Is the feature complete otherwise?

HNOONa-0:
> @HNOONa-0 I see. Is the feature complete otherwise?

Yes, I think it's.

grooverdan:
> > Hi @HNOONa-0, is it still a draft or is it good enough for review? Is there anything you're expecting from MariaDB developers?
> 
> Hello @svoj. I mentioned that I had to install the JSON module in Perl to properly encode the data as JSON. The JSON module does not ship natively with Perl, you have to install it. How should I handle installing the module on the user's machine to ensure everything works correctly? Or is this something I don't need to worry about? Alternatively I could write functions that manually encode the data as JSON.

Can we make the perl module optional like (the following from my AI assisted, and I'm not so good at perl so improvement encouraged):
```
my $have_json;

eval {
    require JSON;
    JSON->import();  # Only if needed
    $have_json = 1;
};

...
if (!$opt{json}) {
if (!$have_json) {
    die "JSON module required for --json";
}
```


Parsing explain https://mariadb.com/kb/en/explain-in-the-slow-query-log/ would probably be too much, but it should handle it in the output.

Optional - can this handle engine stats? https://mariadb.com/kb/en/slow-query-log-extended-statistics/

svoj:
@grooverdan that's what I was going to suggest too.

HNOONa-0:
Hello @grooverdan, I made some modifications to the script. I parsed additional data using regular expressions. Here is a video showing showcasing it:

https://github.com/user-attachments/assets/75420ecd-2ab8-48ad-8924-2ca11753e4d2

Here is the log file I used to test this change:
[slow-queries.log](https://github.com/user-attachments/files/20018073/slow-queries.log)
Here is my configuration:
```
[client-server]
socket = /home/hani/mariadb.sock

[mariadb]
datadir = /home/hani/mariadb-data
slow_query_log
slow_query_log_file=slow-queries.log
log-slow-verbosity=All
long_query_time=0.0
```

HNOONa-0:
Hello @grooverdan , I made minor improvements to the script. Previously, the log headers were always attached to the final query, even in textual output mode. This was because the old regex assumed a slash (`/Version...`) at the start of the string but that is not the case. I’ve adjusted the logic accordingly and it now works as expected.
PTAL!

grooverdan (approved):

grooverdan:
Thanks @HNOONa-0 

Commits squashed up and getting merged now.

PR #3902, opened 2025-03-20
MDEV-36017 Alter table aborts when temporary directory is full

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36017*

## Description
Problem:
=======
- In 10.11, During Copy algorithm, InnoDB does use bulk insert for row by row insert operation. When temporary directory ran out of memory, row_mysql_handle_errors() fails to handle DB_TEMP_FILE_WRITE_FAIL.

- During inplace algorithm, concurrent DML fails to write the log operation into the temporary file. InnoDB fail to mark the error for the online log.

Fix:
===
row_mysql_handle_errors(): Rollback the transaction when InnoDB encounters DB_TEMP_FILE_WRITE_FAIL

convert_error_code_to_mysql(): Report an aborted transaction when InnoDB encounters DB_TEMP_FILE_WRITE_FAIL

row_log_online_op(): Mark the error in online log when InnoDB ran out of temporary space

## How can this PR be tested?
./mtr innodb.alter_temp_fail

## PR quality check

### DIFF
--- a/storage/innobase/btr/btr0cur.cc
+++ b/storage/innobase/btr/btr0cur.cc
@@ -3697,8 +3697,10 @@ btr_cur_optimistic_update(
 	*offsets = rec_get_offsets(rec, index, *offsets, index->n_core_fields,
 				   ULINT_UNDEFINED, heap);
 #if defined UNIV_DEBUG || defined UNIV_BLOB_LIGHT_DEBUG
+	/* Blob pointer can be null if InnoDB was killed or
+	ran out of space while allocating a page. */
 	ut_a(!rec_offs_any_null_extern(rec, *offsets)
-	     || thr_get_trx(thr) == trx_roll_crash_recv_trx);
+	     || thr_get_trx(thr)->in_rollback);
 #endif /* UNIV_DEBUG || UNIV_BLOB_LIGHT_DEBUG */
 
 	if (UNIV_LIKELY(!update->is_metadata())
@@ -4371,7 +4373,12 @@ btr_cur_pessimistic_update(
 					 cursor, offsets, offsets_heap,
 					 new_entry, &rec,
 					 &dummy_big_rec, n_ext, NULL, mtr);
-	ut_a(err == DB_SUCCESS);
+	if (err) {
+		/* This should happen when InnoDB tries to extend the
+		tablespace */
+		ut_ad(err == DB_OUT_OF_FILE_SPACE);
+		return err;
+	}
 	ut_a(rec);
 	ut_a(dummy_big_rec == NULL);
 	ut_ad(rec_offs_validate(rec, cursor->index(), *offsets));
@@ -6240,6 +6247,9 @@ btr_store_big_rec_extern_fields(
 					       FSP_NO_DIR, 0, &mtr, &mtr,
 					       &error);
 
+			DBUG_EXECUTE_IF("btr_page_alloc_fail",
+					block= nullptr;
+					error= DB_OUT_OF_FILE_SPACE;);
 			if (!block) {
 alloc_fail:
                                 mtr.commit();
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -2161,6 +2161,11 @@ convert_error_code_to_mysql(
 		return(HA_ERR_RECORD_FILE_FULL);
 
 	case DB_TEMP_FILE_WRITE_FAIL:
+		/* This error can happen during
+		copy_data_between_tables() or bulk insert operation */
+		innodb_transaction_abort(thd,
+					 innobase_rollback_on_timeout,
+					 error);
 		my_error(ER_GET_ERRMSG, MYF(0),
                          DB_TEMP_FILE_WRITE_FAIL,
                          ut_strerr(DB_TEMP_FILE_WRITE_FAIL),
@@ -15931,7 +15936,7 @@ ha_innobase::extra(
 		}
 		m_prebuilt->table->skip_alter_undo = 0;
 		if (dberr_t err= trx->bulk_insert_apply<TRX_DDL_BULK>()) {
-			m_prebuilt->table->skip_alter_undo = 0;
+			trx->rollback();
 			return convert_error_code_to_mysql(
 				 err, m_prebuilt->table->flags,
 				 trx->mysql_thd);
--- a/storage/innobase/row/row0merge.cc
+++ b/storage/innobase/row/row0merge.cc
@@ -5177,6 +5177,9 @@ dberr_t row_merge_bulk_t::write_to_tmp_file(ulint index_no)
                        m_block, m_crypt_block,
                        buf->index->table->space->id))
     return DB_TEMP_FILE_WRITE_FAIL;
+
+  DBUG_EXECUTE_IF("write_to_tmp_file_fail",
+                  return DB_TEMP_FILE_WRITE_FAIL;);
   MEM_UNDEFINED(&m_block[0], srv_sort_buf_size);
   return DB_SUCCESS;
 }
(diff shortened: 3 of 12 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -2193,6 +2193,9 @@ convert_error_code_to_mysql(
 		return(HA_ERR_RECORD_FILE_FULL);
 
 	case DB_TEMP_FILE_WRITE_FAIL:
+		innodb_transaction_abort(thd,
+					 innobase_rollback_on_timeout,
+					 error);
```
dr-m, line 2198: Is this change really necessary? It is not present in #3903. I suggested this when we were discussing this. As far as I understand, this error could be flagged only during the execution of `copy_data_between_tables()` in `ALTER TABLE…ALGORITHM=COPY` or similar statements, but not during DML. For DML, we realized that if a write to a temporary file fails in trx_t::apply_log()`, then we should flag an error to the `ha_innobase::commit_inplace_alter_table()` thread and let the DML transaction commit normally.

dr-m (commented):

I would suggest to include the fix of [MDEV-18455](https://jira.mariadb.org/browse/MDEV-18455) in the same branch, so that both fixes can be conveniently tested in conditions where `errno=ENOSPC` will be triggered frequently.

storage/innobase/handler/ha_innodb.cc
Thirunarayanan: This DB_TEMP_FILE_WRITE_FAIL error could happen during COPY  algorithm or bulk insert. This code addresses
the issue only for these scenario. For online alter, concurrent DML just marks the index as corrupted and move on.

storage/innobase/row/row0merge.cc
```diff
@@ -5178,6 +5178,8 @@ dberr_t row_merge_bulk_t::write_to_tmp_file(ulint index_no)
                        buf->index->table->space->id))
     return DB_TEMP_FILE_WRITE_FAIL;
   MEM_UNDEFINED(&m_block[0], srv_sort_buf_size);
+  DBUG_EXECUTE_IF("write_to_tmp_file_fail",
+                  return DB_TEMP_FILE_WRITE_FAIL;);
```
dr-m, line 5182: Because we’re skipping the `MEM_UNDEFINED()` if `row_merge_write()` fails, also the fault injection should be before the `MEM_UNDEFINED()`.

storage/innobase/row/row0umod.cc
```diff
@@ -1144,7 +1144,8 @@ row_undo_mod_upd_exist_sec(
 			/* This is only legitimate when
 			rolling back an incomplete transaction
 			after crash recovery. */
-			ut_a(thr_get_trx(thr)->is_recovered);
+			ut_a(thr_get_trx(thr)->is_recovered
+			     || thr_get_trx(thr)->in_rollback);
```
dr-m, line 1148: Doesn’t `trx_t::in_rollback` always hold here? There should be an assertion about it in a code path that leads to here. It seems that the assertion needs to be removed altogether and the comment revised.

storage/innobase/row/row0upd.cc
```diff
@@ -1420,10 +1420,12 @@ row_upd_changes_ord_field_binary_func(
...
 					storing the field. */
 					ut_ad(!thr
-					      || thr->graph->trx->is_recovered);
+					      || thr->graph->trx->is_recovered
+					      || thr->graph->trx->in_rollback);
 					ut_ad(!thr
 					      || thr->graph->trx
-					         == trx_roll_crash_recv_trx);
+					         == trx_roll_crash_recv_trx
+					      || thr->graph->trx->in_rollback);
```
dr-m, line 1428: If `trx->is_recovered` holds, then also `trx->in_rollback` must hold here. The preceding comment (which you must revise) actually mentions rollback. I think that both assertions need to be replaced with the following:
```c++
ut_ad(!thr || thr->graph->trx->in_rollback);
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2166,6 +2166,11 @@ convert_error_code_to_mysql(
 		return(HA_ERR_RECORD_FILE_FULL);
 
 	case DB_TEMP_FILE_WRITE_FAIL:
+		/* This error can happen during
+		ALTER TABLE..ALGORITHM=COPY or bulk insert operation */
+		innodb_transaction_abort(thd,
+					 innobase_rollback_on_timeout,
+					 error);
```
dr-m, line 2173: Can it happen during CREATE…SELECT? Would it be accurate to replace `ALTER TABLE..ALGORITHM=COPY` with `copy_data_between_tables()`?

storage/innobase/btr/btr0cur.cc
```diff
@@ -3697,8 +3697,12 @@ btr_cur_optimistic_update(
...
 				   ULINT_UNDEFINED, heap);
 #if defined UNIV_DEBUG || defined UNIV_BLOB_LIGHT_DEBUG
+	/* Blob pointer can be null if we ran out of space while
+	allocating the extern page. Relax the assertion during
+	rollback as well */
 	ut_a(!rec_offs_any_null_extern(rec, *offsets)
-	     || thr_get_trx(thr) == trx_roll_crash_recv_trx);
+	     || thr_get_trx(thr) == trx_roll_crash_recv_trx
+	     || thr_get_trx(thr)->in_rollback);
 #endif /* UNIV_DEBUG || UNIV_BLOB_LIGHT_DEBUG */
```
dr-m, line 3706: Would the following work? Shouldn’t it be applied also to #3903?
```c++
	/* Null BLOBs can be the result of failed BLOB page allocation. */
	ut_a(!rec_offs_any_null_extern(rec, *offsets)
	     || thr_get_trx(thr)->in_rollback);
```

storage/innobase/btr/btr0cur.cc
```diff
@@ -3697,8 +3697,11 @@ btr_cur_optimistic_update(
 	*offsets = rec_get_offsets(rec, index, *offsets, index->n_core_fields,
 				   ULINT_UNDEFINED, heap);
 #if defined UNIV_DEBUG || defined UNIV_BLOB_LIGHT_DEBUG
+	/* Blob pointer can be null if we ran out of space while
+	allocating the extern page. Relax the assertion during
+	rollback as well */
 	ut_a(!rec_offs_any_null_extern(rec, *offsets)
-	     || thr_get_trx(thr) == trx_roll_crash_recv_trx);
+	     || thr_get_trx(thr)->in_rollback);
```
dr-m, line 3704: A better comment would be as follows:
```
	/* Blob pointer can be null if InnoDB was killed or
	ran out space while allocating a page. */
```
If you want to mention relaxing an assertion, you can do it in the commit message.

storage/innobase/row/row0umod.cc
```diff
@@ -1141,11 +1141,6 @@ row_undo_mod_upd_exist_sec(
...
 			ut_a(dict_table_has_atomic_blobs(index->table));
 
-			/* This is only legitimate when
-			rolling back an incomplete transaction
-			after crash recovery. */
-			ut_a(thr_get_trx(thr)->is_recovered);
-
 			/* The server must have crashed before
 			completing the insert of the new
 			clustered index entry and before
```
dr-m, line 1146: Also the first part of the following comment is needs to be revised:
InnoDB must have run out of space or been killed before …

storage/innobase/row/row0upd.cc
```diff
@@ -1415,15 +1415,13 @@ row_upd_changes_ord_field_binary_func(
...
-					ut_ad(!thr
-					      || thr->graph->trx->is_recovered);
-					ut_ad(!thr
-					      || thr->graph->trx
-					         == trx_roll_crash_recv_trx);
+					storing the field (or)
+					encountered DB_OUT_OF_FILE_SPACE
+					while allocating the external page */
+					ut_ad(thr)
+					ut_ad(thr->graph->trx->in_rollback);
```
dr-m, line 1424: The first line is missing `;`

PR #3903, opened 2025-03-20
MDEV-36017 Alter table aborts when temporary directory is full

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36017*

## Description
Problem:
=======
- During inplace algorithm, concurrent DML fails to write the log operation into the temporary file. InnoDB fail to mark the error for the online log.

Fix:
===
row_log_online_op(): Mark the error in online log when InnoDB ran out of temporary spac

## How can this PR be tested?
./mtr innodb.alter_temp_fail

## PR quality check

### DIFF
--- a/sql/ddl_log.cc
+++ b/sql/ddl_log.cc
@@ -3029,13 +3029,15 @@ static bool ddl_log_write(DDL_LOG_STATE *ddl_state,
   error= ((ddl_log_write_entry(ddl_log_entry, &log_entry)) ||
           ddl_log_write_execute_entry(log_entry->entry_pos,
                                       &ddl_state->execute_entry));
-  mysql_mutex_unlock(&LOCK_gdl);
+  DBUG_EXECUTE_IF("ddl_log_write_fail", error= true;);
   if (error)
   {
     if (log_entry)
       ddl_log_release_memory_entry(log_entry);
+    mysql_mutex_unlock(&LOCK_gdl);
     DBUG_RETURN(1);
   }
+  mysql_mutex_unlock(&LOCK_gdl);
   add_log_entry(ddl_state, log_entry);
   ddl_state->flags|= ddl_log_entry->flags;      // Update cache
   DBUG_RETURN(0);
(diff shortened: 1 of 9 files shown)


dr-m (commented):

I would suggest to include the fix of [MDEV-18455](https://jira.mariadb.org/browse/MDEV-18455) in the same branch, so that both fixes can be conveniently tested in conditions where errno=ENOSPC will be triggered frequently.

sql/ddl_log.cc
```diff
@@ -3029,13 +3029,15 @@ static bool ddl_log_write(DDL_LOG_STATE *ddl_state,
...
-  mysql_mutex_unlock(&LOCK_gdl);
+  DBUG_EXECUTE_IF("ddl_log_write_fail", error= true;);
   if (error)
   {
     if (log_entry)
       ddl_log_release_memory_entry(log_entry);
+    mysql_mutex_unlock(&LOCK_gdl);
     DBUG_RETURN(1);
   }
+  mysql_mutex_unlock(&LOCK_gdl);
```
dr-m, line 3040: I was going to suggest to a comment that explains why `ddl_log_release_memory_entry()` needs to be protected by the mutex. Even better, add `mysql_mutex_assert_owner(&LOCK_gdl)` to the beginning of that function. But, I now see that the function already contains such an assertion. So, this is obviously correct and does not need a comment. This error handling code path was previously uncovered in any tests.

storage/innobase/btr/btr0cur.cc
```diff
@@ -4370,7 +4370,7 @@ btr_cur_pessimistic_update(
 					 cursor, offsets, offsets_heap,
 					 new_entry, &rec,
 					 &dummy_big_rec, n_ext, NULL, mtr);
-	ut_a(err == DB_SUCCESS);
+	if (err) return err;
```
dr-m, line 4373: Can you add a comment that explains when exactly this can happen? Something like the following, perhaps?
```c++
	if (UNIV_UNLIKELY(err != DB_SUCCESS)) {
		/* This should only fail when the log for online ALTER TABLE cannot be written. */
		ut_ad(dict_index_is_online_ddl(index));
		ut_ad(err == DB_TEMP_FILE_WRITE_FAIL);
		return err;
	}
```
I can hardly imagine that the scenario is like that, because I would expect that log to be written at transaction commit time, not here. Maybe we are running out of space when trying to extend the tablespace? If there is a specific error code for that, please add a debug assertion to check for it.

dr-m (approved):
OK to push after improving the `btr_cur_pessimistic_insert()` error handling with a comment and a debug assertion.

PR #3908, opened 2025-03-21
MDEV-36304  InnoDB: Missing FILE_CREATE, FILE_DELETE or FILE_MODIFY error during mariabackup --prepare

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36304*

## Description
Problem:
=======
- During --prepare of partial backup, if InnoDB encounters the redo log for the excluded tablespace then InnoDB stores the space id in dirty tablespace list during recovery, anticipates that it may encounter FILE_* redo log records in the future. Even though we encounter FILE_* record for the partial excluded tablespace then we fail to replace the name in dirty tablespace list. This lead to missing of FILE_* redo log records error.

Fix:
===
srv_prepare_partial_backup: Variable for the server to indicate whether the redo log has been applied on partial backup

xtrabackup_read_backup_info_file(): Reads the mariabackup_info file from target directory and sets srv_prepare_partial_backup value based on partial information exist in file.

fil_name_process(): Rename the file name from "" to name encountered during FILE_* records during prepare on partial backup directory

## How can this PR be tested?
./mtr mariabackup.partial_backup_prepare

## PR quality check


extra/mariabackup/xtrabackup.cc
```diff
@@ -2694,6 +2694,35 @@ static bool innodb_init()
...
+  if (!fp)
+  {
+    msg("Error: cannot open %s", filename);
+    return;
+  }
+
+  while (!feof(fp))
+  {
+    if (fscanf(fp, "%7s = %c\n", key, &value) == 2)
+      if (strcmp(key, "partial") == 0)
```
dr-m, line 2716: The two `if` could be combined, with `&&` between the expressions.

storage/innobase/log/log0recv.cc
```diff
@@ -1232,6 +1231,18 @@ static void fil_name_process(const char *name, ulint len, uint32_t space_id,
...
 
+			/* In case of prepare on partial backup,
+			InnoDB can't find some tablespace on target
+			directory. If InnoDB encounters redo log
+			for the tablespace before FILE_* record then
+			InnoDB should rename the file name in
+			recv_spaces from "" to name which encountered
+			during FILE_* record */
+			if (srv_prepare_partial_backup &&
+			    strcmp(fname.name.c_str(), "") == 0) {
```
dr-m, line 1242: We don’t need `strcmp` to compare an empty string. Actually, the idiomatic C++ equivalent would be `fname.name == ""`. But, it should be more efficient to just check the length.

This function follows the old formatting style, where lines are split before, not after, binary operators.
```c++
			if (srv_prepare_partial_backup
			    && fname.name.size() == 0) {
```

storage/innobase/log/log0recv.cc
```diff
@@ -1751,13 +1773,26 @@ dberr_t recv_sys_t::find_checkpoint()
         continue;
       }
 
+      DBUG_EXECUTE_IF("use_old_checkpoint",
+		      if (log_sys.next_checkpoint_lsn > 0) {
+		        if (checkpoint_lsn > log_sys.next_checkpoint_lsn)
+			  goto skip_lsn;
+			else goto assign_lsn;
+		       }
+		     );
```
dr-m, line 1782: This feels a little unsafe. What if there is only one usable checkpoint header in the file?

dr-m (commented):
Can you please try to revise the logic and the message in the case that a `FILE_` record is actually found, but a file by that name is missing? Currently, we can see a test failure:
```
2025-03-21 16:06:22 0 [Note] InnoDB: Starting crash recovery from checkpoint LSN=60184
2025-03-21 16:06:22 0 [Note] InnoDB: End of log at LSN=67725
2025-03-21 16:06:22 0 [ERROR] InnoDB: Missing FILE_CREATE, FILE_DELETE or FILE_MODIFY before FILE_CHECKPOINT for tablespace 4
[00] 2025-03-21 16:06:22 mariadb-backup: srv_start() returned 37 (Data structure corruption).
```
I would like the message to disclose more detail. Can we use a different marker than an empty file name when the file can’t be found? Could we preserve the latest file name that we found in the log, and output the name in case the file was missing? Only if the log records are missing (the log is corrupted), the above error message would be appropriate.

Thirunarayanan:
This issue was addressed in pr#3990

PR #3914, opened 2025-03-24
Add a sanity check for backups

fauust:
No MariaDB server code is touched here, this is only about GitHub actions CI.

### DIFF
--- a/.github/workflows/backup.yml
+++ b/.github/workflows/backup.yml
@@ -22,27 +22,50 @@ jobs:
           sudo apt-get -y install restic
       - name: Trigger backup export
         run: |
+          # needed for sanity check
+          NUMBER=$(gh api \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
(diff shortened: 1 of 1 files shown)


.github/workflows/backup.yml
```diff
@@ -22,34 +22,58 @@ jobs:
...
+            --jq '.archive_url')
           curl -L -H "Accept: application/vnd.github+json" \
             -H "Authorization: Bearer $GH_TOKEN" \
             -H "X-GitHub-Api-Version: 2022-11-28" \
             -o "archive.tgz" "$ARCHIVE_URL"
+      - name: Sanity check
+        run: |
+          # Make sure that we have the latest PR information
+          # title and corresponding number (from URL)
+          zgrep -a -B3 "$LATEST_PR_TITLE" archive.tgz |
```
grooverdan, line 63: A grep of a tar file looks quite crude, but as long as it works.

grooverdan (approved):

.github/workflows/backup.yml
fauust: Works, thanks to `-a` ;)

PR #3916, opened 2025-03-24
MDEV-36236 Instant alter aborts when InnoDB fails to rollback instant operation correctly

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-_36236*

## Description
Problem:
========
- InnoDB does consecutive instant alter operation, first instant DDL fails, it fails to reset the old instant information in table during rollback. This lead to consecutive instant alter to have wrong assumption about the exisitng instant column information.

Fix:
====
dict_table_t::instant_column(): Duplicate the instant information field of the table. By doing this, InnoDB alter retains the old instant information and reset it during rollback operation

## How can this PR be tested?
./mtr innodb.instant_alter_debug

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/instant_alter_debug.test
+++ b/mysql-test/suite/innodb/t/instant_alter_debug.test
@@ -657,11 +657,19 @@ DROP TABLE t1;
 SET DEBUG_SYNC=RESET;
 
 --echo # End of 10.5 tests
-
 SELECT variable_value-@old_instant instants
 FROM information_schema.global_status
 WHERE variable_name = 'innodb_instant_alter_column';
 
-SET GLOBAL innodb_stats_persistent = @save_stats_persistent;
+CREATE TABLE t1(f1 INT, f2 TEXT)ENGINE=InnoDB;
+INSERT INTO t1 VALUES(1, 'a');
+ALTER TABLE t1 ADD COLUMN f3 TEXT FIRST;
+--error ER_INTERNAL_ERROR
+SET STATEMENT DEBUG_DBUG="+d,instant_insert_fail" FOR
+ALTER TABLE t1 DROP COLUMN f1;
+ALTER TABLE t1 DROP COLUMN f1;
+CHECK TABLE t1;
+DROP TABLE t1;
 
+SET GLOBAL innodb_stats_persistent = @save_stats_persistent;
 --echo # End of 10.6 tests
--- a/storage/innobase/handler/handler0alter.cc
+++ b/storage/innobase/handler/handler0alter.cc
@@ -621,6 +621,16 @@ inline bool dict_table_t::instant_column(const dict_table_t& table,
 	}
 
 	dict_index_t* index = dict_table_get_first_index(this);
+	if (instant) {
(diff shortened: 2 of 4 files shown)


mysql-test/suite/innodb/t/instant_alter_debug.test
```diff
@@ -664,4 +664,15 @@ WHERE variable_name = 'innodb_instant_alter_column';
 
 SET GLOBAL innodb_stats_persistent = @save_stats_persistent;
 
+CREATE TABLE t1(f1 INT, f2 TEXT)ENGINE=InnoDB;
+INSERT INTO t1 VALUES(1, 'a');
+ALTER TABLE t1 ADD COLUMN f3 TEXT FIRST;
+SET DEBUG_DBUG="+d,instant_insert_fail";
+--error ER_INTERNAL_ERROR
+ALTER TABLE t1 DROP COLUMN f1;
+SET DEBUG_DBUG="-d,instant_insert_fail";
```
dr-m, line 673: Can we avoid the problematic `-d` by writing the following?
```sql
--error ER_INTERNAL_ERROR
SET STATEMENT debug_dbug='+d,instant_insert_fail' FOR
ALTER TABLE t1 DROP COLUMN f1;
```

mysql-test/suite/innodb/t/instant_alter_debug.test
```diff
@@ -664,4 +664,15 @@ WHERE variable_name = 'innodb_instant_alter_column';
 
 SET GLOBAL innodb_stats_persistent = @save_stats_persistent;
 
+CREATE TABLE t1(f1 INT, f2 TEXT)ENGINE=InnoDB;
```
dr-m, line 667: I think that the `SET GLOBAL` is part of the after-test cleanup. I would disable the persistent statistics also for this test, to avoid any interference from automatic recalculation.

dr-m (commented):
Please analyze another failure in this area, which I will shortly post to [MDEV-36236](https://jira.mariadb.org/browse/MDEV-36236). Would it be fixed by the same,  or would it require something more?

storage/innobase/handler/handler0alter.cc
```diff
@@ -621,6 +621,16 @@ inline bool dict_table_t::instant_column(const dict_table_t& table,
...
 	dict_index_t* index = dict_table_get_first_index(this);
+	if (instant)
+	{
+	  instant->field_map= static_cast<field_map_element_t*>(
+            mem_heap_dup(heap, instant->field_map,
+                         (index->n_fields - index->first_user_field()) *
+                         sizeof *instant->field_map));
+	  instant= static_cast<dict_instant_t*>(
+            mem_heap_dup(heap, instant, sizeof *instant));
+	}
```
dr-m, line 632: This function uses the K&R style and TAB based formatting. Please follow that formatting here:
```c++
	dict_index_t* index = dict_table_get_first_index(this);
	if (instant) {
		instant->field_map= static_cast<field_map_element_t*>(
			mem_heap_dup(heap, instant->field_map,
```

dr-m (approved):

PR #3917, opened 2025-03-24
MDEV-32084: Assertion in best_extension_by_limited_search() ...

spetrunia:
When subquery with LEFT JOIN is converted into semi-join, it is possible to construct cases where the LEFT JOIN's ON expression refers to a table in the current select but not in the current join nest, for example, here "ON t4.col=t1.col" has this property:

  t1 SEMI JOIN (
    t2
    LEFT JOIN (t3 LEFT JOIN t4 ON t4.col=t1.col) ON expr
  )
Let's denote this as ON-EXPR-HAS-REF-OUTSIDE-NEST.

The optimizer handles LEFT JOINs like so:
- Outer join runtime requires that "inner tables follow outer" in any join order.
- Join optimizer enforces this by following table's dependencies as set in TABLE_LIST::dep_tables.
- The dep_tables are set in simplify_joins() according to the contents of ON expressions and LEFT JOIN structure.

However, the logic in simplify_joins() failed to account for possible ON-EXPR-HAS-REF-OUTSIDE-NEST. It assumed that references outside the current join nest could only be OUTER_REF_TABLE_BIT or RAND_TABLE_BIT.

The fix was to add the missing logic.

- [x] *The Jira issue number for this PR is: MDEV-32084*

### DIFF
--- a/sql/sql_select.cc
+++ b/sql/sql_select.cc
@@ -18628,6 +18628,8 @@ simplify_joins(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool top,
         prev_table->dep_tables|= used_tables;
       if (prev_table->on_expr)
       {
+        /* If the ON expression is still there, it's an outer join */
+        DBUG_ASSERT(prev_table->outer_join);
         prev_table->dep_tables|= table->on_expr_dep_tables;
         table_map prev_used_tables= prev_table->nested_join ?
 	                            prev_table->nested_join->used_tables :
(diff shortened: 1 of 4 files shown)


sanja-byelkin:
OK to push after adding DBUG_ASSERT we have discussed.

sql/sql_select.cc
```diff
@@ -18612,11 +18612,57 @@ simplify_joins(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool top,
...
+                LEFT JOIN (t3 LEFT JOIN t4 ON t4.col=t1.col) ON expr
+              ) ON ...
+
+            here, we would have prev_table=t4, table=t3.  The condition
+            "ON t4.col=t1.col" depends on tables {t1, t4}. To make sure the
+            optimizer puts t3 before t4 we need to make sure t4.dep_tables
+            includes t3.
+          */
+          if (table->embedding)
+          {
```
sanja-byelkin, line 18641: Here probably we should assert left join

sanja-byelkin (approved):
Add the assert

sql/sql_select.cc
```diff
@@ -18642,11 +18644,59 @@ simplify_joins(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool top,
...
+              in the process of recomputing that value.
+              So, we walk the join list and collect the bitmap of peers:
+            */
+            table_map peers= 0;
+            List_iterator_fast<TABLE_LIST> li(*join_list);
+            TABLE_LIST *peer;
+            while ((peer= li++))
+            {
+              table_map curmap= peer->nested_join
+                                    ? peer->nested_join->used_tables
```
Olernov, line 18688: Such formatting of ternary operator is unusual, in other places of code `?` is not carried to another line

sql/sql_select.cc
```diff
@@ -18642,11 +18644,59 @@ simplify_joins(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool top,
...
+              ) ON ...
+
+            here, we would have prev_table=t4, table=t3.  The condition
+            "ON t4.col=t1.col" depends on tables {t1, t4}. To make sure the
+            optimizer puts t3 before t4 we need to make sure t4.dep_tables
+            includes t3.
+          */
+
+          DBUG_ASSERT(table->embedding == prev_table->embedding);
+          if (table->embedding)
```
Olernov, line 18674: By the way, doesn't it make sense to add `DBUG_ASSERT(prev_table->embedding == table->embedding)`? At least would give a useful hint to a reader

PR #3926, opened 2025-03-26
MDEV-33167  ASAN errors in dict_sys_t::load_table / get_foreign_key_info after failing to load a table

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-33167*

## Description
Problem:
=======
- While loading the foreign key constraints for the parent table, if child table wasn't open then InnoDB uses the parent table heap to store the child table name in fk_tables list. If the consecutive foreign key relation for the parent table fails with error, InnoDB evicts the parent table from memory. But it fails to clear the fk_list.

Solution:
========
dict_load_table_one(): Clear the fk_tables list before evicting the table from memory.

## How can this PR be tested?

./mtr innodb.foreign_key

## PR quality check


mysql-test/suite/innodb/t/foreign_key.test
```diff
@@ -1245,8 +1244,30 @@ ALTER TABLE t2 ADD KEY(b), ALGORITHM=NOCOPY;
...
-SET GLOBAL innodb_stats_persistent = @save_stats_persistent;
+set GLOBAL innodb_fast_shutdown=0;
+--let $shutdown_timeout=
+--source include/restart_mysqld.inc
+ALTER TABLE t2 FORCE;
+DROP TABLE t2, t1, t3;
+
+--echo # End of 10.6 tests
 
 --source include/wait_until_count_sessions.inc
```
dr-m, line 1273: Because we restarted the server, there is no need to count the sessions anymore.

storage/innobase/dict/dict0load.cc
```diff
@@ -2456,6 +2456,7 @@ static dict_table_t *dict_load_table_one(const span<const char> &name,
 		ib::error() << "Refusing to load corrupted table "
 			    << table->name;
 evict:
+		fk_tables.clear();
```
dr-m, line 2459: The modified test `innodb.foreign_key` is not failing for me on 1b9d5cdb83f235f313ae19bab1c6978837a566d7 even if I omit this code change.

I may be mistaken here, but it would seem to me that if there is a larger bunch of tables that are connected via `FOREIGN KEY` constraints, and only one table is causing trouble, we had better only remove those entries that had been added by this invocation of `dict_load_table_one()`. Neither the commit message nor the test case are explaining for which table this added code is expected to matter.

Thirunarayanan: Please run it in asan/valgrind build. I will address the comment in the revised patch.

storage/innobase/dict/dict0load.cc
```diff
@@ -2516,7 +2517,7 @@ static dict_table_t *dict_load_table_one(const span<const char> &name,
 		/* Don't attempt to load the indexes from disk. */
 	} else if (err == DB_SUCCESS) {
 		err = dict_load_foreigns(table->name.m_name, nullptr,
-					 0, true, ignore_err, fk_tables);
+					 0, true, ignore_err, fk_list);
 
 		if (err != DB_SUCCESS) {
 			ib::warn() << "Load table " << table->name
```
dr-m, line 2523: Instead of maintaining a local `std::deque`, could we simply save an iterator on `fk_tables` ? Something like the following?
```c++
auto i = fk_tables.end();
err = dict_load_foreigns(...);
if (err != DB_SUCCESS) {
	fk_tables.erase(i, fk_ tables.end());
```

dr-m (commented):
Thank you, this fix looks rather reasonable to me. We’re buffering the elements and only on success (or ignorable failure) filling in the caller-supplied `fk_tables`.

dr-m (approved):
If this passes the ASAN run, this should be OK.

PR #3929, opened 2025-03-27
MDEV-36360: Don't grab table-level X locks for applied inserts

denis-protivensky:
- [x] *The Jira issue number for this PR is: MDEV-36360*

## Description
It prevents a crash in wsrep_report_error() which happened when appliers would run with FK and UK checks disabled and erroneously execute plain inserts as bulk inserts.

Moreover, in release builds such a behavior could lead to deadlocks between two applier threads if a thread waiting for a table-level lock was ordered before the lock holder. In that case the lock holder would proceed to commit order and wait forever for the now-blocked other applier thread to commit before.

## Release Notes
Also fixed behavior of `wsrep_slave_FK_checks` and `wsrep_slave_UK_checks` variables to actually work for appliers.

## How can this PR be tested?

MTR test is provided.

## PR quality check

### DIFF
--- a/mysql-test/suite/galera_3nodes/r/MDEV-36360.result
+++ b/mysql-test/suite/galera_3nodes/r/MDEV-36360.result
@@ -1,0 +1,61 @@
+connection node_2;
+connection node_1;
+connection node_1;
+connection node_2;
+connection node_3;
+connection node_1;
+CREATE TABLE parent (
+id INT PRIMARY KEY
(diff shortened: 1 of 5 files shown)


sysprg:
Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/c01bff4a106326c633f9b899238ab5afe736f64a

PR #3931, opened 2025-03-27
MDEV-36359: Patch NULL deref after disabling semi-sync primary

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-36359](https://jira.mariadb.org/browse/MDEV-36359)*

### The Bug
Under `@@rpl_semi_sync_master_wait_no_slave = 0`, when `rpl_semi_sync_master_clients` decrements to zero, the primary reverts to async replication.
This code did not check whether Semi-Sync is still globally enabled or not as it didn’t matter before.

However, after #3089 ([MDEV-33551](https://jira.mariadb.org/browse/MDEV-33551)) split the transactions’ACK condition variables to per-transaction, this function now needs Semi-Sync’s transaction tracker to unblock these condition variables in batch, but this tracker is `NULL` when Semi-Sync Primary is disabled.

### Background
* FWIW, the Semi-Sync Primary subsystem can “orphanate” its Binlog Dump threads(’ connection to their replicas) when it disables (`SET @@rpl_semi_sync_master_enabled = 0`) while they remain connected.
  ```console
  $ SET @@GLOBAL.rpl_semi_sync_master_enabled = 0;
  $ SELECT STATE FROM information_schema.PROCESSLIST WHERE COMMAND = 'Binlog Dump';
  STATE
  Master has sent all binlog to slave; waiting for more updates
  $ SHOW STATUS LIKE 'Rpl_semi_sync_master_clients';
  Variable_name	Value
  Rpl_semi_sync_master_clients	1
  ```
* *The code for “automatic revert to async” is not an active function, but rather passive behaviors of various functions when the conditions are met.*

Stateful component are okay, but Schrödinger Boxes? Not so much.

## Release Notes
* Fix primary crash when it reverts to async replication after its Semi-Synchronous Replication disabled while still connected to replica(s).

## How can this PR be tested?
As demonstrated the included `rpl.rpl_semi_sync_master_disable_with_slave`:
1. Connect a Semi-Sync replica to a Semi-Sync, `@@rpl_semi_sync_master_wait_no_slave = 0` primary
2. Disable Semi-Sync on the primary
3. Disconnect the replica(’s IO thread) from replication
4. Observe the primary’s wellbeïng as the Binlog Dump thread exits
5. (optional) Reconnect the replica and confirm success

## PR quality check
  * This is a **crash** in a criticals-only version series.
    The higher-ups locked those branches against “unnecessary” pushes, so I added one as a assignee on this PR.

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.test
+++ b/mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.test
@@ -1,0 +1,68 @@
+# MDEV-36359: Master crashes when reverting to async after Semi-Sync disabled.
+#
+# Assert behavior of turning Semi-Sync off on
+# the master when still connected to a slave
+
+--source include/have_binlog_format_mixed.inc # format-agnostic
+
+--echo # Set up Semi-Sync with rpl_semi_sync_master_wait_no_slave=0
+--let $rpl_skip_start_slave= 1
+--source include/master-slave.inc
+
+--let $orig_master_enabled=`SELECT @@GLOBAL.rpl_semi_sync_master_enabled`
+SET @@GLOBAL.rpl_semi_sync_master_enabled= 1;
+--let $orig_wait_no_slave=`SELECT @@GLOBAL.rpl_semi_sync_master_wait_no_slave`
+SET @@GLOBAL.rpl_semi_sync_master_wait_no_slave= 0;
+
+--connection slave
+--let $orig_slave_enabled=`SELECT @@GLOBAL.rpl_semi_sync_slave_enabled`
+SET @@GLOBAL.rpl_semi_sync_slave_enabled= 1;
+--source include/start_slave.inc
+
+--connection master
+# Make sure Semi-Sync is active
+--let $status_var= Rpl_semi_sync_master_status
+--let $status_var_value= ON
+--source include/wait_for_status_var.inc
+
+--sync_slave_with_master
+--connection master
+
+--disable_cursor_protocol
+SELECT ID INTO @binlog_dump_tid
+  FROM information_schema.PROCESSLIST WHERE COMMAND = 'Binlog Dump';
+--enable_cursor_protocol
+
+--echo # Control State
+SELECT STATE FROM information_schema.PROCESSLIST WHERE ID = @binlog_dump_tid;
+SHOW STATUS LIKE 'Rpl_semi_sync_master_clients';
+
+--echo # Disable Semi-Sync while the dump thread is still connected to its slave
+SET @@GLOBAL.rpl_semi_sync_master_enabled = 0;
+--let $status_var_value= OFF
+--source include/wait_for_status_var.inc
+
+SELECT STATE FROM information_schema.PROCESSLIST WHERE ID = @binlog_dump_tid;
+SHOW STATUS LIKE 'Rpl_semi_sync_master_clients';
+
+--echo # Disconnect the slave and wait until the master's dump thread is gone
+--connection slave
+STOP SLAVE;
+# Starting with MDEV-13073,
+# Semi-Sync STOP SLAVE also terminates its dump thread on the master.
+--connection master
+
+# MDEV-36359: The disconnection would crash the master and leave the wait with
+# error 2013 'Lost connection to server during query'
+--let $wait_condition= SELECT COUNT(*)=0 FROM information_schema.PROCESSLIST WHERE ID = @binlog_dump_tid
+--source include/wait_condition.inc
+SHOW STATUS LIKE 'Rpl_semi_sync_master_clients';
+
+--echo # Cleanup
+--eval SET @@GLOBAL.rpl_semi_sync_master_enabled= $orig_master_enabled
+--eval SET @@GLOBAL.rpl_semi_sync_master_wait_no_slave= $orig_wait_no_slave
+--connection slave
+--eval SET @@GLOBAL.rpl_semi_sync_slave_enabled= $orig_slave_enabled
+
+--let $rpl_only_running_threads= 1
+--source include/rpl_end.inc
--- a/mysql-test/suite/rpl/r/rpl_semi_sync_master_disable_with_slave.result
+++ b/mysql-test/suite/rpl/r/rpl_semi_sync_master_disable_with_slave.result
@@ -1,0 +1,41 @@
+# Set up Semi-Sync with rpl_semi_sync_master_wait_no_slave=0
+include/master-slave.inc
+[connection master]
+SET @@GLOBAL.rpl_semi_sync_master_enabled= 1;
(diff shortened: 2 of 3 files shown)


mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.test
ParadoxV5, line 1: new file because uh appearently we test (side-)effects of `--rpl-semi-sync-master-wait-no-slave=0` only when bugs come up (else how’d this crashs slip past QA…)

mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.cnf
```diff
@@ -0,0 +1,8 @@
+!include ../my.cnf
+
+[mysqld.1]
+rpl_semi_sync_master_enabled=1
+rpl_semi_sync_master_wait_no_slave=0
+
+[mysqld.2]
+rpl_semi_sync_slave_enabled=1
```
ParadoxV5, line 8: Huh, does our style uses `-`s? Then why both these options only appear with `_`s in my greps?

ParadoxV5:
Oh BTW @bnestere a friendly ping since you prepared #3089 and I mentioned this changed file in our chat yesterday.

ParadoxV5:
* [ ] Reminder: forgot `Co-authored-by:` for the test

mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.cnf
knielsen: Either `-` or `_` works in configuration options

knielsen: Generally it is better to change the config inside the test case, as it avoids the need for mysql-test-run.pl to restart the server into the new configuration.

It is not a big issue, and if there is a reason for having a .cnf then it's fine, but I did not see in this case why it would be necessary?

mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.test
```diff
@@ -0,0 +1,41 @@
...
+
+SELECT STATE FROM information_schema.PROCESSLIST WHERE ID = @binlog_dump_tid;
+SHOW STATUS LIKE 'Rpl_semi_sync_master_clients';
+
+--echo # Disconnect the slave and wait until the master's dump thread is done
+--connection slave
+STOP SLAVE;
+--connection master
+
+# MDEV-36359: 2013 'Lost connection to server during query'
```
knielsen, line 31: This comment looks out of place here, some forgotten left-over?

mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.test
```diff
@@ -0,0 +1,41 @@
...
+--let $wait_condition= SELECT COUNT(*)=0 FROM information_schema.PROCESSLIST WHERE ID = @binlog_dump_tid
+--source include/wait_condition.inc
```
knielsen, line 33: Does this work and the dump thread disappears immediately when STOP SLAVE is done on the slave?

I thought the dump thread would be waiting for new data to appear in the binlog, and not notice the slave connection is gone until it tries to send new data to it. Are you sure this wait will not sit waiting for some time until some timeout triggers somewhere?

knielsen (commented):
Did you stress-test the testcase for a few hours running massively in parallel like I described in last review, to check that it is robust?

Please reword this part of the commit message:

"split the transactions’ ␆ semaphone to per-transaction, this function now needs Semi-Sync’s transaction tracker to unblock the ␆ semaphones in batch"

No need to use fancy unicode "ACK" here, and I think by "semaphone" you mean maybe semaphore? Better refer to the actual names in the code and use "mutex" as this is the primitive we're using.

mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.test
ParadoxV5: Well, that’s what happened for me, the dump thread starts terminating as soon as the replica disconnects (and, before this fix, leaves the `wait_condition.inc` with `Lost connection to server during query`).
It works in CI too.

Have you tried it out yourself?

mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.cnf
ParadoxV5: `mysql-test-run.pl` doesn’t restart the servers? Huh, I didn’t know that.

ParadoxV5 (commented):
> Did you stress-test the testcase for a few hours running massively in parallel like I described in last review, to check that it is robust?

Whoa? 😳 This test has nothing to do with timing; is it necessary?

> No need to use fancy unicode "ACK" here

stylistic choice 😁

> I think by "semaphone" you mean maybe semaphore?

Yes 🫠

> Better refer to the actual names in the code and use "mutex" as this is the primitive we're using.

By ‘semaphore’ I mean a conditional variable with mutex; ‘mutex’ would only be half of it. ‘Conditional variable’ it is, then.

knielsen:
> > Did you stress-test the testcase for a few hours running massively in parallel like I described in last review, to check that it is robust?
> 
> Whoa? 😳 This test has nothing to do with timing; is it necessary?

Yes. Better spend 2 minutes to start it running and leave it overnight, than risk wasting hours of other developer's time when their push is blocked by a sporadic test failure and they have to investigate.

mysql-test/suite/rpl/t/rpl_semi_sync_master_disable_with_slave.test
knielsen: > Did you find a different behavior when you tried it out yourself?

Interesting, this is something specific to semi-sync. Indeed, I have seen slave dump threads hanging around previously, but that was non-semi sync. Now I learned that semi-sync on the slave explicitly connects to the master and does a KILL CONNECTION on its dump thread, which I did not know about:
```c++
void Repl_semi_sync_slave::kill_connection(MYSQL *mysql) {
...
  size_t kill_buffer_length = my_snprintf(kill_buffer, 30, "KILL %lu",
                                        mysql->thread_id);
  mysql_real_query(kill_mysql, kill_buffer, (ulong)kill_buffer_length);
```
This is apparently some AliBaba patch merged in 2017.

ParadoxV5: 👍 a(n undocumented?) part of [MDEV-13073](https://jira.mariadb.org/browse/MDEV-13073)

knielsen (approved):
Hm, I thought I already approved this after fixing review comments?

PR #3933, opened 2025-03-28
MDEV-36420 Assertion failure in SET GLOBAL innodb_ft_aux_table

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36420*
## Description
`innodb_ft_aux_table_validate()`: If the table is found in InnoDB but not valid for the parameter, only invoke `dict_sys.unlock()` once.

This fixes a regression due to #3856.
## Release Notes
N/A. This regression will not be present in any release.
## How can this PR be tested?
```sh
./mtr innodb_fts.innodb_ft_aux_table
```
## PR quality check


mariadb-DebarunBanerjee (approved):
The patch looks correct. 

This is solving a double unlock dict_sys.unlock() which seems to be serious. It would be nice to write in the MDEV, the actual impact user could face in release mode binary.

PR #3934, opened 2025-03-28
MDEV-36167 Assertion in Item_sum_st with st_collect and group by

DaveGosselin-MariaDB:
The GIS function st_collect can now be used with GROUP BY.  Previously, such queries resulted in a crash because Item_func_stcollect did not specify quick_group = false and did not implement reset_field, update_field methods.

With quick_group as false, an implementation of Item_sum tells the SQL layer that the aggregate can be computed with the TemporaryTableWithPartialSums algorithm.  That is the case for st_collect.  This is convenient because the SQL layer will rely on the Item_sum to produce the correct, updated aggregate value for a given result field.  Otherwise, using reset_field and update_field implementations gives greater flexibility when computing the aggregate but requires more complexity.

### DIFF
--- a/mysql-test/main/spatial_utility_function_collect.combinations
+++ b/mysql-test/main/spatial_utility_function_collect.combinations
@@ -1,0 +1,9 @@
+[myisam]
+default-storage-engine=myisam
+
+[innodb]
+innodb
+default-storage-engine=innodb
+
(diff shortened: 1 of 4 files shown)


holyfoot (approved):
looks ok to push.

PR #3936, opened 2025-03-31
MDEV-32086 Failure to insert from derived table containing target table

mariadb-RexJohnston:
Pull request created in: https://jira.mariadb.org/browse/MDEV-32086


spetrunia:
"Failure to insert" is not descriptive... Please change the MDEV title its used in the patch to "Crash when trying to insert"...

Also when one encounters the resulting condition:
```
+  if (table_list->field_translation && !table_list->is_materialized_derived())
```
the idea that "ah, this table was merged but then we've reconsidered doing the merging but left field_translation behind" is not the first thing that comes to one's mind so please add a comment about this here...

(This does not mean that I consider the patch to be actually correct... no idea about that, yet)

mariadb-RexJohnston:
> "Failure to insert" is not descriptive... Please change the MDEV title and its use in the patch to e.g. `Crash when trying to insert ...`

Done
> +  if (table_list->field_translation && !table_list->is_materialized_derived())
> the idea that "ah, this table was merged but then we've reconsidered doing the merging but left field_translation behind" is not the first thing that comes to one's mind so please add a comment about this here...

Done

> An un-informed question: so if we have reconsidered merging, do we need table_list->field_translation at all? What if one sets it to NULL?

2nd execution problems result.  They are solvable, but not in such a simple manner.

sql/sql_base.cc
```diff
@@ -1184,7 +1184,9 @@ TABLE_LIST* find_dup_table(THD *thd, TABLE_LIST *table, TABLE_LIST *table_list,
     DBUG_PRINT("info",
                ("found same copy of table or table which we should skip"));
   }
-  if (res && res->belong_to_derived)
+  // INSERT SELECT has better dealing with duplicates
+  if (res && res->belong_to_derived &&
+      !(check_flag & CHECK_DUP_FOR_INSERT_SELECT))
```
spetrunia, line 1189: Note the comment right below:

```
 We come here for queries of type:
      INSERT INTO t1 (SELECT tmp.a FROM (select * FROM t1) as tmp);
```
this is no longer true after this patch. The comment needs to be fixed.
(there will be more input).

PR #3940, opened 2025-04-03
MDEV-36469 don't check is_infoschema_db for null db

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36469*

## Description

The is_infoschema_db is a deep character set based comparison. In in many cases the db is still
an empty structure.

Doing this comparison early prevents a UBSAN error
by not performing character set operations on
a null pointer.

## Release Notes

Optimise information_schema check in parser.

## How can this PR be tested?

UBSAN builders (on update) will fail tests per multiple entries listed in MDEV.

Earlier branches have very different `Lex_ident_db` interface therefore a difference solution would be needed for those.

## PR quality check

### DIFF
--- a/sql/lex_ident.h
+++ b/sql/lex_ident.h
@@ -165,6 +165,7 @@ public:
 */
 class Lex_ident_db: public Lex_ident_fs
 {
+public:
   bool is_null() const
   {
     return length == 0 && str == NULL;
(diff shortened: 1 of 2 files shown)


abarkov (approved):
The patch look good for me. Ok to push. Thank you!

PR #3941, opened 2025-04-03
MDEV-20065 parallel replication for galera slave

temeo:
- [x] *The Jira issue number for this PR is: MDEV-20065*

## Description

When replicating transactions from parallel slave replication processing, Galera must respect the commit order of the parallel slave replication. In the current implementation this is done by calling `wait_for_prior_commit()` before the write set is replicated and certified in before-prepare processing. This however establishes a critical section which is held over whole Galera replication step, and the commit rate will be limited by Galera replication latency.

In order to allow concurrency in Galera replication step, the critical section must be released at earliest point where Galera can guarantee sequential consistency for replicated write sets. This change passes a callback to release the critical section by calling `wakeup_subsequent_commits()` to Galera library, which will call the callback once the correct replication order can be established. This functionality will be available from Galera 26.4.22 onwards.

Note that call to `wakeup_subsequent_commits()` at this stage is safe from group commit point of view as Galera uses separate `wait_for_commit` context to control commit ordering.

## How can this PR be tested?

This change is an optimization, and does not have associated MTR or unit tests.

In order to test, Start Galera Cluster and make one of the Galera nodes as a slave to MariaDB master server. 
By varying `slave_parallel_threads` it should be observed that the transaction throughput increases when
more parallel slave threads are added. For this to work, `slave_parallel_mode` should be set to `optimistic` or
`aggressive`. 

- [ x] *This is a new feature or a refactoring, and the PR is based against the `main` branch.*

## PR quality check

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -3998,6 +3998,16 @@ bool THD::wsrep_parallel_slave_wait_for_prior_commit()
   return false;
 }
 
+void wsrep_parallel_slave_wakeup_subsequent_commits(void *thd_ptr)
+{
+  THD *thd = (THD*)thd_ptr;
+  if (thd->rgi_slave && thd->rgi_slave->is_parallel_exec &&
+      thd->wait_for_commit_ptr)
+  {
+    thd->wait_for_commit_ptr->wakeup_subsequent_commits(0);
+  }
+}
+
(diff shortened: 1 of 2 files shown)


sysprg:
@temeo Thanks, the fix has been merged with the head revision: https://github.com/MariaDB/server/commit/07d71fdcf8a9df8bbfd4dd7127f4d7144f161135

PR #3943, opened 2025-04-03
MDEV-36150 Retain original internal 2PC XID in Galera Cluster

temeo:
- [x] *The Jira issue number for this PR is: MDEV-36150*

## Description

Prior to this change, Galera overwrote XID used in internal 2PC with a XID containing wsrep seqno and UUID. This however is not compatible with partition engine and atomic DDL recovery.

In order to avoid overwriting the XID, store the wsrep transaction seqno and UUID into Xid_log_event if the transaction is wsrep transaction. This way the seqno and UUID will be available to reconstruct the wsrep XID for commit during binlog coordinated recovery.

This changes on-disk format of Xid_log_event to contain additional 24 bytes for wsrep seqno and UUID if the transaction is wsrep transaction.

The Xid_log_event output in mariadb-binlog is extended to print also wsrep_seqno and wsrep_uuid if they are set.

Changes in on-disk formats:

* Wsrep seqno and UUID are added into Xid_log_event.

Changes in recovery:

* dry_run flag is set to false if wsrep node is recovering from SST without binlog enabled. This is needed to support the case where Donor runs with binlog enabled and Joiner does not have binlog configured. Without this change the Joiner could fail in recovery if InnoDB recovery would find a transaction in prepared state.

Changes in InnoDB:

* Fix wsrep_xid use in lock_wait_wsrep_kill(). The XID assigned to InnoDB trx was checked whether it is wsrep XID if the victim transaction was found in prepared state. This condition is now never fulfilled because wsrep XID is not used for 2PC. Let wsrep_thd_bf_abort() decide whether the victim should be BF aborted.
* Remove wsrep XID specific code from innobase_rollback(). Wsrep XID is not used for 2PC anymore, so it was dead code.
* Remove wsrep XID specific code from innobase_rollback_by_xid(). If a prepared transaction with wsrep XID is recovered by InnoDB recovery, the recovered wsrep XID is stored in xid_recovery_member making the wsrep specific code in innobase_rollback_by_xid() redundant.
## Release Notes

Prior to this change wsrep/Galera seqno associated to a transaction was visible in mariadb-binlog output as Xid if wsrep was enabled. The output of mariadb-binlog now contains wsrep seqno and uuid in Xid_log_event if the transaction was committed with wsrep enabled.

Example output:
```
#250403 19:08:52 server id 1  end_log_pos 1337 CRC32 0xeec7b894 	Xid = 51, wsrep_seqno = 7, wsrep_uuid = ea2f3364-10a5-11f0-870f-3f79f4ea94c7
```

## How can this PR be tested?

MTR tests in Galera suites.

## PR quality check


sql/handler.cc
```diff
@@ -2719,7 +2746,18 @@ static void xarecover_do_commit_or_rollback(transaction_participant *hton,
...
 
   if (xarecover_decide_to_commit(member, ptr_commit_max))
+  {
+#ifdef WITH_WSREP
+    if (member->wsrep_xid && wsrep_is_wsrep_xid(member->wsrep_xid))
+    {
+      wsrep_recovery_commit_xid= member->wsrep_xid;
+    }
+#endif /* WITH_WSREP */
     rc= hton->commit_by_xid(&x);
```
janlindstrom, line 2756: If I correctly read wsrep_recovery_commit_xid is used only on  wsrep_commit_xid and even there just to see if it is set. Furthermore, x used on commit_by_xid is it same as member->wsrep_xid ?

Actual variable is safe if recovery is run single threaded.

sql/log.cc
```diff
@@ -12243,8 +12274,38 @@ void Recovery_context::process_gtid(int round, Gtid_log_event *gev,
     /* Update the binlog state with any 'valid' GTID logged after Gtid_list. */
     last_gtid_valid= true;    // may flip at Xid when falls to truncate
   }
+#ifdef WITH_WSREP
+  if (last_gtid.domain_id == last_wsrep_gtid_domain_id)
```
janlindstrom, line 12278: In what situation these could be different?

PR #3944, opened 2025-04-03
MDEV-35770 Augment DBUG_ASSERT with __builtin_assume

UltraGreed:
- [x] *The Jira issue number for this PR is: MDEV-35770*

## Description
The DBUG_ASSERT macro tests an invariant at runtime in debug build and aborts if the check fails. Although assertions are disabled in optimized builds, code is usually written in such a manner, that failing an assertion condition would lead to undefined behavior. It therefore makes sense to introduce platform-specific analog of C++23 `[[assume(cond)]]` and use it in these cases, which potentially could lead to generation of more optimized builds in production.
E.g. compiler would be able to optimize `while` loop into `do-while` loop (which means less instructions in assembly) if it knows that the loop will run at least once.

## Special patterns case
There are lots of code fragments like:
```cpp
DBUG_ASSERT(!thd->in_sub_stmt);
if (thd->in_sub_stmt)  {
  ...
}
```
and
```cpp
switch (i)
{
  case 1: x=myisam_sint1korr(from); break;
  case 2: x=myisam_sint2korr(from); break;
  case 3: x=myisam_sint3korr(from); break;
  case 4: x=myisam_sint4korr(from); break;
  default: DBUG_ASSERT(0); x= 0;
}
```
If we simply wrap DBUG_ASSERT in an assume, the compiler will elide both the if-guard and the default branch — removing important fallback code.

There are also cases like:
```cpp
  void acquire()
  {
    IF_DBUG(auto prev= ,)
    ref_count.fetch_add(1);
    DBUG_ASSERT(prev != 0);
  }
```
In this case compilation in release mode would fail, because assume is called with undeclared variable.

This PR:
1. Introduces new platform-dependent macro MY_ASSUME, implemented for GCC >= 13.0, Clang and MSVC.
2. Augments DBUG_ASSERT to use it by default in DBUG. Notably, it will enable MY_ASSUME in the release mode as well.
3. Provides DBUG_ASSERT_NO_ASSUME (identical to the old DBUG_ASSERT) for described above **special** cases.

MY_ASSUME is implemented for GCC >= 13.0, Clang and MSVC.

~~TODO: Measure actual performance improvements and document LLVM/assembly diffs.~~ -- will stay out of scope for this PR

## How can this PR be tested?
This patch doesn't intend to change the behaviour, so it's only important that the existing tests will pass. We assume that test coverage of the **special** cases is out of scope for this work -- it was supposed to have been done in scope of the same task when it was added, if that was feasible at all.  

We acknowledge we may have missed a few “special” patterns, but we’ve covered all the occurrences we could find.

## PR quality check

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,6 +273,9 @@ ENDIF()
 # Be nice to profilers etc
 MY_CHECK_AND_SET_COMPILER_FLAG("-fno-omit-frame-pointer" RELWITHDEBINFO)
 
+# Assert-as-assume makes arbitrary exprs appear in assumptions, causing warnings
+MY_CHECK_AND_SET_COMPILER_FLAG("-Wno-assume")
+
 # enable security hardening features, like most distributions do
 # in our benchmarks that costs about ~1% of performance, depending on the load
 OPTION(SECURITY_HARDENED "Use security-enhancing compiler features (stack protector, relro, etc)" ON)
--- a/client/mysqldump.cc
+++ b/client/mysqldump.cc
@@ -2530,9 +2530,7 @@ static void print_xml_row(FILE *xml_file, const char *row_name,
       {
         create_stmt_ptr= (*row)[i];
         create_stmt_len= lengths[i];
-#ifdef DBUG_ASSERT_EXISTS
         body_found= 1;
-#endif
       }
       else
       {
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -1950,7 +1950,7 @@ ATTRIBUTE_NORETURN static void cleanup_and_exit(int exit_code,
     break;
     default:
       printf("unknown exit code: %d\n", exit_code);
-      DBUG_ASSERT(0);
+      DBUG_ASSERT_NO_ASSUME(0);
     }
   }
 
@@ -11016,7 +11016,7 @@ static void append_session_track_info(DYNAMIC_STRING *ds, MYSQL *mysql)
       }
       else
       {
-        DBUG_ASSERT(0);
+        DBUG_ASSERT_NO_ASSUME(0);
         dynstr_append_mem(ds, STRING_WITH_LEN("Tracker???\n"));
       }
     }
--- a/include/my_compiler.h
+++ b/include/my_compiler.h
@@ -29,6 +29,8 @@
   Compiler-dependent internal convenience macros.
 */
 
+#include <stddef.h>
+
 /* C vs C++ */
 #ifdef __cplusplus
 #define CONSTEXPR constexpr
@@ -79,6 +81,25 @@
 # define MY_ALIGNOF(type)   __alignof__(type)
 /** Determine the alignment requirement of a type. */
 # define MY_ALIGNED(n)      __attribute__((__aligned__((n))))
+#endif
+
+/*
+   An analogue of c++23 [[assume(cond)]]
+   for GCC >= 13.0, Clang and MSVC.
+*/
+#if defined __GNUC__ && __GNUC__ >= 13
+#ifdef __cplusplus
+# define MY_ASSUME(A) [[gnu::assume(!!(A))]]
+#else
+// In gnu-c, statement attributes are ambiguous.
+# define MY_ASSUME(A) do { [[gnu::assume(!!(A))]]; } while (0)
+#endif
+#elif defined __clang__
+# define MY_ASSUME(A) __builtin_assume(!!(A))
+#elif defined _MSC_VER
+# define MY_ASSUME(A) __assume(!!(A))
+#else
+# define MY_ASSUME(A) do { } while(0)
 #endif
 
 /**
--- a/include/my_dbug.h
+++ b/include/my_dbug.h
@@ -17,6 +17,8 @@
 #ifndef _my_dbug_h
 #define _my_dbug_h
 
+#include "my_compiler.h"
+
 #ifndef _WIN32
 #include <signal.h>
 #endif
@@ -116,9 +118,10 @@ extern int (*dbug_sanity)(void);
 #define DBUG_END()  _db_end_ ()
 #define DBUG_LOCK_FILE _db_lock_file_()
 #define DBUG_UNLOCK_FILE _db_unlock_file_()
-#define DBUG_ASSERT(A) do { \
+#define DBUG_ASSERT_NO_ASSUME(A) do { \
   if (unlikely(!(A)) && _db_my_assert(__FILE__, __LINE__, #A)) assert(A); \
 } while (0)
+#define DBUG_ASSERT(A) DBUG_ASSERT_NO_ASSUME(A)
 #define DBUG_SLOW_ASSERT(A) DBUG_ASSERT(A)
 #define DBUG_ASSERT_EXISTS
 #define DBUG_EXPLAIN(buf,len) _db_explain_(0, (buf),(len))
@@ -200,11 +203,13 @@ extern void _db_suicide_(void);
 
 #ifdef DBUG_ASSERT_AS_PRINTF
 extern void (*my_dbug_assert_failed)(const char *assert_expr, const char* file, unsigned long line);
-#define DBUG_ASSERT(assert_expr) do { if (!(assert_expr)) { my_dbug_assert_failed(#assert_expr, __FILE__, __LINE__); }} while (0)
+#define DBUG_ASSERT_NO_ASSUME(assert_expr) do { if (!(assert_expr)) { my_dbug_assert_failed(#assert_expr, __FILE__, __LINE__); }} while (0)
+#define DBUG_ASSERT(assert_expr) DBUG_ASSERT_NO_ASSUME(assert_expr)
 #define DBUG_ASSERT_EXISTS
 #define IF_DBUG_ASSERT(A,B)             A
 #else
-#define DBUG_ASSERT(A)                  do { } while(0)
+#define DBUG_ASSERT(A)                  MY_ASSUME(A)
+#define DBUG_ASSERT_NO_ASSUME(A)        do { } while (0)
 #define IF_DBUG_ASSERT(A,B)             B
 #endif /* DBUG_ASSERT_AS_PRINTF */
 #endif /* !defined(DBUG_OFF) */
--- a/libmysqld/emb_qcache.cc
+++ b/libmysqld/emb_qcache.cc
@@ -399,7 +399,7 @@ void emb_store_querycache_result(Querycache_stream *dst, THD *thd)
       }
     }
   }
-  DBUG_ASSERT(emb_count_querycache_size(thd) == dst->stored_size);
+  DBUG_ASSERT_NO_ASSUME(emb_count_querycache_size(thd) == dst->stored_size);
   DBUG_VOID_RETURN;
 }
 
--- a/libmysqld/libmysql.c
+++ b/libmysqld/libmysql.c
@@ -4424,7 +4424,7 @@ static void stmt_update_metadata(MYSQL_STMT *stmt, MYSQL_ROWS *data)
   {
     if (!(*null_ptr & bit))
       (*my_bind->skip_result)(my_bind, field, &row);
-    DBUG_ASSERT(row <= row_end);
+    DBUG_ASSERT_NO_ASSUME(row <= row_end);
     if (!(bit= (uchar) (bit << 1)))
     {
       bit= 1;					/* To next uchar */
--- a/mysys/charset.c
+++ b/mysys/charset.c
@@ -1542,7 +1542,7 @@ static UINT get_codepage(const char *s)
   UINT cp;
   if (s[0] != 'c' || s[1] != 'p')
   {
-    DBUG_ASSERT(0);
+    DBUG_ASSERT_NO_ASSUME(0);
     return 0;
   }
   cp= strtoul(s + 2, NULL, 10);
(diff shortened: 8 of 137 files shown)


svoj:
Hi @UltraGreed! What is the intention of this pull request? Why is it submitted under MDEV-35770 umbrella? What kind of feedback are you expecting from MariaDB developers?

svoj:
@FooBarrior is mentoring this effort.

UltraGreed:
Hello, @svoj. Sorry for the delayed response. I've updated the PR to better reflect its intention. The only commit uploaded so far is a small fix (related to described in the PR problem), which will be squashed into the main commit later.

svoj:
Reset draft status as the ball is in mentor's hands.

FooBarrior (approved):
I have made the (minor) changes I wanted, now this looks good to me.

@svoj please also take a look as a second reviewer

FooBarrior:
as a good example where it clearly has an effect, I would provide `translog_mark_file_finished` in `ma_loghandler.c`.

There we could see, that not only clang 21 does a loop inversion, but also a partial loop unrolling, even under -O2

```
	movq	log_descriptor+8391888(%rip), %rax
	cmpl	%ebx, (%rax) // <------------- check the first element agressively
	jne	.LBB92_5
	xorl	%ecx, %ecx
	jmp	.LBB92_8
.LBB92_5:
	movl	$1, %ecx  
	cmpl	%ebx, 8(%rax) // <------------- check the second element agressively
	je	.LBB92_8
	movl	$1, %edx
.LBB92_7:
	leaq	1(%rdx), %rcx // <------------ start from 2:  rcx (i) = rdx + 1 (т.е. 1+1=2)
	cmpl	%ebx, 8(%rax,%rdx,8)
	movq	%rcx, %rdx
	jne	.LBB92_7
```

vaintroub:
please take a look on Windows optimized build. CI won't block that for you, on whatever reason, but [appveyor does actually complain.](https://ci.appveyor.com/project/rasmushoj/server/builds/53053467)
``
C:\projects\server\mysys\my_virtual_mem.c(71,5): error C2220: the following warning is treated as an error
    DBUG_ASSERT(is_memory_committed(ptr, size));
    ^
C:\projects\server\mysys\my_virtual_mem.c(71,5): warning C4013: 'is_memory_committed' undefined; assuming extern returning int
``

FooBarrior:
@vaintroub can you tell something about appveyor results now -- is it somehing common, or is it this PR's issue?

vaintroub:
If this does not happen elsewhere, e.g in the base branch for this one, but happens here, then it is PR issue.
If you look at what it complains (destructor never returns), and look at the corresponding code, it might appear obvious (not for GCC, though)

```c++
 virtual ~Sql_cmd()
 {
   /*
     Sql_cmd objects are allocated in thd->mem_root.
     In MySQL, the C++ destructor is never called, the underlying MEM_ROOT is
     simply destroyed instead.
     Do not rely on the destructor for any cleanup.
   */
   DBUG_ASSERT(false);
 }
```

svoj:
> If this does not happen elsewhere, e.g in the base branch for this one, but happens here, then it is PR issue. If you look at what it complains (destructor never returns), and look at the corresponding code, it might appear obvious (not for GCC, though)
> 
> ```c++
>  virtual ~Sql_cmd()
>  {
>    /*
>      Sql_cmd objects are allocated in thd->mem_root.
>      In MySQL, the C++ destructor is never called, the underlying MEM_ROOT is
>      simply destroyed instead.
>      Do not rely on the destructor for any cleanup.
>    */
>    DBUG_ASSERT(false);
>  }
> ```

Should this be rewritten as something like `~Sql_cmd()= delete;`? Or why don't we move `Sql_cmd` cleanup towards destructor and call it explicitly?

FooBarrior:
> Should this be rewritten as something like ~Sql_cmd()= delete;?

yes, and perhaps remove `virtual`?

FooBarrior:
Hmm I guess we can't do it @svoj.
`g++ - <<< 'struct A{ A(){}; ~A()=delete; }; A *a = new A; int main(){}'`
builds just fine.

However adding derivation breaks it:
```
g++ -xc++ - <<< 'struct A{ A(){}; ~A()=delete; }; 
struct B:A{ B():A(){}; int x; ~B()=delete; };
 B *a = new B; int main(){}'
<stdin>: In constructor ‘B::B()’:
<stdin>:1:51: error: use of deleted function ‘A::~A()’
<stdin>:1:18: note: declared here
```

FooBarrior:
So our best bet as i see it is to make those assertions _NO_ASSUME. It seems that C++ requires a reachable destructor (even though we know it will not be reachable,).

svoj:
A few concerns from my side, mostly inline with [assert != assume whitepaper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2064r0.pdf) shared by @FooBarrior.

1. It is primarily advertised as a performance optimisation. Though I do trust that `assume` may give some moderate performance improvement, it has to be proven by benchmarks. E.g. Xiang Fan mentioned in the whitepaper:
> Much to our surprise, the C1xx front end actually got a 1-2% faster with __assume statements removed.

2. There is no clear explanation how to choose between `DBUG_ASSERT_NO_ASSUME()` and `DBUG_ASSERT()`. Even if we come up with one, it most probably going to be quite sophisticated document. I would expect broad misuse of `DBUG_ASSERT()`, which may lead to p.4.

3. There is no way to disable it at compile time. Although this is easily solvable.

4. New set of potential (practically undebuggable) issues in release builds. These are discussed in the whitepaper.

Unless there's a clear performance improvement, I'd definitely vote against "assert is assume" approach. I'd even be very cautious with "assume is assert" approach advertised by the whitepaper: too easy to misuse.

But the good thing is that now we know places where `while` can be replaced with `do ... while`.

vuvova:
I generally disagree with the [assert != assume whitepaper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2064r0.pdf), I think they use asserts incorrectly, and think that assert = assume.

Still, the point on whether it will actually improve the performance is valid. If it doesn't help, it's just useless code change that only makes it more complex.

svoj:
@vuvova we have ~10000 `DBUG_ASSERT` occurrences, ~300 were converted to `DBUG_ASSERT_NO_ASSUME()`, which already gives 3% cases that didn't work with assume. And I guess we'll get some more as PR appears to be still WiP.

A few examples...

debug assert = debug assume, but debug assert != release assume, because it depends on some DBUG specific code.
```
static void remove_ptr_from_dynarray(DYNAMIC_ARRAY *array, void *ptr)
{
  bool found __attribute__((unused))= false;
  for (size_t i= 0; i < array->elements; i++)
  {
    if (ptr == *dynamic_element(array, i, void**))
    {
      DBUG_ASSERT(!found);
      delete_dynamic_element(array, i);
      IF_DBUG_ASSERT(found= true, break);
    }
  }
  DBUG_ASSERT(found);
}
```

Somewhat unrelated, but still... here we don't want to assume mutex is locked, we want to terminate execution. It generally applies to all `ut_a()` calls.
```
ut_a(!pthread_mutex_lock(&m_mutex));
```

We may want to interrupt debug builds on incorrect inputs, but let release builds handle it. We don't want this handling to be optimised away.
```
my_off_t my_get_ptr(uchar *ptr, size_t pack_length)
{
  my_off_t pos;
  switch (pack_length) {
#if SIZEOF_OFF_T > 4
  case 8: pos= (my_off_t) mi_uint8korr(ptr); break;
  case 7: pos= (my_off_t) mi_uint7korr(ptr); break;
  case 6: pos= (my_off_t) mi_uint6korr(ptr); break;
  case 5: pos= (my_off_t) mi_uint5korr(ptr); break;
#endif
  case 4: pos= (my_off_t) mi_uint4korr(ptr); break;
  case 3: pos= (my_off_t) mi_uint3korr(ptr); break;
  case 2: pos= (my_off_t) mi_uint2korr(ptr); break;
  case 1: pos= (my_off_t) *(uchar*) ptr; break;
  default: DBUG_ASSER(0); return 0;
  }
 return pos;
}
```

This one wouldn't be correct assumption either, we want to interrupt here:
```
  /*
    If one day we will support execution of multi-statements
    through PS API we should not issue SHOW WARNINGS until
    we have not read all results...
  */
  DBUG_ASSERT(!mysql_more_results(mysql));
```

Yes, these can be incorrect `DBUG_ASSERT()` use patterns. But we have them, and we will be getting more of them. If we blindly replace them with assume in release builds we'll get code mined with bugs that can probably be debugged in assembler mode only. That's expensive.

vuvova:
1. agree, if we assert an expression that uses debug-only variables, it won't compile without debug, there's nothing to assume.
2. `ut_a(!pthread_mutex_lock(&m_mutex));` — I guess, it's not a debug assert at all, it's production "crash if not true" code.
3. seems safe to assume, `my_get_ptr()` is only used on values that were previously written by `my_store_ptr()`. In memory during this process execution time. It's a  correct usage of an assert, it shows that something that can **never** happen.
4. this is safe to assume as long as we don't support "multi-statements through PS API". When we will, the assert will fire in the debug more, we'll remove it, and this will remove the assume too. *Strictly speaking, it's a wrong assert, as it asserts values read over the socket from another process, but it's `mysqltest`, not a production code, within its domain the assert is fine*

Here only case 1 is a valid reason not to assume asserts

gkodinov (changes requested):
This is a preliminary review. Can you please address the comments by the reviewer? There's also failing build tests that need to be fixed.

gkodinov (changes requested):
I have played a bit with the diff: merged it, cleaned it up etc.

The only thing failing is the clang memsan build atm. And it's failing massively.  I didn't prove it, but my guess is that it's now checking something that depends on the data structures being set to 0 in debug mode and not initialized otherwise. Or there's some side effect of the function calls in DBUG_ASSERT().

I would join @svoj in requesting a benchmark. 

I've grep-ed through the DBUG_ASSERTs: there's about 1.5k of these. A serious amount of them are calling functions. Some of them quite heavy. So having these in assume() I believe doesn't contribute to the performance speedup, but further slows the run in optimized mode since it now needs to evaluate these functions. Also a good % of the ASSERTs do not take advantage of assuming the things that are asserted: most of them are double-checking state of complex data structures before executing an action on them.

So, while I understand and can relate to the basic premise of the PR, I believe that we need to take a different approach: instead of doing that to all DBUG_ASSERTs by default we need to have a DBUG_ASSERT_ASSUME() instead, then evaluate each case where DBUG_ASSERT() is used and replace with _ASSUME() if there's going to be benefits.

IMHO, to have benefit from the assert_assume, the DBUG_ASSERT should:

 * not be calling any functions
 * there should be a similar condition right after the DBUG_ASSERT() or close to it. 
 * ideally the DBUG_ASSERT() should just be checking a single condition over a single variable.
 * The expression should reference a variable. constant expressions are not worth it.

FooBarrior:
@gkodinov,
>  I believe doesn't contribute to the performance speedup, but further slows the run in optimized mode since it now needs to evaluate these functions.

No, actually, everything in assume is supposed to never execute, especially uninlined function call. There was even a massively spamming warning that we had to suppress (-Wno-assume) about that. Something like "warning: function f used in assume statement will not be executed".

Indeed, there should be no side-effects of executing assume, so that's well expected.

I have another point **for** this change, and this was my initial demand that caused this project be:
**Assume-as-assert will help static analysis.** I see a lot of false-positive warnings in my IDE (which runs clang-tidy) caused by ignored assumptions, that are clear from assertions.

`translog_mark_file_finished` is a useful example one more time here:
```
  DBUG_ASSERT(log_descriptor.unfinished_files.elements > 0);
  for (i= 0; i < (int) log_descriptor.unfinished_files.elements; i++)
  {
    fc_ptr= ...
    ...
  }
  
  if (! --fc_ptr->counter)
```
clang-tidy reports `fc_ptr` may be uninitialized at the list line.

That's not the point for enabling it in the optimized builds, though, but to keep it as an option anyway, for example for clang-tidy configs, and perhaps in UBSan, as we've noticed that it catches these problems well.

That can also be a good start before making a decision of enabling it for optimized builds.

FooBarrior:
One more point:
`assume` eliminates the need for `__builtin_unused`/`IF_DBUG`. We have two patterns:
```
IF_DBUG(int result=,) my_something(...);
DBUG_ASSERT(result ==);
```
and 
```
int __attribute__ ((__unused__)) result = ...
DBUG_ASSERT(result == 0);
```
and sometimes it's
```
#ifndef DBUG_OFF
int result=
#endif
my_something(...);
DBUG_ASSERT(result ==);
```

One is creepier than another. All this is needed only decause we like `-Wunused`. The point for `-Wunused` is that sometimes it correctly points to identifiers that became unused after cleanups.

However it results in our tricks above. DBUG_ASSERT is no-op in release, hence the warning. 

assert-as-assume will enable the use of `result`, as that branch won't be a pure no-op anymore.

There is also a c++17 way to do that as an alternative. If the following would be a definition of `DBUG_ASSERT(expr)` in both release and debug builds (actually, in any build)
```
if constexpr (DBUG_ASSERT_ENABLED) {
... all the usual DBUG_ASSERT definition.
}
```
Frankly speaking, I haven't tested this assumption:)

if constexpr will guarantee nothing from assertion code will squeeze into the release builds. And also, `int result` will be kept used (in theory).

gkodinov:
Converting to draft due to inactivity. Please convert it back to "open" when/if you intend to resume working on it.

PR #3945, opened 2025-04-03
MDEV-36469 empty db passed to is_infoschema_db is wasteful

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36469*

## Description

This is backport of #3940 that removes forms of undefined behaviour in 10.11+, but also is a performance improvement. When merging to 11.8 the existing 11.8 code should prevail.

@vuvova acceptable enough for 10.6?

In many cases parsing a table name doesn't include a database name, therefore the empty db isn't going to be the information_schema.

The streq in is_infoschema_db is quite heavy in the so opportunities to bypass this should be taken.

In branches 10.11+ the collation improvements resulting in UBSAN errors when a null pointer reached the lower layers.

In 11.8+ the db is a Lex_ident_db type, and the is_null/is_emtpy usage that will result in the merge conflict with
db5bb6f3339be5a49c0f397eb80a0f259f73f447. The 11.8+ code should remain during the merge.

Alternatives considered:

Implementing check inside is_infoschema_db was rejected as the other callers of this function seemed to have meanisms already to validate that the db passed wasn't null.

Porting is_null/is_empty mechanisms to st_mysql_const_lex_string would create asymmetry with non-const functions. This is also in an ABI compatible sensitive area of the code. As a structure the direct access of str and length as needed is also quite common.

Changing db to a Lex_ident_db in the sql_parse function was incompatible with the other usages.

## Release Notes

SQL parsing speed improved by less overhead in checking if the database is an information_schema.
This eliminated some undefined behaviour with pointers (10.11+)

## How can this PR be tested?

Existing MTR test cover this already.

- [] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*
- [X] * This is a bug fix for 10.11, correcting undefined behaviour, and a parsing performance improvement on 10.6.

## PR quality check

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -8389,7 +8389,11 @@ TABLE_LIST *st_select_lex::add_table_to_list(THD *thd,
     DBUG_RETURN(0);
   else
     fqtn= FALSE;
-  bool info_schema= is_infoschema_db(&db);
+  /*
+    The test for length==0 is for performance purposes as is_infoschema_db()
+    is quite heavy. Without explict db SQL syntax, length==0 is a common case.
+  */
+  bool info_schema= db.length ? is_infoschema_db(&db) : false;
   if (!table->sel && info_schema &&
       (table_options & TL_OPTION_UPDATING) &&
       /* Special cases which are processed by commands itself */


abarkov (changes requested):
Hi Daniel. The change looks OK for me.

Can you please add a comment line (before the changed line) telling that the test for length==0 is for performance purposes as is_infoschema_db() is quite heavy?

Otherwise somebody can think "is_infoschema_db() can handle strings with zero length correctly why extra test before the call?"  and remove the line :)

Thanks.

abarkov (approved):
Okey to push. Thanks.

sql/sql_parse.cc
```diff
@@ -8389,7 +8389,11 @@ TABLE_LIST *st_select_lex::add_table_to_list(THD *thd,
     DBUG_RETURN(0);
   else
     fqtn= FALSE;
-  bool info_schema= is_infoschema_db(&db);
+  /*
+    The test for length==0 is for performance purposes as is_infoschema_db()
+    is quite heavy. Without explict db SQL syntax, length==0 is a common case.
+  */
+  bool info_schema= db.length ? is_infoschema_db(&db) : false;
```
vuvova, line 8396: I don't understand. What is quite heavy? `is_infoschema_db` is
```c++
inline bool is_infoschema_db(const LEX_CSTRING *name)
{
  return lex_string_eq(&INFORMATION_SCHEMA_NAME, name);
}
```
and `lex_string_eq` is
```c++
static inline bool lex_string_eq(const LEX_CSTRING *a, const LEX_CSTRING *b)
{
  if (a->length != b->length)
    return 0;                                   /* Different */
  return strcasecmp(a->str, b->str) == 0;
}
```
note that everything is inlined

grooverdan: Ok, seems like 11.8 is when this changed, which is already fixed. Thanks for the review - closing this aspect of the fix.

PR #3947, opened 2025-04-04
Use C-style block comment for .frm field parsingComment fix v2

Krishnateja4367:
This PR updates the comment above field_data_type_info_array.parse(...) to follow MariaDB’s C-style block comment convention, as per mentor feedback on PR #3938.


svoj (changes requested):
Feels like something went wrong while creating this PR, it shouldn't have 117 commits on behalf of other developers, right?

svoj:
Ah, so this comes from #3938. No need to create a separate pull request, please update original one.

PR #3948, opened 2025-04-04
MDEV-36480 USAN: checking identifier names for 0 length names

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36480*

## Description

mysql_prepare_create_table_finalize returns true if there's an error.

The point of checking the identifier name of the index is to see if there's an error.

The primary key doesn't have name, which is quite valid, so no point in checking empty names.

Checking empty names where the key->name.str is a null pointer results in a UBSAN error in Well_formed_prefix_status further down the stack.

## Release Notes

Avoid checking validity of nameless primary keys.

## How can this PR be tested?

Under UBSAN with a basic bootstrap.

## PR quality check

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -10389,7 +10389,13 @@ bool check_string_char_length(const LEX_CSTRING *str, uint err_msg,
 
 bool check_ident_length(const LEX_CSTRING *ident)
 {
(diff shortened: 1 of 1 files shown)


abarkov (approved):
The patch is OK. Thanks.

abarkov (approved):
Ok to push. Thank you!

PR #3951, opened 2025-04-05
MDEV-36427 - FTBFS with libxml2 2.14.0

svoj:
Connect engine fails to build with libxml2 2.14.0.

Connect engine uses "#ifndef BASE_BUFFER_SIZE" to determine if libxml2 is available. If libxml2 is unavailable it did redefine xmlElementType enum of libxml/tree.h. The reasons for this redefinition is vague, most probably some of these constants were used when connect was compiled with MSXML, while libxml2 was disabled.

However BASE_BUFFER_SIZE constant was removed from libxml2 recently, as a result connect fails to build due to xmlElementType constants redefinition.

Use LIBXML2_SUPPORT instead of BASE_BUFFER_SIZE for libxml2 availability check.

### DIFF
--- a/storage/connect/plgxml.h
+++ b/storage/connect/plgxml.h
@@ -5,7 +5,7 @@
 /******************************************************************/
 /*  Dual XML implementation base classes defines.                 */
 /******************************************************************/
-#if !defined(BASE_BUFFER_SIZE)
+#ifndef LIBXML2_SUPPORT
 enum ElementType {               // libxml2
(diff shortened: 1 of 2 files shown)


vuvova (changes requested):
it doesn't seem to compile

svoj:
Right, sorry. I did some edits to connect CMakeLists.txt, which did hide this error. I will sort it out.

PR #3955, opened 2025-04-07
MDEV-36491 Server crashes in Item_func_group_concat::print

DaveGosselin-MariaDB:
Initialize orig_args with nullptr and allocate it if it hasn't yet been allocated, either during construction or during fix_fields(), depending on use.  The SQL layer calls print() on conditions as a way of validating that they're correct, but in the case of Item_func_group_concat, this crashed because orig_args was not allocated by that point for the test case included in this commit.


DaveGosselin-MariaDB:
@vuvova essentially we need the change from https://github.com/MariaDB/server/pull/3805 and the original change.

vuvova:
I'm not sure about it, other `Item_sum` descendants don't do that. But `Item_sum::print()` starts from
```c++
  /* orig_args is not filled with valid values until fix_fields() */
  Item **pargs= fixed ? orig_args : args;
```
I suppose `Item_func_group_concat::print()` should do the same.

DaveGosselin-MariaDB:
> I'm not sure about it, other `Item_sum` descendants don't do that. But `Item_sum::print()` starts from
> 
> ```c++
>   /* orig_args is not filled with valid values until fix_fields() */
>   Item **pargs= fixed ? orig_args : args;
> ```
> 
> I suppose `Item_func_group_concat::print()` should do the same.

You're right, that works.

PR #3864, opened 2025-03-02
WIP: MDEV-20544: Update bundled Groonga and Mroonga

kou:
This is still in work in progress. I opened this PR just for using CI. I'll update the PR description after this is completed.

- [x] *The Jira issue number for this PR is: MDEV-20544*

## Description
TODO: fill description here

## Release Notes
TODO: What should the release notes say about this change?
Include any changed system variables, status variables or behaviour. Optionally list any https://mariadb.com/kb/ pages that need changing.

## How can this PR be tested?

TODO: modify the automated test suite to verify that the PR causes MariaDB to behave as intended.
Consult the documentation on ["Writing good test cases"](https://mariadb.org/get-involved/getting-started-for-developers/writing-good-test-cases-mariadb-server).

If the changes are not amenable to automated testing, please explain why not and carefully describe how to test manually.

## Basing the PR against the correct MariaDB version
- [ ] *This is a new feature or a refactoring, and the PR is based against the `main` branch.*
- [ ] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [ ] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [ ] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #3956, opened 2025-04-07
MDEV-36506 Build fails with cmake 4.0

FooBarrior:
- [x] *The Jira issue number for this PR is: MDEV-36506*

## Description
After update to cmake 4.0, many distros will fail to compile the code base. There is [MDEV-36422](https://jira.mariadb.org/browse/MDEV-36422) to wait for wsrep fix, but even without it, our code base needed some attention.

The proposed fix enables building the Server with WSREP disabled (`-DWITH_WSREP=0`).

## Release Notes
Add support to build with cmake 4.0, but brop support to cmake prior 3.5.

## PR quality check


vaintroub (approved):
Looks good

kou (approved):
+1

Thanks for notifying me!
Upstream Groonga/Mroonga/groonga-normalizer-mysql require more recent CMake. So we don't need to backport this to https://github.com/MariaDB/server/pull/3864 .

PR #3957, opened 2025-04-07
MDEV-36211 Incorrect result for BLOB type with NOT NULL constraint

DaveGosselin-MariaDB:
During optimize_cond, we incorrectly removed the NOT LIKE condition when attempting to remove any equality conditions.

Item_func_like's override of eq_cmp_result() returns COND_TRUE when its collation is the binary collation.  Item_bool_func2's implementation of remove_eq_conds would then attempt to detect if both arguments were equal to one another and return a NULL condition to optimize_conds.  This removes the condition from ever being evaluated (and Item_func_like::val_bool is never called in this case), rendering the incorrect result.

Fix this by checking the negated condition during eq_cmp_result() to return either COND_FALSE in the negated==true case, or COND_TRUE in the negated==false case which has the effect of not removing the NOT LIKE/LIKE condition for the query.


DaveGosselin-MariaDB:
I don't like the one-liner because it confuses implementation details of the child class (`Item_func_like`) with its parent `Item_bool_func2` and is less maintainable long-term.  The method `on_cmp_result` is a callback: when something happens at the higher level (`Item_bool_func2`) and a derived class needs to take a derived class-specific action on that, then the callback is invoked with the result code.  In this case, `Item_func_like` has to take a different action than its parent, but its parent doesn't need to know what that action is.  In fact any derived type from `Item_func_bool2` can specialize their behavior with respect to that event.

spetrunia (changes requested):
The MDEV title was poor.  It was mentioning Blob, which we've discovered was not essential.  It didn't mention NOT LIKE, although it was essential. 

Pleas change the commit comment to match new MDEV title - don't mention blobs, do mention use of BINARY collation and NOT LIKE.

mysql-test/main/mdev-36211.test
```diff
@@ -0,0 +1,20 @@
...
+DROP TABLE IF EXISTS t1;
+CREATE TABLE t1 (c1 BLOB);
+INSERT INTO t1 (c1) VALUES (1);
+SELECT c1 FROM t1 WHERE c1 NOT LIKE c1;
+
+DROP TABLE IF EXISTS t1;
+CREATE TABLE t1 (c1 BLOB NOT NULL);
+INSERT INTO t1 (c1) VALUES (1);
+SELECT c1 FROM t1 WHERE c1 LIKE c1;
+
```
spetrunia, line 14: So this creates again the table with `c1 BLOB NOT NULL` just to test with LIKE ?  Please move the SELECT query into the top section of the file where  one is testing `SELECT ... NOT LIKE`.
Same for the NULL-able blob section.

mysql-test/main/mdev-36211.test
```diff
@@ -0,0 +1,20 @@
...
+DROP TABLE IF EXISTS t1;
+CREATE TABLE t1 (c1 BLOB);
+INSERT INTO t1 (c1) VALUES (1);
+SELECT c1 FROM t1 WHERE c1 LIKE c1;
+
```
spetrunia, line 19: Also, I think this doesn't deserve its own .test file.  We would have too many files.
Please put it into an existing file, either `main/func_like.test` (I prefer this) or `main/ctype_binary.test`.

spetrunia (changes requested):

mysql-test/main/mdev-36211.test
DaveGosselin-MariaDB: Yes it's repetitive, but I do this to isolate each test case: later, some maintainer could easily copy-paste this one case out for manual testing.  They would not have to selectively copy and paste some lines but not others.

mysql-test/main/mdev-36211.test
DaveGosselin-MariaDB: I will move it to `main/func_like.test`.  But I don't know what you mean by "too many files."  There is a long-used pattern of introducing new tests per bugfix MDEV, and modern filesystems would be practically limited by even a few thousand files.  Separate files makes merging simpler, too.

spetrunia (approved):

PR #3958, opened 2025-04-08
MDEV-36398 Extend SBOM with 'license' and 'copyright'

vaintroub:
- [x] *The Jira issue number for this PR is: MDEV-36398*

## Description
Some users want additional properties in SBOM.  This PR adds them.
The properties (license and copyright statement) are maintained per third party dependency in machine-readable file (which is also human readable).
Also other relevant properties are maintained

## Release Notes
Nothing needs to be in the notes. SBOMs are only product metadata

## How can this PR be tested?

cmake . -DWITH_SBOM=1
examine sbom.json in build directory.

## PR quality check

### DIFF
--- a/cmake/generate_sbom.cmake
+++ b/cmake/generate_sbom.cmake
@@ -36,30 +36,23 @@ ENDMACRO()
 
 # Get CPE ID ( https://en.wikipedia.org/wiki/Common_Platform_Enumeration )
 # for given project name and version
-# Only "known" CPEs are handled here, e.g currently no CPE for rocksdb
+# CPE prefix are stored with other auxilliary info in the 3rdparty_info.cmake
+# file
 FUNCTION(SBOM_GET_CPE name version var)
-  SET(cpe_prefix_map
-    "zlib" "zlib:zlib"
-    "mariadb-connector-c" "mariadb:connector\\\\/c"
-    "wolfssl" "wolfssl:wolfssl"
-    "minizip" "zlib:zlib"
-    "pcre2" "pcre:pcre2"
-    "fmt" "fmt:fmt"
-    "boost" "boost:boost"
-    "thrift" "apache:thrift"
-  )
-  LIST(FIND cpe_prefix_map "${name}" idx_cpe_mapping)
-  # Version needs to have at least one dot character in it.
-  # Otherwise, we assume it is a git  hash, and do not generate CPE
-  STRING(FIND "${version}" "." idx_version_dot)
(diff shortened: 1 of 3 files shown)


3rdparty_info.yaml
```diff
@@ -0,0 +1,50 @@
+# This file is used for SBOM generation.
```
vuvova, line 1: why yaml?

if there's no good reason to use specifically yaml over some other format, I'd prefer one of
* cmake file (set of `SET()` statements) — no need to write a parser for that, just include
* [debian/copyright file](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/) — directly used in debian, `cpe-prefix` can be in a comment

cmake/generate_sbom.cmake
```diff
@@ -277,6 +355,15 @@ FUNCTION(GENERATE_SBOM)
...
     ENDIF()
+
+    GET_YAML_VALUE(THIRD_PARTY "${repo_name_lower}.license" license)
+    IF(NOT license)
+      MESSAGE(FATAL_ERROR "no license for 3rd party dependency ${repo_name_lower}.")
+    ENDIF()
+    GET_YAML_VALUE(THIRD_PARTY "${repo_name_lower}.copyright" copyright)
+    IF(NOT copyright)
+      SET(copyright NOASSERTION)
+    ENDIF()
```
vuvova, line 366: did you double-check that it's the correct behavior? a license must be present, a copyright is optional?

3rdparty_info.yaml
vaintroub: - yaml was just my preference, common use, relatively easy to write an incomplete parser. Extensible, we could extend it with any new field easily
- Debian thing would not take. Not extendible, not known outside of Debian bubble (I do not take misuse of Comment section serious, we might need more fields there, e.g provider)
-  multiple SET is fine by me, if you prefer that to yaml, I can do, it is slightly more verbose, but it does not matter

cmake/generate_sbom.cmake
vaintroub: No, just experience. Every project we depend on has a license, while most projects do not have a copyright statement. We did not have, until 2 weeks ago.

3rdparty_info.yaml
vuvova: just thought of the third option, JSON, this fits into "no need to write a parser" argument, with `string(JSON` command. But it's only from 3.19

PR #3959, opened 2025-04-08
MDEV-36486 Optimizer hints are resolved against the INSERT part of IN…

Olernov:
…SERT..SELECT

## Description
When processing queries like
`  INSERT INTO t1 (..) SELECT .. FROM t1, t2 ...`

there is a single query block (i.e., a single SELECT_LEX) for both INSERT and SELECT parts. During hints resolution, when hints are attached to particular TABLE_LIST's, the search is performed by table name across the whole query block.
So, if a table mentioned in an optimizer hint is present in the INSERT part, the hint is attached to the that tablet. This is obviously wrong as optimizer hints are supposed to only affect the SELECT part of and INSERT..SELECT clause.

This commit disables possible attaching hints to tables in the INSERT part and also disables hints resolution during mysql_prepare_insert_check_table(), which could lead to double resolution and warnings duplication

## How can this PR be tested?

Test cases are added to opt_hints.test

## PR quality check


spetrunia (approved):
Looks good.

Olernov:
I force-pushed an improved version of the patch as the last one broke some tests

spetrunia:
Input:
```
  // INSERT..SELECT allows placing hints next to either INSERT or SELECT, i.e.:
```
Make the comment use `/* ... */` syntax like it's done nearly everywhere.


```
   If hints are present for both INSERT and SELECT
  // parts, SELECT part hints are preserved while INSERT part hints are discarded
```
Let's emit a warning if we swallow a hint.
Later on, we can fix not to swallow it, like MySQL 8.4 does..

Olernov:
> Input:
> 
> ```
>   // INSERT..SELECT allows placing hints next to either INSERT or SELECT, i.e.:
> ```
> 
> Make the comment use `/* ... */` syntax like it's done nearly everywhere.
Not possible here, as we use '/*'  and '*/' inside the comment text and they are misinterpreted by the compiler

PR #3960, opened 2025-04-09
MDEV-28933: Per-table unique FOREIGN KEY constraint names

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-28933*
## Description
`dict_foreign_t::sql_id()`: Return the SQL constraint name. Before MySQL 4.0.18, user-specified constraint names were ignored. Starting with MySQL 4.0.18, the specified constraint name was prepended with the schema name and `/`.  Now we are transforming into a format where the constraint name is prepended with the `dict_table_t::name` and the impossible UTF-8 sequence 0xff. Generated constraint names will be of the form `1`, `2`, …, internally stored as `schemaname/tablename\3771`, `schemaname/tablename\3772`, ….
## Release Notes
InnoDB `FOREIGN KEY` constraint names need not be unique within a schema any more, but only unique within a table. Auto-generated constraint names will be of the form `1`, `2`, ….

The `INFORMATION_SCHEMA` views `INNODB_SYS_FOREIGNS` and `INNODB_SYS_FOREIGN_COLS` will only display the user-specified constraint name, no longer any `databasename/` prefix.
### Upgrade testing
```sh
./mtr --manual-gdb innodb.innodb-fk
```
Before starting the first test step in GDB, specify with `file` an older `mariadbd` executable that does not contain these changes. For the subsequent test steps, specify the updated `mariadbd` executable. The test will pass but display a warning during the execution of `ALTER TABLE t1 ENGINE=InnoDB, ALGORITHM=COPY`:
```
***Warnings generated in error logs during shutdown after running tests: innodb.innodb-fk

2025-04-15  9:20:00 3 [Warning] InnoDB: In ALTER TABLE `test`.`t1` has or is referenced in foreign key constraints which are not compatible with the new table definition.
```
It turns out that this warning is expected by the test, but `call mtr.add_suppression()` apparently has no effect when running `./mtr --manual-gdb`.

For some additional diagnostics, you can apply the following patch:
```diff
diff --git a/mysql-test/suite/innodb/t/innodb-fk.test b/mysql-test/suite/innodb/t/innodb-fk.test
index 28dc4a28ce1..d78d08fecf9 100644
--- a/mysql-test/suite/innodb/t/innodb-fk.test
+++ b/mysql-test/suite/innodb/t/innodb-fk.test
@@ -39,6 +39,7 @@ select * from fk_50;
 let $i = $fk_tables;
 while ($i)
 {
+  eval show create table fk_$i;
   eval drop table fk_$i;
   dec $i;
 }
```
This will correctly display the constraint names `pc120` through `pc1` even though they were internally stored as `test/pc120` through `test/pc1`.
### Downgrade testing
If you run the test `innodb.innodb-fk` as it is adjusted in this PR, and run the unmodified `mariadbd` on the 3rd test step (the server is being started 3 times for running the test), then it will obviously fail because a generated constraint has a different name:
```diff
CURRENT_TEST: innodb.innodb-fk
mysqltest: At line 203: query 'ALTER TABLE `boroda` DROP FOREIGN KEY `2`' failed: ER_CANT_DROP_FIELD_OR_KEY (1091): Can't DROP FOREIGN KEY `2`; check that it exists
```
This happens even if you run the modified test against the old server all the time. We can work around this by commenting out the problematic statement:
```diff
diff --git a/mysql-test/suite/innodb/t/innodb-fk.test b/mysql-test/suite/innodb/t/innodb-fk.test
index 28dc4a28ce1..aba9e0f44c9 100644
--- a/mysql-test/suite/innodb/t/innodb-fk.test
+++ b/mysql-test/suite/innodb/t/innodb-fk.test
@@ -200,7 +200,7 @@ CREATE TABLE `boroda` (
 ALTER TABLE `boroda`
   ADD FOREIGN KEY (`b`) REFERENCES `boroda`(`id`);
 
-ALTER TABLE `boroda` DROP FOREIGN KEY `2`;
+#ALTER TABLE `boroda` DROP FOREIGN KEY `2`;
 
 RENAME TABLE `boroda` TO `#boroda`;
 
```
If we run the old `mariadbd` against that for all test steps, there will be not only result differences for the generated constraint names, but also the following:
```diff
@@ -216,7 +216,7 @@
 ALTER TABLE t1 RENAME TO tm1, ALGORITHM=COPY;
 SELECT * FROM INFORMATION_SCHEMA.INNODB_SYS_FOREIGN;
 ID	FOR_NAME	REF_NAME	N_COLS	TYPE
-fk	test/t1_fk	test/t1	1	4
+test/fk	test/t1_fk	test/t1	1	4
 SET FOREIGN_KEY_CHECKS=0;
 CREATE TABLE t1 (c1 BIGINT NOT NULL, c2 BIGINT NOT NULL, PRIMARY KEY(c1), UNIQUE KEY(c2)) ENGINE=MEMORY;
 ALTER TABLE t1 ENGINE=InnoDB, ALGORITHM=COPY;
```
If we run this modified test with `./mtr --manual-gdb innodb.innodb-fk` and run the updated server only for the first or second step or both, the test will work in the same way.

This confirms that `DROP TABLE` can drop not only the old-format constraints in the updated server, but also the new-format constraints in the old server. Also `DROP DATABASE schemaname` is expected to work, because the `SYS_FOREIGN.ID` and `SYS_FOREIGN_COLS.ID` will continue to start with `schemaname/`. But, `ALTER TABLE…DROP FOREIGN KEY` definitely is not going to work after a downgrade, because it is not expected to be possible for the user to input the invalid UTF-8 sequence 0xff.
## PR quality check


storage/innobase/dict/dict0dict.cc
```diff
@@ -1663,23 +1663,36 @@ dict_table_rename_in_cache(
...
+		/* All table names are internally stored in charset
+		my_charset_filename (except the temp tables and the
+		partition identifier suffix in partition tables). The
+		foreign key constraint names are internally stored
+		in UTF-8 charset.  The variable fkid here is used
+		to store foreign key constraint name in charset
+		my_charset_filename for comparison further below. */
+		char    fkid[MAX_TABLE_NAME_LEN * 2 + 20];
+
+		if (const char* s= strchr(foreign->id, '\377')) {
```
dr-m, line 1675: We do not really need the `fkid[]` buffer here, and we should invoke `foreign->sql_id()` to extract the intended visible part of the constraint name, just like we do everywhere else.

dr-m (commented):
I figured out that we can simply use decimal ASCII numbers for the generated constraint names; an `ibfk_` prefix would not add any value. I will revise this accordingly.

sql/sql_base.cc
```diff
@@ -8997,7 +8997,7 @@ fill_record(THD *thd, TABLE *table_arg, List<Item> &fields, List<Item> &values,
   List_iterator_fast<Item> f(fields),v(values);
   Item *value, *fld;
   Item_field *field;
-  Field *rfield;
+  Field *rfield= nullptr;
```
Thirunarayanan, line 9000: Is this related to this patch?

storage/innobase/dict/dict0dict.cc
```diff
@@ -1580,150 +1557,18 @@ dict_table_rename_in_cache(
...
+					table->name.m_name,
+					foreign->sql_id());
+		char* id = foreign->id;
+		if (fklen > strlen(id)) {
+			id = static_cast<char*>(
+				mem_heap_alloc(foreign->heap, fklen + 1));
 		}
-
+		snprintf(id, fklen + 1, "%s\377%s",
+			 table->name.m_name, foreign->sql_id());
```
Thirunarayanan, line 1569: Do we need to call foreign->sql_id() twice?

sql/sql_base.cc
dr-m: Sorry, no, it was included accidentally. This fix should actually be applied to the 10.5 branch to avoid `-Wmaybe-uninitialized` on `g++ -Og`.

Thirunarayanan (commented):
Do we address upgrade scenario? Does rename table from lower version works? Please point me to any test case.

storage/innobase/dict/dict0dict.cc
dr-m: Not really: we could cache the return value of the first call.

dr-m: d596884d4052c318dd92e0d3f3fdd6755f9b3519 is introducing another local variable `sql_id`. We must in fact avoid reassigning `foreign->id` before we have invoked `table->foreign_set.erase(it)`, because erasing would compare `foreign->id`, and the entry may need to be moved to a different position in the binary tree.

dr-m:
> Do we address upgrade scenario? Does rename table from lower version works? Please point me to any test case.

Upgrades can’t easily be covered by `mtr`, other than by storing a copy of an old data directory in `mysql-test/std_data`, or by introducing some `DBUG_EXECUTE_IF` that would create some constraint names in the old format. On upgrade, basically what happens is that `dict_foreign_t::sql_id()` will return constraint names without any `schemaname/` prefix, and `row_rename_table_for_mysql()` and `dict_table_rename_in_cache()` would drop old constraints or rename them to use the new `schemaname/tablename\377` prefix. After a downgrade from this, the constraint names would likely be displayed as `schemaname/tablename`; at least on my system, the invalid UTF-8 sequence was being treated as a sort of string terminator.

dr-m:
77c31385bde8fbad30a2cb69c65f5308f486e8d6 addresses the regression [MDEV-37077](https://jira.mariadb.org/browse/MDEV-37077).

Thirunarayanan (approved):

PR #3961, opened 2025-04-10
MDEV-36542 Remove compile options for faking initialization

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36542*

## Description

Stated cases for use:
* bypass static analyizers (which should be able to see faults if they are there)
* old gcc bugs that where fixed ~4.7.

Remove defines:
* FORCE_INIT_OF_VARS
* LINT_INIT_STRUCT
* FORCE_INIT_OF_VARS

Define UNINIT_VAR make into a no-op for later branch removal.

mysql.cc - all paths in com_go populate buff before use.

json: Item_func_json_merge::val_str
  LINT_INIT(js2) unneeded as usage in report_error goto label
  only occurs after assignment to js2.

Item_func_json_contains_path::val_bool
  n_found initialized at same location.

_ma_ck_delete - org_key only valid under share->now_transactional likewise with _ma_ck_write_btree_with_log

## Release Notes

nothing - non-functional change from user perspective

## How can this PR be tested?

## PR quality check


storage/connect/tabxml.cpp
```diff
@@ -1611,7 +1611,9 @@ bool XMLCOL::ParseXpath(PGLOBAL g, bool mode)
     if (Tdbp->Xpand)
       n = Tdbp->Limit;
 
+    auto oLong = Long;
```
montywi, line 1614: Please do not use auto.
Please also add a comment why we have to preserver 'Long' as it is not obvious here
Please also use 'save_Long' instead of 'oLong' as we do in most of the other part of the code

I suggest this patch instead that makes this a bit more clear:

+++ b/storage/connect/tabxml.cpp
@@ -1611,7 +1611,7 @@ bool XMLCOL::ParseXpath(PGLOBAL g, bool mode)
     if (Tdbp->Xpand)
       n = Tdbp->Limit;
 
-    new(this) XMULCOL(Value);        // Change the class of this column
+    new(this) XMULCOL(Value, Long);  // Change the class of this column
     } // endif Inod
 
   Valbuf = (char*)PlugSubAlloc(g, NULL, n * (Long + 1));
diff --git a/storage/connect/tabxml.h b/storage/connect/tabxml.h
index 252d2f74d85..9248db58b9b 100644
--- a/storage/connect/tabxml.h
+++ b/storage/connect/tabxml.h
@@ -231,7 +231,7 @@ class XMULCOL : public XMLCOLX {
  public:
   // The constructor must restore Value because XOBJECT has a void
   // constructor called by default that set Value to NULL
-  XMULCOL(PVAL valp) {Value = valp; Mul = true;}
+  XMULCOL(PVAL valp, int buffer_length) {Value = valp; Long= buffer_length ; Mul = true;}
 
   // Methods
   void ReadColumn(PGLOBAL g) override;

storage/connect/filter.cpp
```diff
@@ -1589,8 +1589,9 @@ void FILTER::Prints(PGLOBAL g, char *ps, uint z)
 /***********************************************************************/
 /*  FILTERCMP constructor.                                             */
 /***********************************************************************/
-FILTERCMP::FILTERCMP(PGLOBAL g)
+FILTERCMP::FILTERCMP(PGLOBAL g, OPVAL Opc)
```
montywi, line 1592: Opc -> Opc_arg

storage/connect/filter.cpp
```diff
@@ -1589,8 +1589,9 @@ void FILTER::Prints(PGLOBAL g, char *ps, uint z)
...
   {
+  this->Opc= Opc;
```
montywi, line 1594: Change to Opc= Opc_arg

include/my_global.h
```diff
@@ -440,10 +440,10 @@ extern "C" int madvise(void *addr, size_t len, int behav);
 /*
    Suppress uninitialized variable warning without generating code.
 */
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(__clang__)
 /* GCC specific self-initialization which inhibits the warning. */
 #define UNINIT_VAR(x) x= x
-#elif defined(_lint) || defined(FORCE_INIT_OF_VARS)
+#elif (defined(_lint) || defined(FORCE_INIT_OF_VARS)) && !defined(__clang__)
```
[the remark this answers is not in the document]
grooverdan: Ok, restored operation of `FORCE_INIT_OF_VARS`

The `WITH_UBSAN` is incorrect. The cases in `my_byteorder.h` and `my_pthread.h` related to constructs and additional stack requirements of UBSAN. 

As the macro is explicitly around false warning by compilers and clang doesn't do this, at least in the clang-15 (Apple) and clang-16s of old debian, its the most direct way of mapping the cause to effect. I've added the clang `Werror=(uninit things)` on `RelWithDebInfo` on `mysql_release.cmake` so that clang warnings are 100% caught..

The other problem with WITH_UBSAN is that static analyzer would require this mode which is a bit indirect and has the additional consequence of `my_pthread.h` changes.

storage/innobase/log/log0sync.cc
```diff
@@ -278,6 +278,8 @@ group_commit_lock::lock_return_code group_commit_lock::acquire(value_type num, c
   return lock_return_code::EXPIRED;
 }
 
+PRAGMA_DISABLE_CHECK_STACK_FRAME
```
vaintroub: I think you can reduce the size of the stack by reducing `callbacks` array. That should be fine. the 1000 items is just something large enough, that did not make problems in the past, but if it now creates some problems, it can be reduced. "Large" would still be nice, since it improves performance in highly parallel cases.

grooverdan: I looked, but I don't want to change that in 10.6. I could do this recommendation on the merge to 10.11 and make it more considered change there.

grooverdan: Monty did change this https://github.com/MariaDB/server/commit/49febfad21ab6131a4ca421cd08fb25107d42509

sql/item_jsonfunc.cc
```diff
@@ -1728,9 +1728,16 @@ bool Item_func_json_contains_path::val_bool()
...
   int array_sizes[JSON_DEPTH_LIMIT];
   uint has_negative_path= 0;
+#if defined(FORCE_INIT_OF_VARS)
+  /*
+    Initialization force not required after gcc 13.3
+    where it correctly sees that an uninitialized read
+    of n_found doesn't occur with mode_one being true.
+  */
+  n_found= 0; 
+#endif
```
dr-m, line 1740:
```
/usr/bin/g++-15 -DHAVE_CONFIG_H -DHAVE_EVENT_SCHEDULER -DHAVE_LIBAIO -DHAVE_POOL_OF_THREADS -DHAVE_URING -DMYSQL_SERVER -D_FILE_OFFSET_BITS=64 -I/mariadb/11.4/wsrep-lib/include -I/mariadb/11.4/wsrep-lib/wsrep-API/v26 -I/dev/shm/11.4/include -I/mariadb/11.4/include/providers -I/mariadb/11.4/include -I/mariadb/11.4/sql -I/dev/shm/11.4/extra/wolfssl -I/mariadb/11.4/extra/wolfssl/wolfssl -I/mariadb/11.4/extra/wolfssl/wolfssl/wolfssl -I/dev/shm/11.4/sql -I/mariadb/11.4/tpool -Og -march=native -mtune=native -fPIC -g -DPROTECT_STATEMENT_MEMROOT -U_FORTIFY_SOURCE -fsanitize=address -fPIC -DENABLED_DEBUG_SYNC -D_GLIBCXX_DEBUG -D_GLIBCXX_ASSERTIONS -ggdb3 -DSAFE_MUTEX -Wall -Wenum-compare -Wenum-conversion -Wextra -Wmissing-braces -Wno-format-truncation -Wno-init-self -Wno-nonnull-compare -Wno-unused-parameter -Wnon-virtual-dtor -Woverloaded-virtual -Wvla -Wwrite-strings -Werror -fno-operator-names -Wsuggest-override -std=gnu++11   -DHAVE_OPENSSL -DHAVE_WOLFSSL  -DWOLFSSL_USER_SETTINGS -MD -MT sql/CMakeFiles/sql.dir/item_jsonfunc.cc.o -MF sql/CMakeFiles/sql.dir/item_jsonfunc.cc.o.d -o sql/CMakeFiles/sql.dir/item_jsonfunc.cc.o -c /mariadb/11.4/sql/item_jsonfunc.cc
/mariadb/11.4/sql/item_jsonfunc.cc: In member function ‘virtual bool Item_func_json_contains_path::val_bool()’:
/mariadb/11.4/sql/item_jsonfunc.cc:1819:23: error: ‘n_found’ may be used uninitialized [-Werror=maybe-uninitialized]
 1819 |         if (--n_found == 0)
      |             ~~~~~~~~~~^~~~
/mariadb/11.4/sql/item_jsonfunc.cc:1744:7: note: ‘n_found’ was declared here
 1744 |   int n_found;
      |       ^~~~~~~
```
I hope that `FORCE_INIT_OF_VARS` can be removed. Last time I checked, starting with GCC 7 (the oldest major version that is present in currently supported GNU/Linux distributions), a single 64-bit store can be emitted to initialize two 32-bit variables. We are zero-initializing `uint has_negative_path` above, and we could as well zero-initialize `n_found`. https://godbolt.org makes it very easy to check if there is any overhead.

montywi: One should not use FORCE_INIT_OF_VARS in the normal code. This define is mainly to be used in my_global.h

Please remove #ifdef  FORCE_INIT_OF_VARS and use instead uint UNINIT_VAR(n_found);

sql/item_jsonfunc.cc
```diff
@@ -2412,7 +2414,6 @@ String *Item_func_json_merge::val_str(String *str)
   String *js1= args[0]->val_json(&tmp_js1), *js2=NULL;
   uint n_arg;
   THD *thd= current_thd;
-  LINT_INIT(js2);
```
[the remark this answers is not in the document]
montywi: ok to remove as it's already initialized

sql/item_jsonfunc.cc
```diff
@@ -1565,7 +1565,6 @@ bool Item_func_json_contains_path::val_bool()
   longlong result;
   json_path_t p;
   int n_found;
-  LINT_INIT(n_found);
```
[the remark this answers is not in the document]
montywi: This has still to be fixed

montywi (commented):
Please fix the issues that is not yet resolved (adding UINIT_VAR for one variable).
Ok to push after that.

cmake/maintainer.cmake
```diff
@@ -68,6 +68,14 @@ IF(CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS
   SET(MY_ERROR_FLAGS ${MY_ERROR_FLAGS} -Wno-error=non-virtual-dtor) # gcc bug 7302
 ENDIF()
 
+FOREACH(LANG C CXX)
+  IF(CMAKE_${LANG}_COMPILER_ID MATCHES "Clang")
+    MY_CHECK_AND_SET_COMPILER_FLAG(-Werror=uninitialized-explicit-init)
+    MY_CHECK_AND_SET_COMPILER_FLAG(-Werror=uninitialized-const-reference)
+    SET(CMAKE_${LANG}_FLAGS "${CMAKE_${LANG}_FLAGS} -Werror=uninitialized")
```
montywi, line 75: Please remove SET(CMAKE_${LANG}_FLAGS "${CMAKE_${LANG}_FLAGS} -Werror=uninitialized")
We already set -Werror, in maintainer mode. We don't need -Werror=unintialized

We should not break the compilation in non maintainer mode just for an uninitialized error like we do not break it for other warnings. We will catch the issue in buildbot if needed.

montywi: I checked with Serg and he agrees we should not break on errors outside maintainer mode.

montywi (commented):
Remove the -Werror line and then you can push

cmake/maintainer.cmake
montywi: ok to push after the above is fixed

storage/maria/aria_chk.c
```diff
@@ -236,9 +236,6 @@ int main(int argc, char **argv)
   }
   maria_end();
   my_exit(error);
-#ifndef _lint
```
montywi, line 239: Keep the return 0;
(This was an #ifndef _lint, which means the code should be there for normal builds)

storage/maria/aria_pack.c
```diff
@@ -289,11 +289,9 @@ int main(int argc, char **argv)
   maria_end();
   my_end(verbose ? MY_CHECK_ERROR | MY_GIVE_INFO : MY_CHECK_ERROR);
   exit(error ? 2 : 0);
-#ifndef _lint
```
montywi, line 292: Keep the return 0;

storage/myisam/myisamchk.c
```diff
@@ -124,9 +124,6 @@ int main(int argc, char **argv)
   my_end(check_param.testflag & T_INFO ? MY_CHECK_ERROR | MY_GIVE_INFO : MY_CHECK_ERROR);
   rc= (uchar) error;
   exit(rc);
-#ifndef _lint
```
montywi, line 127: Keep the return 0;

storage/myisam/myisampack.c
```diff
@@ -248,9 +248,6 @@ int main(int argc, char **argv)
   free_defaults(default_argv);
   my_end(verbose ? MY_CHECK_ERROR | MY_GIVE_INFO : MY_CHECK_ERROR);
   exit(error ? 2 : 0);
-#ifndef _lint
```
montywi, line 251: Keep the return 0;

montywi (commented):
Fix the pending issues and then you can push

montywi (commented):
Please fix according to review comments

sql/item_jsonfunc.cc
grooverdan: you already changed this de22bfc2ff908a36ec639f8ca2f57f45b4be4e22

storage/maria/aria_chk.c
grooverdan: After a no-return `my_exit` function the `return 0` is just dead code. Compilers and static analysers rightly complain about after that, normal build or not.

PR #3964, opened 2025-04-10
MDEV-36235 Incorrect result for BETWEEN over unique blob prefix

DaveGosselin-MariaDB:
Disallow range optimization for BETWEEN when casting one of the arguments from STRING to a numeric type would be required to construct a range for the query.

Adds a new method on Item_func_between called can_optimize_range_const which allows range optimization when the types of the arguments to BETWEEN would permit it.


DaveGosselin-MariaDB:
OK @spetrunia your hunch about `BINARY` and `VARBINARY` was correct, so I need to expand this fix to include those cases.

mysql-test/main/type_blob.test
```diff
@@ -808,3 +808,29 @@ DELIMITER ;$$
...
 DROP TABLE t1;
 SET note_verbosity=DEFAULT;
+
+--echo #
+--echo # MDEV-36235 Incorrect query result
+--echo #
+CREATE TABLE t1 (c1 TINYBLOB, UNIQUE (c1(2)));
+INSERT INTO t1 (c1) VALUES (1);
+SELECT c1 FROM t1 WHERE 'a' BETWEEN 0 AND (c1);
+
```
abarkov, line 818: Can you please also add a test with 2 values in the table, here and in the tests below?

sql/field.h
```diff
@@ -4394,7 +4394,11 @@ static inline longlong read_lowendian(const uchar *from, uint bytes)
 
 extern LEX_CSTRING temp_lex_str;
 
-class Field_blob :public Field_longstr {
+class Field_blob :public Field_longstr
+{
+Data_type_compatibility can_optimize_range(const Item_bool_func *cond,
+                                           const Item *item,
+                                           bool is_eq_func) const override;
```
abarkov, line 4401: I don't think overriding Field_blob::can_optimize_range() is correct. The problem is also repeatable with VARBINARY and BINARY. See the MDEV for details, and please add tests for the other data types mentioned in the comment in the MDEV.

sql/item_cmpfunc.h
```diff
@@ -1003,6 +1014,7 @@ class Item_func_opt_neg :public Item_bool_func
 
 class Item_func_between :public Item_func_opt_neg
 {
+  bool can_optimize_range(const Item *item, bool is_eq_func) const override;
```
abarkov, line 1017: I don't think we need a new virtual method.

The problem can be fixed in Item_func_between::get_mm_tree(), in this block:
```
  for (uint i= 1 ; i < arg_count ; i++)
  {
    if (arguments()[i]->real_item()->type() == Item::FIELD_ITEM)
    {
```

sql/opt_range.cc
```diff
@@ -7910,6 +7910,20 @@ SEL_TREE *Item_func_ne::get_func_mm_tree(RANGE_OPT_PARAM *param,
...
+bool Item_func_between::can_optimize_range(const Item *item,
+                                           bool is_eq_func) const
+{
+  /*
+    BETWEEN cannot allow the range optimization when the argument type is a
+    string and BETWEEN's declared comparison type is not.  This would require
+    a casting operation for the argument.  Rather than cast, disallow the range
+    optimization in this case.
+  */
+  return cmp_type() == item->cmp_type() ||
```
abarkov, line 7922: Calling Item_func_between::cmp_type() is not relevant here. It is cmp_type() of the Item_func_between itself. It return the BOOLEAN data type and it's cmp_type() is always INT_RESULT, no matter what data types of arguments are.

sql/item_cmpfunc.h
abarkov: This code should do:
```
--- a/sql/opt_range.cc
+++ b/sql/opt_range.cc
@@ -8653,6 +8653,13 @@ Item_func_between::get_mm_tree(RANGE_OPT_PARAM *param, Item **cond_ptr)
     if (arguments()[i]->real_item()->type() == Item::FIELD_ITEM)
     {
       Item_field *field_item= (Item_field*) (arguments()[i]->real_item());
+      Type_handler_hybrid_field_type cmp(field_item->type_handler_for_comparison());
+      if (cmp.aggregate_for_comparison(args[0]->type_handler_for_comparison()) ||
+          cmp.type_handler() != m_comparator.type_handler())
+        continue;
       SEL_TREE *tmp= get_full_func_mm_tree(param, field_item,
                                            (Item*)(intptr) i);
```

abarkov: It seems no other changes except this one ^^^ are needed.

abarkov (changes requested):
Please fix the patch and make sure all additional problems reported in the MDEV are gone. Thanks.

sql/field.h
DaveGosselin-MariaDB: Thanks, @spetrunia pointed this out to me during our call yesterday when I was discussing this solution.

mysql-test/main/type_blob.test
DaveGosselin-MariaDB: Good idea, will do.

mysql-test/main/type_blob.test
```diff
@@ -808,3 +808,54 @@ DELIMITER ;$$
...
+SELECT c1 FROM t1 WHERE 'a' BETWEEN 0 AND (c1);
+SELECT c1 FROM t1 IGNORE KEY(c1) WHERE 'a' BETWEEN 0 AND (c1);
+DROP TABLE t1;
+
+CREATE TABLE t1 (c1 BINARY(16), UNIQUE (c1));
+INSERT INTO t1 (c1) VALUES (1);
+INSERT INTO t1 (c1) VALUES (2);
+SELECT c1 FROM t1 WHERE 'a' BETWEEN 0 AND (c1);
+SELECT c1 FROM t1 IGNORE KEY(c1) WHERE 'a' BETWEEN 0 AND (c1);
+DROP TABLE t1;
```
abarkov, line 861: There is a file `type_binary.test` for tests with the `BINARY` data type. This chunk should probably go there.
I also suggest to move the previous chunk for `VARBINARY` to a new file `type_varbinary.test`

sql/item_cmpfunc.h
```diff
@@ -1003,6 +1003,20 @@ class Item_func_opt_neg :public Item_bool_func
 
 class Item_func_between :public Item_func_opt_neg
 {
+  /*
+    If the types of the arguments to BETWEEN permit, then:
+
+    WHERE const1 BETWEEN expr2 AND field1
+      can be optimized to:
+    WHERE const1 <= field1
+
```
abarkov, line 1012: Can you please also add a comment the other way around:
`WHERE const1 BETWEEN field1 AND expr2`
can be optimized like if it was just:
`WHERE const1 >= field1`
?

sql/item_cmpfunc.h
```diff
@@ -1003,6 +1003,20 @@ class Item_func_opt_neg :public Item_bool_func
...
+    as expr2 could be an arbitrary expression.  More generally,
+    this optimization is permitted if aggregation for comparison
+    for three expressions (const1,const2,field1) and for two
+    expressions (const1,field1) return the same type handler.
+   */
+  bool can_optimize_range(Item_field *field_item) const;
+
```
abarkov, line 1019: I think the most usual case for BETWEEN is `field1 BETWEEN expr1 AND expr2`.

`can_optimize_range()` sounds too generic for a method called in a less frequent case.

Perhaps the method could be renamed to something like:
`bool can_optimize_range_const_between_x_y(Item_field *field_item) const;`

Or maybe you have an idea for a better name.

abarkov (changes requested):
The patch generally looks ok to push for me. I have only very small suggestions about tests and comments.

Thanks.

sql/item_cmpfunc.h
DaveGosselin-MariaDB: Good point, but I think this comment is outdated because I have such a comment on the method's prototype in the class (in the latest revision).

sql/item_cmpfunc.h
abarkov: Yes, it supersedes. Thanks.

sql/item_cmpfunc.h
DaveGosselin-MariaDB: `between` is implied because this is in the `Item_func_between` class, so `can_optimize_range_const` is probably best.

abarkov (approved):
Hi Dave. Ok to push after updating the commit comment.
It mentions the old method name: can_optimize_range().

sql/item_cmpfunc.h
```diff
@@ -1003,6 +1003,20 @@ class Item_func_opt_neg :public Item_bool_func
 
 class Item_func_between :public Item_func_opt_neg
 {
+  /*
```
spetrunia, line 1006: Please move the comment to function declaration.

spetrunia (changes requested):
Please look at the comment before `Item_bool_func::get_full_func_mm_tree`.  Does it need to be adjusted after this fix? 

I see there statements like
```
    The validity of the processing of the predicate (c NOT BETWEEN f1i AND f2i)
      which equivalent to (f1i > c OR f2i < c) 
```
which do not seem to be correct.

mysql-test/main/type_binary.result
```diff
@@ -397,3 +397,28 @@ indexed_col	not_indexed_col
...
+#
+# MDEV-36235 Incorrect result for BETWEEN over unique blob prefix
+#
+CREATE TABLE t1 (c1 BINARY(16), UNIQUE (c1));
+INSERT INTO t1 (c1) VALUES (1);
+INSERT INTO t1 (c1) VALUES (2);
+SELECT c1 FROM t1 WHERE 'a' BETWEEN 0 AND (c1);
+c1
+1               
+2               
```
spetrunia, line 409: Note that here you actually have this in the diff:
```
1^@^@^@^@^@^@^@^@^@
2^@^@^@^@^@^@^@^@^@
```
github just doesn't display `\0` correctly.  There are other issues with some merge tools that don't handle \0.
Please use `SELECT hex(c1)` like the rest of this file so this is avoided.

spetrunia (changes requested):

sql/item_cmpfunc.h
DaveGosselin-MariaDB: This is the site of the method's declaration.  Are you asking me to move it to the method's definition, instead?  We aren't consistent in our codebase, so it seems impossible to know which is preferred before code review time.

mysql-test/main/type_binary.result
DaveGosselin-MariaDB: That's a good idea, will do.

DaveGosselin-MariaDB:
> Please look at the comment before `Item_bool_func::get_full_func_mm_tree`. Does it need to be adjusted after this fix?
> 
> I see there statements like
> 
> ```
>     The validity of the processing of the predicate (c NOT BETWEEN f1i AND f2i)
>       which equivalent to (f1i > c OR f2i < c) 
> ```
> 
> which do not seem to be correct.

I'll look at it and adjust as needed.  I don't expect much to change here as a result of this patch because this patch prevents tree construction for unexpected types, but does not fundamentally change how trees are constructed.

I'm not sure what I'm missing as `c NOT BETWEEN x AND y` is equivalent to `(x > c OR y < c)` but I think that's an unusual way to expressit it; I'd rather express the excluded range as the inner part of the disjunction instead, like this: `(c < x OR y < c)`.

abarkov (approved):
Ok to push. Thanks.

PR #3965, opened 2025-04-11
MDEV-36558 Performance schema fails to compile on AIX

grooverdan:
try compile tests are flawed, particularly on AIX as they succeed with an "warning: implicit declaration of function".

Upon compulation of these detected feature there is a link error for the missing function.

AIX can use the HAVE_INTEGER_PTHREAD_SELF implementation

Interface per:
https://www.ibm.com/docs/en/aix/7.3?topic=t-thread-self-kernel-service

- [x] *The Jira issue number for this PR is: MDEV-36558*

## Description

Enable Performance Schema to compile on AIX.

## Release Notes

Corrected compilation of performance schema on AIX

## How can this PR be tested?

Compile under AIX:

With patch:
```
-- Performing Test HAVE_PTHREAD_THREADID_NP
-- Performing Test HAVE_PTHREAD_THREADID_NP - Failed
-- Looking for gettid
-- Looking for gettid - not found
-- Performing Test HAVE_SYS_GETTID
-- Performing Test HAVE_SYS_GETTID - Failed
-- Performing Test HAVE_GETTHRID
-- Performing Test HAVE_GETTHRID - Failed
-- Performing Test HAVE_PTHREAD_GETTHREADID_NP
-- Performing Test HAVE_PTHREAD_GETTHREADID_NP - Failed
-- Performing Test HAVE_INTEGER_PTHREAD_SELF
-- Performing Test HAVE_INTEGER_PTHREAD_SELF - Success

buildbot@mariadb:[/home/buildbot/build_10.6]make -j16 perfschema

[100%] Linking CXX static library libperfschema.a
[100%] Built target perfschema
```

## PR quality check

### DIFF
--- a/storage/perfschema/CMakeLists.txt
+++ b/storage/perfschema/CMakeLists.txt
@@ -280,6 +280,11 @@ table_replication_applier_status_by_worker.cc
 #table_replication_group_member_stats.cc
 )
 
+# These may warning on warning: implicit declaration of function
+# which isn't good enough as a compile test.
+set(SAVED_CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
+set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -Wall -Werror")
+
 # Check for pthread_threadid_np()
 CHECK_C_SOURCE_COMPILES("
 #include <pthread.h>
@@ -325,6 +330,9 @@ int main(int ac, char **av)
   return (tid != 0 ? 0 : 1);
 }"
 HAVE_PTHREAD_GETTHREADID_NP)
+
+# Restore original flags
+set(CMAKE_REQUIRED_FLAGS "${SAVED_CMAKE_REQUIRED_FLAGS}")
 
 # Check for pthread_self() returning an integer type
 CHECK_C_SOURCE_COMPILES("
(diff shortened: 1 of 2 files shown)


storage/perfschema/CMakeLists.txt
```diff
@@ -280,6 +280,11 @@ table_replication_applier_status_by_worker.cc
 #table_replication_group_member_stats.cc
 )
 
+# These may warning on warning: implicit declaration of function
+# which isn't good enough as a compile test.
+set(SAVED_CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
```
vaintroub, line 285: Nah, this is proprietary, an the amount of warnins depends on the version of GCC. Better, CHECK_CXX_SOURCE_COMPILES ,  i.e C++ instead of C, there are no "implicit declaration of a function" in C++.
Also, a link error should by caught by CHECK_C_SOURCE_COMPILES  (From the documentation "Check once that the source supplied in <code> can be built, linked as an executable, and then run. The <code> must contain at least a main() function") ,but perhaps AIX linker returns 0 even when it fails, I remember some irregularities.

vaintroub: There  is probably something about this funny linker option 
storage/perfschema/CMakeLists.txt:  
```cmake
SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-berok")
```
that allows linker to return 0 if there are unresolved symbols :)
We can blame cmake, and try_compile, but if we use this option, all bets are off

vaintroub: I suggest removing these "beroks", and either do not build the test executables (there is really not much value in them)
or, if absolutely necessary, to add
``cmake
if(AIX)
  target_link_options(target PRIVATE "-Wl,-berok") for targets that fail.
endif()
```
This is preferable to changing CMAKE_EXE_LINKER_FLAGS, and has less side-effect

Though, personally, I'd rather *not* workaround linker bugs, and build unimportant unit-and-other-test executables with if(NOT AIX)

PR #3966, opened 2025-04-11
MDEV-19749 MDL scalability regression after backup locks

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-19749*

## Description
The `MDL_BACKUP_COMMIT` lock used in `ha_commit_trans()`, which ensures mutual exclusion of `COMMIT` and some `XA` statements with concurrent `FLUSH TABLES WITH READ LOCK` or `BACKUP STAGE BLOCK_COMMIT`, can have a notable performance impact.

To avoid acquiring an expensive MDL on each transaction, we will introduce an atomic counter `backup_commit_lock_enabled`, which counts in-progress `FLUSH TABLES WITH READ LOCK` or `BACKUP STAGE BLOCK_COMMIT` operations. Only if this counter is nonzero, `MDL_BACKUP_COMMIT` lock will be acquired.

The member variable `THD::backup_commit_lock` is changed from a pointer to `std::atomic<backup_locking_state>`, which will indicate whether a protected transaction state change is in progress and whether that operation is being protected by `THD::mdl_backup`.

`enable_backup_commit_locks()`: Increment the `backup_commit_lock_enabled` counter and wait until there is no `THD` in `server_threads` for which `check_if_unblocked()` would hold.

`disable_backup_commit_locks()`: Decrement the `disable_backup_commit_locks` counter.

`THD::protect_against_backup()`: If `backup_commit_lock_enabled`, try to acquire `mdl_backup` lock.  Set the `backup_commit_lock` state.

`THD::unprotect_against_backup()`: Clear the `backup_commit_lock` state and release `mdl_backup` if it had been acquired.

The effect is that we replace unconditional MDL acquisition and release with an extra load, conditional branch and a store in the fast path of `THD::protect_against_backup()`.  An extra penalty is paid in `BACKUP STAGE START` and `FLUSH TABLES WITH READ LOCK`, in the function `enable_backup_commit_locks()`.
## Release Notes
The `information_schema.processlist.state` values `'Enabling backup commit lock'` and `'Backup commit lock enabled'` were introduced to the `FLUSH TABLES WITH READ LOCK` and `BACKUP STAGE START` statements.
## How can this PR be tested?
Performance testing, similar to that one conducted for #3925.

This bug affects older major versions as well, but we avoid intrusive changes for older branches, such as 10.6.
## PR quality check


sql/backup.cc
```diff
@@ -191,16 +211,120 @@ static bool backup_start(THD *thd)
...
+    mdl backup locks until both are finished.
+  */
+  mysql_mutex_lock(&LOCK_backup_commit);
+  backup_commit_lock_enabled.fetch_add(1, std::memory_order_acq_rel);
+  mysql_mutex_unlock(&LOCK_backup_commit);
+
+  /*
+    Ensure that all threads are doing commits with mdl backup locks active.
+  */
+  while (server_threads.iterate(check_if_unblocked, static_cast<void*>(0)))
```
dr-m, line 266: @montywi In your original version of this (8691412d6bc6d8e40b7fa15ea22f7afc9d4fd9bd), `backup_commit_lock_enabled` was a regular variable. Increments and decrements were protected by a mutex, but `have_backup_commit_locks()` was not. I was unsure if the mutex is requierd for any other reason than for preventing race conditions between increments and decrements, so I did not remove the mutex acquisition here.

I reached the conclusion that we may suffer from an unlikely race condition here. Maybe the mutex is protecting against that; I didn’t check it. The scenario would be as follows:
1. A thread enters `protect_against_backup()` and reads `backup_commit_lock_enabled` as 0.
2. That thread will be blocked for quite a while, allowing `enable_backup_commit_locks()` to complete.
3. The thread resumes execution and skips the MDL.

The additional `while` loop in 8691412d6bc6d8e40b7fa15ea22f7afc9d4fd9bd does not seem to provide any additional security against this.

The race condition should be unlikely if `enable_backup_commit_locks()` is executed as part of `BACKUP STAGE START` and the backup locks are not being consulted until `BACKUP STAGE BLOCK_COMMIT`, which would be executed much later. That is, the committing thread should have a fair chance of completing before the actual critical section is entered.

sql/sql_class.h
```diff
@@ -2736,10 +2736,49 @@ class THD: public THD_count, /* this must be first */
...
+  bool protect_against_backup() noexcept
+  {
+    DBUG_ASSERT(!backup_commit_lock.load(std::memory_order_relaxed));
+    backup_commit_lock.store(IN_COMMIT_NO_LOCK, std::memory_order_release);
+    /* We must have updated our state before the load below, to guarantee that
+    enable_backup_commit_locks() will account for us. */
+    return
+      unlikely(backup_commit_lock_enabled.load(std::memory_order_acquire)) &&
+      protect_against_backup_slow();
+  }
```
dr-m, line 2768: @montywi pointed out in https://lists.mariadb.org/hyperkitty/list/commits@lists.mariadb.org/thread/2UXN53DK54E7MDTXIHP5S7AYIPP4RCMO/ that there is a possibility of starvation here. It seems that we must invoke `backup_commit_lock_enabled.load()` before assigning `backup_commit_lock`. Maybe the following would work:
```c++
    if (unlikely(backup_commit_lock_enabled.load(std::memory_order_relaxed)))
    slow:
      return protect_against_backup_slow();
    backup_commit_lock.store(IN_COMMIT_NO_LOCK, std::memory_order_release);
    /* We must have updated our state before the load below, to guarantee that
    enable_backup_commit_locks() will account for us. */
    if (unlikely(backup_commit_lock_enabled.load(std::memory_order_acquire)))
      goto slow;
    return false;
```

dr-m:
d1ee025d5554601f88d3b8f2bcc10b2f3a151fbe supercedes this.

dr-m:
This was basically just a clean-up of some changes that did not show any improvement in performance tests. #4048 supercedes this.

PR #1674, opened 2020-10-04
MDEV-23077: flashbacking the update/delete + join statements has some bugs

lukemakeit:
1. Extend the --table --database option of mysqlbinlog to allow --table and --database to support plural values;
2. Fix bug 2 and bug 3 described by [MDEV-23077](https://jira.mariadb.org/browse/MDEV-23077);

PR #3968, opened 2025-04-12
MDEV-23077: --flashback + --table produces buggy results

ParadoxV5:
* Split from and built upon #1674
  * [x] *The Jira issue number for this PR is: [MDEV-23077](https://jira.mariadb.org/browse/MDEV-23077)*

## What problem is the patch trying to solve?
* Events of skipped tables were still queued and generates extraneous “Row event for unknown table” when flashing back.
* Skipping a statement-end event left prior non-skipped events processed incompletely.

## How can this PR be tested?
#1674 provided a `< CASE 8 >` for `binlog.flashback` which I simplified and corrected (?) to cover both bullet points.
It also adds a `< CASE 9 >` checking the similar `--database` excludes the not-specified database.

## PR quality check
  * #1674 included changes that adds multi-choice support for `--database` and `--table`.
    This PR excludes them.

### DIFF
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -884,7 +884,7 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
   char ll_buff[21];
   bool result= 0;
 
-  if (opt_flashback)
+  if (opt_flashback && !skip_event)
   {
     Rows_log_event *e= (Rows_log_event*) ev;
     // The last Row_log_event will be the first event in Flashback
@@ -936,7 +936,7 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
        result_file (as it would happen in ev->print(...) if
        event was not skipped).
     */
-    if (skip_event)
+    if (skip_event && !opt_flashback)
     {
       // append END-MARKER(') with delimiter
       IO_CACHE *const body_cache= &print_event_info->body_cache;
@@ -955,14 +955,14 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
   }
 
   /* skip the event check */
-  if (skip_event)
+  if (skip_event && !(opt_flashback && is_stmt_end))
     return 0;
 
   if (!opt_flashback)
     result= print_base64(print_event_info, ev);
   else
   {
-    if (is_stmt_end)
+    if (is_stmt_end && events_in_stmt.elements)
     {
       Log_event *e= NULL;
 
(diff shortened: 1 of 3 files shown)


mysql-test/suite/binlog/t/flashback.test
```diff
@@ -364,13 +364,94 @@ FLUSH LOGS;
...
+DROP DATABASE d2;
+
 --echo #
 --echo # MDEV-30698 Cover missing test cases for mariadb-binlog options
 --echo #            --raw [and] --flashback
 --echo #
 
 --error 1 # --raw mode and --flashback mode are not allowed
---exec $MYSQL_BINLOG -vv -B --raw --read-from-remote-server --user=root --host=127.0.0.1 --port=$MASTER_MYPORT master-bin.000003> $MYSQLTEST_VARDIR/tmp/mysqlbinlog_row_flashback_8.sql
+--exec $MYSQL_BINLOG -vv -B --raw --read-from-remote-server --user=root --host=127.0.0.1 --port=$MASTER_MYPORT master-bin.000003> $MYSQLTEST_VARDIR/tmp/mysqlbinlog_row_flashback_0.sql
```
ParadoxV5, line 454: We have a real Case 8 now.

mysql-test/suite/binlog/t/flashback.test
```diff
@@ -364,13 +364,94 @@ FLUSH LOGS;
...
+--let SEARCH_FILE= $MYSQLTEST_VARDIR/tmp/mysqlbinlog_row_flashback_8_2.sql
+--let SEARCH_PATTERN= ### Row event for unknown table
+--source include/search_pattern_in_file.inc
+
+DROP TABLE t1;
+DROP TABLE t2;
+
+--echo # < CASE 9 >
+--echo # --flashback + --database
+--echo #
```
ParadoxV5, line 417: TODO: I’m concerned if `--database` also suffer*ed* from the same issue.

client/mysqlbinlog.cc
```diff
@@ -955,14 +955,14 @@ static bool print_row_event(PRINT_EVENT_INFO *print_event_info, Log_event *ev,
   }
 
   /* skip the event check */
-  if (skip_event)
+  if (skip_event && !(opt_flashback && is_stmt_end))
```
ParadoxV5, line 958: What #1674 did was copy-pasting the `else` branch of the next `if` to the previous `if` tree as a new branch for `(skip_event && opt_flashback && is_stmt_end)`.
Here, I reimplement it by skipping this intermediate “step” (?) for that exact condition.

While the diff is now smaller, it has become proof that our code is 🍝.

mysql-test/suite/binlog/t/flashback.test
ParadoxV5: ⚠️ Reproduced both bugs for both `--table` __and `--database`__ on one of my `main`-based branches

ParadoxV5:
* [ ] TODO: I forgot to clarify in Commit 2’s descripton that the bugs described were already fixed in Commit 1.
  If I drop by the past again, I should amend it back in.

mysql-test/suite/binlog/t/flashback.test
ParadoxV5: I tested that `--database` is also affected by the same bugs, but is also fixed by the same patch.

ParadoxV5:
There were reports of binlog event memory leaks at the time I ported this contributed bug fix.
I don’t remember my build configs at that time, and apparently don’t have MSAN on my new device, so I don’t know if this latest commit is it.
All I know is that this delta does not appear to be necessary for passing the new test cases, so I might as well skip it.

PR #3974, opened 2025-04-15
MDEV-36340: Reset `Connects_Tried` with `Master_Retry_Count=X`

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-36340](https://jira.mariadb.org/browse/MDEV-36340)*
  * Follow-up of [MDEV-35304](https://jira.mariadb.org/browse/MDEV-35304)

## Description
CHANGE MASTER can set `Master_Retry_Count` to be lower than `Connects_Tried`, which’ll look strange to those unaware of that CHANGE.

Now, setting `Master_Retry_Count` also resets `Connects_Tried` to 0.

## Release Notes
Amend MDEV-35304’s with:
* STOP REPLICA does not reset `Connects_Tried` to 0, but RESET REPLICA does, and also does changing `Master_Retry_Count` with CHANGE MASTER.

## How can this PR be tested?
This PR adds a test case to `multi_source.connects_tried`.

This PR also
* swaps some other tests’ SELECTs so they’re all `FROM status_before JOIN status_after USING(Connection_name)`; no function change
* “simplifies” START/STOP SLAVE + `wait_for_slave_to_start`/`stop.inc` to `start`/`stop_slave.inc`

## PR quality check
The bug is about a feature exclusive to the `main` branch. Or is this a new sub-feature?

### DIFF
--- a/mysql-test/suite/multi_source/connects_tried.test
+++ b/mysql-test/suite/multi_source/connects_tried.test
@@ -122,27 +122,37 @@ SELECT *
   FROM status_after JOIN status_stop USING(Connection_name)
   WHERE status_stop.Connects_Tried <> status_after.Connects_Tried;
 
-START SLAVE;
---source include/wait_for_slave_to_start.inc
+--source include/start_slave.inc
 --echo # START SLAVE recounts `Connects_Tried` from 1 (for the restarted connection only).
 CREATE TEMPORARY TABLE status_restart AS
   SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
 SELECT *
-  FROM status_restart JOIN status_stop USING(Connection_name)
+  FROM status_stop JOIN status_restart USING(Connection_name)
   WHERE status_restart.Connects_Tried NOT BETWEEN IF(
     Connection_name = '', 1, status_stop.Connects_Tried
   ) AND status_stop.Connects_Tried;
 
-STOP SLAVE;
---source include/wait_for_slave_to_stop.inc
+--source include/stop_slave.inc
+# MDEV-36340 Master_Retry_Count should never be less than Connects_Tried
+CHANGE MASTER TO Master_Retry_Count=777;
+--echo # Setting `Master_Retry_Count` resets `Connects_Tried` to 0 (for the changed connection only).
+CREATE TEMPORARY TABLE status_change AS
+  SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
+SELECT *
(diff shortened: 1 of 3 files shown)


mysql-test/suite/multi_source/connects_tried.test
```diff
@@ -105,27 +105,37 @@ SELECT *
...
 RESET SLAVE;
 --echo # RESET SLAVE resets `Connects_Tried` to 0 (for the resetted connection only).
 CREATE TEMPORARY TABLE status_reset AS
   SELECT Connection_name, Connects_Tried FROM information_schema.SLAVE_STATUS;
 SELECT *
-  FROM status_reset JOIN status_restart USING(Connection_name)
+  FROM status_change JOIN status_reset USING(Connection_name)
   WHERE status_reset.Connects_Tried <>
-    IF(LENGTH(Connection_name), status_restart.Connects_Tried, 0);
+    IF(Connection_name = '', 0, status_change.Connects_Tried);
```
ParadoxV5, line 138: This format to match #3897.
Conflict Resolution: this wins.

mariadb-SusilBehera (approved):
Looks good!

sql/sql_repl.cc
```diff
@@ -4052,7 +4052,10 @@ bool change_master(THD* thd, Master_info* mi, bool *master_info_added)
   if (lex_mi->connect_retry)
     mi->connect_retry = lex_mi->connect_retry;
   if (lex_mi->retry_count)
+  {
     mi->retry_count= lex_mi->retry_count;
+    mi->connects_tried= 0;
```
bnestere, line 4057: Would it make sense to reset this for all `CHANGE MASTER` executions, rather than just when 
 MASTER_RETRY_COUNT` is provided? That is, if we are changing the connection parameters (e.g. `MASTER_HOST`), it would make sense to me that we shouldn't report the old `connects_tried` for a new connection.

bnestere (commented):
Hi @ParadoxV5 ! 

Thanks for a good patch and test - I just left one question.

sql/sql_repl.cc
ParadoxV5: I say naw, because other parameters, including `MASTER_HOST` and co., don’t relate to `Connects_Tried`, and so have no reason to clear it when it’s intended to remain past STOP REPLICA.

bnestere (approved):

ParadoxV5:
@bnestere says in JIRA that QA’s done. Now, (still) waiting for the main branch to switch to DST… 🫤

ParadoxV5:
> This branch cannot be rebased due to conflicts

manual update time

PR #3976, opened 2025-04-15
MDEV-36234: Dual stack libaio/liburing

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36234*
## Description
FIXME: Introduce a parameter for explicitly specifying which one to use: innodb_linux_aio=libaio/liburing/auto

FIXME: In `srv_start()`, indicate which implementation was chosen
## Release Notes
TBD
## How can this PR be tested?
TBD

We may want to base this on 10.11 instead of 10.6, if there is no `liburing` capable GNU/Linux distribution that would by default ship 10.6.
## PR quality check


tpool/aio_linux.cc
```diff
@@ -174,11 +170,20 @@ class aio_linux final : public aio
...
+}
+
+namespace tpool
+{
 
-aio *create_linux_aio(thread_pool *pool, int max_io)
+#ifdef HAVE_URING
+aio *create_libaio(thread_pool *pool, int max_io)
+#else
+aio *create_linux_aio(thread_pool *pool, int max_io, aio_implementation implementation __attribute__((unused)))
```
vaintroub, line 185: Drop the `implementation __attribute__((unused))` from this line. Just `aio_implementation`. It's cleaner

It is C++, if you do not have to name parameter, if you do not use them.

vaintroub (approved):
Fixed Windows compilation. Other than that, and minor commit on unnecessary __attribute__(unused), I do not have any objections

storage/innobase/os/os0file.cc
```diff
@@ -3387,35 +3387,50 @@ int os_aio_init() noexcept
...
+#ifdef __linux__
+  if (srv_use_native_aio)
   {
-    ut_ad(srv_use_native_aio);
-disable:
-    ib::warn() << "Linux Native AIO disabled.";
-    srv_use_native_aio= false;
-    ret= srv_thread_pool->configure_aio(false, max_events);
-  }
+    switch (srv_linux_aio_method)
```
vaintroub, line 3395: Suggestion : I think most of this `if linux` fallback logic (uring to libaio) could be handled inside tpool::configure_aio itself, you got the OS_DEFAULT for it, right? Will be less clutter here and you can put OS_DEFAULT  to use

tpool/tpool_generic.cc
```diff
@@ -300,12 +299,12 @@ class thread_pool_generic : public thread_pool
   void wait_begin() override;
   void wait_end() override;
   void submit_task(task *task) override;
-  aio *create_native_aio(int max_io) override
+  aio *create_native_aio(int max_io, aio_implementation implementation) override
   {
 #ifdef _WIN32
     return create_win_aio(this, max_io);
```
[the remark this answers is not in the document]
dr-m:
```
/home/buildbot/amd64-debian-11-msan-clang-16/build/tpool/tpool_generic.cc:307:12: error: no matching function for call to 'create_linux_aio'
    return create_linux_aio(this, max_io, implementation);
           ^~~~~~~~~~~~~~~~
/home/buildbot/amd64-debian-11-msan-clang-16/build/tpool/tpool_generic.cc:45:8: note: candidate function not viable: requires 2 arguments, but 3 were provided
  aio *create_linux_aio(thread_pool *, int) { return nullptr; };
```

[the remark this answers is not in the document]
grooverdan: per MDEV I did an update on bb-10.6-MDEV-36234-pkgtest (which is actually 10.11, opps) that fixes this difference. Is probably enough for testing (resizing has a fault if i remember) though I need to take @vaintroub's feedback and refactor the handling into tpool.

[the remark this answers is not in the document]
dr-m: Fixed in 322b6a4b15595219dc0db9219d44c8d22da12bbe. I could reproduce this by the following:
```bash
cmake -DCMAKE_DISABLE_FIND_PACKAGE_{URING,LIBAIO}=ON …
```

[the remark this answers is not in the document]
dr-m: The Copilot warning is actually about `-Wunused-parameter` on Windows.

storage/innobase/os/os0file.cc
```diff
@@ -3387,35 +3387,50 @@ int os_aio_init() noexcept
...
+      else
+        break;
+      /* fallthough */
+#endif /* HAVE_URING */
+    case SRV_LINUX_AIO_LIBAIO:
+      ret= srv_thread_pool->configure_aio(srv_use_native_aio,
+                                          max_events,
+                                          tpool::OS_AIO);
+      if (!ret && !is_linux_native_aio_supported())
+      {
```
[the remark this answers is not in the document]
dr-m: According to `strace -p $(pgrep mariadbd)` in my environment, `libaio` on tmpfs will work just fine; I did see `io_submit()` calls for page writes on InnoDB shutdown.

What was the `is_linux_native_aio_supported()` originally about? It had been refactored in mysql/mysql-server@98909cefbc37e54efc6452c7e95bccbf64ac9213 in 2015 and originally added in mysql/mysql-server@0da310b69db3a39ba2e6d63ff62ef3c027cd63ff in 2012 to address [MySQL Bug #58421 InnoDB temporary table crash when tmpdir on Linux tmpfs](https://bugs.mysql.com/bug.php?id=58421). The system that is reported there is CentOS 5.5, and it is about temporary tables, which actually were handled mostly as persistent tables back then. I believe that this check got useless in [MySQL WL#6560: InnoDB: separate tablespace for innodb-temp-tables](https://dev.mysql.com/worklog/task/?id=6560). Temporary tables would be written to the `innodb_temp_data_file_path` (default file name `ibtmp1`), no longer into any `/tmp/#sql-something.ibd` file. The error was that `io_submit()` would return `EINVAL` when writing to a file on `tmpfs`.

In https://stackoverflow.com/questions/34572559/asynchronous-io-io-submit-latency-in-ubuntu-linux/46377629 the accepted answer from 2017 suggests that `libaio` would always work (after `io_setup()` succeeded), but the requests could silently be converted to synchronous ones when `O_DIRECT` is not being used.  This does not seem to entirely hold for CentOS 5, but CentOS 5, which was originally released in 2007, reached its end-of-life in March 2017.  I find it reasonable to expect that any currently maintained GNU/Linux distributions, even including the recently-EOLed RHEL 7, would just work. In the worst case, someone would have to specify `innodb_use_native_aio=OFF` in order to be able to start up InnoDB when their data directory or the `innodb_temp_data_file_path` are in tmpfs. Both would seem to be rather insane choices in a production environment.

I hoped that we could test RHEL 7 on our CI system, which runs the `mtr` tests on `/dev/shm` (tmpfs), but it turns out that https://buildbot.mariadb.org/#/builders/222 reports a way too new kernel version (that of the Docker host system):
```
Linux 2930783f10e5 6.8.0-59-generic #61-Ubuntu SMP PREEMPT_DYNAMIC Fri Apr 11 23:16:11 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```
In any case, I will remove `is_linux_native_aio_supported()`. It would be useful to test that on an actual RHEL 7 kernel to be on the safe side. After all, the problem was claimed to affect "old kernels" already in 2012, and RHEL 7 was released in 2014.

storage/innobase/os/os0file.cc
dr-m: `sql_print_warning()` is not being used in `tpool`. What we can do is to remove some `#ifdef LINUX_NATIVE_AIO` checks that I believe are redundant by now.

grooverdan (commented):
With Native AIO only the mariabackup tests failed:

```
$ mysql-test/mtr  --strace --client-strace  mariabackup.slave_provision_nolock
Logging: /home/dan/repos/mariadb-server-10.11/mysql-test/mariadb-test-run.pl  --strace --client-strace mariabackup.slave_provision_nolock
VS config: 
vardir: /home/dan/repos/build-mariadb-server-10.11/mysql-test/var
Checking leftover processes...
Removing old var directory...
Creating var directory '/home/dan/repos/build-mariadb-server-10.11/mysql-test/var'...
MariaDB Version 10.11.12-MariaDB
 - SSL connections supported
 - binaries built with wsrep patch
Collecting tests...
Installing system database...

mariabackup.slave_provision_nolock       [ fail ]
        Test ended at 2025-05-21 16:43:58

CURRENT_TEST: mariabackup.slave_provision_nolock
Warning: /home/dan/repos/build-mariadb-server-10.11/client//mariadb-test: unknown option '--loose-debug-gdb'
Warning: /home/dan/repos/build-mariadb-server-10.11/client//mariadb-test: unknown option '--loose-skip-stack-trace'
[00] 2025-05-21 16:38:48 Connecting to MariaDB server host: localhost, user: root, password: set, port: 19000, socket: /home/dan/repos/build-mariadb-server-10.11/mysql-test/var/tmp/mysqld.1.sock
[00] 2025-05-21 16:38:48 Using server version 10.11.12-MariaDB-log
....
[00] 2025-05-21 16:43:51 Finished backing up non-InnoDB tables and files
[00] 2025-05-21 16:43:51 Waiting for log copy thread to read lsn 282202708
[00] 2025-05-21 16:43:52 Retrying read of log at LSN=48736
[00] 2025-05-21 16:43:53 Retrying read of log at LSN=48736
[00] 2025-05-21 16:43:54 Retrying read of log at LSN=48736
[00] 2025-05-21 16:43:55 Retrying read of log at LSN=48736
[00] 2025-05-21 16:43:56 Was only able to copy log from 48736 to 48736, not 282202708; try increasing innodb_log_file_size
mariabackup: Stopping log copying thread[00] 2025-05-21 16:43:56 Retrying read of log at LSN=48736
.
mysqltest: At line 100: exec of '/home/dan/repos/build-mariadb-server-10.11/extra/mariabackup/mariabackup --defaults-file=/home/dan/repos/build-mariadb-server-10.11/mysql-test/var/my.cnf --backup --target-dir=/home/dan/repos/build-mariadb-server-10.11/mysql-test/var/tmp/backup --no-lock' failed, error: 256, status: 1, errno: 11
Output from before failure:
*** Doing backup...



The result from queries just before the failure was:
call mtr.add_suppression("Can't init tc log");
call mtr.add_suppression("Aborting");
# restart
RESET MASTER;
CREATE TABLE t1(a varchar(60) PRIMARY KEY, b VARCHAR(60)) ENGINE INNODB;
INSERT INTO t1 VALUES(1, NULL);
CREATE TABLE t2 (val INT) ENGINE=InnoDB;
INSERT INTO t2 VALUES (0);
connect  con1,localhost,root,,;
*** Start a background load...
CALL gen_load();
connection default;
*** Doing backup...

 - saving
 ```

Once `io_uring` is linked it functions correctly - on and off `--mem` (ext4 alternative).

I find myself on a old machine - 6.4.12-200.fc38.x86_64 - will test again after kernel upgrade (because fc38 is definitely EOL)

```
$ firejail  --noprofile  '--seccomp=io_uring_setup:ENOSYS,!io_setup'  sql/mariadbd --no-defaults --bootstrap --datadir=$(mktemp -d) 
Seccomp list in: io_uring_setup:ENOSYS,!io_setup, check list: @default-keep, prelist: io_uring_setup:ENOSYS,unknown,
Parent pid 655069, child pid 655070
Seccomp list in: io_uring_setup:ENOSYS,!io_setup, check list: @default-keep, prelist: io_uring_setup:ENOSYS,unknown,
Child process initialized in 18.72 ms
2025-05-21 17:49:15 0 [Note] Starting MariaDB 10.11.12-MariaDB source revision 1e033232260abae44089544877e1fc4462c8219c server_uid tluFI/7BR1nZc5vueEzXfY8fsIE= as process 4
2025-05-21 17:49:15 0 [Note] InnoDB: The first data file './ibdata1' did not exist. A new tablespace will be created!
2025-05-21 17:49:15 0 [Note] InnoDB: Compressed tables use zlib 1.2.13
2025-05-21 17:49:15 0 [Note] InnoDB: Number of transaction pools: 1
2025-05-21 17:49:15 0 [Note] InnoDB: Using crc32 + pclmulqdq instructions
2025-05-21 17:49:15 0 [Warning] mariadbd: io_uring_queue_init() failed with ENOSYS: check seccomp filters, and the kernel version (newer than 5.1 required)
2025-05-21 17:49:15 0 [Warning] InnoDB: io_uring failed: falling back to libaio
io_getevents returned -1
250521 17:49:15 [ERROR] sql/mariadbd got signal 6 ;
```

Bit of an eccentric case that io_getevents failed.

File: include/my_time.h [hunk 5]
```diff
@@ -230,7 +230,6 @@ static inline longlong sec_part_unshift(longlong second_part, uint digits)
 /* Date/time rounding and truncation functions */
 static inline long my_time_fraction_remainder(long nr, uint decimals)
 {
-  DBUG_ASSERT(decimals <= TIME_SECOND_PART_DIGITS);
   return nr % (long) log_10_int[TIME_SECOND_PART_DIGITS - decimals];
```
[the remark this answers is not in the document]
dr-m: That change was included accidentally. Removing the assertion removes the following compilation warning for me:
```
FAILED: sql/CMakeFiles/sql.dir/field.cc.o 
/usr/bin/c++ -DHAVE_CONFIG_H -DHAVE_EVENT_SCHEDULER -DHAVE_POOL_OF_THREADS -DMYSQL_SERVER -D_FILE_OFFSET_BITS=64 -I/mariadb/10.11/wsrep-lib/include -I/mariadb/10.11/wsrep-lib/wsrep-API/v26 -I/dev/shm/10.11/include -I/mariadb/10.11/include/providers -I/mariadb/10.11/include -I/mariadb/10.11/sql -I/dev/shm/10.11/sql -I/mariadb/10.11/tpool -O2 -march=native -pie -fPIC -fstack-protector --param=ssp-buffer-size=4 -fPIC -g -DPROTECT_STATEMENT_MEMROOT -DENABLED_DEBUG_SYNC -D_GLIBCXX_DEBUG -D_GLIBCXX_ASSERTIONS -ggdb3 -DSAFE_MUTEX -DTRASH_FREED_MEMORY -Wall -Wenum-compare -Wenum-conversion -Wextra -Wmissing-braces -Wno-format-truncation -Wno-init-self -Wno-nonnull-compare -Wno-unused-parameter -Wnon-virtual-dtor -Woverloaded-virtual -Wsuggest-override -Wvla -Wwrite-strings -Werror -fno-operator-names -std=gnu++11   -DHAVE_OPENSSL -DOPENSSL_API_COMPAT=0x10100000L -MD -MT sql/CMakeFiles/sql.dir/field.cc.o -MF sql/CMakeFiles/sql.dir/field.cc.o.d -o sql/CMakeFiles/sql.dir/field.cc.o -c /mariadb/10.11/sql/field.cc
In file included from /mariadb/10.11/sql/structs.h:26,
                 from /mariadb/10.11/sql/handler.h:34,
                 from /mariadb/10.11/sql/log.h:20,
                 from /mariadb/10.11/sql/sql_class.h:28,
                 from /mariadb/10.11/sql/procedure.h:31,
                 from /mariadb/10.11/sql/sql_select.h:31,
                 from /mariadb/10.11/sql/field.cc:32:
In function 'long int my_time_fraction_remainder(long int, uint)',
    inlined from 'long int Timestamp::fraction_remainder(uint) const' at /mariadb/10.11/sql/sql_type.h:2815:38,
    inlined from 'Timestamp& Timestamp::round(uint, time_round_mode_t, int*)' at /mariadb/10.11/sql/sql_type.h:2831:7,
    inlined from 'virtual int Field_timestamp::store_timestamp_dec(const timeval&, uint)' at /mariadb/10.11/sql/field.cc:5411:42:
/mariadb/10.11/include/my_time.h:234:67: error: array subscript 4294901767 is above array bounds of 'ulonglong [20]' {aka 'long long unsigned int [20]'} [-Werror=array-bounds=]
  234 |   return nr % (long) log_10_int[TIME_SECOND_PART_DIGITS - decimals];
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
/mariadb/10.11/include/my_time.h: In member function 'virtual int Field_timestamp::store_timestamp_dec(const timeval&, uint)':
/mariadb/10.11/include/my_time.h:30:38: note: while referencing 'log_10_int'
   30 | extern MYSQL_PLUGIN_IMPORT ulonglong log_10_int[20];
      |                                      ^~~~~~~~~~
```

dr-m:
> Bit of an eccentric case that io_getevents failed.

I’m not sure if it is related, but there is a relationship between `io_setup(2)` failures and `/proc/sys/fs/aio-max-nr`. Did you try a larger parameter in the kernel?

grooverdan:
> > Bit of an eccentric case that io_getevents failed.
> 
> I’m not sure if it is related, but there is a relationship between `io_setup(2)` failures and `/proc/sys/fs/aio-max-nr`. Did you try a larger parameter in the kernel?

I should of explained this more - the `firejail` command was allowing `io_setup` however the other syscall of aio, `io_getevents` was still blocked hence it returning -1 and. I Checked the tpool and the `abort()` is an explicit call. so probably nothing required (unless there's an easy fatal termination that doesn't incur stack traces).

On retesting the mariabackup, with a firejail disable (below) and with libaio natively as option (validated by server error logs) both where passing mariabackup tests. No idea what was failing yesterday.

On a uring enabled build you can disable libaio as follows to test.
```
firejail  --noprofile  --seccomp-error-action=ENOSYS '--seccomp=io_uring_setup,!io_setup,!io_getevents,!io_submit,!io_cancel,!io_destroy'  mysql-test/mtr --suite=mariabackup
```

grooverdan (approved):
Sorry for confusing. mariabackup running successful on libaio native.

tpool/tpool.h
```diff
@@ -256,6 +277,10 @@ class thread_pool
   {
     m_aio.reset();
   }
+  const char *get_implementation() const
```
vaintroub, line 280: please call it get_aio_implementation. It is not threadpool implementation

tpool/tpool.h
```diff
@@ -244,7 +260,12 @@ class thread_pool
     assert(m_aio);
     if (use_native_aio)
     {
-      auto new_aio = create_native_aio(max_io);
+      const aio_implementation impl=
+#ifdef LINUX_NATIVE_AIO
```
vaintroub, line 264: Can this ifdef be put  inside create_native_aio? So that LINUX_NATIVE_AIO would not have to be defined by the API caller, and can be an internal detail of the tpool.

storage/innobase/os/os0file.cc
```diff
@@ -3084,168 +3080,53 @@ static void write_io_callback(void *c)
...
-    ut_ad(srv_use_native_aio);
-disable:
-    ib::warn() << "Linux Native AIO disabled.";
-    srv_use_native_aio= false;
-    ret= srv_thread_pool->configure_aio(false, max_events);
-  }
+    switch (srv_linux_aio_method) {
+    case SRV_LINUX_AIO_AUTO:
+    case SRV_LINUX_AIO_IO_URING:
+#ifdef HAVE_URING
```
vaintroub, line 3098: I'd prefer HAVE_URING and  HAVE_LINUX_NATIVE_AIO to be contained inside tpool and not used outside of it.
my_printf_error  with ME_WARNING inside tpool is used by uring  already, though I think fprintf is less dependencies. if you need some reporting

storage/innobase/srv/srv0start.cc
```diff
@@ -1292,14 +1292,10 @@ dberr_t srv_start(bool create_new_db)
...
+#if defined(LINUX_NATIVE_AIO) || defined(HAVE_URING)
 	if (srv_use_native_aio) {
-		ib::info() << "Using Linux native AIO";
-	}
-#endif
-#ifdef HAVE_URING
-	if (srv_use_native_aio) {
-		ib::info() << "Using liburing";
+		sql_print_information("InnoDB: Using %s",
+				      srv_thread_pool->get_implementation());
```
vaintroub, line 1298: Yes, get_aio_implementation is more appropriate name here.

storage/innobase/srv/srv0start.cc
```diff
@@ -1292,14 +1292,10 @@ dberr_t srv_start(bool create_new_db)
 		return(srv_init_abort(DB_ERROR));
 	}
 
-#ifdef LINUX_NATIVE_AIO
+#if defined(LINUX_NATIVE_AIO) || defined(HAVE_URING)
```
vaintroub, line 1295: I think we can live without those ifdefs, then we'll get some output for non-Linux OSes.

extra/mariabackup/xtrabackup.cc
```diff
@@ -378,6 +378,10 @@ extern const char *innodb_checksum_algorithm_names[];
 extern TYPELIB innodb_checksum_algorithm_typelib;
 extern const char *innodb_flush_method_names[];
 extern TYPELIB innodb_flush_method_typelib;
+#ifdef __linux__
+extern const char *innodb_linux_aio_names[];
+extern TYPELIB innodb_linux_aio_typelib;
```
vaintroub, line 383: Do we want backup to choose AIO implementation? Does really which AIO it uses, is it performance relevant? Can't we just use default one? Or even simulated one?

dr-m: It could be relevant for `mariadb-backup --prepare`. I’ll try to test a [MDEV-29911](https://jira.mariadb.org/browse/MDEV-29911) or [MDEV-34062](https://jira.mariadb.org/browse/MDEV-34062) like scenario to assess the impact.

tpool/tpool.h
dr-m: I can’t see how it could be moved inside `create_native_aio()`. Here the intention is to replace `m_aio` with an object where some attributes differ, but the type of the implementation is preserved. The function `create_native_aio()` may also be invoked for constructing the very first `thread_pool::m_aio`. This is the cleanest solution that I was able to come up with. Can you post a tested patch that corresponds to what you are suggesting?

storage/innobase/os/os0file.cc
vaintroub: Only on Linux, innodb_use_native_aio=ON *can* fail in the first place. You do not need to apply logic for other OSes at all, even if introduce it, it won't change anything .

tpool/tpool.h
vaintroub: Ok, will write and post, but this gonna need some time (since I'd need to spend time in an environment I'm not most productive in)

vaintroub: The tested patch, is 18bb45f4b1550d9fc8eaa91edb66e9d365de81af . 
This also fixes several other things I did I was not happy with in your patch, like complicated mazes made out of `_WIN32,__linux__,HAVE_URING,LINUX_NATIVE_AIO`, spread throughout Innodb code. also some by now necessary build cleanups are there.

vaintroub (changes requested):
The functionality needs a better encapsulation. 
That means - no randomly spread `_WIN32,__linux__, HAVE_URING, LINUX_NATIVE_AIO` spread throughout the Innodb code. If you want to know if tpool supports native aio.
there should be tpool::native_aio_supported function (constexpr or not), not an "educated guess", based on combination of ifdefs, which is also copied around into xtrabackup. There should not be a single HAVE_URING outside of the tpool.
 
Also, fallback from failed uring initialization to libaio, needs to be in tpool, not inside Innodb. 
(Maybe there should also be a fallback from failed native aio init, to simulated aio inside tpool as well, but this is for another day)

Please take a look at https://github.com/MariaDB/server/commit/18bb45f4b1550d9fc8eaa91edb66e9d365de81af , thats a patch upon yours, that fixes all that.

dr-m:
@vaintroub, thank you for the suggestions–much appreciated. 038966697038b01992cb872054fc31c11c829cd6 includes them. Please review.

vaintroub (approved):
Looks good to me.

PR #3978, opened 2025-04-15
MDEV-9343 Copying from YEAR to DATE result in '0000-00-00'

HazemRawi:
When converting YEAR values to DATE/DATETIME, explicitly set month and day to zero (e.g., 2000 → 2000-00-00). Validate against SQL modes:
- Allow `2000-00-00` if `NO_ZERO_IN_DATE` is disabled
- Reject if `NO_ZERO_IN_DATE` is enabled

- [x] *The Jira issue number for this PR is: MDEV-9343*

## Description
This fix explicitly targets `YEAR` to `DATE` or `DATETIME` conversions (setting `month=0`, `day=0`).
Queries involving `YEAR()`, `MONTH()`, or `DAY()` functions will change its behavior because they rely on the same conversion logic (`number_to_datetime_or_date`, `str_to_datetime_or_date_body`) used for date parsing.

## Release Notes
The server now preserves the year value when converting `YEAR` to `DATE` or `DATETIME`, resulting in `YYYY-00-00`.

## How can this PR be tested?

   Run `./mtr main.type_date`
  

## PR quality check


HazemRawi:
Ping 👀

svoj:
@HazemRawi Hi! This pull request was assigned to @abarkov for review. [MDEV-9434](https://jira.mariadb.org/browse/MDEV-9343) received Critical priority, which means it is constantly on radar.

In the meantime, please rebase on top of fresh 10.11, it contains fix for massive `main.timezone` failure. Also please check buildbot status, there failures that are very likely caused by this patch.

svoj (changes requested):
As requested in the comment: rebase and check bb failures.

ottok:
@HazemRawi If you want to have this re-reviewed, press on the arrow thing in the reviewer list to request re-review, or leave a comment saying this is ready for re-review.

image

mysql-test/main/temporal_literal.result
```diff
@@ -10,7 +10,8 @@ ERROR HY000: Incorrect DATE value: '01'
 SELECT DATE'01-01';
 ERROR HY000: Incorrect DATE value: '01-01'
 SELECT DATE'2001';
-ERROR HY000: Incorrect DATE value: '2001'
+DATE'2001'
```
gkodinov: This is a sign of an functional issue with the proposed change IMHO: DATE literals are defined to be either 'YYYY-MM-DD' or 'YY-MM-DD' in https://mariadb.com/docs/server/reference/sql-structure/sql-language-structure/date-and-time-literals. 
This is because you consider '2001' a valid date literal now. And it's not.

mysql-test/main/temporal_literal.result
```diff
@@ -10,7 +10,8 @@ ERROR HY000: Incorrect DATE value: '01'
 SELECT DATE'01-01';
 ERROR HY000: Incorrect DATE value: '01-01'
 SELECT DATE'2001';
-ERROR HY000: Incorrect DATE value: '2001'
```
gkodinov, line 13: This is IMHO a functional issue with the proposed fix: You consider DATE'2001' a valid literal now. And it's not a valid one, at least according to https://mariadb.com/docs/server/reference/sql-structure/sql-language-structure/date-and-time-literals, that says:

> A DATE string is a string in one of the following formats: 'YYYY-MM-DD' or 'YY-MM-DD'. 

..

> A DATE literal can also be an integer, in one of the following formats: YYYYMMDD or YYMMDD.

I'm out of my depth here, but it looks like every string passed to any of these literals is more or less parsed by the same internal function and then converted. This seems wrong: there should be some constraints to what is possible.

mysql-test/main/func_extract.result
```diff
@@ -425,9 +425,7 @@ Warnings:
 Warning	1292	Incorrect interval value: '100000-02-03'
 SELECT EXTRACT(DAY FROM '1111');
 EXTRACT(DAY FROM '1111')
-NULL
```
gkodinov, line 428: See my other comment: This is a demonstration how every literal (prefixed or not) is going to that same string parsing functionality. In this particular case there's no defined constraint on what string might be that is getting converted to a date.

mysql-test/main/temporal_literal.test
```diff
@@ -14,7 +14,6 @@ SELECT DATE'xxxx';
 SELECT DATE'01';
 --error ER_WRONG_VALUE
 SELECT DATE'01-01';
---error ER_WRONG_VALUE
```
gkodinov, line 17: ditto here: 2001 is not a valid DATE literal according to the docs.

mysql-test/main/type_date.result
```diff
@@ -123,7 +123,7 @@ select @d:=1111;
 1111
 select year(@d), month(@d), day(@d), cast(@d as date);
 year(@d)	month(@d)	day(@d)	cast(@d as date)
-2000	11	11	2000-11-11
```
gkodinov, line 126: My best guess here is that both the old and the new recordings are not correct according to the docs. I believe we should have had a error here.

mysql-test/main/type_date.result
```diff
@@ -135,12 +135,7 @@ select @d:=1311;
 1311
 select year(@d), month(@d), day(@d), cast(@d as date);
 year(@d)	month(@d)	day(@d)	cast(@d as date)
-NULL	NULL	NULL	NULL
-Warnings:
-Warning	1292	Incorrect datetime value: '1311'
-Warning	1292	Incorrect datetime value: '1311'
-Warning	1292	Incorrect datetime value: '1311'
-Warning	1292	Incorrect datetime value: '1311'
```
gkodinov, line 143: This here was returning an error. But it is not doing that now. See my comment for the change above: both should return an error IMHO.

mysql-test/main/type_date.result
```diff
@@ -934,8 +929,7 @@ EXPLAIN EXTENDED SELECT * FROM t1 WHERE 1995.0000000 BETWEEN f AND '2012-12-12';
 id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
 1	SIMPLE	t1	range	f	f	4	NULL	1	100.00	Using where; Using index
 Warnings:
-Warning	1292	Truncated incorrect datetime value: '1995.0000000'
-Note	1003	select `test`.`t1`.`f` AS `f` from `test`.`t1` where '0000-00-00' between `test`.`t1`.`f` and <cache>('2012-12-12')
+Note	1003	select `test`.`t1`.`f` AS `f` from `test`.`t1` where '1995-00-00' between `test`.`t1`.`f` and <cache>('2012-12-12')
```
gkodinov, line 932: This is IMHO also a functional degradation: here it's converting a decimal (probably) first to a string and then to a date. And now, with the change proposed, "1995.000000" is considered a valid date literal. And it isn't.

mysql-test/main/type_date.result
```diff
@@ -1089,14 +1083,14 @@ SET sql_mode=DEFAULT;
 #
 SELECT NULLIF(CAST(1012.5 AS DATE), 1);
 NULLIF(CAST(1012.5 AS DATE), 1)
-2000-10-12
```
gkodinov, line 1092: this should have produced a conversion warning IMHO. "1012.5" is not a valid string literal either as an integer or as a string.

mysql-test/main/type_datetime.result
```diff
@@ -1134,12 +1134,12 @@ CREATE TABLE t1 (a TIME);
 INSERT INTO t1 VALUES ('00:00:00'),('00:01:00');
 SELECT 1 FROM t1 WHERE 2016 > SOME (SELECT CAST(a AS DATETIME) FROM t1);
 1
-Warnings:
```
gkodinov, line 1137: This should have kept the warning IMHO: 2016 is not a valid date time literal.

mysql-test/main/type_enum.result
```diff
@@ -2399,16 +2399,6 @@ SELECT * FROM t1 WHERE indexed_col=DATE'2001-01-01';
 indexed_col
 Warnings:
 Note	1105	Cannot use key `indexed_col` part[0] for lookup: `test`.`t1`.`indexed_col` of type `enum` = "DATE'2001-01-01'" of type `date`
-Warning	1292	Truncated incorrect datetime value: '2001'
```
gkodinov, line 2402: the warning should have stayed: the enum options are not valid date time literals IMHO.

gkodinov:
After some internal discussions I believe the right approach consists of:

- Field_year should override save_in_field().
- Type_handler_year should override Item_save_in_field().
- Both of the above should try to downcast their target to Type_handler_temporal
- if it's castable then both should initialize a MYSQL_TIME structure, e.g. tm.year=year_value; and use target_field->store_store_time_dec(&tm)
- Issue a note message if any of the date/time source parts are non-0.

Also, it'd be prudent to retarget towards main. It's a bit too risky for 10.x. 

The above approach won't cover YEAR value -> DATETIME value conversions. But a similar (reverse) approach could be devised for these too as a separate task.

gkodinov:
It looks like there's no answer to my review for a month. Closing the PR for now. Please re-open if you plan on working on it again.

PR #3982, opened 2025-04-16
MDEV-36613 Incorrect undo logging for indexes on virtual columns

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36613*
## Description
Starting with mysql/mysql-server@02f8eaa9988dadb73dd68630dd82393cfa19bfb8 and 2e814d4702d71a04388386a9f591d14a35980bfe the index ID of indexes on virtual columns was being encoded insufficiently in InnoDB undo log records.  Only the least significant 32 bits were being written.  This could lead to some corruption of the affected indexes on `ROLLBACK`, as well as to missed chances to remove some history from such indexes when purging the history of committed transactions that included `DELETE` or an `UPDATE` in the indexes.

`dict_hdr_create()`: In debug instrumented builds, initialize the `DICT_HDR_INDEX_ID` close to the 32-bit barrier, instead of initializing it to `DICT_HDR_FIRST_ID` (10).  This will allow the changed code to be exercised while running the following:
```sh
./mtr --suite=gcol,vcol
```

`trx_undo_log_v_idx()`: Encode large `index->id` in a similar way as mysql/mysql-server@e00328b4d068c7485ac2ffe27207ed1f462c718d but using a different implementation.

`trx_undo_read_v_idx_low()`: Decode large `index->id` in a similar way as `mach_u64_read_much_compressed()`.
## Release Notes
Indexes on `VIRTUAL` columns with `INFORMATION_SCHEMA.INNODB_SYS_INDEXES.ID>4294967295` could become corrupted due to the incorrect way of writing undo log records.
## How can this PR be tested?
```sh
./mtr --suite=gcol,vcol
```
## PR quality check

### DIFF
--- a/storage/innobase/trx/trx0rec.cc
+++ b/storage/innobase/trx/trx0rec.cc
@@ -148,7 +148,9 @@ trx_undo_log_v_idx(
 	ulint n_idx = 0;
 	for (const auto& v_index : vcol->v_indexes) {
 		n_idx++;
-		/* FIXME: index->id is 64 bits! */
+		if (uint32_t hi= uint32_t(v_index.index->id >> 32)) {
+			size += 1 + mach_get_compressed_size(hi);
+		}
 		size += mach_get_compressed_size(uint32_t(v_index.index->id));
 		size += mach_get_compressed_size(v_index.nth_field);
 	}
@@ -175,10 +177,14 @@ trx_undo_log_v_idx(
 	ptr += mach_write_compressed(ptr, n_idx);
 
 	for (const auto& v_index : vcol->v_indexes) {
-		ptr += mach_write_compressed(
-			/* FIXME: index->id is 64 bits! */
-			ptr, uint32_t(v_index.index->id));
-
+		/* This is compatible with
+		ptr += mach_u64_write_much_compressed(ptr, v_index.index-id)
+		(the added "if" statement is fixing an old regression). */
+		if (uint32_t hi= uint32_t(v_index.index->id >> 32)) {
+			*ptr++ = 0xff;
+			ptr += mach_write_compressed(ptr, hi);
+		}
+		ptr += mach_write_compressed(ptr, uint32_t(v_index.index->id));
 		ptr += mach_write_compressed(ptr, v_index.nth_field);
 	}
 
@@ -217,7 +223,15 @@ trx_undo_read_v_idx_low(
 	dict_index_t*	clust_index = dict_table_get_first_index(table);
 
 	for (ulint i = 0; i < num_idx; i++) {
-		index_id_t	id = mach_read_next_compressed(&ptr);
+		index_id_t	id = 0;
+		/* This is like mach_u64_read_much_compressed(),
+		but advancing ptr to the next field. */
+		if (*ptr == 0xff) {
+			ptr++;
+			id = mach_read_next_compressed(&ptr);
+			id <<= 32;
+		}
+		id |= mach_read_next_compressed(&ptr);
 		ulint		pos = mach_read_next_compressed(&ptr);
 		dict_index_t*	index = dict_table_get_next_index(clust_index);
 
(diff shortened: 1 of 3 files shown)


storage/innobase/dict/dict0boot.cc
```diff
@@ -148,7 +148,7 @@ dict_hdr_create(
 	mtr->write<8>(*d, DICT_HDR + DICT_HDR_TABLE_ID + d->frame,
 		      DICT_HDR_FIRST_ID);
 	mtr->write<8>(*d, DICT_HDR + DICT_HDR_INDEX_ID + d->frame,
-		      DICT_HDR_FIRST_ID);
+		      IF_DBUG(0xffffff00, DICT_HDR_FIRST_ID));
```
mariadb-DebarunBanerjee, line 152: May be better to have DBUG_EXECUTE_IF("simulate_index_id_exceed_uint32") and drive it via debug tests ? Otherwise, we would no longer be testing the majority scenario when index_id is less than 32 BIT max.

storage/innobase/trx/trx0rec.cc
```diff
@@ -175,10 +177,11 @@ trx_undo_log_v_idx(
 	ptr += mach_write_compressed(ptr, n_idx);
 
 	for (const auto& v_index : vcol->v_indexes) {
-		ptr += mach_write_compressed(
-			/* FIXME: index->id is 64 bits! */
-			ptr, uint32_t(v_index.index->id));
-
+		if (uint32_t hi= uint32_t(v_index.index->id >> 32)) {
+			*ptr++ = 0xff;
+			ptr += mach_write_compressed(ptr, hi);
```
mariadb-DebarunBanerjee, line 182: If we generate undo log this way, it is certain to fail on a downgrade. I understand for such index ID the previous version is buggy but this could possibly cause unpredictable behaviour (possibly corruption) on downgrade since there is nothing other than debug check while reading back from the file.

Can we show an warning here (not repeating) that one should do a slow shutdown before any attempt for downgrade ?

storage/innobase/trx/trx0rec.cc
```diff
@@ -217,7 +220,13 @@ trx_undo_read_v_idx_low(
 	dict_index_t*	clust_index = dict_table_get_first_index(table);
 
 	for (ulint i = 0; i < num_idx; i++) {
-		index_id_t	id = mach_read_next_compressed(&ptr);
+		index_id_t	id = 0;
+		if (*ptr == 0xff) {
```
mariadb-DebarunBanerjee, line 224: The 0xFF works as it is not conflicting with constant values (0x80, 0xC0 ... 0xF0) the `mach_read_next_compressed` for first byte. May be a comment here and/or in `mach_read_next_compressed` is worth.

dr-m: I could also add source code comments that this encoding is compatible with `mach_u64_write_much_compressed()`.

storage/innobase/dict/dict0boot.cc
dr-m: I was really too lazy about this. 43882731421d6e983a0973ea6d605ad7ab9b99db is now injecting the precondition in a non-debug test using some Perl code. This change should also prevent failures of tests that are looking for some exact `SYS_INDEXES.ID` values after bootstrap.

storage/innobase/trx/trx0rec.cc
dr-m: 1c9f64e54ffb109bb6cf6a189e863bfa54e46510 includes a test that would reproduce a crash on 10.5 when the code fixes are not present. If we are downgrading to an old version while using larger `SYS_INDEXES.ID` for indexes on virtual columns, we should be in a crash-prone territory anyway. Therefore, I don’t think that we’d really need to issue any warning to the error log here. Downgrades should be extremely rare anyway.

mariadb-DebarunBanerjee (approved):

PR #3983, opened 2025-04-17
MDEV-36182 revert incorrect 5.14 kernel warnings and correct liburing interface usage

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36182*

## Description

The kernel warnings adding in MDEV-26674 where at the very early stages of kernel development. Since then RHEL9 and its clones and compatible distros have stabilised on the 5.14 kernel which no longer has broken behaviour as the initial kernels but users are still receiving warnings and disabling of innodb_native_aio.

Because of this, its safe to revert the kernel checks. We've checked that the kernel of RHEL9 works correctly with io_uring. During this it was discovered that the default RHEL9 kernels will have sysctl kernel.io_uring_disabled=2 set which disable io_uring usage for all users. This is either out of compatibility or from the early days where there where more io_uring bug reports. Either way, we don't need to provide another slightly broken layer of control around this, the end user has the option (via sysctl) and reverting this change isn't going to do anything except remove a false warning in the MariaDB logs.

We also correct a bug in our liburing usage. It is the return value of `io_uring_queue_init` that has the error number and not the `errno` variable. The obvious case this was presented was with the EPERM error, which meant users where getting a warning "mariadbd: io_uring_queue_init() failed with errno 0" which is very unintuitive. So correcting the usage of this function and adding EPERM description as it applies to MariaDB is a necessary change for those warning to use liburing on RHEL kernels.

The revert commits are unmodified in content from what git-revert produced with no conflicts.

Removed an used variable from tpool because it was on the list of things to clean-up and it is as risk free as it gets.

## Release Notes

Remove out of date warning about the 5.14 kernel about io_uring bugs, and produce useful error messages if the io_uring_disabled is set and leave the rest to user choice.

## How can this PR be tested?

```
$ sudo sysctl kernel.io_uring_disabled=2

and start MariaDB
2025-04-17 15:51:26 0 [Warning] mysqld: io_uring_queue_init() failed with EPERM: sysctl kernel.io_uring_disabled has the value 2, or 1 and the user of the process is not a member of sysctl kernel.io_uring_group. (see man 2 io_uring_setup).
2025-04-17 15:51:26 0 [Warning] InnoDB: liburing disabled: falling back to innodb_use_native_aio=OFF

$ sudo sysctl kernel.io_uring_disabled=0

...
2025-04-17 15:52:16 0 [Note] InnoDB: Using liburing
```

no incorrect warnings, useful descriptions, and with disable=0, a functioning working MariaDB

Tick one of the following boxes [x] to help us understand if the base branch for the PR is correct.
-->

## PR quality check

### DIFF
--- a/extra/mariabackup/xtrabackup.cc
+++ b/extra/mariabackup/xtrabackup.cc
@@ -1581,11 +1581,6 @@ uint xb_client_options_count = array_elements(xb_client_options);
 static const char *dbug_option;
 #endif
 
-#ifdef HAVE_URING
-extern const char *io_uring_may_be_unsafe;
-bool innodb_use_native_aio_default();
-#endif
-
 struct my_option xb_server_options[] =
 {
   {"datadir", 'h', "Path to the database root.", (G_PTR*) &mysql_data_home,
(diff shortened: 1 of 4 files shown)


dr-m (approved):

grooverdan:
Thanks @dr-m and @vaintroub

PR #3989, opened 2025-04-21
MDEV-36649 dict_acquire_mdl_shared() aborts when table mode is DICT_TABLE_OP_OPEN_ONLY_IF_CACHED

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36649*

## Description
- InnoDB fails to check the table is being dropped or evicted while acquiring the MDL for the table when table open operation mode is DICT_TABLE_OP_OPEN_ONLY_IF_CACHED.

Fix:
===
dict_acquire_mdl_shared(): Check whether the table name length before retrying to acquire MDL for the table. In case of evicted or dropped table, new table length is 0.

## How can this PR be tested?
./mtr innodb.stats_persistent

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/stats_persistent.test
+++ b/mysql-test/suite/innodb/t/stats_persistent.test
@@ -1,4 +1,5 @@
 --source include/have_innodb.inc
+--source include/have_sequence.inc
 --source include/have_debug.inc
 --source include/have_debug_sync.inc
 --source include/count_sessions.inc
@@ -26,3 +27,14 @@ SET DEBUG_SYNC= 'RESET';
 DROP TABLE t1;
 
 --source include/wait_until_count_sessions.inc
+
+--echo #
+--echo # MDEV-36649 dict_acquire_mdl_shared() aborts when table
+--echo #      mode is DICT_TABLE_OP_OPEN_ONLY_IF_CACHED
+--echo #
+set @old_defragment_stats_accuracy= @@innodb_defragment_stats_accuracy;
+SET GLOBAL innodb_defragment_stats_accuracy=1;
+CREATE TABLE t (a INT ) ENGINE=INNODB;
+INSERT INTO t SELECT * FROM seq_1_to_1000;
+DROP TABLE t;
+set global innodb_defragment_stats_accuracy= @old_defragment_stats_accuracy;
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -795,6 +795,8 @@ lookup:
       dict_sys.freeze(SRW_LOCK_CALL);
     goto return_without_mdl;
   }
+  else
+    goto return_without_mdl;
 
   if (*mdl)
   {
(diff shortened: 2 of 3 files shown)


storage/innobase/dict/dict0dict.cc
```diff
@@ -816,6 +816,11 @@ dict_acquire_mdl_shared(dict_table_t *table,
   db_len= db1_len;
   tbl_len= tbl1_len;
 
+  /* Table could be dropped or evicted when table open operation mode is
+  DICT_TABLE_OP_OPEN_ONLY_IF_CACHED */
+  if (tbl_len == 0)
+    goto return_without_mdl;
```
dr-m, line 822: Did I misunderstand something, or could `tbl1_len` and therefore also `tbl_len` be uninitialized here? Could we also add debug assertions before the `goto`?
```c++
ut_ad(!table);
ut_ad(table_op == DICT_TABLE_OP_OPEN_ONLY_IF_CACHED);
```
In this way, we would not even need a source code comment.

If my understanding is correct, the following fix would seem to work. Can you please check that?
```diff
diff --git a/storage/innobase/dict/dict0dict.cc b/storage/innobase/dict/dict0dict.cc
index 4f293c41854..ef4c24f6fb5 100644
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -795,6 +795,8 @@ dict_acquire_mdl_shared(dict_table_t *table,
       dict_sys.freeze(SRW_LOCK_CALL);
     goto return_without_mdl;
   }
+  else
+    goto return_without_mdl;
 
   if (*mdl)
   {
```

mysql-test/suite/innodb/t/stats_persistent.test
```diff
@@ -26,3 +27,13 @@ SET DEBUG_SYNC= 'RESET';
...
+
+--echo #
+--echo # MDEV-36649 dict_acquire_mdl_shared() aborts when table
+--echo #      mode is DICT_TABLE_OP_OPEN_ONLY_IF_CACHED
+--echo #
+SET GLOBAL innodb_defragment_stats_accuracy=1;
+CREATE TABLE t (a INT ) ENGINE=INNODB;
+INSERT INTO t SELECT * FROM seq_1_to_1000;
+DROP TABLE t;
+SET GLOBAL innodb_defragment_stats_accuracy=default;
```
dr-m, line 39: I think that we should assign the original parameter to a `@save` variable and restore from it.

storage/innobase/dict/dict0dict.cc
```diff
@@ -795,6 +795,7 @@ dict_acquire_mdl_shared(dict_table_t *table,
       dict_sys.freeze(SRW_LOCK_CALL);
     goto return_without_mdl;
   }
+  else goto return_without_mdl;
```
dr-m, line 798: Usually the body of `if` or `else` is on a separate line.

dr-m (approved):

PR #3993, opened 2025-04-21
MDEV-35816 ASAN use-after-poison in st_select_lex::print

DaveGosselin-MariaDB:
Semi-joins with CTEs as the derived table will leave dangling pointers to the JOIN instance, from the corresponding SELECT_LEX instance, after first execution has completed.  Clean these up during statement reinitialization, before subsequent executions occur.


DaveGosselin-MariaDB:
Closing this in favor of https://github.com/MariaDB/server/pull/4306

PR #4009, opened 2025-04-24
MDEV-36684 - main.mdl_sync fails under valgrind (test for Bug#42643)

svoj:
Valgrind is single threaded and only changes threads as part of system calls or waits.

I found some busy loops where the server assumes that some other thread will change the state, which will not happen with valgrind.

Added VALGRIND_YIELD to the loops, which calls pthread_yield() if HAVE_VALGRIND is defined.

Added pthread_yield() to the loops in table_cache.

We should consider changing some of the VALGRIND_YIELD calls to call pthread_yield() as busy loop without any sleep() is usually a bad thing.

Reviewer: svojtovich@gmail.com

### DIFF
--- a/sql/sql_base.cc
+++ b/sql/sql_base.cc
@@ -4460,6 +4460,7 @@ restart:
             goto error;
 
           error= FALSE;
+          std::this_thread::yield();
           goto restart;
         }
         goto error;
@@ -4522,6 +4523,7 @@ restart:
               goto error;
 
             error= FALSE;
+            std::this_thread::yield();
             goto restart;
           }
           /*
--- a/sql/table_cache.cc
+++ b/sql/table_cache.cc
@@ -901,6 +901,7 @@ retry:
   {
     mysql_mutex_unlock(&element->LOCK_table_share);
     lf_hash_search_unpin(thd->tdc_hash_pins);
+    std::this_thread::yield();
     goto retry;
   }
   lf_hash_search_unpin(thd->tdc_hash_pins);
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -1076,6 +1076,7 @@ static void trx_purge_close_tables(purge_node_t *node, THD *thd) noexcept
 
 void purge_sys_t::wait_FTS(bool also_sys)
 {
+  std::this_thread::yield();
   for (const uint32_t mask= also_sys ? ~0U : ~PAUSED_SYS; m_FTS_paused & mask;)
     std::this_thread::sleep_for(std::chrono::milliseconds(10));
 }
--- a/sql/sql_class.h
+++ b/sql/sql_class.h
@@ -20,6 +20,7 @@
 /* Classes in mysql */
 
 #include <atomic>
+#include <thread>
 #include "dur_prop.h"
 #include <waiting_threads.h>
 #include "sql_const.h"


storage/innobase/trx/trx0purge.cc
```diff
@@ -1189,7 +1189,10 @@ dict_table_t *purge_sys_t::close_and_reopen(table_id_t id, THD *thd,
 
   dict_table_t *table= trx_purge_table_open(id, mdl_context, mdl);
   if (table == reinterpret_cast<dict_table_t*>(-1))
+  {
+    VALGRIND_YIELD;
     goto retry;
+  }
```
dr-m, line 1195: We could avoid some code duplication by introducing another label here and replacing the subsequent `goto retry` jump here. But then again, this only affects Valgrind builds.

dr-m (approved):
I’d use `std::this_thread::yield()` in C++ code. But I don’t think it will make much difference, other than removing a dependency on POSIX or our POSIX simulation layer (`#define pthread_yield()`).

svoj:
> I’d use `std::this_thread::yield()` in C++ code. But I don’t think it will make much difference, other than removing a dependency on POSIX or our POSIX simulation layer (`#define pthread_yield()`).

Yes, I also suggested `std::this_thread::yield()` when reviewing it, same thinking.

dr-m (approved):

svoj:
@dr-m retargeted for `10.6`, moved to `std::this_thread::yield()`. Please merge if you agree.

sql/sql_base.cc
```diff
@@ -4522,6 +4523,7 @@ bool open_tables(THD *thd, const DDL_options_st &options,
               goto error;
 
             error= FALSE;
+            std::this_thread::yield();
             goto restart;
```
dr-m, line 4527: This is duplicating the section of code 63 lines earlier. In fact, a wider section of the code block starting with `if (unlikely(error))` is being duplicated. I think that a follow-up fix would be useful where we’d do the following:
```c++
  if (unlikely(error))
  {
    if (!(error= open_tables_handle_error(error, …)))
      goto restart;
    goto error;
  }
```
and there would be a new function `ATTRIBUTE_COLD static bool open_tables_error_handler(bool, …)` that would include the duplicated code, including the call to `std::this_thread::yield()`.

sql/table_cache.cc
```diff
@@ -901,6 +901,7 @@ TABLE_SHARE *tdc_acquire_share(THD *thd, TABLE_LIST *tl, uint flags,
   {
     mysql_mutex_unlock(&element->LOCK_table_share);
     lf_hash_search_unpin(thd->tdc_hash_pins);
+    std::this_thread::yield();
     goto retry;
   }
   lf_hash_search_unpin(thd->tdc_hash_pins);
```
dr-m, line 907: Would the following be a correct optimization?
```c++
  mysql_mutex_lock(&element->LOCK_table_share);
  lf_hash_search_unpin(thd->tdc_hash_pins);
  if (!(share= element->share))
  {
    mysql_mutex_unlock(&element->LOCK_table_share);
    std::this_thread::yield();
    goto retry;
  }
```

storage/innobase/trx/trx0purge.cc
```diff
@@ -1176,6 +1176,7 @@ dict_table_t *purge_sys_t::close_and_reopen(table_id_t id, THD *thd,
   MDL_context *mdl_context= static_cast<MDL_context*>(thd_mdl_context(thd));
   ut_ad(mdl_context);
  retry:
+  std::this_thread::yield();
```
dr-m, line 1179: If we are going to insert this yield unconditionally (outside Valgrind builds), I think that the correct place for it would be right after closing all tables:
```c++
  m_active= false;
  wait_FTS(false);
  m_active= true;
```
In fact, maybe the following should be rewritten to be Valgrind friendly:
```c++
void purge_sys_t::wait_FTS(bool also_sys)
{
  for (const uint32_t mask= also_sys ? ~0U : ~PAUSED_SYS; m_FTS_paused & mask;)
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
}
```
Maybe we should just add a call to `std::this_thread::yield()` at the beginning of this function? Because `purge_sys.m_FTS_paused` is being modified by atomic operations, I think that we need an actual sleep in the loop; invoking only a yield would likely result in excessive CPU consumption.

dr-m (commented):
I think that we need a better solution for `purge_sys_t::close_and_reopen()` where the thread would yield while not holding any metadata locks or table handles. My other comments are suggestions of follow-up improvements, not directly related to improving the Valgrind experience.

sql/table_cache.cc
svoj: Well, depending on what you're trying to achieve. `LOCK_table_share` is a per-share mutex, it is not locked for table DML, but still can become contested for views. The less we have under its protection the better. So, this will make things slightly worse in this regard, and that was the main reason to keep it out of `LOCK_table_share`.

Speaking of correctness: `lf_hash_search_unpin()` marks `element` unused. In current `LF_HASH` implementation it means `element` is available for reuse. Taking it into account as well as considering how table definition cache is using it, nothing bad is going to happen. But I'd say it is like relying on undefined behaviour. Once it is unpinned it becomes property of `LF_HASH`.  If it happens *before* `element->share` check, in theory we can expect `LF_HASH` to free/corrupt its memory. So it is not good from correctness point of view either.

What we get if we do `lf_hash_search_unpin()` earlier? It means we mark `element` for reuse earlier, that is another thread may reuse it rather than allocating it anew.

storage/innobase/trx/trx0purge.cc
svoj: That's a good point. Under complex locking scenarios, like:
```
purge: locks t2
con1: holds MDL_EXCLUSIVE on t1, waiting for MDL_EXCLUSIVE on t2
purge: locks t1, deadlocks (t1 lock not acquired), yields
con1: can't proceed as t2 is still locked
purge: locks t1, deadlocks (t1 lock not acquired), yields
...
```

In this case yielding before `trx_purge_close_tables()` won't fix the problem we're trying to solve.

I don't feel like `wait_FTS()` is the right name for this method, but I agree that `yield` should be moved there. Will get it fixed.

sql/sql_base.cc
svoj: Right. There was a change in 10.11 though, that makes it different by one line at least. I wouldn't dare doing it in GA versions though.

svoj:
@dr-m PR updated. I can see Monty pushed his aae9b50a53463d1ebfbf7b22cd78eda097536b46 to `main`, so you'd have to cleanup `VALGRIND_YIELD` once this one is merged up to `main`.

sql/table_cache.cc
dr-m: What we would gain is avoiding some code duplication and instruction cache bloat. Another variant would avoid the potential race condition:
```c++
  mysql_mutex_lock(&element->LOCK_table_share);
  share= element->share;
  lf_hash_search_unpin(thd->tdc_hash_pins);
  if (!share)
  {
    mysql_mutex_unlock(&element->LOCK_table_share);
    std::this_thread::yield();
    goto retry;
  }
```

dr-m (approved):
This will need to be rebased once more, because a failing builder was made mandatory for branch protection.

svoj:
There is no fix for this test case in 10.6, I added a separate commit that fixes it to this PR.

sql/table_cache.cc
svoj: It won't work either: `mysql_mutex_unlock(&element->LOCK_table_share)` would access potentially corrupt memory.

Presence of `element->share` guarantees there is no concurrent delete ongoing. Or if there is, it is waiting on `element->LOCK_table_share` mutex and thus it is safe to unpin it immediately after this block.

I understand your concern, I'd have implemented it as you suggested originally (back 10+ years ago), but I didn't see the way to arrange code better. Nor do I see it now.

svoj:
Removed test case fix, it is in 10.6 now.

PR #4010, opened 2025-04-24
MDEV-36461 Remove join_execution nodes from the trace, and instead add range-checked-for-each-record when needed

bsrikanth-mariadb:
Pull request created in: https://jira.mariadb.org/browse/MDEV-36461


sql/sql_select.cc
```diff
@@ -23934,10 +23929,16 @@ test_if_quick_select(JOIN_TAB *tab)
...
+  range_fer.add("loop", tab->join->explain->time_tracker.get_loops());
+
+  Json_writer_array rows_est(tab->join->thd, "rows_estimation");
   res= tab->select->test_quick_select(tab->join->thd, tab->keys,
-                                      (table_map) 0, HA_POS_ERROR, 0,
-                                      FALSE, /*remove where parts*/FALSE,
-                                      FALSE,
+                                      (table_map) 0, HA_POS_ERROR, 0, FALSE,
+                                      /*remove where parts*/ FALSE, FALSE,
                                       /* no unusable key notes */
```
spetrunia, line 23942: Is this a meaningful change or just re-formatting?  I don't think this changes anything, please restore the original formatting. That way, your patch won't show up when somebody will look at who was the last to change test_quick_select() call parameters.

mysql-test/main/opt_trace.test
```diff
@@ -1195,3 +1195,42 @@ set optimizer_trace='enabled=off';
...
+
+select * 
+from 
+  t1, t2
+where
+  t2.a<t1.a and t2.b<t1.b and t1.a+t2.a > 2000;
+
+--echo # The trace must contain 3 objects:
+select json_detailed(json_extract(trace, '$**.range-checked-for-each-record'))
+    from information_schema.optimizer_trace;
```
spetrunia, line 1217: So does it contain them?

mysql-test/main/opt_trace.test
```diff
@@ -1195,3 +1195,42 @@ set optimizer_trace='enabled=off';
...
+--echo # The trace must be empty:
+select json_detailed(json_extract(trace, '$**.join_execution'))
+    from information_schema.optimizer_trace;
+
+select *
```
spetrunia, line 1222: Is it a repetition of the same test as above? If yes please remove.

mysql-test/main/opt_trace.test
spetrunia: Please make sure that there is a test case showing that indeed `range-checked-for-each-record` is emitted for every execution with different `loop` number, and it has `range_analysis` inside it.  At the same time, please make .result file not contain irrelevant parts of the trace.

spetrunia (changes requested):
Please address input.

sql/sql_select.cc
bsrikanth-mariadb: unintentional change. will restore.

mysql-test/main/opt_trace.test
bsrikanth-mariadb: not a repetition, slightly different test with filters t2.a+1<t1.a+1 and t2.b+1<t1.b+1. So, range_check doesn't happen here. And so, neither range_check not join_execution objects should be present

mysql-test/main/opt_trace.test
bsrikanth-mariadb: sure, will have to add a new test for this case.

bsrikanth-mariadb: > So does it contain them?

yes

spetrunia (changes requested):
Please take a look at this commit: 
https://github.com/MariaDB/server/pull/4010/commits/6e917acf4cfb6ab0ebe3eb9c1f7ae474810083a6

You didn't have DROP TABLE in your tests, which produced errors, e.g.
https://ci.appveyor.com/project/rasmushoj/server/builds/51947118

I've also made the testcase smaller.  The contents of "range_analysis" nodes is stripped now.  also changed SELECT to ANALYZE to highlight the query plan used .

Please take a look and squash into your commit.

bsrikanth-mariadb:
> Please take a look at this commit: [6e917ac](https://github.com/MariaDB/server/commit/6e917acf4cfb6ab0ebe3eb9c1f7ae474810083a6)
> 
> You didn't have DROP TABLE in your tests, which produced errors, e.g. https://ci.appveyor.com/project/rasmushoj/server/builds/51947118
> 
> I've also made the testcase smaller. The contents of "range_analysis" nodes is stripped now. also changed SELECT to ANALYZE to highlight the query plan used .
> 
> Please take a look and squash into your commit.

got it. Thanks!

Also, squashed the commits

spetrunia (changes requested):
The commit comment

```
MDEV-36461:Remove join_execution nodes from the trace
```

Does not describe all what the patch does. Just "trace" can be confused with "debug trace".

Please change into something like (feel free to edit):

```
MDEV-36461: Optimizer trace: remove join_execution node

In non-EXPLAIN queries with subqueries, the trace was flooded 
with empty "join_execution":{} nodes. Now, they are gone. 

The "Range checked for each record" optimization still prints content
into trace on join execution. Now, we wrap it into 
"range-checked-for-each-record" to delimit the invocations.
```
and use similar level of detail for other patches.

bsrikanth-mariadb:
> The commit comment
> 
> ```
> MDEV-36461:Remove join_execution nodes from the trace
> ```
> 
> Does not describe all what the patch does. Just "trace" can be confused with "debug trace".
> 
> Please change into something like (feel free to edit):
> 
> ```
> MDEV-36461: Optimizer trace: remove join_execution node
> 
> In non-EXPLAIN queries with subqueries, the trace was flooded 
> with empty "join_execution":{} nodes. Now, they are gone. 
> 
> The "Range checked for each record" optimization still prints content
> into trace on join execution. Now, we wrap it into 
> "range-checked-for-each-record" to delimit the invocations.
> ```
> 
> and use similar level of detail for other patches.

Sure, added description in the commit message.

spetrunia (approved):

PR #4012, opened 2025-04-24
MDEV-36397 Record change_user command in MTR output

FarihaIS:
## Description

MTR `.result` files currently do not contain output to indicate if a `change_user` command has been executed in the corresponding `.test` files.

Record `change_user` command in the following format in MTR output only if `disable_query_log` is set to false: `change_user <user>,<password>,<db>;`

## Release Notes

N/A

## How can this PR be tested?

Execute the main suite in mysql-test-run. This commit modifies tests in that suite.

For example, a `change_user root;` line in the `.test` file will generate a new `change_user root,,;` line in the `.result` file if no password or db is set. 

If a parameter (`user`, `db`, or `password`) is not specified, the current value for that parameter will be used in the output.

* [x] _This is a new feature and the PR is based against the latest MariaDB development branch._

## Copyright


client/mysqltest.cc
```diff
@@ -4655,6 +4655,24 @@ void do_change_user(struct st_command *command)
       dynstr_set(&ds_db, mysql->db);
   }
 
+  /* Connection logging if enabled */
+  if (!disable_result_log && !disable_connect_log && !disable_query_log)
```
vuvova, line 4659: this is getting too complex, let's just have `if (!disable_query_log)` here

client/mysqltest.cc
```diff
@@ -4655,6 +4655,24 @@ void do_change_user(struct st_command *command)
...
+    replace_dynstr_append(ds, ds_user.str);
+    dynstr_append_mem(ds, ",", 1);
+
+    if (ds_passwd.length)
+      replace_dynstr_append(ds, ds_passwd.str);
+    dynstr_append_mem(ds, ",", 1);
+
+    if (ds_db.length)
+      replace_dynstr_append(ds, ds_db.str);
+    dynstr_append_mem(ds, ";\n", 2);
```
vuvova, line 4673: generally prefer this pattern:
```c
dynstr_append_mem(ds, STRING_WITH_LEN(";\n"));
```
when you need to specify a string and its length. Here and above too.

mysql-test/main/change_user.result
```diff
@@ -84,13 +95,15 @@ SELECT IS_USED_LOCK('bug31418') = CONNECTION_ID();
 IS_USED_LOCK('bug31418') = CONNECTION_ID()
 1
 change_user
```
vuvova, line 97: you can now remove these `--echo change_user` lines from the `change_user.test` file

vuvova (changes requested):
Good, thanks! Result files became much easier to read!

See few comments inline.

client/mysqltest.cc
FarihaIS: Thank you, I've changed my code to use `dynstr_append_mem(ds, STRING_WITH_LEN("..."));` where applicable

mysql-test/main/change_user.result
FarihaIS: Removed, thank you for catching that

client/mysqltest.cc
FarihaIS: Good point, I've changed it as per your suggestion. 

I did notice, however, that excluding `!disable_result_log` from the check causes `main.plugin_vars` to fail either locally or in the PR's CI/CD run.

Locally, I'm seeing a result length mismatch like so:
```
connection con2;
CALL p_show_vars(100);
connection default;
+change_user root,,test;
...
+change_user root,,test;
connection con1;
 ```
 
But after running the test with `--record`, in the CI/CD run, I'm seeing a result length mismatch like so:
```
connection con2;
CALL p_show_vars(100);
connection default;
+change_user root,,;
...
+change_user root,,;
connection con1;
 ```

vuvova: Note, that what fails is **debug-emb test** step. On the **amd64-debian-12-debug-embedded** builder. It runs the mtr with the `--embedded` switch, you can see it in the log. That is,
* run `cmake -DWITH_EMBEDDED_SERVER=1`
* `make` (or `ninja` or whatever you're using)
* run `./mtr --emb main.plugin_vars`
* figure out why it's different in embedded

grooverdan:
it looks like your main.plugin_vars test is still having a different result under embedded.

FarihaIS:
> it looks like your main.plugin_vars test is still having a different result under embedded.

I haven't pushed any changes to rectify the failure yet, I'm still working on the fix - thanks

FarihaIS:
I have made the necessary changes for `main.plugin_vars` to work in embedded mode. I believe the only test failing now might be a flaky one (`perfschema.lowercase_fs_off`). Please let me know if there is anything else I should change, thank you.

Edit: I just modified the commit message and pushed again and seems `perfschema.lowercase_fs_off` is not failing now.

libmysqld/lib_sql.cc
```diff
@@ -789,9 +789,22 @@ int check_embedded_connection(MYSQL *mysql, const char *db)
...
+
   /* Change database if necessary */
-  if (!(result= (db && db[0] && mysql_change_db(thd, &db_str, FALSE))))
+  result = 0;
+  if (db && db[0])
+  {
+    result = mysql_change_db(thd, &db_str, FALSE);
+    if (!result)
+    {
+      if (mysql->db)
```
vuvova: you don't need this `if()`. `free(NULL)` is well defined and does nothing, it's safe. So, simply write
```c
my_free(mysql->db);
mysql->db = my_strdup(PSI_NOT_INSTRUMENTED, db, MYF(0));
```

vuvova (approved):
looks good. Just one nitpick. Feel free to ignore it, if you don't fix by the release date we'll merge as is, no big deal.

libmysqld/lib_sql.cc
FarihaIS: Fixed, thank you for the advice!

PR #4013, opened 2025-04-24
MDEV-36666 - atomic.alter_table still times out often

svoj:
According to buildbot cross-reference atomic.alter_table is failing ~10 times a day due to 900 seconds test case timeout. Split it into two tests: atomic.alter_table_myisam and atomic.alter_table_innodb. It should reduce failure frequency down to once a day or so, similarly to atomic.alter_table_aria test.

Diabling InnoDB for MyISAM/Aria/RocksDB tests makes them 20-35% faster.


dr-m (approved):
Thank you, this looks OK to me. But, should we target the earliest applicable version (10.6) with this?

PR #4016, opened 2025-04-25
MDEV-28823 Secure mariadb-secure-installation output file with chmod

akshatnehra:
## Description

This commit addresses a security issue in the `mariadb-secure-installation` script where the temporary output file containing SQL commands and potentially password hashes was being created with default permissions (typically world-readable).

The fix involves modifying the `prepare()` function to:

1. Create the `$output` file explicitly using `touch` before it's used
2. Apply `chmod 600` permissions to restrict access to owner only
3. Maintain consistency with how `$config` and `$command` files are already handled

## How can this PR be tested?

1. Run the original `mariadb-secure-installation` script and observe file permissions:

   ```bash
   ./build/scripts/mariadb-secure-installation --socket=/tmp/mysql.sock
   # In another terminal
   ls -la .my* | grep -v .mysql_history
   ```

   The `.my.output.*` file will have `-rw-r--r--` permissions
2. With the patched version:

   ```bash
   # After applying the fix and rebuilding
   ./build/scripts/mariadb-secure-installation --socket=/tmp/mysql.sock
   # In another terminal 
   ls -la .my* | grep -v .mysql_history
   ```

   The `.my.output.*` file will have `-rw-------` permissions

## Results from my testing

1. Before changes

   ```
   root@03b5517f4303:/quick-rebuilds# ls -la .my* | grep -v .mysql_history
   -rw------- 1 root root  70 Apr 22 16:55 .my.cnf.15643
   -rw-r--r-- 1 root root 130 Apr 22 16:55 .my.output.15643
   -rw------- 1 root root  32 Apr 22 16:55 .mysql.15643
   ```
2. After Changes

   ```
   root@03b5517f4303:/quick-rebuilds# ls -la .my* | grep -v .mysql_history
   -rw------- 1 root root  70 Apr 22 17:04 .my.cnf.16290
   -rw------- 1 root root 130 Apr 22 17:04 .my.output.16290
   -rw------- 1 root root  32 Apr 22 17:04 .mysql.16290
   ```

* ✅ _This is a security fix applicable to multiple versions, and the PR is based against the latest MariaDB development branch._

## PR quality check

* ✅ I have checked the `CODING_STANDARDS.md` file and my PR conforms to this where appropriate.
* ✅ For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

### DIFF
--- a/scripts/mysql_secure_installation.sh
+++ b/scripts/mysql_secure_installation.sh
@@ -1,16 +1,16 @@
 #!/bin/sh
 
 # Copyright (c) 2002, 2016, Oracle and/or its affiliates. All rights reserved.
-# 
+#
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation; version 2 of the License.
-# 
+#
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
-# 
+#
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1335  USA
@@ -218,8 +218,8 @@ validate_reply () {
 }
 
 prepare() {
-    touch $config $command
-    chmod 600 $config $command
+    touch $config $command $output
+    chmod 600 $config $command $output
 }
 
 do_query() {


grooverdan:
Thanks for the fix. I'm thinking the original premise of touch/chmod is incorrect. There there's a very small race condition where the file can be opened, before the chmod is applied, and then read when its its later populated.

Probably best fixed with `umask` and omit `chmod`. Does that sound better? As security sensitive fix, a rebase back to the 10.6 branch is ok and edit this PR (near title) to target 10.6 branch.

akshatnehra:
> Thanks for the fix. I'm thinking the original premise of touch/chmod is incorrect. There there's a very small race condition where the file can be opened, before the chmod is applied, and then read when its its later populated.
> 
> Probably best fixed with `umask` and omit `chmod`. Does that sound better? As security sensitive fix, a rebase back to the 10.6 branch is ok and edit this PR (near title) to target 10.6 branch.

Thanks for reviewing my CR and providing suggestion, I will make the required changes soon.

ottok:
Side note: I believe this script is a historical artifact from 20+ years ago and no longer serves any purpose. Anyone installing MariaDB will have a secure installation out-of-the-box. We should not perpetuate any assumptions that the default installation is somehow insecure and running this script somehow helps, as it does not. This whole script could probably be replaced with an `echo` statement saying that "MariaDB is secure by default and allows only a local root user to access it. To create a database and a user for an application, and to allow MariaDB to accept connections over the network and to have all connections TLS encrypted, please see tutorial at https://...".

grooverdan:
> Side note: I believe this script is a historical artifact from 20+ years ago and no longer serves any purpose. Anyone installing MariaDB will have a secure installation out-of-the-box. We should not perpetuate any assumptions that the default installation is somehow insecure and running this script somehow helps, as it does not. This whole script could probably be replaced with an `echo` statement saying that _"MariaDB is secure by default and allows only a local root user to access it. To create a database and a user for an application, and to allow MariaDB to accept connections over the network and to have all connections TLS encrypted, please see tutorial at https://..."_.

I don't disagree. Attempts and making it non-breaking where apparently too invasive ([bb-10.4-anel-mysql-secureinstall](https://github.com/MariaDB/server/compare/bb-10.4-anel-mysql-secureinstall))

akshatnehra:
Thank you Otto for the detailed explanation about the script's historical context. I have made the requested changes - using umask instead of touch/chmod to avoid the race condition, and rebased the PR to target the 10.6 branch as this is a security fix. I apologize for the delay in making these updates. Please let me know if any further changes are needed. Thank you

svoj:
The fix was approved by @grooverdan on zulip Oct 16: "This minor fix in the pr looks ok, for 10.11."

PR #4019, opened 2025-04-25
MDEV-36092 New-style hint: [NO_]SPLIT_MATERIALIZED

DaveGosselin-MariaDB:
Support for optimizer hints NO_SPLIT_MATERIALIZED and SPLIT_MATERIALIZED.  These hints allow fine-grained control of the "lateral derived" optimization within a query.

### DIFF
--- a/mysql-test/main/opt_hints_split_materialized.test
+++ b/mysql-test/main/opt_hints_split_materialized.test
@@ -1,0 +1,525 @@
+--echo #
+--echo # MDEV-36092 New-style hint: [NO_]SPLIT_MATERIALIZED
+--echo #
+
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+
+set @save_optimizer_switch=@@optimizer_switch;
+
+--echo #
+--echo # case 1
+--echo #
+create table t1 (
+  n1 int(10) not null,
+  n2 int(10) not null,
+  c1 char(1) not null,
+  key c1 (c1),
+  key n1_c1_n2 (n1,c1,n2)
+) engine=innodb charset=latin1;
+insert into t1 values (0, 2, 'a'), (1, 3, 'a');
+insert into t1 select seq+1,seq+2,'c' from seq_1_to_1000;
+
+analyze table t1;
+
+--echo # default showing lateral derived optimization in use
+explain
+select t1.n1 from t1, (select n1, n2 from t1 where c1 = 'a' group by n1) as t
+  where t.n1 = t1.n1 and t.n2 = t1.n2 and c1 = 'a' group by n1;
+
+select t1.n1 from t1, (select n1, n2 from t1 where c1 = 'a' group by n1) as t
+  where t.n1 = t1.n1 and t.n2 = t1.n2 and c1 = 'a' group by n1;
+
+--echo # cases where hint disables lateral derived optimization
+explain
+select /*+ no_split_materialized(t) */ t1.n1 from t1,
+ (select n1, n2 from t1 where c1 = 'a' group by n1) as t
+  where t.n1 = t1.n1 and t.n2 = t1.n2 and c1 = 'a' group by n1;
+
+select /*+ no_split_materialized(t) */ t1.n1 from t1,
+ (select n1, n2 from t1 where c1 = 'a' group by n1) as t
+  where t.n1 = t1.n1 and t.n2 = t1.n2 and c1 = 'a' group by n1;
+
+--echo # query having two subqueries, both of which are subject to LATERAL DERIVED optimization
+--echo # explain and result without hints
+explain
+with cte as (
+select t1.n1 from t1, (select n1, n2 from t1 where c1 = 'a' group by n1) as t
+  where t.n1 = t1.n1 and t.n2 = t1.n2 and c1 = 'a' group by n1
+union
+select t1.n1 from t1, (select n1, n2 from t1 where c1 = 'a' group by n1) as t
+  where t.n1 = t1.n1 and t.n2 = t1.n2 and c1 = 'a' group by n1)
+select * from cte;
+
+with cte as (
+select /*+ QB_NAME(qb1) */ t1.n1 from t1, (select n1, n2 from t1 where c1 = 'a' group by n1) as t
+  where t.n1 = t1.n1 and t.n2 = t1.n2 and c1 = 'a' group by n1
+union
+select /*+ QB_NAME(qb2) */ t1.n1 from t1, (select n1, n2 from t1 where c1 = 'a' group by n1) as t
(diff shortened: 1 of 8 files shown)


mysql-test/main/opt_hints_split_materialized.test
Olernov, line 1: Nice test cases, but I can see only `explain` outputs are tested. Would be great to also see that result sets are not affected by the hints (i.e., outputs of `select`s)

mysql-test/main/opt_hints_split_materialized.test
```diff
@@ -0,0 +1,337 @@
+--echo #
+--echo # MDEV-36092 New-style hint: [NO_]SPLIT_MATERIALIZED
+--echo #
+
+--source include/have_innodb.inc
+--source include/have_sequence.inc
+
+set @save_optimizer_switch=@@optimizer_switch;
+
+### case 1
```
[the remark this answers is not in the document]
DaveGosselin-MariaDB: Will do.

mysql-test/main/opt_hints_split_materialized.test
DaveGosselin-MariaDB: Thanks! 😊 Will add result sets.

Olernov, line 1: Is it possible to construct a query having two subqueries, where both of them are subject for the lateral derived optimization? Preferably, using a common subset of tables in both subqueries. Then adding hints to either one or another subquery could possibly highlight something interesting. Before the hints we either applied the optimization to the whole statement or didn't apply it at all, but selective application is something new.

sql/opt_split.cc
```diff
@@ -367,8 +368,8 @@ bool JOIN::check_for_splittable_materialized()
   ORDER *partition_list= 0;
   st_select_lex_unit *unit= select_lex->master_unit();
   TABLE_LIST *derived= unit->derived;
-  if (!(optimizer_flag(thd, OPTIMIZER_SWITCH_SPLIT_MATERIALIZED)) ||  // !(1)
-      !(derived && derived->is_materialized_derived()) ||             // !(2)
+
+  if (!(derived && derived->is_materialized_derived()) ||             // !(2)
```
Olernov, line 372: Numeration now starts from 2, I'd suggest adjusting it here and in the function description above. The new condition `if (!is_split_materialized_allowed)` can also be marked with some number and reflected in the description

mysql-test/main/opt_hints_split_materialized.test
DaveGosselin-MariaDB: This is an interesting idea, I'll experiment for a bit and see what I come up with.

DaveGosselin-MariaDB: Yes, I was able to construct such a query using `UNION` and the new hint to control lateral derived behaviors on each half of the union.  It behaves as expected and produces correct results.  I'll update the PR with the new test cases.

spetrunia (commented):
@DaveGosselin-MariaDB , example where hint doesn't force it:

```sql

create table ten(a int);
insert into ten select seq from seq_1_to_10;

create table one_k(a int);
insert into one_k select seq from seq_1_to_1000;

CREATE TABLE `t1000` (
  `grp` int(11) DEFAULT NULL,
  `val` int(11) DEFAULT NULL,
  KEY `grp` (`grp`)
);

insert into t1000 select A.seq, B.seq from seq_1_to_100 A, seq_1_to_10 B;
analyze table t1000;
```
This works:
```sql
explain 
select /*+ SPLIT_MATERIALIZED(TBL) */ * 
from 
  ten T1, (select grp, count(*) from t1000 group by grp) TBL where TBL.grp=T1.a;

+------+-----------------+------------+------+---------------+------+---------+---------+------+-------------+
| id   | select_type     | table      | type | possible_keys | key  | key_len | ref     | rows | Extra       |
+------+-----------------+------------+------+---------------+------+---------+---------+------+-------------+
|    1 | PRIMARY         | T1         | ALL  | NULL          | NULL | NULL    | NULL    | 10   | Using where |
|    1 | PRIMARY         | <derived2> | ref  | key0          | key0 | 5       | j4.T1.a | 1    |             |
|    2 | LATERAL DERIVED | t1000      | ref  | grp           | grp  | 5       | j4.T1.a | 10   | Using index |
+------+-----------------+------------+------+---------------+------+---------+---------+------+-------------+
```
This doesn't (the query is the same except table ten is changed to one_k)
```
explain 
select /*+ SPLIT_MATERIALIZED(TBL) */ * 
from 
  one_k T1, (select grp, count(*) from t1000 group by grp) TBL where TBL.grp=T1.a;
+------+-------------+------------+-------+---------------+------+---------+---------+------+----------------------------------------------+
| id   | select_type | table      | type  | possible_keys | key  | key_len | ref     | rows | Extra                                        |
+------+-------------+------------+-------+---------------+------+---------+---------+------+----------------------------------------------+
|    1 | PRIMARY     | T1         | ALL   | NULL          | NULL | NULL    | NULL    | 1000 | Using where                                  |
|    1 | PRIMARY     | <derived2> | ref   | key0          | key0 | 5       | j4.T1.a | 10   |                                              |
|    2 | DERIVED     | t1000      | index | grp           | grp  | 5       | NULL    | 1000 | Using index; Using temporary; Using filesort |
+------+-------------+------------+-------+---------------+------+---------+---------+------+----------------------------------------------+

```

DaveGosselin-MariaDB:
> @DaveGosselin-MariaDB , example where hint doesn't force it:

Fixed in latest patch, thank you for this great example!

spetrunia:
Ok, retrying the example above on the latest patch and indeed one can use a hint to force split-materialized...

```sql
analyze format=json select  * from one_k T1, (select grp, count(*) from t1000 group by grp) TBL where TBL.grp=T1.a\G
analyze format=json select /*+ SPLIT_MATERIALIZED(TBL) */ * from one_k T1, (select grp, count(*) from t1000 group by grp) TBL where TBL.grp=T1.a\G
```

It's interesting that the cost becomes *lower*:
* without hint: query_block.cost= 2.047789395
* with hint: query_block.cost= 0.4491132

Looks like two things are wrong: if the cost is lower with split-materialized why didn't we pick it by default,  second, why is the cost lower if we're going to access all  GROUP BY groups?

Also filed https://jira.mariadb.org/browse/MDEV-36913. 
Well, all of this seems to be outside of the scope of this MDEV.

spetrunia (commented):
Please check and apply this patch https://gist.github.com/spetrunia/efdf4394e386a367e441a75d5c27ea22

it does this:
- print "forced_by_hint" only if it's true (which is rare)
- move hint_table_state() declaration to be together with others.
-  fix coding style in Table_levelhint_table_state_hint::resolve and 
- make  is_split_materialized_allowed use pointer for OUT parameter

Ok to push after this is done

PR #4023, opened 2025-04-28
MDEV-30831 fixing, aws_key_management can compile with new aws sdk on UBUNTU 22.04, 24.04

csenki:
- [x] *The Jira issue number for this PR is: MDEV-30831

## Description
MDEV-30831 fixing, aws_key_management plugin can compile with new aws sdk on UBUNTU 24.04 and 22.04
Its fix working with mariadb-10.11.

## Release Notes
Little changes, only change 2 CMakeLists.txt and some include added to aws_key_management_plugin.cc

## How can this PR be tested?
I tested build process.
cmake ../mariadb-serever -DPLUGIN_AWS_KEY_MANAGEMENT=DYNAMIC -DAWS_SDK_EXTERNAL_PROJECT=1 -DNOT_FOR_DISTRIBUTION=1
cd plugin/aws_key_management
make
I tested the plugin:
Install plugin
alter table .... ENGINE=INNODB encrypted=yes;
alter table .... ENGINE=INNODB encrypted=no;
I tested mariadb start and stop.

## PR quality check


plugin/aws_key_management/CMakeLists.txt
```diff
@@ -9,9 +9,34 @@ ENDIF()
 MYSQL_ADD_PLUGIN(aws_key_management
   aws_key_management_plugin.cc
   COMPONENT aws-key-management)
+set(AWS_SDK_CPP_LIB_DIR ${CMAKE_BINARY_DIR}/extra/aws_sdk/aws_sdk_cpp/lib)
+
+target_link_libraries(aws_key_management
```
vaintroub, line 14: you do not know whether target exists, but you're still using target_link_libraries.
You'd need to move it under IF(TARGET) section.

Also, the idea was that USE_AWS_SDK_LIBS would be everything one needs, and not refering to more obscure libraries you listed. So whatever adjustments are needed, should be ideally moved into USE_AWS_SDK_LIBS.

Also, it worked on Windows in the past. Can you make it work there, again?

csenki: The USE_AWS_SDK_LIBS() macro only links in the main components of the AWS SDK, but does not include the necessary CRT (aws-c-*) and utility libraries (libs2n.a ...) that are required for linking. For minimal functionality, an additional target_link_libraries call is required.
Windows question ! The original ticket was covered only linux build. (Sorry alos I haven't Windows for testing and  build)

csenki:
if everything went well in f8e908d is complete (sorry but I'm not a git guru)

vaintroub:
Hi @csenki,

Thank you for taking the time to put together this commit — I appreciate the effort and research you put into it.

That said, I'm considering closing this PR without merging in its current form. While it addresses an issue on Linux(where the original bug was reported), this plugin has been cross-platform in the beginning, unfortunately it  broke since as stopped build it on CI . Fix limited to Ubuntu, or to Linux, or to *nix with GNU linker, it isn’t sufficient.

My point,  To move forward, we’d need a more robust solution that:

- Works across platforms, with ExternalProject, but also works with package managers that include the AWS C++ SDK (I can think only of vcpkg now)

- Avoids workarounds like the `-Wl,start-group`, which should be replaced with correctly specified link dependencies between AWS SDK libraries, declared as imported targets in CMake.  Bulk of the logic (newly introduced intermediate SDK libraries) needs to be in aws_sdk.cmake

- Is validated using CI, especially if you don’t have access to a Windows or macOS machine locally. I understand that current Buildbot CI is not flexible to allow you to change, but one  can *temporarily* manipulate CMake variables  (e.g do not require NOT_FOR_DISTRIBUTION) to see this plugin built on variety of OS.

If you’re interested in addressing these points, I’m happy to keep the PR open while you revise it. Otherwise, I’m okay updating the code myself, as the changes aren’t too complex.

Let me know how you’d like to proceed.

vaintroub:
closing, because a comprehensive fix dd2982dc333c05c04dac0b3970a0176bd5f809b2 was pushed.

PR #4027, opened 2025-04-29
MDL BF-BF conflict on ALTER and UPDATE with multi-level foreign key parents

hemantdangi-gc:
Issue:
Mariadb acquires additional MDL locks on UPDATE/INSERT/DELETE statements
on table with foreign keys. For example, table t1 references t2, an
UPDATE to t1 will MDL lock t2 in addition to t1.
A replica may deliver an ALTER t1 and UPDATE t2 concurrently for
applying. Then the UPDATE may acquire MDL lock for t1, followed by a
conflict when the ALTER attempts to MDL lock on t1. Causing a BF-BF
conflict.

Solution:
Additional keys for the referenced/foreign table needs to be added
to avoid potential MDL conflicts with concurrent update and DDLs.


sql/sql_base.cc
```diff
@@ -3846,6 +3846,45 @@ open_and_process_table(THD *thd, TABLE_LIST *tables, uint *counter, uint flags,
...
+      While doing sematic analysis it opens all required tables to check access
+      rights. These operations to test prepared statements does not require to
+      append table level shared key, they will be added later.
+    */
+    if (tables->prelocking_placeholder == TABLE_LIST::PRELOCK_FK &&
+        WSREP(thd) && !thd->wsrep_applier && !error &&
+        wsrep_is_active(thd) &&
+        thd->wsrep_cs().state() == wsrep::client_state::s_exec &&
+        (sql_command_flags[lex->sql_command] &
+         (CF_UPDATES_DATA | CF_DELETES_DATA)))
```
FooBarrior, line 3869: You might want to use trg_event_map or slave_fk_event_map in TABLE_LIST. Take a look at `DML_prelocking_strategy::handle_table` and `prepare_fk_prelocking_list` - that might be a more apropriate place for this code.

FooBarrior: I will be extra happy if you will find the way to contain all this designation and handling code inside some wsrep*.cc file. Anyways, it's reasonable to extract it into a separate function.

FooBarrior:
Hi, @hemantdangi-gc ! I don't know the reason you've closed it, but please consider this my proposal to changes, if you are still going to push your fix.

PR #4028, opened 2025-04-29
MDEV-36714: Rows_log_event::write_data_header overallocates buffer size

bnestere:
Nothing is broken by this, but the function Rows_log_event::write_data_header slightly over-allocates its buf by 2 bytes. It allocates 10 (ROWS_HEADER_LEN_V2), but only 8 are ever written.

This patch fixes the buf size to match what is written (i.e. ROWS_HEADER_LEN, or 8 bytes).

### DIFF
--- a/sql/log_event_server.cc
+++ b/sql/log_event_server.cc
@@ -6501,7 +6501,7 @@ Rows_log_event::do_update_pos(rpl_group_info *rgi)
 
 bool Rows_log_event::write_data_header()
 {
-  uchar buf[ROWS_HEADER_LEN_V2];        // No need to init the buffer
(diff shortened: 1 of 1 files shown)


knielsen (approved):
Hi Brandon,

The patch is ok, but better use ROWS_HEADER_LEN_V1 to match the call to
write_data() at the end of the function (ROWS_HEADER_LEN is defined in
log_event_old.h and gone in 11.4 AFAICS).

Also, why do this non-critical cleanup in stable 10.11, I'd suggest to do it
in main instead.

 - Kristian.

bnestere:
@knielsen Good points, will do on both of those, thanks for reviewing so quickly!

bnestere:
Pushed into main (12.0) branch as 88cc11dc54b.

PR #4030, opened 2025-05-02
MDEV-36729: ha_example::show_func_example is incorrectly defined

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36729*

## Description

In the main.plugin this function is called assuming the function prototype int (*)(THD *, st_mysql_show_var *, void *, system_status_var *, enum_var_type)' as changed in b4ff64568c88ab3ce559e7bd39853d9cbf86704a.

We update the ha_example::show_func_example to match the prototype on which it is called.

## Release Notes

internal test issue.

## How can this PR be tested?

Under clang UBSAN tests run the unit test main.plugin.

- [X] *This is a bug fix of a type than cannot be filtered with UBSAN filters that would otherwise block testing on the 10.6 branch with an updated UBSAN tester (MDBF-741).

## PR quality check


montywi (approved):
Ok to push

PR #4031, opened 2025-05-02
MDEV-35009: Initialize affected_rows in SQL service

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-35009*

## Description

The SQL service leaves the affected rows uninitialized.

The initialization of the spider plugin that uses
this service will fail under MSAN because there isn't an initialized value to return at the end of the query.

## Release Notes

## How can this PR be tested?

## PR quality check


mariadb-YuchenPei (approved):
LGTM, thanks!

Maybe I should demand a video explanation for every PR now

PR #4037, opened 2025-05-07
MDEV-18983 Port rpl_semi_sync_master_wait_for_slave_count from MySQL

ParadoxV5:
# WIP

* [x] *The Jira issue number for this PR is: [MDEV-18983](https://jira.mariadb.org/browse/MDEV-18983)*

An amazing description should answer some questions like:
## Description
TODO: fill description here

## Release Notes
TODO

## How can this PR be tested?
TODO

## PR quality check


sql/sys_vars.cc
```diff
@@ -3826,6 +3826,15 @@ static bool fix_rpl_semi_sync_master_wait_point(sys_var *self, THD *thd,
   return false;
 }
 
+static bool fix_rpl_semi_sync_master_wait_for_slave_count
+  (sys_var *self, THD *thd, enum_var_type type)
+{
+  mysql_mutex_unlock(&LOCK_global_system_variables);
+    repl_semisync_master.refresh_wait_for_slave_count(thd->variables.server_id);
```
ParadoxV5, line 3833: Ideally, the refresh is only required when the requirement decreases.
However, it seems that the Sys.Var. system has a limitation that the variable must have an ‘old’ duplicate that an `ON_UPDATE` callback can compare to tell whether it was decreased.

Intriguingly, other Semi-Sync Primary variables _are_ duplicated and updated through `ON_UPDATE` even though they don’t have special handling.
This is likely a remnant of Semi-Sync’s past as a plugin.

ParadoxV5 (commented):
The feature should be implemented. While it’s ready for review besides testing, I would prefer we first refactor our code to solve the hindrances I encountered.

sql/sys_vars.cc
```diff
@@ -3883,6 +3892,18 @@ Sys_semisync_master_wait_point(
...
        ON_UPDATE(fix_rpl_semi_sync_master_wait_point));
 
+static Sys_var_on_access_global<Sys_var_uint,
+           PRIV_SET_SYSTEM_GLOBAL_VAR_RPL_SEMI_SYNC_MASTER_WAIT_FOR_SLAVE_COUNT>
+Sys_semisync_master_wait_for_slave_count(
+       "rpl_semi_sync_master_wait_for_slave_count",
+       "The number of slaves that need to acknowledge that they have received "
+       "a transaction before the transaction can complete on the master",
+       GLOBAL_VAR(rpl_semi_sync_master_wait_for_slave_count),
+       CMD_LINE(REQUIRED_ARG), VALID_RANGE(1, 0xFFFF),
```
ParadoxV5, line 3902: matches MySQL

sql/semisync_master.cc
```diff
@@ -230,23 +250,16 @@ bool Active_tranx::is_tranx_end_pos(const char *log_file_name,
...
-
-  DBUG_PRINT("semisync", ("%s: probe (%s, %lu) in entry(%u)",
-                          "Active_tranx::is_tranx_end_pos",
-                          log_file_name, (ulong)log_file_pos, hash_val));
-
-  DBUG_RETURN(entry != NULL);
+Tranx_node *Active_tranx::find_acked_tranx_node()
+{
+  Tranx_node *new_front;
+  for (Tranx_node *entry= m_trx_front; entry; entry= entry->next)
```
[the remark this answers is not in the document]
ParadoxV5: `Active_tranx` is a *naturally*-ordered, block-allocating, hashtable with linked list, specialized for managing … its elements.

bnestere:
Hi @ParadoxV5 !

Thanks for submitting the PR! A couple quick questions:

 1. Why did you not include any tests? Regardless of whether or not we refactor, we'd need MTR tests for the feature (which wouldn't change whether or not a refactor is included).
 
 2. Do you have a specific refactor design in mind? I agree the semi-sync code-base is spaghetti-adjacent, and I'm open to improvements.

ParadoxV5:
Hi @bnestere,
 
> 1. Why did you not include any tests? Regardless of whether or not we refactor, we'd need MTR tests for the feature (which wouldn't change whether or not a refactor is included).

It is simply because I have not begun writing them. I wanted to share those thoughts with the team early.

> 2. Do you have a specific refactor design in mind? I agree the semi-sync code-base is spaghetti-adjacent, and I'm open to improvements.

Besides improving the inconveniences I had, I don’t have exact designs in mind, especially when also considering adjacent topics such as proofing for #3931 and coherence with async.

bnestere:
> It is simply because I have not begun writing them. I wanted to share those thoughts with the team early.

Ah ok. Agreed that early feedback is ideal :) Overall, I'd say your patch aligns with what I had in mind for the feature, though I didn't do a detailed review. To the general early-feedback process, if you're just looking for design-level feedback (which it seems you are doing here), it should be ok to just write-up/sketch out the idea of the implementation (preferably before starting implementation, as to save time in-case the design is changed). Then later on when you share code, I'd say to at least provide a minimal test to show the code works (even for early feedback).

> Besides improving the inconveniences I had, I don’t have exact designs in mind, especially when also considering adjacent topics such as proofing for https://github.com/MariaDB/server/pull/3931 and coherence with async.

Ok. My thoughts on the current semi-sync implementation aren't that the implementation is necessarily wrong, just that it is hard to understand the details of what is happening, and when. I wonder if this doesn't really even need that large of a refactoring, but perhaps just some small encapsulations, similar to what you did in this patch, to improve readability. Or at least, it would be a start (as you've already started doing :)). Perhaps if there is time left-over at the end of the sprint, we can address it.

ParadoxV5:
> … just that it is hard to understand the details of what is happening, and when.

The “what” is straightforward to encapsulate, but the “when”… 🤔

sql/semisync_master.cc
```diff
@@ -705,16 +673,19 @@ int Repl_semi_sync_master::report_reply_binlog(uint32 server_id,
...
-    if (m_active_tranxs->is_empty())
-      m_wait_file_name_inited= false;
+    entry= m_active_tranxs->get_tranx_node(log_file_name, log_file_pos);
+    if (entry && ++(entry->acks) >= rpl_semi_sync_master_wait_for_slave_count)
+    {
+      /* Remove all active transaction nodes before this point. */
+      m_active_tranxs->clear_active_tranx_nodes(entry->next);
+      if (m_active_tranxs->is_empty())
+        m_wait_file_name_inited= false;
+    }
```
bnestere, line 688: I think now that we could need multiple ACKs, this logic here which does the
actual clearing and signalling probably shouldn't be grouped in the
need_copy_send_pos conditional. I imagine that if there are a couple
transactions actively awaiting acks, and one slave is fast, and another is
slow, that when handling ACKs from the slow slave, need_copy_send_pos would
be false because m_reply_file_{pos,name} would already be updated from the
fast slave. It'd be good to have a test case for this scenario too.

sql/semisync_master.cc
```diff
@@ -1499,6 +1469,16 @@ void Repl_semi_sync_master::await_all_slave_replies(const char *msg)
   DBUG_VOID_RETURN;
 }
 
+void Repl_semi_sync_master::clear_acked_tranx_nodes()
+{
+  DBUG_ENTER("refresh_wait_for_slave_count");
```
bnestere, line 1474: This DBUG_ENTER function name isn't consistent with the real function name (and
note my above comments about changing the function name too).

bnestere (changes requested):
Thanks for the PR submission @ParadoxV5!

I know the review is a bit preliminary, as you haven't officially requested it and you have a couple more changes you wanted to make before doing so - but I won't be around tomorrow, so I figured I'd at least review what is currently up to give some early feedback. Most of my comments are minor, with one exception being that I think ACKs can be lost if one slave is lagging behind another.

sql/semisync_master.cc
```diff
@@ -230,42 +237,24 @@ bool Active_tranx::is_tranx_end_pos(const char *log_file_name,
...
-                          "Active_tranx::is_tranx_end_pos",
-                          log_file_name, (ulong)log_file_pos, hash_val));
-
-  DBUG_RETURN(entry != NULL);
+  DBUG_RETURN(get_tranx_node(log_file_name, log_file_pos));
 }
 
-void Active_tranx::clear_active_tranx_nodes(
-    const char *log_file_name, my_off_t log_file_pos,
-    active_tranx_action pre_delete_hook)
```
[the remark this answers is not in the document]
ParadoxV5: 🤔

sql/semisync_master.cc
```diff
@@ -641,6 +611,7 @@ int Repl_semi_sync_master::report_reply_packet(uint32 server_id,
 
   rpl_semi_sync_master_get_ack++;
   report_reply_binlog(server_id, log_file_name, log_file_pos);
+  DEBUG_SYNC(current_thd, "semisync_got_reply");
```
ParadoxV5, line 614: @bnestere can you help me with Debug Sync? The program appears never to hit this sync point.

Good call on that there’s only one singular ACK receiver thread, by the way.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
```diff
@@ -0,0 +1,111 @@
...
+  SHOW STATUS LIKE 'Rpl\_semi\_sync\_master\_%\_tx';
+
+  --echo # Case 2.4: returns to Semi-Sync when the requirement is set just right
+  SET @@GLOBAL.rpl_semi_sync_master_wait_for_slave_count= 2;
+  SET DEBUG_SYNC= 'semisync_got_reply SIGNAL wait WAIT_FOR continue EXECUTE 2';
+  --eval INSERT INTO t VALUES (4$wait_no_slave) # ask for ACK(s)
+  # Wait for 2 *complete* ACK processings
+  SET DEBUG_SYNC= 'now WAIT_FOR wait';
+  SET DEBUG_SYNC= 'now SIGNAL continue WAIT_FOR wait';
+  SET DEBUG_SYNC= 'now SIGNAL continue';
```
ParadoxV5, line 101: This is because it is ACKs that automatically turn Semi-Sync back on.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
```diff
@@ -0,0 +1,111 @@
...
+SHOW STATUS LIKE 'Rpl\_semi\_sync\_master\_%\_tx';
+
+
+--let $wait_no_slave= 0
+while ($wait_no_slave < 2)
+{
+  --echo # Cases 2.x Matrix:
+  --eval SET @@GLOBAL.rpl_semi_sync_master_wait_no_slave= $wait_no_slave
+
+  --echo # Case 2.1: falls back to Async when the requirement is set too high
```
ParadoxV5, line 69: Should I `SET @@GLOBAL.rpl_semi_sync_master_timeout` to a short(er) wait?

sql/sys_vars.cc
```diff
@@ -3826,6 +3826,15 @@ static bool fix_rpl_semi_sync_master_wait_point(sys_var *self, THD *thd,
   return false;
 }
 
+static bool fix_rpl_semi_sync_master_wait_for_slave_count
+  (sys_var *self, THD *thd, enum_var_type type)
+{
+  mysql_mutex_unlock(&LOCK_global_system_variables);
```
ParadoxV5, line 3832: I wonder how important this is, exactly?

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
```diff
@@ -0,0 +1,140 @@
...
+  --send FLUSH GLOBAL STATUS
+  if ($wait_no_slave)
+  {
+    --echo # Case 2.1.1: An up-to-date slaves does not ACK when it reconnects
+    --connection default
+    --sync_slave_with_master server_3
+    --source include/stop_slave.inc
+    --source include/start_slave.inc
+    --connection server_1
+  }
```
ParadoxV5, line 112: ### About this strange Case 2.1.1
As noted in https://github.com/MariaDB/server/pull/4037#discussion_r2087499079 point 2, `Repl_semi_sync_master::report_reply_packet()` isn’t the only caller of `Repl_semi_sync_master::report_reply_binlog()` – `Repl_semi_sync_master::dump_start()` also calls it.
This makes me concerned that the connection of a semi-sync replica also counts as an ACK, at least for whatever code reüse convenience it was before `…_wait_for_slave_count`.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
```diff
@@ -0,0 +1,68 @@
...
+
+  --echo # Case 2.3: returns to Semi-Sync when the requirement is set just right
+  SET @@GLOBAL.rpl_semi_sync_master_wait_for_slave_count= 2;
+  --eval INSERT INTO t VALUES (3$wait_no_slave)
+  SHOW STATUS LIKE 'Rpl\_semi\_sync\_master\_%\_tx';
+
+  --inc $wait_no_slave
+}
+
+
```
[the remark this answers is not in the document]
ParadoxV5: I expanded Test Case 1 (now renumbered as Case 0 to signify its fundamentality) to cover the “slow” replica situation.
And you were right, this is a bug, and it has been fixed.

sql/semisync_master.cc
ParadoxV5: To fix the “slow replica” bug (https://github.com/MariaDB/server/pull/4037#discussion_r2118545543), I moved the `m_reply_file_{pos,name}` updates to (a reüsable method that’s called) under the `if (entry …)` condition.
Now, there is practically nothing between `if (need_copy_send_pos)` and this `if`.

Although I think, when `entry= …->get_tranx_node(…)` exists, `need_copy_send_pos` would always be true.
I think the converse is only false when this code is called from `Repl_semi_sync_master::dump_start()` (see https://github.com/MariaDB/server/pull/4037#discussion_r2118543247).
Overall, I think the `m_reply_file_{name,pos}` variables are now only significant for the states of `is_no_slave()` that flush transactions without updating this position.

`Repl_semi_sync_master::report_reply_binlog()` might have an identity crisis if we have more to refactor away.

ParadoxV5 (commented):
Cheers, @bnestere, the project will be ready once CI passes and I squash everything and rebase.

<details>
<summary>

### Commit Message

</summary>

```md
MDEV-18983 Port rpl_semi_sync_master_wait_for_slave_count from MySQL

`rpl_semi_sync_master_wait_for_slave_count` is a long-requested
variable specifying the minimum number of acknowledging replicas
before a semi-synchronous transaction can complete on the primary.

Consequently, its purpose also leads it to indirectly control the
number of replicas required to keep semi-synchronous replication
active when `rpl_semi_sync_master_wait_no_slave` is set to disable
immediately (i.e. auto-revert to asynchronous).

This commit also deduplicates code in `sql/semisync_master.cc`:
* Create reüsable functions
  `is_no_slave()` & `Active_tranx::get_tranx_node()`
  * Changed `Active_tranx:clear_active_tranx_nodes()` from locating
    nodes with the given position to use a pointer given by callers
    so they make use of this `get_tranx_node()` or an alternative.
    * Inline `signal_waiting_transaction()`
* Replace `Active_tranx::is_thd_waiter()`
  with equivalent method `is_tranx_end_pos()`

Reviewed-by: Brandon Nesterenko <brandon.nesterenko@mariadb.com>
```
</details>

sql/semisync_master.cc
```diff
@@ -641,6 +612,13 @@ int Repl_semi_sync_master::report_reply_packet(uint32 server_id,
 
   rpl_semi_sync_master_get_ack++;
   report_reply_binlog(server_id, log_file_name, log_file_pos);
+  DBUG_EXECUTE_IF("sync_semisync_report_reply",
+    debug_sync_set_action(current_thd, STRING_WITH_LEN(
+      /*FIXME: no DBUG_ASSERT due to
+        MDEV-32748 debug_sync_set_action() Fails if Entered in an Error State */
+      "now SIGNAL report_reply_wait WAIT_FOR report_reply_continue"
+    ));
+  );
```
ParadoxV5: Oops, I placed the comment on the wrong line.
```suggestion
  /*FIXME: no DBUG_ASSERT due to
    MDEV-32748 debug_sync_set_action() Fails if Entered in an Error State */
  DBUG_EXECUTE_IF("sync_semisync_report_reply",
    debug_sync_set_action(current_thd, STRING_WITH_LEN(
      "now SIGNAL report_reply_wait WAIT_FOR report_reply_continue"
    ));
  );
```

But really, we should figure out what happened to the ACK receiver thread that it managed to scrape up a handled
“Got an error reading communication packets” and lingers it in `thd->is_error()`.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
```diff
@@ -0,0 +1,140 @@
...
+SET @@GLOBAL.debug_dbug= '+d,simulate_delay_semisync_slave_reply';
+
+# Queue 2 transactions and wait for their *complete* `server_2` ACK processings
+--let $rpl_connection_name= master
+--let $rpl_server_number= 1
+--source include/rpl_connect.inc
+--connection master
+SET @@GLOBAL.debug_dbug= '+d,sync_semisync_report_reply';
+# Case 2.1 expects a timeout, but this suppression is also replicated
+--send CALL mtr.add_suppression('Timeout waiting for reply of binlog')
```
bnestere: Using an `mtr.add_suppression` call as a transaction isn't straightforward (though it is clever :joy:). Our convention is to list the suppressions in the pre-amble/setup (usually with `set statement sql_log_bin=0 for call...` so it is specific to the server which is expecting the error), and then make the transactions based on DDL/DML. I'd say to follow the convention, as clarity is more important when going through tests/results.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
bnestere: I think the `dump_start()` case is ok. It calls `report_reply_packet()` using the coordinates the the slave connected with, which means the slave has already received up to that point. Which is correct - imagine the case where a slave disconnects right before it sends an `ACK`; the slave has received and relay-logged all events for the transaction, but the master never got the `ACK`. Now when the slave re-connects, it will provide its current state; and if that matches with the currently-waiting-for-an-ACK connection, it can be released.

sql/semisync_master.cc
```diff
@@ -705,30 +683,33 @@ int Repl_semi_sync_master::report_reply_binlog(uint32 server_id,
...
+l_end:
   unlock();
-
-
   DBUG_RETURN(0);
 }
 
+void Repl_semi_sync_master::report_reply_trx(const Tranx_node &entry)
+{
+  strmake_buf(m_reply_file_name, entry.log_name);
```
bnestere, line 707: Now that the context of these updates (i.e how it is called) isn't immediately obvious, it'd be good to assert the locking state is correct (in addition to your `.h` comment mentioning the `@pre`).


```
    mysql_mutex_assert_owner(&LOCK_binlog);
```

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
bnestere: That's a good idea, it's always good to reduce test time, when you can.

sql/semisync_master.h
```diff
@@ -697,6 +707,12 @@ class Repl_semi_sync_master
   /*called before reset master*/
   int before_reset_master();
 
+  /**
+    With lock() held, call Active_tranx::clear_active_tranx_nodes()
+    with Active_tranx::find_acked_tranx_node() if there is one
+  */
```
bnestere, line 715: For header-level comments, it's better to summarize the purpose of a function, rather than how it is implemented. Implementations often change, and these types of comments are not usually changed with them, easily leading to inconsistencies and harder to maintain code. The purpose of a function shouldn't ever really change though.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
ParadoxV5: 😅 like `FLUSH GLOBAL STATUS`, them beïng replicated gave my test results off-by-ones.
But yeah, it’s a good idea to use `sql_log_bin=0` as the replicas don’t require this suppression.
Case 0 asserts with two *concurrent* transactions, though, so no `INSERT INTO t` here.

The `FLUSH GLOBAL STATUS`s in Cases 2.x are mandatory to erase the stats of the non-consistent steps, yet they are also themselves “transactions” in those state-sensitive tests.
Update: Looks like I can also wrap them with `SET STATEMENT sql_log_bin=0 FOR`!

sql/semisync_master.cc
ParadoxV5: Sure. It’s not currently mandated, but it’s good to do so for assurance.
Both `report_reply_binlog()` and `clear_active_tranx_nodes()` now call `report_reply_trx()`; though other places could also use a validation using it 🤔…

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
ParadoxV5: \* `report_reply_binlog`, not `report_reply_packet`

So the replica connects with the start position of the unreceived, which is the end position of the last received.
What about the `Active_tranx` entries, do they use start or end positions?

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
ParadoxV5: Except I’m not sure about how short to reduce it to, when our overloaded CI can time itself out just because it doesn’t have time to run a process.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
bnestere: The Active_tranx entries use the end of the transaction. Note too that slaves only send ACKs for end-of-transaction binlog events, e.g. `COMMIT` Query_log_events and Xid_log_events; they don't ACK each and every event they receive.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
bnestere: I think tests which rely on timing granularities of `2s` or lower are usually the ones that fail individually. Otherwise, when the CI overloads, usually the whole chunk of all tests all concurrently running on that builder will fail. So I'd bet anywhere in the range of `3s`-`5s` would be ok; then if we're wrong, we'll just adjust it :shrug:.

ParadoxV5: > then if we're wrong, we'll just adjust it 🤷.

Many Hearthstone cards today get buffed in one patch but get reverted in the next immediate patch. 😁

sql/sys_vars.cc
bnestere: It's still a good idea, as `repl_semisync_master.clear_acked_tranx_nodes();` takes a lock of its own. Looking through other functions, it looks like the _only_ function which doesn't release `LOCK_global_system_variables` is `update_slave_skip_counter` - I wonder if that is intentional, or if it is vulnerable to deadlock/slowing down other connections which are updating variables.

mysql-test/suite/rpl/t/rpl_semi_sync_master_wait_for_slave_count.test
ParadoxV5: Done, though I moved the `mtr.add_suppression` call to above the 2.x matrix to signify that they’re the only ones expecting it.

ParadoxV5:
I’ll squash & rebase and queue for QA after CI.
In the meantime, @bnestere, you can take a look at the changes in the latest commits.
(Plural: missed one item 😅)

ParadoxV5:
It failed with “Warning 1639 debug sync point wait timed out” again, this time on [amd64-ubuntu-2004-debug/31281](https://buildbot.mariadb.org/#/builders/369/builds/31281).
It might be very “normal”… 🫤

ParadoxV5:
Force Push:
* Adjusted the feature test: The setups for cases 2.2 and 2.3 were different.
* Edited the commit message

PR #4038, opened 2025-05-08
MDEV-36760 log_t::append_prepare_wait(): Bogus assertion on write_lsn

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36760*
## Description
`log_t::append_prepare_wait()`: Do not attempt to read `log_sys.write_lsn` because it is not protected by `log_sys.latch` but by `write_lock`, which we cannot hold here. The assertion could fail if `log_t::write_buf()` is executing concurrently, and it has not yet executed `log_write_buf()` or updated `log_sys.write_lsn`.

Fixes up #3925
## Release Notes
N/A. This replaces a bogus debug assertion.
## How can this PR be tested?
Run a write-heavy workload with the default `innodb_log_buffer_size` and possibly `innodb_flush_log_at_trx_commit=0`. Possibly, add a sleep to `log_write_buf()`. If running on Linux `/dev/shm`, be sure to compile with `cmake -DWITH_INNODB_PMEM=OFF` in order to cover the correct code path.

I tested this by running the default `mtr` suites while `mysql-test/var` was a symbolic link to a SATA device, as well as when it was a regular directory on `/dev/shm`. That covered both code paths that are affected by these changes.
## PR quality check

### DIFF
--- a/storage/innobase/mtr/mtr0mtr.cc
+++ b/storage/innobase/mtr/mtr0mtr.cc
@@ -916,17 +916,16 @@ ATTRIBUTE_COLD void log_t::append_prepare_wait(bool late, bool ex) noexcept
   {
   got_ex:
     const uint64_t l= write_lsn_offset.load(std::memory_order_relaxed);
-    const lsn_t lsn{base_lsn.load(std::memory_order_relaxed)};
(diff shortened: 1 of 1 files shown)


mariadb-DebarunBanerjee (approved):
The change looks correct to me. The issue should affect debug build only.

PR #4047, opened 2025-05-16
MDEV-32570: Fragment ROW replication events larger than slave_max_allowed_packet

bnestere:
This PR is currently a draft.

This PR solves two problems:
  1. Rows log events cannot be transmitted to the slave if their
     size exceeds slave_max_packet_size (max 1GB at the time of
     writing this patch, i.e. MariaDB 12.1)
  2. Rows log events cannot be binlogged if they are larger than
     4GB because the common binlog event header field event_len is
     32-bits.

This PR adds support for fragmenting large Rows_log_events
through a new event type, Partial_rows_log_event. When any given
instantiation of a Rows_log_event (e.g. Write_rows_log_event, etc)
is too large to be sent to a replica (i.e. larger than the value
slave_max_allowed_packet, as configured on a replica), then the rows
event must be fragmented into sub-events (i.e.
Partial_rows_log_events), each of size slave_max_allowed_packet, so
the event can be transmitted to the replica. The replica will then
take the content of each of these Partial_rows_log_events, and join
them together into a large Rows_log_event to be executed as normal.
Partial_rows_log_events are written to the binary log sequentially,
and the replica assembles the events in the order they are 
binlogged.

Remaining things to be done to remove draft status of the PR:
  1. Implement error handling
  2. Implementing Partial_rows_log_event::pack_info()
  3. Extend MTR testing

* Commits 1 - 3 commits add code preparations to make the actual
   feature commits easier to read
* Commit 4 adds the Partial_rows_log_event type and the server logic
   to support fragmenting and re-assembling/applying large
   Rows_log_events.
* Commit 5 adds client (mysqlbinlog) logic to support output and
   replay of Partial_rows_log_events through piping to the
    mariadb client
 * Commit 6 adds the MTR test (still working on this) for the feature
 * Commit 7 adjusts existing MTR tests to pass

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_fragment_row_event.result
+++ b/mysql-test/suite/rpl/r/rpl_fragment_row_event.result
@@ -1,0 +1,1153 @@
+include/rpl_init.inc [topology=1->2, 2->3]
+# Setup Defaults for rpl_fragment_row_event.inc
+#
+#
+# Test case ) Boundary Testing
+#
+# A rows event at its maximum unfragmented length (i.e. the length of
+# slave_max_allowed_packet minus the Rows_log event metadata (e.g.
+# headers)) should not fragment, and replication replay should work both
+# via traditional replication as well as mysqlbinlog.
+connection server_1;
+set @@global.slave_max_allowed_packet= 1024;
+set @@global.net_buffer_length= 1024;
+disconnect server_1;
+connect  server_1,127.0.0.1,root,,test,$SERVER_MYPORT_1,;
+connection server_2;
+include/stop_slave.inc
+set @@global.slave_max_allowed_packet= 1024;
+set @@global.net_buffer_length= 1024;
+disconnect server_2;
+connect  server_2,127.0.0.1,root,,test,$SERVER_MYPORT_2,;
+connection server_3;
+include/stop_slave.inc
+set @@global.slave_max_allowed_packet= 1024;
+set @@global.net_buffer_length= 1024;
+disconnect server_3;
+connect  server_3,127.0.0.1,root,,test,$SERVER_MYPORT_3,;
+#
+# Initialize test data
+connection server_1;
+create table t1 (a int, b longblob) engine=innodb;
+#
+# Test Write_rows_log_event fragmentation
+connection server_2;
+include/start_slave.inc
+include/sync_with_master_gtid.inc
+include/stop_slave.inc
+connection server_3;
+include/start_slave.inc
+include/sync_with_master_gtid.inc
+include/stop_slave.inc
+# Ensuring Write_rows replication event replay is consistent..
+include/diff_tables.inc [server_1:test.t1, server_2:test.t1, server_3:test.t1]
+# Ensuring Write_rows data was fragmented into Partial_rows_log_events
+# Server_1 binlogs for Write_rows event
+connection server_1;
+flush logs;
+# MYSQL_BINLOG /home/brandon/workspace/server/build/mysql-test/var/mysqld.1/data//master-bin.000001 --result-file=/home/brandon/workspace/server/build/mysql-test/var/binlog_insert_frags.out --base64-output=never
+NOT FOUND matches in binlog_insert_frags.out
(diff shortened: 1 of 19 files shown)


mysql-test/suite/rpl/r/rpl_fragment_row_event.result
```diff
@@ -0,0 +1,1645 @@
...
+include/start_slave.inc
+include/sync_with_master_gtid.inc
+include/stop_slave.inc
+# Ensuring Write_rows replication event replay is consistent..
+include/diff_tables.inc [server_1:test.t1, server_2:test.t1, server_3:test.t1]
+# Ensuring Write_rows data was fragmented into Partial_rows_log_events
+# Server_1 binlogs for Write_rows event
+connection server_1;
+flush logs;
+# MYSQL_BINLOG /home/brandon/workspace/server/build/mysql-test/var/mysqld.1/data//master-bin.000001 --result-file=/home/brandon/workspace/server/build/mysql-test/var/binlog_insert_frags.out --base64-output=never
```
hemantdangi-gc: change path, this is local to your machine.

mysql-test/suite/rpl/t/rpl_fragment_row_event.inc
```diff
@@ -0,0 +1,301 @@
...
+--echo # Initialize test data
+--connection server_1
+create table t1 (a int, b longblob) engine=innodb;
+
+--echo #
+--echo # Test Write_rows_log_event fragmentation
+--disable_query_log
+if ($insert_size)
+{
+  # subtract 8 from the insert for the first column
```
hemantdangi-gc: mentioned 'subtract 8 from the insert for the first column', but subtracted 9.

sql/log_event_server.cc
```diff
@@ -5582,6 +5582,302 @@ bool Rows_log_event::write_data_body(Log_event_writer *writer)
...
+    }
+  }
+  assembler= rgi->assembler;
+
+  /*
+    TODO: Needs stage
+  */
+  if ((res= assembler->append(this)))
+  {
+    goto end;
```
hemantdangi-gc, line 5698: In case of error, allocated assembler is not getting deleted. Please add:
```
    delete rgi->assembler;
    rgi->assembler= NULL;
```

sql/log_event_server.cc
```diff
@@ -5582,6 +5582,302 @@ bool Rows_log_event::write_data_body(Log_event_writer *writer)
...
+                     rgi->gtid_info(),
+                     ER_THD(rgi->thd, ER_PARTIAL_ROWS_LOG_EVENT_BAD_STREAM),
+                     buf, this->last_fragment_seen + 1, this->total_fragments);
+    return ER_PARTIAL_ROWS_LOG_EVENT_BAD_STREAM;
+  }
+
+  if (this->last_fragment_seen == 0)
+  {
+    /*
+      Alloc the size of the string (overestimating the size, as the last
```
hemantdangi-gc: The total size of whole event is available when we are fragmenting into multiple Partial_rows_log_event. So better to add new total_size field into Partial_rows_log_event and send instead of allocating extra memory. For high load this can afffect memory uisage. If not now may be add it as todo for future.

sql/log_event_server.cc
ParadoxV5: We use C++ extensively, but use a lot more `goto end`s than RAII. 😐

sql/log_event_server.cc
bnestere: Good idea! We have until January to get it done for the 12.3 release, so I should be able to get it done by then.

sql/log_event_server.cc
bnestere: It gets deleted in the `rpl_group_info` destructor (say just in case the stream ends and not all Partial log event fragments are found), but it is a good idea to delete it here too.

bnestere:
This was manually merged: c580d17f290ecf1b774aec33b393eb71378ae37a...1eff7ddd810ada007fd6c4e5bf302eaf77708b72

PR #4050, opened 2025-05-19
MDEV-36330: SERIALIZABLE read inconsistency

dr-m:
- [x] *The Jira issue number for this PR is: MDEV-36330*
## Description
`trans_begin()`: At `SERIALIZABLE` isolation level, handle `START TRANSACTION` in the same way as `START TRANSACTION WITH CONSISTENT SNAPSHOT` so that write/read conflicts can be detected.

`innobase_start_trx_and_assign_read_view()`: At any other isolation level than `READ UNCOMMITTED`, do create a read view. This is needed so that at `SERIALIZABLE` isolation level and `innodb_snapshot_isolation=ON` we are able to flag `ER_CHECKREAD` when reading data that was committed after the transaction started.
## Release Notes
Transactions running at `SERIALIZABLE` isolation level failed to return `ER_CHECKREAD` for non-serializable executions (reading data that was committed after the current transaction started) even if `innodb_snapshot_isolation=ON`.
## How can this PR be tested?
```sh
./mtr innodb.lock_isolation
```
## PR quality check


sql/transaction.cc
```diff
@@ -217,7 +217,8 @@ bool trans_begin(THD *thd, uint flags)
 #endif //EMBEDDED_LIBRARY
 
   /* ha_start_consistent_snapshot() relies on OPTION_BEGIN flag set. */
-  if (flags & MYSQL_START_TRANS_OPT_WITH_CONS_SNAPSHOT)
+  if ((flags & MYSQL_START_TRANS_OPT_WITH_CONS_SNAPSHOT) ||
+      thd->tx_isolation == ISO_SERIALIZABLE)
```
vuvova: I am not sure it's correct. The explanation looks very InnoDB-specific, why would a serializable mode need to start a consistent snapshot "so that write/read conflicts can be detected" in a more general case?

Perhaps it's something that can be done inside InnoDB? Like if `ha_innobase::external_lock()` detects that the isolation mode is SERIALIZABLE, it starts a consistent snapshot.

vuvova (changes requested):

sql/transaction.cc
dr-m: Thank you for the prompt review.

Which other storage engines support transactions than InnoDB? I will try to extend the test case for RocksDB. The logical place for creating a snapshot for the purpose of isolation (no matter if it is serializable isolation) is at the start of a transaction. If a transaction involves multiple storage engines, then the snapshot really has to be created in all storage engines at the same point of time. Inside an implementation of the virtual member function `handler::external_lock()` it would be more error-prone to try to tell other storage engines to start a snapshot.

I find the original InnoDB or MySQL design choice to not create a read view at the start of transaction rather questionable. I understand that we can’t change that because of compatibility. But, to me, a read view pretty much defines a transaction. The deferred read view creation feels almost like violating isolation or consistency (the CI in ACID) on purpose.

dr-m: This change cannot affect `ENGINE=RocksDB` because of the following error:
```
ER_ISOLATION_MODE_NOT_SUPPORTED (4057): MyRocks supports only READ COMMITTED and REPEATABLE READ isolation levels. Please change from current isolation level SERIALIZABLE
```
My test would provoke the following response from both `ENGINE=Aria` and `ENGINE=MyISAM`:
```
Warning	1105	This MariaDB server does not support any consistent-read capable storage engine
```
My test was as follows:
```sql
CREATE TABLE t1 (c0 INT);
CREATE TABLE t2 (c0 INT PRIMARY KEY AUTO_INCREMENT, c1 INT);
INSERT INTO t1(c0) VALUES (1),(2),(3);
--connect con1,localhost,root
SET SESSION TRANSACTION ISOLATION LEVEL SERIALIZABLE;
BEGIN; INSERT INTO t2 (c1) VALUES(1);
--connect con2,localhost,root
SET SESSION TRANSACTION ISOLATION LEVEL SERIALIZABLE;
BEGIN; INSERT INTO t2 (c1) VALUES(2),(3);
COMMIT;
--disconnect con2
--connection con1
UPDATE t1 SET c0 = 10 WHERE c0 in
(SELECT c1 FROM (SELECT * FROM t2 ORDER BY c0 LIMIT 2) AS t);
COMMIT;
--disconnect con1
--connection default
SELECT * FROM t2;
SELECT * FROM t1;
DROP TABLE t1,t2;
```
Which transactional storage engine do you have in mind? I can only think of `ENGINE=PBXT` and `ENGINE=BDB`, perhaps `ENGINE=Gemini` as well, but I don’t think that any of them are currently supported by MariaDB Server.

vuvova: I didn't mean any in particular (although, I think, ColumnStore, FederatedX and Spider can do transactions too), more like if we'll be changing the server to be InnoDB-specific, it'll be a hell of a task to add a new transactional storage engine, if we'd want to do it. A "consistent view" is a concept of an MVCC engine, 2PL engine doesn't necessarily have it.

PBXT, BDB, and Gemini are, indeed, not supported by MariaDB, and the last two never were :smile: very few know they even existed.

I don't think a consistent snapshot starts simultaneously over all engines anyway, it's sequential, so I wasn't talking about multi-engine case. One single engine, just not necessarily InnoDB. In that case, creating a read view on the first access is fine and is completely in line with the SQL Standard.

That's why I think it's InnoDB-specific detail and must be done in `ha_innobase::external_lock`

dr-m: To my understanding, InnoDB currently is the only storage engine that supports the `SERIALIZABLE` isolation level. I thought that when we get another such storage engine, we can adjust this logic. An option could be to extend `handler::Table_flags` with another flag, say, `HA_CAN_SERIALIZE`. Unfortunately, it is already using close to 64 bits.

vuvova: My point mainly is that we can tell ourselves "we get another such storage engine, we can adjust this logic", but then if we'll really need to do that, all these special cases will be spread everywhere through the code, and it'll be next to impossible to get rid of them, there's no list of assumptions (and there can never be, nobody will maintain it).
So my preference would be not to introduce them in the first place. If you'd like I can try to fix it inside InnoDB

dr-m: Based on [this comment](https://github.com/MariaDB/server/pull/3676#discussion_r1882096400) in #3676, it should be possible to implement a hack in `ha_innobase::store_lock()` to create a read view and to start the transaction in case it is not yet started and the isolation level is `SERIALIZABLE`. This function is already responsible for closing any current read view if the isolation level is `READ COMMITTED` or `READ UNCOMMITTED`.

vlad-lesin (approved):
Looks good to me. I would also be good to document the innodb_snapshot_isolation influence on isolation levels other than REPEATABLE READ.

PR #4054, opened 2025-05-21
MDEV-31949 III. Innodb flush_log_later for XA commit,rollback

andrelkin:
This commit carries changes to Innodb for the user XA's binlog-coordinated
 ordered commit. It's the part III of the series aimed at
improving on parallel slave performance and providing crash-recovery.

As the part I has unified binlog coordinated completion (commit and
rollback) of the XA transactions with the normal ones a remained
task is to optimize flush to disk call away from the XA completion execution
in engine like it's done to the normal transaction.
Namely innobase_{rollback,commit}_by_xid()^\footnote{Unlike the normal
transaction case, XA-rollback also needs a similar addressing} now
raise a flag blocking the flush before innobase_commit_low().
The flag is restored upon that, which of course what the normal
transaction does in its execution path past the low-commit.

Tests demonstrate the changes are effectual, e.g
binlog_xa_prepared_disconnect.test " failed ", to require it to
adapted to possible loss of the commit/rollback upon crash.
The actual binlog recovery in such scenarios is only going to be recovered in the part
IV of MDEV-33168. 

Feasibility of successful recovery of
XA-"COMPLETE" is based on a plain observation, that in order to decide
whether to commit an XA or not it's sufficient to follow
the normal transaction recovery rule: when `xid` is found in binlog (that's MDEV-33168)
the transaction gets completed (committed in the normal transaction case).


storage/innobase/handler/ha_innodb.cc
```diff
@@ -2953,8 +2953,19 @@ static int innobase_rollback_by_xid(XID *xid) noexcept
...
+    ut_ad(!trx->mysql_thd);
+    THD* thd = current_thd;
+    if (thd)
+    {
+      /* Symmetrically to innobase_commit_by_xid the flush is
+      delayed in binlog coordinated order commit. */
+      thd_binlog_pos(thd, &trx->mysql_log_file_name, &trx->mysql_log_offset);
+      if (trx->mysql_log_offset > 0)
+        trx->flush_log_later = true;
+    }
```
dr-m, line 2965: How exactly is this logic supposed to change the existing InnoDB behaviour? Unfortunately, the pull request description is not following the template, in particular the section
# How can this PR be tested?
I would like to see some output, such as stack traces of calls to `pwrite`, `fdatasync` or `fsync` on the `ib_logfile0` and the binlog files, or some output of `strace`, which shows the difference of this change in some particular scenarios. The following part of `trx_rollback_for_mysql()` is relevant here:
```c++
			mtr_t		mtr;
			mtr.start();
			if (trx_undo_t* undo = trx->rsegs.m_redo.undo) {
				trx_undo_set_state_at_prepare(trx, undo, true,
							      &mtr);
			}
			/* Write the redo log for the XA ROLLBACK
			state change to the global buffer. It is
			not necessary to flush the redo log. If
			a durable log write of a later mini-transaction
			takes place for whatever reason, then this state
			change will be durable as well. */
			mtr.commit();
			ut_ad(mtr.commit_lsn() > 0);
```

PR #4055, opened 2025-05-21
MDEV-36125 [NO_]INDEX_MERGE Hint

DaveGosselin-MariaDB:
This PR consists of two commits.  The first is to address some small cleanups and refactors which make the second (the commit that implements the hints) somewhat clearer.

    MDEV-36125 [NO_]INDEX_MERGE Hint

    Introduces NO_INDEX_MERGE and INDEX_MERGE, which control whether or
    not index merge strategies are used during query optimization.  Here
    is an example query from the tests:

      SET optimizer_switch='index_merge_intersection=off';
      EXPLAIN SELECT /*+ INDEX_MERGE(t1 f4, f2) */ COUNT(*) FROM t1
      WHERE f4 = 'h' AND f2 = 2;

    with the hint in place, the query plan will employ the index_merge
    intersect strategy (abbreviated EXPLAIN output):

      type          Extra
      index_merge   Using intersect(f2,f4); Using where; Using index

    The presence of the [NO_]INDEX_MERGE hint overrides the optimizer's
    choice of keys during the index merge optimization.  As we see in
    the above example, keys f2 and f4 and given and the optimizer will
    consider only those keys for this query.

    When the hint is given without any particular keys, as in
    INDEX_MERGE(table), then all keys are considered.  In this case, the
    cheapest index merge among the keys should be used.  When
    NO_INDEX_MERGE(table) is given, then index merge is disabled for
    that table.

    When the hint is given with one or more keys, then only those keys
    are considered.  In the case of NO_INDEX_MERGE, those keys are
    excluded.  This can lead to no merged indexes at all, because
    there may not be sufficient row-ordered read columns available for
    consideration.

    The index merge strategies of intersection, union, and sort union
    cannot themselves be directly controlled via the hints.  In combination
    with the optimizer switches for the same, the strategy may be
    indirectly controlled but this is not guaranteed.

    When the hint directs the optimizer such that insufficient ROR scans
    are available, thus leading to a situation where the INDEX_MERGE hint
    cannot be honored, the server will emit a warning to that effect.

    In the hints module (opt_hints*{cc,h}), this commit adds some
    index merge-specific functionality to make interpreting hint state
    at callsites in the optimizer cleaner and more intuitive.
    Particularly, we add a bit field to the table hints class which
    indicates the keys that are marked by an [NO_]INDEX_MERGE hint, if
    present.  A new function, index_merge_hint (and associated new
    helper functions) relies on this field when interpreting index merge
    hint state for the optimizer.

    If there are no index merges available prior to attemping to find
    a suitable union/sort union, then the optimizer will not attempt
    it.  This change results in optimizer trace output which does not
    include the 'analyzing_index_merge_union' block when there are no
    merges.

    Parts of this implementation based on MySQL commit
    ebcb981807e3d91a64782e89d48e1a25622eafea


spetrunia:
Looking at the cleanup patch, first.
test_if_quick_select() before the patch:

```cpp
  if (!keys_to_use.is_clear_all())
  {
    // a lot here ...
   }
  set_if_smaller(records, table_records);
```
After:
```cpp
  if (keys_to_use.is_clear_all())
    DBUG_RETURN(returnval);
  ...
  set_if_smaller(records, table_records);
```
So, set_if_smaller() won't be done when keys_to_use.is_clear_all()==true.
Is this intentional?

spetrunia:
@DaveGosselin-MariaDB , is the following behavior expected:

```sql
create table ten(a int primary key);
insert into ten values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
create table one_k(a int primary key);
insert into one_k select A.a + B.a* 10 + C.a * 100 from ten A, ten B, ten C;
create table t1 (a int, b int, c int, d int, index(a),index(b),index(c),index(d));
insert into t1 select a,a,a,a from test.one_k;
```

```sql
MariaDB [test]> explain select /*+ INDEX_MERGE(t1 c,d) */ *  from t1 where (a <3 or b <3 ) and (c=3 and d=4);
+------+-------------+-------+-------------+---------------+------+---------+------+------+-----------------------------------+
| id   | select_type | table | type        | possible_keys | key  | key_len | ref  | rows | Extra                             |
+------+-------------+-------+-------------+---------------+------+---------+------+------+-----------------------------------+
|    1 | SIMPLE      | t1    | index_merge | a,b,c,d       | c,d  | 5,5     | NULL | 1    | Using intersect(c,d); Using where |
+------+-------------+-------+-------------+---------------+------+---------+------+------+-----------------------------------+
1 row in set (0.002 sec)
```
OK
```
MariaDB [test]> explain select /*+ INDEX_MERGE(t1 a,b) */ *  from t1 where (a <3 or b <3 ) and (c=3 and d=4);
+------+-------------+-------+------+---------------+------+---------+-------+------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref   | rows | Extra       |
+------+-------------+-------+------+---------------+------+---------+-------+------+-------------+
|    1 | SIMPLE      | t1    | ref  | a,b,c,d       | c    | 5       | const | 1    | Using where |
+------+-------------+-------+------+---------------+------+---------+-------+------+-------------+
1 row in set (0.002 sec)
```
Here, it seems ref access is used regardless of the hint to use index_merge(a,b)...

while in MySQL:
```sql
mysql> explain select /*+ INDEX_MERGE(t1 a,b) */ *  from t1 where (a <3 or b <3 ) and (c=3 and d=4);
+----+-------------+-------+------------+-------------+---------------+------+---------+------+------+----------+------------------------------------+
| id | select_type | table | partitions | type        | possible_keys | key  | key_len | ref  | rows | filtered | Extra                              |
+----+-------------+-------+------------+-------------+---------------+------+---------+------+------+----------+------------------------------------+
|  1 | SIMPLE      | t1    | NULL       | index_merge | a,b,c,d       | a,b  | 5,5     | NULL |    6 |     0.83 | Using sort_union(a,b); Using where |
+----+-------------+-------+------------+-------------+---------------+------+---------+------+------+----------+------------------------------------+
1 row in set, 1 warning (0,00 sec)
```
Note that  `INDEX_MERGE(t1 a b)`  can work against range access:
```
MariaDB [test]> explain select /*+ INDEX_MERGE(t1 a,b) */ *  from t1 where (a <3 or b <3 ) and (c<2 and d<2);
+------+-------------+-------+-------------+---------------+------+---------+------+------+------------------------------------+
| id   | select_type | table | type        | possible_keys | key  | key_len | ref  | rows | Extra                              |
+------+-------------+-------+-------------+---------------+------+---------+------+------+------------------------------------+
|    1 | SIMPLE      | t1    | index_merge | a,b,c,d       | a,b  | 5,5     | NULL | 4    | Using sort_union(a,b); Using where |
+------+-------------+-------+-------------+---------------+------+---------+------+------+------------------------------------+
1 row in set (0.002 sec)
```
This follows the  hint, without hint I get range(c).


EDIT:
Looking at the trace of the problematic query,  I see that `chosen_range_access_summary`  does show the requested index merge. 
The problem is that best_access_path then ignores this choice...

spetrunia:
for the query
```sql
explain select /*+ INDEX_MERGE(t1 a,b) */ *  from t1 where (a <3 or b <3 or c<3);
```
both MySQL and MariaDB use full table scan (Ok)
In the trace, MariaDB has:
```js
                          {
                            "range_scan_alternatives": [
                              {
                                "index": "c"
                              }
                            ],
                            "chosen": false,
                            "cause": "cost"
                          }
```
while MySQL has
```js
                          "range_scan_alternatives": [
                            {
                              "index": "c",
                              "chosen": false,
                              "cause": "index_merge_hint"
                            }
                          ],
                          "chosen": false,
                          "cause": "cost"
                        }
```
I'm not yet sure if I should request the trace to be adjusted...

spetrunia:
I get assertion failure with this:
```sql
create table t2 ( c1 int, c2 int, c3 int, index(c1),index(c2),index(c3));
insert into t2 select mod(a,1000), mod(a, 250), mod(a, 75) from test.one_k;
insert into t2 select mod(a,1000), mod(a, 250), mod(a, 75) from test.one_k;
set optimizer_switch='rowid_filter=off';
alter table t2 add c1a int; 
update t2 set c1a=c1;
alter table t2 add index(c1a);
set optimizer_trace=1;
```

```sql
-- Attempt to get union(intersect(c1,c2), c1a):
explain select /*+ INDEX_MERGE(t2 c1,c2,c1a) */ * from t2 where c1=1 and c2=1 and c3=1 or c1a=1;
```

```
sql/my_json_writer.cc:163: Json_writer& Json_writer::add_member(const char*, size_t): Assertion  `is_uniq_key' failed.
```

spetrunia:
This line in get_key_scans_params() is very suspicious:

```cpp
      else if (ihb == index_merge_behavior::TABLE_DISABLED)
        DBUG_RETURN(nullptr);
```
This basically means NO_INDEX_MERGE(tbl) will prevent construction of any range access, too? 

Indeed:
```
MariaDB [test]>  explain select * from t2 where c1< 100 and c2<1;
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------------+
| id   | select_type | table | type  | possible_keys | key  | key_len | ref  | rows | Extra                              |
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------------+
|    1 | SIMPLE      | t2    | range | c1,c2         | c2   | 5       | NULL | 8    | Using index condition; Using where |
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------------+
```

and now the hint will cause full scan instead of range scan:
```
MariaDB [test]>  explain select /*+ NO_INDEX_MERGE(t2) */ * from t2 where c1< 100 and c2<1;
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows | Extra       |
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
|    1 | SIMPLE      | t2    | ALL  | c1,c2         | NULL | NULL    | NULL | 2000 | Using where |
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
1 row in set (0.002 sec)
```

EDIT: it seems, MySQL does the same. I don't think we should follow them here.

spetrunia:
```cpp
index_merge_behavior index_merge_hint(const TABLE *table,
                                     uint keyno,
                                     bool &force_index_merge,
                                     bool &use_cheapest_index_merge)
```
Remember Monty saying that OUT parameters should be passed by reference as it makes it unclear they are out?
```
index_merge_hint( .... x, y);
```
doesn't explicitly say x and y are modified by the function. Compare with
```
index_merge_hint( .... &x, &y);
```
where it's much more clear that x and y are likely to be modified.

Please change.

sql/opt_range.cc
```diff
@@ -5475,10 +5547,11 @@ TABLE_READ_PLAN *get_best_disjunct_quick(PARAM *param, SEL_IMERGE *imerge,
...
 
-  if (imerge_too_expensive || (imerge_cost > read_time) ||
-      ((non_cpk_scan_records+cpk_scan_records >=
-        param->table->stat_records()) &&
-       read_time != DBL_MAX))
+  if (imerge_too_expensive || (((imerge_cost > read_time) ||
+                               ((non_cpk_scan_records+cpk_scan_records >=
+                                 param->table->stat_records()) &&
+                                read_time != DBL_MAX)) &&
+                               !force_index_merge))
```
spetrunia, line 5532: Do I understand it correctly:  if `imerge_too_expensive==true`  then we will take this if-branch, execute this code

```cpp
    trace_best_disjunct.add("chosen", false).add("cause", "cost");
    DBUG_RETURN(NULL);
```

ignoring the fact that `force_index_merge=true` due to hints? 
Is this intended behavior? 

Also, I think the condition in the if statement became unreadable and needs to be written in a more human-readable way.

DaveGosselin-MariaDB:
> So, set_if_smaller() won't be done when keys_to_use.is_clear_all()==true. Is this intentional?

Unintentional, fixed.

DaveGosselin-MariaDB:
> @DaveGosselin-MariaDB , is the following behavior expected:
> 
> ...
>
> This follows the hint, without hint I get range(c).
> 
> EDIT: Looking at the trace of the problematic query, I see that `chosen_range_access_summary` does show the requested index merge. The problem is that best_access_path then ignores this choice...

It is not expected.  As you surmised, the problem was in `best_access_path` and I fixed it. The branch starting with the following comment
```
     Don't test table scan if it can't be better.
     Prefer key lookup if we would use the same key for scanning.
```
needed to be updated to check if there is a quick select present and it has the `force_index_merge` member set.  Then we get the expected result without any regressions in other test cases.

DaveGosselin-MariaDB:
> I get assertion failure with this:
> 
> ...
> 
> ```
> sql/my_json_writer.cc:163: Json_writer& Json_writer::add_member(const char*, size_t): Assertion  `is_uniq_key' failed.
> ```

Fixed.

DaveGosselin-MariaDB:
> This line in get_key_scans_params() is very suspicious:
> 
> ```c++
>       else if (ihb == index_merge_behavior::TABLE_DISABLED)
>         DBUG_RETURN(nullptr);
> ```
>
>... 
> 
> EDIT: it seems, MySQL does the same. I don't think we should follow them here.

Fixed this by removing this line from the method (no regressions introduced) and the generated plan is now:
```
MariaDB [test]> explain select * from t2 where c1< 100 and c2<1;
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------------+
| id   | select_type | table | type  | possible_keys | key  | key_len | ref  | rows | Extra                              |
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------------+
|    1 | SIMPLE      | t2    | range | c1,c2         | c2   | 5       | NULL | 8    | Using index condition; Using where |
+------+-------------+-------+-------+---------------+------+---------+------+------+------------------------------------+
1 row in set (0.008 sec)

```

DaveGosselin-MariaDB:
> ```c++
> index_merge_behavior index_merge_hint(const TABLE *table,
>                                      uint keyno,
>                                      bool &force_index_merge,
>                                      bool &use_cheapest_index_merge)
> ```
> 
> Remember Monty saying that OUT parameters should be passed by reference as it makes it unclear they are out?

These are passed by reference, but if you prefer them by pointer I'm happy to change it.

sql/opt_range.cc
DaveGosselin-MariaDB: Yes, you are correct.  The equivalent logic in MySQL is from `range_optimizer.cc` around line 1080 (MySQL 9):
```
 1080     trace_best_disjunct.add(key: "cost_of_reading_ranges", value: imerge_cost);
 1081     if (imerge_too_expensive || (((imerge_cost > read_cost) ||
 1082                                   ((non_cpk_scan_records + cpk_scan_records >=
 1083                                     table->file->stats.records) &&
 1084                                    read_cost != DBL_MAX)) &&
 1085                                  !force_index_merge)) {
```
So we mimic the same behavior for compatibility.

spetrunia:
```
commit 998e4b4f582dec28e46098e9453b74f678ef4af5
Author: Dave Gosselin <dave.gosselin@mariadb.com>
Date:   Thu May 1 14:55:46 2025 -0400

    MDEV-36125 Cleanup ahead of [NO_]INDEX_MERGE
    
    Introduces code cleanup and small refactorings, such as:
    
      - Flip keys_to_use.is_clear_all() and dedent
      - Rename best_trp to best_read_plan
      - Fix key_parts assignment spacing
      - find_shortest_key keeps happy path to the left
      - Cleanup in get_best_ror_intersect, remove_nonrange_trees,
        and function restore_nonrange_trees functions
      - Reorganization of locals within test_quick_select
```
Here 

dedent = de-ident ? (not as obvious for non native speakers)

>         and function restore_nonrange_trees functions

this needs rewording.

It's really good that these changes are in separate commit.

sql/opt_hints.cc
```diff
@@ -810,6 +808,26 @@ void Opt_hints_table::append_hint_arguments(THD *thd, opt_hints_enum hint,
...
+void Opt_hints_table::set_index_hint(Opt_hints *hint, uint arg)
+{
+  hint->set_fixed();
+  keyinfo_array[arg]= static_cast<Opt_hints_key *>(hint);
+  incr_fully_fixed_children();
+
+  /*
+    Update the index_merge_map to note that the key
+    is referenced by a [NO_]INDEX_HINT associated with
+    the query.
```
spetrunia: Does it mean "with the table" ?

sql/opt_hints.cc
```diff
@@ -905,10 +923,150 @@ hint_state hint_table_state(const THD *thd,
...
+}
+
+
+// See comment in header file.
+index_merge_behavior index_merge_hint(const TABLE *table,
+                                      uint keyno)
+{
+    bool force_index_merge_IGNORED= false,
+        use_cheapest_index_merge_IGNORED= false;
+    return index_merge_hint(table,
```
spetrunia: * Please use correct indentation.
* What is the purpose of initializing variables that are only OUT parameters?

sql/opt_hints.cc
DaveGosselin-MariaDB: Ultimately, yes, and maybe wording it that way is clearer.

sql/opt_hints.cc
DaveGosselin-MariaDB: I have an ingrained habit of initializing variables cultivated over the years by bugs arising from uninitialized variables.  There (should be) no runtime cost for basic types initialized inline.

DaveGosselin-MariaDB:
> dedent = de-ident ? (not as obvious for non native speakers)
> 
> > ```
> >     and function restore_nonrange_trees functions
> > ```
> 
> this needs rewording.

Will hyphenate and reword. 

> It's really good that these changes are in separate commit.

Agreed, thanks.

spetrunia:
The printout of INDEX_MERGE hint  in EXPLAIN EXTENDED has room for improvement.


```sql
create table ten(a int primary key);
insert into ten values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);

create table one_k(a int primary key);
insert into one_k select A.a + B.a* 10 + C.a * 100 from ten A, ten B, ten C;

create table ten_k(a int primary key);
insert into ten_k select A.a + B.a* 1000 from one_k A, ten B;

create table t1 (a int, b int, c int, filler varchar(100), index(a),index(b),index(c));
insert into t1 select a, a, a, uuid() from ten_k;
```
```sql
explain extended select /*+ INDEX_MERGE(t1 b,c) */  * from t1 where a<2 and (b<100 or c <100);
```

```
+------+-------------+-------+-------------+---------------+------+---------+------+------+----------+------------------------------------+
| id   | select_type | table | type        | possible_keys | key  | key_len | ref  | rows | filtered | Extra                              |
+------+-------------+-------+-------------+---------------+------+---------+------+------+----------+------------------------------------+
|    1 | SIMPLE      | t1    | index_merge | a,b,c         | b,c  | 5,5     | NULL | 200  |     1.00 | Using sort_union(b,c); Using where |
+------+-------------+-------+-------------+---------------+------+---------+------+------+----------+------------------------------------+
1 row in set, 1 warning (0.002 sec)

Note (Code 1003): select /*+ INDEX_MERGE(`t1`@`select#1`) INDEX_MERGE(`t1`@`select#1` `b`) INDEX_MERGE(`t1`@`select#1` `c`) */ `j200`.`t1`.`a` AS `a`,`j200`.`t1`.`b` AS `b`,`j200`.`t1`.`c` AS `c`,`j200`.`t1`.`filler` AS `filler` from `j200`.`t1` where `j200`.`t1`.`a` < 2 and (`j200`.`t1`.`b` < 100 or `j200`.`t1`.`c` < 100)
````
Note the 
```
INDEX_MERGE(`t1`@`select#1`) INDEX_MERGE(`t1`@`select#1` `b`) INDEX_MERGE(`t1`@`select#1` `c`) 
```
If I actually try using such syntax
```sql
 explain extended select /*+  INDEX_MERGE(t1) INDEX_MERGE(t1 b) INDEX_MERGE(t1 c) */  * from t1 where a<2 and (b<100 or c <100);
```
I get 
```
Warning (Code 4238): Hint INDEX_MERGE(`t1` `b`) is ignored as conflicting/duplicated (an index hint of the same type has already been specified for this table)
Warning (Code 4238): Hint INDEX_MERGE(`t1` `c`) is ignored as conflicting/duplicated (an index hint of the same type has already been specified for this table)
```

Compare with MySQL:
```sql
mysql> explain select /*+ INDEX_MERGE(t1 b,c) */  * from t1 where a<2 and (b<100 or c <100);
```
...
```
| Note  | 1003 | /* select#1 */ select /*+ INDEX_MERGE(`t1`@`select#1` `b`, `c`) */ `j200`.`t1`.`a` AS `a`,`j200`.`t1`.`b` AS `b`,`j200`.`t1`.`c` AS `c`,`j200`.`t1`.`filler` AS `filler` from `j200`.`t1` where ((`j200`.`t1`.`a` < 2) and ((`j200`.`t1`.`b` < 100) or (`j200`.`t1`.`c` < 100))) |
```

sql/opt_range.cc
```diff
@@ -7508,7 +7617,22 @@ TRP_ROR_INTERSECT *get_best_ror_intersect(const PARAM *param, SEL_TREE *tree,
       The next ror scan is only accepted if the total cost went down
       (Enough rows where reject to offset the intersect cost)
     */
-    if (intersect->total_cost < min_cost)
+    if (intersect->total_cost < min_cost ||
+        (force_index_merge &&
+         /*
+           If INDEX_MERGE hint is used without only specified index,
+           index merge is forced and the cheapest combination of indexes
```
spetrunia, line 7601: Does "without only" mean "without any" ?

sql/opt_range.cc
```diff
@@ -7525,41 +7649,56 @@ TRP_ROR_INTERSECT *get_best_ror_intersect(const PARAM *param, SEL_TREE *tree,
...
+                        ihb == index_merge_behavior::NO_HINT);
+  }
+
+  if (cpk_merge_allowed && !intersect->is_covering)
+  {
+    if (ror_intersect_add(intersect, cpk_scan, &trace_cpk, TRUE, true) &&
+        ((intersect->total_cost < min_cost) ||
+         (force_index_merge &&
+          (!use_cheapest_index_merge ||
+           (num_scans == 1 && force_index_merge_result)))))
```
spetrunia, line 7671: The logic here looks odd:

```cpp
if (ror_intersect_add(... )  && SOME_COND) 
{
     min_cost= intersect->total_cost;
     // write clustered_pk_scan_added_to_intersect to trace
     ...
}
```
What if ror_intersect_add() returned true, but SOME_COND is false?  
the CPK scan is added, but the trace would say that it wasn't ?

spetrunia:
Please take the code  from branch `12.0-mdev-36125-index-merge-hint-cleaner` and push it into this PR.

In that branch, I took the remaining "code-cleanup" changes out of this patch and put it into a separate commit, "Cleanup ahead of [NO_]INDEX_MERGE, part 2.".

```
16f2b7e3df8 (HEAD -> 12.0-mdev-36125-index-merge-hint-cleaner, origin/12.0-mdev-36125-index-merge-hint-cleaner, tmp2) MDEV-36125 [NO_]INDEX_MERGE Hint
43faedc3d98 MDEV-36125 Cleanup ahead of [NO_]INDEX_MERGE, part 2.
998e4b4f582 MDEV-36125 Cleanup ahead of [NO_]INDEX_MERGE
e02f4d7e311 (origin/main, origin/bb-12.2-frm-staging, origin/bb-12.1-bumpversion) 12.2 branch
```
This makes it easier to review.

spetrunia:
I have a table like this:
```sql
CREATE TABLE `t1` (
  `a` int(11) DEFAULT NULL,
  `b` int(11) DEFAULT NULL,
  `c` int(11) DEFAULT NULL,
  `filler` varchar(100) DEFAULT NULL,
  `non_key` int(11) DEFAULT NULL,
  KEY `a` (`a`),
  KEY `b` (`b`),
  KEY `c` (`c`)
```
I try `where a=1 and non_key<2`  and I get a warning about "insufficient ROR scans":

```
MariaDB [j200]> explain extended select /*+ INDEX_MERGE(t1 a,b) */  * from t1 where a=1 and non_key<2;
+------+-------------+-------+------+---------------+------+---------+-------+------+----------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref   | rows | filtered | Extra       |
+------+-------------+-------+------+---------------+------+---------+-------+------+----------+-------------+
|    1 | SIMPLE      | t1    | ref  | a             | a    | 5       | const | 1    |   100.00 | Using where |
+------+-------------+-------+------+---------------+------+---------+-------+------+----------+-------------+
1 row in set, 2 warnings (0.002 sec)

Warning (Code 4207): Insufficient ROR scans available for keys given in INDEX_MERGE hint
Note (Code 1003): select /*+ INDEX_MERGE(`t1`@`select#1`) INDEX_MERGE(`t1`@`select#1` `a`) INDEX_MERGE(`t1`@`select#1` `b`) */ `j200`.`t1`.`a` AS `a`,`j200`.`t1`.`b` AS `b`,`j200`.`t1`.`c` AS `c`,`j200`.`t1`.`filler` AS `filler`,`j200`.`t1`.`non_key` AS `non_key` from `j200`.`t1` where `j200`.`t1`.`a` = 1 and `j200`.`t1`.`non_key` < 2
```
now change OR into AND: `where a=1 or non_key<2` and no warning is given:
```
MariaDB [j200]> explain extended select /*+ INDEX_MERGE(t1 a,b) */  * from t1 where a=1 or non_key<2;
+------+-------------+-------+------+---------------+------+---------+------+------+----------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra       |
+------+-------------+-------+------+---------------+------+---------+------+------+----------+-------------+
|    1 | SIMPLE      | t1    | ALL  | a             | NULL | NULL    | NULL | 9897 |   100.00 | Using where |
+------+-------------+-------+------+---------------+------+---------+------+------+----------+-------------+
1 row in set, 1 warning (0.002 sec)

Note (Code 1003): select /*+ INDEX_MERGE(`t1`@`select#1`) INDEX_MERGE(`t1`@`select#1` `a`) INDEX_MERGE(`t1`@`select#1` `b`) */ `j200`.`t1`.`a` AS `a`,`j200`.`t1`.`b` AS `b`,`j200`.`t1`.`c` AS `c`,`j200`.`t1`.`filler` AS `filler`,`j200`.`t1`.`non_key` AS `non_key` from `j200`.`t1` where `j200`.`t1`.`a` = 1 or `j200`.`t1`.`non_key` < 2
```

DaveGosselin-MariaDB:
> I have a table like this:
> 
> ```sql
> CREATE TABLE `t1` (
>   `a` int(11) DEFAULT NULL,
>   `b` int(11) DEFAULT NULL,
>   `c` int(11) DEFAULT NULL,
>   `filler` varchar(100) DEFAULT NULL,
>   `non_key` int(11) DEFAULT NULL,
>   KEY `a` (`a`),
>   KEY `b` (`b`),
>   KEY `c` (`c`)
> ```
> ...

That we emit a warning is a difference in behavior from MySQL.  I wasn't sure at the time whether we should silently fail the hint's directive if it was impossible, or inform the user (I went with the latter).  If we attempt the same queries in MySQL, we get the same query plans (but without a warning):

```
mysql> explain select /*+ INDEX_MERGE(t1 a,b) */  * from t1 where a=1 and non_key<2;
+----+-------------+-------+------------+------+---------------+------+---------+-------+------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref   | rows | filtered | Extra       |
+----+-------------+-------+------------+------+---------------+------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | t1    | NULL       | ref  | a             | a    | 5       | const |    1 |    33.33 | Using where |
+----+-------------+-------+------------+------+---------------+------+---------+-------+------+----------+-------------+
1 row in set, 1 warning (0.004 sec)

mysql> show warnings;
+-------+------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Level | Code | Message                                                                                                                                                                                                                                                                                    |
+-------+------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Note  | 1003 | /* select#1 */ select /*+ INDEX_MERGE(`t1`@`select#1` `a`, `b`) */ `test`.`t1`.`a` AS `a`,`test`.`t1`.`b` AS `b`,`test`.`t1`.`c` AS `c`,`test`.`t1`.`filler` AS `filler`,`test`.`t1`.`non_key` AS `non_key` from `test`.`t1` where ((`test`.`t1`.`a` = 1) and (`test`.`t1`.`non_key` < 2)) |
+-------+------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.002 sec)

mysql> explain select /*+ INDEX_MERGE(t1 a,b) */  * from t1 where a=1 or non_key<2;
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra       |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
|  1 | SIMPLE      | t1    | NULL       | ALL  | a             | NULL | NULL    | NULL |    6 |    44.44 | Using where |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
1 row in set, 1 warning (0.003 sec)

mysql> show warnings;
+-------+------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Level | Code | Message                                                                                                                                                                                                                                                                                   |
+-------+------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Note  | 1003 | /* select#1 */ select /*+ INDEX_MERGE(`t1`@`select#1` `a`, `b`) */ `test`.`t1`.`a` AS `a`,`test`.`t1`.`b` AS `b`,`test`.`t1`.`c` AS `c`,`test`.`t1`.`filler` AS `filler`,`test`.`t1`.`non_key` AS `non_key` from `test`.`t1` where ((`test`.`t1`.`a` = 1) or (`test`.`t1`.`non_key` < 2)) |
+-------+------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.003 sec)

sql/opt_range.cc
DaveGosselin-MariaDB: Fixed!

sql/opt_range.cc
DaveGosselin-MariaDB: In this case, `ror_intersect_add` doesn't add the CPK scan actually.  Because is_cpk_scan is TRUE, `ror_intersect_add` will compute a cost for scanning the current set of index records and add that to the intersection's index_scan_costs.  This is CPK scan merging.  After `ror_intersect_add` returns, either the CPK scan is successfully merged to the current intersection (at which point it doesn't really exist anymore) or it is dropped entirely.  I don't think that it is possible, in the case that CPK merging is allowed and the intersection isn't covering, for the trace to be incorrect here.

spetrunia:
I'm not sure if I should request changing this MDEV's patch with this, but  here's an observation of behavior that I think a user would not find desiarable, either in MySQL or in MariaaDB: 

https://jira.mariadb.org/browse/MDEV-36125?focusedCommentId=312388&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-312388

DaveGosselin-MariaDB:
> I'm not sure if I should request changing this MDEV's patch with this, but here's an observation of behavior that I think a user would not find desiarable, either in MySQL or in MariaaDB:
> 
> https://jira.mariadb.org/browse/MDEV-36125?focusedCommentId=312388&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-312388

Right now we are compatible with MySQL (the goal of this work), so I advocate your suggestion as a new optimization feature.

spetrunia (approved):

PR #4049, opened 2025-05-18
Replace deprecated CMAKE_COMPILER_IS_GNU(CC|CXX) with CMAKE_(C|CXX)_COMPILER_ID

brad0:
## Description
Replace deprecated use of CMAKE_COMPILER_IS_GNU(CC|CXX) with CMAKE_(C|CXX)_COMPILER_ID
throughout the tree. Some places already use CMAKE_(C|CXX)_COMPILER_ID.

## How can this PR be tested?

Build and check the various paths are taken as appropriate.

## Basing the PR against the correct MariaDB version
- [ ] *This is a new feature or a refactoring, and the PR is based against the `main` branch.*
- [x] *This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.*

## PR quality check
- [x] I checked the [CODING_STANDARDS.md](https://github.com/MariaDB/server/blob/-/CODING_STANDARDS.md) file and my PR conforms to this where appropriate.
- [x] For any trivial modifications to the PR, I am ok with the reviewer making the changes themselves.

PR #4056, opened 2025-05-22
Replace deprecated CMAKE_COMPILER_IS_GNU(CC|CXX) with CMAKE_(C|CXX)_COMPILER_ID

brad0:
## Description
Replace deprecated use of CMAKE_COMPILER_IS_GNU(CC|CXX) with CMAKE_(C|CXX)_COMPILER_ID
throughout the tree. Some places already use CMAKE_(C|CXX)_COMPILER_ID.

As of CMake 3.24 CMAKE_COMPILER_IS_GNU(CC|CXX) are deprecated and should be replaced with
CMAKE_(C|CXX)_COMPILER_ID which were introduced with CMake 2.6.

## How can this PR be tested?

Build and check the various paths are taken as appropriate.

## PR quality check


brad0:
cc @grooverdan

svoj:
@brad0 thanks for updating your patch. There is no need to create new PR. You can force update original branch and change target version. It is better to keep discussion in one place.

vaintroub:
You are aware that  CMAKE_COMPILER_IS_GNU(CC|CXX) and CMAKE_(C|CXX)_COMPILER_ID STREQUAL "GNU". Is not the same

That  CMAKE_COMPILER_IS_GNU(CC|CXX) is either gcc or command-line compatible clang?
In some contexts one just wants to set GNU-compatible flags. And it seems that at least mroonga might want to use it like this.

vaintroub:
Indeed, sorry for confusion, heh. this seems not to be the case already since Oct 2013  (https://cmake.org/pipermail/cmake/2013-October/056071.html) and version 2.8.10 or so, but I was sure it is still the case.

brad0:
> Indeed, sorry for confusion, heh. this seems not to be the case already since Oct 2013 (https://cmake.org/pipermail/cmake/2013-October/056071.html) and version 2.8.10 or so, but I was sure it is still the case.

Ah, I see why you thought that. When CMake didn't understand identifying Clang it made sense why it did that, but the corrections in behavior make sense to be able to differentiate the two properly (and other variants).

brad0:
Ping.

I'd like to get this merged so I can send my functional changes on top of this.

grooverdan:
ack. Did you have a reason for 10.6 from #4049?

Despite the number of my commits on 10.6 I can't actually push them.

If you want to rebase to 10.11 and then edit the title of the PR to change the base to 10.11 I can process this quickly.

svoj:
@grooverdan base branch changed

brad0:
> ack. Did you have a reason for 10.6 from #4049?

As I've seen mentioned somewhere in the docs to try and push fixes as far back as possible.

PR #4058, opened 2025-05-22
MDEV-35856: implement index hints

Olernov:
- [x] *The Jira issue number for this PR is: MDEV-35856*

## Description
This commit implements optimizer hints affecting
the use of indexes:
  - ORDER_INDEX, NO_ORDER_INDEX
  - GROUP_INDEX, NO_GROUP_INDEX
  - JOIN_INDEX, NO_JOIN_INDEX
  - INDEX, NO_INDEX

## How can this PR be tested?

`./mtr opt_hints_index`

## PR quality check

### DIFF
--- a/mysql-test/main/opt_hints_index.result
+++ b/mysql-test/main/opt_hints_index.result
@@ -1,0 +1,249 @@
+CREATE TABLE t1 (a INT, b INT, c INT, d INT,
+KEY i_a(a), KEY i_b(b),
+KEY i_ab(a,b), KEY i_c(c), KEY i_d(d));
+INSERT INTO t1 VALUES
+(1,1,1,1),(2,2,2,1),(3,3,3,1),(4,4,4,1),
+(5,5,5,1),(6,6,6,1),(7,7,7,1),(8,8,8,1);
+INSERT INTO t1 SELECT a,b, c + 10, d FROM t1;
+INSERT INTO t1 SELECT a,b, c + 20, d FROM t1;
+INSERT INTO t1 SELECT a,b, c + 40, d FROM t1;
+INSERT INTO t1 SELECT a,b, c + 80, d FROM t1;
+INSERT INTO t1 SELECT a,b, c + 160, d FROM t1;
+ANALYZE TABLE t1;
+Table	Op	Msg_type	Msg_text
+test.t1	analyze	status	Engine-independent statistics collected
+test.t1	analyze	status	OK
+# Check behavior of duplicated/intersected hints.
+# First specified hint is applied and next conflicting/intersected hints
+# are ignored with warning.
+# JOIN_INDEX(t1) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1) JOIN_INDEX(t1) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint JOIN_INDEX(`t1` ) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# INDEX(t1 i_d) is ignored as duplicated.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a, i_b, i_c) INDEX(t1 i_d) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint INDEX(`t1` `i_d`) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`,`i_b`,`i_c`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# JOIN_INDEX(t1 i_a, i_b) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a) JOIN_INDEX(t1 i_a, i_b) JOIN_INDEX(t1 i_b) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint JOIN_INDEX(`t1` `i_a`,`i_b`) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`) JOIN_INDEX(`t1`@`select#1` `i_b`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# GROUP_INDEX(t1 i_a, i_b) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a) GROUP_INDEX(t1 i_a, i_b) GROUP_INDEX(t1 i_b) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint GROUP_INDEX(`t1` `i_a`,`i_b`) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`) GROUP_INDEX(`t1`@`select#1` `i_b`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# GROUP_INDEX(t1) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a) GROUP_INDEX(t1) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint GROUP_INDEX(`t1` ) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# ORDER_INDEX(t1 i_a, i_b) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a) ORDER_INDEX(t1 i_a, i_b) ORDER_INDEX(t1 i_b) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint ORDER_INDEX(`t1` `i_a`,`i_b`) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`) ORDER_INDEX(`t1`@`select#1` `i_b`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# ORDER_INDEX(t1) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a) ORDER_INDEX(t1) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint ORDER_INDEX(`t1` ) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# ORDER_INDEX(t1 i_b) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+  ORDER_INDEX(t1 i_a, i_b) ORDER_INDEX(t1 i_b) INDEX(t1 i_c)*/ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint ORDER_INDEX(`t1` `i_b`) is ignored as conflicting/duplicated
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_c`) ORDER_INDEX(`t1`@`select#1` `i_a`,`i_b`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# INDEX(t1) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+  ORDER_INDEX(t1 i_a, i_b) GROUP_INDEX(t1 i_b) INDEX(t1)*/ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint INDEX(`t1` ) is ignored as conflicting/duplicated
+Note	1003	select /*+ GROUP_INDEX(`t1`@`select#1` `i_b`) ORDER_INDEX(`t1`@`select#1` `i_a`,`i_b`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# INDEX(t1) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+  ORDER_INDEX(t1) GROUP_INDEX(t1) INDEX(t1)*/ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint INDEX(`t1` ) is ignored as conflicting/duplicated
+Note	1003	select /*+ GROUP_INDEX(`t1`@`select#1`) ORDER_INDEX(`t1`@`select#1`) */ `test`.`t1`.`a` AS `a` from `test`.`t1`
+# Check the use of index hints.
+# Force the use of i_a, i_b indexes, intersect(i_a,i_b) is used.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a, i_b) */ a FROM t1 WHERE a = 1 AND b = 2 AND c = 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index_merge	i_a,i_b	i_a,i_b	5,5	NULL	2	0.39	Using intersect(i_a,i_b); Using where
+Warnings:
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`,`i_b`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` = 1 and `test`.`t1`.`b` = 2 and `test`.`t1`.`c` = 3
+# Force the use of i_a, i_ab indexes, i_ab index is used.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a, i_ab) */ a FROM t1 WHERE a = 1 AND b = 2 AND c = 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	ref	i_a,i_ab	i_ab	10	const,const	1	0.39	Using where
+Warnings:
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`,`i_ab`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` = 1 and `test`.`t1`.`b` = 2 and `test`.`t1`.`c` = 3
+# Force the use of i_a, i_b, i_c indexes, i_c index is used.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_a, i_b) JOIN_INDEX(t1 i_c) */ * FROM t1 WHERE a = 1 AND b = 2 AND c = 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	ref	i_a,i_b,i_c	i_c	5	const	1	1.03	Using where
+Warnings:
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_a`,`i_b`) JOIN_INDEX(`t1`@`select#1` `i_c`) */ `test`.`t1`.`a` AS `a`,`test`.`t1`.`b` AS `b`,`test`.`t1`.`c` AS `c`,`test`.`t1`.`d` AS `d` from `test`.`t1` where `test`.`t1`.`a` = 1 and `test`.`t1`.`b` = 2 and `test`.`t1`.`c` = 3
+# Test range index access
+EXPLAIN EXTENDED SELECT a FROM t1 WHERE a > 1 AND a < 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	range	i_a,i_ab	i_a	5	NULL	26	100.00	Using where; Using index
+Warnings:
+Note	1003	select `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` > 1 and `test`.`t1`.`a` < 3
+# Indexes are forbidden, full scan is employed
+EXPLAIN EXTENDED SELECT /*+ NO_INDEX(t1)*/a FROM t1 WHERE a > 1 AND a < 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	256	12.50	Using where
+Warnings:
+Note	1003	select /*+ NO_INDEX(`t1`@`select#1`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` > 1 and `test`.`t1`.`a` < 3
+EXPLAIN EXTENDED SELECT /*+ NO_JOIN_INDEX(t1)*/a FROM t1 WHERE a > 1 AND a < 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	12.50	Using where; Using index
+Warnings:
+Note	1003	select /*+ NO_JOIN_INDEX(`t1`@`select#1`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` > 1 and `test`.`t1`.`a` < 3
+# Indexes are forbidden for grouping and ordering but are usable for data access
+EXPLAIN EXTENDED SELECT /*+ NO_GROUP_INDEX(t1) NO_ORDER_INDEX(t1)*/a FROM t1 WHERE a > 1 AND a < 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	range	i_a,i_ab	i_a	5	NULL	26	100.00	Using where; Using index
+Warnings:
+Note	1003	select /*+ NO_GROUP_INDEX(`t1`@`select#1`) NO_ORDER_INDEX(`t1`@`select#1`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` > 1 and `test`.`t1`.`a` < 3
+# Index "i_a" is forbidden, "i_ab" is used instead
+EXPLAIN EXTENDED SELECT /*+ NO_INDEX(t1 i_a)*/a FROM t1 WHERE a > 1 AND a < 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	range	i_ab	i_ab	5	NULL	36	100.00	Using where; Using index
+Warnings:
+Note	1003	select /*+ NO_INDEX(`t1`@`select#1` `i_a`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` > 1 and `test`.`t1`.`a` < 3
+# Usable indexes are forbidden, full scan is employed
+EXPLAIN EXTENDED SELECT /*+ NO_INDEX(t1 i_a, i_ab)*/a FROM t1 WHERE a > 1 AND a < 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	256	12.50	Using where
+Warnings:
+Note	1003	select /*+ NO_INDEX(`t1`@`select#1` `i_a`,`i_ab`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` > 1 and `test`.`t1`.`a` < 3
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1 i_ab)*/a FROM t1 WHERE a > 1 AND a < 3;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	range	i_ab	i_ab	5	NULL	36	100.00	Using where; Using index
+Warnings:
+Note	1003	select /*+ INDEX(`t1`@`select#1` `i_ab`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` where `test`.`t1`.`a` > 1 and `test`.`t1`.`a` < 3
+# Ignore i_ab index, i_b index is used.
+EXPLAIN EXTENDED SELECT /*+ NO_INDEX(t1 i_ab) */ a, max(b) FROM t1 WHERE b = 2 GROUP BY a;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	ref	i_b	i_b	5	const	26	100.00	Using where; Using temporary; Using filesort
+Warnings:
+Note	1003	select /*+ NO_INDEX(`t1`@`select#1` `i_ab`) */ `test`.`t1`.`a` AS `a`,max(`test`.`t1`.`b`) AS `max(b)` from `test`.`t1` where `test`.`t1`.`b` = 2 group by `test`.`t1`.`a`
+# Ignore i_ab index, i_b index is used.
+EXPLAIN EXTENDED SELECT /*+ NO_JOIN_INDEX(t1 i_ab) */ a, max(b) FROM t1 WHERE b = 2 GROUP BY a;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	ref	i_b	i_b	5	const	26	100.00	Using where; Using temporary; Using filesort
+Warnings:
+Note	1003	select /*+ NO_JOIN_INDEX(`t1`@`select#1` `i_ab`) */ `test`.`t1`.`a` AS `a`,max(`test`.`t1`.`b`) AS `max(b)` from `test`.`t1` where `test`.`t1`.`b` = 2 group by `test`.`t1`.`a`
+# Force i_ab index for GROUP BY, i_ab index scan is used.
+EXPLAIN EXTENDED SELECT /*+ GROUP_INDEX(t1 i_ab) */ a, max(b) FROM t1 GROUP BY a;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_ab	10	NULL	256	100.00	Using index
+Warnings:
+Note	1003	select /*+ GROUP_INDEX(`t1`@`select#1` `i_ab`) */ `test`.`t1`.`a` AS `a`,max(`test`.`t1`.`b`) AS `max(b)` from `test`.`t1` group by `test`.`t1`.`a`
+# Force i_ab index for JOIN, i_ab loose index scan is used.
+EXPLAIN EXTENDED SELECT /*+ JOIN_INDEX(t1 i_ab) */ a, max(b) FROM t1 GROUP BY a;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_ab	10	NULL	256	100.00	Using index; Using temporary; Using filesort
+Warnings:
+Note	1003	select /*+ JOIN_INDEX(`t1`@`select#1` `i_ab`) */ `test`.`t1`.`a` AS `a`,max(`test`.`t1`.`b`) AS `max(b)` from `test`.`t1` group by `test`.`t1`.`a`
+# Ignore i_ab for for sorting rows. i_a index is used for sorting.
+EXPLAIN EXTENDED SELECT /*+ NO_ORDER_INDEX(t1 i_ab) */ a FROM t1 ORDER BY a;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Note	1003	select /*+ NO_ORDER_INDEX(`t1`@`select#1` `i_ab`) */ `test`.`t1`.`a` AS `a` from `test`.`t1` order by `test`.`t1`.`a`
+# Ignore i_a for for sorting rows. i_ab is used for sorting.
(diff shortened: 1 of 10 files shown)


mysql-test/main/opt_hints_index.test
```diff
@@ -0,0 +1,105 @@
...
+
+--echo # Ignore i_ab index, i_b index is used.
+EXPLAIN EXTENDED SELECT /*+ NO_INDEX(t1 i_ab) */ a, max(b) FROM t1 WHERE b = 2 GROUP BY a;
+--echo # Ignore i_ab index, i_b index is used.
+EXPLAIN EXTENDED SELECT /*+ NO_JOIN_INDEX(t1 i_ab) */ a, max(b) FROM t1 WHERE b = 2 GROUP BY a;
+--echo # Force i_ab index for GROUP BY, i_ab index scan is used.
+EXPLAIN EXTENDED SELECT /*+ GROUP_INDEX(t1 i_ab) */ a, max(b) FROM t1 GROUP BY a;
+--echo # Force i_ab index for JOIN, i_ab loose index scan is used.
+EXPLAIN EXTENDED SELECT /*+ JOIN_INDEX(t1 i_ab) */ a, max(b) FROM t1 GROUP BY a;
+--echo # Ignore i_ab for for sorting rows. i_a index is used for sorting.
```
mariadb-YuchenPei: s/for for/for, same below

sql/mem_root_array.h
```diff
@@ -44,7 +45,9 @@
           compilers we use.
 */
 
-template<typename Element_type, bool has_trivial_destructor>
+template<typename Element_type,
+         bool has_trivial_destructor=
+                     std::is_trivially_destructible<Element_type>::value>
```
mariadb-YuchenPei, line 50: Question: why does `std::is_trivially_destructible<Opt_hints*>::value` resolve to true but `std::is_trivially_destructible<Parser::Join_order_hint*>::value` resolve to false?

storage/maria/ha_maria.cc
```diff
@@ -1838,8 +1839,17 @@ int ha_maria::assign_to_keycache(THD * thd, HA_CHECK_OPT *check_opt)
...
+  if (!table_list->opt_hints_table ||
+      !table_list->opt_hints_table->update_index_hint_maps(thd, table))
+  {
+    /*
+      Old-style index hints are processed only if there
+      new-style hints are not specified
+    */
+    if (table_list->process_index_hints(table))
+      DBUG_RETURN(HA_ADMIN_FAILED);
+  }
```
mariadb-YuchenPei: This block appears 5 times (including one above with a return of 1) - can you factor it into a function?

sql/sql_base.cc
```diff
@@ -8363,12 +8372,22 @@ bool setup_tables(THD *thd, Name_resolution_context *context,
...
+        table->map= table_list->map_exec;
+        table->maybe_null= table_list->maybe_null_exec;
+        table->pos_in_table_list= table_list;
+
+        if (!table_list->opt_hints_table ||
+            !table_list->opt_hints_table->update_index_hint_maps(thd, table))
+        {
+          /*
+            Old-style index hints are processed only if there
+            new-style hints are not specified
```
mariadb-YuchenPei: Remove "there"

sql/opt_hints.h
```diff
@@ -563,14 +587,83 @@ class Opt_hints_qb : public Opt_hints
...
+  Auxiliary class for compound key objects.
+  Used for processing INDEX, JOIN_INDEX, GROUP_INDEX and ORDER_INDEX hints.
+
+  Represents a bitmap of keys specified in the hint along with methods for
+  handling it.
+  The class cannot be instantiated, can only be used as a base class
+  for derived ones, see `Index_key_hint`, `Global_index_key_hint`
+*/
+
+class Compound_key_hint : public Sql_alloc
```
mariadb-YuchenPei: Why is `Compound_key_hint` separate from the class hierarchy of `Opt_hints_*`? For example, why is it not defined as a descendant class of `Opt_hints` like the other `Opt_hints_*` classes?

sql/opt_hints.cc
```diff
@@ -517,6 +518,209 @@ bool Opt_hints_table::fix_hint(TABLE *table)
...
+  if (!is_fixed(INDEX_HINT_ENUM) && !is_fixed(JOIN_INDEX_HINT_ENUM) &&
+      !is_fixed(GROUP_INDEX_HINT_ENUM) && !is_fixed(ORDER_INDEX_HINT_ENUM))
+    return false;  // No index hint is specified
+
+  Key_map usable_index_map(tbl->s->usable_indexes(thd));
+  tbl->keys_in_use_for_query= tbl->keys_in_use_for_group_by=
+      tbl->keys_in_use_for_order_by= usable_index_map;
+
+  const bool force_index= is_force_index_hint(INDEX_HINT_ENUM);
+  tbl->force_index= (force_index || is_force_index_hint(JOIN_INDEX_HINT_ENUM));
```
mariadb-YuchenPei: Shouldn't the left hand side be tbl->force_index_join?

mariadb-YuchenPei:
It would be good to provide some further description of the implemented hints in the commit message. For example, do the hints not prefixed by "NO_" have the same effect with FORCE INDEX, or USE INDEX?

sql/opt_hints_parser.h
```diff
@@ -383,7 +396,15 @@ class Optimizer_hint_parser: public Optimizer_hint_tokenizer,
...
-             id == TokenID::keyword_NO_MRR;
+             id == TokenID::keyword_NO_MRR ||
+             id == TokenID::keyword_INDEX ||
+             id == TokenID::keyword_NO_INDEX ||
+             id == TokenID::keyword_JOIN_INDEX ||
+             id == TokenID::keyword_NO_JOIN_INDEX ||
+             id == TokenID::keyword_ORDER_INDEX ||
+             id == TokenID::keyword_NO_ORDER_INDEX ||
+             id == TokenID::keyword_GROUP_INDEX ||
+             id == TokenID::keyword_NO_GROUP_INDEX;
```
mariadb-YuchenPei, line 411: Update the comment above "// index_level_hint_type ::= ...." to reflect the new additions here

sql/opt_hints.h
mariadb-YuchenPei: Relatedly, there are multiple places indicating these new hints are index level (Index_level_hint_type_cond::allowed_token_id, Parser::Index_level_hint::resolve) - why not have them as part of Opt_hints_key? why call them compound not index level?

sql/opt_hints.cc
```diff
@@ -992,6 +1196,51 @@ bool Opt_hints_global::fix_hint(THD *thd)
...
+  already specified JOIN_INDEX, GROUP_INDEX, ORDER_INDEX
+  hints.
+
+  @param table_hint         pointer to table hint
+  @param key_hint           pointer to key hint
+
+  @return false if no conflict, true otherwise.
+*/
+
+bool Global_index_key_hint::is_hint_conflicting(Opt_hints_table *table_hint,
```
mariadb-YuchenPei: This function does not depend on fields of the Global_index_key_hint object, why not make it static member function to reduce noise, or better yet, add a bool parameter is_global and make it a free standing function.

Ditto for Index_key_hint::is_hint_conflicting

sql/opt_hints.cc
```diff
@@ -992,6 +1196,51 @@ bool Opt_hints_global::fix_hint(THD *thd)
 }
 
 
+/**
+  Function checks if INDEX hint is conflicting with
+  already specified JOIN_INDEX, GROUP_INDEX, ORDER_INDEX
+  hints.
```
mariadb-YuchenPei: I find the code comment here a bit confusing.

How about something like this:

```
Function checks if an INDEX hint conflicts with
any JOIN_INDEX, GROUP_INDEX, ORDER_INDEX
hints, by checking if any of the latter is already 
specified at table level or index level.
```

storage/myisam/ha_myisam.cc
```diff
@@ -1455,8 +1456,17 @@ int ha_myisam::assign_to_keycache(THD* thd, HA_CHECK_OPT *check_opt)
 
   table->keys_in_use_for_query.clear_all();
 
-  if (table_list->process_index_hints(table))
-    DBUG_RETURN(HA_ADMIN_FAILED);
+  if (!table_list->opt_hints_table ||
```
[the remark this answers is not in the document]
Olernov: OK, reverted the changes

mysql-test/main/opt_hints_index.test
```diff
@@ -0,0 +1,105 @@
...
+INSERT INTO t1 SELECT a,b, c + 40, d FROM t1;
+INSERT INTO t1 SELECT a,b, c + 80, d FROM t1;
+INSERT INTO t1 SELECT a,b, c + 160, d FROM t1;
+ANALYZE TABLE t1;
+
+--echo # Check behavior of duplicated/intersected hints.
+--echo # First specified hint is applied and next conflicting/intersected hints
+--echo # are ignored with warning.
+
+--echo # JOIN_INDEX(t1) is ignored as intersected.
```
mariadb-YuchenPei: It would be nice to explain "intersected". For example (see my other comment summarising conflict checking):

```
--echo # JOIN_INDEX(t1) is ignored as intersected (opposite kind same table)
```

sql/opt_hints.cc
```diff
@@ -517,6 +518,209 @@ bool Opt_hints_table::fix_hint(TABLE *table)
 }
 
 
+bool Opt_hints_table::is_hint_conflicting(Opt_hints_key *key_hint,
```
mariadb-YuchenPei: Can you add some comment documentation for this function? For example:

```
Check whether a hint of type is already specified. 
Check at table level if key_hint is NULL, otherwise check at index level.
```

In fact given the symmetry between `this` and `key_hint` for this function, it would be much more readable if it were a freestanding function, like so:

```
static bool table_or_key_hint_type_specified(Opt_hints_table *table_hint, Opt_hints_key *key_hint,
                                          opt_hints_enum type) 
{
  assert(table_hint != NULL || key_hint != NULL);
  return key_hint ? key_hint->is_specified(type) : table_hint->is_specified(type);
}
```

sql/opt_hints.cc
```diff
@@ -517,6 +518,209 @@ bool Opt_hints_table::fix_hint(TABLE *table)
...
+      }
+    }
+  }
+}
+
+
+/**
+  Function updates keys_in_use_for_query, keys_in_use_for_group_by,
+  keys_in_use_for_order_by depending on INDEX, JOIN_INDEX, GROUP_INDEX,
+  ORDER_INDEX hints.
```
mariadb-YuchenPei: Suggest comment describing at higher level and more concretely what this function does:

```
For each index hint that is not ignored, include the index in
    - tbl->keys_in_use_for_query if the hint is INDEX or JOIN_INDEX
    - tbl->keys_in_use_for_group_by if the hint is INDEX or
      GROUP_INDEX
    - tbl->keys_in_use_for_order_by if the hint is INDEX or
      ORDER_INDEX
    conversely, subtract the index from the corresponding
    tbl->keys_in_use_for_... map if the hint is prefixed with NO_.
```

mysql-test/main/opt_hints_index.result
```diff
@@ -0,0 +1,249 @@
...
+test.t1	analyze	status	OK
+# Check behavior of duplicated/intersected hints.
+# First specified hint is applied and next conflicting/intersected hints
+# are ignored with warning.
+# JOIN_INDEX(t1) is ignored as intersected.
+EXPLAIN EXTENDED SELECT /*+ INDEX(t1) JOIN_INDEX(t1) */ a FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
+1	SIMPLE	t1	index	NULL	i_a	5	NULL	256	100.00	Using index
+Warnings:
+Warning	4219	Hint JOIN_INDEX(`t1` ) is ignored as conflicting/duplicated
```
mariadb-YuchenPei: It would be nice to include a bit of explanation how JOIN_INDEX(`t1` ) conflicts with previous hints in the warning, something like:

```
An INDEX hint was already specified for the same table.
```

sql/opt_hints_parser.cc
```diff
@@ -458,37 +514,96 @@ bool Parser::Index_level_hint::resolve(Parse_context *pc) const
   if (!tab)
     return false;
 
-  if (is_empty())  // Table level hint
```
mariadb-YuchenPei, line 461: Please include some comments describing exactly what constitutes conflicts.

Here's my proposal:

```
Taxonomy of index hints

- 2 levels of hints:
  - table level hints: no index names, i.e. for a table
  - index level hints: with index names, i.e. for an index
- 2 kinds of hints:
  - global: [NO_]INDEX
  - non-global: [NO_]JOIN_INDEX, [NO_]GROUP_INDEX, [NO_]ORDER_INDEX
- 4 types of hints:
  - [NO_]JOIN_INDEX
  - [NO_]GROUP_INDEX
  - [NO_]ORDER_INDEX
  - [NO_]INDEX

Conflict checking

- A conflict happens iff
  - for a table level hint
    - a hint of the opposite kind has already been specified for the
      same table
  - for a index level hint
    - the same type of hint has already been specified for the same
      table or for the same index, OR
    - the opposite kind of hint has already been specified for the
      same index
- For a multi index hint like JOIN_INDEX(t1 i1, i2, i3), it conflicts
  with a previous hint if any of the JOIN_INDEX(t1 i1), JOIN_INDEX(t1
  i2), JOIN_INDEX(t1 i3) conflicts with a previous hint
- When a hint type is specified for an index, it is also marked as
  specified with the same switch state for the table
```

sql/opt_hints_parser.cc
```diff
@@ -458,37 +514,96 @@ bool Parser::Index_level_hint::resolve(Parse_context *pc) const
...
-    Opt_hints_key *idx= (Opt_hints_key *)tab->find_by_name(index_name_sys);
-    if (!idx)
+    Opt_hints_key *key= (Opt_hints_key *)tab->find_by_name(index_name_sys);
+    if (!key)
     {
-      idx= new (pc->thd->mem_root)
+      key= new (pc->thd->mem_root)
         Opt_hints_key(index_name_sys, tab, pc->thd->mem_root);
-      tab->register_child(idx);
+      tab->register_child(key);
```
mariadb-YuchenPei: Shouldn't the child be de-registered if the loop is aborted in a subsequent iteration due to conflicts?

mariadb-YuchenPei (changes requested):

mariadb-YuchenPei:
> Good idea, I expanded the commit message in a force-pushed commit.

Thanks. Where can I find this new commit?

> I think it's too much for a commit message to specify the rules of conflicting but I agree it should be stated somewhere, probably in JIRA.

How about something like the description of the conflicting rules in the following comment?

https://github.com/MariaDB/server/pull/4058#discussion_r2120204671

Olernov:
> > Good idea, I expanded the commit message in a force-pushed commit.
> 
> Thanks. Where can I find this new commit?
> 
Oh, I pushed it to a wrong branch initially. Now it's here.

> > I think it's too much for a commit message to specify the rules of conflicting but I agree it should be stated somewhere, probably in JIRA.
> 
> How about something like the description of the conflicting rules in the following comment?
> 
> [#4058 (comment)](https://github.com/MariaDB/server/pull/4058#discussion_r2120204671)

Great comment, what do you think is the best place to put it? The expanded commit message is already large...

sql/mem_root_array.h
Olernov: I guess it's because `Join_order_hint` is inherited from a class having a virtual destructor (`Printable_parser_rule`), similarly to [what is described here](https://stackoverflow.com/questions/39528522/is-trivially-destructible-does-not-work-with-inherited-class) .  Can you check any other class inherited from `Opt_hints` , say `Opt_hints_table`? If this suggestion is correct then it won't be trivially destructible as `Opt_hints` also has a virtual destructor

storage/maria/ha_maria.cc
Olernov: No need any more, 4 of those occurrences were eliminated.

sql/opt_hints.h
Olernov: Those are valid questions. In fact, initially `Opt_hints_key` for handling NO_ICP, MRR/NO_MRR and NO_RANGE_OPTIMIZATION hints probably wasn't designed correctly. The parser split `INDEX_LEVEL_HINT(t1, key1, key2, key3)` into 3 separate independent objects of `Opt_hints_key`. However, the new hints (JOIN_INDEX, ORDER_INDEX and so on) could not be handled this way, and `Compound_key_hint` was introduced to act as some kind of proxy.

sql/mem_root_array.h
mariadb-YuchenPei: Exactly, I read the definition at cplusplus too, and indeed Opt_hints having a virtual destructor - so why does it resolve to true?

If you do `b Mem_root_array` in a debugger then it tells you the value of each resolution:

```
  2 breakpoint keep   y         <MULTIPLE>    0 
2.1                   y 0x0000563cb7988e3a      in Mem_root_array<SEL_ARG*, true>::Mem_root_array(st_mem_root*, unsigned long, SEL_ARG* const&) of /home/ycp/source/mariadb-server/push-main/src/sql/mem_root_array.h:63
2.2                   y 0x0000563cb7c3dc7a      in Mem_root_array<st_order*, true>::Mem_root_array(st_mem_root*) of /home/ycp/source/mariadb-server/push-main/src/sql/mem_root_array.h:58
2.3                   y 0x0000563cb83f8722      in Mem_root_array<Opt_hints*, true>::Mem_root_array(st_mem_root*) of /home/ycp/source/mariadb-server/push-main/src/sql/mem_root_array.h:58
2.4                   y 0x0000563cb83f8c6c      in Mem_root_array<Opt_hints_key*, true>::Mem_root_array(st_mem_root*) of /home/ycp/source/mariadb-server/push-main/src/sql/mem_root_array.h:58
2.5                   y 0x0000563cb840537c      in Mem_root_array<Optimizer_hint_parser::Join_order_hint*, true>::Mem_root_array(st_mem_root*) of /home/ycp/source/mariadb-server/push-main/src/sql/mem_root_array.h:58
2.6                   y 0x0000563cb85de702      in Mem_root_array<Alter_inplace_info::Rename_key_pair, true>::Mem_root_array(st_mem_root*) of /home/ycp/source/mariadb-server/push-main/src/sql/mem_root_array.h:58
2.7                   y 0x0000563cb8654a20      in Mem_root_array<Item_param*, true>::Mem_root_array(st_mem_root*) of /home/ycp/source/mariadb-server/push-main/src/sql/mem_root_array.h:58
```

There's no Opt_hints_table probably because no fields or variables of `Mem_root_array<Opt_hints_table*>` is declared. But it resolves to true for both Opt_hints and Opt_hints_key.

mariadb-YuchenPei:
> Great comment, what do you think is the best place to put it? The expanded commit message is already large...

I think the best place to put it would be in the code in the resolve function where my comment is, or as documentation for that function.

But it would also be good to have maybe a brief description of the conflicting rules in the commit message.

sql/sql_base.cc
Olernov: OK, will remove

sql/opt_hints.cc
Olernov: Good catch, I think it should

sql/opt_hints_parser.h
Olernov: OK, will update

sql/opt_hints.cc
Olernov: OK, will replace with your version

mysql-test/main/opt_hints_index.test
Olernov: OK, will do

sql/opt_hints.cc
Olernov: Nice improvement, will update

sql/opt_hints_parser.cc
Olernov: OK, will add after slight editing

sql/opt_hints.cc
Olernov: virtual functions cannot be static, and I don't like the idea of adding the new `is_global` parameter, as it would reduce readability and would require additional comments about what this new parameter is for. I believe the current implementation is a good practice of using virtual functions: we have descendant classes which differ in behavior and we cover this difference with polymorphism.

Olernov: On the other hand, there is `Opt_hints_table::is_hint_conflicting()` which is not virtual, and this looks messy. Maybe a templated implementation could be a solution?

mariadb-YuchenPei: I meant something like this:

```
-         tab->get_compound_key_hint(hint_type)->is_hint_conflicting(tab,
-                                                                    nullptr)) ||
+         is_index_hint_conflicting(tab, nullptr, hint_type)) ||
```

This improves readability because methods imply involvement of object states. When I see `tab->get_compound_key_hint(hint_type)` I wonder what of `tab` is used for this call. When I see `tab->get_compound_key_hint(hint_type)->is_hint_conflicting(tab, nullptr))` I further wonder what of `tab->get_compound_key_hint(hint_type)` is used in the method `Some_class::get_compound_key_hint` (I also don't know what is `Some_class` without checking what tab->get_compound_key_hint(...) returns). But when I see `is_index_hint_conflicting(tab, nullptr, hint_type)` I know it is a function that only relies on two items: tab and hint_type, and nothing else. No polymorphism and no subclasses. It takes less mental space to understand.

See this commit (in the branch bb-12.0-MDEV-35856-index-hints-v2-ycp-cleanup) for the full (demo) proposal:

```
30de1c13598af2d1beed6facd3c24ef6e2e2790e
MDEV-35856 Clean up index hint conflicting check

Change
bool Opt_hints_table::is_hint_conflicting,
Global_index_key_hint::is_hint_conflicting, and
Index_key_hint::is_hint_conflicting

to freestanding functions.
```

Interestingly, after this change, we can get rid of the subclasses:

```
upstream/bb-12.0-MDEV-35856-index-hints-v2-ycp-cleanup 6aeb8b3697ef8f4047de9e867d36dd85036152af
MDEV-35856 Remove Index_key_hint and Global_key_hint

Use Compound_key_hint directly.
```

What do you think?

Olernov:
> > Great comment, what do you think is the best place to put it? The expanded commit message is already large...
> 
> I think the best place to put it would be in the code in the resolve function where my comment is, or as documentation for that function.
> 
> But it would also be good to have maybe a brief description of the conflicting rules in the commit message.

OK, added a brief description to the commit message.

mysql-test/main/opt_hints_index.result
Olernov: Pushed it as a separate commit `MDEV-35856 expand warning messages for conflicting index hints`. Does it look better now ?

sql/opt_hints.cc
Olernov: Indeed, very nice simplification!

sql/opt_hints.cc
Olernov: After the simplication you suggested I guess it's not relevant any more

sql/opt_hints_parser.cc
Olernov: Sure, I think it's better to register the child after the checks for conflicts. Fixed

mysql-test/main/opt_hints_index.test
mariadb-YuchenPei: looks like it's fixed in the current commit c2a428dbe255f4d1cf083b4ca436060821ce8451

sql/mem_root_array.h
mariadb-YuchenPei: But the list does not contain `Parser::Join_order_hint*`, strange...

sql/opt_hints.cc
mariadb-YuchenPei: Thanks. However I don't see any changes to tests in the commit (5e61e76b9b37b93bd391339268d36ead4efd569c) that updates this - can you add test coverage for this issue?

sql/opt_hints_parser.h
mariadb-YuchenPei: I can see it fixed at the current patch.

sql/opt_hints.h
```diff
@@ -59,6 +59,8 @@
       Opt_hints_qb
         Opt_hints_table
           Opt_hints_key
+          Index_key_hint
+          Global_index_key_hint
```
mariadb-YuchenPei: These two no longer exist. Change to `Compound_hint`

sql/opt_hints.h
mariadb-YuchenPei: With the cleanup that leaves only Compound_key_hint and no subclasses, does this reasoning still hold?

Also, given the subclasses are gone, can we rename Compound_key_hint to something more descriptive? Like Opt_hints_use_key?

sql/sql_base.cc
mariadb-YuchenPei: "there" still there?

sql/opt_hints_parser.cc
mariadb-YuchenPei: It still looks like a child can be registered before the loop aborts for the same hint:

INDEX(t i1, i2)

i1 not conflicting, register
i2 conflicting, break

i1 is still registered

mariadb-YuchenPei:
> OK, added a brief description to the commit message.

Thanks. Looks good. Can you add one sentence explaining "kind" which is not obvious, like "we say there are two kinds of index, the global kind [NO_]INDEX and the others being the non-global kind"

sql/opt_hints_parser.cc
```diff
@@ -411,10 +436,39 @@ bool Parser::Table_level_hint::resolve(Parse_context *pc) const
...
+  - 4 types of hints:
+    - [NO_]JOIN_INDEX
+    - [NO_]GROUP_INDEX
+    - [NO_]ORDER_INDEX
+    - [NO_]INDEX
+
+  Conflict checking
+  - A conflict happens if and only if
+    - for a table level hint
+      - a hint of the same or opposite kind has already been specified for the
```
mariadb-YuchenPei: Should be "a hint of the same type or opposite kind has already been specified for the same table"

Please also fix it in the commit message

sql/opt_hints_parser.cc
```diff
@@ -458,37 +544,106 @@ bool Parser::Index_level_hint::resolve(Parse_context *pc) const
...
+  if (is_empty())  // Empty list of index names, i.e. it is a table level hint
   {
-    if (tab->set_switch(hint_state, hint_type, false))
+    if ((is_compound_hint(hint_type) &&
+           is_index_hint_conflicting(tab, nullptr, hint_type)) ||
+         tab->set_switch(hint_state, hint_type, false))
     {
       print_warn(pc->thd, ER_WARN_CONFLICTING_HINT, hint_type, hint_state,
-                 &qb_name_sys, &table_name_sys, nullptr, nullptr);
+                 &qb_name_sys, &table_name_sys, nullptr, this, &table_conflict);
```
mariadb-YuchenPei: The message is not precise enough and can confuse users because `JOIN_INDEX(t) GROUP_INDEX(t)` does not trigger it. It should be "an index hint of the same type or opposite kind was already specified for this table".

sql/opt_hints_parser.cc
```diff
@@ -458,37 +544,106 @@ bool Parser::Index_level_hint::resolve(Parse_context *pc) const
...
+    {
+      bool is_specified= tab->is_specified(hint_type) ||
+                         key->is_specified(hint_type);
+      if (is_specified || is_index_hint_conflicting(tab, key, hint_type))
+      {
+        is_conflicting= true;
+        print_warn(pc->thd, ER_WARN_CONFLICTING_HINT,hint_type, hint_state,
+                   &qb_name_sys, &table_name_sys, nullptr, this,
+                   tab->is_specified(hint_type) ? &table_conflict :
+                                                  &key_conflict);
```
mariadb-YuchenPei: Like above, the messages here are not accurate. As noted there are two possibilities

If is_specified is true, the warning message should be like "the same type of hint has already been specified for the same table or for the same index"

Otherwise the message should be like "the opposite kind of hint has already been specified for the same index"

Vague messages like "another hint was already specified for this index" are confusing - people might ask "how come `JOIN_INDEX(t i) GROUP_INDEX(t i)` is ok?"

mysql-test/main/opt_hints_index.test
mariadb-YuchenPei: I'll check this after the new review comments about warning messages are addressed.

mysql-test/main/opt_hints_index.result
mariadb-YuchenPei: Yes it looks better now, however they can still be a bit confusing - see my new comments at the warning messages. Resolving this thread meanwhile.

sql/mem_root_array.h
Olernov: Maybe multiple inheritance makes `Join_order_hint` non-trivially destructible. `Opt_hints_*` classes are inherited from a single class and probably the compiler manages to make them trivially destructible despite having virtual destructors

sql/sql_base.cc
Olernov: Oh yes, initially removed them in another branch. Now fixed here

sql/opt_hints_parser.cc
Olernov: I just tried to make the messages more concise. Otherwise we'll end up with warnings like
`Hint NO_ORDER_INDEX(`t1` `i_b`) is ignored as conflicting/duplicated (an index hint of the same type or opposite kind was already specified for this table)`
Do we print such long warnings anywhere else? I thought they are usually not so verbose

sql/opt_hints.h
Olernov: You mean `Compound_key_hint` I think? OK, will do

Olernov:
> > OK, added a brief description to the commit message.
> 
> Thanks. Looks good. Can you add one sentence explaining "kind" which is not obvious, like "we say there are two kinds of index, the global kind [NO_]INDEX and the others being the non-global kind"

OK, will do

sql/opt_hints_parser.cc
Olernov: You're right, unlike simple index hints compound ones are either accepted or rejected as a whole. Will fix.

sql/opt_hints_parser.cc
Olernov: "The opposite kind of hint" may also be confusing for a user. Basically, we can print 3 different messages here without complicating the logic, let's agree on their texts:
`is_specified == true`:
 -  `tab->is_specified(hint_type) == true`:  "the same type of hint was already specified for this table"
 -  `key->is_specified(hint_type) == true`:  "the same type of hint was already specified for this index"

`is_index_hint_conflicting(tab, key, hint_type))==true`:   ??

sql/opt_hints.cc
Olernov: I couldn't yet construct a test case for that, probably because `TABLE::force_index` and `TABLE::force_index_join` are somewhat equivalent, see this comment:
```
TABLE_LIST::process_index_hints()
....
    /*
      TODO: get rid of tbl->force_index (on if any FORCE INDEX is specified)
      Use the correct force_index_XX in all places instead of the global one.
    */
    tbl->force_index= (tbl->force_index_order | tbl->force_index_group |
                       tbl->force_index_join);

```
and `force_join_index` is a recent addition. I'll take a second look whether we have test to make sure `JOIN_INDEX` doesn't act like `GROUP_INDEX` or 'ORDER_INDEX`

sql/opt_hints_parser.cc
mariadb-YuchenPei: looks like it's fixed now at 76c7564e07576cb41cd2dd3b9b5f432daa34888a, thank you

sql/opt_hints_parser.cc
mariadb-YuchenPei: I don't know about long warnings anywhere else. Maybe you are right and we should just stick with "Hint NO_ORDER_INDEX(t1 i_b) is ignored as conflicting/duplicated" and refer the user to the documentation for the rules.

sql/opt_hints_parser.cc
mariadb-YuchenPei: True. Users will need to refer to documentation to understand what "kind" means here. Either way users will need to consult the documentation. Maybe the original messages like in https://github.com/MariaDB/server/pull/4058#discussion_r2131450274?

sql/opt_hints_parser.cc
Olernov: Maybe @spetrunia has an opinion on this topic?

sql/opt_hints_parser.cc
Olernov: It would be great if we could print to the exact hint causing the conflict. However, this is not trivial to implement

sql/opt_hints.cc
Olernov: Once again I revisited the code in `Opt_hints_table::update_index_hint_maps()` and found it not always accurate. I fixed the inaccuracies, and after that a couple of test cases from `opt_hints_index` began to execute more correctly. I applied changes to the commit "MDEV-35856: Address code review comments 2"

spetrunia:
Just an observation about different behavior from MySQL. Not sure if this should be fixed or not:

```sql
create table ten(a int primary key);
insert into ten values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);

create table one_k(a int primary key);
insert into one_k select A.a + B.a* 10 + C.a * 100 from ten A, ten B, ten C;

create table t1 (a int, b int, c int, index i1(a), index i2(b));
insert into t1 select a,a,a from one_k;
```
By default, both MariaDB and MySQL use full table scan:
```
explain select  * from t1 where a <700 and b<700;
MariaDB [j1]> explain select  * from t1 where a <700 and b<700;
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows | Extra       |
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
|    1 | SIMPLE      | t1    | ALL  | i1,i2         | NULL | NULL    | NULL | 1000 | Using where |
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
1 row in set (0.001 sec)
```

Here MariaDB will still use full table scan:
```
MariaDB [j1]> explain select  /*+ INDEX(t1) */ * from t1 where a <700 and b<700;
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows | Extra       |
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
|    1 | SIMPLE      | t1    | ALL  | i1,i2         | NULL | NULL    | NULL | 1000 | Using where |
+------+-------------+-------+------+---------------+------+---------+------+------+-------------+
1 row in set (0.001 sec)
```

While MySQL would use range access:
```
mysql> explain select  /*+ INDEX(t1) */ * from t1 where a <700 and b<700;
+----+-------------+-------+------------+-------+---------------+------+---------+------+------+----------+------------------------------------+
| id | select_type | table | partitions | type  | possible_keys | key  | key_len | ref  | rows | filtered | Extra                              |
+----+-------------+-------+------------+-------+---------------+------+---------+------+------+----------+------------------------------------+
|  1 | SIMPLE      | t1    | NULL       | range | i1,i2         | i1   | 5       | NULL |  700 |    70.00 | Using index condition; Using where |
+----+-------------+-------+------------+-------+---------------+------+---------+------+------+----------+------------------------------------+
1 row in set, 1 warning (0,00 sec)
```

spetrunia:
```cpp
    const Lex_ident_sys table_conflict(
        STRING_WITH_LEN("an index hint was already specified for this table"));
    const Lex_ident_sys key_conflict(
        STRING_WITH_LEN("another hint was already specified for this index"));
```
It is not a good idea to have strings in the code as this prevents error message translation.  The start of the error message will get translated, the end of the message will remain in English.

spetrunia:
... note to self:  NO_JOIN_INDEX hint would conflict with itself even if specified for two different indexes. It's the same in MySQL: 

```
mysql> explain select  /*+ NO_JOIN_INDEX(t1 i1) NO_JOIN_INDEX(t1 i2) */ * from t1 where a <7 and b<700;
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra       |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
|  1 | SIMPLE      | t1    | NULL       | ALL  | i2            | NULL | NULL    | NULL | 1000 |    23.33 | Using where |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+-------------+
1 row in set, 2 warnings (0,00 sec)
Warning (Code 3126): Hint NO_JOIN_INDEX(`t1`  `i2`) is ignored as conflicting/duplicated
Note (Code 1003): /* select#1 */ select /*+ NO_JOIN_INDEX(`t1`@`select#1` `i1`) */ `j22`.`t1`.`a` AS `a`,`j22`.`t1`.`b` AS `b`,`j22`.`t1`.`c` AS `c` from `j22`.`t1` where ((`j22`.`t1`.`a` < 7) and (`j22`.`t1`.`b` < 700))
```
Other hints do not conflict with themselves:
```sql
explain extended select  /*+ NO_MRR(t1 i1) NO_MRR(t1 i2) */ * from t1 where a <7 and b<700;
explain extended select /*+ JOIN_ORDER(t1, t2)  JOIN_ORDER(t2,t3) */ * from t1,t2,t3 ;
```

spetrunia:
Looking at Opt_hints::print(), I see two checks
* `opt_hint_info[i].irregular_hint`
* `ignore_print(hint)`

Do they achieve the same goal? Can we use just one instead of two?

Olernov:
> Just an observation about different behavior from MySQL. Not sure if this should be fixed or not:
> 
> ```sql
> create table ten(a int primary key);
> insert into ten values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
> 
> create table one_k(a int primary key);
> insert into one_k select A.a + B.a* 10 + C.a * 100 from ten A, ten B, ten C;
> 
> create table t1 (a int, b int, c int, index i1(a), index i2(b));
> insert into t1 select a,a,a from one_k;
> ```
> 
> By default, both MariaDB and MySQL use full table scan:
> 
> ```
> explain select  * from t1 where a <700 and b<700;
> MariaDB [j1]> explain select  * from t1 where a <700 and b<700;
> +------+-------------+-------+------+---------------+------+---------+------+------+-------------+
> | id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows | Extra       |
> +------+-------------+-------+------+---------------+------+---------+------+------+-------------+
> |    1 | SIMPLE      | t1    | ALL  | i1,i2         | NULL | NULL    | NULL | 1000 | Using where |
> +------+-------------+-------+------+---------------+------+---------+------+------+-------------+
> 1 row in set (0.001 sec)
> ```
> 
> Here MariaDB will still use full table scan:
> 
> ```
> MariaDB [j1]> explain select  /*+ INDEX(t1) */ * from t1 where a <700 and b<700;
> +------+-------------+-------+------+---------------+------+---------+------+------+-------------+
> | id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows | Extra       |
> +------+-------------+-------+------+---------------+------+---------+------+------+-------------+
> |    1 | SIMPLE      | t1    | ALL  | i1,i2         | NULL | NULL    | NULL | 1000 | Using where |
> +------+-------------+-------+------+---------------+------+---------+------+------+-------------+
> 1 row in set (0.001 sec)
> ```
> 
> While MySQL would use range access:
> 
> ```
> mysql> explain select  /*+ INDEX(t1) */ * from t1 where a <700 and b<700;
> +----+-------------+-------+------------+-------+---------------+------+---------+------+------+----------+------------------------------------+
> | id | select_type | table | partitions | type  | possible_keys | key  | key_len | ref  | rows | filtered | Extra                              |
> +----+-------------+-------+------------+-------+---------------+------+---------+------+------+----------+------------------------------------+
> |  1 | SIMPLE      | t1    | NULL       | range | i1,i2         | i1   | 5       | NULL |  700 |    70.00 | Using index condition; Using where |
> +----+-------------+-------+------------+-------+---------------+------+---------+------+------+----------+------------------------------------+
> 1 row in set, 1 warning (0,00 sec)
> ```

Yes, it's an error. I didn't copy this piece of code from `TABLE_LIST::process_index_hints()` as I thought this `TODO` is already done.
```
    /*
      TODO: get rid of tbl->force_index (on if any FORCE INDEX is specified)
      Use the correct force_index_XX in all places instead of the global one.
    */
    tbl->force_index= (tbl->force_index_order | tbl->force_index_group |
                       tbl->force_index_join);
```
However, this code is still required. Will fix it

Olernov:
> ```c++
>     const Lex_ident_sys table_conflict(
>         STRING_WITH_LEN("an index hint was already specified for this table"));
>     const Lex_ident_sys key_conflict(
>         STRING_WITH_LEN("another hint was already specified for this index"));
> ```
> 
> It is not a good idea to have strings in the code as this prevents error message translation. The start of the error message will get translated, the end of the message will remain in English.

Should we then spawn multiple instances of messages for the duplication/conflict warnings?   There's  [a discussion about that](https://github.com/MariaDB/server/pull/4058#discussion_r2127897719).   Yuchen doesn't insist on expanding warnings messages as it's not clear what is the reasonable length should be and how to avoid confusion about the terminology.

sql/opt_hints_parser.cc
Olernov: Expanded warning messages in the recent commit f380225f7f95203a7d97744b0a8877f22cf58de9.

sql/opt_hints_parser.cc
Olernov: Expanded warning messages in the recent commit f380225f7f95203a7d97744b0a8877f22cf58de9.

sql/opt_hints.h
Olernov: I thought `Opt_hints_key_bitmap` should be more or less descriptive and renamed the class.

Olernov:
> Looking at Opt_hints::print(), I see two checks
> 
> * `opt_hint_info[i].irregular_hint`
> * `ignore_print(hint)`
> 
> Do they achieve the same goal? Can we use just one instead of two?

Yes, `ignore_print(hint)` may check `irregular_hint` as well.

mysql-test/main/opt_hints_index.test
mariadb-YuchenPei: I haven't checked all the echo messages but resolving since the patch is already in testing.

sql/opt_hints.cc
mariadb-YuchenPei: Is the code here regarding force_index_join vs force_index covered by tests now? I'm not sure which commit is "MDEV-35856: Address code review comments 2" but since the latest patch is in testing I'm resolving this thread now.

Olernov: Yes, it is covered now. Two test existing test cases started executing with different plans after the fix.

PR #4060, opened 2025-05-23
MDEV-36864 crc32 compatibility on Windows on ARM64

coneco-cy:
MDEV-36864 Makes Windows on ARM64 compile again (MSVC)

Followup for [4052 pull request](https://github.com/MariaDB/server/pull/4052#issuecomment-2902729240), rebased on 11.4,
as requested in review.

### DIFF
--- a/mysys/crc32/crc32_arm64.c
+++ b/mysys/crc32/crc32_arm64.c
@@ -4,31 +4,38 @@
 #include <stddef.h>
 
 typedef unsigned (*my_crc32_t)(unsigned, const void *, size_t);
+unsigned crc32_aarch64(unsigned, const void *, size_t);
 
 #ifdef HAVE_ARMV8_CRC
 
-#ifdef _WIN32
-#include <windows.h>
+# ifdef HAVE_ARMV8_CRYPTO
+static unsigned crc32c_aarch64_pmull(unsigned, const void *, size_t);
+# endif
+
+# ifdef _WIN32
+#  include <windows.h>
+#  ifdef __clang__
+#   include <arm_acle.h>
+#   include <arm_neon.h>
+# endif
 int crc32_aarch64_available(void)
 {
   return IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE);
 }
 
-const char *crc32c_aarch64_available(void)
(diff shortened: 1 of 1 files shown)


dr-m (commented):
This looks reasonable to me. The bulk of the changes is making the indentation of nested preprocessor directives consistent, which I had suggested in #4052.

However, I’d let @vaintroub check the `crc32c_aarch64()` and `crc32c_aarch64_available()` changes on Windows. [This piece of Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent) does not mention `pmull` at all. The flag `PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE` could be close, but according to [this post from 2014](https://community.arm.com/arm-community-blogs/b/operating-systems-blog/posts/runtime-detection-of-cpu-features-on-an-armv8-a-cpu) on Linux there are separate capability flags for the features. My understanding is that on ARM, there is no user-mode instruction such the i386 `cpuid`, and it is the job of the operating system kernel to expose the available ARM ISA extensions via some user mode interface.

I believe that in the worst case, this will reintroduce [MDEV-23030](https://jira.mariadb.org/browse/MDEV-23030) on Windows. Apparently, it is possible to [install Windows 11 on Raspberry Pi 4](https://www.tomshardware.com/how-to/install-windows-11-raspberry-pi), so this could be worth testing.

vaintroub:
Ok, those were basically my changes. pmull was not used, could not get it compiled. only the CRC32 native instructions, via intrinsics.I believe `IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE)` is the right check for that.
CI support can/should  be added now (not  just cross-compiling, but native, with test) after [this annoucement](https://blogs.windows.com/windowsdeveloper/2025/04/14/github-actions-now-supports-windows-on-arm-runners-for-all-public-repos/), I can try this after compilation is fixed.

vaintroub (approved):

coneco-cy:
Thank you, everyone.

svoj:
@coneco-cy, JFYI: we're about to enable CI for Windows in #4071.

coneco-cy:
@svoj, Awesome! Thanks.

PR #4072, opened 2025-05-28
MDEV-34425 mroonga files are not copied by mariabackup

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34425 *

## Description

Backing up with mariabackup a datadir containing
ENGINE=Mroonga tables leaves behind the corresponding *.mrn* files. Those tables are therefore broken once such backup is restored.

minor style/mtr changes by Daniel Black

## Release Notes

Added mroonga files to mariadb-backup

## How can this PR be tested?

per mtr test.

## PR quality check


grooverdan:
FYI @fdiary - thanks for patch. Can you check-in with the the CLA bot to permit contribution.

extra/mariabackup/backup_copy.cc
```diff
@@ -2068,6 +2072,47 @@ bool backup_files_from_datadir(ds_ctxt_t *ds_data,
...
+		if (const char *last = strrchr(info.name, '\\'))
+		{
+			if (!pname || last > pname)
+				pname = last;
+		}
+#endif
+		if (!pname)
+			pname = info.name;
+
+		if (!strstr(pname, ".mrn"))
```
vaintroub, line 2100: Can't you just use  `ends_with(info.name, ".mrn")` ,  instead of manipulations with pname? We only want to check file extension, right? and `ends_with` is  used exactly for it.

fdiary: Unfortunately, mroonga files have filenames like
```
table.mrn
table.mrn.000000
table.mrn.000100
table.mrn.001
table.mrn.conf
table.mrn.options
```

/cc @kou @komainu8

kou: @fdiary is (almost) correct. `database.mrn*` not `table.mrn*`. If we know the target database name here, we may be able to use `starts_with(info.name, "${database_name}.mrn")`.

vaintroub (approved):
Looks OK, after clarifications by mroonga devs.

PR #4073, opened 2025-05-28
MDEV-36487  Fix ha_innobase::check() for sequences

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36487*

## Description
InnoDB does the following check for sequence table during check
    table command:
    - There should be only one index should exist on sequence table
    - There should be only one row should exist on sequence table
    - The leaf page must be the root page for the sequence table
    - Delete marked record should not exist
    - DB_TRX_ID and DB_ROLL_PTR of the record should be 0 and 1U << 55
## How can this PR be tested?
./mtr innodb.sequence_check

## PR quality check


sql/handler.cc
```diff
@@ -5775,6 +5775,9 @@ handler::check_if_supported_inplace_alter(TABLE *altered_table,
       (table->s->row_type != create_info->row_type))
     DBUG_RETURN(HA_ALTER_INPLACE_NOT_SUPPORTED);
 
+  if (create_info->sequence)
+    DBUG_RETURN(HA_ALTER_INPLACE_NOT_SUPPORTED);
+
```
mariadb-DebarunBanerjee, line 5782: Good, I guess we don't want to support in-place alter for sequence. However, we also have overridden ha_innobse::check_if_supported_inplace_alter(). Shouldn't we add similar check there ?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15206,6 +15206,120 @@ ha_innobase::optimize(
...
+  DBUG_EXECUTE_IF("skip_no_rollback_check",
+                  goto skip_no_rollback;);
+  if (!table->no_rollback())
+  {
+    push_warning_printf(thd, Sql_condition::WARN_LEVEL_WARN, ER_NOT_KEYFILE,
+			"InnoDB: Sequence table %s doesn't have NO ROLLBACK "
+                        "flag enabled.", table->name.m_name);
+    corruption= true;
+    goto func_exit;
+  }
```
mariadb-DebarunBanerjee: 1. Suggest to make the error more user friendly. Something like `"InnoDB: Sequence table %s has ROLLBACK enabled"`
2. Suggest to move this internal flag check at the end to raise other user friendly errors earlier,

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15206,6 +15206,120 @@ ha_innobase::optimize(
...
+  root_block= buf_page_get_gen(page_id_t(space->id, clust_index->page),
+                               space->zip_size(), RW_S_LATCH, nullptr, BUF_GET,
+                               &mtr);
+  if (!root_block)
+  {
+    push_warning_printf(thd, Sql_condition::WARN_LEVEL_WARN, ER_NOT_KEYFILE,
+			"InnoDB: Sequence table %s is corrupted.",
+                        table->name.m_name);
+    corruption= false;
+    goto err_exit;
```
mariadb-DebarunBanerjee: Here and in many other "if" conditions above, we raising corruption warning but setting `corruption= false;` 

1. Did you intend to do `corruption= true;` ? 
2. Ideally, we should have test to pass through each if this error condition. We could also have a debug

mysql-test/suite/sql_sequence/check.result
```diff
@@ -124,7 +124,7 @@ delete from s;
 FLUSH TABLES;
 CHECK TABLE s;
 Table	Op	Msg_type	Msg_text
-test.s	check	Error	Fewer than one row in the table
+test.s	check	Warning	InnoDB: Sequence table test/s doesn't have NO ROLLBACK flag enabled.
```
mariadb-DebarunBanerjee: This one should revert back to raising more common error once the order is changed in code,

mariadb-DebarunBanerjee (changes requested):
The changes look good overall. Please check the comments as we discussed.

I have some general observation on the sequence feature.
1. It would have been nice to have `CHECK SEQUENCE` syntax inline with `CREATE/DROP SEQUENCE` instead of  `CHECK TABLE ` on a sequence object.
2. We should probably not allow user to ALTER and convert a sequence into a table and then mess with it this way.
> create sequence s;
> alter table s sequence=0;

I see it is being used in our test suite to simulate sequence corruption but not sure why we would allow user to shoot at his own foot this way. Perhaps best to be a debug only option,

mysql-test/suite/sql_sequence/check.result
```diff
@@ -124,7 +124,7 @@ delete from s;
 FLUSH TABLES;
 CHECK TABLE s;
 Table	Op	Msg_type	Msg_text
-test.s	check	Error	Fewer than one row in the table
+test.s	check	Warning	InnoDB: Encountered delete marked record in sequence table test/s.
```
mariadb-DebarunBanerjee, line 130: Please make sure purge is blocked so that we get consistent result.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15197,6 +15197,117 @@ ha_innobase::optimize(
...
+    corruption= true;
+    goto err_exit;
+  }
+
+  if (page_get_n_recs(root_block->page.frame) != 1)
+  {
+    push_warning_printf(thd, Sql_condition::WARN_LEVEL_WARN, ER_NOT_KEYFILE,
+			"InnoDB: Should have only one record in sequence "
+                        "table %s. But it has %d records.", table->name.m_name,
+                        page_get_n_recs(root_block->page.frame));
```
mariadb-DebarunBanerjee: page_get_n_recs return uint16_t. I think %u is more appropriate here.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15197,6 +15197,117 @@ ha_innobase::optimize(
...
+                        table->name.m_name);
+    corruption= true;
+  }
+err_exit:
+  mtr.commit();
+func_exit:
+  if (corruption)
+  {
+    dict_set_corrupted(clust_index, "Table corruption");
+    return false;
```
mariadb-DebarunBanerjee, line 15306: Can you please add a case where we check the sequence operation like NEXTVAL() after marking it corrupt. We should return a right error message. This would probably need you to alter it back to sequence and use the old frm file.

mariadb-DebarunBanerjee (approved):
Thanks for addressing the comments. Added some more minor comments for your consideration.

PR #4075, opened 2025-05-28
MDEV-29474 - main.lock_sync fails with timeout

svoj:
Reimplement test for MDEV-28567 such that it behaves consistently and consistently covers the code that it was intended to cover. The test itself moved to mdl_sync.test, because lock_sync.test is supposed for testing THR_LOCK, not MDL. Verified with DBUG_ASSERT(0) in open_tables() (from 92bfc0e8c4bb) restored.

The test is failing with various symptoms as it was designed to be sporadic. It did cover the code that it was supposed to cover in less than 0.1% of runs. That is the following command succeeds often with DBUG_ASSERT(0) in open_tables() (from 92bfc0e8c4bb) restored: ./mtr --repeat=25 --parallel=40 lock_sync{,,,,}{,,,,}

Test is also reimplemented such that it doesn't rely on buggy "MDL overweight" behaviour as described in MDEV-36887. Previously it did require preceding statements that fall in deadlock in default connection. Now it works consistently in both cases: if there was preceding deadlock and if there was none.

mdl_after_find_deadlock sync point added to avoid premature deadlock handling by connections that don't intend to handle it. ALTER VIEW connection is supposed to handle it, this makes test behaviour consistent.

Tracking history:
92bfc0e8c4bb MDEV-17554, assert added
f88511647aa MDEV-28567, test added, assertion modified
fe3adde2507 MDEV-29060, test moved, fix attempted
8ee93b9cb48 MDEV-29060, fix attempted
611d4425108 fix attempted
77c465d5aacd assertion modified
4c695c85bdb9 assertion removed
MDEV-29474 - subj
MDEV-36665 - duplicate of MDEV-29474


vuvova:
could you please explain in the commit comment or in the comment in the test, what the test does? Though your test is much simpler than what you removed, I still don't see why one needs four connections to get `open_tables()` to fail

vuvova (changes requested):
could you please explain in the commit comment or in the comment in the test, what the test does? Though your test is much simpler than what you removed, I still don't see why one needs four connections to get `open_tables()` to fail

vuvova (approved):

PR #4076, opened 2025-05-29
MDEV-34388: Stack overflow on Alpine Linux (postfix) old gcc+Debug+MSAN

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-34388*

## Description

Unoptimized CMAKE_BUILD_TYPE=Debug with ASAN build on GCC 7.5, 10.2.1, 12.2.0 filed to compile fail with:
sql/sql_show.cc: In function ‘void __static_initialization_and_destruction_0(int, int)’: sql/sql_show.cc:10461:1: error: the frame size of 75216 bytes is larger than 65536 bytes [-Werror=frame-larger-than=] 10461 | }
      | ^

Disable this frame-larger-than for the sql_show.cc file on older gcc compilers under Debug.

Fix from Marko per bug report.

## Release Notes

not required.

## How can this PR be tested?

compile with ASAN and a gcc with CMAKE_BUILD_TYPE=Debug

## PR quality check


vuvova (changes requested):
I don't see how your change is "Debug+MSAN".

As discussed elsewhere, please, just disable `-Wframe-larger-than=` **globally** if sanitizers are used.

vuvova (approved):
thanks. basically ok to push, I'm approving the PR, but please check the following:

the original commit by @dr-m had `=16384`, you changed it first to `=16384` or `=65536` but now you're changing it to `=16k`, so, using "k" suffix for the first time. Does it even work?

... checked documentation. It seems to say that "kB" works (not "k"). But since what version? May be it's safer to use just `=16384` as before?

grooverdan:
> thanks. basically ok to push, I'm approving the PR, but please check the following:
> 
> the original commit by @dr-m had `=16384`, you changed it first to `=16384` or `=65536` but now you're changing it to `=16k`, so, using "k" suffix for the first time. Does it even work?
> 
> ... checked documentation. It seems to say that "kB" works (not "k"). But since what version? May be it's safer to use just `=16384` as before?

sure. Updated.

I can't push to 10.6 so leaving that to you.

PR #4077, opened 2025-05-29
MDEV-34493 memory leak when using galera node with only replication thread

temeo:
- [x] *The Jira issue number for this PR is: MDEV-34493*

## Description

Applying of write set containing many row events causes applier `thd->mem_root` to grow until applying is over. This may contribute to memory fragmentation problem, producing large chunks of memory which are not released back to operating system.

In order to reduce memory fragmentation, free `thd->mem_root` after applying row events which have flag STMT_END_F set.

## How can this PR be tested?

For this fix there is no deterministic test case as there is not general way to observer memory allocator internal state. In order to verify that the fix is effective, use reported SQL in MDEV-34493 comments to reproduce the problem and observe process memory usage with standard system tools.

## PR quality check

### DIFF
--- a/sql/log_event_server.cc
+++ b/sql/log_event_server.cc
@@ -5969,7 +5969,11 @@ int Rows_log_event::do_apply_event(rpl_group_info *rgi)
       slave_rows_error_report(ERROR_LEVEL, thd->is_error() ? 0 : error,
                               rgi, thd, table, get_type_str(),
                               RPL_LOG_NAME, log_pos);
-    if (thd->slave_thread)
+    if (thd->slave_thread
+#ifdef WITH_WSREP
+        || (WSREP(thd) && wsrep_thd_is_applying(thd))
+#endif /* WITH_WSREP */
+    )
       free_root(thd->mem_root, MYF(MY_KEEP_PREALLOC));
   }
 


sql/log_event_server.cc
```diff
@@ -5969,7 +5969,9 @@ int Rows_log_event::do_apply_event(rpl_group_info *rgi)
       slave_rows_error_report(ERROR_LEVEL, thd->is_error() ? 0 : error,
                               rgi, thd, table, get_type_str(),
                               RPL_LOG_NAME, log_pos);
-    if (thd->slave_thread)
+    if (IF_WSREP(thd->slave_thread ||
```
bnestere: Looks a little strange having `thd->slave_thread` twice for both A and B conditions - perhaps just `#ifdef WITH_WSREP` around your conditions would look nicer.

bnestere (approved):
Looks good to me @temeo. I only left one minor cosmetic consideration for you :)

sql/log_event_server.cc
mariadb-TeemuOllakka: This is always a bit matter of taste ;) But yes, in this case `#ifdef WITH_WSREP` looks nicer. Changed as suggested.

mariadb-TeemuOllakka:
@bnestere Thanks for the review!

PR #4078, opened 2025-05-29
Make building of the INFO_SRC and INFO_BIN files optional

FaramosCZ:
## Description

This is an effort to upstream the Fedora patch introduced in this commit:
  https://src.fedoraproject.org/rpms/mariadb10.11/c/7e36d4?branch=rawhide

MariaDB upstream does not pack it, neither we want to pack it in Fedora.

## Release Notes
None.
This change introduce a conditional the downstream can make a use of. The default value is set to the current behavior.

## How can this PR be tested?
Test build without `-DWITH_INFO_FILES` CMake option.
Test build with `-DWITH_INFO_FILES=ON` CMake option.
Test build with `-DWITH_INFO_FILES=OFF` CMake option.

## PR quality check

## Discussion
The proposed change is minimum that allows me to cleanly abandon the code in downstream Fedora and RHEL packages.

Bases on an old discussion in the https://jira.mariadb.org/browse/MDEV-7584 you might later want to set the default to OFF, or ditch the whole INFO_{BIN,SRC} code as whole.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -542,36 +542,39 @@ IF(SOURCE_REVISION OR
                ${PROJECT_BINARY_DIR}/include/source_revision.h )
 ENDIF()
 
-CONFIGURE_FILE(
-    ${CMAKE_SOURCE_DIR}/cmake/info_macros.cmake.in
-    ${CMAKE_BINARY_DIR}/info_macros.cmake @ONLY)
+OPTION (WITH_INFO_FILES "Control creation of INFO_SRC and INFO_BIN files" ON)
+IF(WITH_INFO_FILES)
+  CONFIGURE_FILE(
+      ${CMAKE_SOURCE_DIR}/cmake/info_macros.cmake.in
+      ${CMAKE_BINARY_DIR}/info_macros.cmake @ONLY)
 
-# Handle the "INFO_*" files.
-INCLUDE(${CMAKE_BINARY_DIR}/info_macros.cmake)
-# Source: This can be done during the cmake phase, all information is
-# available, but should be repeated on each "make" just in case someone
-# does "cmake ; make ; git pull ; make".
-CREATE_INFO_SRC(${CMAKE_BINARY_DIR}/Docs)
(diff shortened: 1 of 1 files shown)


svoj:
@FaramosCZ why `-DWITH_INFO_FILES=OFF` was needed? Isn't it enough to update spec file to ignore `INFO_*`?

FaramosCZ:
Hmm ... :thinking: 

Author blindness?
Now I see we are deliberately calling `install` on top of them, not that they are installed by the CMake.

I will test it tomorrow, thanks for the second opinion :+1:

FaramosCZ:
You were right @svoj, I don't need the new CMake option.

So what about changing this PR so it does the same thing, but the CMake option defaults to `OFF`?

We may cut a bit of wasted computation time that way.

svoj:
I have no good answer atm. I'd check if upstream rpms/debs/win/bintar/srctar don't include these. INFO_SRC seems to be redundant, we already have mariadbd reporting server revision as a first thing. INFO_BIN can probably be useful.

@vaintroub your thoughts?

svoj:
@vaintroub I support your thinking. Are our support engineers going to be happy with this change? Are they going to have sufficient information to identify build details?

vaintroub:
Now, if support engineers want details of the build, they could find a log of the CI build (which is best, verbose as in cmake --build ... --verbose, or make VERBOSE=1, or IIRC in ninja -v ).  Everything is there, the version of cmake , the version of the compiler, and the exact compile  and link commands for every single binary we produce,  not a Linux-oriented "digest" that this INFO thing was trying to present.

FaramosCZ:
small note:
in case of complete removal, it's also used in the `file_contents.test`, now disabled for years by https://jira.mariadb.org/browse/MDEV-6526.
So the test might be dropped too.

vaintroub:
No, 10.11 I would not really touch

svoj:
@FaramosCZ now that we have #4090, I guess we can close this one, right?

PR #4079, opened 2025-05-29
Make sure the 'lex_hash.h' and 'lex_token.h' header files are packed

FaramosCZ:
## Description

In some cases the '.' path used to pack the header files may not equal the '${CMAKE_CURRENT_BINARY_DIR}' path where these two header files are.

This is the case of Fedora 33 and later:
  https://fedoraproject.org/wiki/Changes/CMake_to_do_out-of-source_builds

## Release Notes
none

## How can this PR be tested?
Build must pass on both upstream and downstream

## PR quality check


FaramosCZ:
@svoj 
> These are generated files, there seems to be no need to install them for the purpose of building at least

Not for the purpose of building the server, but instead for the purpose of packing them into the resulting RPMs, so 3rd party server plugins have them available when building against them.


The problem is simply these files are not installed in our case.
And I strongly suspect the $description is the culprit.

We didn't even know it was happening. We discovered it when doing a diff between MariaDB upstream and Fedora downstream RPMs. It's easy to miss that they are not present, as it's completely silent.

If you do out-of-source builds, could you please double-check that the files are correctly installed and are present into the resulting packages?

svoj:
I need to double check, but it looks like these headers are pretty much guts of MariaDB server and are not intended to be used by plugins. But if you say we have them  in MariaDB upstream RPMs, it is definitely something to analyse.

svoj:
Alright, so it appears to be that @vuvova suggests that we should have `lex_hash.h` and `lex_token.h` included. One question less, we just need to verify correctness of the fix and be done with it.

FaramosCZ:
Any progress with the review?

svoj:
@FaramosCZ sorry for the delay, I was vacationing. I will discuss this with colleagues.

sql/CMakeLists.txt
```diff
@@ -543,6 +543,11 @@ ENDIF(WIN32)
 IF(NOT WITH_WSREP)
   SET(EXCL_WSREP "wsrep_[a-np-z]*.h")
 ENDIF()
+INSTALL(FILES
```
vaintroub, line 520: Note, that for in-source case, the same files are installed twice: here, and couple of lines below.
It has never been a problem so far, but you might want  to be aware. if you want exceptional future-bug-safety, you might add a guard  `IF( NOT ("${CMAKE_CURRENT_SOURCE_DIR}" STREQUAL ${CMAKE_CURRENT_BINARY_DIR}") ) ` or something like that.

vuvova:
moved the PR to 10.6

PR #4080, opened 2025-06-01
MDEV-36765: JSON Histogram cannot handle >1 byte characters

grooverdan:
- [x] *The Jira issue number for this PR is: MDEV-36765*

## Description

When json_escape changed[1] to return a -1 in the case of a character that didn't match the character set, json_unescape_to_string assumed the -1 meant out of memory and just looped with more memory.

Problem 1 - json_escape needs to return a different code so that the different between charset incompatibility and out of memory needs to occur. This enables json_escape_to_string to handle the it correctly (ignore and fail seems the best option).

Problem 2 - JSON histograms need to support character with where the column json min/maximum value aren't a character set represented by a single byte.

Problem 2 was previously hidden as ? was a result of the conversion.

As JSON histograms can relate to columns when have an explict character set, use that and fall back to bin which was the previous default for non-string columns.

[1] regression from: f699010c0fc570786f6fe271f4dc3b2c84f8521d

## Release Notes

Per MDEV

## How can this PR be tested?

test included.

## PR quality check

### DIFF
--- a/sql/item_jsonfunc.cc
+++ b/sql/item_jsonfunc.cc
@@ -907,7 +907,7 @@ error:
 String *Item_func_json_unquote::val_str(String *str)
 {
   json_engine_t je;
-  int c_len;
+  int c_len= JSON_ERROR_OUT_OF_SPACE;
   String *js;
 
   if (!(js= read_json(&je)))
@@ -930,7 +930,15 @@ String *Item_func_json_unquote::val_str(String *str)
   return str;
 
 error:
-  report_json_error(js, &je, 0);
+  if (current_thd)
+  {
+    if (c_len == JSON_ERROR_OUT_OF_SPACE)
+      my_error(ER_OUTOFMEMORY, MYF(0), je.value_len);
+    else if (c_len == JSON_ERROR_ILLEGAL_SYMBOL)
+      push_warning_printf(current_thd, Sql_condition::WARN_LEVEL_WARN,
+                          ER_JSON_BAD_CHR, ER_THD(current_thd, ER_JSON_BAD_CHR),
+                          0, "unquote", 0);
+  }
   return js;
 }
 
@@ -4195,9 +4203,14 @@ int Arg_comparator::compare_json_str_basic(Item *j, Item *s)
                                (uchar *) (value2.ptr() + je.value_len))) < 0)
        {
          if (current_thd)
-           push_warning_printf(current_thd, Sql_condition::WARN_LEVEL_WARN,
-                               ER_JSON_BAD_CHR, ER_THD(current_thd, ER_JSON_BAD_CHR),
-                               0, "comparison", (int)((const char *) je.s.c_str - js->ptr()));
+         {
+           if (c_len == JSON_ERROR_OUT_OF_SPACE)
+             my_error(ER_OUTOFMEMORY, MYF(0), je.value_len);
+           else if (c_len == JSON_ERROR_ILLEGAL_SYMBOL)
+             push_warning_printf(current_thd, Sql_condition::WARN_LEVEL_WARN,
+                                 ER_JSON_BAD_CHR, ER_THD(current_thd, ER_JSON_BAD_CHR),
+                                 0, "comparison", (int)((const char *) je.s.c_str - js->ptr()));
+         }
          goto error;
        }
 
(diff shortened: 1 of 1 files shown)


spetrunia:
> Problem 2 - JSON histograms need to support character with where the column json min/maximum value aren't a character set represented by a single byte.

This will need a more precise wording. 
"aren't a character set represented by a single byte" .
I roll back the server before https://github.com/MariaDB/server/commit/f699010c0fc570786f6fe271f4dc3b2c84f8521d so that I can see the bug.
Then, I try this example with Russian characters (each represented by two bytes in UTF-8)
```sql
create table t1 (
  col1 varchar(10)
);

insert into t1 values
('а'),('б'),('в'),('г'),('д'),('е'),('ж'),('з'),('и'),('й');
select sin(0);
analyze table t1 persistent for all;
select hex(col1) from t1;
explain extended select * from t1 where col1 < 'а';
explain extended select * from t1 where col1 < 'в';
explain extended select * from t1 where col1 < 'д';
explain extended select * from t1 where col1 < 'ж';
explain extended select * from t1 where col1 < 'й';

drop table t1;
```
 and the execution doesn't even enter json_unescape_to_string(), because je->value_escaped=0 in read_bucket_endpoint().

Then, I add quotes and make it un-escape and it still works:
```sql
 create table t1 (
  col1 varchar(10)
);

insert into t1 values
('"а'),('"б'),('"в'),('"г'),('"д'),('"е'),('"ж'),('"з'),('"и'),('"й');
select sin(0);
analyze table t1 persistent for all;
select hex(col1) from t1;       
explain extended select * from t1 where col1 < '"а';
explain extended select * from t1 where col1 < '"в';
explain extended select * from t1 where col1 < '"д';
explain extended select * from t1 where col1 < '"ж';
explain extended select * from t1 where col1 < '"й';

drop table t1;
```
apparently  any character with my_wc_t(c) < 256 is still fine...  I'm not sure what's the name for these characters...

spetrunia:
... wait, it seems to work but actually the handling is not correct...

spetrunia:
Discovered https://jira.mariadb.org/browse/MDEV-36946.  The histogram data can be incorrect, both before and after this patch.  

The testcase for this patch passes when I run it against 459dfe99d148271834a0fcbc895fc7b9508d1451.  (The last commit before  the https://github.com/MariaDB/server/commit/f699010c0fc570786f6fe271f4dc3b2c84f8521d ) However when I tried to amend the testcase I've hit MDEV-36496....

*EDIT*:  was incorrect about MDEV-36946.

spetrunia:
.. so, in 10.11.12 everything was fine with the histograms...
misread the commit comment, thought 
> Problem 2 was previously hidden as ? was a result of the conversion.

applied before that

spetrunia:
Wait,
```sql 
set @val=CONCAT('"ab', UNHEX('FFFFFFFF') ,'"');
select json_unquote(@val);
```
gives 
```
mysqltest: At line 4: query 'select json_unquote(@val)' failed: ER_OUTOFMEMORY (1037): Out of memory; restart server and try again (needed 6 bytes)
```
the error is reported wrongly. Found this when trying to add test coverage for the added code.

spetrunia:
Another testcase, with simpler characters:

```sql
set names utf8;

select collation('"ФФ"');
select JSON_UNQUOTE('"ФФ"');

--echo #
select CONVERT('"ФФ"' using cp1251);
select collation(CONVERT('"ФФ"' using cp1251));
select JSON_UNQUOTE(CONVERT('"ФФ"' using cp1251));
```

In revision 459dfe99d148271834a0fcbc895fc7b9508d1451, this gives:
```
set names utf8;
select collation('"ФФ"');
collation('"ФФ"')
utf8mb3_general_ci
select JSON_UNQUOTE('"ФФ"');
JSON_UNQUOTE('"ФФ"')
ФФ
#
select CONVERT('"ФФ"' using cp1251);
CONVERT('"ФФ"' using cp1251)
"ФФ"
select collation(CONVERT('"ФФ"' using cp1251));
collation(CONVERT('"ФФ"' using cp1251))
cp1251_general_ci
select JSON_UNQUOTE(CONVERT('"ФФ"' using cp1251));
JSON_UNQUOTE(CONVERT('"ФФ"' using cp1251))
"ФФ"
```

in this PR:
```
set names utf8;
select collation('"ФФ"');
collation('"ФФ"')
utf8mb3_general_ci
select JSON_UNQUOTE('"ФФ"');
JSON_UNQUOTE('"ФФ"')
ФФ
#
select CONVERT('"ФФ"' using cp1251);
CONVERT('"ФФ"' using cp1251)
"ФФ"
select collation(CONVERT('"ФФ"' using cp1251));
collation(CONVERT('"ФФ"' using cp1251))
cp1251_general_ci
select JSON_UNQUOTE(CONVERT('"ФФ"' using cp1251));
...
mysqltest: At line 10: query 'select JSON_UNQUOTE(CONVERT('"ФФ"' using cp1251))' failed: ER_OUTOFMEMORY (1037): Out of memory; restart server and try again (needed 2 bytes)
```

spetrunia:
Merged into the main tree.

PR #4086, opened 2025-06-03
MDEV-9345: Replication to enable filtering on master

MohanadKh03:
This PR is part of GSoC 2025 for MDEV-9345 and will consist of multiple parts/milestones and will update the progress of each one after each one is reviewed and squash each and every milestone later on 

- [x] The Jira issue number for this PR is: [MDEV-9345](https://jira.mariadb.org/browse/MDEV-9345)

## Description
This patch moves replication filtering to the binlog dump thread so that filtering happens on the primary before events are sent to replicas, instead of being handled on the replica side using ```replicate_*``` options.

Unlike ```binlog_do_*``` and ```binlog_ignore_*```, which prevent events from being written to the binlog at all, this approach logs everything as usual and applies filtering only when dumping to the replica. This helps preserve a complete binlog, which is important for features like Point-in-Time Recovery.

## Release Notes
New dynamically changed system variables that shall be added in https://mariadb.com/kb/en/replication-filters/
- ```binlog_dump_do_db```
- ```binlog_dump_ignore_db```
- ```binlog_dump_do_table```
- ```binlog_dump_ignore_table```
- ```binlog_dump_wild_do_table```
- ```binlog_dump_wild_ignore_table```

## How can this PR be tested?

To be added later as this is currently a draft

## PR quality check


sql/sys_vars.inl
```diff
@@ -799,6 +800,22 @@ public:
...
+  int opt_id;
+
+  const uchar *global_value_ptr(THD *thd, const LEX_CSTRING *base)
+    const override;
+  bool set_filter_value(const char *value, Master_info *mi);
+};
+
+// No additional functionality all behavior is inherited from Sys_var_rpl_filter to
+// avoid code duplication. This class inherits Sys_var_rpl_filter since setting the
+// global variable should also be dynamic unlike Sys_var_binlog_filter.
```
bnestere: Comment needs to be aligned with the convention in this file

sql/sys_vars.cc
```diff
@@ -5913,6 +5988,47 @@ static Sys_var_rpl_filter Sys_replicate_rewrite_db(
        "Example: replicate-rewrite-db=master_db_name->slave_db_name",
        PRIV_SET_SYSTEM_GLOBAL_VAR_REPLICATE_REWRITE_DB);
 
+static Sys_var_binlog_dump_filter Sys_binlog_dump_do_db(
+  "binlog_dump_do_db", OPT_BINLOG_DUMP_DO_DB,
+  "Tells the primary to restrict binlog dump to updates of tables "
+  "whose names appear in the comma-separated list. "
+  "This filtering is applied only at the binlog and is not replicated to the relay log",
```
bnestere: This is good for the description to mention both what is filtered and where :smile:. Though I'd suggest being a bit more specific so users know _exactly_ where this filtering happens. I'd suggest the following:

Tells the primary's binlog dump threads to only transmit binary log events that update databases who names appear in this comma-separated list. Applies for all replicas that connect to this server. Note that unlike option binlog_do_db, this option does not prevent any transactions from being written into the local binary log file.

Then I'd use this as a template for the other options as well (and making sure the semantics match the option name).

sql/mysqld.cc
```diff
@@ -666,6 +666,9 @@ THD_list server_threads;
 Rpl_filter* cur_rpl_filter;
 Rpl_filter* global_rpl_filter;
 Rpl_filter* binlog_filter;
+// Used for writing to the binlog only and filter what is sent to the relay log
```
bnestere: This comment doesn't align with the comment-style for the rest of the file. Unfortunately, due to how long the MariaDB source code has been around with various developers, many of our files follow different stylistic conventions (e.g. comments, indentations, etc). Our current general rule is when modifying any given file, is to follow the existing conventions set forth by that file (and this is sometimes difficult, as sometimes things slip through and then there are multiple conventions within the same file :disappointed:).

Also the content doesn't read quite right. "Used for writing to the binlog" makes it seem like we are actually working at the level of writing to the binlog (i.e the `binlog_filter` semantics), but we are filtering at transmission time. I'd also remove "sent to the relay log". The binlog dump thread doesn't care what happens to the events it sends out, e.g. we can also connect to the server via `mariadb-binlog --read-from-remote-server` and we'd encounter the same filtering logic.

sql/mysqld.cc
```diff
@@ -666,6 +666,9 @@ THD_list server_threads;
...
+// Unlike binlog_filter which prevents writing to the binlog entirely
+Rpl_filter* binlog_dump_thread_filter;
```
bnestere: We can shorten this a bit to just `binlog_dump_filter` :smile:  which is also more consistent with the option names themselves.

sql/sys_vars.cc
```diff
@@ -5809,6 +5809,47 @@ bool Sys_var_rpl_filter::set_filter_value(const char *value, Master_info *mi)
   return status;
 }
 
+bool Sys_var_binlog_dump_filter::global_update(THD *thd, set_var *var)
+{
+  bool result= true;                            // Assume error
+
+  mysql_mutex_unlock(&LOCK_global_system_variables);
+
+  // passing NULL as Master_info is just a workaround to follow the Sys_var_rpl_filter's overridden method
```
bnestere: Hmm, reading this makes me realize my suggestion of overriding `Sys_var_rpl_filter` is a bad suggestion - the `Master_info` class is strictly a slave-side concept, we shouldn't introduce it to any master-side code. _The original concept of the macro `HAVE_REPLICATION` was to be able to compile a server which could function as a master (including binary logging) without any slave internals. That principle is long broken, but we should try to at least uphold it when we can :innocent:_

I'd be ok with either
 1) Making `Sys_var_binlog_dump_filter` the new super-class, which `Sys_var_rpl_filter` would inherit. Then `Sys_var_binlog_dump_filter` wouldn't have any reference to `Master_info` (I suppose it wouldn't even need the function `set_filter_value`).
 2) Just make `Sys_var_binlog_dump_filter` standalone (with the boilerplate).

Noting that we discussed on Zulip that it would be good to refactor all `Sys_var_*_filter` classes anyway, option 2 is probably the quickest for now.

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_db.test
```diff
@@ -0,0 +1,86 @@
...
+DROP TABLE t1;
+USE ignored_db;
+DROP TABLE t2;
+USE test;
+DROP DATABASE ignored_db;
+--source include/sync_slave_sql_with_master.inc
+--source include/rpl_end.inc
+
+connection slave;
+SET @@GLOBAL.binlog_dump_do_db="";
```
bnestere: When setting/returning global variables to their old values in tests, please store the original value in a variable and restore it. For example

```
SET @old_binlog_dump_do_db= @@GLOBAL.binlog_dump_do_db;
SET @@GLOBAL.binlog_dump_do_db = 'test';
...
SET @@GLOBAL.binlog_dump_do_db = @old_binlog_dump_do_db;
```

It seems a bit cumbersome, but we have tools which run tests with different configuration variables, and if some variable is set before you run the test (which your test overrides), the value of the variable needs to exactly match the original value or the test fails.

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_db.test
```diff
@@ -0,0 +1,86 @@
...
+--echo # Cleanup
+connection master;
+USE test;
+DROP TABLE t1;
+USE ignored_db;
+DROP TABLE t2;
+USE test;
+DROP DATABASE ignored_db;
+--source include/sync_slave_sql_with_master.inc
+--source include/rpl_end.inc
```
bnestere:
```
--source include/rpl_end.inc
```
should always be last (after restoring `binlog_dump_do_db`)

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_db.test
```diff
@@ -0,0 +1,86 @@
...
+# which events appear in the master's binlog and the slave's relay log.
+#
+
+--source include/have_innodb.inc
+--source include/master-slave.inc
+
+--echo # Stop slave to set replicate_do_db
+connection slave;
+--source include/stop_slave.inc
+SET @@GLOBAL.binlog_dump_do_db = 'test';
```
bnestere: The `binlog_dump_*` options would only affect the `master` connection, but here you are setting them on the slave (just FYI for when you begin implementing the filter on the dump thread, and add substance to your tests :smile:).

bnestere (commented):
Hi @MohanadKh03 !

Thanks for such an early draft :) Generally things look very good! I've given some initial pointers.

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_db-master.opt
```diff
@@ -0,0 +1 @@
+--binlog-dump-do-db=test --binlog-dump-ignore-db=ignored_db
```
bnestere: Add a new line at the end please (and for the other `.opt` file as well)

mysql-test/main/mysqld--help.result
```diff
@@ -80,6 +80,50 @@ The following specify which files/extra groups are read (specified before remain
...
+ appear in this comma-separated list. Applies for all
+ replicas that connect to this server. Note: Unlike
+ binlog_do_db, this does not prevent transactions from
+ being written to the local binary log file
+ --binlog-dump-do-table=name 
+ Tells the primary's binlog dump threads to only transmit
+ binary log events that update tables whose names appear
+ in this comma-separated list. Applies for all replicas
+ that connect to this server. Note: Unlike
+ binlog_do_table, this does not prevent transactions from
```
bnestere: There is no binlog_do_table option. You can probably just say "Note: this does not prevent transactions..."

mysql-test/main/mysqld--help.result
```diff
@@ -80,6 +80,50 @@ The following specify which files/extra groups are read (specified before remain
...
+ replicas that connect to this server. Note: Unlike
+ binlog_ignore_db, this does not prevent transactions from
+ being written to the local binary log file
+ --binlog-dump-ignore-table=name 
+ Tells the primary's binlog dump threads to not transmit
+ any binary log events that update tables whose names
+ appear in this comma-separated list, even if other tables
+ might be updated by the same statement. Applies for all
+ replicas that connect to this server. Note: Unlike
+ binlog_ignore_table, this does not prevent transactions
```
bnestere: No such option (same suggestion as above)

mysql-test/main/mysqld--help.result
```diff
@@ -80,6 +80,50 @@ The following specify which files/extra groups are read (specified before remain
...
+ from being written to the local binary log file
+ --binlog-dump-wild-do-table=name 
+ Tells the primary's binlog dump threads to only transmit
+ binary log events for statements where any of the updated
+ tables match the specified database and table name
+ patterns (wildcards allowed). Applies for all replicas
+ that connect to this server. Note: Unlike
+ binlog_wild_do_table, this does not prevent transactions
```
bnestere: No such option (same suggestion as above)

mysql-test/main/mysqld--help.result
```diff
@@ -80,6 +80,50 @@ The following specify which files/extra groups are read (specified before remain
...
+ from being written to the local binary log file
+ --binlog-dump-wild-ignore-table=name 
+ Tells the primary's binlog dump threads to not transmit
+ binary log events for tables that match the given
+ wildcard pattern. Applies for all replicas that connect
+ to this server. Note: Unlike binlog_wild_ignore_table,
```
bnestere: No such option (same suggestion as above)

mysql-test/suite/rpl/include/rpl_filter_binlog_dump_db.test
```diff
@@ -0,0 +1,83 @@
+#
+# Purpose:
+#
+# This test verifies the correct behavior of replication database filters,
+# specifically: binlog_dump_do/ignore_db.
+#
+# The test compares the master's binary log (binlog) and the slave's relay log
+# to confirm how these filters affect data replication.
+#
```
bnestere, line 9: In the description (somewhere), it would be good to mention the filtering configuration specified by the `-master.opt` file.

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_table.test
```diff
@@ -0,0 +1,81 @@
...
+# This test ensures that these filters behave as expected, by examining
+# which events appear in the master's binlog and the slave's relay log.
+#
+
+--source include/have_innodb.inc
+--source include/master-slave.inc
+
+connection master;
+
+--echo # On master: create two databases and insert data
```
bnestere: s/databases/tables/g

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_table.test
```diff
@@ -0,0 +1,81 @@
...
+CREATE TABLE t1 (id INT PRIMARY KEY, data VARCHAR(50));
+CREATE TABLE t2 (id INT PRIMARY KEY, data VARCHAR(50));
+INSERT INTO t1 VALUES (1, 'row in t1');
+INSERT INTO t2 VALUES (1, 'row in t2');
+
+--echo # Capture current binlog file name
+--let $binlog_file= SELECT @@global.log_bin_basename
+--let $binlog_index= query_get_value(SHOW MASTER STATUS, File, 1)
+
+SET @@default_storage_engine='innodb';
```
bnestere: Is the intent for this to apply to the above tables? You might need to move this above those CREATE TABLE statements if so :)

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_db.test
```diff
@@ -0,0 +1,83 @@
...
+--echo # Cleanup
+connection master;
+
+
+use ignored_db;
+DROP TABLE t2;
+USE test;
+DROP TABLE t1;
+DROP DATABASE ignored_db;
+--source include/rpl_end.inc
```
bnestere: At the end of each test (after `rpl_end.inc`), can you add an echo outputting that it is the end of the test? E.g.

```
--echo # End of rpl_filter_binlog_dump_db.test
```

_And to update the echo with the appropriate test name

mysql-test/suite/rpl/t/rpl_filter_replicate_db.test
```diff
@@ -0,0 +1,93 @@
+#
```
bnestere, line 1: Why add tests for replicate_* filters? They already exist, e.g. rpl_filter_tables_dynamic.

mysql-test/suite/rpl/t/rpl_filter_binlog_dump_table.test
```diff
@@ -0,0 +1,81 @@
...
+--let $assert_text= INSERT INTO t2 must not be present in the Slave's relaylog
+--let $assert_select= INSERT INTO t2
+--let $assert_count= 0
+--source include/assert_grep.inc
+
+--echo # Cleanup
+connection master;
+DROP TABLE t1;
+DROP TABLE t2;
+--source include/rpl_end.inc
```
bnestere: Thanks for writing tests with clear context/expectations set up! Note the following tests will need to be added as well (the [ ] to become [x] when done :)):
 [ ]  Multi-statement transactions which target multiple tables (Both all and partially filtered should be tested)
 [ ]  Multi-statement transactions which target multiple databases (Both all and partially filtered should be tested)
 [ ]  Multi-statement transactions which target multiple tables and databases
 [ ]  A trasaction targeting a non-transaction table (e.g. MyISAM)
 [ ]  A transaction targeting multiple tables of different transactional capabilities (e.g. MyISAM (non-trans) and Innodb (trans)); where filtering combinations span the configuration space. That is, for booleans (N,T) where <N=non-transaction table is filtered> and <T=transactional table is filtered>, combinations (1,0), (0,1), and (1, 1) should be tested.
 [ ]  A two-phase XA transaction targeting a filtered table/db filters the event content, but never filters any `XA ...` commands nor `Xa_prepare_log_event` types
 [ ]  A two-phase alter table command which targets a filtered table
 [ ]  A semi-sync enabled connection should not wait if the transaction is filtered (this is actually probably a bug right now, we should check & test).
 [ ]  Slave configuration of replicate_marked_for_skip=FILTER_ON_MASTER doesn't interfere with master

sql/sql_repl.cc
```diff
@@ -2232,15 +2232,17 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
   /*
-    Skip events with the @@skip_replication flag set, if slave requested
-    skipping of such events.
+    Skip events that have either
+      1) @@skip_replication flag set (it has a negation value with @@skip_parallel so only one of them should be used)
+      2) Set during binlogging (set using binlog_dump_* filter) 
+
   */
-  if (info->thd->variables.option_bits & OPTION_SKIP_REPLICATION)
+  if (info->thd->variables.option_bits & OPTION_SKIP_REPLICATION || !binlog_dump_filter->is_db_empty() || binlog_dump_filter->is_on())
```
bnestere: Weird that `Rpl_filter::is_on()` only considers table-level filters :thinking: - though we don't need to address that by your patch. We can perhaps refactor that another time.

sql/log.cc
```diff
@@ -6390,6 +6342,41 @@ bool stmt_has_updated_non_trans_table(const THD* thd)
   return (thd->transaction->stmt.modified_non_trans_table);
 }
 
+/**
+  Updates the skip status of the binlog cache based on the current event.
+
+  If the event has the LOG_EVENT_SKIP_REPLICATION_F flag set, the cache is marked
+  as skippable (CACHE_SKIP_ALL). Otherwise, it is marked as non-skippable (CACHE_SKIP_NONE),
+  indicating that at least one event must be replicated later when the transaction is committed and
+  the GTID_EVENT is to be written inside write_gtid_event()
```
bnestere: A lot of your lines are over 80 characters (that's the max set by our [coding standards](https://github.com/MariaDB/server/blob/main/CODING_STANDARDS.md))

sql/log_cache.h
```diff
@@ -301,6 +303,23 @@ class binlog_cache_data
...
+    have the LOG_EVENT_SKIP_REPLICATION_F flag. This determines
+    whether the entire transaction/event group can be skipped when writing to the binary log
+    
+    - CACHE_SKIP_UNDECIDED: Initial state, typically before any events are cached. e.g: The first GTID_EVENT
+    - CACHE_SKIP_ALL: All cached events have the LOG_EVENT_SKIP_REPLICATION_F flag.
+    - CACHE_SKIP_NONE: At least one cached event does not have the skip flag,
+                       so the transaction must not be skipped.
+  */
+  enum skip_cache_status {
+      CACHE_SKIP_UNDECIDED= 0,
```
bnestere: Do we need `CACHE_SKIP_UNDECIDED`? I think the default can be `CACHE_SKIP_ALL`, and when anything is _not_ filtered, we just negate it.

I also think the names may need to be changed, I think `CACHE_SKIP_*` implies that the transaction data is skipping being written to the binlog cache entirely. This would apply to the rest of the patch too, e.g. in variable names like `skip_annotate`, etc. Perhaps `WITH_SKIP_REPLICATE` would be better? I'm open to other suggestions too.

sql/log_cache.h
```diff
@@ -384,3 +403,85 @@ class binlog_cache_data
   binlog_cache_data& operator=(const binlog_cache_data& info);
   binlog_cache_data(const binlog_cache_data& info);
 };
+
+class binlog_cache_mngr {
```
bnestere, line 428: When moving code (or doing general cleanups), we make those separate preparatory commits (to make it clear that it isn't something actually new, but you are re-structuring existing code to prepare for your patch, and it also makes merges much easier to manage).

sql/log_event_server.cc
```diff
@@ -1397,6 +1397,48 @@ Query_log_event::Query_log_event(THD* thd_arg, const char* query_arg,
   */
   flags2= (uint32) (thd_arg->variables.option_bits &
                     (OPTIONS_WRITTEN_TO_BIN_LOG & ~OPTION_NOT_AUTOCOMMIT));
+  #ifdef HAVE_REPLICATION
```
bnestere: Why is this surrounded by `HAVE_REPLICATION` blocks? It is still master-side logic (usually the distinction is `HAVE_REPLICATION` is for slave-side logic _though this concept has been broken historically, so it sadly isn't consistent_.

Note too that macros (e.g. `#ifdef`) shouldn't be indent aligned at all

sql/log_event_server.cc
```diff
@@ -1397,6 +1397,48 @@ Query_log_event::Query_log_event(THD* thd_arg, const char* query_arg,
...
+
+  MYSQL_LOCK *locks[2], **locks_end= locks;
+
+  if ((*locks_end= thd_arg->extra_lock))
```
bnestere: Can you write a comment explaining what this part does with getting locks and why it is done this way?

sql/log_event_server.cc
```diff
@@ -4688,7 +4730,6 @@ Rows_log_event::Rows_log_event(THD *thd_arg, TABLE *tbl_arg,
   DBUG_ASSERT((tbl_arg && tbl_arg->s &&
                (table_id & MAX_TABLE_MAP_ID) != UINT32_MAX) ||
               (!tbl_arg && !cols && (table_id & MAX_TABLE_MAP_ID) == UINT32_MAX));
-
```
bnestere, line 4691: New line needs to be restored

sql/sql_class.h
```diff
@@ -3421,7 +3421,7 @@ class THD: public THD_count, /* this must be first */
                         enum_binlog_row_image row_image,
                         const uchar *old_data, const uchar *new_data);
   bool prepare_handlers_for_update(uint flag);
-  bool binlog_write_annotated_row(Log_event_writer *writer);
+  bool binlog_write_annotated_row(Log_event_writer *writer, bool skip_annotate= false);
```
bnestere: Param name should be changed per a previous comment

sql/sql_repl.cc
```diff
@@ -2232,15 +2232,17 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
   {
-    uint16 event_flags= uint2korr(&((*packet)[FLAGS_OFFSET + ev_offset]));
-
-    if (event_flags & LOG_EVENT_SKIP_REPLICATION_F)
-      return NULL;
+      uint16 event_flags= uint2korr(&((*packet)[FLAGS_OFFSET + ev_offset]));
```
bnestere: Wrong indentation

sql/sys_vars.inl
```diff
@@ -805,6 +806,69 @@ protected:
...
+  }
+
+  bool do_check(THD *thd, set_var *var) override
+
+  {
+
+    DBUG_ASSERT(FALSE);
+
+    return true;
+
```
bnestere, line 840: Weird spacing should be removed

sql/log.cc
```diff
@@ -1972,6 +1892,16 @@ binlog_commit_flush_stmt_cache(THD *thd, bool all,
 
   Query_log_event end_evt(thd, STRING_WITH_LEN("COMMIT"),
                           FALSE, TRUE, TRUE, 0);
+
+  if (cache_mngr)
```
bnestere, line 1901: In all the `binlog_commit_flush_*` functions that have this change, can you add a comment explaining why this is needed?

sql/log.cc
```diff
@@ -6691,6 +6682,42 @@ bool THD::binlog_write_table_maps()
   if ((*locks_end= lock))
     locks_end++;
 
+
+  bool filter_all_tables= false;
+
+  for (MYSQL_LOCK **cur_lock= locks ; cur_lock < locks_end ; cur_lock++)
```
bnestere: Two things here:
 1. Before looping through all tables to check for binlog filters, we should first check that the filter is configured (otherwise it is just wasted effort).
 2. Can you add a comment explaining why this loop is separate from the next loop, despite looping over the same thing? In other words, why it isn't combined with the previous loop.

sql/log.cc
```diff
@@ -6699,6 +6726,7 @@ bool THD::binlog_write_table_maps()
          ++table_ptr)
     {
       TABLE *table= *table_ptr;
+
```
bnestere: Extra line to remove

sql/log.cc
```diff
@@ -6691,6 +6682,42 @@ bool THD::binlog_write_table_maps()
...
+      /* 
+        If a single table map event is affected and NOT replicated (should get filtered) then every other
+        table map event (and as a consequence all other corresponding *_ROWS_EVENTs) should also get filtered
+        as well as the ANNOTATE_EVENT ... this will skip the whole transaction starting from GTID_EVENT which should
+        check the cache isnide write_gtid_event and realize that everything in the cache has the skip flag so it would 
+        also get filtered (same thing with the end events such as XID COMMIT or QUERY EVENT with COMMIT string and ROLLBACK
+        
+        Basically a single table affected by not being replicated should skip the whole transaction being binlogged with
+        the filtering
```
bnestere: Just a reminder that we don't want this in the final patch. It needs to be per event (and for the STMT_END case, the binlog dump thread will need to re-write the data with null data).

sql/log.cc
```diff
@@ -7118,6 +7165,43 @@ MYSQL_BIN_LOG::write_gtid_event(THD *thd, bool standalone,
...
+          gtid_event.flags|= LOG_EVENT_SKIP_REPLICATION_F;
+        }
+        else
+        {
+          DBUG_PRINT("info", ("GTID_EVENT: skip_cache_status != CACHE_SKIP_ALL, not setting LOG_EVENT_SKIP_REPLICATION_F"));
+        }
+      }
+      else
+      {
+        DBUG_PRINT("info", ("GTID_EVENT: cache_mngr is NULL, not setting LOG_EVENT_SKIP_REPLICATION_F"));
```
bnestere: You can parameterize all these prints into one `DBUG_PRINT`, e.g. something like 
```
... ("skip_cache_status: %s; %ssetting", enum == ALL ? "SKIP_ALL" : "SKIP_NONE", enum == ALL ? "" "not ");
```

sql/log.cc
```diff
@@ -13214,8 +13309,8 @@ inline bool Binlog_commit_by_rotate::should_commit_by_rotate(
      file. It happens in the case binlog cache buffer is larger than
      threshold
    */
-  if (cache_data->file_reserved_bytes() == 0 ||
-      cache_data->cache_log.disk_writes == 0)
+  if (cache_data && (cache_data->file_reserved_bytes() == 0 ||
```
bnestere: Was this problematic or is this preventative? We shouldn't need to check cache_data existence here (binlog commit by rotate feature assumes it exists), though `DBUG_ASSERT(cache_data);` would be good to ensure it logically. Though I'd suggest doing it as a separate commit (as it is cleanup, not really related to MDEV-9345 AFAIK).

sql/log.cc
```diff
@@ -13244,7 +13339,7 @@ bool Binlog_commit_by_rotate::commit(MYSQL_BIN_LOG::group_commit_entry *entry)
     cache_data= cache_mngr->get_binlog_cache_data(false);
 
   /* Call them before enter log_lock to avoid holding the lock long */
-  if (cache_data->sync_temp_file())
+  if (cache_data && cache_data->sync_temp_file())
```
bnestere: Same

mysql-test/suite/rpl/t/rpl_filter_replicate_db.test
MohanadKh03: These new tests are meant to cover how the binlog itself behaves 

The replicate_filter_*_dynamic does not check it at the binlog-level/relay-log level

sql/log.cc
MohanadKh03: Preventative yes just as a sanity check as it failed in other places related to filtering of 9345.. Assert seems better but ill remove it for a later patch other than 9345

sql/log_cache.h
MohanadKh03: Before we got to the filter all-or-nothing discussions I had UNDECIDED just to make it more flexible but right now it's not needed yes

As for the naming how about
EVENT_GROUP_SKIP
and
EVENT_GROUP_KEEP

This would not conflict with how partial filtering could happen later on as this status is only concerned with all-or-nothing event group filtering

sql/log_cache.h
MohanadKh03: Oh yeah my bad this was needed because there was a conflict of #include's in log.cc and log_cache.h when using binlog_cache_mngr inside ```log_event_server.cc`` although right it has been removed and not used in that file anymore so no more conflicts

Technically This is not needed anymore but overall since this patch deals heavily with binlog caches I thought it didnt really make much sense to have ```binlog_cache_mngr``` inside ```log.cc``` right ? Makes much more sense to be inside a .h file + to be in a file that is related to cache ```log_cache``` instead of ```log.cc```

Do you still think it should be in another commit ?  I could mention in the commit message that this is kind of a clean-up

sql/sql_class.h
MohanadKh03: ```bool filter_all_event_group= false``` seems more appropriate ? It should not conflict with the upcoming partial filtering as this parameter currently deals with all-or-nothing filtering

sql/sql_class.h
```diff
@@ -3421,7 +3421,8 @@ class THD: public THD_count, /* this must be first */
                         enum_binlog_row_image row_image,
                         const uchar *old_data, const uchar *new_data);
   bool prepare_handlers_for_update(uint flag);
-  bool binlog_write_annotated_row(Log_event_writer *writer);
+  bool binlog_write_annotated_row(Log_event_writer *writer, 
+                                  bool filter_all_event_group= false);
```
bnestere: It'd be good to add a comment here to describe `filter_all_event_group` meaning

sql/log.h
```diff
@@ -1129,7 +1132,7 @@ class MYSQL_BIN_LOG: public TC_LOG, private Event_log
   bool write_gtid_event(THD *thd, bool standalone, bool is_transactional,
                         uint64 commit_id,
                         bool commit_by_rotate,
-                        bool has_xid= false, bool ro_1pc= false);
+                        bool has_xid= false, bool ro_1pc= false, bool skip_all_events= false);
```
bnestere: `skip_all_events` I think is confusing terminology, as there is the option `--skip-annotate-row-events` which actually means to not write annotate row events into the binary log at all. Whereas this patch is writing the `skip_replicate` flag. I'd suggest a naming convention that relates to filtering (as that is the overall semantic) or `write_skip_rpl` (as that is the actual impact).

Also I think this line goes beyond 80 characters, so please split the line.

sql/log.h
```diff
@@ -1023,7 +1025,8 @@ class MYSQL_BIN_LOG: public TC_LOG, private Event_log
   bool write_incident_already_locked(THD *thd);
   bool write_incident(THD *thd);
   void write_binlog_checkpoint_event_already_locked(const char *name, uint len);
-  bool write_table_map(THD *thd, TABLE *table, bool with_annotate);
+  bool write_table_map(THD *thd, TABLE *table, bool with_annotate, 
+                       bool filter_all_tables= false);
```
bnestere: A comment would be good to describe filter_all_tables

sql/log_event_server.cc
```diff
@@ -1397,6 +1397,63 @@ Query_log_event::Query_log_event(THD* thd_arg, const char* query_arg,
...
   flags2= (uint32) (thd_arg->variables.option_bits &
                     (OPTIONS_WRITTEN_TO_BIN_LOG & ~OPTION_NOT_AUTOCOMMIT));
+  #ifdef HAVE_REPLICATION
+
+  bool is_end_event= false;
+
+  if (q_len == 6 && strncasecmp(query, "COMMIT", 6) == 0)
+      is_end_event= true;
+  else if (q_len == 8 && strncasecmp(query, "ROLLBACK", 8) == 0)
+      is_end_event= true;
```
bnestere, line 1404: There are convenience functions `is_commit()` and `is_rollback()`

sql/log_event_server.cc
```diff
@@ -1397,6 +1397,63 @@ Query_log_event::Query_log_event(THD* thd_arg, const char* query_arg,
...
+          table_ptr != end_ptr ;
+          ++table_ptr)
+      {
+        TABLE *table= *table_ptr;
+    
+        /* 
+          If there is a single tables affected by this query then all of the 
+          tables will get filtered as well the whole event group should
+          get filtered. NOTE: Partial Filtering is only allowed in the
+          ROW format not STATEMENT format
```
bnestere, line 1435: Nice comment note! I don't know if I saw it in any of your higher level comments (those in `.h` files), can you make sure that limitation is also visible in an appropriate header file?

sql/mysqld.cc
```diff
@@ -667,6 +667,12 @@ Rpl_filter* cur_rpl_filter;
 Rpl_filter* global_rpl_filter;
 Rpl_filter* binlog_filter;
 
+/**
+  Filter used by the binlog dump thread to determine which events to transmit to replicas or clients.
+  This is separate from binlog_filter, which controls which events are written to the binary log.
```
bnestere: Lines are too long (>80 chars)

sql/sql_repl.cc
```diff
@@ -139,7 +139,13 @@ struct binlog_send_info {
   bool send_fake_gtid_list;
   bool slave_gtid_ignore_duplicates;
   bool using_gtid_state;
-
+  /*
+    True if any earlier row events in the current row-based statement
+    were sent and not fully filtered. Used to detect partially filtered
+    event groups and ensure that row events with STMT_END_F are sent correctly
+  */
+  bool row_events_sent_before;
```
bnestere: I think the variable name and comment here can be improved. The name should make it clear this is for binlog_dump filtering and also partial filtering. 

Though more generally, do we need to track this? If the GTID event itself has `@@skip_replication`, then we know there won't be partial filtering. Because the GTID event gets `@@skip_replication` when all of its events are filtered, and not otherwise.

sql/sql_repl.cc
```diff
@@ -2232,15 +2239,118 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
   }
 
   /*
-    Skip events with the @@skip_replication flag set, if slave requested
-    skipping of such events.
+    Skip events that have either
+      1) @@skip_replication flag
+      2) Set during binlogging (set using binlog_dump_* filter)
```
bnestere: Isn't the `@@skip_replication` flag set during binlogging? So the 2nd point is really just the 1st. I'd reword as something like

> Skip events with the @@skip_replication flag set, i.e. either if slave requested skipping of such events, or if filtered via the binlog_dump_* filter options.

sql/sql_repl.cc
```diff
@@ -2232,15 +2239,118 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
-    Skip events with the @@skip_replication flag set, if slave requested
-    skipping of such events.
+    Skip events that have either
+      1) @@skip_replication flag
+      2) Set during binlogging (set using binlog_dump_* filter) 
+
   */
-  if (info->thd->variables.option_bits & OPTION_SKIP_REPLICATION)
+  if (info->thd->variables.option_bits & OPTION_SKIP_REPLICATION ||
+    !binlog_dump_filter->is_db_empty() || binlog_dump_filter->is_on())
```
bnestere, line 2258: I don't think we need to check the binlog_dump_filter at this point in the life-cycle, only
`OPTION_SKIP_REPLICATION` because that should be sorted out at binlog time.

sql/sql_repl.cc
```diff
@@ -2232,15 +2239,118 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+          table_id= uint4korr(post_start);
+          post_start += 4;
+      }
+      else
+      {
+        table_id  = uint4korr(post_start);
+        table_id |= ((uint64) uint2korr(post_start + 4)) << 32;
+        // skip 6 bytes
+        post_start += RW_FLAGS_OFFSET;
+      }
```
bnestere: Are the table_id or post_start used at all? It looks like all you are reading is `m_flags`, you should just be able to set up the pointer directly and read the flags in.

sql/sql_repl.cc
```diff
@@ -2232,15 +2239,118 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+      else {
+        if (!info->row_events_sent_before) {
+          return NULL;
+        }
+
+        /* 
+          TODO: Clear payload and rewrite the checksum
+        */
+        // const uchar *row_data_start = buf + common_header_len + post_header_len;
```
bnestere: The post header includes the table_id and m_flags, which should also be zeroed. So I'd think you could write `0`'s starting at `buf + common_header_len`

sql/sql_repl.cc
```diff
@@ -2232,15 +2239,118 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        //                       != BINLOG_CHECKSUM_ALG_OFF && 
+        //                       info->current_checksum_alg 
+        //                       != BINLOG_CHECKSUM_ALG_UNDEF) 
+        //               ? BINLOG_CHECKSUM_LEN : 0;
+        // size_t row_data_len = len - (row_data_start - buf) - checksum_len;
+
+        // memset((uchar*)row_data_start, 0, row_data_len); // zero out row payload
+        // if (checksum_len > 0) 
+        // {
+        //   uint32 crc = my_checksum(0, buf,  len - BINLOG_CHECKSUM_LEN);
```
bnestere: Hmm, I'm not sure if it makes a difference, but in the code, the header checksum happens separately from the data checksum. So if you're running into a problem with this, it might make more sense to do:

```
uint32 crc= my_checksum(0, buf, common_header_len);
crc= my_checksum(crc, buf + common_header_len, len - common_header_len - BINLOG_CHECKSUM_LEN);
```

sql/log_cache.h
```diff
@@ -128,6 +142,9 @@ class binlog_cache_data
     status= 0;
     incident= FALSE;
     before_stmt_pos= MY_OFF_T_UNDEF;
+    // Since the cache_data is reused so we should reset it to not conflict 
+    event_group_rpl_filter= false;
```
bnestere: Is `partial_filtered_table_ids` intentionally left out of being freed/reset here? E.g. as an optimization to not have to re-build it when the table_ids don't change? I think that would be more involved of an optimization, otherwise it would leak for deleted tables (and those infrequently touched).

sql/log.cc
```diff
@@ -6390,6 +6357,46 @@ bool stmt_has_updated_non_trans_table(const THD* thd)
...
+  Updates the skip status of the binlog cache based on the current event.
+
+  If the event has the LOG_EVENT_SKIP_REPLICATION_F flag set,
+  Then there are two options either this is is part of a
+  query that affects a single table/db or it would be part 
+  of a partial filter.
+
+  If it's the first case then all events in an event group should be
+  filtered , this is done by updating the cache's event group filter
+  flag. Otherwise then it's part of the partial filter which should have
```
bnestere: Something is off in this description. There's some gap in logic between "part of a query that affects a single table/db" and "all events in an event group should be filtered".

sql/log.cc
```diff
@@ -6649,7 +6656,7 @@ void THD::binlog_prepare_for_row_logging()
    Write annnotated row event (the query) if needed
 */
 
-bool THD::binlog_write_annotated_row(Log_event_writer *writer)
+bool THD::binlog_write_annotated_row(Log_event_writer *writer, bool skip_annotate)
```
bnestere: Ambiguous variable name (due to `--skip-annotate-row-events`, that I previously mentioned)

sql/log.cc
```diff
@@ -6691,6 +6702,85 @@ bool THD::binlog_write_table_maps()
...
+              TODO: should this be done only if filter_all_tables is false ? 
+            */
+            bool is_transactional= table->file->row_logging_has_trans;
+            is_transactional|= this->variables.option_bits & OPTION_GTID_BEGIN;
+
+            binlog_cache_mngr *const cache_mngr= this->binlog_get_cache_mngr();
+            binlog_cache_data *cache_data= (cache_mngr->
+                                  get_binlog_cache_data(is_transactional));
+            ulonglong table_id= table->s->table_map_id;
+            uchar *stored= (uchar*) my_malloc(PSI_INSTRUMENT_ME,
```
bnestere: Could we pre-allocate a chunk of memory to hold all table_ids up front? Then we aren't allocating lots of small chunks individually (which could contribute to memory fragmentation). I'd think we could make use of the `THD::mem_root` here too, instead of `my_malloc()`.

sql/log.cc
```diff
@@ -6691,6 +6702,85 @@ bool THD::binlog_write_table_maps()
...
+                                    sizeof(ulonglong), MYF(0));
+            memcpy(stored, &table_id, sizeof(ulonglong));
+
+            my_hash_insert(&cache_data->partial_filtered_table_ids, stored);
+            
+          }
+        }
+      }
+    filter_all_tables= found_filtered_table && !found_unfiltered_table;
```
bnestere: Do we need this `filter_all_tables` variable (and parameter to `write_table_map`? Can we just set `cache_data->event_group_rpl_filter= true` here? Then use that going forward.

sql/log.cc
```diff
@@ -6785,6 +6879,18 @@ bool MYSQL_BIN_LOG::write_table_map(THD *thd, TABLE *table, bool with_annotate)
     }
   });
 
+  if (filter_all_tables)
+  {
+    the_event.flags|= LOG_EVENT_SKIP_REPLICATION_F;
+    cache_data->event_group_rpl_filter= true;
```
bnestere: I think it makes more sense to set `cache_data->event_group_rpl_filter= true;` in the parent function, then 1) we don't redundantly set this for each table map, and 2) we don't need `filter_all_tables` as a parameter at all.

sql/log.cc
```diff
@@ -7042,6 +7162,19 @@ Event_log::prepare_pending_rows_event(THD *thd, TABLE* table,
       flush the pending event and replace it with the newly created
       event...
     */
+    binlog_cache_mngr *cache_mngr= thd->binlog_get_cache_mngr();
+    
+    if (cache_mngr)
+    {
+      binlog_cache_data *cache_data= (cache_mngr->get_binlog_cache_data(is_transactional));
+      /*
+        If the ANNOTATE and TABLE_MAPS have been filtered earlier using binlog_dump_* then
```
bnestere:
```suggestion
        If all the ANNOTATE and TABLE_MAPS have been filtered earlier using binlog_dump_* then 
```

Also the line is too long

sql/log.cc
```diff
@@ -7460,7 +7636,17 @@ bool MYSQL_BIN_LOG::write(Log_event *event_info, my_bool *with_annotate)
...
+        We should gain access to the standalone event's flags to 
+        check whether it was skipped or not since there is only a single 
+        statement following it so if that event has the skip flag then 
+        the current GTID_EVENT should also be skipped to avoid sending 
+        multiple GTIDs with empty event groups which was solved 
+        by MDEV-27697 as explained in the transactional case
+        later MYSQL_BIN_LOG::inside write_gtid_event()
+      */
+      bool should_skip_gtid= (event_info->flags & LOG_EVENT_SKIP_REPLICATION_F);
+      res= write_gtid_event(thd, true, using_trans, commit_id, false, false, false, should_skip_gtid);
```
bnestere: Line too long

mysql-test/suite/rpl/include/rpl_filter_binlog_dump_db.test
MohanadKh03: Right this file is an include one that is kind of a template for storage engines to use , I added it now in the comments of ```rpl_filter_binlog_dump_db/table_STORAGE_ENGINE.test``` files where STORAGE_ENGINE is either innodb/myisam

sql/log_cache.h
MohanadKh03: Resolving this since it was discussed long ago that using a boolean is sufficient instead of using an ENUM

sql/log.cc
MohanadKh03: Resolving this comment as partial filtering is addressed in later commits + reviews

sql/sql_repl.cc
MohanadKh03: Actually this is for partial filtering only , there has to be a way to detect whether the last row event (which has STMT_END_F) should get filtered or not since there is no way to know if this last event should only have its payload cleared and then sent over the network , which is the behavior that should happen ONLY if it's part of a partial filter **and this last row event has the skip flag over it** , or to just not send anything at all , which is the behavior that should happen the event group as a whole got filtered 

First case if it's part of a partial filter then if at least one event gets sent over the network this flag will be true indicating that if the last row event (which has STMT_END_F) does have the skip flag then it should have its payload cleared , otherwise if ALL events (or if there is no events before this last event at all) then by default this flag is false and would just get filtered as a whole and not partially filtered 

BTW the resetting part is per-event-group because we must make sure this flag is true only in a single event group (e.g: if there are multiple row events in the event group then if only one of them gets sent over the network we would be sure that this last row event would be part of a partial filtering and therefore we need a way to know it will not get filtered and we just clear the payload)

sql/log_cache.h
MohanadKh03: Yes I am assuming the table_id never changes but now that you mention I don't think the optimization would matter much though since it's just 8 bytes inserted each time so yeah it's better to just free it up

sql/sql_repl.cc
MohanadKh03: If the post header metadata are cleared the slave wouldnt know if this event has the STMT_END_F flag so I think we should clear the actual row changes themselves not the metadata , I tried both options and clearing the post headers caused the slave to crash since it expected the last row event to contain STMT_END_F

sql/sql_repl.cc
MohanadKh03: I think mathematically both should be the same ? + nvm it turns out the problem was using len which is the packet length although we need the event length not the packet length **this can be gotten by using ```size_t event_len= uint4korr(buf + EVENT_LEN_OFFSET);```**
 that's what was causing the issue but now it's all fine although I dont think it would matter if we do crc over header then the payload both would have the same output

sql/sql_repl.cc
MohanadKh03: I would agree but for some reason the OPTION_SKIP_REPLICATION doesnt seem to be set each time , if i remove the filtering rules check it would fail the replication filtering so we would need to check over it here

bnestere: That sounds like a bug in the setting of `OPTION_SKIP_REPLICATION` :sweat_smile: I'd debug through the cases when it isn't set, but should be, and see why it isn't.

sql/sql_repl.cc
bnestere: Oh right, good catch! I overlooked that `STMT_END_F` was in those `m_flags`. Perhaps you can still reset those fields and just set `STMT_END_F`? It'd still be best practice to clear everything we can.

sql/sql_repl.cc
bnestere: Yeah I would have also thought it should have been the same, but sometimes these libraries have bugs where it isn't :innocent:. Glad you figured it out though

sql/sql_repl.cc
MohanadKh03: https://github.com/MariaDB/server/blob/5ed3668890390060526434b92271756392d77869/sql/log_event_server.cc#L584-L586

Aside from binlog dump filter rules the only way the master could have known how to filter would be using this option that is only set when setting ```@@skip_replication``` to true , so technically it's not a bug of setting ```OPTION_SKIP_REPLICATION``` since it shouldnt be set if one is using binlog dump filters since these dump rules/filters would be useless because all events from the master will get filtered regardless of the rules 

I think that we should stick with checking for either the @@skip_replication option OR if there is any binlog dump filtering rules exist **basically I mean we should leave the changes above as it is**  instead of trying to set the ```OPTION_SKIP_REPLICATION``` , WDYT ? @bnestere

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,130 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+    !binlog_dump_filter->is_db_empty() || binlog_dump_filter->is_on())
   {
+
     uint16 event_flags= uint2korr(&((*packet)[FLAGS_OFFSET + ev_offset]));
 
-    if (event_flags & LOG_EVENT_SKIP_REPLICATION_F)
-      return NULL;
+    const uchar *buf= (const uchar*) packet->ptr() + ev_offset;
+
+      Log_event_type etype= (Log_event_type) buf[EVENT_TYPE_OFFSET];
```
bnestere: the indentation here is broken until line 2277, and the event type is also passed in as a parameter to this function as `event_type`, so you don't need to re-find it.

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,130 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        if the current event has the STMT_F flag as well as the
+        LOG_EVENT_SKIP_REPLICATION_F flag then we should clear the buffer
+        data of this event and rewrite this *_ROW_EVENT to avoid sending
+        filtered data to the replica otherwise if it does not have
+        the STMT_F flag then it should just get filterd 
+        (return NULL basically filters and does not send the event
+        over the network)
+      */
+
+      if (!is_rows_event) {
```
bnestere: if starting bracket should be on new line

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,130 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+      uint16 m_flags= uint2korr(post_start);
+
+      bool has_stmt_end_f= (m_flags & Rows_log_event::STMT_END_F);
+
+
+      if (!has_stmt_end_f)
+      {
+        return NULL;
+      }
+      else {
```
bnestere: else starting bracket should be on new line (and everywhere else too, I'll stop commenting about this)

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,130 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        // skip 6 bytes
+        post_start+= RW_FLAGS_OFFSET;
+      }
+
+      uint16 m_flags= uint2korr(post_start);
+
+      bool has_stmt_end_f= (m_flags & Rows_log_event::STMT_END_F);
+
+
+      if (!has_stmt_end_f)
```
bnestere: For easy comparisons like this, it is easier for others to read to just have the comparison:

```suggestion
if (!(m_flags & Rows_log_event::STMT_END_F))
```

Then there is also weird spacing above

sql/sql_repl.cc
bnestere: Ah I misunderstood what this check was doing. The check for `(info->thd->variables.option_bits & OPTION_SKIP_REPLICATION)` is only the server option, I had it in my head it was _also_ the check for `(event_flags & LOG_EVENT_SKIP_REPLICATION_F)` (i.e. the actual binlog-flagged option). And I also forgot that the `Rpl_filter` API is confusing :( I would think `is_on()` covers both db and table rules, but it doesn't.

So then you are right. It should be ok (I'll file something to clean up the `Rpl_filter` API for another time, I think).

sql/sql_repl.cc
bnestere: Ping, this is still relevant

sql/log_cache.h
```diff
@@ -301,6 +322,25 @@ class binlog_cache_data
...
+    Boolean used to track whether all events inside the transaction cache
+    have the LOG_EVENT_SKIP_REPLICATION_F flag. This determines
+    whether the entire transaction/event group can be skipped when
+    writing to the binary log.
+
+    - true: All cached events have the LOG_EVENT_SKIP_REPLICATION_F flag.
+    - false: At least one cached event does not have the skip flag,
+             so the transaction must not be skipped.
+  */
+  bool event_group_rpl_filter= false;
```
bnestere: `event_group_rpl_filter` reads like there is some `rpl_filter` _for_ some current `event_group` context. I'd suggest a name that makes it clear this is a boolean, e.g. `entire_event_group_has_skip_rpl_f` (though that's pretty lengthy, maybe you can come up with a more concise name :))

sql/log_cache.h
```diff
@@ -301,6 +322,25 @@ class binlog_cache_data
...
+
+  /*
+    Hash set that tracks table_ids from TABLE_MAP_EVENTs which are subject
+    to partial filtering. It is later used to identify matching *_ROWS_EVENTs 
+    that reference those same table_ids.
+  */
+  HASH partial_filtered_table_ids;
```
bnestere: I think `partial_` in the variable name here seems a bit redundant. I think simply `filtered_table_ids` would be more concise. _Or perhaps, because this is in the log_cache file, `binlog_dump_filtered_table_ids` would be more clear too, because the context isn't situationally obvious (as opposed to if we were in `sql_repl.cc`, which specifically deals with the binlog_dump thread)_

sql/log.cc
```diff
@@ -6691,6 +6711,91 @@ bool THD::binlog_write_table_maps()
...
+            WRITE_ROWS_EVENT , if the first one is the only one
+             that has to be filtered then there has to be a way
+             to know from the table_id of WRITE_ROWS_EVENT 
+          */
+          bool should_replicate= should_replicate_db && should_replicate_table;
+          if(should_replicate)
+          {
+            found_unfiltered_table= true;
+          }
+          else {
```
bnestere:
```suggestion
          else
          {
```

sql/log.cc
```diff
@@ -6691,6 +6711,91 @@ bool THD::binlog_write_table_maps()
...
+            
+          }
+        }
+      }
+
+    /*
+      In case all tables are filtered then the whole event group
+      should get filtered
+    */
+    if(found_filtered_table && !found_unfiltered_table)
```
bnestere: I don't think you need two variables to track this. If you assume `found_filtered_table= true` from the start, any time you find something that _isn't_ filtered, you'd wouldn't be able to set `cache_data->event_group_rpl_filter= true`. 

In other words / naming conventions...

```
  all_events_filtered= true;
  foreach (TABLE t : tables)
    if (should_replicate_db && should_replicate_table)
      all_events_replicate= false;
  cache_data->event_group_rpl_filter= all_events_replicate;
```

Technically you don't even need the intermediary `all_events_replicate` :)

sql/log.cc
```diff
@@ -6659,6 +6666,19 @@ bool THD::binlog_write_annotated_row(Log_event_writer *writer)
     DBUG_RETURN(0);
 
   Annotate_rows_log_event anno(this, 0, false);
+
+  bool is_transactional= this->variables.option_bits & OPTION_GTID_BEGIN;
```
bnestere: The `is_transactional` check also needs to consider if the table itself is transactional - but this check is actually done in the parent function, `MYSQL_BIN_LOG::write_table_map()`, I'd just pass it in as a parameter here.

sql/log.cc
```diff
@@ -6691,6 +6711,91 @@ bool THD::binlog_write_table_maps()
...
+        }
+      }
+
+    /*
+      In case all tables are filtered then the whole event group
+      should get filtered
+    */
+    if(found_filtered_table && !found_unfiltered_table)
+    {
+      bool is_transactional= this->variables.option_bits & OPTION_GTID_BEGIN;
```
bnestere: _As we've already discussed offline, this check isn't complete, but I'm just mentioning it again for records sake_

sql/log.cc
```diff
@@ -6691,6 +6711,91 @@ bool THD::binlog_write_table_maps()
...
+          }
+          else {
+            found_filtered_table= true;
+            bool is_transactional= table->file->row_logging_has_trans
+                      || (this->variables.option_bits & OPTION_GTID_BEGIN);
+            binlog_cache_mngr *const cache_mngr= this->binlog_get_cache_mngr();
+            binlog_cache_data *cache_data= (cache_mngr->
+                                  get_binlog_cache_data(is_transactional));
+            ulonglong table_id= table->s->table_map_id;
+            ulonglong *stored = (ulonglong*) alloc_root(this->mem_root,
```
bnestere: Another thing we discussed offline that I'm just noting for records sake: this doesn't work in stored procedures, as it overrides the `mem_root` (though I was the one who told you to use `mem_root` in the first place, so the blame lies on me :innocent:). Instead, here I think we can just insert `table->s->table_map_id` directly into the map, and then when clearing the map, _don't_ free the memory (because the `TABLE_SHARE` is shared globally).

bnestere (changes requested):
Hi @MohanadKh03 !

I've let some minor notes on the PR, along with things we've already discussed offline. We also need to make this work with semi-sync (i.e. extending the case of MDEV-38486 here). 

And I really appreciate your continued diligence on this feature, it has been a long journey, but the end is in-sight! Thank you :)

sql/sql_repl.cc
MohanadKh03: Oh I forgot to reply to this my bad , right the ```table_id``` is not used ill remove it but ```post_start``` is used to get the position of m_flags which is used to know whether there is an existing STMT_F over the ROWS_EVENT or not

sql/log.cc
```diff
@@ -7619,11 +7822,12 @@ bool MYSQL_BIN_LOG::write(Log_event *event_info, my_bool *with_annotate)
       mysql_mutex_assert_owner(&LOCK_after_binlog_sync);
       mysql_mutex_assert_not_owner(&LOCK_commit_ordered);
 #ifdef HAVE_REPLICATION
-      if (repl_semisync_master.wait_after_sync(log_file_name, offset))
-      {
-        error=1;
-        /* error is already printed inside hook */
-      }
+        if (!(event_info->flags & LOG_EVENT_SKIP_REPLICATION_F) &&
```
bnestere: Indentation here is broken

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,123 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        return NULL;
+      }
+      else
+      {
+        if (!info->sent_unfiltered_row_event)
+        {
+          return NULL;
+        }
+
+        size_t checksum_len= (info->current_checksum_alg
```
bnestere: This logic to clear the payload of a `Rows_log_event` would be good as a separate function.

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,123 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        (event_type == UPDATE_ROWS_EVENT)       ||
+        (event_type == DELETE_ROWS_EVENT)       ||
+        (event_type == WRITE_ROWS_EVENT_V1)     ||
+        (event_type == UPDATE_ROWS_EVENT_V1)    ||
+        (event_type == DELETE_ROWS_EVENT_V1);
+    bool is_gtid_event= (event_type == GTID_EVENT);
+
+    if (is_gtid_event)
+    {
+      info->sent_unfiltered_row_event= false; // reset per event group
```
bnestere, line 2276: I think the amount of brackets in this code area for single-line statements makes it hard to read  (because it adds so much space). I'd just do 

```
    if (is_gtid_event)
      info->sent_unfiltered_row_event= false; // reset per event group
```

for the many new conditionals in this block.

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,123 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        {
+          uint32 crc= my_checksum(0, buf,  event_len - BINLOG_CHECKSUM_LEN);
+          uchar *footer= (uchar*)buf + (event_len - BINLOG_CHECKSUM_LEN);
+          
+          int4store(footer, crc);
+        }
+
+        // reset per ROW_EVENT with stmt_f
+        info->sent_unfiltered_row_event= false;
+
```
bnestere: remove this line

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,123 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        memset(data_start, 0, data_len);
+        uint16 new_m_flags= Rows_log_event::STMT_END_F;
+        int2store(post_start, new_m_flags);
+
+
+        if (checksum_len > 0) 
+        {
+          uint32 crc= my_checksum(0, buf,  event_len - BINLOG_CHECKSUM_LEN);
+          uchar *footer= (uchar*)buf + (event_len - BINLOG_CHECKSUM_LEN);
+
```
bnestere: remove this line

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,123 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+          slave will be checking for it to make sure the event group ends with
+          this flag 
+          
+          After modifying the payload the checksum is recomputed
+        */
+        memset(data_start, 0, data_len);
+        uint16 new_m_flags= Rows_log_event::STMT_END_F;
+        int2store(post_start, new_m_flags);
+
+
```
bnestere: remove this line (one is ok, two makes it harder to read from too much space)

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,123 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+      */
+
+      if (!is_rows_event)
+      {
+        return NULL;
+      }
+
+      /*
+        Extract the STMT_END_F of the rows_event flag
+      */
```
bnestere: This could probably also be a function to make it easier to read the block/flow overall. E.g., some static function, like `is_stmt_end_f()` in the `Rows_log_event` class that takes in the raw buffer data and returns a boolean. You probably don't need spaces between each line either :)

sql/sql_repl.cc
```diff
@@ -2232,15 +2250,123 @@ send_event_to_slave(binlog_send_info *info, Log_event_type event_type,
...
+        (event_type == WRITE_ROWS_EVENT_V1)     ||
+        (event_type == UPDATE_ROWS_EVENT_V1)    ||
+        (event_type == DELETE_ROWS_EVENT_V1);
+    bool is_gtid_event= (event_type == GTID_EVENT);
+
+    if (is_gtid_event)
+    {
+      info->sent_unfiltered_row_event= false; // reset per event group
+    }
+
```
bnestere, line 2277: There is bad spacing here. GitHub viewer won't show it, but if you examine the patch contents in `git diff`, it will highlight all bad spacing in red blocks.

There are also other places in your patch with bad spacing like here. Please go over the whole patch contents via `git diff` on the command line (`.result` files exempt as those are generated and wouldn't have such spacing).

PR #4088, opened 2025-06-04
MDEV-35163  InnoDB persistent statistics fail to update after ALTER TABLE...ALGORITHM=COPY

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-35163*

## Description
Problem:
=======
- InnoDB statistics calculation for the table is done after every 10 seconds by default in background thread dict_stats_thread()

- Doing multiple ALTER TABLE..ALGORITHM=COPY causes the dict_stats_thread() to lag behind, therefore calculation of stats for newly created intermediate table gets delayed

Fix:
====
- Stats calculation for newly created intermediate table is made independent of background thread. After copying gets completed, stats for new table is calculated as part of ALTER TABLE ... ALGORITHM=COPY.

dict_stats_rename_table(): Rename the table statistics from temporary table to non-temporary table

This is a cherry-pick fix of mysql commit@cfe5f287ae99d004e8532a30003a7e8e77d379e3

## How can this PR be tested?
./mtr innodb.alter_copy_stats

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb_fts/t/versioning.test
+++ b/mysql-test/suite/innodb_fts/t/versioning.test
@@ -65,6 +65,9 @@ if ($MTR_COMBINATION_UPGRADE)
 {
 --disable_query_log
 call mtr.add_suppression("InnoDB: Table `mysql`.\`innodb_(table|index)_stats`");
+call mtr.add_suppression("InnoDB: Table mysql\\.innodb_table_stats is not readable");
+call mtr.add_suppression("InnoDB: Fetch of persistent statistics requested for table `test`\\.`articles[0-9]*` but the required system tables mysql.innodb_table_stats and mysql.innodb_index_stats are not present or have unexpected structure. Using transient stats instead.");
+call mtr.add_suppression("InnoDB: Recalculation of persistent statistics requested for table `test`\\.`#sql-alter.*` but the required persistent statistics storage is not present or is corrupted. Using transient stats instead.");
 --enable_query_log
   --source include/shutdown_mysqld.inc
   --exec rm -f $datadir/test/*.ibd $datadir/ib*
--- a/storage/innobase/dict/dict0stats.cc
+++ b/storage/innobase/dict/dict0stats.cc
@@ -380,7 +380,15 @@ dict_table_schema_check(
 		return DB_TABLE_NOT_FOUND;
 	}
 
-	if (!table->is_readable() && !table->space) {
+	if (!table->is_readable()) {
+		/* table is not readable */
+		snprintf(errstr, errstr_sz,
+			 "Table %s is not readable.",
+			 req_schema->table_name_sql);
+		return DB_ERROR;
+	}
+
+	if (!table->space) {
 		/* missing tablespace */
 		snprintf(errstr, errstr_sz,
 			 "Tablespace for table %s is missing.",
@@ -3921,8 +3929,9 @@ dberr_t dict_stats_rename_table(const char *old_name, const char *new_name,
   dict_fs2utf8(old_name, old_db, sizeof old_db, old_table, sizeof old_table);
   dict_fs2utf8(new_name, new_db, sizeof new_db, new_table, sizeof new_table);
 
-  if (dict_table_t::is_temporary_name(old_name) ||
-      dict_table_t::is_temporary_name(new_name))
+  /* Delete the stats only if renaming the table from old table to
+  intermediate table during COPY algorithm */
+  if (dict_table_t::is_temporary_name(new_name))
   {
     if (dberr_t e= dict_stats_delete_from_table_stats(old_db, old_table, trx))
       return e;
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -15787,16 +15787,17 @@ ha_innobase::extra(
 	/* Warning: since it is not sure that MariaDB calls external_lock()
 	before calling this function, m_prebuilt->trx can be obsolete! */
 	trx_t* trx;
+	THD* thd = ha_thd();
 
 	switch (operation) {
 	case HA_EXTRA_FLUSH:
-		(void)check_trx_exists(ha_thd());
+		(void)check_trx_exists(thd);
 		if (m_prebuilt->blob_heap) {
 			row_mysql_prebuilt_free_blob_heap(m_prebuilt);
 		}
 		break;
 	case HA_EXTRA_RESET_STATE:
-		trx = check_trx_exists(ha_thd());
+		trx = check_trx_exists(thd);
 		reset_template();
 		trx->duplicates = 0;
 	stmt_boundary:
@@ -15804,23 +15805,23 @@ ha_innobase::extra(
 		trx->bulk_insert = false;
 		break;
 	case HA_EXTRA_NO_KEYREAD:
-		(void)check_trx_exists(ha_thd());
+		(void)check_trx_exists(thd);
 		m_prebuilt->read_just_key = 0;
 		break;
 	case HA_EXTRA_KEYREAD:
-		(void)check_trx_exists(ha_thd());
+		(void)check_trx_exists(thd);
 		m_prebuilt->read_just_key = 1;
 		break;
 	case HA_EXTRA_KEYREAD_PRESERVE_FIELDS:
-		(void)check_trx_exists(ha_thd());
+		(void)check_trx_exists(thd);
 		m_prebuilt->keep_other_fields_on_keyread = 1;
 		break;
 	case HA_EXTRA_INSERT_WITH_UPDATE:
-		trx = check_trx_exists(ha_thd());
+		trx = check_trx_exists(thd);
 		trx->duplicates |= TRX_DUP_IGNORE;
 		goto stmt_boundary;
 	case HA_EXTRA_NO_IGNORE_DUP_KEY:
-		trx = check_trx_exists(ha_thd());
+		trx = check_trx_exists(thd);
 		trx->duplicates &= ~TRX_DUP_IGNORE;
 		if (trx->is_bulk_insert()) {
 			/* Allow a subsequent INSERT into an empty table
@@ -15829,11 +15830,11 @@ ha_innobase::extra(
 		}
 		goto stmt_boundary;
 	case HA_EXTRA_WRITE_CAN_REPLACE:
-		trx = check_trx_exists(ha_thd());
+		trx = check_trx_exists(thd);
 		trx->duplicates |= TRX_DUP_REPLACE;
 		goto stmt_boundary;
 	case HA_EXTRA_WRITE_CANNOT_REPLACE:
-		trx = check_trx_exists(ha_thd());
+		trx = check_trx_exists(thd);
 		trx->duplicates &= ~TRX_DUP_REPLACE;
 		if (trx->is_bulk_insert()) {
 			/* Allow a subsequent INSERT into an empty table
@@ -15842,7 +15843,7 @@ ha_innobase::extra(
 		}
 		goto stmt_boundary;
 	case HA_EXTRA_BEGIN_ALTER_COPY:
-		trx = check_trx_exists(ha_thd());
+		trx = check_trx_exists(thd);
 		m_prebuilt->table->skip_alter_undo = 1;
 		if (m_prebuilt->table->is_temporary()
 		    || !m_prebuilt->table->versioned_by_id()) {
@@ -15855,7 +15856,7 @@ ha_innobase::extra(
 			.first->second.set_versioned(0);
 		break;
 	case HA_EXTRA_END_ALTER_COPY:
-		trx = check_trx_exists(ha_thd());
+		trx = check_trx_exists(thd);
 		if (!m_prebuilt->table->skip_alter_undo) {
 			/* This could be invoked inside INSERT...SELECT.
 			We do not want any extra log writes, because
@@ -15874,6 +15875,7 @@ ha_innobase::extra(
 			handler::extra(HA_EXTRA_BEGIN_ALTER_COPY). */
 			log_buffer_flush_to_disk();
 		}
+		alter_stats_rebuild(m_prebuilt->table, thd);
 		break;
 	default:/* Do nothing */
 		;
@@ -21290,4 +21292,23 @@ ulint buf_pool_size_align(ulint size) noexcept
   } else {
     return (ulint)((size / m + 1) * m);
   }
+}
+
+/** Adjust the persistent statistics after rebuilding ALTER TABLE.
+Remove statistics for dropped indexes, add statistics for created indexes
+and rename statistics for renamed indexes.
+@param table InnoDB table that was rebuilt by ALTER TABLE
+@param thd   alter table thread */
+void alter_stats_rebuild(dict_table_t *table, THD *thd)
+{
+  DBUG_ENTER("alter_stats_rebuild");
+  if (!table->space || !dict_stats_is_persistent_enabled(table))
+    DBUG_VOID_RETURN;
+
+  dberr_t ret= dict_stats_update(table, DICT_STATS_RECALC_PERSISTENT);
+  if (ret != DB_SUCCESS)
+    push_warning_printf(thd, Sql_condition::WARN_LEVEL_WARN,
+                        ER_ALTER_INFO, "Error updating stats for table after"
+                        " table rebuild: %s", ut_strerr(ret));
+  DBUG_VOID_RETURN;
 }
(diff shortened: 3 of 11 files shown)


storage/innobase/dict/dict0stats.cc
```diff
@@ -3921,8 +3921,10 @@ dberr_t dict_stats_rename_table(const char *old_name, const char *new_name,
...
   dict_fs2utf8(new_name, new_db, sizeof new_db, new_table, sizeof new_table);
 
-  if (dict_table_t::is_temporary_name(old_name) ||
-      dict_table_t::is_temporary_name(new_name))
+  bool is_old_temp= dict_table_t::is_temporary_name(old_name);
+  bool is_new_temp= dict_table_t::is_temporary_name(new_name);
+  bool old_to_new= is_old_temp && !is_new_temp;
+  if (!old_to_new && (is_old_temp || is_new_temp))
   {
     if (dberr_t e= dict_stats_delete_from_table_stats(old_db, old_table, trx))
```
mariadb-DebarunBanerjee: > `if (!old_to_new && (is_old_temp || is_new_temp))`

Looks like the condition translates to simply ...
`if (is_new_temp)`

So, we delete stat data if and only if new table is temp(intermediate table) table. Can you please confirm and then simplify the condition ?

mysql-test/suite/innodb_fts/r/versioning.result
```diff
@@ -290,7 +292,10 @@ AGAINST ('this' IN NATURAL LANGUAGE MODE);
 drop index idx on articles2;
 Warnings:
 Warning	1082	InnoDB: Table test/articles2 contains 3 indexes inside InnoDB, which is different from the number of indexes 1 defined in the MariaDB 
+Warning	1088	Error updating stats for table 'test/#sql-alter' after table rebuild: Persistent statistics do not exist
 create fulltext index idx on articles2(title, body);
+Warnings:
+Warning	1088	Error updating stats for table 'test/#sql-alter' after table rebuild: Persistent statistics do not exist
 SELECT * FROM articles2 WHERE MATCH (title,body)
```
mariadb-DebarunBanerjee: We are raising warning for an intermediate table created during DDL. This could be confusing for user and need to fix,

sql/sql_table.cc
```diff
@@ -11817,8 +11817,12 @@ copy_data_between_tables(THD *thd, TABLE *from, TABLE *to, bool ignore,
   }
   bulk_insert_started= 0;
   if (!ignore)
+  {
+    bool save_abort_on_warning= thd->abort_on_warning;
+    thd->abort_on_warning= false;
     to->file->extra(HA_EXTRA_END_ALTER_COPY);
-
+    thd->abort_on_warning= save_abort_on_warning;
+  }
```
mariadb-DebarunBanerjee: Please add a comment here as to why would not abort on warning i.e. If the statistics data is missing for some reason we want to allow the operation with an warning returned to user.

mysql-test/suite/innodb_fts/t/versioning.test
```diff
@@ -80,6 +80,8 @@ SELECT * FROM articles WHERE MATCH (title,body)
 AGAINST ('Database' IN NATURAL LANGUAGE MODE);
 
 call mtr.add_suppression("test/articles.? contains 3 indexes inside InnoDB");
+call mtr.add_suppression("updating stats for table .*");
+--replace_regex /#sql-alter-[0-9a-f_\-]*/#sql-alter/
```
mariadb-DebarunBanerjee: I think you need to update `#sql-alter` here and 2 places below using replace_regex as we have changed the message.

mariadb-DebarunBanerjee (approved):
Thanks for the changes. Please check one open comment.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15858,7 +15858,9 @@ ha_innobase::extra(
 			.first->second.set_versioned(0);
 		break;
 	case HA_EXTRA_END_ALTER_COPY:
-		trx = check_trx_exists(ha_thd());
+        {
+		THD *user_thd = ha_thd();
+		trx = check_trx_exists(user_thd);
```
dr-m: The `{` and `}` need to be indented by TAB, just like everything else in this function.

Better yet, if `default: break;` was added to the start of the `switch` statement body, there would be no need to add any braces. Also, there are very many calls to `ha_thd()` in the function. If the goal is to avoid duplicating those calls, we should assign the variable at the start of the function and replace all occurrences of `ha_thd()` with references to the local variable.

storage/innobase/handler/ha_innodb.h
```diff
@@ -942,3 +942,10 @@ ib_push_frm_error(
 @return true if index column length exceeds limit */
 MY_ATTRIBUTE((warn_unused_result))
 bool too_big_key_part_length(size_t max_field_len, const KEY& key);
+
+/** Adjust the persistent statistics after rebuilding ALTER TABLE.
+Remove statistics for dropped indexes, add statistics for created indexes
+and rename statistics for renamed indexes.
+@param table_name Table name in MySQL
+@param thd MySQL connection */
```
dr-m: What is MySQL?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21308,3 +21312,22 @@ ulint buf_pool_size_align(ulint size) noexcept
...
+{
+  DBUG_ENTER("alter_stats_rebuild");
+  if (!table->space || !dict_stats_is_persistent_enabled(table))
+    DBUG_VOID_RETURN;
+
+  dberr_t ret= dict_stats_update(table, DICT_STATS_RECALC_PERSISTENT);
+  if (ret != DB_SUCCESS)
+    push_warning_printf(thd, Sql_condition::WARN_LEVEL_WARN,
+			ER_ALTER_INFO, "Error updating stats for table after"
+                        " table rebuild: %s", ut_strerr(ret));
```
dr-m: Compared to the old implementation of `alter_stats_rebuild()`, this is losing the table name. Is there a particular reason for this? This change is not being mentioned in the commit message either.

Thirunarayanan: Because we may throw warning for intermediate table during alter table. This could be confusing for user. Better
to mention that warning without any intermediate table. Do you have any suggestion?

dr-m: Assuming that `ALTER TABLE` is a rarely executed operation, there is a good chance that the user would know which table they were trying to alter when the error was being reported. In fact, the `push_warning_printf()` will be sent to the client connection that executed the `ALTER TABLE`. The messages says "error", though, and users can disable warnings. Should we perhaps send this also to the error log?

What the user can’t easily know is the name of the intermediate table that was assigned during `ALTER TABLE`. It could be relevant information for troubleshooting or for cleanup, in case the statistics for that intermediate table can’t be dropped either. Based on the message, the user would be able to delete the garbage statistics later.

Other than avoiding some potential confusion, is there a reason for omitting the detail?

mariadb-DebarunBanerjee: My thoughts. 

1. The internal/intermediate table name is good information in server error log - may be as a NOTE as it is not a warning at server level.
2. For sending warning message to client it is better to refer to the client table name if possible.

dr-m: In some hosted environments, users might not have access to the server error log at all, but they would be able to see the response sent to their SQL statements. Because DDL operations support a single table only, repeating the table name in an error log to the client connection does not make much sense. (Reporting the name of another table, such as one that is not compatible with a `FOREIGN KEY` constraint that was requested to be added, might make sense.)

There is a known problem with `alter_stats_rebuild()`: It may consume a lot of time on a large table. When it is being executed in `ALTER TABLE…ALGORITHM=INPLACE`, we’re holding `MDL_EXCLUSIVE` on the table name. This is rather disruptive, because it could block the execution of a statement like `SHOW TABLES` for several minutes.

The commit message fails to mention it, but an important part of the change is that `alter_stats_rebuild()` is no longer `static`. It used to be exclusively for the table-rebuild code path of `ALTER TABLE…ALGORITHM=INPLACE`. The change will invoke it also at the end of `ALTER TABLE…ALGORITHM=COPY`. Until now, this type of operation should have triggered a recalculation of the InnoDB persistent statistics, similar to a large `INSERT` operation. Now, we are adding an extra step of scanning through the table after completing the operation:
```
#0  dict_stats_update (table=0x781f340828f8, stats_upd_option=DICT_STATS_RECALC_PERSISTENT) at /home/thiru/server/10.6/storage/innobase/dict/dict0stats.cc:3637
#1  0x00005cfda97f61b0 in alter_stats_rebuild (table=0x781f340828f8, thd=0x781f34001ea8) at /home/thiru/server/10.6/storage/innobase/handler/ha_innodb.cc:21325
#2  0x00005cfda97ebe66 in ha_innobase::extra (this=0x781f3407e7d8, operation=HA_EXTRA_END_ALTER_COPY) at /home/thiru/server/10.6/storage/innobase/handler/ha_innodb.cc:15881
#3  0x00005cfda909899e in copy_data_between_tables (thd=0x781f34001ea8, from=0x781f3414bfb8, to=0x781f34083a18, ignore=0x0, order_num=0x0, order=0x0, copied=0x781f65879d38, deleted=0x781f65879d40, 
    alter_info=0x781f6587c0e0, alter_ctx=0x781f6587b390) at /home/thiru/server/10.6/sql/sql_table.cc:11826
#4  0x00005cfda9095b2f in mysql_alter_table (thd=0x781f34001ea8, new_db=0x781f34006b50, new_name=0x781f34006fc0, create_info=0x781f6587c250, table_list=0x781f340168b8, recreate_info=0x781f6587c0a0, 
    alter_info=0x781f6587c0e0, order_num=0x0, order=0x0, ignore=0x0, if_exists=0x0) at /home/thiru/server/10.6/sql/sql_table.cc:11011
#5  0x00005cfda915763b in Sql_cmd_alter_table::execute (this=0x781f34016fc8, thd=0x781f34001ea8) at /home/thiru/server/10.6/sql/sql_alter.cc:675
#6  0x00005cfda8f771db in mysql_execute_command (thd=0x781f34001ea8, is_called_from_prepared_stmt=0x0) at /home/thiru/server/10.6/sql/sql_parse.cc:6167
#7  0x00005cfda8f7d706 in mysql_parse (thd=0x781f34001ea8, rawbuf=0x781f340167c0 "alter table t1 force, algorithm=copy", length=0x24, parser_state=0x781f6587d370)
    at /home/thiru/server/10.6/sql/sql_parse.cc:8200
```
It seems to me that we are only holding a shared MDL at this point of time. But, I think that we’d need a further change to ensure that `dict_stats_recalc_pool_add()` cannot ever be invoked on the table while it is being copied:
```
#0  dict_stats_update_if_needed (table=0x781f340828f8, trx=...) at /home/thiru/server/10.6/storage/innobase/dict/dict0stats_bg.cc:138
#1  0x00005cfda9a09189 in row_insert_for_mysql (mysql_rec=0x781f3407efd8 "\377\001", prebuilt=0x781f340873a8, ins_mode=ROW_INS_NORMAL) at /home/thiru/server/10.6/storage/innobase/row/row0mysql.cc:1395
#2  0x00005cfda97d5880 in ha_innobase::write_row (this=0x781f3407e7d8, record=0x781f3407efd8 "\377\001") at /home/thiru/server/10.6/storage/innobase/handler/ha_innodb.cc:7849
#3  0x00005cfda9366384 in handler::ha_write_row (this=0x781f3407e7d8, buf=0x781f3407efd8 "\377\001") at /home/thiru/server/10.6/sql/handler.cc:7802
#4  0x00005cfda90984e2 in copy_data_between_tables (thd=0x781f34001ea8, from=0x781f3414bfb8, to=0x781f34083a18, ignore=0x0, order_num=0x0, order=0x0, copied=0x781f65879d38, deleted=0x781f65879d40, 
    alter_info=0x781f6587c0e0, alter_ctx=0x781f6587b390) at /home/thiru/server/10.6/sql/sql_table.cc:11738
```
Now, I think that the interesting errors would be those where `dict_stats_save()` is failing. If it fails, it looks like the transaction for updating the statistics tables will always be rolled back, if anything was written in the first place. In some cases we would also display the internal name of the table in the error log.

So, I retract my request and agree that it does not make any sense to display the exact name of the table to the user. It would not be left over as garbage in the InnoDB statistics tables. What matters is that the statistics could not be updated during the `ALTER TABLE` operation.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -21308,3 +21310,22 @@ ulint buf_pool_size_align(ulint size) noexcept
...
```
dr-m: The second line of the `push_warning_printf()` is indented by TABs instead of spaces. Please use uniform formatting.

dr-m (commented):
The commit message needs to mention that the `copy_data_between_tables()` code path is invoking `alter_stats_rebuild()` under a shared MDL, on an `#sql-alter-` table name. Previously this function had been invoked only during a table-rebuilding `ALGORITHM=INPLACE` operation, using the user-visible table name.

Please make sure that `dict_stats_update_if_needed()` will be a no-op during `copy_data_between_tables()`. We do not want the statistics to be prematurely calculated while the table is being rebuilt.

sql/sql_insert.cc
```diff
@@ -4379,7 +4379,14 @@ bool select_insert::prepare_eof()
   if (info.ignore || info.handle_duplicates != DUP_ERROR)
       if (table->file->ha_table_flags() & HA_DUPLICATE_POS)
         table->file->ha_rnd_end();
+
+  /* Don't throw an error during InnoDB DDL if persistent
+  stats doesn't exist while updating the statistics */
+  bool save_abort_on_warning= thd->abort_on_warning;
+  thd->abort_on_warning= false;
   table->file->extra(HA_EXTRA_END_ALTER_COPY);
+  thd->abort_on_warning= save_abort_on_warning;
```
vuvova: Why InnoDB returns an error, if the server is expected to ignore it?

dr-m: As far as I understand, this mimics `mysql_inplace_alter_table()`, which had been modified by you in b2af0ddcf95ade838e67595fd09e9c4a60efbb04. Specifically:
>set `thd->abort_on_warning` for inplace alter, just like it's set for `copy_data_between_tables` - to have warnings converted into errors identically in all alter algorithms

In order to answer the question, @Thirunarayanan must test both `sql_mode=STRICT_ALL_TABLES` as well as a non-strict mode, with different storage engines.

Thirunarayanan:
dict_stats_update_if_needed() is no-op for the intermediate table during copy alter algorithm.
```
        if (dict_stats_is_persistent_enabled(table)) {
                if (table->name.is_temporary()) {
                        return;
                }
```
Since intermediate table has temporary table name, we don't do anything.

mysql-test/suite/innodb_fts/t/versioning.test
```diff
@@ -65,6 +65,9 @@ if ($MTR_COMBINATION_UPGRADE)
 {
 --disable_query_log
 call mtr.add_suppression("InnoDB: Table `mysql`.\`innodb_(table|index)_stats`");
+call mtr.add_suppression("InnoDB: Table mysql.innodb_table_stats is not readable");
+call mtr.add_suppression("InnoDB: Fetch of persistent statistics requested for table `test`.`articles[0-9]*` but the required system tables mysql.innodb_table_stats and mysql.innodb_index_stats are not present or have unexpected structure. Using transient stats instead.");
+call mtr.add_suppression("InnoDB: Recalculation of persistent statistics requested for table `test`.`#sql-alter.*` but the required persistent statistics storage is not present or is corrupted. Using transient stats instead.");
```
dr-m: The meta-character `.` needs to be escaped.

mysql-test/suite/innodb_fts/t/versioning.test
```diff
@@ -80,6 +83,7 @@ SELECT * FROM articles WHERE MATCH (title,body)
 AGAINST ('Database' IN NATURAL LANGUAGE MODE);
 
 call mtr.add_suppression("test/articles.? contains 3 indexes inside InnoDB");
+call mtr.add_suppression("updating stats for table .*");
```
dr-m: Leading or trailing `.*` does not have any effect and had better be omitted.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15787,40 +15787,41 @@ ha_innobase::extra(
 	/* Warning: since it is not sure that MariaDB calls external_lock()
 	before calling this function, m_prebuilt->trx can be obsolete! */
 	trx_t* trx;
+        THD* thd = ha_thd();
```
dr-m: This function has been formatted using TAB, not spaces.

PR #4093, opened 2025-06-06
MDEV-37107 - Optimise dot_product by loop-unrolling by a factor of 4

mikejuliet13:
This patch introduces loop unrolling by a factor of 4 in the
dot_product() function used in vector-based distance calculations.

The goal is to improve SIMD utilization and overall performance during
high-throughput vector operations, particularly in indexing and search
routines that rely on this function.

Observations from benchmarking (ann-benchmark):
- Query Performance (QPS) improved by 4–10% across datasets.
- Indexation time reduced by 22–28%.
- Loop unrolling factors of 8 or 16 yielded similar performance to
  factor-4 but made the code less readable. Hence, a factor of 4 was
  chosen to maintain a balance between performance and code clarity.

This change is architecture-specific (PowerPC) and should not
introduce any behavioural regressions or side effects in unrelated
parts of the codebase.

### DIFF
--- a/sql/vector_mhnsw.cc
+++ b/sql/vector_mhnsw.cc
@@ -241,6 +241,7 @@ struct FVector
     // Round up to process full vector, including padding
     size_t base= ((len + POWER_dims - 1) / POWER_dims) * POWER_dims;
 
+    #pragma GCC unroll 4
     for (size_t i= 0; i < base; i+= POWER_dims)
     {
       vector short x= vec_ld(0, &v1[i]);


mikejuliet13:
Hi @svoj ,

This change improves performance on Power architecture by unrolling the loop by a factor of 4. Kindly review and approve the PR when possible — your feedback would be appreciated.

sql/vector_mhnsw.cc
```diff
@@ -237,28 +237,44 @@ struct FVector
   static float dot_product(const int16_t *v1, const int16_t *v2, size_t len)
   {
     // Using vector long long for int64_t accumulation
-    vector long long ll_sum= {0, 0};
-    // Round up to process full vector, including padding
-    size_t base= ((len + POWER_dims - 1) / POWER_dims) * POWER_dims;
+    vector long long ll_sum1 = {0, 0};
```
svoj: One thing that immediately hits my eyes is coding style: `var = val` should be `var= val`. Same for `var+= val`.

svoj (changes requested):
Please try to avoid merge commits, use rebase if needed. Initially there were none. Now that you have merge commits in your branch it will be harder to amend original patch. Merge commits will have to be reverted.

I'm going to review it more carefully soon.

sql/vector_mhnsw.cc
mikejuliet13: I have made the required changes

mikejuliet13: I’ve squashed the commits and cleaned up the merge commits via interactive rebase, so now the branch has a single clean commit.

mikejuliet13:
> Please try to avoid merge commits, use rebase if needed. Initially there were none. Now that you have merge commits in your branch it will be harder to amend original patch. Merge commits will have to be reverted.
> 
> I'm going to review it more carefully soon.

Hi @svoj,
Thanks for your earlier feedback — I’ve now rebased the branch to remove the merge commits, and the PR contains only a single clean commit on top of main.

Could you please take another look and let me know if it’s good to go

svoj:
@mikejuliet13 looks much better now, thanks! Did you happen to analyse  how exactly loop unrolling improves performance? Did it use different instructions for vector arithmetic? Or was that due to some sort of optimistic execution? Better vector registers utilisation? Or pre-loading of multiple vectors at once did the trick?

I have some suggestions on my mind, but I need to understand what can be changed in your patch and what can't. Alternatively we can take empiristic approach, try my suggestions and see if they provide similar benefits.

mikejuliet13:
> @mikejuliet13 looks much better now, thanks! Did you happen to analyse how exactly loop unrolling improves performance? Did it use different instructions for vector arithmetic? Or was that due to some sort of optimistic execution? Better vector registers utilisation? Or pre-loading of multiple vectors at once did the trick?
> 
> I have some suggestions on my mind, but I need to understand what can be changed in your patch and what can't. Alternatively we can take empiristic approach, try my suggestions and see if they provide similar benefits.

Yes, I did some analysis on where the performance improvement is coming from.

The primary benefit comes from increased parallelism due to loop unrolling. In the original implementation, the compiler ends up using a single register (or vector register) repeatedly for the accumulation — so each multiplication, add operation is serialised through one register.

In my patch, by unrolling the loop by a factor of 4, the compiler can use multiple registers in parallel — one for each unrolled segment. This reduces dependency chains and allows better instruction-level parallelism (ILP). As a result, the same vector instructions are issued multiple times in parallel, which improves throughput significantly.

There’s no change in the type of instructions — it’s not due to new vector instructions, prefetching, or optimistic execution. The same arithmetic instructions are used, just invoked more efficiently across registers, making better use of available execution units.

sql/vector_mhnsw.cc
```diff
@@ -237,28 +237,49 @@ struct FVector
...
+    size_t elements_per_loop= POWER_dims * 4;
 
-      // Extend vector int to vector long long for accumulation
-      vector long long llhi1= vec_unpackh(product_hi);
-      vector long long llhi2= vec_unpackl(product_hi);
-      vector long long lllo1= vec_unpackh(product_lo);
-      vector long long lllo2= vec_unpackl(product_lo);
+    // Round up to nearest multiple of elements_per_loop (with padding)
+    size_t base= ((len + elements_per_loop - 1) / elements_per_loop) *
+                 elements_per_loop;
```
svoj: You overrun buffers here, allocation is `POWER_dims` aligned, not `POWER_dims * 4`. Feels like you have to raise `POWER_bytes= 128 / 8 * 4`.

sql/vector_mhnsw.cc
```diff
@@ -237,28 +237,49 @@ struct FVector
...
 
-      ll_sum+= llhi1 + llhi2 + lllo1 + lllo2;
+    for (size_t i= 0; i < base; i+= elements_per_loop)
```
svoj: It would be good if you could check if either of these gives similar boost with original code:
```
#pragma unroll 4
#pragma GCC unroll 4
#pragma clang loop unroll_count(8)
for (...)
```

sql/vector_mhnsw.cc
```diff
@@ -237,28 +237,49 @@ struct FVector
...
+    {
+      vector short x1= vec_ld(0, &v1[i + 0 * POWER_dims]);
+      vector short y1= vec_ld(0, &v2[i + 0 * POWER_dims]);
+      vector short x2= vec_ld(0, &v1[i + 1 * POWER_dims]);
+      vector short y2= vec_ld(0, &v2[i + 1 * POWER_dims]);
+      vector short x3= vec_ld(0, &v1[i + 2 * POWER_dims]);
+      vector short y3= vec_ld(0, &v2[i + 2 * POWER_dims]);
+      vector short x4= vec_ld(0, &v1[i + 3 * POWER_dims]);
+      vector short y4= vec_ld(0, &v2[i + 3 * POWER_dims]);
```
svoj: It would be good if you could check if moving this code to a separate method gives similar boost:
```C++
static void dot_product_step(const int16_t *v1, const int16_t *v2,
                             int i, vector long long &ll_sum)
{
  vector short x= vec_ld(i, v1);
  vector short y= vec_ld(i, v2);
  vector int hi= vec_mule(x, y);
  vector int lo= vec_mulo(x, y);
  ll_sum+= vec_unpackh(hi) + vec_unpackl(hi) +
           vec_unpackh(lo) + vec_unpackl(lo);
}

...

for (...)
{
  dot_product_step(v1, v2, i + 0 * POWER_dims, ll_sum1);
  dot_product_step(v1, v2, i + 1 * POWER_dims, ll_sum2);
  dot_product_step(v1, v2, i + 2 * POWER_dims, ll_sum3);
  dot_product_step(v1, v2, i + 3 * POWER_dims, ll_sum4);
}
```

svoj (changes requested):
One potential bug outlined.
Need your opinion on the other 2 items.

sql/vector_mhnsw.cc
mikejuliet13: I've updated the alloc_size() and fix_tail() functions to account for POWER_dims * 4, ensuring proper padding for the loop’s access pattern and preventing any potential buffer overrun.
I’ve kept POWER_bytes = 128 / 8 as-is to accurately reflect the actual Power vector width (16 bytes), rather than scaling it up by 4. Defining it as 128 / 8 * 4 might be misleading for future readers since Power vectors are still 128-bit wide — we're just processing 4 of them per loop.

sql/vector_mhnsw.cc
mikejuliet13: Hi @svoj 
I have replaced the manual unrolling with the pragma 
`#pragma GCC unroll 4`
It gave similar performance gains compared to the manually unrolled version. Given this, I’ve updated the code to use #pragma GCC unroll 4 for better readability and maintainability

sql/vector_mhnsw.cc
mikejuliet13: Hi @svoj 
I've now switched to compiler-directed unrolling using #pragma GCC unroll 4 and removed the manual unrolling for better readability and maintainability. With this change, I think there is no longer a need for the separate dot_product_step() function, as the loop is now cleaner and doesn't repeat the same block multiple times.

svoj (changes requested):
The code LGTM, thanks!

Could you also update commit message to something like:
```
MDEV-37107 - Optimise dot_product by loop-unrolling by a factor of 4

This patch introduces loop unrolling by a factor of 4 in the
dot_product() function used in vector-based distance calculations.

The goal is to improve SIMD utilization and overall performance during
high-throughput vector operations, particularly in indexing and search
routines that rely on this function.

Observations from benchmarking (ann-benchmark):
- Query Performance (QPS) improved by 4–10% across datasets.
- Indexation time reduced by 22–28%.
- Loop unrolling factors of 8 or 16 yielded similar performance to
  factor-4 but made the code less readable. Hence, a factor of 4 was
  chosen to maintain a balance between performance and code clarity.

This change is architecture-specific (PowerPC) and should not
introduce any behavioral regressions or side effects in unrelated
parts of the codebase.
```

mikejuliet13:
Hi @svoj,
The commit message has been updated as per your suggestion.
Could you please review and merge the PR when convenient?
Thanks!

PR #4094, opened 2025-06-08
MDEV-18827 Create utility to parse frm files and print their DDL

hp77-creator:
## Add FRM parser as a standalone utility to mariaDB server
- [x] The Jira issue number for this PR is: [MDEV-18827](https://jira.mariadb.org/browse/MDEV-18827)

## Description
The following tests are going to be added and run with mtr:
- [x]  simple columns, at least int, text, varchar(value)
- [x]  keys, `using btree`
- [ ]  keys `using rtree`
- [x]  sinple vcol expressions: `a int, x int as (a *2)` 
- [x]  same for DEFAULT: `a int default (whatever)`
- [x]  DEFAULT CURRENT_TIMESTAMP is a special case
- [x]  field-level comments, table-level comments

==================
- [x]  innodb engine
- [X]  more exotic engines
- [X]  engine options
- [ ] partitions -- currently failing so added with error clause

==================
- [ ] Expressions from plugins, like json, Item_geometry_func, inet6 -- currently failing so added with error clause
- [x] Sequences: default(nextval(s))

==================
- [x] Foreign keys in innodb -- foreign key constraints are not printed properly
## Release Notes

## How can this PR be tested?

run 
```
> ./mtr --suite=client --do-test=mariadb-frm
```

## PR quality check


extra/frm_parser.cc
```diff
@@ -0,0 +1,373 @@
...
+  }
+  if (frm_data)
+  {
+    my_free(frm_data);
+  }
+  if (db_name.str)
+  {
+    my_free((void*)db_name.str);
+  }
+  if (table_name.str)
```
FooBarrior: `free` pattern is designed to work fine with null pointers passed, so the `if` is obsolete

hp77-creator: understood, will add this change

hp77-creator: done, so I will only keep check for `share`

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -20,18 +20,79 @@
...
+#include "mysql_com.h"
+#include "m_ctype.h"
 #include "field.h"
 
 
+class Security_context;
+class LEX;
+class Transaction_state;
+
+class THD {
```
FooBarrior: No, You can't add THD like this. It'll be available only for mariadb_frm.cc. Other modules will not see it like this. Think of what will happen with access to some field if it has a different layout in  another module's function!

FooBarrior (changes requested):

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,531 @@
...
+
+extern int mysqld_server_started;
+extern int mysqld_server_initialized;
+extern int sf_leaking_memory;
+
+
+static handlerton mock_hton;
+
+static st_maria_plugin mock_plugin = {
+  .type = MYSQL_STORAGE_ENGINE_PLUGIN,
```
FooBarrior: Formally, designated initialization is only possible since c++20. I think nothing can stop us to enable it in cmake. Please try to make it, but only for frm_parser!

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,531 @@
...
+  memset(&table_list, 0, sizeof(table_list));
+  printf("DEBUG: table_list initialized\n");
+  fflush(stdout);
+
+  printf("DEBUG: About to read FRM file: %s\n", frm_path);
+  fflush(stdout);
+  
+  frm_data= read_frm_file(frm_path, &frm_length);
+  if (!frm_data)
+    goto cleanup;
```
FooBarrior: We have different opinions about gotos acros our communities, and mine is -- they are completely unnecessary, they make the code harder to read, and they make it impossible to add a variable initialization in-between the function (after a first goto).

Try to rewrite cleanups with the use of RAII. We have an ATEXIT convenience macro for that.
THD uses constructor and destructor, so use  unique_ptr for a THD pointer.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,531 @@
...
+  }
+  
+  printf("DEBUG: Names extracted - db: %.*s, table: %.*s\n", 
+         (int)db_name.length, db_name.str, (int)table_name.length, table_name.str);
+  fflush(stdout);
+
+  printf("DEBUG: Allocating TABLE_SHARE manually\n");
+  fflush(stdout);
+  
+  share = (TABLE_SHARE*)my_malloc(PSI_NOT_INSTRUMENTED, sizeof(TABLE_SHARE), MYF(MY_ZEROFILL));
```
FooBarrior: No need. TABLE_SHARE is small enough to allocate it on stack. You'll get rid of an extra my_free as a bonus

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,531 @@
...
+  {
+    fprintf(stderr, "Error: Cannot parse FRM file - init_from_binary_frm_image failed\n");
+    goto cleanup;
+  }
+  
+  printf("DEBUG: init_from_binary_frm_image completed successfully\n");
+  fflush(stdout);
+  
+
+  table= static_cast<TABLE *>(
```
FooBarrior: same as for TABLE_SHARE

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,531 @@
...
+
+  printf("DEBUG: About to initialize THD...\n");
+  fflush(stdout);
+
+
+  if (init_early_variables() ||
+     init_thread_environment() ||
+     mysql_init_variables() ||
+     init_character_sets() ||
+     init_plugin_system_complete())  // This is the key addition
```
FooBarrior: only mysql_init_variables can really fail. I think it's reasonable to convert the rest of the functions to void and move outside if

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,531 @@
...
+extern int sf_leaking_memory;
+
+
+static handlerton mock_hton;
+
+static st_maria_plugin mock_plugin = {
+  .type = MYSQL_STORAGE_ENGINE_PLUGIN,
+  .info = &mock_hton,
+  .name = "MOCK_ENGINE",
+  .author = "Mock Author",
```
FooBarrior: if mock_plugin survives our refactorings, you can put your name here:)

FooBarrior (changes requested):
Now it looks directed to the right way.

I noticed there are some junk files included into PR -- pls rollback them.
Add a test: create a table with a real server and invoke your frm parser with MTR's `--exec`

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,531 @@
...
+  
+  TABLE_LIST table_list;
+  uchar *frm_data= NULL;
+  size_t frm_length= 0;
+  TABLE_SHARE *share= NULL;
+  TABLE *table= NULL;
+  LEX_CSTRING db_name, table_name;
+  bool error= true;
+  
+  memset(&table_list, 0, sizeof(table_list));
```
FooBarrior: better call a constructor, as soon as you can.

FooBarrior: the non-default one

extra/mariadbfrm/mariadb_frm.cc
hp77-creator: understood, will refactor this.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+    .status_vars= NULL,
+    .system_vars= NULL,
+    .version_info= "1.0",
+    .maturity= MariaDB_PLUGIN_MATURITY_STABLE};
+
+static st_plugin_int mock_plugin_int= {
+    .name= {const_cast<char *>("mock_storage_engine"), 19},
+    .plugin= &mock_plugin,
+    .data= nullptr,
+    .plugin_dl= NULL,
```
FooBarrior: Make sure your initializers are in the declaration order. `plugin_dl` appears before `data`.

Also no need to initialize zeores, since it's the default value.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+st_plugin_int *hton2plugin[MAX_HA];
+
+static st_maria_plugin mock_plugin= {
+    .type= MYSQL_STORAGE_ENGINE_PLUGIN,
+    .info= nullptr,
+    .name= "MOCK_ENGINE",
+    .author= "Nikita Malyavin",
+    .descr= "Mock storage engine for FRM parsing",
+    .license= PLUGIN_LICENSE_GPL,
+    .init= NULL,
```
FooBarrior: Figure what you like more: NULL or nullptr, and rewrite in a chosen style.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+extern int mysqld_server_initialized;
+extern int sf_leaking_memory;
+extern handlerton *get_frm_mock_handlerton();
+st_plugin_int *hton2plugin[MAX_HA];
+
+static st_maria_plugin mock_plugin= {
+    .type= MYSQL_STORAGE_ENGINE_PLUGIN,
+    .info= nullptr,
+    .name= "MOCK_ENGINE",
+    .author= "Nikita Malyavin",
```
FooBarrior: Not true, the author is you ☺️

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+  mysql_mutex_init(key_LOCK_user_conn, &LOCK_user_conn, MY_MUTEX_INIT_FAST);
+  mysql_mutex_init(key_LOCK_thread_id, &LOCK_thread_id, MY_MUTEX_INIT_FAST);
+  mysql_cond_init(key_COND_start_thread, &COND_start_thread, NULL);
+
+  return 0;
+}
+
+static int mysql_init_variables()
+{
+  global_system_variables.character_set_client= &my_charset_utf8mb3_general_ci;
```
FooBarrior: you may like the idea of making a shortcut:
```
auto &gv  = global_system_variables;
gv. ... = ...
...
```

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+static st_plugin_int mock_plugin_int= {
+    .name= {const_cast<char *>("mock_storage_engine"), 19},
+    .plugin= &mock_plugin,
+    .data= nullptr,
+    .plugin_dl= NULL,
+    .state= PLUGIN_IS_READY,
+    .ref_count= 1,
+    .load_option= PLUGIN_ON,
+    .locks_total= 0,
+    .mem_root= {0},
```
FooBarrior: no need to put zero in `{}` in c++

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+{
+  system_charset_info= &my_charset_utf8mb3_general_ci;
+  files_charset_info= &my_charset_utf8mb3_general_ci;
+  national_charset_info= &my_charset_utf8mb3_general_ci;
+  table_alias_charset= &my_charset_bin;
+  character_set_filesystem= &my_charset_bin;
+  default_charset_info= &my_charset_utf8mb3_general_ci;
+
+  return 0;
+}
```
FooBarrior, line 138: This all can be initialized in definitions, as long as you have them. Alternatively, `#include <mysqld.h>` and remove the definitions, since all this is already defined in mysqld.cc, which you do link.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+  mock_plugin_int.data= actual_hton;
+
+  actual_hton->flags= HTON_CAN_RECREATE;
+  actual_hton->db_type= DB_TYPE_BLACKHOLE_DB;
+  actual_hton->slot= 0;
+  hton2plugin[0]= &mock_plugin_int;
+  if (!plugin_array)
+    return 1;
+
+  hton2plugin[0]= &mock_plugin_int;
```
FooBarrior: a duplicate of two lines above

extra/mariadbfrm/frm_mocks.cc
```diff
@@ -0,0 +1,210 @@
...
+  frm_mock_hton_struct.rollback= nullptr;
+  frm_mock_hton_struct.prepare= nullptr;
+  frm_mock_hton_struct.recover= nullptr;
+  frm_mock_hton_struct.commit_by_xid= nullptr;
+  frm_mock_hton_struct.rollback_by_xid= nullptr;
+  frm_mock_hton_struct.drop_database= nullptr;
+  frm_mock_hton_struct.panic= nullptr;
+  frm_mock_hton_struct.start_consistent_snapshot= nullptr;
+  frm_mock_hton_struct.flush_logs= nullptr;
+  frm_mock_hton_struct.show_status= nullptr;
```
FooBarrior: No need to specify those that are nulls, they are such by default

extra/mariadbfrm/frm_mocks.cc
```diff
@@ -0,0 +1,210 @@
...
+enum legacy_db_type ha_checktype(THD *thd, enum legacy_db_type database_type,
+                                 bool no_substitute, bool report_error)
+{
+  return DB_TYPE_MYISAM;
+}
+
+handler *get_new_handler(TABLE_SHARE *share, MEM_ROOT *alloc,
+                         handlerton *db_type)
+{
+  return new (alloc) FRM_Mock_Handler(db_type, share);
```
FooBarrior, line 143: 👍

extra/mariadbfrm/frm_mocks.cc
```diff
@@ -0,0 +1,210 @@
...
+
+plugin_ref ha_lock_engine(THD *thd, const handlerton *hton)
+{
+  return global_system_variables.table_plugin;
+}
+
+enum legacy_db_type ha_checktype(THD *thd, enum legacy_db_type database_type,
+                                 bool no_substitute, bool report_error)
+{
+  return DB_TYPE_MYISAM;
```
FooBarrior: yeah whatever 👍

extra/mariadbfrm/frm_mocks.cc
```diff
@@ -0,0 +1,210 @@
...
+{
+  return new (mem_root) FRM_Mock_Handler(hton, table);
+}
+
+static void init_frm_mock_handlerton()
+{
+  frm_mock_hton_struct.create= frm_mock_create_handler;
+  frm_mock_hton_struct.db_type= DB_TYPE_UNKNOWN;
+
+  frm_mock_hton_struct.flags= HTON_NO_FLAGS;
```
FooBarrior, line 93: Put this all in frm_mock_hton_struct designated initializer.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+extern mysql_cond_t COND_start_thread;
+extern struct system_variables global_system_variables;
+
+PSI_mutex_key key_LOCK_start_thread= 0;
+PSI_mutex_key key_LOCK_status= 0;
+PSI_mutex_key key_LOCK_global_system_variables= 0;
+PSI_mutex_key key_LOCK_user_conn= 0;
+PSI_mutex_key key_LOCK_thread_id= 0;
+PSI_cond_key key_COND_start_thread= 0;
+PSI_mutex_key key_LOCK_plugin= 0;
```
FooBarrior: do we really need these? I think they are already initialized to zeroes in mysqld.cc.

FooBarrior (changes requested):
Overall looks good, I just commented a few creeps about the style, and I think many declarations might be redundand. (I guess that could be left from our past trials with bringing up the minimal linkage...)

You can test it all as follows:
```
create table t (x int); # will be executed by server
 --exec $MARIADB_FRM $MYSQL_TEST_DIR/std_data/frm/table_simple.frm
drop table t;

# etc, etc, etc...
```
to test different storage engine, add in the beginning:
```
# and for example
--source include/have_blackhole.inc
```

What needs to be tested:

*  simple columns, at least int, text, varchar(value)
*  keys, `using btree`, `using rtree`
*  sinple vcol expressions: `a int, x int as (a *2)`
*  same for DEFAULT: `a int default (whatever)`
*  DEFAULT CURRENT_TIMESTAMP is a special case
*  field-level comments, table-level comments

Later on:
*  innodb engine
*  more exotic engines
*  engine options
*  Partitions!!

Later on:
* Expressions from plugins, like json, Item_geometry_func, inet6
* sequences: default(nextval(s))

Later on:
* Foreign keys in innodb

This is the full roadmap remaining, as I see it. I think we won't reach full support for everythink written in time, but at least:
* We shouldn't crash on those cases.
* We need to add an information on what's not supported in the commit message
* A test showing it's not crasing is still needed.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,498 @@
...
+    .ref_count= 1,
+    .load_option= PLUGIN_ON,
+    .locks_total= 0,
+    .mem_root= {0},
+    .system_vars= NULL,
+    .nbackups= 0,
+    .ptr_backup= NULL};
+
+static st_plugin_int *mock_plugin_ptr= &mock_plugin_int;
+static plugin_ref mock_plugin_ref= &mock_plugin_ptr;
```
FooBarrior: You may find that many CI builds fail:
```
/home/buildbot/aarch64-debian-11/build/extra/mariadbfrm/mariadb_frm.cc:110:36: error: cannot convert ‘st_plugin_int**’ to ‘plugin_ref’ {aka ‘st_plugin_int*’} in initialization
  110 | static plugin_ref mock_plugin_ref= &mock_plugin_ptr;
      |                                    ^~~~~~~~~~~~~~~~
      |                                    |
      |                                    st_plugin_int**
```
That's because sometimes `plugin_ref` is defined as `st_plugin_int*`, and sometimes it's `st_plugin_int**`.

To make it right, use `plugin_int_to_ref(pl)` from `sql_plugins.h`

extra/mariadbfrm/mariadb_frm.cc
FooBarrior: alternativelty, 
```
global_system_variables = { 
 // designited initializers
};
```
should work in c++20. Worth checking? 🤔

extra/mariadbfrm/mariadb_frm.cc
hp77-creator: `key_LOCK_thread_id` and `key_LOCK_plugin` are required, others I will remove.

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,550 @@
...
+
+FrmErrorDetails FrmErrorHandler::analyze_default_expression(const Field* field)
+{
+    if (!field)
+        return FrmErrorDetails();
+
+    if (field->default_value && field->default_value->expr)
+    {
+          String default_str;
+          field->default_value->expr->print(&default_str, QT_ORDINARY);
```
FooBarrior: Are you sure this code is working? If the expression->print is normally callable, then it should be perfercly outputtable by show_create_table

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,550 @@
...
+                                   default_str.c_ptr_safe());
+          }
+    }
+    if (field->default_value)
+    {
+        const char* default_expr_str = nullptr;
+        if (field->default_value->expr)
+        {
+            String expr_string;
+            field->default_value->expr->print(&expr_string, QT_ORDINARY);
```
FooBarrior: duplicate of the above??

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+struct UnsupportedPattern
+{
+  const char *pattern;
+  FrmErrorType error_type;
+  const char *component;
+  const char *description;
+  const char *suggestion;
+};
+
+static const UnsupportedPattern UNSUPPORTED_PATTERNS[]= {
```
FooBarrior: It's impossible to enumerate all unsupported functions! There is only list of supported that is strictly defined. You could iterate all the arrays you have an access to and determine if it's supported.

However I think You won't need it if you'll realize how to support an arbitrary function. See the last (global) comment for the idea.

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+  {
+    return FrmErrorDetails(
+        FrmErrorType::CORRUPT_FORMAT, 1, "FRM Header",
+        "FRM file too small or corrupted - invalid header size",
+        "Verify the FRM file is not truncated or corrupted");
+  }
+
+  if (frm_data[0] != 254 || frm_data[1] != 1)
+  {
+    return FrmErrorDetails(
```
FooBarrior: worth having a test for these cases (including one above)

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+
+  if (error.technical_info)
+    fprintf(stderr, "Technical Info: %s\n", error.technical_info);
+
+  fprintf(stderr, "\n");
+}
+
+int FrmErrorHandler::get_exit_code(FrmErrorType type)
+{
+  switch (type)
```
FooBarrior: Function's not needed. Enums are convertible to int

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+  }
+}
+
+bool FrmErrorHandler::contains_unsupported_functions(
+    const char *expr_str, FrmErrorDetails *error_details)
+{
+  if (!expr_str || !error_details)
+    return false;
+
+  for (const UnsupportedPattern *pattern= UNSUPPORTED_PATTERNS;
```
FooBarrior: next time use sets or dictionaries for such tasks, unless it violate some rules.

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+  case DB_TYPE_MYISAM:
+  case DB_TYPE_INNODB:
+  case DB_TYPE_ARIA:
+  case DB_TYPE_ARCHIVE_DB:
+  case DB_TYPE_CSV_DB:
+  case DB_TYPE_HEAP:
+  case DB_TYPE_BLACKHOLE_DB:
+  case DB_TYPE_FEDERATED_DB:
+  case DB_TYPE_MRG_MYISAM:
+    return FrmErrorDetails();
```
FooBarrior: It could be return {};

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+//
+// Private Implementation Methods
+//
+
+FrmErrorDetails
+FrmErrorHandler::check_engine_support(enum legacy_db_type db_type)
+{
+  switch (db_type)
+  {
+  case DB_TYPE_SEQUENCE:
```
FooBarrior: forget about db_type. In theory, your code could be compiled even withou innodb support, then this switch won't help.

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+  if (share->comment.length > 0 && share->comment.str)
+  {
+    const char *comment= share->comment.str;
+    size_t len= share->comment.length;
+    for (size_t i= 0; i < len; i++)
+    {
+      unsigned char ch= (unsigned char) comment[i];
+      if (ch > 127)
+      {
+        return FrmErrorDetails(
```
FooBarrior: It's better to output anything we find how we find, than just failing because charset is 'not supported'. Though we don't know if it's really not supported - maybe it'll output just fine in the user's console. Besides, almost all the users from non-latin countries will be dumped by this check.

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+        uint2korr(frm_data + FRM_EXTRA_REC_BUF_LENGTH_OFFSET);
+    if (extra_rec_buf_length > 100 &&
+        length > static_cast<size_t>(FRM_HEADER_SIZE + extra_rec_buf_length))
+    {
+      const char *extra_data= (const char *) (frm_data + FRM_HEADER_SIZE);
+      size_t search_len= std::min((size_t) extra_rec_buf_length,
+                                  (size_t) 1024); // Limit search
+      for (size_t i= 0; i < search_len - 10; i++)
+      {
+        if (strncmp(extra_data + i, "PARTITION", 9) == 0 ||
```
FooBarrior: the better way would be to just ignore the feature, but I'm sure partitions and other storages are trivially implementable

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+      size_t search_len=
+          std::min((size_t) extra_rec_buf_length, (size_t) 1024);
+
+      for (size_t i= 0; i < search_len - 10; i++)
+      {
+        if (strncmp(extra_data + i, "FOREIGN", 7) == 0 ||
+            strncmp(extra_data + i, "REFERENCE", 9) == 0 ||
+            strncmp(extra_data + i, "CONSTRAINT", 10) == 0)
+        {
+          return FrmErrorDetails(
```
FooBarrior: Why?? Again, you could just ignore the feature.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,633 @@
...
+#include "sql_show.h"
+#include "table_cache.h"
+#include "sql_table.h"
+#include "sql_plugin.h"
+#include "mysqld.h"
+#include "frm_error_handler.h"
+#include <memory>
+#include <functional>
+
+#define DEBUG(fmt, ...)                                                       \
```
FooBarrior, line 41: In final version, DEBUG should be a no-op in release build, at least by default. Or you can implement debug outputting with passing `--debug` if you want

extra/mariadbfrm/frm_error_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+        "unsupported format",
+        "Try recreating the table or exporting with 'SHOW CREATE TABLE'");
+
+  case 2:
+    return FrmErrorDetails(FrmErrorType::UNKNOWN_ERROR, 3, "Memory Allocation",
+                           "Failed to allocate memory for parsing FRM file",
+                           "Check system memory availability");
+
+  case 3:
+    return FrmErrorDetails(
```
FooBarrior: where did you get these error codes from?? All `init_from_binary_frm_image` returns is `HA_ERR_NOT_A_TABLE`!!

FooBarrior (changes requested):
Thank you for your work @hp77-creator!


I really like a steady collection of tests.

About the code - I see some functions handle unexistent error codes, etc, making me think that it's AI generated. If that's so, it should be reverted.

_AI is nothing bad in general, but the author should know what are they doing and validate the output._

For supporting the arbitrary expressions, I suggest you examine the following approach

1. Inherit an own Item, like Frm_parser_item
2. This item will do nothing but output an expression string it was supplied with through a constructor, in Frm_parser_item::print()
3. In unpack_vcol_info_from_frm:
#ifdef FRM_PARSER
return new Virtual_column_info(new Frm_parser_item(expr_str->c_ptr_safe(), expr_str->length())));
#endif
3.1. To make it work, youll have to add table.cc explicitly in the sources list! So that the macro would take effect.
3.2 Alternatively, make a global variable denoting an frm parser mode
3.3 Alternatively again, open_table_from_share->...->unpack_vcol_info_from_frm call chaing with an extra argument, denoting frm parser mode. This is the cleanest solution out of proposed.

Should work!

About the storage engines -- given that we already have a working mockup, an arbitrary engine should just work. The question is only in how to test it. 

For expample, EXAMPLE storage engine can be enabled with supplying `-DWITH_EXAMPLE_STORAGE_ENGINE=1` in the cmake line.

Then have_example_plugin.inc should work in tests after you'll build the target.

hp77-creator:
> About the code - I see some functions handle unexistent error codes, etc, making me think that it's AI generated. If that's so, it should be reverted.

> AI is nothing bad in general, but the author should know what are they doing and validate the output.

100% agree, I didn't validate everything it gave and just pushed it, sorry for that, I will go through the comments and fix whatever's hallucinated.

extra/mariadbfrm/mariadb_frm.cc
hp77-creator: adding mysqld in include led to lot of linker errors so didn't change this much @FooBarrior

extra/mariadbfrm/CMakeLists.txt
```diff
@@ -0,0 +1,84 @@
...
+    )
+    IF(MSVC)
+        TARGET_LINK_OPTIONS(mariadb-frm PRIVATE
+                /INCREMENTAL:NO
+                /FORCE:MULTIPLE
+                /IGNORE:4006
+        )
+    ELSEIF(APPLE)
+    TARGET_LINK_OPTIONS(mariadb-frm PRIVATE
+            "-Wl,-multiply_defined,suppress"
```
FooBarrior: mac build seems to fail. 
I see this: `ld: warning: -multiply_defined is obsolete`
Maybe it's more important which linker is used, rather than a platform? I'm not sure, but maybe you have an access to mac and can check?

hp77-creator: i tried searching `ld` doesn't have replacement for `-multiply_defined`, trying with `-flat_namespace`

RazvanLiviuVarzaru:
Hi @hp77-creator !

`MTR` will complain about `mariadb-frm` when invoked after installing the packages as in: https://buildbot.mariadb.org/#/builders/889/builds/1596

```bash
mysql-test-run: *** ERROR: Could not find any of /usr/extra/mariadbfrm/mariadb-frm /usr/extra/mariadbfrm/mariadb-frm /usr/bin/mariadbfrm/mariadb-frm
```

I don't have a lot of server knowledge, but I'm sure @FooBarrior  can help you with this.

hp77-creator:
Hi @RazvanLiviuVarzaru 
Thanks for noting that, actually I had fixed this, let me see in the latest version if it fails, I have added the path in the post install config to install it in the correct location. Thanks for pointing it out

hp77-creator:
@RazvanLiviuVarzaru  fixed it: https://buildbot.mariadb.org/#/builders/554/builds/14113

hp77-creator:
@FooBarrior shall I rebase all the commits into one for the final merge? All the CI are green now, one that is failing that is not because of my changes and it will pass if its re-triggered again.

FooBarrior:
@hp77-creator yes, please, make a single commit and rebase. Please mention in the commit what is implemented, and what is only tested.

Also mention our approach to reduce the size of the binary (so that there'd be an explanation for the linker flags)

hp77-creator:
@FooBarrior done.

hp77-creator:
Hey @FooBarrior any plans for this PR?

FooBarrior:
@hp77-creator yes it's queued for adoption and merge

gkodinov (changes requested):
This seems to have fallen through and didn't get merged. I'm terribly sorry for that on behalf of the project. 
And I fully intend to get it moving again. 
I'd like to ask you one more thing before we proceed with the testing required: Can you please rebase on main? This is a new feature and it needs to go to that branch.

hp77-creator:
Hi @gkodinov , shall I rebase it from MariaDB/server `main` and then force push the rebased commit? 
Although there's only one commit, I am getting conflicts when I am trying to push the local branch which is rebased with MariaDB/server `main`

hp77-creator:
i also changed the PR's base branch to `main` since `staging` one was 920 commits behind the current main, let me know if something else needs to be done

extra/mariadbfrm/frm_mocks.cc
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  int safe_cond_timedwait(pthread_cond_t *cond, safe_mutex_t *mp,
+                          const struct timespec *abstime, const char *file,
+                          uint line)
+  {
+    return pthread_cond_timedwait(cond, &mp->mutex, abstime);
+  }
+#endif // SAFE_MUTEX
+}
```
gkodinov: please always terminate the last line with a end of line.

gkodinov (changes requested):
You did the rebase correctly. Unfortunately, when doing that UI alone won't do it (afaik). You also need to rebase your branch by yourself. 

Please ensure that the buildbot tests are passing. Right now there's a compilation error:

 ```
/home/buildbot/aarch64-debian-11/build/sql/table.cc: In function ‘bool parse_vcol_defs(THD*, MEM_ROOT*, TABLE*, bool*, vcol_init_mode, bool)’:
/home/buildbot/aarch64-debian-11/build/sql/table.cc:1273:96: error: too many arguments to function ‘Virtual_column_info* unpack_vcol_info_from_frm(THD*, TABLE*, String*, Virtual_column_info**, bool*)’
 1273 |                                     &((*field_ptr)->vcol_info), error_reported, frm_parser_mode);
```

**Comment by gkodinov** [comment]
Do you intend to keep working on this? It's been 3 weeks since I've supplied review comments. Changing to draft due to inactivity. Please change it back when/if you are ready with the next batch.

File: extra/mariadbfrm/frm_mocks.cc (hunk 37 above)
Comment (hp77-creator) [reply extra/mariadbfrm/frm_mocks.cc]: addressed

**Comment by hp77-creator** [comment]
> Do you intend to keep working on this? It's been 3 weeks since I've supplied review comments. Changing to draft due to inactivity. Please change it back when/if you are ready with the next batch.

yes, I inted to get this merged. sorry for delay. I was occupied in some work stuff. will be more active here

**Comment by FooBarrior** [comment]
@hp77-creator there is a (unintended) concatenation of various commit messages in your commit message, please check

**Comment by hp77-creator** [comment]
> @hp77-creator there is a (unintended) concatenation of various commit messages in your commit message, please check

snap! I did multiple git rebase, probably that appended messages of every instant, i will `amend` it, thanks @FooBarrior

**Comment by hp77-creator** [comment]
some tests are failing, converting to draft to resolve those. will make it RFR once done.

**Comment by hp77-creator** [comment]
@gkodinov @FooBarrior please review.
@FooBarrior I have added support for vcols as well and removed conditional compilation for utility by using functional pointers.
Please let me know if this is alright or not.

File: extra/mariadbfrm/CMakeLists.txt [hunk 38]
```diff
@@ -0,0 +1,90 @@
...
+                /IGNORE:4088
+                /IGNORE:4217
+                /WX:NO
+        )
+    ELSEIF(APPLE)
+        TARGET_COMPILE_OPTIONS(mariadb-frm PRIVATE
+                -fno-common  
+        )
+        SET_TARGET_PROPERTIES(mariadb-frm PROPERTIES
+                LINK_FLAGS "-Wl,-w -Wl,-no_pie -Wl,-dead_strip -mmacosx-version-min=14.0"
```
gkodinov, line 79: Are you absolutely positive you need all of these options? E.g. -mmacosx-version-min has a cmake equivalent: CMAKE_OSX_DEPLOYMENT_TARGET. 
I'd strongly suggest to either remove all of these or reuse it from some other places.

extra/mariadbfrm/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+        )
+    ELSE()
+        TARGET_LINK_OPTIONS(mariadb-frm PRIVATE -flto
+                "LINKER:--allow-multiple-definition"
```
gkodinov, line 83: LINKER:--allow-multiple-definition seems like silencing the messenger to me. You most probably are specifying libraries twice or source files twice. 
I'd rather try to see what are the duplicates and attack that.

extra/mariadbfrm/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+
+    ADD_DEFINITIONS(-DFRM_PARSER)
+
+    INCLUDE_DIRECTORIES(
+        ${CMAKE_SOURCE_DIR}/include
+        ${CMAKE_SOURCE_DIR}/sql
+        ${CMAKE_BINARY_DIR}/sql
+        ${CMAKE_BINARY_DIR}/include
+        ${LIBFMT_INCLUDE_DIR}
+        ${PCRE_INCLUDE_DIRS}
```
gkodinov, line 11: are you using PCRE? Why do you need to include this?

extra/mariadbfrm/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
+IF(NOT WITHOUT_SERVER)
+
+    ADD_DEFINITIONS(-DFRM_PARSER)
+
+    INCLUDE_DIRECTORIES(
+        ${CMAKE_SOURCE_DIR}/include
+        ${CMAKE_SOURCE_DIR}/sql
+        ${CMAKE_BINARY_DIR}/sql
```
gkodinov, line 8: why do you need to include /sql and /include twice, but in different order?

extra/mariadbfrm/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+        ${CMAKE_BINARY_DIR}/include
+        ${LIBFMT_INCLUDE_DIR}
```
gkodinov, line 10: are you using libfmt in your utility? I sincerely doubt that.

extra/mariadbfrm/mariadb_frm.cc
```diff
@@ -0,0 +1,658 @@
...
+  {                                                                           \
+    if (debug_enabled)                                                        \
+    {                                                                         \
+      fprintf(stderr, "DEBUG: " fmt, ##__VA_ARGS__);                          \
+      fflush(stderr);                                                         \
+    }                                                                         \
+  } while (0)
+
+extern mysql_mutex_t LOCK_thread_id;
+extern handlerton *get_frm_mock_handlerton();
```
gkodinov, line 52: this is very bad style! please put these in common header(s) so that they're actually checked against their definitions.

extra/mariadbfrm/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
+IF(NOT WITHOUT_SERVER)
```
gkodinov, line 1: you will need a copyright header I'd guess.

extra/mariadbfrm/frm_mocks.cc
```diff
@@ -0,0 +1,253 @@
+#include "handler.h"
```
gkodinov, line 1: this one will also need a copyright header.

sql/handler.cc
```diff
@@ -264,6 +271,8 @@ handlerton *ha_default_tmp_handlerton(THD *thd)
 plugin_ref ha_resolve_by_name(THD *thd, const LEX_CSTRING *name,
                               bool tmp_table)
 {
+  if (ha_resolve_by_name_hook)
```
gkodinov, line 274: This feels like a very bad idea. As it will be done for the actual server in an actual open context every time you open a table.

gkodinov (approved):
Thank you for working on this. It's a very interesting idea. 
I'm somewhat terrified by the approach taken though: starting a gutted server with hooks that sway otherwise deeply layered code towards doing different things. But this is not for me to decide on. 

I'm guessing that, if approved, this needs a lot of cleanup. I've highlighted some of it below. 

But I'd strongly suggest considering getting an approval for the architecture first before we dive any deeper. 

Formal preliminary review criteria met, hence approving it.

PR #4096, opened 2025-06-09
GSOC'25 - MDEV-36100 Automatic embedding generation and "expensive" functions 

apostolis1:
This DRAFT PR is part of the GSOC'25 project "[MDEV-36100](https://jira.mariadb.org/browse/MDEV-36100) Generate vector embeddings automatically on INSERT".
This is a DRAFT PR, meant to be used for mentor feedback and will be updated as the project progresses.

There are two main aspects on this PR:
1. Plugin to automatically generate vector emebeddings
2. Avoidance of recalculation of stored expensive virtual columns

- [x] *The Jira issue number for this PR is: MDEV-36100*

## Description

TODO: This will be updated 

## Release Notes

The plugin uses the following session variables:
1. `generate_embedding_openai_host`: The API endpoint the request will be sent to
2. `generate_embedding_openai_api_key`: The API key used for the request, provided via [HTTP Bearer authentication](https://platform.openai.com/docs/api-reference/authentication) 

and provides the following status variables to help system admins keep track of requests:
1. `generate_embedding_openai_successful_http_requests`
2. `generate_embedding_openai_total_http_requests`

## How can this PR be tested?

This PR creates a new testing suite, `gen_embedding`,which includes tests both for the plugin and the expensive functionality.
Expensive tests have the prefix `expensive_`, to distinguish them from the plugin tests.

## Future Work

In terms of the plugin, the status variables are not being reset by `FLUSH STATUS`, it remains to be investigated why that is and whether there is special support needed from the plugin.

As far as the stored expensive virtual columns are concerned:

- Currently, changes in the virtual columns data type in `ALTER` statements are not detected.
So, something like the following incorrectly does not recompute the virtual column:
```sql
create table t (
    doc text,
    embedding vector(1536) as (generate_embedding_openai(doc, "text-embedding-ada-002")) stored
) engine=MyISAM;
alter table t modify embedding vector(2000);
```

- We would also like to avoid recomputing stored expensive virtual columns in replication setups whenever possible.
Initially, we would want to do this for [Row-Based Logging](https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/binary-log-formats#row-based-logging) and then for [Mixed Logging](https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/binary-log-formats#mixed-logging).
## PR quality check


plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/*************************************************************************/
+
+maria_declare_plugin(type_test)
+{
+  MariaDB_FUNCTION_PLUGIN,       // the plugin type
+  Item_func_gen_embedding::plugin_descriptor(),
+  "GENERATE_EMBEDDING_OPENAI",// plugin name
+  "MariaDB Corporation",        // plugin author
```
vuvova: you might want to put yourself here

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,356 @@
...
+  String *read_json(String *str, json_value_types *type, char **out_val);
+  json_path_with_flags *paths;
+  
+  public:
+    using Item_str_func::Item_str_func;
+    bool fix_length_and_dec(THD *thd) override
+    {
+      // TODO: I am grabbing the session variables here, because this is the function that has access to the THD
+      // Is there a better place to do this?
+      host = THDVAR(thd, host);
```
apostolis1: Regarding the plugin variables and access to the thread descriptor, I just came across `_current_thd`
This allows me to create a getter, for example
`const char* get_openai_host(THD* thd)
{
  return THDVAR(thd, host);
}`
which can then be used as `const char* myHost = get_openai_host(_current_thd());` where appropriate. 
I will use this pattern, which seems to be the correct way to handle plugin variables

vuvova: the common pattern is `current_thd` because of
```c
#define current_thd _current_thd()
```
in one of the headers

FooBarrior: It's actually fine to do that in `fix_length_and_dec`. If you want, you can override fix_fields instead, as a more dedicated pick, just don't forget to call the "super" definition.

Also, I believe `host` should be a global var.

apostolis1: For transparency: This has been discussed internally, and we agreed that it makes sense for `host` to be a session var

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+#include <sql_class.h>
+#include <mysql/plugin_function.h>
+#include <curl/curl.h>
+#include "sql_type_vector.h"
+#include "item_jsonfunc.cc"
+#include <unordered_map>
+#include "json_lib.h"
+
+static char *host, *api_key;
+ulonglong curl_requests= 0; /* Number of total curl requests */
```
apostolis1, line 29: Adding this comment from another commit here as well for transparency
> HTTP requests, it is called. curl is not a protocol, it is very much possible to implement the same without curl

This makes sense, will probably replace `curl` with `http`

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    uint32 actual_len = copy_and_convert(converted_input_buff, dst_length, &cs_openai, input->ptr(), input->length(),
+      cs, &errors);
+    if (errors != 0) {
+      my_printf_error(1, "GENERATE_EMBEDDING_OPENAI: "
+        "Error converting input string from %s to UTF-8 charset", ME_ERROR_LOG | ME_WARNING, cs->cs_name.str);
+      null_value = true;
+      return 1;
+    }
+    converted_input_buff[actual_len] = '\0';
+    String *converted_input_str = new String(converted_input_buff, actual_len, &cs_openai);
```
apostolis1: This can be changed to `String converted_input_str(converted_input_buff, actual_len, &cs_openai);` if we don't want the `new` allocation

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+class Item_func_gen_embedding: public Item_str_func 
+{
+  String tmp_js, api_response, tmp_str;
+  std::unordered_map<std::string, int> model_dimensions = {
+    {"text-embedding-3-small", 1536},
+    {"text-embedding-ada-002", 1536},
+    {"text-embedding-3-large", 3072}
+  };
+
+  int make_openai_request() {
```
apostolis1: The `make_openai_request` method is a rather big method, but it does what it promises. It could be split into smaller methods if needed.

plugin/gen_embedding/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2019, MariaDB corporation. All rights reserved.
```
vuvova: Some day before the PR is merged, you'll need to update copyright lines to say 2025

plugin/gen_embedding/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1335  USA
+FIND_PACKAGE(CURL)
+IF(NOT CURL_FOUND)
+  # Can't build plugin
+  MESSAGE_ONCE(WARNING "Generate OpenAI Embeddings plugin requires curl development package")
+  RETURN()
+ENDIF()
+
+SET_PACKAGE_PROPERTIES(CURL PROPERTIES TYPE REQUIRED)
```
vuvova: I think it's confusing, `cmake` will say
```
-- The following REQUIRED packages have been found:
 
 * Curses
 * Threads
 * CURL
```
which isn't exactly correct, because CURL isn't required, the server will build fine without it.
I know you've copied it from hashicorp plugin, but it's wrong there too.

Remove this line.

plugin/gen_embedding/mysql-test/gen_embedding/minimal_socket_server.py
```diff
@@ -0,0 +1,88 @@
...
+
+def handle_request(request):
+    try:
+        request_line = request.splitlines()[0]
+        method, path, _ = request_line.split()
+        if path == "/success" and method in ("GET", "POST"):
+            parsed_input = json.loads(request.splitlines()[-1])["input"]
+            print(parsed_input)
+            with open(SUCCESS_RESPONSES_FILENAME, 'r') as file:
+                responses = json.load(file)
```
vuvova: Why not to put these two lines on the global scope?

plugin/gen_embedding/mysql-test/gen_embedding/minimal_socket_server.py
```diff
@@ -0,0 +1,88 @@
...
+                print(f"Input '{parsed_input}' found in {SUCCESS_RESPONSES_FILENAME}.")
+                return_text = responses[parsed_input]
+                return build_response(json.dumps(return_text), 200)
+            else:
+                print(f"Input '{parsed_input}' not found in {SUCCESS_RESPONSES_FILENAME}.")
+                # In our tests we always provide a valid input, but we can return some form of error here
+                return build_response(json.dumps({"error": "Not Found"}), 400)
+        elif path == "/errorcode" and method in ("GET", "POST"):
+            # The response body is irreleveant for this endpoint, the tests only care about the status code
+            return build_response(json.dumps({"error": "Not Found"}), 400)
```
vuvova: okay, although now, when you changed responces into a dict, your /success endpoint can return "Not found",
so /errorcode is a bit redundant

plugin/gen_embedding/mysql-test/gen_embedding/suite.pm
```diff
@@ -0,0 +1,62 @@
...
+     error        => $http_server_log,
+     append       => 1,
+     nocore       => 1,
+    );
+}
+
+sub openai_wait {
+    my ($opneai, $test) = @_; # My::Config::Group, My::Test
+    my $cmd= "echo Waiting for openai mockup server at port $ENV{OPENAI_PORT}"; 
+    system $cmd;
```
vuvova: eventually we'll want here a proper waiting logic. It doesn't have to be complex, for example
```perl
open(LOG, '<', $http_server_log);
do {
  sleep 1 and seek LOG,0,1 until $_=<LOG>;
} until /known string/;
```
this waits for a known string  to appear in the log and is, of course, completely untested.

plugin/gen_embedding/plugin.cc
vuvova: users don't care how you call your internal variables, you can keep calling it curl if you think it describes it better. Just rename the user-visible name in the `status_variables[]` array

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+  size_t realsize = size * nmemb;
+  std::ostringstream *read_data = static_cast<std::ostringstream *>(userp);
+  read_data->write(static_cast<char *>(contents), realsize);
+  return realsize;
+}
+
+class Item_func_gen_embedding: public Item_str_func 
+{
+  String tmp_js, api_response, tmp_str;
+  std::unordered_map<std::string, int> model_dimensions = {
```
vuvova: could be global and static, I suppose.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+      null_value= true;
+      return 1;
+    }
+
+    std::string model_name= (std::string) args[1]->val_str()->c_ptr();
+    // Before performing the request, check if the model is supported
+    // This might allow us to avoid a request if the model is not supported
+    if (model_dimensions.find(model_name) == model_dimensions.end()) {
+      my_printf_error(1, "GENERATE_EMBEDDING_OPENAI: "
+        "Model %s is not supported", ME_ERROR_LOG | ME_WARNING, model_name.c_str());
```
vuvova: generally user errors aren't printed into the error log

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+      null_value= true;
+      return 1;
+    }
+
+    // Convert the input to cs_openai charset
+    uint errors= 0;
+    CHARSET_INFO *cs= input->charset();
+    CHARSET_INFO cs_openai = my_charset_utf8mb4_general_ci;
```
vuvova: talked about that already, don't copy a charset, use a pointer to it:
```c
CHARSET_INFO *cs_openai = &my_charset_utf8mb4_general_ci;
```

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    // Convert the input to cs_openai charset
+    uint errors= 0;
+    CHARSET_INFO *cs= input->charset();
+    CHARSET_INFO cs_openai = my_charset_utf8mb4_general_ci;  
+    size_t dst_length= cs_openai.mbmaxlen * input->length() + 1;
+    MEM_ROOT *root= current_thd->active_stmt_arena_to_use()->mem_root;
+    char *converted_input_buff;
+    converted_input_buff= (char *) alloc_root(root, dst_length);
```
vuvova: This is wrong on a couple of level.
memroot is an allocator that is always freed as a whole, individual entries cannot be freed. This conveniently helps to avoid memory leaks, you allocate and forget.

But you cannot keep allocating in a memroot in a loop. thd->mem_root is usually freed at the end of the statement execution. So, in
```sql
select generate_embedding_openai(txt, @model) from t1;
```
you'll be allocating memory for every row. From the same memroot.

A second problem — you use "statement arena", not "execution arena". There is no difference in your tests, but in prepared statements:
```sql
prepare stmt1 from 'select generate_embedding_openai(txt, ?) from t1';
execute stmt1 using 'text-embedding-3-small';
execute stmt1 using 'text-embedding-3-small';
```
the execution arena is freed after every execution, while statement arena has the same lifetime as stmt1.

Both are completely unnecessary, you only need memory for the duration of your `make_openai_request()` call. A simplest approach would be to use good old malloc/free. A better solution would be to have a
```c++
String converted_buffer;
```
member in your `Item_func_gen_embedding` class. This way you'll keep the buffer when the function is invoked for every next row, avoiding per-row malloc calls.

plugin/gen_embedding/plugin.cc
vuvova: Don't want indeed. But you can also avoid allocations by making it a member of `Item_func_gen_embedding`

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    std::ostringstream read_data_stream;
+    std::string response, post_fields, escaped_input_str;
+    struct curl_slist *slist1;
+    long http_response_code;
+    std::string authorization = std::string("Authorization: Bearer ") + api_key;
+    
+    // Escape the input string
+    const int jsLen = converted_input_str->length();
+    const char* rawJS = converted_input_str->ptr();
+    int strLen = jsLen * 12 * cs_openai.mbmaxlen / cs_openai.mbminlen;
```
vuvova: add a comment, why 12

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    char* buf = (char*)alloca(strLen);
```
vuvova: an argument might be big, `alloca` is not a good idea here. Either use `my_safe_alloca` or maloc/free or, like above, a member of `Item_func_gen_embedding`

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    if ((strLen = json_escape(&cs_openai, (const uchar*)rawJS, (const uchar*)rawJS + jsLen, &cs_openai, (uchar*)buf,
```
vuvova: not sure, but it seems you don't need to convert the string from its original charset to utf8mb4, because `json_escape` can take an argument in any charset, it looks like it does the conversion internally

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    escaped_input_str = std::string(buf);
+    // It is essential to escape the input string to ensure the input of the API call is a valid JSON
+    // The model name does not need escaping as it is a valid string
+    // Validity of the model name is checked later in this function 
+    post_fields =
+      std::string("{\"input\": \"") +
+      escaped_input_str + 
+      "\", \"model\": \"" +
+      model->c_ptr() +
+      "\",\"encoding_format\": \"float\"}";
```
vuvova: you don't really need to convert `char*` to `std::string` if all you want is to concatenate it with something. You can do, for example
```c
strxnmov(post_fields, post_fields_size, "{\"input\": \"", buf, "\", \"model\": \"", model->c_ptr(), "\",\"encoding_format\": \"float\"}", NULL);
```
or you can use `my_snprintf`

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    curl_easy_setopt(hnd, CURLOPT_POSTFIELDS, post_fields.c_str());
+    curl_easy_setopt(hnd, CURLOPT_POSTFIELDSIZE_LARGE, (curl_off_t)post_fields.length());
+    curl_easy_setopt(hnd, CURLOPT_HTTPHEADER, slist1);
+    curl_easy_setopt(hnd, CURLOPT_USERAGENT, "curl/8.5.0");
+    curl_easy_setopt(hnd, CURLOPT_MAXREDIRS, 50L);
+    curl_easy_setopt(hnd, CURLOPT_HTTP_VERSION, (long)CURL_HTTP_VERSION_2TLS);
+    curl_easy_setopt(hnd, CURLOPT_FTP_SKIP_PASV_IP, 1L);
+    curl_easy_setopt(hnd, CURLOPT_TCP_KEEPALIVE, 1L);
+    curl_easy_setopt(hnd, CURLOPT_WRITEFUNCTION, WriteMemoryCallback);
+    curl_easy_setopt(hnd, CURLOPT_WRITEDATA, &read_data_stream);
```
vuvova, line 180: would it make sense to keep `slist` and `hnd` between `val_str()` calls to not construct them for every row?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    {
+      host = THDVAR(thd, host);
+      api_key = THDVAR(thd, api_key);
+      uint max_dimensions = 3072; // Default to the largest embedding size
+      if (args[1]->const_item()) { // If a const, we can parse the model name here and determine the max dimensions 
+        std::string model_name= (std::string) args[1]->val_str()->c_ptr();
+        if (model_dimensions.find(args[1]->val_str()->c_ptr()) != model_dimensions.end()) {
+          max_dimensions = model_dimensions[model_name];
+        }
+        else {
```
vuvova: if you use mariadb indentation style, use it consistently everywhere, please

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+          // TODO: Can we also early exit here?
+          my_printf_error(1, "GENERATE_EMBEDDING_OPENAI: "
+            "Model %s is not supported", ME_ERROR_LOG | ME_WARNING, model_name.c_str());
+        }
+      }
+
+      decimals= 0;
+      MEM_ROOT *root= thd->active_stmt_arena_to_use()->mem_root;
+      uint n_paths = 1;
```
vuvova: not sure, why you need `n_paths` here

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+                                          char **out_val)
+{
+  String *js= &api_response;
+  json_engine_t je;
+  json_path_t p;
+  const uchar *value;
+  int not_first_value= 0, count_path= 0;
+  size_t v_len;
+  int array_size_counter[JSON_DEPTH_LIMIT];
+  json_path_with_flags *c_path= paths;
```
vuvova: why?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+          // TODO: Can we also early exit here?
+          my_printf_error(1, "GENERATE_EMBEDDING_OPENAI: "
+            "Model %s is not supported", ME_ERROR_LOG | ME_WARNING, model_name.c_str());
+        }
+      }
+
+      decimals= 0;
+      MEM_ROOT *root= thd->active_stmt_arena_to_use()->mem_root;
+      uint n_paths = 1;
+      paths= (json_path_with_flags *) alloc_root(root, sizeof(json_path_with_flags) * n_paths);
```
vuvova: ok, this is allocated once per statement, not per row. so ok to do on a memroot. but still, not on a statement arena, on execution arena.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+  int not_first_value= 0, count_path= 0;
+  size_t v_len;
+  int array_size_counter[JSON_DEPTH_LIMIT];
+  json_path_with_flags *c_path= paths;
+
+  if ((null_value= args[0]->null_value))
+    return 0;
+
+  c_path->p.types_used= JSON_PATH_KEY_NULL;
+  const uchar* s_p = (const uchar*) JSON_EMBEDDING_PATH.c_str();
```
vuvova: why do you create a static std::string, if all you ever need is char* ?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+  if (s_p)
```
vuvova: how can it be NULL?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+                      s_p + strlen((const char *) s_p)))
+    {
+    //  report_path_error(s_p, &c_path->p, n_arg);
+      goto return_null;
+    }
+  }
+
+  *type= JSON_VALUE_NULL;
+
+  if (str)
```
vuvova: this is redundant, `str` — as an argument of `val_str()` — can not be NULL

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+
+    if (*type == JSON_VALUE_NULL)
+    {
+      *type= je.value_type;
+      *out_val= (char *) je.value;
+      // value_len= je.value_len;
+    }
+    if (!str)
+    {
+      /* If str is NULL, we only care about the first found value. */
```
vuvova: I don't understand that

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+  return 0;
+}
+
+int Item_func_gen_embedding::parse_vector(String *buf, json_engine_t je, CHARSET_INFO *cs, const uchar *start, const uchar *end)
+{
+  /* 
+    This function is used to parse the vector from the JSON response
+    It should be called after the embedding is located, for the start and end positions 
+    of the json_engine to be known
+    The vector is expected to be in binary format
```
vuvova: this was a bit confusing, took me a while to understand.
the vector that you parse is not in binary format, it's json array of numbers. the result of the parsing — that vector is in binary format.
so, better say "the output vector" or "the result vector" or something along those lines

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,455 @@
...
+    //  report_path_error(s_p, &c_path->p, n_arg);
+      goto return_null;
+    }
+  }
+
+  *type= JSON_VALUE_NULL;
+
+  if (str)
+  {
+    str->set_charset(js->charset());
```
vuvova: not really? `str` is the resulting vector, not json. it should be always `my_charset_bin`, as far as I understand.

plugin/gen_embedding/plugin.cc
apostolis1: Yes, I am referring to the status variable name - should have put the comment near the  `status_variables[] array`

plugin/gen_embedding/plugin.cc
apostolis1: Should we just exit then, or should the error be logged somewhere else?

plugin/gen_embedding/plugin.cc
apostolis1: It can't, this is probably leftover code

plugin/gen_embedding/plugin.cc
apostolis1: Doesn't make sense indeed, probably leftover code from another version which modified `str`

plugin/gen_embedding/plugin.cc
apostolis1: This is from the json parsing function for the general case, where we could be looking for multiple paths. I agree though that we don't need it here, the intermediate variable `c_path` will be removed - and I will rename `paths` to `json_path` since it is a single path.

plugin/gen_embedding/plugin.cc
vuvova: I mean, just remove `ME_ERROR_LOG`, the warning will be sent to the user, but won't be written into the system-wide error log

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,452 @@
...
+  int make_openai_request()
+  {
+    /* 
+      Make the request to OpenAI API to generate the embedding
+      Store the result in api_response
+    */
+    CHARSET_INFO *cs_openai= &my_charset_utf8mb4_general_ci;
+    CURLcode ret;
+    CURL *hnd=NULL;
+    std::ostringstream read_data_stream;
```
FooBarrior, line 90: add `#include <sstream>`

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,452 @@
...
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA */
+
+#define MYSQL_SERVER
+
+#include <my_global.h>
+#include <sql_class.h>
+#include <mysql/plugin_function.h>
+#include <curl/curl.h>
+#include "sql_type_vector.h"
+#include "item_jsonfunc.cc"
```
FooBarrior: include .h, otherwise you'll end up with various problems on some builds, like duplicate symbol definitions

sql/item.cc
```diff
@@ -965,6 +965,27 @@ bool Item_field::register_field_in_write_map(void *arg)
   return 0;
 }
 
+/*
+  Processor used in UPDATE statements for expensive virtual stored columns
+  Walks down the expression tree
+  Returns true if any of the fields in the expression tree are in the write set
+*/
+
+bool Item_field::check_field_in_write_map(void *arg)
```
FooBarrior: Let's refactor it into `check_field_in_map`, where `arg` will be `pair<TABLE*, my_bitmap*>`

usage: 
```
auto arg= std::make_pair({table, table->write_set});
res= field->vcol_info->expr->walk(&Item::check_dependencies_in_write_set, 1, &arg);
```

sql/item.cc
```diff
@@ -965,6 +965,27 @@ bool Item_field::register_field_in_write_map(void *arg)
...
+{
+  TABLE *table= (TABLE *) arg;
+  bool res = false;
+  if (field->table == table || !table)
```
FooBarrior: `vcol`s can't refer to other tables. Is it really needed (well, except for being more generic)?

apostolis1: It is not needed, but since we have also refactored this to `check_field_in_map`, I guess it might be used in the future in more generic use cases than vcols, so I believe we can keep it like this

plugin/gen_embedding/plugin.cc
apostolis1: I agree.
The problem is I need functions like `path_exact` which are declared and defined in `item_jsonfunc.cc`, I suspect because initially there was no need for them to be used outside of `item_jsonfunc.cc`

sql/item.cc
FooBarrior: I think it's the opposite: a bitmap can only refer to a single table. So my suggestion is actually narrowing this walker's possibility.

plugin/gen_embedding/plugin.cc
FooBarrior: Of course, the developer couldn't predict a usage outside their scope

FooBarrior: So the valid approach will be declaring them

sql/item.cc
apostolis1: By "suggestion", do you mean removing the check `if (field->table == table || !table)`?

FooBarrior: no, I mean my suggestion to refactor this as `check_field_in_map`

apostolis1: Ah I see.
But still, I think refactoring to `check_field_in_map` would allow checking if the field is in other bitmaps potentially.

FooBarrior: Here's another point: that line can't be tested by any test case.

I'll add another point on top of it: why not to check the `field->table->write_set` overall? Do you know, would the caller want checking only in some given table's write_set, or in any table's write_set?

I'll explain. We have a multi-table update option:
UPDATE t1, t2 set t2.x = t1.y, t1.x = t2.y;

I guess in such scenarios the caller might want checking if the field is in it own table's write_set.
Though AFAIK multi-update creates a new fake table where the update happens, and then splits the results into real tables (@vuvova fix me if I'm wrong), My refactoring suggestion is based on that (thoug I can be wrong indeed!).

I'd suggest converting this check into DBUG_ASSERT(...). This way it will be useful, won't provoke extra questions, and besides we'll find out if I'm right about multi-update! :grin:

sql/sql_table.cc
```diff
@@ -12698,6 +12708,54 @@ copy_data_between_tables(THD *thd, TABLE *from, TABLE *to,
...
+      {
+        /*
+          Remove the vcol from the write_set, as check_dependencies_in_write_set checks if the bit is set
+          The vcol will be added back to the write_set later if it needs to be recomputed 
+        */
+        bitmap_clear_bit(to->write_set, (*ptr)->field_index);
+        /* 
+          If there is no change in the dependencies and the vcol expression and type (see TODO below), we can copy directly
+        */
+        if (!to->check_dependencies_in_write_set(*ptr) && def->field->vcol_info && (*ptr)->vcol_info->is_equal(def->field->vcol_info, true))
```
apostolis1, line 12741: This check does not detect changes in the column's data type, for example changes like
`alter table t modify embedding vector(2000);`
What would be a good way to detect those?

sql/item.cc
apostolis1: I agree with changing this to DBUG_ASSERT(...), it has been implemented

plugin/gen_embedding/mysql-test/gen_embedding/minimal_socket_server.py
apostolis1: Yes that's correct. The idea behind it is that since we want to move away from a python server in the future, we might want as little logic as possible encoded in the server itself.
However, this could be changed to check if the text provided is of a special value (e.g. an API request for the document `"error"`), and then return the error code in that case.

apostolis1:
For anyone interested, I have created [this gist](https://gist.github.com/apostolis1/8c17712a256f86969efb7d8e4cb4c591) as well with some additional information on this work.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+      {
+        // TODO: Can we also early exit here?
+        my_printf_error(1, "GENERATE_EMBEDDING_OPENAI: "
+          "Model %s is not supported", ME_ERROR_LOG | ME_WARNING, model_name.c_str());
+      }
+    }
+
+    decimals= 0;
+    MEM_ROOT *root= thd->mem_root;
+    json_path= (json_path_with_flags *) alloc_root(root, sizeof(json_path_with_flags));
```
svoj, line 248: Either use `thd->alloc<json_path_with_flags>(1)` or `thd_alloc(thd, sizeof(json_path_with_flags))`.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+int Item_func_gen_embedding::parse_vector(String *buf, json_engine_t je, CHARSET_INFO *cs, const uchar *start, const uchar *end)
+{
+  /* 
+    This function is used to parse the vector from the JSON response
+    It should be called after the embedding is located, for the start and end positions 
+    of the json_engine to be known
+    The vector being parsed is a JSON array of numbers
+    The output vector is in binary format, will be stored in buf by this function
+    Returns 0 on success, 1 on error
+  */
```
svoj, line 393: Function comment should be moved before function itself.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    {"text-embedding-ada-002", 1536},
+    {"text-embedding-3-large", 3072}
+  };
+
+  int make_openai_request()
+  {
+    /* 
+      Make the request to OpenAI API to generate the embedding
+      Store the result in api_response
+    */
```
svoj, line 86: Function comment should be moved before function itself.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    slist1= NULL;
+    response= read_data_stream.str();
+
+    api_response.copy(response.c_str(), response.length(), &my_charset_utf8mb4_general_ci);
+    return 0;
+cleanup:
+    curl_easy_cleanup(hnd);
+    hnd= NULL;
+    curl_slist_free_all(slist1);
+    slist1= NULL;
```
svoj, line 213: Resetting `hnd` and `slist1` is redundant. `cleanup` label duplicates preceding code, this can be avoided.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+  const uchar *value;
+  int not_first_value= 0, count_path= 0;
+  size_t v_len;
+  int array_size_counter[JSON_DEPTH_LIMIT];
+  
+  if ((null_value= args[0]->null_value))
+    return 0;
+
+  json_path->p.types_used= JSON_PATH_KEY_NULL;
+  const uchar *s_p= (const uchar *) JSON_EMBEDDING_PATH;
```
svoj, line 321: Redundant variable.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+  size_t v_len;
+  int array_size_counter[JSON_DEPTH_LIMIT];
+  
+  if ((null_value= args[0]->null_value))
+    return 0;
+
+  json_path->p.types_used= JSON_PATH_KEY_NULL;
+  const uchar *s_p= (const uchar *) JSON_EMBEDDING_PATH; 
+  if (json_path_setup(&json_path->p,&my_charset_utf8mb4_general_ci, s_p,
+                    s_p + strlen((const char *) s_p)))
```
svoj, line 323: `strlen()` can be avoided.

plugin/gen_embedding/mysql-test/gen_embedding/t/bad_response.test
```diff
@@ -0,0 +1,20 @@
...
+
+# This file contains tests for unexpected/bad API responses
+
+call mtr.add_suppression("GENERATE_EMBEDDING_OPENAI: Bad http response code: 400");
+
+# Currently only checks that an HTTP error code is handled correctly,
+# without examining the response body
+
+--replace_result $OPENAI_PORT OPENAI_PORT
+eval set generate_embedding_openai_host="http://127.0.0.1:$OPENAI_PORT/success";
```
svoj, line 11: Replace with `evalp` all similar occurrences.

plugin/gen_embedding/mysql-test/gen_embedding/include/have_gen_embedding.opt
```diff
@@ -0,0 +1 @@
+--plugin-load-add=$GEN_EMBEDDING_SO
```
svoj, line 1: Should be moved to `suite.opt`?

plugin/gen_embedding/mysql-test/gen_embedding/include/have_gen_embedding.inc
```diff
@@ -0,0 +1,9 @@
+if (`SELECT COUNT(*)=0 FROM INFORMATION_SCHEMA.PLUGINS WHERE PLUGIN_NAME = 'GENERATE_EMBEDDING_OPENAI' AND PLUGIN_STATUS='ACTIVE'`)
+{
+  --skip Test requires active GENERATE_EMBEDDING_OPENAI plugin
+}
+
+# Helper function
+let $get_expensive_calls= (select variable_value 
+   from information_schema.session_status 
+   where variable_name like 'Generate_embedding_openai_successful_http_requests');
```
svoj, line 9: We don't normally put helper functions into `have_*` files.
`like` is redundant here, use `=`.

plugin/gen_embedding/mysql-test/gen_embedding/minimal_socket_server.py
```diff
@@ -0,0 +1,83 @@
...
+        print(f"Error binding to port {PORT}: {e}", flush=True)
+        sys.exit(0)
+    server_socket.listen(1)
+
+    print(f"Started mockup API server on PORT:{PORT} ...", flush=True)
+
+    while True:
+        client_connection, client_address = server_socket.accept()
+        with client_connection:
+            request = client_connection.recv(4096).decode('utf-8')
```
svoj, line 78: IIUC, this code does not handle HTTP requests that arrive in multiple chunks, which may lead to sporadic test failures.

plugin/gen_embedding/mysql-test/gen_embedding/minimal_socket_server.py
```diff
@@ -0,0 +1,83 @@
...
+
+def get_response(filename, status_code):
+    if not os.path.exists(filename):
+        print(f"File {filename} not found.", flush=True)
+        return build_response(json.dumps({"error": "File not found"}), 404)
+
+    with open(filename, 'r') as file:
+        content = file.read()
+
+    return build_response(content, status_code, content_type="application/json")
```
svoj, line 23: `get_response` looks like unused function?

plugin/gen_embedding/mysql-test/gen_embedding/minimal_socket_server.py
```diff
@@ -0,0 +1,83 @@
...
+
+    except Exception as e:
+        return build_response(json.dumps({"error": str(e)}), 500)
+
+def build_response(body, status_code, content_type="application/json"):
+    reason_phrases = {
+        200: "OK",
+        400: "Error",
+        404: "Not Found"
+    }
```
svoj, line 51: 500 is missing here

plugin/gen_embedding/mysql-test/gen_embedding/suite.pm
```diff
@@ -0,0 +1,64 @@
...
+
+@ISA = qw(My::Suite);
+
+sub skip_combinations {
+  my %skip;
+  my $api_key_file = "$ENV{HOME}/openai_api_key.inc";
+  $skip{'t/openai_server.test'}="No OpenAI API key" unless -r $api_key_file;
+  $skip{'t/openai_greek_charset.test'}="No OpenAI API key" unless -r $api_key_file;
+  %skip;
+}
```
svoj, line 15: I'd rather make `include/openai_api_key.inc` that'd do `skip` and read some file that contains the key and then do appropriate step to set it up.

`$ENV{HOME}/openai_api_key.inc` contents is a mystery for us, we can only guess which commands it is supposed to execute.

plugin/gen_embedding/mysql-test/gen_embedding/my.cnf
```diff
@@ -0,0 +1,8 @@
+!include include/default_my.cnf
+
+[openai]
+listen = @ENV.OPENAI_PORT
+api_response_filename = @ENV.MTR_SUITE_DIR/json_responses/api_responses.json
```
svoj, line 5: `api_response_filename` looks redundant. I'm unsure if `listen` is required too.

plugin/gen_embedding/mysql-test/gen_embedding/t/greek_mockup.test
```diff
@@ -0,0 +1,18 @@
+source include/have_gen_embedding.inc;
+--disable_query_log
+source $HOME/openai_api_key.inc;
+--enable_query_log
```
svoj, line 4: This test is failing without `openai_api_key.inc`.

plugin/gen_embedding/mysql-test/gen_embedding/suite.pm
```diff
@@ -0,0 +1,64 @@
...
+    my $python_path="python3"; # TODO: Maybe use a config option, could also be "python" or a specific path
+    my $success_file_path=$openai->value('api_response_filename');
+    my $python_server_path="$ENV{MTR_SUITE_DIR}/minimal_socket_server.py";
+    my $args;
+    &::mtr_init_args(\$args);
+    &::mtr_add_arg($args, $python_server_path);
+    &::mtr_add_arg($args, "$ENV{OPENAI_PORT}");
+    &::mtr_add_arg($args, $success_file_path);
+    # This can be changed to include the PORT or other info, perhaps in combination with append => 0
+    $http_server_log= "openai.log";
```
svoj, line 32: It create `mysql-test/openai.log` in the source tree, whereas this file has to be created somewhere in `MYSQLTEST_VARDIR`.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+#include "json_lib.h"
+
+ulonglong curl_requests= 0; /* Number of total curl requests */
+ulonglong successful_curl_requests= 0; /* Number of successful curl requests */
+static const char *JSON_EMBEDDING_PATH = "$.data[0].embedding";
+
+static MYSQL_THDVAR_STR(host,
+  PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_MEMALLOC,
+  "OpenAI API host, can be set to 'https://api.openai.com/v1/embeddings' or a custom endpoint",
+  NULL, NULL, "");
```
svoj, line 36: Given that this plugin is implemented specifically for OpenAI, shouldn't it have sane default `host` value?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    post_fields.append(STRING_WITH_LEN("\", \"model\": \""));
+    post_fields.append(model->c_ptr(), model->length());
+    post_fields.append(STRING_WITH_LEN("\",\"encoding_format\": \"float\"}"));
+    
+    // Prepare the request to OpenAI API
+    slist1= NULL;
+    slist1= curl_slist_append(slist1, authorization.c_str());
+    slist1= curl_slist_append(slist1, "Content-Type: application/json; charset=utf-8");
+    hnd= curl_easy_init();
+    curl_easy_setopt(hnd, CURLOPT_BUFFERSIZE, 102400L);
```
svoj, line 168: `curl_easy_setopt()` return value should be checked too, e.g. see hashicorp plugin.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    post_fields.length(post_fields.length()+strLen); // We need to manually update the length
+    post_fields.append(STRING_WITH_LEN("\", \"model\": \""));
+    post_fields.append(model->c_ptr(), model->length());
+    post_fields.append(STRING_WITH_LEN("\",\"encoding_format\": \"float\"}"));
+    
+    // Prepare the request to OpenAI API
+    slist1= NULL;
+    slist1= curl_slist_append(slist1, authorization.c_str());
+    slist1= curl_slist_append(slist1, "Content-Type: application/json; charset=utf-8");
+    hnd= curl_easy_init();
```
svoj, line 167: `curl_easy_init()` may perform lazy global initialization by calling `curl_global_init()`, which is not thread safe. That is if multiple threads call `generate_embedding_openai()` for the first time it may cause server crash.

Another problem is that multiple plugins use of curl. Even if we move `curl_global_init()` out to plugin init routines, in theory they may call `curl_global_init()` concurrently.

plugin/gen_embedding/mysql-test/gen_embedding/suite.pm
```diff
@@ -0,0 +1,64 @@
...
+  $skip{'t/openai_greek_charset.test'}="No OpenAI API key" unless -r $api_key_file;
+  %skip;
+}
+
+sub openai_start {
+    my ($openai, $test) = @_; # My::Config::Group, My::Test
+    if ($openai->{proc}) {
+      # Already started
+      return ;
+    }
```
svoj, line 22: It means single mockup API server for all mtr workers. It is probably alright for now.

One thing that I'm worried about: there can be race condition here. If multiple mtr workers hit this function simultaneously and both see empty `$openai->{proc}`.

plugin/gen_embedding/mysql-test/gen_embedding/suite.pm
```diff
@@ -0,0 +1,64 @@
...
+    my $python_path="python3"; # TODO: Maybe use a config option, could also be "python" or a specific path
+    my $success_file_path=$openai->value('api_response_filename');
+    my $python_server_path="$ENV{MTR_SUITE_DIR}/minimal_socket_server.py";
+    my $args;
+    &::mtr_init_args(\$args);
+    &::mtr_add_arg($args, $python_server_path);
+    &::mtr_add_arg($args, "$ENV{OPENAI_PORT}");
+    &::mtr_add_arg($args, $success_file_path);
```
svoj, line 30: I'd avoid redundant variables and pass values directly to `mtr_add_arg()`.

plugin/gen_embedding/mysql-test/gen_embedding/my.cnf
```diff
@@ -0,0 +1,8 @@
...
+
+[ENV]
+OPENAI_PORT = @OPT.port
```
svoj, line 8: This needs to be double checked. On my host mtr allocates 30 ports to the first worker: 19000..19029. mockup server uses port 19001. How it avoids port conflicts with e.g. replication or spider?

plugin/gen_embedding/mysql-test/gen_embedding/suite.pm
```diff
@@ -0,0 +1,64 @@
...
+     append       => 1,
+     nocore       => 1,
+    );
+}
+
+sub openai_wait {
+    my ($openai, $test) = @_; # My::Config::Group, My::Test
+    open(LOG, '<', $http_server_log);
+    do {
+      sleep 1 and seek LOG,0,1 until $_=<LOG>;
```
svoj, line 50: I'd do shorter `sleep(0.1)` here.

plugin/gen_embedding/mysql-test/gen_embedding/t/bad_response.test
```diff
@@ -0,0 +1,20 @@
...
+select vec_totext(generate_embedding_openai('text that does not exist in responses json', 'text-embedding-ada-002'));
+
+# Test that if some element of the API response changes in the future, for example the path of the embedding
+# vector the server does not crash and a NULL value is returned
+# responses.json includes a special entry with the key "wrong_embedding_path" for this purpose
+
+--replace_result $OPENAI_PORT OPENAI_PORT
+eval set generate_embedding_openai_host="http://127.0.0.1:$OPENAI_PORT/success";
```
svoj, line 19: `generate_embedding_openai_host` is already set to required value, so this is redundant.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_alter.test
```diff
@@ -0,0 +1,56 @@
...
+    c text, d text, 
+    b text as (concat(c,d)),
+    embedding vector(1536) as (generate_embedding_openai(b, "text-embedding-ada-002")) stored
+) engine=MyISAM;
+
+let $expensive_calls_begin= `$get_expensive_calls`;
+insert into exp_test_new (c,d) values ("ab", "bc");
+let $expensive_calls= `$get_expensive_calls`;
+let $res= `select $expensive_calls - $expensive_calls_begin`;
+echo $res;
```
svoj, line 16:
```
evalp select $expensive_calls - $expensive_calls_begin;
```

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_alter.test
```diff
@@ -0,0 +1,56 @@
...
+
+# Should not call the expensive function
+alter table exp_test_new add x int;
+let $expensive_calls= `$get_expensive_calls`;
+let $res= `select $expensive_calls - $expensive_calls_begin`;
+echo $res;
+select c, d, b, embedding is null from exp_test_new;
+
+# Should call the expensive function
+alter table exp_test_new modify c varchar(50);
```
svoj, line 26: Hmm... should it actually call the expensive function in this case? Data itself hasn't modified.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_alter.test
```diff
@@ -0,0 +1,56 @@
...
+let $expensive_calls= `$get_expensive_calls`;
+let $res= `select $expensive_calls - $expensive_calls_begin`;
+echo $res;
+select c, d, b, embedding is null from exp_test_new;
+
+# TODO: Should this call the expensive function since the type of the vcol is changed?
+# TODO: Currently it doesn't
```
svoj, line 33: ATTN @vuvova

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_alter.test
```diff
@@ -0,0 +1,56 @@
...
+alter table exp_test_new modify embedding vector(1537);
+let $expensive_calls= `$get_expensive_calls`;
+let $res= `select $expensive_calls - $expensive_calls_begin`;
+echo $res;
+select c, d, b, embedding is null from exp_test_new;
+
+# TODO: Same as above, what do we want to happen here, where a dep's type changed and expression?
```
svoj, line 40: I'd say "dep's" type change as such shouldn't involve "expensive function call", unless data wasn't truncated or somehow modified otherwise.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_alter_simple.test
```diff
@@ -0,0 +1,24 @@
+source include/have_gen_embedding.inc;
+
+--replace_result $OPENAI_PORT OPENAI_PORT
+eval set generate_embedding_openai_host="http://127.0.0.1:$OPENAI_PORT/success";
+
+drop table if exists exp_test_new;
```
svoj, line 6: Redundant.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_alter_simple.test
```diff
@@ -0,0 +1,24 @@
...
+insert into exp_test_new (doc) values ("abc");
+let $expensive_calls= `$get_expensive_calls`;
+let $res= `select $expensive_calls - $expensive_calls_begin`;
+echo $res;
+
+alter table exp_test_new add x int;
+let $expensive_calls= `$get_expensive_calls`;
+let $res= `select $expensive_calls - $expensive_calls_begin`;
+echo $res;
+select doc, x, embedding is null from exp_test_new;
```
svoj, line 22: It is essentially a copy of one of the tests from `expensive_alter.test`, right?

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_other_vcols.test
```diff
@@ -0,0 +1,93 @@
...
+
+# Make sure that updating another vcol that is not related to the expensive vcol does not update the expensive vcol as well
+# We also verify that the fields that are meant to be updated, are indeed updated by executing select statements
+# We don't verify the contents of the embedding vector in those select statements, as the status variables are 
+# sufficient indications that the expensive function is (or isn't) called
+
+--replace_result $OPENAI_PORT OPENAI_PORT
+eval set generate_embedding_openai_host="http://127.0.0.1:$OPENAI_PORT/success";
+
+create table expensive_test (idx int, doc text, model text,
```
svoj, line 12: Frankly, I’ve gotten so used to t1, t2, etc. that reading other table names hurts my eyes.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_other_vcols.test
```diff
@@ -0,0 +1,93 @@
...
+
+# This should not call the expensive function
+update expensive_test set d = "whatever2";
+select d, doc, c from expensive_test;
+let $expensive_calls= `$get_expensive_calls`;
+let $res= `select $expensive_calls - $expensive_calls_begin`;
+echo $res;
+
+# Cleanup
+drop table expensive_test;
```
svoj, line 93: I'd move all expensive tests to a single test file.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_rpl.cnf
```diff
@@ -0,0 +1,15 @@
+!include include/default_my.cnf
+
+[openai]
+listen = @ENV.OPENAI_PORT
+api_response_filename = @ENV.MTR_SUITE_DIR/json_responses/api_responses.json
+wrong_json_path_filename = @ENV.MTR_SUITE_DIR/json_responses/wrong_json_path.json
```
svoj, line 6: I don't see `wrong_json_path_filename` used anywhere.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_rpl.test
```diff
@@ -0,0 +1,68 @@
+# Test to check how expensive functions are treated in replication
```
svoj, line 1: Normally files like this are called `rpl_something`, not `something_rpl`.

plugin/gen_embedding/mysql-test/gen_embedding/t/expensive_rpl.test
```diff
@@ -0,0 +1,68 @@
...
+--connection master
+--replace_result $OPENAI_PORT OPENAI_PORT
+eval set generate_embedding_openai_host="http://127.0.0.1:$OPENAI_PORT/success";
+let $expensive_calls_begin= `$get_expensive_calls`;
+
+# Set slave's host to the correct value
+--connection slave
+source include/stop_slave.inc;
+--replace_result $OPENAI_PORT OPENAI_PORT
+eval set global generate_embedding_openai_host="http://127.0.0.1:$OPENAI_PORT/success";
```
svoj, line 21: Why slave requires this setting? ATTN @vuvova

plugin/gen_embedding/mysql-test/gen_embedding/t/greek_mockup.test
```diff
@@ -0,0 +1,18 @@
...
+
+set names greek;
+set character_set_results=utf8;
+
+--replace_result $OPENAI_PORT OPENAI_PORT
+eval set generate_embedding_openai_host="http://127.0.0.1:$OPENAI_PORT/success";
+set @model="text-embedding-ada-002";
+--disable_query_log
+set @text="�� ��� ��� ���� ��������� ��������� ��� ������� ��� ����������� ������������� �������� ��� ����� ����� ������ � �������� ��������� ���� ���������� ��� ��������.";
+--enable_query_log
```
svoj, line 15: I'd say disabling query log in this case can be misleading.

plugin/gen_embedding/mysql-test/gen_embedding/t/openai_greek_charset.test
```diff
@@ -0,0 +1,20 @@
+source include/have_gen_embedding.inc;
+--disable_query_log
+source $HOME/openai_api_key.inc;
+--enable_query_log
+
+if (!$GEN_EMBEDDING_OPENAI_PLUGIN) {
+  skip GEN_EMBEDDING_OPENAI_PLUGIN is not set;
+}
```
svoj, line 8: Why?

plugin/gen_embedding/mysql-test/gen_embedding/t/openai_greek_charset.test
```diff
@@ -0,0 +1,20 @@
...
+set character_set_results=utf8;
+
+eval set generate_embedding_openai_host="https://api.openai.com/v1/embeddings";
+set @model="text-embedding-ada-002";
+--disable_query_log
+set @text="�� ��� ��� ���� ��������� ��������� ��� ������� ��� ����������� ������������� �������� ��� ����� ����� ������ � �������� ��������� ���� ���������� ��� ��������.";
+--enable_query_log
+select vec_totext(generate_embedding_openai(@text, @model)) is null as is_null;
+
+set names default;
```
svoj, line 20: IIUC the test is essentially equivalent to `greek_mocup.test`, the only difference is `generate_embedding_openai_host` setting. There're facilities that allow to avoid such duplication.

plugin/gen_embedding/mysql-test/gen_embedding/t/openai_server.test
```diff
@@ -0,0 +1,48 @@
+source include/have_gen_embedding.inc;
+--disable_query_log
+source $HOME/openai_api_key.inc;
+--enable_query_log
+
+# This file contains tests that use the actual OpenAI API endpoint (instead of our mockup server)
+
+if (!$GEN_EMBEDDING_OPENAI_PLUGIN) {
+  skip GEN_EMBEDDING_OPENAI_PLUGIN is not set;
+}
```
svoj, line 10: Looks redundant.

plugin/gen_embedding/mysql-test/gen_embedding/t/openai_server.test
```diff
@@ -0,0 +1,48 @@
...
+
+drop table t1;
+
+# Test that an error HTTP response code from OpenAI is handled correctly
+# The error might be caused by a wrong api key, which is tested here
+
+set generate_embedding_openai_api_key="wrong-api-key";
+
+call mtr.add_suppression("GENERATE_EMBEDDING_OPENAI: Bad http response code: 401");
+select vec_totext(generate_embedding_openai(@text, @model));
```
svoj, line 48: This test covers more than basic OpenAI server verification would need. I'd either simplify it or make it work with both mockup server and OpenAI.

plugin/gen_embedding/mysql-test/gen_embedding/t/reassign_variables.test
```diff
@@ -0,0 +1,19 @@
+source include/have_gen_embedding.inc;
+
+# Verify that plugin session variables can be assigned multiple times without issues
+# This test is created because a previous plugin design had was handling
```
svoj, line 4: s/had//

plugin/gen_embedding/mysql-test/gen_embedding/t/status_variables.test
```diff
@@ -0,0 +1,44 @@
+source include/have_gen_embedding.inc;
+
+--skip "Skipping for now due to plugin status variables not being reset by flush status"
```
svoj, line 3: This needs to be verified.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+
+  
+  class Create_func : public Create_func_arg2
+  {
+  public:
+    Item *create_2_arg(THD *thd, Item *arg1, Item *arg2) override
+    {
+      return new (thd->mem_root) Item_func_gen_embedding(thd, arg1, arg2);
+    }
+  };
```
svoj, line 296: I doubt OpenAI promises stable embedding generation. It is unclear if it is safe for caching, but it must be unsafe for SBR and needs something like:
```
thd->lex->set_stmt_unsafe(LEX::BINLOG_STMT_UNSAFE_SYSTEM_FUNCTION);
thd->lex->uncacheable(UNCACHEABLE_RAND);
```

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+  read_data->write(static_cast<char *>(contents), realsize);
+  return realsize;
+}
+
+
+class Item_func_gen_embedding: public Item_str_func 
+{
+  String tmp_js, api_response, tmp_str, post_fields;
+  json_path_with_flags *json_path;
+  std::unordered_map<std::string, int> model_dimensions= {
```
svoj, line 75: `static constexpr`?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+#include <mysql/plugin_function.h>
+#include <curl/curl.h>
+#include <sstream>
+#include "sql_type_vector.h"
+#include "item_jsonfunc.h"
+#include <unordered_map>
+#include "json_lib.h"
+
+ulonglong curl_requests= 0; /* Number of total curl requests */
+ulonglong successful_curl_requests= 0; /* Number of successful curl requests */
```
svoj, line 30: `static`

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    curl_easy_setopt(hnd, CURLOPT_POSTFIELDS, post_fields.ptr());
+    curl_easy_setopt(hnd, CURLOPT_POSTFIELDSIZE_LARGE, (curl_off_t)post_fields.length());
+    curl_easy_setopt(hnd, CURLOPT_HTTPHEADER, slist1);
+    curl_easy_setopt(hnd, CURLOPT_USERAGENT, "curl/8.5.0");
+    curl_easy_setopt(hnd, CURLOPT_MAXREDIRS, 50L);
+    curl_easy_setopt(hnd, CURLOPT_HTTP_VERSION, (long)CURL_HTTP_VERSION_2TLS);
+    curl_easy_setopt(hnd, CURLOPT_FTP_SKIP_PASV_IP, 1L);
+    curl_easy_setopt(hnd, CURLOPT_TCP_KEEPALIVE, 1L);
+    curl_easy_setopt(hnd, CURLOPT_WRITEFUNCTION, WriteMemoryCallback);
+    curl_easy_setopt(hnd, CURLOPT_WRITEDATA, &read_data_stream);
```
svoj, line 180: I'd add also `CURLOPT_CONNECTTIMEOUT` / `CURLOPT_TIMEOUT`. Also think how to make `curl_easy_perform()` killable. Otherwise it is possible to exhaust MariaDB server with unkillable connections.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+      null_value= true;
+      return 1;
+    }
+    // Model for the OpenAI API
+    String *model= args[1]->val_str(&tmp_str);
+    if (!model) 
+    {
+      null_value= true;
+      return 1;
+    }
```
svoj, line 109: Should we make `model` argument optional and use some sane default model?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    curl_easy_cleanup(hnd);
+    hnd= NULL;
+    curl_slist_free_all(slist1);
+    slist1= NULL;
+    return 1;
+  }
+
+  
+  int parse_vector(String *buf, json_engine_t je, CHARSET_INFO *cs, const uchar *start, const uchar *end);
+  String *read_json(String *str, json_value_types *type, char **out_val);
```
svoj, line 219: It looks inconsistent that `make_openai_request()` definition is kept here while `parse_vector()` and `read_json()` are moved out.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    static Create_func creator;
+    static Plugin_function descriptor(&creator);
+    return &descriptor;
+  }
+};
+
+
+String *Item_func_gen_embedding::read_json(String *str,
+                                          json_value_types *type,
+                                          char **out_val)
```
svoj, line 310: Looks like `type` and `out_val` are unused.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+  size_t realsize= size * nmemb;
+  std::ostringstream *read_data= static_cast<std::ostringstream *>(userp);
+  read_data->write(static_cast<char *>(contents), realsize);
+  return realsize;
+}
+
+
+class Item_func_gen_embedding: public Item_str_func 
+{
+  String tmp_js, api_response, tmp_str, post_fields;
```
svoj, line 70: `tmp_js` unused.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+      and model_name is the name of the model to be used for embedding generation.
+      We need to allocate enough space for the JSON string
+      At worst case, each input character needs to be escaped, hence the 2 multiplier below
+    */
+    max_chars= (2 * input->length()/input->charset()->mbminlen +
+                      model->length()/model->charset()->mbminlen +
+                      strlen("{\"input\": \"") + strlen("\", \"model\": \"") +
+                      strlen("\",\"encoding_format\": \"float\"}") + 1);
+    strLen= max_chars * cs_openai->mbmaxlen;
+    post_fields.realloc_with_extra_if_needed(strLen);
```
svoj, line 138: This calculation should be reevaluated.
1. escape multiplier is much bigger for `json_escape()`, `st_append_escaped()` it claims value of 12
2. `strlen()` on constant strings should be avoided
3. `realloc_with_extra_if_needed()` looks excessive

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+      null_value= true;
+      goto cleanup;
+    }
+    post_fields.length(post_fields.length()+strLen); // We need to manually update the length
+    post_fields.append(STRING_WITH_LEN("\", \"model\": \""));
+    post_fields.append(model->c_ptr(), model->length());
+    post_fields.append(STRING_WITH_LEN("\",\"encoding_format\": \"float\"}"));
+    
+    // Prepare the request to OpenAI API
+    slist1= NULL;
```
svoj, line 164: Redundant.

plugin/gen_embedding/plugin.cc
svoj: I support this idea.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+  std::ostringstream *read_data= static_cast<std::ostringstream *>(userp);
+  read_data->write(static_cast<char *>(contents), realsize);
+  return realsize;
+}
+
+
+class Item_func_gen_embedding: public Item_str_func 
+{
+  String tmp_js, api_response, tmp_str, post_fields;
+  json_path_with_flags *json_path;
```
svoj, line 71: It doesn't have to be a pointer, right? Moreover it is a `read_json()` local thing, it doesn't have to be class member.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+
+String *Item_func_gen_embedding::read_json(String *str,
+                                          json_value_types *type,
+                                          char **out_val)
+{
+  String *js= &api_response;
+  json_engine_t je;
+  json_path_t p;
+  const uchar *value;
+  int not_first_value= 0, count_path= 0;
```
svoj, line 314: Both variables are useless.

plugin/gen_embedding/plugin.cc
svoj, line 109: Should we also add `dimensions` argument?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+                        "Error converting input string from %s to UTF-8 charset", 
+                        ME_ERROR_LOG | ME_WARNING, input->charset()->cs_name.str);
+      }
+      null_value= true;
+      goto cleanup;
+    }
+    post_fields.length(post_fields.length()+strLen); // We need to manually update the length
+    post_fields.append(STRING_WITH_LEN("\", \"model\": \""));
+    post_fields.append(model->c_ptr(), model->length());
+    post_fields.append(STRING_WITH_LEN("\",\"encoding_format\": \"float\"}"));
```
svoj, line 161: I'd rather use `base64` instead of `float`: 2-2.5x less data to transfer and much cheaper to parse.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+  *type= JSON_VALUE_NULL;
+
+  // str will never be NULL here, since it comes from val_str()
+  str->set_charset(&my_charset_bin);
+  str->length(0);
+
+  json_get_path_start(&je, js->charset(),(const uchar *) js->ptr(),
+                      (const uchar *) js->ptr() + js->length(), &p);
+
+  while (json_get_path_next(&je, &p) == 0)
```
svoj, line 335: I'd avoid this loop unless we really want to handle multiple embeddings at once. We probably don't within the scope of this SQL function.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+        my_printf_error(1, "GENERATE_EMBEDDING_OPENAI: "
+          "Model %s is not supported", ME_ERROR_LOG | ME_WARNING, model_name.c_str());
+      }
+    }
+
+    decimals= 0;
+    MEM_ROOT *root= thd->mem_root;
+    json_path= (json_path_with_flags *) alloc_root(root, sizeof(json_path_with_flags));
+    // We want to store max_dimensions floats, each 4 bytes 
+    fix_length_and_charset(max_dimensions * 4, &my_charset_bin);
```
svoj, line 250: Why not `max_dimensions * sizeof(float)`?

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+    {
+      null_value= true;
+      return NULL;
+    }   
+    return read_json(buf, &type, &value);
+  }
+  
+  LEX_CSTRING func_name_cstring() const override
+  {
+    static LEX_CSTRING name= "GENERATE_EMBEDDING_OPENAI"_LEX_CSTRING;
```
svoj, line 282: Given that most modern LLMs expose REST APIs, I’d prefer a generic function such as `AI_EMBEDDING(data[, model])` that abstracts away provider-specific differences. This would avoid maintaining multiple plugins with nearly identical implementations.

plugin/gen_embedding/plugin.cc
```diff
@@ -0,0 +1,449 @@
...
+
+static MYSQL_THDVAR_STR(host,
+  PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_MEMALLOC,
+  "OpenAI API host, can be set to 'https://api.openai.com/v1/embeddings' or a custom endpoint",
+  NULL, NULL, "");
+
+static MYSQL_THDVAR_STR(api_key,
+  PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_MEMALLOC,
+  "OpenAI API key",
+  NULL, NULL, "");
```
svoj, line 41: The fact that OpenAI key is exposed via session variable needs to be reevaluated.

1. value may shortly become visible to `SHOW PROCESSLIST`
2. value may be written to query log
3. unsure if it matters, but DBA can't set per-server key without exposing it to users
4. are there other ways for the value to become unintentionally visible?

svoj (changes requested):
This review spans only `plugin/gen_embedding` changes.

Coding style is inconsistent: spacing, lines over 80 characters, trailing spaces, variable names.

I'd move tests from `plugin/gen_embedding/mysql-test/gen_embedding/t/` `plugin/gen_embedding/mysql-test/gen_embedding/r/` to `plugin/gen_embedding/mysql-test/gen_embedding`.

I'd move the whole plugin either to `plugin/ai/generate_embedding_openai` or `plugin/func_generate_embedding_openai`.

sql/table.cc
```diff
@@ -8320,7 +8320,33 @@ bool TABLE::mark_virtual_columns_for_write(bool insert_fl)
...
+            needs to be updated so we put it in the write_set
+            To update it, we need the values of all the deps, so we put them
+            in the read_set
+           */
+          bitmap_set_bit(write_set, tmp_vfield->field_index);
+          mark_virtual_column_with_deps(tmp_vfield);
+          bitmap_updated= true;
+        }
+      }
+      else if (insert_fl)
```
svoj, line 8349: Looks like this change is redundant after a4e4a56720c.

sql/sql_table.cc
```diff
@@ -12703,6 +12718,54 @@ copy_data_between_tables(THD *thd, TABLE *from, TABLE *to,
...
+              or 
+              def->field->vcol_info && (*ptr)->vcol_info->is_equal(def->field->vcol_info)) 
+            */
+        {
+          /* 
+            Check if vcol can be copied
+            If not stored, it cannot be copied
+            If not expensive, we decided for now to not copy
+          */ 
+          if ((*ptr)->vcol_info->is_stored() && (*ptr)->vcol_info->expr->is_expensive())
```
svoj, line 12761: `is_expensive()` existed prior to this patch:
```
      This function is now used to prevent evaluation of expensive subquery
      predicates during the optimization phase. It also prevents evaluation
      of predicates that are not computable at this moment.
```

It is reused here with a different meaning and it spans more functions like `get_lock()`, `sleep()`, UDF, etc.

Do we need to re-evaluate stored virtual columns at all?

svoj (changes requested):
`sql/` comments inline.

PR #4097, opened 2025-06-09
Fix bootstrapping Galera nodes on FreeBSD 15.0

fsmeets:
## Description
Fix bootstrapping Galera nodes on FreeBSD 15.0

Commit https://github.com/MariaDB/server/commit/606c867e7fec4507bd1225f9abf2d0aa664b6ab6
broke bootstrapping nodes on FreeBSD. We should not need LD_LIBRARY_PATH
trickery on FreeBSD, so skip setting it same as on Darwin. Alternatively
putting /lib before /usr/lib also fixes it.

WSREP_SST: [INFO] Preparing the backup at /var/db/mysql/.sst (20250608 14:44:04.000)
WSREP_SST: [INFO] Evaluating /usr/local/bin/mariadb-backup --prepare --log-bin='/var/db/mysql-logs/mysql-bin' --use-memory=12G --target-dir='/var/db/mysql/.sst' --datadir='/var/db/mysql/.sst' --mysqld-args '--basedir=/usr/local' '--plugin-dir=/usr/local/lib/mysql/plugin' '--wsrep_provider=/usr/local/lib/libgalera_smm.so' '--wsrep_on=ON' '--wsrep_provider=/usr/local/lib/libgalera_smm.so' '--wsrep_on=ON' '--log-error=/var/db/mysql/db4.nexxtmobile.de.err' '--pid-file=/var/run/mysql/mysqld.pid' '--wsrep_start_position=00000000-0000-0000-0000-000000000000:-1,0-0-0' > '/var/db/mysql/mariabackup.prepare.log' 2>&1 (20250608 14:44:04.000)
WSREP_SST: [ERROR] Cleanup after exit with status: 1 (20250608 14:44:04.000)

With the following error in /var/db/mysql/mariabackup.prepare.log:
ld-elf.so.1: /usr/local/bin/mariadb-backup: Undefined symbol "_ZNSt3__117bad_function_callD1Ev"

I did not look for the root cause in FreeBSD 15.0 that changed this
behaviour. Calling mariadb-backup directly from the shell works fine on
FreeBSD 14.x and 15.x when LD_LIBRARY_PATH is not set (when using
FreeBSD packages or installing from FreeBSD ports). Setting
LD_LIBRARY_PATH manually like the script does, breaks mariadb-backup
with the Undefined symbol error on FreeBSD 15.0 but not on 14.x.

## Release Notes
Fix bootstrapping Galera nodes with mariadb-backup on FreeBSD 15.0

## How can this PR be tested?
Try to bootstrap a new Galera node running on FreeBSD 15.0.

## PR quality check


fsmeets:
Something went wrong. It has 6 spurious commits. I will reopen a fixed PR.

svoj:
@fsmeets you can force push to `freebsd_galera_mariadb-backup_bootstrap` branch to update PR.
Your branch is based on 10.5, while PR is created against main. These are unmerged changes.

fsmeets:
> @fsmeets you can force push to `freebsd_galera_mariadb-backup_bootstrap` branch to update PR. Your branch is based on 10.5, while PR is created against main. These are unmerged changes.

Thank you for the suggestion, honestly I don't know what to do. Even after working with git for 5 years, I still loathe it. I can deal with FreeBSD's git workflows, but that's it.
I first had the patch in main, but the PR template says to create it against the earliest maintained branch exhibiting the issue, then "Submitting a pull request" suggests creating the PR against main. Now I'm stuck. 

It would have been so much simpler to just dump this as a local patch in FreeBSD ports ;)

svoj:
@fsmeets no problem, I have edited PR now it looks alright. MariaDB developers will reevaluate target version.

grooverdan:
@brad0 does the same apply to OpenBSD?

brad0:
@grooverdan Sorry I saw this and meant to reply but forgot.

I think the logic on this should be inverted. This is typically something you'd see on Linux or maybe Solaris.

grooverdan:
> @grooverdan Sorry I saw this and meant to reply but forgot.
> 
> I think the logic on this should be inverted. This is typically something you'd see on Linux or maybe Solaris.

I'm loosing track too :-).

Thanks for getting back.

Seems the `LD_LIBRARY_PATH` was introduced because of Centos in commit 606c867e7fec4507bd1225f9abf2d0aa664b6ab6 for some reason extending the PATH and adding the same to LD_LIBRARY_PATH.

@sysprg why did this occur?

If you hand compiled mariabackup for Centos with /usr/local or non-standard paths you could use a ` -Wl,-rpath=/usr/local/...` and not need a `LD_LIBRARY_PATH`. So I can't see why `LD_LIBRARY_PATH` to non-distro locations would be set in this script.

sysprg:
@fsmeets Thanks for finding the cause of this non-trivial failure, I merged your patch into the main branch for 10.6 (https://github.com/MariaDB/server/commit/b5d79b86b29c126381907507d43b7c44908ee58e), but then added a rework that solves the problem by rearranging the default paths in accordance with historical practice, because this approach will be more reliable for users who do not have all paths set in LD_LIBRARY_PATH on any system, not just FreeBSD (new patch is https://github.com/MariaDB/server/commit/03ae1f5ab394a0a57b1eb9245a0d62dc93b5b731).

sysprg:
@grooverdan Almost all of our SST scripts included some kind of PATH and/or LD_LIBRARY_PATH setting, previously this was done each time in some new way and with an individual combination of paths. I tried to remove these settings from each individual script and put them all in one point in the main wsrep_sst_common.sh script, but I violated the historically established "standard" according to which /lib goes before /usr/lib, and not vice versa. It would be possible to not set anything at all, but these settings in scripts have been around for many years, if not a couple of decades, there were probably many systems or users who did not list some important path in their PATH or LD_LIBRARY_PATH, and then they could not find or run this or that system utility, so in ancient times someone added something to each of the scripts (as problems were solved), and as a result we have what we have. I looked at the discussions on the forums about the correct order of paths and now I have put it in line with established practice, since there is apparently no clear standard for this.

PR #4098, opened 2025-06-10
MDEV-36977: Histogram code lacks coverage for non-latin characters

mariadb-lenastartseva:
- [x] *The Jira issue number for this PR is: MDEV-36977*

## Description
Added new testcases for histogram:
- 3-byte characters
- characters outside Basic Multilingual Plane

## Release Notes
Per MDEV

## How can this PR be tested?
The change contains only tests

## PR quality check


mysql-test/main/statistics_json.test
```diff
@@ -531,4 +531,157 @@ drop table t1;
 # JSON_UNQUOTE was touched by this patch also
 select JSON_UNQUOTE(CONVERT('"ФФ"' using cp1251));
 
+
+--echo #
+--echo # MDEV-36977 Histogram code lacks coverage for non-latin characters
+--echo #
+
+create table t1 (
+  col1 varchar(10) charset utf8
```
spetrunia: Please specify the collation explicitly here (otherwise it uses the default which changes across versions).

mysql-test/main/statistics_json.test
```diff
@@ -531,4 +531,157 @@ drop table t1;
...
+);
+set names utf8;
+select hex('Ꙃ'), collation('Ꙃ');
+insert into t1 values
+('Ꙩ'),('Ꙛ'),('ꙮ'),('Ꙃ'),('Ꚛ'),('ꘐ'),('ꘜ'),('ꕫ'),('ꖿ'), ('ꙛ');
+
+analyze table t1 persistent for all;
+select hex(col1) from t1;
```
spetrunia: Please use  (here and in other places) `ORDER BY col1` so that one can see how the symbols are ordered.

mysql-test/main/statistics_json.test
```diff
@@ -531,4 +531,157 @@ drop table t1;
...
+select json_detailed(json_extract(histogram, '$**.histogram_hb'))
+from mysql.column_stats where db_name=database() and table_name='t1';
+
+explain extended select * from t1 where col1 < 'ꕫ';
```
spetrunia: Instead of EXPLAIN EXTENDED, please use ANALYZE.  It prints filtered and r_filtered which makes the [mis]matches between statistics and data much clearer.

spetrunia (approved):

PR #4103, opened 2025-06-12
MDEV-36840 Seconds_Behind_Master Spike at Log Rotation on Parallel Replica

ParadoxV5:
* [x] *The Jira issue number for this PR is: [MDEV-36840](https://jira.mariadb.org/browse/MDEV-36840)*

## Description
TODO: An amazing description should answer some questions like:

## Release Notes

## How can this PR be tested?

## PR quality check

### DIFF
--- a/mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike.test
+++ b/mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike.test
@@ -18,6 +18,8 @@
 #
 # References:
 #   MDEV-16091: Seconds_Behind_Master spikes to millions of seconds
+#   MDEV-25999: Unrealistic "Seconds behind master" value
+#   MDEV-36840: Seconds_Behind_Master Spike at Log Rotation on Parallel Replica
 #
 --source include/have_debug.inc
 --source include/have_debug_sync.inc
@@ -29,12 +31,10 @@
 # Using_Gtid needs to start as NO before updating debug_dbug
 --source include/stop_slave.inc
 CHANGE MASTER TO MASTER_USE_GTID=NO;
---source include/start_slave.inc
-
---source include/stop_slave.inc
-SET @save_dbug= @@GLOBAL.debug_dbug;
-SET @@global.debug_dbug="+d,pause_sql_thread_on_relay_fde_after_trans";
-SET @@global.debug_dbug="+d,negate_clock_diff_with_master";
+if ($slave_parallel_threads)
+{
+  --eval SET @@GLOBAL.slave_parallel_threads= $slave_parallel_threads
+}
 --source include/start_slave.inc
 
 --let $sleep_time=2
@@ -45,55 +45,36 @@ SET @@global.debug_dbug="+d,negate_clock_diff_with_master";
 --connection master
 --echo # Write events to ensure slave will be consistent with master
 create table t1 (a int);
-insert into t1 values (1);
---let $t_master_events_logged= `SELECT UNIX_TIMESTAMP()`
+--sync_slave_with_master
 
+--echo # Enable breakpoints
+SET @save_dbug= @@GLOBAL.debug_dbug;
+SET @@GLOBAL.debug_dbug=
+  "+d,pause_sql_thread_on_fde,negate_clock_diff_with_master";
+
+--connection master
 --echo # Flush logs on master forces slave to generate a Format description
 --echo # event in its relay log
 flush logs;
 
 --connection slave
 --echo # On the next FDE, the slave should have the master CREATE/INSERT events
+--sleep $sleep_time
 SET DEBUG_SYNC='now WAIT_FOR paused_on_fde';
-select count(*)=1 from t1;
 
 --echo # The relay log FDE has been processed - here we check to ensure it was
 --echo # not considered in Seconds_Behind_Master calculation
---connection slave1
-let $sbm= query_get_value(SHOW SLAVE STATUS, Seconds_Behind_Master, 1);
---let $t_now= `SELECT UNIX_TIMESTAMP()`
+# `Seconds_Behind_Master` must remains 0.
+# * MDEV-16091:
+#   It should not point at the FDE's `start_slave.inc` timestamp.
+# * MDEV-25999 & MDEV-36840:
+#   It should not point at all in a new stream, for FDEs are not content.
+--let $status_items= Seconds_Behind_Master
+--source include/show_slave_status.inc
 
-# Ensure Seconds_Behind_Master does not point beyond when we have proven the
-# events we have proven to have executed. The extra second is needed as a
-# buffer because the recorded times are not exact with when the events were
-# recorded on the master.
-if(`select $sbm > $t_now - $t_master_events_logged + 1`)
-{
-    die "A relay log event was incorrectly used to set Seconds_Behind_Master";
-}
-
---echo # Safely resume slave SQL thread
-
---let $dbug_wait_state="debug sync point: now"
---echo # Prove SQL thread is in state $dbug_wait_state
---let $wait_condition= SELECT STATE=$dbug_wait_state from information_schema.PROCESSLIST where COMMAND="Slave_SQL"
---source include/wait_condition.inc
-
+--echo # Resume slave SQL (worker) thread
 SET @@global.debug_dbug="-d,pause_sql_thread_on_fde";
 SET DEBUG_SYNC='now SIGNAL sql_thread_continue';
-
-# We have to wait for the SQL thread to acknowledge the sql_thread_continue
-# signal. Otherwise the below RESET command can overwrite the signal before
-# the SQL thread is notified to proceed, causing it to "permanently" become
-# stuck awaiting the signal (until timeout is reached).
-
---echo # Wait for SQL thread to continue into normal execution
---let $wait_condition= SELECT STATE!= $dbug_wait_state from information_schema.PROCESSLIST where COMMAND="Slave_SQL"
---source include/wait_condition.inc
-
-# Reset last sql_thread_continue signal
-SET DEBUG_SYNC='RESET';
-
 
 --echo #
 --echo # MDEV-29639
@@ -178,6 +159,10 @@ while (!$caught_up)
 --source include/stop_slave.inc
 set debug_sync='RESET';
 SET @@global.debug_dbug=$save_dbug;
+if ($slave_parallel_threads)
+{
+  SET @@GLOBAL.slave_parallel_threads= DEFAULT;
+}
 --source include/start_slave.inc
 --source include/sync_with_master_gtid.inc
 
--- a/sql/rpl_parallel.cc
+++ b/sql/rpl_parallel.cc
@@ -104,6 +104,14 @@ handle_queued_pos_update(THD *thd, rpl_parallel_thread::queued_event *qev)
       (e->force_abort && !rli->stop_for_until))
     return;
 
+  #ifdef ENABLED_DEBUG_SYNC
+    DBUG_EXECUTE_IF("pause_sql_thread_on_fde",
+      DBUG_ASSERT(!debug_sync_set_action(thd, STRING_WITH_LEN(
+        "now SIGNAL paused_on_fde WAIT_FOR sql_thread_continue"
+      )));
+    );
+  #endif
+
   mysql_mutex_lock(&rli->data_lock);
   cmp= compare_log_name(rli->group_relay_log_name, qev->event_relay_log_name);
   if (cmp < 0)
--- a/sql/slave.cc
+++ b/sql/slave.cc
@@ -7808,9 +7808,9 @@ static IO_CACHE *reopen_relay_log(Relay_log_info *rli, const char **errmsg)
 
 /**
   Reads next event from the relay log.  Should be called from the
-  slave IO thread.
+  slave SQL thread.
 
-  @param rli Relay_log_info structure for the slave IO thread.
+  @param rgi rpl_group_info structure for the slave SQL thread.
 
   @return The event read, or NULL on error.  If an error occurs, the
   error is reported through the sql_print_information() or
--- a/sql/slave.cc
+++ b/sql/slave.cc
@@ -4511,15 +4511,6 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli,
     {
       Gtid_log_event *gev= static_cast<Gtid_log_event *>(ev);
 
-#ifdef ENABLED_DEBUG_SYNC
-    DBUG_EXECUTE_IF(
-        "pause_sql_thread_on_relay_fde_after_trans",
-        {
-          DBUG_SET("-d,pause_sql_thread_on_relay_fde_after_trans");
-          DBUG_SET("+d,pause_sql_thread_on_next_relay_fde");
-        });
-#endif
-
       /*
         For GTID, allocate a new sub_id for the given domain_id.
         The sub_id must be allocated in increasing order of binlog order.
@@ -4671,16 +4662,13 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli,
     wsrep_after_statement(thd);
 #endif /* WITH_WSREP */
 #ifdef ENABLED_DEBUG_SYNC
-    DBUG_EXECUTE_IF(
-        "pause_sql_thread_on_next_relay_fde",
-        if (ev && typ == FORMAT_DESCRIPTION_EVENT &&
-            ((Format_description_log_event *) ev)->is_relay_log_event()) {
-          DBUG_ASSERT(!debug_sync_set_action(
-              thd,
-              STRING_WITH_LEN(
-                  "now SIGNAL paused_on_fde WAIT_FOR sql_thread_continue")));
-          DBUG_SET("-d,pause_sql_thread_on_next_relay_fde");
-        });
+    // Note: Parallel Replication does not hit this point.
+    DBUG_EXECUTE_IF("pause_sql_thread_on_fde",
+      if (typ == FORMAT_DESCRIPTION_EVENT)
+        DBUG_ASSERT(!debug_sync_set_action(thd, STRING_WITH_LEN(
+          "now SIGNAL paused_on_fde WAIT_FOR sql_thread_continue"
+        )));
+    );
 #endif
 
     DBUG_RETURN(exec_res);
--- a/mysql-test/suite/rpl/r/rpl_seconds_behind_master_spike-parallel.result
+++ b/mysql-test/suite/rpl/r/rpl_seconds_behind_master_spike-parallel.result
@@ -3,6 +3,7 @@ include/master-slave.inc
 connection slave;
 include/stop_slave.inc
 CHANGE MASTER TO MASTER_USE_GTID=NO;
+SET @@GLOBAL.slave_parallel_threads= 1;
 include/start_slave.inc
 # Future events must be logged at least 2 seconds after
 # the slave starts
@@ -23,7 +24,7 @@ connection slave;
 SET DEBUG_SYNC='now WAIT_FOR paused_on_fde';
 # The relay log FDE has been processed - here we check to ensure it was
 # not considered in Seconds_Behind_Master calculation
-Seconds_Behind_Master = '0'
+Seconds_Behind_Master = '2'
 # Resume slave SQL (worker) thread
 SET @@global.debug_dbug="-d,pause_sql_thread_on_fde";
 SET DEBUG_SYNC='now SIGNAL sql_thread_continue';
@@ -50,6 +51,7 @@ set debug_sync='now wait_for paused_on_event';
 include/stop_slave.inc
 set debug_sync='RESET';
 SET @@global.debug_dbug=$save_dbug;
+SET @@GLOBAL.slave_parallel_threads= DEFAULT;
 include/start_slave.inc
 include/sync_with_master_gtid.inc
 connection master;
--- a/mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike-parallel.test
+++ b/mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike-parallel.test
@@ -1,0 +1,2 @@
+--let $slave_parallel_threads= 1
+--source ./rpl_seconds_behind_master_spike.test


ParadoxV5:
`rpl.rpl_parallel_sbm` failed on [amd64-ubuntu-2004-debug/31634](https://buildbot.mariadb.org/#/builders/369/builds/31634) but not locally.

Update: And now it passes. 🤨

sql/rpl_parallel.cc
```diff
@@ -3050,13 +3065,12 @@ rpl_parallel::stop_during_until()
...
 
 
-bool
-rpl_parallel::workers_idle(Relay_log_info *rli)
+void rpl_parallel::update_workers_idle(Relay_log_info *rli)
 {
   mysql_mutex_assert_owner(&rli->data_lock);
-  return !rli->last_inuse_relaylog ||
-    rli->last_inuse_relaylog->queued_count ==
-    rli->last_inuse_relaylog->dequeued_count;
```
ParadoxV5, line 3067: This previous iteration checked null, but when does `Relay_log_info::last_inuse_relaylog` become `nullptr`?

And how could the old code only check the last in-use relay log rather than all of them?
Wouldn’t the last one have a blank queue after a relay log switch?

sql/slave.cc
```diff
@@ -4448,13 +4448,11 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli,
...
+        The check for `rli->worker_threads_caught_up` ensures the worker threads
+        are all practically idle (i.e. all user events have been executed).
       */
       if ((unlikely(rli->last_master_timestamp == 0) ||
-           (rli->sql_thread_caught_up &&
-            (rli->last_inuse_relaylog->queued_count ==
-             rli->last_inuse_relaylog->dequeued_count))) &&
+           (rli->sql_thread_caught_up && rli->worker_threads_caught_up)) &&
           event_can_update_last_master_timestamp(ev))
       {
```
ParadoxV5: This section is curiously similar to the serial branch fifty lines above, but without adding the `ev->exec_time`.
Was it intended?

sql/rpl_rli.h
```diff
@@ -621,10 +627,14 @@ struct inuse_relaylog {
...
-  /* Number of events completed by worker threads. */
-  Atomic_counter<int64> dequeued_count;
+  /** Number of events in this relay log queued for worker threads.
+    @pre Hold `rli->data_lock` while accessing
+  */
+  int64_t queued_count;
+  /** Number of events completed by worker threads.
+    @pre Hold `rli->data_lock` while accessing
+  */
+  int64_t dequeued_count;
```
ParadoxV5: Now that every access has a reason to hold `rli->data_lock`, this pair no longer needs to be atomic.

sql/rpl_parallel.cc
```diff
@@ -104,6 +104,17 @@ handle_queued_pos_update(THD *thd, rpl_parallel_thread::queued_event *qev)
...
+    DBUG_EXECUTE_IF("pause_sql_thread_on_fde",
+      DBUG_ASSERT(!debug_sync_set_action(thd, STRING_WITH_LEN(
+        "now SIGNAL paused_on_fde WAIT_FOR sql_thread_continue"
+      )));
+      DBUG_ASSERT(!debug_sync_set_action(thd, STRING_WITH_LEN(
+        "now SIGNAL main_sql_thread_continue"
+      )));
+    );
+  #endif
+
```
ParadoxV5, line 117: > Because Parallel Replication queues Format Description Events as position update metaevents, this variant uses a separate set of API-compatible breakpoints to pause both the main and worker threads.
> ⸺ commit message

sql/rpl_parallel.cc
```diff
@@ -2034,7 +2045,11 @@ rpl_parallel_thread::inuse_relaylog_refcount_update()
   inuse_relaylog *ir= accumulated_ir_last;
   if (ir)
   {
-    ir->dequeued_count+= accumulated_ir_count;
+    Relay_log_info *rli= ir->rli;
+    mysql_mutex_lock(&rli->data_lock);
+      ir->dequeued_count+= accumulated_ir_count;
+      rpl_parallel::update_workers_idle(rli);
+    mysql_mutex_unlock(&rli->data_lock);
```
ParadoxV5: #### Note from MDEV-25999
If the replica is not idle, it should remain not idle even if the next event is internal, since it doesn’t want to dip `Seconds_Behind_Master` to 0 in case there are more non-internal events after this one.
Only at EOF can the replica return to idle with certainty.

Therefore, worker threads *should*n’t care if they’re dequeuing internal (meta)events either.

mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike.combinations
ParadoxV5, line 1: Looks like the configs in a `.combinations` file apply to every server in a test.

mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike.test
ParadoxV5, line 1: I see more room for optimization in this test… but this refactor goes to `main` after this PR merges up.

sql/rpl_parallel.cc
bnestere: A couple things cause `last_inuse_relaylog` to nullify:
 1. At SQL thread start, there is a brief moment in-between `reset_inuse_relaylog()` and `alloc_inuse_relaylog()` where this was NULL. Prior to 13a9395a6acf1bd90370bd6fe7b4637597b921dd, this section was not locked, and `SHOW SLAVE STATUS` would result in a segfault in this point.
 2. When the SQL shuts down
 3. When the `Master_info` parent object is destroyed (e.g. `RESET SLAVE` or `CHANGE MASTER TO`). Note the `Relay_log_info` member variable is not a reference, but belongs within the memory of the `Master_info`.

bnestere: And to

> And how could the old code only check the last in-use relay log rather than all of them?
> Wouldn’t the last one have a blank queue after a relay log switch?

Hmm good observation, there may indeed be a small gap of time right after rotating, and before the new binlog's `Format_description_event` is queued where we temporarily appear up-to-date. Can you test that? I think your patch would still be vulnerable to this.

sql/slave.cc
bnestere: Do you mean should _this_ check have `+ ev->exec_time`? If so no, it is intentional to leave it out. The idea being that parallel replication has special semantics for SBM, where it is updated at transaction commit. Then this `lmt` update here is done at the beginning of transaction execution, so we try to keep SBM consistent with the running state and use the transaction begin time of the event.

bnestere (changes requested):
Thanks for a well thought out patch, @ParadoxV5 !

Generally I think it is ok, but I think @knielsen should also review it to see if he has another idea which wouldn't incur additional locking (I don't have any more ideas apart from the ones I mentioned on the JIRA).

Though one thing to address is that when the SQL thread increments `queued_count`, that also needs to be locked (both in `rpl_parallel::do_event()` and `rpl_parallel_entry::queue_master_restart()`).

sql/rpl_parallel.cc
ParadoxV5: > Prior to 13a9395a6acf1bd90370bd6fe7b4637597b921dd, this section was not locked, and `SHOW SLAVE STATUS` would result in a segfault in this point.

Well, it locks `rli->data_lock` now (:.

> 2. When the SQL shuts down

It locks too.

> 3. When the `Master_info` parent object is destroyed (e.g. `RESET SLAVE` or `CHANGE MASTER TO`).

I think it is `purge_relay_logs()` of `rpl_rli.cc`, which also claims the lock.

> Hmm good observation, there may indeed be a small gap of time right after rotating, and before the new binlog's `Format_description_event` is queued where we temporarily appear up-to-date.

Hmm, only `alloc_inuse_relaylog()` changes `last_inuse_relaylog()`, and it has three calls:
1. One during the initialization of the SQL thread, which should occur before this post-event callback.
2. Two in `next_event()` of `slave.cc` which is always called with `rli->data_lock` held.

ParadoxV5 (commented):
> […] another idea which wouldn't incur additional locking (I don't have any more ideas apart from the ones I mentioned on the JIRA).

From JIRA:

> […] any other ideas which don't involve additional locking per-enqueue and per-commit.

While working on the patch, I looked around and found that I can leverage `rli->data_lock` instead of a new mutex for guarding `rpl_parallel::update_workers_idle()`.

I was concerned that `rli->data_lock` could conflict with the `Atomic_counter`s, but I “found” that the pair are already accessed with the lock held, so I dragging the pair into the mix. (https://github.com/MariaDB/server/pull/4103/files#r2151240919)
If mutexes implement `Atomic_counter`, then the patch doesn’t introduce additional locking; it only switches the locks.

Except—

> Though one thing to address is that when the SQL thread increments `queued_count`, that also needs to be locked (both in `rpl_parallel::do_event()` and `rpl_parallel_entry::queue_master_restart()`).

I missed a spot in my search.
`rpl_parallel::do_event()`, which is also the entry point to `rpl_parallel_entry::queue_master_restart()`, does not always hold the lock during the execution, for `sql_delay_event()` of `slave.cc` yoinks this carpet under its feet early in the progress.

Incrementing this counter without a lock may or may not be concerning; e.g., I don’t think a data race is impactful here.
Notably, it wasn’t an `Atomic_counter` before 8dad51481b of MDEV-10653; The thought was,
> By separating the queued and dequeued count, only the dequeued_count needs multi-thread synchronisation; the completed flag and queued_count fields are only accessed by the SQL driver thread and need no synchronisation.
> ⸺ `struct inuse_relaylog` documentation

bnestere:
@knielsen 

> What if the IO thread updated one atomic counter to be the latest timestamp of events from the master, and the worker thread commit updated another atomic to be the latest timestamp of a committed event, then Seconds_behind_master could be simply the difference between these two counters , which can be done without any mutex?

MDEV-33856 has already added that exact functionality in 11.6.

 > I think a step back should be taken to rethink what the real problem here is.

I think an easy summary here is that we have users that have scripts monitoring `Seconds_Behind_Master` with certain thresholds set, and when `SBM` reaches beyond their threshold, it sends off alerts throughout their system saying something is wrong. When there's a bug in `SBM` that artificially spikes it really high, it sets off these alarms. Despite that we already have the correct MDEV-33856 solution in 11.6+, we still need to fix problems in the ill-conceived `Seconds_Behind_Master`. With that said, I agree that adding a mutex as a band-aid for this minor issue seems quite heavy-weight. 

> The whole way that the old-style (non-GTID) position is updated could do with a look-over, that is quite expensive, and not very useful in GTID-only setups. Is this position update relevant to this issue?

Unfortunately, I do think this position update is relevant, as we still have users affected by this `SBM` spike who don't use GTID replication :/  Though generally speaking, I agree it would be good to re-look at how we handle old-style replication. Perhaps the fix could be multi-faceted. One fix for GTID-based configurations to not queue these events, and then for non-GTID setups we could still fix it, but in a less-optimal way (IMO some atomic counter would probably suffice).

ParadoxV5:
> This needs to be avoided. The whole parallel replication architecture goes to great pain to minimize global mutex contention between the replication threads.

This patch changed the atomicity of the counters from `std::atomic`, which might’ve been implemented with a mutex, to a real mutex.
However, `rli->data_lock` is a __global__ mutex and depending on it can be a bottleneck. The replacement should continue to be a unique mutex.
(In this design, a mutex is mandatory around the entire `rpl_parallel::update_workers_idle()` to guard against time-of-check to time-of-*write*)

> Is this position update relevant to this issue? Otherwise I'm not sure why we even queue the Format Description events and such for the parallel worker threads, using `handle_queued_pos_update()`. Could we simply drop queueing these events (position updates) for slave threads?

[Your previous note in MDEV-36840](https://jira.mariadb.org/browse/MDEV-36840?focusedCommentId=305419#comment-305419) explained:
> It would be wrong to update the replication position to point past the Format_description or Rotate event in the SQL driver thread, as this would end up skipping all queued-but-not-executed events in case of a slave restart.
> Therefore the update of the position is scheduled to be done in a worker thread, so that it happens at the correct time as coordinated with the wait_for_prior_commit() between the worker threads.

> Why have this notion of "empty queue" at all? What if the IO thread updated one atomic counter to be the latest timestamp of events from the master, and the worker thread commit updated another atomic to be the latest timestamp of a committed event, then Seconds_behind_master could be simply the difference between these two counters , which can be done without any mutex?

Yes, what he said: [`Master_last_event_time - Slave_last_event_time = Master_Slave_time_diff`](https://mariadb.com/resources/blog/a-new-consistent-way-to-monitor-replication-lag-in-mariadb-community-server-11-8-lts-release/).

ParadoxV5:
> > I think a step back should be taken to rethink what the real problem here is.
> 
> I think an easy summary here is that we have users that have scripts monitoring `Seconds_Behind_Master` with certain thresholds set, and when `SBM` reaches beyond their threshold, it sends off alerts throughout their system saying something is wrong. When there's a bug in `SBM` that artificially spikes it really high, it sets off these alarms. Despite that we already have the correct MDEV-33856 solution in 11.6+, we still need to fix problems in the ill-conceived `Seconds_Behind_Master`.

~~So, the real problem is users not switching to monitor `Master_Slave_time_diff` instead (after upgrading to 11.6+ if not already).~~

> With that said, I agree that adding a mutex as a band-aid for this minor issue seems quite heavy-weight.

What other designs are available? 🤔 Note that *some* weight isn’t going anywhere as long as `dequeued_count` is an `Atomic_counter` (and `queued_count` too since 8dad51481b).

> (IMO some atomic counter would probably suffice)

Chances are that you are referring to your option 2 in JIRA, correct?
> Add an atomic counter to specifically track the number of {{rpl_parallel_thread::queued_event::QUEUED_POS_UPDATE}} that are queued and awaiting dequeue. Not that it is a lot of overhead, but it is overhead for a small use-case.

ParadoxV5:
* [P.S.] Atomics may be implemented with mutexes, or it may not; it can even be the other way around.
  It depends on the architecture we build for, which is likely beyond our concerns.

knielsen:

> alarms. Despite that we already have the correct MDEV-33856 solution
> in 11.6+, we still need to fix problems in the ill-conceived
> `Seconds_Behind_Master`. With that said, I agree that adding a mutex

I disagree, we don't have to fix something that's basically unfixable, since
it was never well defined how it should work in the first place.

On the contrary, any changes done pre-11.6 around Seconds_Behind_Master must
only be done if they can be guaranteed to not cause any further regressions,
including performance regressions.

> Unfortunately, I do think this position update is relevant, as we
> still have users affected by this `SBM` spike who don't use GTID
> replication :/ Though generally speaking, I agree it would be good to

But isn't the problem that the queued Format_description and other events
are causing seconds_behind_master to change? If we don't queue these, then
the problem should go away? And if the old-style position doesn't update for
a Format_description or similar event, it means that a slave that
disconnects and reconnects at that point will need to re-apply the event,
but is that really a problem?

And BTW, claiming that atomics "may be implemented with mutexes" is a
strawman's argument, that is not the case on any platform relevant for high
peformance database operations. Introducing a global mutex, which can cause
thread waits and context switching, to do yet another tweak to
Seconds_Behind_Master, is not reasonable.

 - Kristian.

ParadoxV5:
> I disagree, we don't have to fix something that's basically unfixable, since it was never well defined how it should work in the first place.

So, is the real problem actually the `Seconds_Behind_Master` users?
Then, now that `Master_Slave_time_diff` is around, should we deprecate this `Seconds_Behind_Master` because all it contributes is confusion and technical debt?

We don’t. Hurdles like this are not enough to justify giving up a feature and its remaining value.

Especially that I’m going to revive a scrapped idea where `worker_threads_idle= true` takes place not in `rpl_parallel_thread::inuse_relaylog_refcount_update()` but on demand.
Considering now, this would only depend on the existing `rli->data_lock` locking and avoid mutex locking in there.

bnestere:
> On the contrary, any changes done pre-11.6 around Seconds_Behind_Master must
> only be done if they can be guaranteed to not cause any further regressions,
> including performance regressions.

Agreed.

> But isn't the problem that the queued Format_description and other events
> are causing seconds_behind_master to change? If we don't queue these, then
> the problem should go away? And if the old-style position doesn't update for
> a Format_description or similar event, it means that a slave that
> disconnects and reconnects at that point will need to re-apply the event,
> but is that really a problem?

Would that be safe for 10.6? I agree in the general case, the user probably wouldn't notice anything. Though the problem I'd envision is that we would start ignoring log rotations, where previously we hadn't been. One edge case (albeit unlikely)  would be if a user configured with old-style coordinates had already purged the rotated-on-master binlog but not rotated-on-slave. Would we be able to know if the user could just manually switch to the new binlog? _I mean, probably, as MariaDB would still track the GTID state and a user could manually compare the `Gtid_list_log_event` of the new master binlog vs the binlog state of the slave.. but it would still lead to an awkward case for users to find that out (albeit it is a very unlikely case)_ .

And another thought is that the slave wouldn't rotate at the same place as the master.  IMO this actually makes sense, I don't actually understand the reason a slave rotates with its master in the first place - its binlog is its own. But again, do we want to change that in a GA release? 


> Then, now that Master_Slave_time_diff is around, should we deprecate this Seconds_Behind_Master because all it contributes is confusion and technical debt?

I do think we should deprecate `Seconds_Behind_Master`. The alternative (MDEV-33856) is present in the latest 11.8 LTS already, so I would think we could deprecate it by the next planned LTS (I think 12.4?), and potentially remove it by the LTS after that (13.4?)

sql/rpl_parallel.cc
```diff
@@ -3050,13 +3061,13 @@ rpl_parallel::stop_during_until()
...
-bool
-rpl_parallel::workers_idle(Relay_log_info *rli)
+bool Relay_log_info::are_worker_threads_caught_up()
 {
-  mysql_mutex_assert_owner(&rli->data_lock);
-  return !rli->last_inuse_relaylog ||
-    rli->last_inuse_relaylog->queued_count ==
-    rli->last_inuse_relaylog->dequeued_count;
+  mysql_mutex_assert_owner(&data_lock);
+  if (!worker_threads_caught_up && (!last_inuse_relaylog ||
```
ParadoxV5: Re. https://github.com/MariaDB/server/pull/4103#discussion_r2151239208, I still wonder if we can skip this null check at any of the two places.

ParadoxV5 (commented):
Here’s the alternative I spoke of, @bnestere and @knielsen.
I moved `worker_threads_caught_up= true` from eagerly by the worker threads to lazily when using the state.
The reads already hold `rli->data_lock`, and I avoid adding a new mutex acquisition in the worker threads this way.

I also moved `update_workers_idle()` inside the getter `are_worker_threads_caught_up()` to enforce this method of access.

knielsen (approved):
Looks good ParadoxV5, good call finding a way to do this within how the existing code is holding mutexes. This looks to have much lower impact (if any) on parallel worker thread performance, so I'm fine with this.

I did not try to understand why/if this patch solves the problem with Seconds_behind_master, nor look at the test case, but I'm fine with the patch like this. So not appropriate with "Reviewed-by:", but _is_ appropriate with "Acked-by".

mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike.test
```diff
@@ -46,33 +42,35 @@ SET @@global.debug_dbug="+d,negate_clock_diff_with_master";
...
-{
-    die "A relay log event was incorrectly used to set Seconds_Behind_Master";
-}
-
---echo # Safely resume slave SQL thread
+# `Seconds_Behind_Master` must remains 0.
+# * MDEV-16091:
+#   It should not point at the FDE's `start_slave.inc` timestamp.
+# * MDEV-25999 & MDEV-36840:
+#   It should not point at all in a new stream, for FDEs are not content.
```
bnestere: I don't know what this last line means. Maybe add some specifics? _And generally I'd suggest avoiding the word "it" in technical writing (which code comments are), it is often ambiguous/unclear as to what "it" refers to :innocent:._

mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike.combinations
```diff
@@ -0,0 +1,5 @@
+[serial]
+--slave-parallel-threads=0
+
+[parallel]
+--slave-parallel-threads=1
```
bnestere, line 5: Good realization that we basically already tested for this regression just with a different configuration, and nice idea to re-use the test with just a `.combinations` file :sunglasses:

sql/rpl_parallel.cc
bnestere: This condition is starting to get a bit complex (and not easily self-documenting anymore) - it'd be good to add a comment before it now explaining what is being checked.

sql/rpl_rli.h
```diff
@@ -588,6 +607,13 @@ class Relay_log_info : public Slave_reporting_capability
     relay log.
   */
   uint32 m_flags;
+
+  /**
```
bnestere, line 615: Can you also add a description to this comment on the purpose of the variable? i.e. what it is actually tracking, how it is used, and why it is updated lazily.

sql/slave.cc
```diff
@@ -4448,13 +4448,12 @@ static int exec_relay_log_event(THD* thd, Relay_log_info* rli,
...
-        The check for queued_count == dequeued_count ensures the worker threads
-        are all idle (i.e. all events have been executed).
+        The check for `rli->are_worker_threads_caught_up()` ensures the worker
+        threads are practically idle (i.e. all user events have been executed).
       */
       if ((unlikely(rli->last_master_timestamp == 0) ||
            (rli->sql_thread_caught_up &&
-            (rli->last_inuse_relaylog->queued_count ==
-             rli->last_inuse_relaylog->dequeued_count))) &&
+            rli->are_worker_threads_caught_up())) &&
```
bnestere: It might be better to keep this as it was. I don't think the value of `worker_threads_caught_up` matters here; and for busy servers (i.e. rarely idle), this adds an extra function call and conditional check for each event (only as long as the SQL thread is caught up too, I suppose). _Just for your consideration, I'd doubt it would be really noticeable_

bnestere (approved):
Hi @ParadoxV5 !

To echo Kristian, nice work! This is quite a creative solution :) I've left a few minor comments.

Also, when you write the git commit message, please explain the life-cycle of `worker_threads_caught_up` and why it is so (i.e. how & why this fixes the `SBM` spikes). 

Thanks too for the cleanup :+1:

sql/slave.cc
ParadoxV5: > It might be better to keep this as it was. I don't think the value of `worker_threads_caught_up` matters here;

Let's see…
When the first user event of a stream comes after all SQL threads (main and workers) are “caught up”, this code should set the `last_master_timestamp` to the event’s timestamp.
But, because `worker_threads_caught_up` returning to `true` is now lazy, if it does not refresh to reflect that workers are in fact “caught up”, this section will never run again once it is set `false`.
That means, during the small window in `rpl_parallel::do_event()` between it unlocks `rli->data_lock` (as [you warned](https://github.com/MariaDB/server/pull/4103#pullrequestreview-2947004825)) and the event enqueues, if we refresh `worker_threads_caught_up= true` with SHOW REPLICA STATUS, it looks like we would miss `sql_thread_caught_up= false` as if the event is internal, and together **they hide `Seconds_Behind_Master` as 0** until at least the next user event.
That isn’t right.
On a related note, if the SHOW REPLICA STATUS is between the event enqueues and a worker completes it, `Seconds_Behind_Master` would instead continue to reflect the time since the last event *from the previous stream* completed.

sql/rpl_rli.h
ParadoxV5: Obsolete: The new solution has reverted the reasons that are new to this change because [they were a performance toll](https://github.com/MariaDB/server/pull/4103#issuecomment-2998129259).

sql/rpl_parallel.cc
ParadoxV5: Indeed…

For the record, once this section was:
```c++
if (worker_threads_caught_up)
  return true;
if (!last_inuse_relaylog ||
    last_inuse_relaylog->queued_count == last_inuse_relaylog->dequeued_count)
  return worker_threads_caught_up= true; // Refresh
return false;
```

ParadoxV5 (commented):
💜

sql/slave.cc
bnestere: > and together they hide Seconds_Behind_Master as 0 until at least the next user event.
> That isn’t right.

Hmm indeed - perhaps we should include the check for `sql_thread_caught_up` in `are_worker_threads_caught_up` (and perhaps rename the function) to make it so we can't set `worker_threads_caught_up= true` unless *both* 1) the sql thread is idle, and 2) there is nothing queued; then we can't overwrite the variable in this in-between.

bnestere: And for 

> On a related note, if the SHOW REPLICA STATUS is between the event enqueues and a worker completes it, Seconds_Behind_Master would instead continue to reflect the time since the last event from the previous stream completed.

If I understand you correctly, this is expected and documented (implemented in this commit: 75dc2671011ba53e4f4531752c213ced7f9012ff). _With the notable exception being the first transaction after idling, where we update last_master_timestamp right away to avoid an arbitrarily large gap where a replica could be idle between events_

ParadoxV5: > perhaps we should include the check for `sql_thread_caught_up` in `are_worker_threads_caught_up` (and perhaps rename the function)

I thought about that…
My concern was that the new encapsulation design prevents setting `worker_threads_caught_up= true` independently of `sql_thread_caught_up`.
But, you highlighted from my analysis that it’s not right to do so, not to mention that I annotated `@pre Check @ref sql_thread_caught_up first, …` on `are_worker_threads_caught_up()`.

The next thought is what I would name the new version as… Because it doesn’t need to check `mi->using_parallel()` since it’s done at a higher frame:
https://github.com/MariaDB/server/blob/c8cf8d352899a264203b4f49085b013b29a4a59f/sql/slave.cc#L4441-L4444
How about simply `are_sql_threads_caught_up()`, and note the parallel assumption as a `@pre`requisite?

mysql-test/suite/rpl/t/rpl_seconds_behind_master_spike.test
```diff
@@ -46,33 +42,38 @@ SET @@global.debug_dbug="+d,negate_clock_diff_with_master";
 --echo # Write events to ensure slave will be consistent with master
 create table t1 (a int);
 insert into t1 values (1);
---let $t_master_events_logged= `SELECT UNIX_TIMESTAMP()`
+--sync_slave_with_master
+let $wait_condition= SELECT STATE LIKE '%waiting for more updates'
+  FROM information_schema.PROCESSLIST WHERE COMMAND = 'Slave_SQL';
+--source include/wait_condition.inc
```
ParadoxV5, line 48: For your awareness, @bnestere, this is a noteworthy addition since your review.

I queued the PR for merging yesterday, but the modified test sporadically [failed](https://buildbot.mariadb.org/#/builders/369/builds/32039) *in serial mode*.
I suspect it’s the same “up-to-date” assumption I had that led to #4155.

PR #4108, opened 2025-06-12
GSOC'25 - MDEV-12513 Implementation of Native INTERVAL Data Type

HazemRawi:

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check


mysql-test/main/type_interval.test
```diff
@@ -55,6 +55,33 @@ FROM t1;
 
 DROP TABLE t1;
 
+--echo #
+--echo # MDEV-12513: Support for INTERVAL data types in MariaDB
+--echo #
+CREATE TABLE t2 (a INTERVAL YEAR );
```
DaveGosselin-MariaDB, line 67: Please add `SHOW CREATE TABLE` for each of these as we discussed.  For example:
```
SHOW CREATE TABLE t2;
```

mysql-test/main/type_interval.test
```diff
@@ -55,6 +55,33 @@ FROM t1;
...
+CREATE TABLE t6 (a INTERVAL HOUR(3) );
+CREATE TABLE t7 (a INTERVAL MINUTE(2) );
+CREATE TABLE t9 (a INTERVAL DAY TO HOUR );
+CREATE TABLE t10 (a INTERVAL DAY TO MINUTE );
+CREATE TABLE t11 (a INTERVAL DAY(4) TO SECOND(3) );
+CREATE TABLE t12 (a INTERVAL HOUR TO MINUTE );
+CREATE TABLE t13 (a INTERVAL HOUR(3) TO SECOND(2) );
+CREATE TABLE t14 (a INTERVAL MINUTE TO SECOND(4) );
+
+DROP TABLE t2;
```
DaveGosselin-MariaDB: Consolidate these to one statement, like this (but with all your new tables):
```
DROP TABLE t2, t3, t4;
```

sql/sql_yacc.yy
```diff
@@ -6624,8 +6625,28 @@ field_type_temporal:
           {
             $$.set(thd->type_handler_for_datetime(), $2);
           }
+        | INTERVAL_SYM interval_qualifier
+        {
+            $$.set(&type_handler_interval_DDhhmmssff);
+
```
DaveGosselin-MariaDB: Remove unneeded blank line.

DaveGosselin-MariaDB (commented):
Hi @HazemRawi , thanks for the PR.  Here are my first impressions, please update the PR with changes.  Thanks!

HazemRawi:
Thanks you! updated with the requested changes ;

sql/sql_yacc.yy
```diff
@@ -15708,6 +15747,13 @@ temporal_literal:
                                                             YYCSCL, true))))
               MYSQL_YYABORT;
           }
+        | INTERVAL_SYM %prec EMPTY_FROM_CLAUSE TEXT_STRING interval_qualifier
+                    {
+                      if (unlikely(!($$= type_handler_datetime.create_literal_item(thd,
```
HazemRawi: Delegating interval handling to the `type_handler_datetime.create_literal_item`  is the reason why some some tests fail -- This will be resolved in the next commit by implementing `type_handler_interval.create_literal_item`

sql/field.cc
```diff
@@ -88,20 +89,24 @@ bool Field::marked_for_write_or_computed() const
   following #defines describe that gap and how to calculate number of fields
   and index of field in this array.
 */
-const int FIELDTYPE_TEAR_FROM= (MYSQL_TYPE_BIT + 1);
-const int FIELDTYPE_TEAR_TO=   (MYSQL_TYPE_NEWDECIMAL - 1);
-const int FIELDTYPE_LAST=      254;
-const int FIELDTYPE_NUM=       FIELDTYPE_TEAR_FROM + (FIELDTYPE_LAST -
-                                                      FIELDTYPE_TEAR_TO);
+const int FIELDTYPE_TEAR_FROM = (MYSQL_TYPE_INTERVAL + 1);
```
DaveGosselin-MariaDB, line 92: The `varname= ` syntax in the original is our standard way of writing assignment statements, please revert these changes.

sql/field.cc
```diff
@@ -88,20 +89,24 @@ bool Field::marked_for_write_or_computed() const
...
+const int FIELDTYPE_TEAR_TO =   (MYSQL_TYPE_NEWDECIMAL - 1);
+const int FIELDTYPE_LAST =      254;
+const int FIELDTYPE_NUM =       (FIELDTYPE_TEAR_FROM) + (FIELDTYPE_LAST - FIELDTYPE_TEAR_TO - 3);
 
 static inline int merge_type2index(enum_field_types merge_type)
 {
   DBUG_ASSERT(merge_type < FIELDTYPE_TEAR_FROM ||
               merge_type > FIELDTYPE_TEAR_TO);
   DBUG_ASSERT(merge_type <= FIELDTYPE_LAST);
+  if (merge_type > MYSQL_TYPE_BIT && merge_type < MYSQL_TYPE_INTERVAL)
```
DaveGosselin-MariaDB: Please explain the intent behind this change.

mysql-test/main/type_interval.result
```diff
@@ -81,172 +81,6 @@ DROP TABLE t1;
 #
 # MDEV-12513: Support for INTERVAL data types in MariaDB
 #
-CREATE TABLE t2 (a INTERVAL YEAR );
```
DaveGosselin-MariaDB: Why did you remove the contents of this file and `if(0)` the corresponding `.test` file?

sql/field.cc
```diff
@@ -5479,7 +5516,29 @@ int Field_timestamp::store(const char *from,size_t len,CHARSET_INFO *cs)
...
+{
+    THD *thd = get_thd();
+    Interval interval(from, len, m_interval_type, true);
+    my_timeval tm;
+    interval_to_timeval(&interval, &tm, dec, thd);
+    store_TIMEVAL(tm);
+    return 0;
+}
+
+String *Field_interval::val_str(String *val_buffer, String *) {
```
DaveGosselin-MariaDB: Opening brace goes on new line.

sql/field.cc
```diff
@@ -5479,7 +5516,29 @@ int Field_timestamp::store(const char *from,size_t len,CHARSET_INFO *cs)
...
+  my_timeval tm;
+  my_timestamp_from_binary(&tm, ptr, decimals());
+
+Interval* iv = new Interval();
```
DaveGosselin-MariaDB: Indent two spaces here.

sql/field.cc
```diff
@@ -5479,7 +5516,29 @@ int Field_timestamp::store(const char *from,size_t len,CHARSET_INFO *cs)
...
+  timeval_to_interval(tm, iv);
 
+  char buf[64];
```
DaveGosselin-MariaDB: Why hard-coded `64`, and is it sufficient in the general case?

storage/mroonga/ha_mroonga.cpp
```diff
@@ -1542,6 +1542,7 @@ static grn_builtin_type mrn_grn_type_from_field(grn_ctx *ctx, Field *field,
     break;
   case MYSQL_TYPE_VARCHAR_COMPRESSED:
   case MYSQL_TYPE_BLOB_COMPRESSED:
+  case MYSQL_TYPE_INTERVAL:
```
DaveGosselin-MariaDB, line 1545: Why is this change, or any storage engine specific change, required?  At this point, we were supposed to be focusing on stored procedures.

DaveGosselin-MariaDB (commented):
Where are the tests demonstrating stored procedures?

mysql-test/main/type_interval.result
HazemRawi: > Why did you remove the contents of this file and `if(0)` the corresponding `.test` file?

That was the latest working commit that pass the mtr 
the next commit after it introduced ambiguous results (resolved now)

storage/mroonga/ha_mroonga.cpp
HazemRawi: > Why is this change, or any storage engine specific change, required? At this point, we were supposed to be focusing on stored procedures.

To avoid build errors

sql/field.cc
HazemRawi: 4th comment:
`merge_type2index` function modified because inserting `MYSQL_TYPE_INTERVAL` after `MYSQL_TYPE_TIME2` shifted values of all subsequent enums
This shift broke `merge_type2index` so to keep the MTR working I updated merge_type2index.

I first tried placing `MYSQL_TYPE_INTERVAL` before `MYSQL_TYPE_BIT` but this caused problems because the positions of `MYSQL_TYPE_TIMESTAMP2`, `MYSQL_TYPE_DATETIME2`, `MYSQL_TYPE_TIME2` are fixed and cannot change because they have fixed values(I didn't investigate in this matter too much), This value  prevented placing `MYSQL_TYPE_INTERVAL` before `MYSQL_TYPE_BIT`, because doing so would break MySQL 5.6 to MariaDB 5.3 replication tests.

sql/sql_yacc.yy
```diff
@@ -6624,7 +6671,37 @@ field_type_temporal:
           {
             $$.set(thd->type_handler_for_datetime(), $2);
           }
-        ;
+                 | INTERVAL_SYM interval_qualifier
+                 {
+                     uint8 frac_prec = $2.end_prec.length();
```
HazemRawi: 1st comment:
I want to pass `INTERVAL_TYPE`, `Length` and `decimal` to the type_handler_interval 
so instead of introducing `interval_type` within `Lex_length_and_dec_st`  ,SO I stored the interval_type  in the top 5 bits of the existing `m_dec` field
2nd comment: will explain how decimal & interval_type extracted

sql/sql_type.cc
```diff
@@ -8214,6 +8254,22 @@ Item *Type_handler_interval_DDhhmmssff::
...
                                                                attr.decimals());
 }
+bool Type_handler_interval_DDhhmmssff::
+ Column_definition_set_attributes(THD *thd,
+                                             Column_definition *def,
+                                             const Lex_field_type_st &attr,
+                                             column_definition_type_t type)
+                                             const
+{
+  Lex_length_and_dec_st alter_attr;
```
HazemRawi: 2nd comment:

The 5 most significant bits store the interval type which extracted assigned it to pack_flag.
The 3 LSBs hold a decimal value that is less than or equal to six.

The Earlier ambiguous results occurred because I stored the interval type in pack_flag then Column_definition_prepare_stage2_legacyoverwrote it with:
 def->pack_flag = f_settype((uint)type);
Now I can get correct result
3rd comment: I will explain how interval_type is stored within pack_flag

sql/field.h
```diff
@@ -6068,6 +6140,8 @@ bool check_expression(Virtual_column_info *vcol, const Lex_ident_column &name,
 #define f_bit_as_char(x)        ((x) & FIELDFLAG_TREAT_BIT_AS_CHAR)
 #define f_is_hex_escape(x)      ((x) & FIELDFLAG_HEX_ESCAPE)
 #define f_visibility(x)         (static_cast<field_visibility_t> ((x) & INVISIBLE_MAX_BITS))
+#define f_set_interval_type(x)  ((((x) & 0x1U) << 1) | (((x) & 0xEU) << 10))
```
HazemRawi: 3rd comment:
I stored the interval type in `pack_flag` using bits 1,11,12, and 13. because:

Bit 1: Always set to 1 for MYSQL_TYPE_INTERVAL (using ternary logic).
Bit 11: is used to store decimal values which in this case is redundant as the max decimal value is 6.
Bits 12-13: Store FIELDFLAG_TREAT_BIT_AS_CHAR and FIELDFLAG_LONG_DECIMAL flags, They are redundant for intervals.

I went on this direction to avoid expanding pack_flag, Please review this bit layout and share your thoughts.

HazemRawi:
> Where are the tests demonstrating stored procedures?

SP tests aren't included in this commit because they cause MTR to fail. I want to maintain passing MTR status for every commit, I'll add these tests once they're working and giving correct results (before our upcoming meeting).

sql/field.cc
HazemRawi: All comments related to Coding Standards will be resolved on the upcoming commit.

storage/mroonga/ha_mroonga.cpp
DaveGosselin-MariaDB: I see, when I revert the change I get this:
```
ha_mroonga.cpp:1383:10: warning: enumeration value ‘MYSQL_TYPE_INTERVAL’ not handled in switch [-Wswitch]
```

sql/field.cc
HazemRawi: For testing purposes only, I resolved it later by calculating its value depending on the MAX_INTERVAL_WIDTH for every interval type.

sql/field.cc
```diff
@@ -10884,6 +10945,19 @@ bool Column_definition::fix_attributes_temporal_with_time(uint int_part_length)
   return false;
 }
 
+bool Column_definition::fix_attributes_interval(interval_type itype)
+{
+  if (decimals > MAX_DATETIME_PRECISION)
+  {
+    my_error(ER_TOO_BIG_PRECISION, MYF(0), field_name.str,
```
DaveGosselin-MariaDB, line 11189: Please add a test case that exercises this case.

sql/field.cc
```diff
@@ -10884,6 +10945,19 @@ bool Column_definition::fix_attributes_temporal_with_time(uint int_part_length)
...
+             MAX_DATETIME_PRECISION);
+    return true;
+  }
+
+  //length= calc_interval_display_width(itype, length, decimals);
```
DaveGosselin-MariaDB: Is this still needed?

sql/sql_interval.cc
```diff
@@ -0,0 +1,523 @@
...
+  minute= second= second_part= 0;
+  neg= false;
+}
+
+
+Interval::Interval(const char *str, size_t length,
+                   enum interval_type itype, bool allow_fractional)
+{
+  if (str_to_interval(str, length, this, itype, allow_fractional) != 0)
+    throw std::invalid_argument("Invalid interval format");
```
DaveGosselin-MariaDB: We don't support exceptions in the `server` code.  To emit an error, call `my_error`.  You will find examples in `sql_select.cc`.

sql/sql_interval.cc
```diff
@@ -0,0 +1,523 @@
...
+  while (*str < end && my_isdigit(&my_charset_latin1, *(*str)))
+  {
+    *dest= (*dest) * 10 + (*(*str) - '0');
+    (*str)++;
+  }
+
+  return (*str == start);
+}
+
+int str_to_interval(const char *str, size_t length, Interval *to,
```
DaveGosselin-MariaDB, line 213: This is called by `Field_interval::store`, but this doesn't account for the character set.  Do we need to do that here?

sql/sql_interval.cc
```diff
@@ -0,0 +1,523 @@
...
+                    enum interval_type itype, bool allow_fractional)
+{
+  const char *p= str;
+  const char *end= str + length;
+  ulong values[6]= {0};
```
DaveGosselin-MariaDB: Please comment magic numbers.

sql/sql_interval.cc
```diff
@@ -0,0 +1,523 @@
+/* Copyright (C) 2025 Google Summer of Code 2025
```
DaveGosselin-MariaDB: This can't be the copyright.  For any contribution we need MCA or new BSD.

sql/sql_type.cc
```diff
@@ -401,6 +401,38 @@ my_decimal *Temporal::bad_to_decimal(my_decimal *to) const
   return NULL;
 }
 
+static size_t interval_default_length(enum interval_type type) {
+  switch(type) {
+  case INTERVAL_YEAR:
+  case INTERVAL_YEAR_MONTH:
+    return 4;
```
DaveGosselin-MariaDB: Comment magic numbers please.

sql/sql_interval.h
```diff
@@ -0,0 +1,70 @@
+/* Copyright (C) 2025 Google Summer of Code 2025
```
DaveGosselin-MariaDB: This can't be the copyright. For any contribution we need MCA or new BSD.

DaveGosselin-MariaDB (commented):
Comment magic numbers throughout.

mysql-test/main/type_interval.test
```diff
@@ -84,6 +84,128 @@ SHOW CREATE TABLE t14;
...
+
+  SELECT CASE
+    WHEN x < INTERVAL  '2025-06-19 15:00:00' HOUR THEN 'A'
+    WHEN x BETWEEN INTERVAL  '2025-06-19 15:00:00' HOUR AND INTERVAL  '2025-06-19 15:00:00' HOUR THEN 'B'
+    ELSE 'C'
+  END AS duration_category;
+END;
+/
+DELIMITER ;/
 --echo #
```
DaveGosselin-MariaDB: Your tests are not part of 10.4, but 12.1.  Please add your tests to their own 12.1 block.

DaveGosselin-MariaDB:
Please fix this error uncovered by the Windows builder:
```
Building Custom Rule C:/Buildbot/amd64-windows/build/sql/CMakeLists.txt
         [BISON][gen_mariadb_cc_hh] Building parser with bison 2.7
         C:/Buildbot/amd64-windows/build/sql/yy_mariadb.yy:1359.1-11: error: invalid directive: '%precedence'
```

sql/sql_type.cc
HazemRawi: TEST

sql/field.cc
HazemRawi: Okay, I'll include full set of tests after implementing validation across the server.

sql/field.cc
HazemRawi: No,I left it to tell you that I wanted to store the interval display width in `length`, but later I preferred to use `field_length` to store interval width, What are your thoughts on this?

sql/sql_interval.cc
HazemRawi: Understood, it was for testing purposes only.

sql/sql_interval.cc
HazemRawi: I left some details intentionally for simpler development phase, later I was planning to implement charset info, ErrConvString, validation on the parser side and within Field_interval, However, Here we are I must implement them now.

sql/sql_interval.cc
HazemRawi: Understood

mysql-test/main/type_interval.result
```diff
@@ -81,3 +81,254 @@ DROP TABLE t1;
 #
 # End of 10.4 tests
 #
+#
```
HazemRawi, line 84: Please review these tests to validate supported interval precision & values

mysql-test/main/type_interval.result
```diff
@@ -79,87 +79,256 @@ Warning	1292	Incorrect interval value: '876494160000.000000000' for column `test
...
+CALL sp_valid_9('999');
+a
+999
+CALL sp_valid_10('9999999');
+a
+9999999
+CALL sp_valid_11('99 23:59:59');
+a
+99 23:59:59
+CALL sp_valid_12('999:59:999999');
```
DaveGosselin-MariaDB: As discussed on Zulip and in our call, the seconds precision is for sub-second resolution:
```
SECONDS(3)
```
means
```
00:00:00.123
```

mysql-test/main/type_interval.result
```diff
@@ -79,87 +79,256 @@ Warning	1292	Incorrect interval value: '876494160000.000000000' for column `test
...
-@out_hs,
-@out_hm,
-@out_ms;
-@out_ym	@out_y	@out_m	@out_ds	@out_d	@out_h	@out_n	@out_s	@out_dh	@out_dm	@out_hs	@out_hm	@out_ms
-5-03	10	12	100 12:30:45.123456	100	120	90	45	100 12	100 12:30	120:30:45.123456	120:30	90:45.123456
 #
-# End of 10.4 tests
+# MDEV-12513: Support for INTERVAL data types in MariaDB
 #
+Parser Validation - Unsupported Interval Types & Invalid combination of INTERVAL Range.
```
DaveGosselin-MariaDB, line 167: When you add new tests, don't remove the old ones.  What happened to the pass-thorough `CREATE PROCEDURE` tests?  Why did you delete them?

DaveGosselin-MariaDB (changes requested):
Comments on precision for seconds.

mysql-test/main/type_interval.result
HazemRawi: Understood, it's clear now

mysql-test/main/type_interval.result
HazemRawi: You are right, I should have included them. however it is not a common practice to delete existing tests whenever developing a new feature because development must be incremental.

DaveGosselin-MariaDB: It is common practice to incrementally add tests as you develop your feature.  If the behavior of the code has changed and the tests are no longer needed, then it is fine to remove them.  But I don't think that is the case here.

include/mysql_com.h
```diff
@@ -501,6 +501,7 @@ enum enum_field_types { MYSQL_TYPE_DECIMAL, MYSQL_TYPE_TINY,
                         MYSQL_TYPE_TIMESTAMP2,
                         MYSQL_TYPE_DATETIME2,
                         MYSQL_TYPE_TIME2,
+                        MYSQL_TYPE_INTERVAL,
```
DaveGosselin-MariaDB, line 504: Why isn't this at the end of the enumeration?  In the case that it cannot be at the end:  there's a comment just a few lines up indicating that these are for mysql-5.6 compatibility.  Please add an empty line break before your new enum entry, it will indicate that your new type is not for compatibility.

sql/compat56.cc
```diff
@@ -473,4 +473,85 @@ void my_timestamp_to_binary(const struct my_timeval *tm, uchar *ptr, uint dec)
...
+  tm->tv_sec= mi_uint5korr(ptr);
+  switch (dec)
+  {
+  case 0:
+  default:
+    tm->tv_usec= 0;
+    break;
+  case 1:
+  case 2:
+    tm->tv_usec= ((uint) ptr[5]) * 10000;
```
DaveGosselin-MariaDB, line 494: Please make `5` into a constant at the start of this method.  By giving it a useful name, you also document its meaning.  You can declare it as `constexpr const ...` (where `...` is the type, identifier, etc) on the first line of the body of this function.  As a `constexpr`, the compiler will try to optimize its use during compilation and it will retain its type.

sql/compat56.cc
```diff
@@ -473,4 +473,85 @@ void my_timestamp_to_binary(const struct my_timeval *tm, uchar *ptr, uint dec)
...
+    break;
+  case 3:
+  case 4:
+    tm->tv_usec= (uint) mi_uint2korr(ptr + 5) * 100;
+    break;
+  case 5:
+  case 6:
+    tm->tv_usec= (uint) mi_uint3korr(ptr + 5);
+  }
+if (tm->tv_sec & INTERVAL_SIGN_BIT)
```
DaveGosselin-MariaDB, line 504: Code formatting is off, please indent by two spaces.  Also, put a blank line between this `if` and the end of the `switch` statement.

sql/compat56.cc
```diff
@@ -473,4 +473,85 @@ void my_timestamp_to_binary(const struct my_timeval *tm, uchar *ptr, uint dec)
...
+  case 0:
+  default:
+    break;
+  case 1:
+  case 2:
+    ptr[5]= (unsigned char) (char) (microseconds / 10000);
+    break;
+  case 3:
+  case 4:
+    mi_int2store(ptr + 5, microseconds / 100);
```
DaveGosselin-MariaDB, line 545: (Same comment regarding `5` here as above)

sql/compat56.cc
```diff
@@ -473,4 +473,85 @@ void my_timestamp_to_binary(const struct my_timeval *tm, uchar *ptr, uint dec)
...
+  }
+  mi_int5store(ptr, seconds);
+  switch (dec)
+  {
+  case 0:
+  default:
+    break;
+  case 1:
+  case 2:
+    ptr[5]= (unsigned char) (char) (microseconds / 10000);
```
DaveGosselin-MariaDB, line 541: Why are two casts necessary?  In other places, you use `static_cast` but why not here?

sql/compat56.cc
```diff
@@ -473,4 +473,85 @@ void my_timestamp_to_binary(const struct my_timeval *tm, uchar *ptr, uint dec)
...
+/**
+  Convert MariaDBxx in-memory interval representation to on-disk representation.
+
+  @param        tm   The value to convert.
+  @param  OUT   ptr  The pointer to store the value to.
+  @param        dec  Precision.
+*/
+void my_interval_to_binary(const struct my_timeval *tm, uchar *ptr, uint dec)
+{
+  long long seconds= tm->tv_sec, microseconds= (long long)tm->tv_usec;
```
DaveGosselin-MariaDB, line 524: `static_cast` ?

sql/compat56.cc
```diff
@@ -473,4 +473,85 @@ void my_timestamp_to_binary(const struct my_timeval *tm, uchar *ptr, uint dec)
...
+    break;
+  case 5:
+  case 6:
+    mi_int3store(ptr + 5, microseconds);
+  }
+}
+
+uint my_interval_binary_length(uint dec)
+{
+  return 5 + (dec + 1) / 2;
```
DaveGosselin-MariaDB, line 555: Please add a comment explaining the rationale for this code.

sql/compat56.h
```diff
@@ -32,6 +32,7 @@
 #define MY_PACKED_TIME_GET_FRAC_PART(x)    ((x) % (1LL << 24))
 #define MY_PACKED_TIME_MAKE(i, f)          ((((ulonglong) (i)) << 24) + (f))
 #define MY_PACKED_TIME_MAKE_INT(i)         ((((ulonglong) (i)) << 24))
+#define INTERVAL_SIGN_BIT (1LL << 39)      // 1 For positive & 0 For negative
```
DaveGosselin-MariaDB, line 35: For consistency, please name this `MY_INTERVAL_SIGN_BIT`.

sql/field.cc
DaveGosselin-MariaDB: Please fix this.

sql/field.cc
DaveGosselin-MariaDB: Why can't we place `MYSQL_TYPE_INTERVAL` at the end of the enumeration in which it is declared?  How will that impact the change here in `merge_type2index`?

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
+/* Copyright (C) 2025
+
+This program is free software; you can redistribute it and/or modify
```
DaveGosselin-MariaDB, line 3: Indent is incorrect here.

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
...
+  : INTERVAL(other),
+    m_interval_type(other.m_interval_type),
+    start_prec(other.start_prec),
+    end_prec(other.end_prec)
+{}
+
+
+Interval& Interval::operator=(const Interval &other)
+{
+  if (this != &other)
```
DaveGosselin-MariaDB, line 37: It's easier to read code when control flow is as de-dented as possible.  This is known as keeping the happy path to the left.  In this case, I would invert the conditional so that we return early:
```
if (this == &other)
  return *this;

static_cast<INTERVAL&>(*this)= other;
m_interval_type= other.m_interval_type;
start_prec= other.start_prec;
end_prec= other.end_prec;
```

Second comment:  Is this not going to result in infinite recursion?
```
static_cast<INTERVAL&>(*this)= other;
```

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
...
+  return *this;
+}
+
+
+Interval::Interval()
+  : m_interval_type(INTERVAL_LAST),
+    start_prec(0),
+    end_prec(0)
+{
+  year= month= day= hour= 0;
```
DaveGosselin-MariaDB, line 54: Why are these not initialized in the initializer list along with `start_prec` and `end_prec` ?

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
...
+}
+
+
+Interval::Interval(const char *str, size_t length,
+                  enum interval_type itype, CHARSET_INFO *cs, uint8 start_prec, uint8 end_prec)
+    : m_interval_type(itype),
+      start_prec(start_prec),
+      end_prec(end_prec)
+{
+  TemporalAsciiBuffer tmp(str, length, cs);
```
DaveGosselin-MariaDB, line 66: What about initializing other data members of the class?

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
...
+  str_to_interval(tmp.str, tmp.length, this, itype, start_prec, end_prec);
+  if (!is_valid_interval(itype, start_prec, end_prec, this))
+    my_error(ER_INVALID_DEFAULT_PARAM, MYF(0));
+}
+
+Interval::Interval(const Sec6 &sec6, enum interval_type itype, uint8 start_prec, uint8 end_prec)
+{
+  Sec6_to_interval(sec6, this, itype, start_prec, end_prec);
```
DaveGosselin-MariaDB, line 74: (Same comment for this constructor and the remaining `Interval` constructors)

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
...
+         uint8 end_prec)
+    : Interval(Sec6(d), itype, start_prec, end_prec) {}
+Interval::Interval(THD *thd, Item *item) {  }
+Interval::Interval(Interval_native val) {  }
+Interval::Interval(Native *val) {  }
+
+int Interval::cmp(const Interval &other) const
+{
+  my_timeval tm1= this->to_TIMEVAL(), tm2= other.to_TIMEVAL();
+  return tm1.tv_sec < tm2.tv_sec ? -1 : tm1.tv_sec > tm2.tv_sec ? 1 : tm1.tv_usec < tm2.tv_usec ? -1 : tm1.tv_usec > tm2.tv_usec ? 1 : 0;
```
DaveGosselin-MariaDB, line 103: Please clarify the formatting of this.  Alternatively, use plain `if` statements if that would be clearer.

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
...
+  return 0;
+}
+
+int Sec6_to_interval(const Sec6 &sec6, Interval *to,
+                    enum interval_type itype, uint8 start_prec, uint8 end_prec) {
+    to->year = to->month = to->day = to->hour = to->minute = to->second = to->second_part = 0;
+    to->neg = sec6.neg();
+    ulonglong value = sec6.sec();
+
+    switch (itype) {
```
DaveGosselin-MariaDB, line 387: The indent and formatting throughout this `switch` statement is inconsistent.  I think you probably have tabs enabled in your editor.  Please make sure that your tab key inserts spaces, preferably two spaces.  Don't change your editor to automatically fix all tabs as spaces because we don't want to update lines of code that you didn't add or modify.

sql/sql_interval.cc
```diff
@@ -0,0 +1,954 @@
...
+    MY_ASSERT_UNREACHABLE();
+    return 1;
+  }
+  to->round(current_thd, end_prec, to->default_round_mode(current_thd));
+
+  return 0;
+}
+
+int Sec6_to_interval(const Sec6 &sec6, Interval *to,
+                    enum interval_type itype, uint8 start_prec, uint8 end_prec) {
```
DaveGosselin-MariaDB, line 382: Opening curly brace on new line.

sql/sql_time.h
```diff
@@ -32,6 +32,14 @@ typedef struct st_known_date_time_format KNOWN_DATE_TIME_FORMAT;
 #define WEEK_YEAR            2
 #define WEEK_FIRST_WEEKDAY   4
 
+class TemporalAsciiBuffer : public LEX_CSTRING {
```
DaveGosselin-MariaDB, line 35: Please add a comment explaining this class.

sql/sql_type.cc
```diff
@@ -8684,6 +8955,24 @@ Field *Type_handler_datetime2::
...
+  return new (mem_root) Field_interval(
+      rec.ptr(),
+      16,
+      rec.null_ptr(),
+      rec.null_bit(),
+      attr->unireg_check,
+      name,
+      (enum interval_type)f_get_interval_type(attr->pack_flag),
+      attr->length &15,
+attr->length>>4      );
```
DaveGosselin-MariaDB, line 9096: formatting

sql/field.cc
```diff
@@ -7365,8 +7416,244 @@ longlong Field_datetimef::val_datetime_packed(THD *thd)
...
+  my_timeval tm{};
+  interval_to_timeval(&interval, &tm, thd);
+  store_TIMEVAL(tm);
+  return 0;
+}
+
+
+int Field_interval::store_time_dec(const MYSQL_TIME *ltime, uint dec)
+{
+  // How can we convert?
```
DaveGosselin-MariaDB, line 7482: Is this a TODO?

sql/field.cc
```diff
@@ -7365,8 +7416,244 @@ longlong Field_datetimef::val_datetime_packed(THD *thd)
...
+bool Field_interval::validate_value_in_record(THD *thd, const uchar *record) const
+{
+  Interval iv;
+  return get_INTERVAL(&iv, ptr_in_record(record));
+}
+
+
+void Field_interval::sort_string(uchar *to, uint length)
+{
+  to[0]= ptr[4];
```
DaveGosselin-MariaDB, line 7642: Why not the following?
```
DBUG_ASSERT(length == 5);  // some nice comment explaining why here
memcpy(to, ptr, length);
```
or
```
memcpy(to, ptr, 5);  // some nice comment explaining 5
```

sql/item.h
```diff
@@ -7909,6 +8052,73 @@ class Item_cache_timestamp: public Item_cache
...
+     m_native()
+  {}
+
+  Item *do_get_copy(THD *thd) const override
+  { return get_item_copy<Item_cache_interval>(thd, this); }
+  Item *do_build_clone(THD *thd) const override { return get_copy(thd); }
+
+  bool cache_value() override
+  {
+    if (!example) return false;
```
DaveGosselin-MariaDB, line 8071: Our coding style requires this to be formatted as:
```
if (!example)
  return false;
```

sql/item_cmpfunc.cc
```diff
@@ -4085,6 +4125,34 @@ Item *in_temporal::create_item(THD *thd)
   return new (thd->mem_root) Item_datetime(thd);
 }
 
+in_interval::in_interval(THD *thd, uint elements)
+  :in_vector(thd, elements, sizeof(Value), cmp_timestamp, 0)
+{}
+
+
+bool in_interval::set(uint pos, Item *item)
```
DaveGosselin-MariaDB, line 4133: Are these methods (set, get_value, create_item, value_to_item) future TODOs?

sql/sql_yacc.yy
```diff
@@ -132,6 +132,88 @@ static Item* escape(THD *thd)
   return new (thd->mem_root) Item_string_ascii(thd, esc, MY_TEST(esc[0]));
 }
 
+inline const char* interval_error_message(enum interval_type type) {
```
DaveGosselin-MariaDB, line 135: These functions belong in a `.cc` file someplace.  That will likely fix the inconsistent error message handling in the buildbots as well as locally on my machine (and probably other dev machines, too).

HazemRawi: Okay, I'll try to figure out how to move it to a .cc file while making sure it still does its job

sql/item_cmpfunc.cc
HazemRawi: Yes, but properly implementing it would be preferable once I understand its functionality. It is vague to implement some functions without writing a simple query to test them

sql/item.h
HazemRawi: Understood. I’ll review the entire PR and double-check everything

sql/field.cc
HazemRawi: Hmm, okay, but for now I haven't come up with sql query to trigger this function to execute. I just implemented it in the same way as its sisters for other data types

sql/field.cc
HazemRawi: Additionally, I wanted to ask if we can insert a MYSQL_TIME value into an INTERVAL column by first converting it into an interval calculated from the epoch (1970-01-01)

sql/sql_time.h
HazemRawi: This class was previously only available in sql_time.cc. I've made it public and moved its implementation to sql_time.h to make it accessible for use within sql_interval.cc

sql/field.cc
HazemRawi: If you mean placing MYSQL_TYPE_INTERVAL at the end of the enumeration, after MYSQL_TYPE_GEOMETRY = 255, then it wouldn’t be a problem to put it there. In fact, there were only two possible positions: either right after MYSQL_TYPE_GEOMETRY, or in the place where I have already added it. I didn’t place it after MYSQL_TYPE_GEOMETRY because that would assign it the value 256, but I preferred to keep it within the 0–255 range

sql/field.cc
DaveGosselin-MariaDB: In the case that the user wants to put a MYSQL_TIME value into an INTERVAL column, I think we should require the user to use the `CAST` function.

PR #4113, opened 2025-06-13
MENT-2297: Inconsistency detected - create sequence

hemantdangi-gc:
Issue:
The Create SEQUENCE object on the first/active galera node: CREATE SEQUENCE seq_moni_num start with 1 minvalue 1 maxvalue 9223372036854775806 increment by 0 cache 1000 nocycle ENGINE=InnoDB

Execute the following query for two or three times on the other nodes: SELECT NEXT VALUE FOR seq_moni_num;

Resulted in below error on applier node:
[ERROR] Slave SQL: Could not execute Write_rows_v1 event on table monitor.seq_moni_num; Unknown error, Error_code: 1105; handler error No Error!; the event's master log FIRST, end_log_pos 0, Internal MariaDB error code: 1105 [Warning] WSREP: Event 3 Write_rows_v1 apply failed: 195, seqno 28222

Solution:
The error HA_ERR_SEQUENCE_INVALID_DATA 195 happens in sequence_definition::check_and_adjust() but not enough log are provided to know which limit reached or condition failed. So more logs
 are added on failure in sequence_definition::check_and_adjust().

## Description
TODO: fill description here

## Release Notes

## How can this PR be tested?

## PR quality check


sql/sql_sequence.cc
```diff
@@ -133,6 +133,12 @@ bool sequence_definition::check_and_adjust(bool set_reserved_until)
        (real_increment < 0 && reserved_until <= max_value)))
     DBUG_RETURN(FALSE);
 
+  sql_print_error("Check failed, sequence is invalid. "
+                  "max_value: %lld min_value: %lld start: %lld "
+                  "cache: %lld real_increment: %lld max_increment: %lld "
+	          "reserved_until: %lld next_free_value: %lld.",
+                  max_value, min_value, start, cache, real_increment,
+                  max_increment, reserved_until, next_free_value);
```
janlindstrom: Move this before above if-clause (and reword as it will be only information)

sql/sql_sequence.cc
```diff
@@ -372,6 +378,15 @@ bool sequence_insert(THD *thd, LEX *lex, TABLE_LIST *org_table_list)
 
   seq->reserved_until= seq->start;
   error= seq->write_initial_sequence(table);
+  if (seq->max_value == 0 || seq->min_value == 0)
```
janlindstrom: Here also, can we unconditionally write information, this build will not be used on production long.

sql/sql_sequence.cc
```diff
@@ -1064,6 +1079,17 @@ bool Sql_cmd_alter_sequence::execute(THD *thd)
   else
     table->file->print_error(error, MYF(0));
   seq->write_unlock(table);
+  if (seq->max_value == 0 || seq->min_value == 0)
```
janlindstrom: Remove condition.

janlindstrom (changes requested):
Please add instrumentation I requested on slack and make these unconditional.

sql/sql_sequence.cc
hemantdangi-gc: Done

sql/sql_sequence.cc
hemantdangi-gc: Done

sql/sql_sequence.cc
hemantdangi-gc: Done

mysql-test/main/backup_locks.test
```diff
@@ -8,6 +8,7 @@
 --source include/not_embedded.inc
 --source include/no_view_protocol.inc
 
+
```
janlindstrom: Remove these unnecessary line additions.

sql/ha_sequence.cc
```diff
@@ -259,7 +263,16 @@ int ha_sequence::write_row(const uchar *buf)
                                              lock_wait_timeout))
         DBUG_RETURN(ER_LOCK_WAIT_TIMEOUT);
 
+    sql_print_information("write_row read_fields ");
     tmp_seq.read_fields(table);
+
+#ifdef WITH_WSREP
```
janlindstrom: This is not a good idea. Instead why not add sync point and
(1) Start local sequence operation that will block after it has taken exclusive lock but before it has replicated, maybe local commit monitor, wait until it has on there ?
(2) Start remote sequence operation and see that it will report error and retry
(3) release local sync point

sql/ha_sequence.cc
```diff
@@ -295,6 +308,7 @@ int ha_sequence::write_row(const uchar *buf)
 
   if (likely(!(error= file->update_first_row(buf))))
   {
+    sql_print_information("write_row update_first_row");
```
janlindstrom: Remove these extra info.

janlindstrom (changes requested):
Test case does not represent what customer is hitting, in customer case some local is holding lock on sequence concurrently on applier executing sequence operation that fails.

sql/ha_sequence.cc
hemantdangi-gc: Added testcases as you suggested but exclusive locks taken in one query, the other query will fail with  ER_LOCK_DEADLOCK and ER_LOCK_WAIT_TIMEOUT.

sql/ha_sequence.cc
hemantdangi-gc: will remove after resolution of new exclusive lock galera_3nodes.MENT-2297-local and galera_3nodes.MENT-2297 testcases.

mysql-test/main/backup_locks.test
hemantdangi-gc: done

hemantdangi-gc:
> Test case does not represent what customer is hitting, in customer case some local is holding lock on sequence concurrently on applier executing sequence operation that fails.

If customer is holding lock here then it would not have failed with min_value, max_value and other sequence variables as 0. It will fail with ER_LOCK_DEADLOCK and ER_LOCK_WAIT_TIMEOUT as shown in galera_3nodes.MENT-2297-local and galera_3nodes.MENT-2297.

The issue is 'SELECT NEXT VALUE FOR seq_moni_num',  on applier node reads min_value, max_value and other sequence variables as 0. This can happen because value returned in sequence_definition::read_fields() -> Item_func_nextval::val_int() can be 0 on errors which is not handled in current code. So on certain errors retry read and/or transaction retry.

sql/ha_sequence.cc
janlindstrom: Very good, please add that test case also when this PR is ready.

janlindstrom: It would be good to know that is the actual error that happens here.

PR #4114, opened 2025-06-13
MDEV-36107: Add expression support to mysqltest

KhaledR57:
- [x] *The Jira issue number for this PR is: MDEV-36107*

## Description
The `let` command can now evaluate mathematical and logical expressions.

## Release Notes
This patch integrates a recursive descent parser to handle arithmetic, comparisons, and logical operators.

TODO: Still working on it
To maintain backward compatibility, a heuristic function first checks if the right-hand side of an assignment looks like an expression. If not, the original string assignment logic is used.

## How can this PR be tested?

`
$ ./mtr main.mysqltest_expression_evaluation
`

## PR quality check

### DIFF
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -666,6 +666,16 @@ int multi_reg_replace(struct st_replace_regex* r,char* val);
 void free_win_path_patterns();
 #endif
 
+/* For expression parsing */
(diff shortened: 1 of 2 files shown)


sanja-byelkin:
I merged the commit in 10.6

PR #1624, opened 2020-07-15
MDEV-10267 Add "ngram" support to MariaDB

i-rinat:
Ngram is a fulltext parser plugin that splits words into overlapping segments of fixed lengths. For example, if 3-grams are in use, string "abcdef" is split into "abc", "bcd", "cde", "def", thus allowing an efficient substring search for CJK texts where splitting by words in not feasible. Only word characters are going into n-grams. Spaces and punctuation characters are treated as separators.

PR #4118, opened 2025-06-16
MDEV-10267/pre: move fulltext example plugin into fulltext/example

sitano:
follow up to https://github.com/MariaDB/server/pull/1624.

the patch moves fulltext example plugin to the example directory to synchronize plugin dirs structure with [MySQL](https://github.com/mysql/mysql-server/tree/trunk/plugin/fulltext) before proceeding with ngram parser plugin.


sitano:
The failed CI builds should be attributed to flaky tests from main, as my changes have nothing to do with the failed cases. I am not sure what is the proper way to restart CI failed builds.

```
6230:main.mysqladmin                          w20 [ fail ]
10671:innodb.innodb-lock-inherit-read_commited w26 [ fail ]
10665:main.lock_sync                           w13 [ fail ]  timeout after 900 seconds
```

vuvova:
I don't see a point. Not even sure our `plugin.cmake` supports deeply nested directory structure for plugins. Flat directory structure, as now, is fine.

PR #4123, opened 2025-06-17
MDEV-36650  Unexpected checkpoint in the test innodb.doublewrite

Thirunarayanan:
- [x] *The Jira issue number for this PR is: MDEV-36650*

## Description
innodb.doublewrite: Added sleep to ensure page cleaner thread wake up from my_cond_wait

## How can this PR be tested?
Tested in s390x-ubuntu platform and no failure

Added the following code:
```
set global innodb_log_checkpoint_now=1;              
--let CLEANUP_IF_CHECKPOINT=XA COMMIT 'x';drop table t1;                
--source ../include/no_checkpoint_end.inc
```

After adding the above code, InnoDB gracefully skips the test case.
```
innodb.doublewrite '32k,strict_crc32'    [ skipped ]  Extra checkpoint 1 after 61047 (77875,77842)
innodb.doublewrite '64k,strict_crc32'    [ skipped ]  Extra checkpoint 1 after 61048 (94277,94228)
innodb.doublewrite '32k,strict_full_crc32' [ skipped ]  Extra checkpoint 1 after 60901 (77729,77696)
innodb.doublewrite '64k,strict_full_crc32' [ skipped ]  Extra checkpoint 1 after 60902 (94082,94131)
```

## PR quality check

### DIFF
--- a/mysql-test/suite/innodb/t/doublewrite.test
+++ b/mysql-test/suite/innodb/t/doublewrite.test
@@ -60,7 +60,10 @@ connection default;
 
 flush table t1 for export;
 
---let CLEANUP_IF_CHECKPOINT=drop table t1, unexpected_checkpoint;
+# If we are skipping the test at this point due to an unexpected
+# checkpoint then page cleaner could be active after reading the
+# initial checkpoint information
+--let CLEANUP_IF_CHECKPOINT=XA COMMIT 'x'; drop table t1;
 --source ../include/no_checkpoint_end.inc
 
 --copy_file $MYSQLD_DATADIR/ibdata1 $MYSQLD_DATADIR/ibdata1.bak


mysql-test/suite/innodb/t/doublewrite.test
```diff
@@ -49,6 +49,9 @@ FROM information_schema.global_status
 WHERE variable_name = 'INNODB_BUFFER_POOL_PAGES_DIRTY';
 --source include/wait_condition.inc
 SET GLOBAL innodb_max_dirty_pages_pct=99;
+# Ensure that buf_flush_page_cleaner() has woken up from its
+# my_cond_timedwait() and gone idle.
+sleep 1;
```
svoj: Meh, it doesn't really ensure anything. Nothing guarantees that awakening from timedwait lasts less than a second. No sleeps in tests please, they're unreliable and slowdown tests.

mariadb-DebarunBanerjee: I agree, we should avoid sleep in tests.

@Thirunarayanan Can you please provide more details on how waking up page cleaner helps here ?

I see following error in log of one of the test failure. Do we know the exact reason for the lock wait timeout here ?
> 'drop table t1, unexpected_checkpoint' failed: ER_LOCK_WAIT_TIMEOUT (1205): Lock wait timeout exceeded; try restarting transaction

Thirunarayanan: since we don't have unexpected_checkpoint table, we're making test case to fail if we encounter any checkpoint. But LOCK_WAIT_TIMEOUT error happens when the  drop table tries to drop the table when XA PREPARE helds the MDL on the table. 

```
# If we are skipping the test at this point due to an unexpected                
# checkpoint then page cleaner could be active after reading the                
# initial checkpoint information                                                
--let CLEANUP_IF_CHECKPOINT=XA COMMIT 'x'; drop table t1; 
```

Thirunarayanan: We could avoid this error and skip the test case gracefully. In release build, there is no other way to control page cleaner thread. Anyway, this checkpoint happens rarely in few builds. I think it is a safer to skip the test if we encounter any unexpected checkpoint

mariadb-DebarunBanerjee: Thanks @Thirunarayanan. Yes, the intent of the test was to gracefully exit but I guess the rare scenario (checkpoint) is never tested. This is in right direction.

PR #4124, opened 2025-06-17
MDEV-36389 Incorrect query results for an indexed text column

DaveGosselin-MariaDB:
Fixes a scenario where an IN subquery returned the wrong result because the pushed WHERE clause was not retained for downstream result filtering.  For example:
  CREATE TABLE t1 (c1 TEXT, UNIQUE (c1(1)));
  INSERT INTO t1 (c1) VALUES ('a');
  SELECT 'abc' IN (SELECT c1 FROM t1);
Internally, he 'abc' IN subquery condition becomes the constant condition:
  'abc' = t1.c1 or t1.c1 is null
Prior to this patch, this condition was incorrectly removed when converting the subquery engine to an index lookup-based engine. Now eligible conditions are preserved during such engine rewrites.


mysql-test/main/subselect.result
```diff
@@ -991,9 +991,9 @@ a	t1.a in (select t2.a from t2)
 explain extended SELECT t1.a, t1.a in (select t2.a from t2) FROM t1;
 id	select_type	table	type	possible_keys	key	key_len	ref	rows	filtered	Extra
 1	PRIMARY	t1	index	NULL	PRIMARY	4	NULL	4	100.00	Using index
-2	SUBQUERY	t2	index_subquery	a	a	5	func	2	100.00	Using index
+2	DEPENDENT SUBQUERY	t2	index_subquery	a	a	5	func	2	100.00	Using index; Using where
 Warnings:
-Note	1003	/* select#1 */ select `test`.`t1`.`a` AS `a`,<expr_cache><`test`.`t1`.`a`>(<in_optimizer>(`test`.`t1`.`a`,<exists>(<index_lookup>(<cache>(`test`.`t1`.`a`) in t2 on a checking NULL having `test`.`t2`.`a` is null)))) AS `t1.a in (select t2.a from t2)` from `test`.`t1`
+Note	1003	/* select#1 */ select `test`.`t1`.`a` AS `a`,<expr_cache><`test`.`t1`.`a`>(<in_optimizer>(`test`.`t1`.`a`,<exists>(<index_lookup>(<cache>(`test`.`t1`.`a`) in t2 on a checking NULL where <cache>(`test`.`t1`.`a`) = `test`.`t2`.`a` or `test`.`t2`.`a` is null having `test`.`t2`.`a` is null)))) AS `t1.a in (select t2.a from t2)` from `test`.`t1`
 CREATE TABLE t3 (a int(11) default '0');
```
spetrunia, line 997: It's interesting that  t1.a=t2.a is not remove here now.  both are integers... are there cases left where the WHERE removal code can still remove the conditions?

sql/opt_subselect.cc
```diff
@@ -5801,9 +5805,14 @@ static Item *remove_additional_cond(Item* conds)
   Remove the predicates pushed down into the subquery
 
   SYNOPSIS
-    remove_subq_pushed_predicates()
-      where   IN  Must be NULL
-              OUT The remaining WHERE condition, or NULL
+    remove_subq_pushed_predicates_impl()
+
+    @where   IN  If not NULL and predicate eligible for removal, this function
```
spetrunia, line 5810: `@where` ? I think the right doxygen format is `@param` ...

The relationship between `item` and `*where` is not clear.  does `item==*where` or not?

mysql-test/main/unique.test
```diff
@@ -23,3 +23,39 @@ DELETE FROM t2 WHERE n = 1;
 DROP TABLE t2;
 
 --source include/test_db_charset_restore.inc
+
+#
+# Start of 11.8 tests
+#
+
```
spetrunia, line 30: Do these tests add any value? 

I've tried this:

Get the commit before this patch and compile:
```
git checkout 63cbca3fba053b8fb291f7dd7f20adf877bf5869
make 
```
Then, get this patch so that tests are updated:
```
git checkout f4833ceb4133099a501605c9688e34259a0aaa97
``` 
and do NOT compile.
Now, run
```
./mtr unique 
```
and it still passes ....

spetrunia: Note that running other tests, like
```
./mtr subselect
```
predictably fails.

spetrunia: Another question:  Is `unique.test` the right place to put the tests?  I think the problem is not in unique indexes or their handling but rather in the subquery optimizations.

DaveGosselin-MariaDB: I sometimes add tests as a form of documentation, to show expected behaviors.  Yes, in this case the tests pass with or without the changes, but they illustrate some aspect of how we expect the system to behave at this point in time.  I don't think we should remove them, and I'm not sure where else they should live.

sql/opt_subselect.cc
DaveGosselin-MariaDB: Fixed both doxygen to be `@param where` and clarified comment.

spetrunia:
This PR is against 11.8, while MDEV-36389 has fixVersion starting from 11.4.  @DaveGosselin-MariaDB , is this intentional?

DaveGosselin-MariaDB:
> This PR is against 11.8, while MDEV-36389 has fixVersion starting from 11.4. @DaveGosselin-MariaDB , is this intentional?

The "Affects Version" is 11.8, so I assumed that 11.4 was a mistake in the "Fixed Version" field.  I'll check 11.4 and backport if necessary.

spetrunia:
One other observation:  There are plenty of changes like this:

image

the  `SUBQUERY` -> `DEPENDENT SUBQUERY` looks odd.  I would say, the original `SUBQUERY` one is incorrect.  For the shown query, `index_subquery` does use the value of `t2.a`.   How could it be not `DEPENDENT SUBQUERY` ?  


(The interpretation of `DEPENDENT SUBQUERY` vs `SUBQUERY` that I know and am using is :  `SUBQUERY` is computed once as it doesn't depend on outside values.  `DEPENDENT SUBQUERY` is, well, dependent and so is recomputed many times).

I don't think we can fix anything in this fix here.  Perhaps this should be mentioned in the commit comment.

DaveGosselin-MariaDB:
> This PR is against 11.8, while MDEV-36389 has fixVersion starting from 11.4. @DaveGosselin-MariaDB , is this intentional?

@spetrunia this issue was introduced by git commit `89f5d28191a8ecf7a1511de22d00538b063acda7` in 11.8.  Testing shows that 11.4 is unaffected.  Please see my analysis in [this Jira](https://jira.mariadb.org/browse/MDEV-36389?focusedCommentId=306552&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-306552) comment.  I re-tested 11.4 to be sure.

sql/opt_subselect.cc
DaveGosselin-MariaDB: Verified that `item == *where` and added an assertion in `remove_subq_pushed_predicates`:
```
  DBUG_ASSERT(!where || !*where || join->conds == *where);
```
Manually verified that in cases when `join->conds` is set that indeed it points to `*where`.

mysql-test/main/subselect.result
DaveGosselin-MariaDB: In this case, `t1.a = t2.a` internally is `<cache>(t1.a) = t2.a` and the first argument to the equality is a `CACHE_ITEM`.  The logic in `remove_subq_pushed_predicates_impl` requires that the first argument of the equality is a `REF_ITEM`.  Allowing the first argument to be either a ref or a cache item still fails the `test_if_ref` check because `part_of_refkey` returns NULL.   `test_if_ref` makes it to this point
```
    if (join_tab->ref.is_access_triggered())
      return FALSE;

-->Item *ref_item=part_of_refkey(field->table,field);
    if (ref_item && (ref_item->eq(right_item,1) || ......
```
to call `part_of_refkey`.  While we can find the field in the key within `part_of_refkey`, this conditional is false:
```
        /*
          Found the field in the key. Check that 
           1. ref_or_null doesn't alternate this component between a value and
              a NULL
           2. index fully covers the key
        */
-->        if (part != join_tab->ref.null_ref_part &&            // (1)
                !(key_part->key_part_flag & HA_PART_KEY_SEG))     // (2)
```

DaveGosselin-MariaDB:
Closing this PR in favor of #4344 because this issue affects 10.11 and should be delivered against that version and merged up later.

PR #4126, opened 2025-06-18
MDEV-35913 Assertion `m_comparator.cmp_type() != ROW_RESULT' in Item_…

DaveGosselin-MariaDB:
…func_in

Check earlier that a column is a virtual column, before checking that all arguments are constants, to avoid this assertion in the case of non-virtual columns.

### DIFF
--- a/mysql-test/main/type_row.test
+++ b/mysql-test/main/type_row.test
@@ -88,3 +88,22 @@ SET sql_mode=DEFAULT;
 --echo #
 --echo # End of 10.7 tests
 --echo #
+
+--echo #
+--echo # Start of 11.8 tests
+--echo #
+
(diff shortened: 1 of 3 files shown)


mysql-test/main/type_row.test
```diff
@@ -88,3 +88,17 @@ SET sql_mode=DEFAULT;
 --echo #
 --echo # End of 10.7 tests
 --echo #
+
+--echo #
+--echo # Start of 11.8 tests
```
spetrunia, line 93: Please also add a `--echo # MDEV-nnnn: .....` comment.

spetrunia (approved):
Ok to push after making the change.

